PR #2802, opened 2017-09-18
Add Gaussian filter-based antialiasing to resize

ahojnnes:
Previously, we only provided functions to resize/rescale images with
pure interpolation. This results in aliasing effects when reducing the
image size. The new functions properly account for this effect by
applying a Gaussian filter prior to interpolation.

## For reviewers

  later.

### DIFF
--- a/doc/examples/transform/plot_rescale.py
+++ b/doc/examples/transform/plot_rescale.py
@@ -10,30 +10,27 @@ along each axis.
 `Resize` serves the same purpose, but allows to specify an output image shape
 instead of a scaling factor.
 
-`Downscale` operation serves the purpose of downsampling an n-dimensional image
-by Gaussian smoothing and by calculating local mean on the elements of each
-block of the size factors given as a parameter to the function.
+`Downscale` serves the purpose of downsampling an n-dimensional image by
+calculating local mean on the elements of each block of the size factors given
+as a parameter to the function.
 
-Note that when downsampling an image, `resize` and `rescale` might produce
-aliasing effects since they are not appropriately resampling the input image.
-Please, refer to the `downsize`, `downscale`, and `downscale_local_mean`
-functions in this case.
+Note that when downsampling an image, `resize` and `rescale` should perform
+Gaussian smoothing to avoid aliasing artifacts.
 
 """
 
 import matplotlib.pyplot as plt
 
 from skimage import data
-from skimage.transform import rescale, resize, downscale, downscale_local_mean
+from skimage.transform import rescale, resize, downscale_local_mean
 
 image = data.camera()
 
 image_rescaled = rescale(image, 0.5)
 image_resized = resize(image, (400, 400), mode='reflect')
-image_downscaled = downscale(image, 2)
-image_downscaled_local_mean = downscale_local_mean(image, (2, 2))
+image_downscaled = downscale_local_mean(image, (2, 2))
 
-fig, axes = plt.subplots(nrows=3, ncols=2, sharex=True, sharey=True)
+fig, axes = plt.subplots(nrows=2, ncols=2, sharex=True, sharey=True)
 
 ax = axes.ravel()
 
@@ -47,10 +44,7 @@ ax[2].imshow(image_resized, cmap='gray')
 ax[2].set_title("Resized image")
 
 ax[3].imshow(image_downscaled, cmap='gray')
-ax[3].set_title("Downscaled image using Gaussian smoothing")
-
-ax[4].imshow(image_downscaled_local_mean, cmap='gray')
-ax[4].set_title("Downscaled image using local averaging")
+ax[3].set_title("Downscaled image")
 
 ax[0].set_xlim(0, 512)
 ax[0].set_ylim(512, 0)
--- a/doc/examples/transform/plot_rescale.py
+++ b/doc/examples/transform/plot_rescale.py
@@ -3,31 +3,37 @@
 Rescale, resize, and downscale
 ==============================
 
-`Rescale` operation resizes an image by a given scaling factor.
-The scaling factor can either be a single floating point value,
-or multiple values - one along each axis.
+`Rescale` operation resizes an image by a given scaling factor. The scaling
+factor can either be a single floating point value, or multiple values - one
+along each axis.
 
-`Resize` serves the same purpose, but allows to specify an output
-image shape instead of a scaling factor.
+`Resize` serves the same purpose, but allows to specify an output image shape
+instead of a scaling factor.
 
-`Downscale` operation serves the purpose of downsampling an
-n-dimensional image by calculating local mean on the elements of
-each block of the size factors given as a parameter to the function.
+`Downscale` operation serves the purpose of downsampling an n-dimensional image
+by Gaussian smoothing and by calculating local mean on the elements of each
+block of the size factors given as a parameter to the function.
+
+Note that when downsampling an image, `resize` and `rescale` might produce
+aliasing effects since they are not appropriately resampling the input image.
+Please, refer to the `downsize`, `downscale`, and `downscale_local_mean`
+functions in this case.
+
 """
 
 import matplotlib.pyplot as plt
 
 from skimage import data
-from skimage.transform import rescale, resize, downscale_local_mean
+from skimage.transform import rescale, resize, downscale, downscale_local_mean
 
 image = data.camera()
 
 image_rescaled = rescale(image, 0.5)
 image_resized = resize(image, (400, 400), mode='reflect')
-image_downscaled = downscale_local_mean(image, (2, 3))
+image_downscaled = downscale(image, 2)
+image_downscaled_local_mean = downscale_local_mean(image, (2, 2))
 
-fig, axes = plt.subplots(nrows=2, ncols=2,
-                         sharex=True, sharey=True)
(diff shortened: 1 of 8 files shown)


skimage/transform/_warps.py
```diff
@@ -325,6 +354,162 @@ def rotate(image, angle, resize=False, center=None, order=1, mode='constant',
                 mode=mode, cval=cval, clip=clip, preserve_range=preserve_range)
 
 
+def downsize(image, output_shape, sigma=None, order=1, mode="reflect", cval=0,
```
stefanv: Should this not become the default for resize and rescale?

ahojnnes: Not sure I understand the question correctly. Are you referring to all parameters or a specific one?

stefanv: Sorry, I meant the Gaussian blurring should eventually be the default (I think it's the "correct" way). As such, would it not make sense to have a binary flag added to the existing downsampling functions, instead of adding another two with vaguely similar names?

ahojnnes: Makes sense. This way, we could also smooth the image after interpolation when upsampling. Not sure about the recent policies in scikit-image but deprecation path is still with setting a parameter to None and warning the user?

stefanv: Yes, still: https://github.com/scikit-image/scikit-image/blob/master/CONTRIBUTING.txt#L383

Glad to have you back for a bit :)

skimage/transform/_warps.py
```diff
@@ -74,6 +74,13 @@ def resize(image, output_shape, order=1, mode=None, cval=0, clip=True,
     preserve_range : bool, optional
         Whether to keep the original range of values. Otherwise, the input
         image is converted according to the conventions of `img_as_float`.
+    anti_aliasing : bool
+        Whether to apply a Gaussian filter to smooth the image prior to
+        down-scaling. It is crucial to filter when resampling the image to
+        avoid aliasing artifacts.
+    anti_aliasing_stddev : {float, tuple of floats}
+        Standard deviation for Gaussian filtering to avoid aliasing artifacts.
+        By default, this value is chosen as 0.25 times the scale factor.
```
stefanv: Is this just based on intuition, or did you find a paper to guide the size of the Gaussian?

skimage/transform/_warps.py
```diff
@@ -97,26 +104,48 @@ def resize(image, output_shape, order=1, mode=None, cval=0, clip=True,
...
+        if anti_aliasing_stddev is None:
+            anti_aliasing_stddev = np.maximum(0, (factors - 1) / 2)
+            anti_aliasing_stddev[factors <= 1] = 0
+        else:
+            anti_aliasing_stddev = \
+                np.atleast_1d(anti_aliasing_stddev) * np.ones_like(factors)
+            if np.any(anti_aliasing_stddev < 0):
+                raise ValueError("Anti-aliasing standard deviation must be "
+                                 "greater than zero")
+            elif np.any((anti_aliasing_stddev > 0) & (factors <= 1)):
```
stefanv: What is this check doing?

doc/examples/transform/plot_rescale.py
```diff
@@ -3,16 +3,20 @@
...
-image shape instead of a scaling factor.
+`Resize` serves the same purpose, but allows to specify an output image shape
+instead of a scaling factor.
+
+`Downscale` serves the purpose of downsampling an n-dimensional image by
+calculating local mean on the elements of each block of the size factors given
+as a parameter to the function.
+
+Note that when downsampling an image, `resize` and `rescale` should perform
+Gaussian smoothing to avoid aliasing artifacts.
```
jni: It's not clear from this whether the users should do this manually, or whether we provide a function to do so. Perhaps add the sentence, "See the `anti_aliasing` and `anti_aliasing_stddev` arguments to these functions.

And perhaps one of the calls in the example should use the new kwargs?

skimage/transform/_warps.py
```diff
@@ -74,6 +74,13 @@ def resize(image, output_shape, order=1, mode=None, cval=0, clip=True,
     preserve_range : bool, optional
         Whether to keep the original range of values. Otherwise, the input
         image is converted according to the conventions of `img_as_float`.
+    anti_aliasing : bool
+        Whether to apply a Gaussian filter to smooth the image prior to
+        down-scaling. It is crucial to filter when resampling the image to
+        avoid aliasing artifacts.
+    anti_aliasing_stddev : {float, tuple of floats}
```
jni: My personal preference here would be `anti_aliasing_sigma`, because sigma is such a common parameter name for Gaussian filtering. But I'm not set on this, just a suggestion.

skimage/transform/_warps.py
jni: For some combination of parameters, the image could be upscaled in one dimension but downscaled in another. If the user provides their own anti-aliasing sigma as a scalar, the image will be blurred along both the downscaled and upscaled dimensions. I was worried about this as I made my way through the implementation, so it was good to see this check! =)

skimage/transform/_warps.py
```diff
@@ -97,26 +104,48 @@ def resize(image, output_shape, order=1, mode=None, cval=0, clip=True,
...
 
-    input_shape = np.asarray(input_shape, dtype=float)
-    dim_scales = input_shape / np.asarray(output_shape)
+    factors = (np.asarray(input_shape, dtype=float) /
+               np.asarray(output_shape, dtype=float))
+
+    if anti_aliasing:
+        if anti_aliasing_stddev is None:
+            anti_aliasing_stddev = np.maximum(0, (factors - 1) / 2)
+            anti_aliasing_stddev[factors <= 1] = 0
```
jni: Isn't this line redundant with the one above it?

skimage/transform/_warps.py
```diff
@@ -97,26 +104,48 @@ def resize(image, output_shape, order=1, mode=None, cval=0, clip=True,
...
+        else:
+            anti_aliasing_stddev = \
+                np.atleast_1d(anti_aliasing_stddev) * np.ones_like(factors)
+            if np.any(anti_aliasing_stddev < 0):
+                raise ValueError("Anti-aliasing standard deviation must be "
+                                 "greater than zero")
```
jni: greater than or equal to

skimage/transform/_warps.py
ahojnnes: @jni Described this properly.

skimage/transform/_warps.py
ahojnnes: FreeImage uses the same convention and I checked the results on some examples. I couldn't find any reference that suggests a theoretically proven value. If you have a better suggestion, I'd be happy to incorporate it. If you cannot find a good reference either, I'd appreciate if you could also play around with the parameter and check the results. This is important enough to be checked by more than one person.

doc/examples/transform/plot_rescale.py
ahojnnes: Done.

doc/examples/transform/plot_rescale.py
```diff
@@ -24,10 +28,9 @@
 
 image_rescaled = rescale(image, 0.5)
 image_resized = resize(image, (400, 400), mode='reflect')
-image_downscaled = downscale_local_mean(image, (2, 3))
+image_downscaled = downscale_local_mean(image, (2, 2))
```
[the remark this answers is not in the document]
ahojnnes: This happened accidentally. Reverted.

skimage/transform/_warps.py
ahojnnes: Agreed, done.

skimage/transform/_warps.py
ahojnnes: Indeed, fixed.

skimage/transform/_warps.py
ahojnnes: Fixed.

skimage/transform/_warps.py
stefanv: Glad to have more competent brains than my own on the problem :)

stefanv:
Please remember to update the TODO and add the deprecation to the release notes.

jni:
... And it looks like the travis failures are related to the PR. Not sure why the AppVeyor build didn't find the same errors!

Specifically, the calls to resize inside the Gaussian and Laplacian pyramids need to be updated so that they don't raise a warning.

skimage/transform/_warps.py
```diff
@@ -74,6 +74,14 @@ def resize(image, output_shape, order=1, mode=None, cval=0, clip=True,
     preserve_range : bool, optional
         Whether to keep the original range of values. Otherwise, the input
         image is converted according to the conventions of `img_as_float`.
+    anti_aliasing : bool
```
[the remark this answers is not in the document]
ahojnnes: Fixed.

ahojnnes:
Thanks for the feedback. This should be ready to merge. Maybe someone give my default choice for the standard deviation a sanity check. I think the results look good and compare well with the downscale_locale_mean results, which should do the optimal thing for integer scaling factors.

stefanv:
Thanks, this is extremely useful.

stefanv:
BTW, I did try to find a published method for determining kernel size, but didn't come up with any.

ahojnnes:
Thanks, me neither. I compared against several existing other libraries and the results look very similar, so the default choice should be good enough.

Issue #2804, opened 2017-09-20
Hough Transform: hough_line_peaks error


seheatwole:
In developing an image processing script scikit-image(0.13.0-2) that grabs images off a camera and detects lines using the hough_transform with hough_line_peaks, an error appeared related to a lack of lines in the image.  If there is no output from _prominent_peaks() on line 62 of hough_transform.py, the following error occurs:

Traceback (most recent call last):
  File "theodoliteAlign.py", line 82, in <module>
    a, angle, dist = hough_line_peaks(h,theta,d)
  File "/usr/lib/python3.6/site-packages/skimage/transform/hough_transform.py", line 66, in hough_line_peaks
    return (h, angles[a], dists[d])
IndexError: arrays used as indices must be of integer (or boolean) type

This is due to the fact that h, a, d returned from _prominent_peaks are empty matrices and a and d are being used as indexes for the function return.  This happens when grabbing a blank image but previous versions I have used of scikit-image (0.12.3-8) had a different algorithm that didn't fail on blank images.

stefanv:
Thanks for catching that; do you mind submitting a pull request for a fix?  We'd be happy to guide you through that process.

seheatwole:
Sure.  I am new to GitHub but not git.  How do I submit a pull request for
a fix?

stefanv:
Thank you!  Please take a look at this document: http://scikit-image.org/docs/stable/contribute.html

Let me know if anything is unclear.

PR #2805, opened 2017-09-21
Return empty array if hough_line_peaks detects nothing

seheatwole:
## Description
…w errors

## Checklist
- [ ] Unit tests

## References
Closes https://github.com/scikit-image/scikit-image/issues/2804

## For reviewers

  later.

### DIFF
--- a/skimage/transform/hough_transform.py
+++ b/skimage/transform/hough_transform.py
@@ -61,7 +61,7 @@ def hough_line_peaks(hspace, angles, dists, min_distance=9, min_angle=10,
                                min_ydistance=min_distance,
                                threshold=threshold,
                                num_peaks=num_peaks)
-    if a:
+    if a.any():
         return (h, angles[a], dists[d])
     else:
         return (np.array([]), np.array([]), np.array([]))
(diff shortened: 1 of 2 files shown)


soupault:
@seheatwole good! Please, add also a test with the case from your issue to https://github.com/scikit-image/scikit-image/blob/master/skimage/transform/tests/test_hough_transform.py . This way we'll be sure that this edge case will be handled properly in the future.

seheatwole:
Sorry, I was swamped at work and couldn't get to this.  You are right, (h,
np.array([]), np.array([])) is more correct.

thrdr:
I think there's a bug in this code.

The bug shows up in the case where  `_prominent_peaks()` detects peaks and the all peaks have angle 0. 

Those are valid results but the `a.any()` test evaluates to false for zero values. So the returned angle and distance values with get replaced with empty arrays.

rfezzani:
Hello @thrdr and thank you for your interset in `skimage`. Can you open a new issue reporting this bug to track it? It would be great If you can come with a failing sample code demonstrating this bug :wink:

pdinnen:
I've opened issue #6187 @rfezzani.

PR #2815, opened 2017-09-28
Restandardize handling of Multi-Image files

### LINKED ISSUE #2811: Inconsistent behaviour of ImageCollection
# Problem
Calling `ImageCollection` with a string file pattern vs. a list of filenames produces different results when using a custom `load_func` to read files not supported by pillow. 

I ran into this while trying to use `open_rasterio` from [xarray](https://github.com/pydata/xarray)  as a load_func for `ImageCollection`, though this will undoubtedly come up with any files not readable with `Image.open` from PIL.

## Examples

This works:
```python
files = glob('*.dat')
ims = ImageCollection(files, load_func=open_rasterio)
ims[0] # prints the returned xarray
```

This does not:
```python
ims = ImageCollection('*.dat', load_func=open_rasterio)
ims[0] #  `IndexError: There are only 0 images in the collection`
```
## Cause

The reason is that with the pattern, `_find_images` sets `_numframes` to 0, since PIL cannot open any of the files (even though the load_func has no problems opening them):
https://github.com/scikit-image/scikit-image/blob/8c79a2c96f6e20bcfe4a3e6e4ea812ca74899b80/skimage/io/collection.py#L159-L169

https://github.com/scikit-image/scikit-image/blob/8c79a2c96f6e20bcfe4a3e6e4ea812ca74899b80/skimage/io/collection.py#L197-L221

This in turn results in an error later when `_numframes` is checked.

maaleske:
## Description
- Fixes #2811 for the described use case.
- Adds simple unit tests for pattern expansion with PIL-readable and non-PIL-readable files.
- Also fixes previously nonsensical string_sort test

## Checklist
- [x] Unit tests

## References
Closes #2811

## For reviewers

  later.


stefanv:
The tests are legitimately failing on Travis-CI for this PR.

maaleske:
I think I resolved the test failures by explicitly passing load_func=None in MultiImage to trigger the old behaviour. I could not find a reason for the special imread definition there, and removing it doesn't seem to affect tests. This feels slightly odd and also makes MultiImage quite redundant (all it does is set the _filename attribute).

The whole thing feel slightly fragile, since even specifying just load_func=imread automatically removes the flattening of multi-frame files. It seems no tests catch this with imread_collection, which does exactly that. I don't what the better way would be. Should there perhaps be an explicit parameter for enabling the flattening behaviour?

stefanv:
My intuition is that we should never flatten.  Each file should map to one image (array).

We can build a special interface to load multi-images and unpack them, but as you suggest, that should be explicit.

maaleske:
I separated all the multi-image functionality from ImageCollection to a inherited class FrameCollection and implemented MultiImage on top of that. Now ImageCollection is completely frame-agnostic, and just sees each frame as its own image, with _files corresponding to the _frame_index. MultiImage is now just a simple passthrough to Framecollection, and I think it could probably be deprecated as it adds no real functionality (as was apparently discussed before: https://github.com/scikit-image/scikit-image/pull/1200). Any comments on the class organization?

There's still some niggles to fix before it's ready for merging:
 - The current _files should probably be renamed to _ids for clarity, since it doesn't consistently refer to files (possibly adding a property which always does?)
 - _frame_index needs to be updated when slicing FrameCollections (was in ImageCollection)
 - FrameCollection does not yet accept any special load_func

maaleske:
I added the `load_func` parameter for FrameCollection, and resolved the slicing behaviour of `_frame_index` by removing it altogether (since it was redundant with `_files` in the returned collection). The original list of searched files can now be accessed as `_searched_files`, which should be helpful for anyone trying to debug further issues with multi-images.

Existing functionality should now work as it was, just with FrameCollection instead of ImageCollection.

stefanv:
Doesn't FrameCollection and MultiImage now do the same when called on a multi-frame image?

maaleske:
@stefanv Yes, FrameCollection is just the existing MultiImage, that now also handles multiple images. MultiImage probably should be deprecated in favor of FrameCollection, with warnings before removing it completely?

stefanv:
Why not just replace MultiImage with the FrameCollection implementation?  Do they have differing APIs?

maaleske:
The API is different in that FrameCollection handles multiple images. I'm not sure what you mean by replacing, do you want to keep the MultiImage class name (even though it now handles multiple files as well)?

stefanv:
I think, since we only added functionality, and since the old calls will still work, we can probably keep the MultiImage name, yes.

maaleske:
I figured it would be more clear to make a new class that explicitly handles any sets of frames, but I guess keeping MultiImage as the name is not too confusing.

stefanv:
Another option is to add an `unpack_frames` keyword to ImageCollection, and deprecate `MultiImage`.

maaleske:
Adding the parameter would make for a lot of branching in ImageCollection. Having the FrameCollection/MultiImage just inherit ImageCollection makes the latter much cleaner.

maaleske:
There is the small API change of MultiImage having the `filename` property which doesn't make sense in a multiple file setting, and would be best removed (which I guess was why I thought about making the separate class in the first place).

stefanv:
OK, then let's keep MultiImage, and just set the filename parameter to the file pattern, for backward compatibility.  We can make it a property, and warn that that property is being deprecated.

stefanv:
Or, just keep it, and have it contain the file pattern :)  Since this is new functionality, no-one should notice.

skimage/io/collection.py
```diff
@@ -377,16 +332,23 @@ def imread_collection(load_pattern, conserve_memory=True):
...
+    """A class containing all frames from multi-frame images.
 
     Parameters
     ----------
-    filename : str
-        The complete path to the image file.
+    load_pattern : str or list
+        Pattern glob or filenames to load. The path can be absolute or
+        relative.  Multiple patterns should be separated by os.pathsep,
+        e.g. '/tmp/work/*.png:/tmp/other/*.jpg'.
```
stefanv: Backticks for code, ``` ``/tmp/work/*.png:/tmp/other/*.jpg`` ```

stefanv: Same for ``` ``os.pathsep`` ``` above

skimage/io/collection.py
```diff
@@ -396,8 +358,14 @@ class MultiImage(ImageCollection):
     The last accessed frame is cached, all other frames will have to be read
     from file.
 
-    The current implementation makes use of ``tifffile`` for Tiff files and
-    PIL otherwise.
+    By default, MultiImage builds a list of frames in each image
+    and calls ``imread(filename, img_num=n)`` to read the n:th frame in
```
stefanv: Some indication that it is being called *for each n*

maaleske:
Done. To summarize the current version as compared to master: 
 - ImageCollection now always deals with single items (files, indices, etc.) set by load_pattern (and possibly load_func) and never flattens images.
 - MultiImage implementation is now inherited from ImageCollection and just passes it the _frame_index (as built previously by _find_images) and a wraps imread or load_func to accept the pairs as input.
 - As a resulting new feature, MultiImage can now handle multiple files with a special load_func (which must then accept the img_num parameter)

stefanv:
This looks great; thanks @maaleske!

stefanv:
@scikit-image/core This PR went through review and has been approved; if another person could take a look, rebase and merge, please.

stefanv:
@scikit-image/core This is ready to go (test failure seems unrelated).

jni:
@stefanv there appears to be a non-trivial merge conflict currently... Could you have a look? I didn't follow what was happening in this PR...

sciunto:
@scikit-image/core We must consider this PR for 0.16, otherwise, it will fall into the cracks.

stefanv:
@emmanuelle Could you take a look at https://github.com/scikit-image/scikit-image/pull/2815/commits/4c8012e63e3c2213112a46ab97f5f620b23866eb? Seems to be where the merge conflict happened.

mkcor:
@maaleske @stefanv @jni @sciunto @emmanuelle I'm trying to rescue this PR: I've resolved merge conflicts using the online editor just so I can pull this branch and it's (formally) clean. I have arbitrated according to my understanding of the PR, but I have to work on it locally to make sure it made sense. Please sync up your local branch if you are working on it as well.

skimage/io/collection.py
```diff
@@ -386,6 +349,52 @@ def imread_collection(load_pattern, conserve_memory=True):
...
+        """Return an `ImageCollection` from files matching the given pattern.
+
+        Note that files are always stored in alphabetical order. Also note that
+        slicing returns a new ImageCollection, *not* a view into the data.
+
+        See `skimage.io.ImageCollection` for details.
+
         Parameters
         ----------
         load_pattern : str or list
```
mkcor, line 400:
```suggestion

```

mkcor: I need to remove lines down to the next occurrence of 'Notes' (something I didn't see well while editing online).

skimage/io/collection.py
```diff
@@ -389,6 +352,39 @@ def imread_collection(load_pattern, conserve_memory=True):
...
+        True.
+
+    Other parameters
+    ----------------
+    load_func : callable
+        ``imread`` by default.  See notes below.
+
             Pattern glob or filenames to load. The path can be absolute or
             relative.  Multiple patterns should be separated by a colon,
             e.g. '/tmp/work/*.png:/tmp/other/*.jpg'.  Also see
```
mkcor, line 390:
```suggestion

```

skimage/io/collection.py
```diff
@@ -391,7 +354,37 @@ def imread_collection(load_pattern, conserve_memory=True):
...
+        True.
+
+    Other parameters
+    ----------------
+    load_func : callable
+        ``imread`` by default.  See notes below.
+
             implementation notes below.
         conserve_memory : bool, optional
             If True, never keep more than one in memory at a specific
```
mkcor, line 390:
```suggestion

```

skimage/io/collection.py
```diff
@@ -403,6 +366,33 @@ def imread_collection(load_pattern, conserve_memory=True):
...
+            time.  Otherwise, images will be cached once they are loaded.
+
+        """
+        return ImageCollection(load_pattern, conserve_memory=conserve_memory,
+                               load_func=imread)
+    return imread_collection
+
+
 class MultiImage(ImageCollection):
     """A class containing a single multi-frame image.
```
mkcor, line 398:
```suggestion

```

skimage/io/collection.py
```diff
@@ -404,8 +367,23 @@ def imread_collection(load_pattern, conserve_memory=True):
...
+        Whether to conserve memory by only caching a single frame. Default is
+        True.
+
+    Other parameters
+    ----------------
+    load_func : callable
+        ``imread`` by default.  See notes below.
     Parameters
     ----------
     filename : str
```
mkcor, line 389:
```suggestion

```

skimage/io/collection.py
```diff
@@ -404,11 +367,23 @@ def imread_collection(load_pattern, conserve_memory=True):
...
+        Whether to conserve memory by only caching a single frame. Default is
+        True.
+
+    Other parameters
+    ----------------
+    load_func : callable
+        ``imread`` by default.  See notes below.
         The complete path to the image file.
     conserve_memory : bool, optional
         Whether to conserve memory by only caching a single frame. Default is
```
mkcor, line 389:
```suggestion

```

skimage/io/collection.py
```diff
@@ -404,16 +367,25 @@ def imread_collection(load_pattern, conserve_memory=True):
...
         True.
 
+    Other parameters
+    ----------------
+    load_func : callable
+        ``imread`` by default.  See notes below.
+        True.
+
```
mkcor, line 388:
```suggestion

```

mkcor:
Sorry for the gross series of 'Remove extra lines' commits... Following my manual resolution of merge conflicts (using GitHub's editor), this extra block shouldn't have been left there (and was breaking the syntax):

```
    Notes
    -----
    If ``conserve_memory=True`` the memory footprint can be reduced, however
        """Return an `ImageCollection` from files matching the given pattern.

        Note that files are always stored in alphabetical order. Also note that
        slicing returns a new ImageCollection, *not* a view into the data.

        See `skimage.io.ImageCollection` for details.

        Parameters
        ----------
        load_pattern : str or list
            Pattern glob or filenames to load. The path can be absolute or
            relative.  Multiple patterns should be separated by a colon,
            e.g. '/tmp/work/*.png:/tmp/other/*.jpg'.  Also see
            implementation notes below.
        conserve_memory : bool, optional
            If True, never keep more than one in memory at a specific
            time.  Otherwise, images will be cached once they are loaded.

        """
        return ImageCollection(load_pattern, conserve_memory=conserve_memory,
                               load_func=imread)
    return imread_collection


class MultiImage(ImageCollection):
    """A class containing a single multi-frame image.

    Parameters
    ----------
    filename : str
        The complete path to the image file.
    conserve_memory : bool, optional
        Whether to conserve memory by only caching a single frame. Default is
        True.
```

mkcor:
I have deleted my comment starting with "Sorry for the gross series of 'Remove extra lines' commits..." -- I thought I wouldn't be able to push from my local branch and, hence, had to work with the online editor (because I couldn't when I wasn't core, but anyway).

mkcor:
I have tried updating this PR in line with #3928 but I'm afraid I have lost the original spirit of this contribution along the way... :disappointed:  There are now 3 failing tests in `test_collection.py`, which were added by the original author, i.e.:

* FAILED io/tests/test_collection.py::TestImageCollection::test_custom_load - T...
    as expected since `load_pattern = [(1, 'one'), (2, 'two')]` is not a string nor a list of strings;

* FAILED io/tests/test_collection.py::TestImageCollection::test_custom_pattern
    indeed, `glob('PREFIX/.cache/scikit-image/master/data/b??ck.png')` results in `['PREFIX/.cache/scikit-image/master/data/brick.png']` and not in `['PREFIX/.cache/scikit-image/master/data/block.png', 'PREFIX/.cache/scikit-image/master/data/brick.png']` as 'expected;'

* FAILED io/tests/test_collection.py::TestImageCollection::test_custom_pattern_not_pil
    where pattern `PREFIX/.cache/scikit-image/master/data/chessboard_*_U8.npz'` doesn't get matched to the expected `['chessboard_GRAY_U8.npz', 'chessboard_RGB_U8.npz']`

Note that function `_find_files` doesn't get used at this point.

@jni could you please have a look? Thanks :grimacing:

mkcor:
Forgot to loop in @alexdesiqueira, especially when referring to #3928 🙏

mkcor:
I realize this PR is three years old, so it might be unrealistic to include it at this point... I thought I would give it a try, since it was tagged with `action: mrg+1` as well as `0.18`.

> Note that function `_find_files` doesn't get used at this point.

Removed in fad090cfa48697229f2f89baae43abeb7c5cbcf2.

>     * FAILED io/tests/test_collection.py::TestImageCollection::test_custom_load - T...
>       as expected since `load_pattern = [(1, 'one'), (2, 'two')]` is not a string nor a list of strings;

Removed in 335be36fab0ea11d21bfbb2662081f1990d3d06a.

maaleske:
Since the pattern parameter was changed, the tests I added are probably outdated since they were meant to check for differences w.r.t. the original behaviour. It's probably best to disregard them and just make sure that you test whatever is needed to check for the issues in #5069 and #2811 (if that one still happens).

mkcor:
@maaleske great to read you!! Sure. Specifically, we want `test_custom_load` to fail, because the current approach is to support only str and list of str patterns. That's why I removed it. But, in my understanding so far, I think we would want `test_custom_pattern` to pass... Good point anyway, about focusing on the issues which are currently open. Thank you!

alexdesiqueira:
@maaleske great to see you around, brother! Hope we can get a beer one of these days.
@mkcor I'm sorry, I didn't see we're working on this at #4859 — I said @jni I would check but it slipped away. Could we check that one instead?

mkcor:
Oh, I wasn't aware of #4859... :woman_facepalming: It's okay: I've spent enough time in this code by now that I would be fast at reworking either PR. Which one of the two PRs do we want to close, @jni? @alexdesiqueira I don't think we want to keep both in parallel, indeed.

My general comment remains the same: My understanding is that we want to remain in the spirit of #3928 and build from there. Notably, we want to accept loading patterns which are either strings or lists of strings (not lists of anything as in the original contribution). Besides, if we keep (as I suggest we do) the handling of multi-pattern brought by #3928, the detection of frames as implemented in the original contribution fails at splitting frames within a given (multi-image) file. Therefore, we can only keep a marginal part of this contribution on the functionality side. But we should keep most tests brought by this contribution (yay!) even if some need adapting.

jni:
@mkcor @alexdesiqueira I gotta say I don't understand this stuff anymore! :sweat_smile: To me it seems that either PR can go forward, #4859 was a simple rebase. Since @mkcor you have made some nice updates here, I think we can just close #4859 and keep going here!

alexdesiqueira:
@mkcor @jni I apologize, that was on me. Thank you for understanding Juan!

mkcor:
@jni I don't know if we can squeeze this fix into `0.18`... Anyway, I'm happy to be done with this resurrected PR! :relieved: 

@cosinus11 eeffe8a fixes #5069.

In the end, I respected the original vision outlined by @maaleske and @stefanv:

> My intuition is that we should never flatten. Each file should map to one image (array).
> We can build a special interface to load multi-images and unpack them, but as you suggest, that should be explicit.

Namely, an image collection of (multiple) multi-frame file(s) has a tree structure when specifying a loading function (`load_func` parameter): `ic[0][0]` is the first frame from the first file, `ic[0][1]` is the second frame from the first file, etc.

The changes brought by this PR don't overwrite those brought by #3928 -- still, it would be great if you could review, @alexdesiqueira! :smile:

mkcor:
Additional notes:

* d3a968e draws from https://github.com/scikit-image/scikit-image/pull/3928#discussion_r303211107;
* 03d2936 is pretty much the same as the example I updated in the [docstring](https://scikit-image.org/docs/dev/api/skimage.io.html?highlight=imagecollection#skimage.io.ImageCollection) at eeffe8a.

cosinus11:
> @jni I don't know if we can squeeze this fix into `0.18`... Anyway, I'm happy to be done with this resurrected PR! 😌
> 
> @cosinus11 [eeffe8a](https://github.com/scikit-image/scikit-image/commit/eeffe8a19acfe6bdee4b93b612a82095b886792b) fixes #5069.
> 
> In the end, I respected the original vision outlined by @maaleske and @stefanv:
> 
> > My intuition is that we should never flatten. Each file should map to one image (array).
> > We can build a special interface to load multi-images and unpack them, but as you suggest, that should be explicit.
> 
> Namely, an image collection of (multiple) multi-frame file(s) has a tree structure when specifying a loading function (`load_func` parameter): `ic[0][0]` is the first frame from the first file, `ic[0][1]` is the second frame from the first file, etc.
> 
> The changes brought by this PR don't overwrite those brought by #3928 -- still, it would be great if you could review, @alexdesiqueira! 😄

Thank you very much for your attention and your reactivity, amazing!
For your information, I use ImageCollection for reading 3D images for tomography (one file for 2048x2048x2048 voxels), in connection with PIL I read this images directly in raw format provided by the reconstruction software.

stefanv:
Thank you all for your feedback and work put into this PR; given that this has been reviewed by more eyes than most other PRs, I am putting it in.  Thanks @maaleske !

jni:

jni:
Wow, thank you so much @maaleske for the original work, and @alexdesiqueira @mkcor for the resurrection, and @stefanv for the merge! Amazing to not have this looming over my head anymore! 😂

alexdesiqueira:
Thank you very much everyone! Sorry if I didn't review @mkcor, thank you for your work!

Issue #2777, opened 2017-08-27
Reorganizing files in the root directory


soupault:
Originally posted by @jarrodmillman in https://github.com/scikit-image/scikit-image/pull/2775#issuecomment-325161048.

If you want to declutter the root directory, you could move the following files:
1. [x] __( See https://github.com/scikit-image/scikit-image/pull/2800 )__ ``TASKS.txt`` could be included directly in ``doc/source/contribute.rst``.  The name ``TASKS`` doesn't seem to match the content and looking at it from the commandline doesn't seem to be very helpful for a new contributor.
2. [ ] ``TODO.txt`` could be moved to ``doc/release/``.  It looks like it would be more relevant to an experienced developer or release manager anyway, so it isn't clear it needs to have so much visibility.
3. [x] __( See https://github.com/scikit-image/scikit-image/pull/2819 )__ ``DEPENDS.txt`` could be included directly in ``doc/source/install.rst``.  When I set up a development environment, I usually expect that the ``requirements.txt`` or ``pip install`` should handle dependencies for me.
4. [ ] ``RELEASE.txt`` could be moved to ``doc/release/``.  I suspect that only a few people will need to look at this file.  And those people should probably be familiar with where to find things in the repo.
5. [x] __( See https://github.com/scikit-image/scikit-image/pull/2781 )__ Change ``requirements.docs.txt`` and ``requirements.optional.txt`` to ``requirements/docs.txt`` and ``requirements/optional.txt``.  I would expect to find requirements files in a directory named requirements, so I don't think this would add confusion.  You could also add a ``requirements/README.md`` explaining what the various requirement files contain, if you wish.
6. [x] __( See https://github.com/scikit-image/scikit-image/pull/2779 )__ ``appveyor.yml`` could become ``.appveyor.yml`` so it doesn't show up with ``ls``.  This is the type of things I would expect to be hidden and shouldn't be necessary for new contributors to find from the commandline.~
7. [x] __( See https://github.com/scikit-image/scikit-image/pull/2776 )__ Remove ``bento.info`` and content in related files.  I think it is a dead project (correct me if I am wrong).  NumPy stopped supporting it a couple of years ago:  https://github.com/numpy/numpy/pull/6268

soupault:
For the discussion on `TODO.txt` future see https://github.com/scikit-image/scikit-image/issues/2651.

soupault:
I think this issue has been sufficiently addressed.

PR #2819, opened 2017-10-01
Move INSTALL to top-level

jarrodmillman:
## Description

This was supposed to declutter the root directory (related to #2777) by merging DEPENDS.txt with doc/source/install.  However, I think it makes sense to have an INSTALL file at the top of the repo.  I still merged DEPENDS with it, so it doesn't make the root directory any more cluttered.

Here is what README and INSTALL look like now:
- https://github.com/jarrodmillman/scikit-image/blob/depends/README.md
- https://github.com/jarrodmillman/scikit-image/blob/depends/INSTALL.rst


stefanv:
Having requirements specified in two places probably means we'll eventually forget to update one or the other.  Now that you've split the requirements files, why don't we just include these inside a literal block?

PR #2822, opened 2017-10-03
The gallery now points to the stable docs

stefanv:
I already took the liberty of updating the main webpage with the same.

### DIFF
--- a/doc/source/_templates/navbar.html
+++ b/doc/source/_templates/navbar.html
@@ -1,6 +1,6 @@
 <li><a href="/download.html">Download</a></li>
-<li><a href="/docs/dev/auto_examples">Gallery</a></li>
-<li><a href="/docs/dev">Documentation</a></li>
+<li><a href="{{ pathto('auto_examples/index') }}">Gallery</a></li>
+<li><a href="{{ pathto('index') }}">Documentation</a></li>
 <li><a href="/community_guidelines.html">Community Guidelines</a></li>
 
 <li><a href="https://github.com/scikit-image/scikit-image">


soupault:
Continuing the discussions from https://github.com/scikit-image/scikit-image-web/issues/40, https://github.com/scikit-image/scikit-image-web/issues/22, doesn't it make more sense to:
1. Either link to `stable docs` and `dev gallery` (https://github.com/scikit-image/scikit-image-web/issues/40#issuecomment-244751612) (exposing new features);
2. Or link to `stable docs` and `stable gallery` (for consistency)?

On a separate topic, I'd make `stable` and `dev` docs reachable more easily:
image

jni:
I agree with @soupault's comment: the gallery and the docs should point to the same version, Dev or Stable. I'd suggest Stable, as it will save lots of developer time answering questions about unreleased features.

stefanv:
The docs and gallery *do* point to the same version (both stable).

stefanv:
OK, how's this?

ahojnnes:
Agreeing on this as well and remembering the discussion back then. Pointing to stable is the way to go. Not merging and updating the website, since I don't have a setup for building the documentation at the moment.

stefanv:
Note that the PR has now been updated to stick to the version you are in 
(which will be stable, if the reader never chooses another version).

ahojnnes:
That sounds even better to me.

soupault:
@jni I think this fix will require rebuilding the docs for 0.13.

jni:
@stefanv has the secret sauce for building the docs without LaTeX errors everywhere. =P

stefanv:
I really hope not---this isn't supposed to be black magic!  It works out of the box on my Linux machine.

jni:
"works out of the box on my Linux machine" is black magic for those of us on other OSs. =) I do need to get a Linux box at home but in the meantime my LaTeX install is broken and I don't know how to fix it. In the future it would be awesome to switch to MathJAX in the docs so that the build doesn't depend on the builder's LaTeX install. Perhaps this has already been solved elsewhere in SciPy-land?

stefanv:
I will be build the docs on my MacOS box to double check. I reiterate that 
I do not have any special setup.

PR #2823, opened 2017-10-04
Adapt AppVeyor to use Python.org dist, and remove install script

stefanv:


stefanv:
We're green!!

ahojnnes:
Looks great!

Issue #2825, opened 2017-10-04
remove_small_objects with single label


Phlya:
When applied to a labelled image with only one label `skimage.morphology.remove_small_objects` treats all of pixels with the one label as a single object. It does show a warning, that perhaps a boolean array was intended to be used, where the behaviour is different, where it treats all connected components separately.

First of all, the documentation is a bit vague about this - the description of the method just says it removes connected components, and only in the description of the `ar` variable it says "If the array type is int, it is assumed that it contains already-labeled objects" - which also doesn't directly say that the behaviour is then different.
http://scikit-image.org/docs/stable/api/skimage.morphology.html#skimage.morphology.remove_small_objects

Second of all, this inconsistency creates a problem during batch processing of many images, when after labelling you don't know how many objects you will actually have, and just a warning is not very helpful - for a long time I was just thinking it was related to efficiency (i.e. with a single label it's more efficient to use boolean arrays) and completely ignored it until I discovered it actually wasn't doing what I expected it to.

Therefore, I think, at the very least the documentation should be more explicit about this, and ideally the behaviour should be changed. I am not sure what would be the best solution here, perhaps adding a keyword to define which behaviour is desired? Or splitting this method into to, with different behaviours?

Thanks!

jni:
Hi @Phlya, thanks for the report!

I agree 100% that the documentation should be updated. Would you like to submit a PR to that effect? Otherwise we'll get to it, but it's hard to predict when. =)

Regarding the change in behaviour, we'd have to be careful. Right now, other than the confusing warning, the function can be used both with connected-components labelling, and with user labelling (which allows for disjoint objects, which I can imagine in many scenarios). The behaviour will be consistent across a batch, so the user just needs to decide whether to convert their images to int or to bool.

What is your use case? Perhaps you want to add a flag that would silence the warnings? Lastly, whenever people ask for another flag, I always show them this page to make them ponder the costs and whether they are really necessary:

http://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html

Note that temporarily suppressing warnings is quite easy:

https://github.com/scikit-image/scikit-image/blob/b9e6a69a839bc65b3c70e0aad7b8e0ff06f81740/skimage/_shared/_warnings.py#L76-L105

We should probably expose that utility, I often find myself rummaging for it...

Phlya:
Thanks for the reply @jni! I can try to submit a PR for this, yes.

[...]

PR #2830, opened 2017-10-09
Better document remove_small_objects behaviour: int vs bool

Phlya:
Fixing documentation according to https://github.com/scikit-image/scikit-image/issues/2825

Also added an example showing the different in behaviour together with the warning that I get.

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/misc.py
+++ b/skimage/morphology/misc.py
@@ -46,22 +46,21 @@ def _check_dtype_supported(ar):
 
 def remove_small_objects(ar, min_size=64, connectivity=1, in_place=False):
     """Remove objects smaller than the specified size.
-    If ar is bool, labels connected components and removes objects smaller than
-    min_size. If ar is int, assumes objects have already been labelled. This
-    means that int arrays with only one value above 0 are treated differently
-    than boolean arrays, and all non-zero pixels are considered a single object
-    even if they are not touching.
-
+    Expects ar to be an array with labelled objects, and removes objects
+    smaller than min_size. If ar is bool, the image is first labelled.
+    This leads to potentially different behavior for bool and 0-and-1
+    arrays.
+    
     Parameters
     ----------
     ar : ndarray (arbitrary shape, int or bool type)
         The array containing the objects of interest. If the array type is
-        int, it is assumed that it contains already-labeled objects.
-        The ints must be non-negative.
+        int, the ints must be non-negative.
     min_size : int, optional (default: 64)
         The smallest allowable object size.
     connectivity : int, {1, 2, ..., ar.ndim}, optional (default: 1)
-        The connectivity defining the neighborhood of a pixel.
+        The connectivity defining the neighborhood of a pixel. Used during
+        labelling if ar is bool.
     in_place : bool, optional (default: False)
         If `True`, remove the objects in the input array itself.
         Otherwise, make a copy.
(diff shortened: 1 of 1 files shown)


stefanv:
How about a slightly different approach: we change the docstring to make it explicit that only labeled arrays should ever be provided.  We then keep the special behavior for boolean arrays, but warn that we are labeling the array before use.

This should simplify things quite a bit.

skimage/morphology/misc.py
```diff
@@ -45,20 +45,24 @@ def _check_dtype_supported(ar):
                         "Got %s." % ar.dtype)
 
 def remove_small_objects(ar, min_size=64, connectivity=1, in_place=False):
-    """Remove connected components smaller than the specified size.
+    """Remove objects smaller than the specified size.
+    Expects ar to be an array with labelled objects, and removes objects
```
stefanv: Insert newline after first descriptive line (NumpyDoc format)

skimage/morphology/misc.py
```diff
@@ -45,20 +45,24 @@ def _check_dtype_supported(ar):
...
+    smaller than min_size. If ar is bool, the image is first labelled.
```
stefanv: labeled

stefanv: ar -> ``` `ar` ```

skimage/morphology/misc.py
```diff
@@ -45,20 +45,24 @@ def _check_dtype_supported(ar):
...
-        The ints must be non-negative.
+        The array containing the objects of interest. If the array type is
+        int, the ints must be non-negative.
     min_size : int, optional (default: 64)
-        The smallest allowable connected component size.
+        The smallest allowable object size.
     connectivity : int, {1, 2, ..., ar.ndim}, optional (default: 1)
-        The connectivity defining the neighborhood of a pixel.
+        The connectivity defining the neighborhood of a pixel. Used during
+        labelling if ar is bool.
```
stefanv: ar -> ``` `ar` ```

skimage/morphology/misc.py
```diff
@@ -45,20 +45,24 @@ def _check_dtype_supported(ar):
...
     in_place : bool, optional (default: False)
-        If `True`, remove the connected components in the input array itself.
+        If `True`, remove the objects in the input array itself.
```
stefanv: If True or ``` If ``True`` ```

skimage/morphology/misc.py
```diff
@@ -92,6 +96,15 @@ def remove_small_objects(ar, min_size=64, connectivity=1, in_place=False):
     >>> d = morphology.remove_small_objects(a, 6, in_place=True)
     >>> d is a
     True
+    >>> e = morphology.remove_small_objects(a.astype(int), 6)
+    /usr/local/lib/python3.5/dist-packages/skimage/morphology/misc.py:135:
```
stefanv: This won't pass the test suite.  Let's remove this example.

Phlya: OK, that's a shame, I would have found this very useful when trying to figure out my problem...

stefanv:
I think I meant that we should warn in the code, not in the documentation.  The docs should only describe one way of input: a labeled array.

jni:
@stefanv I kind of like the bool polymorphism, but I see your point. However if we really want to discourage the use of bool then I suggest we simply deprecate it. @Phlya as one of the people bitten by the dual behaviour, what do you prefer?

stefanv:
@jni Agreed.  For consistency, let's rename the parameter `label_image` and raise when it is of dtype `bool`.

PR #2603, opened 2017-04-09
Add nD support to image moments computation

jni:
## Description
Image moments can now be computed for images of any dimension. Many thanks to @jaimefrio for his optimization help on the mailing list for these functions. Reference: https://mail.python.org/pipermail/scikit-image/2017-April/005169.html

PR #2832, opened 2017-10-11
Switch from LaTeX to MathJax in doc build

stefanv:


soupault:
Everything builds fine, although there are some rendering issues. I don't remember all the places in the docs where we have formulas, but here is a faulty one:
image

soupault:
>I don't remember all the places in the docs where we have formulas
Nevermind, I can just check every generated html-file manually :smile: ...

The one above appears to be the only issue. Fixed, and pushed to @stefanv 's branch.

JDWarner:
Looks good to me.  We still need to squash the hyperlinking error, which is being seen in other builds as well (#2603), so I'm pretty sure it's unrelated.

PR #2835, opened 2017-10-12
Corrected error with scales attribute in ORB.detect_and_extract

AlexisMignon:
Corrected error in the way scales were built in the ORB.detect_and_extract method

## Description

When using ORB.detect_and_extract there was an error in the way the scales attribute was built. It doesn't take into account the fact that other attributes are built using a mask and relies on the size of the size of an unmasked array instead of a masked one.
## Checklist
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/orb.py
+++ b/skimage/feature/orb.py
@@ -312,11 +312,12 @@ class ORB(FeatureDetector, DescriptorExtractor):
             descriptors, mask = self._extract_octave(octave_image, keypoints,
                                                      orientations)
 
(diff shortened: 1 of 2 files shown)


stefanv:
Thanks for catching this @AlexisMignon .  Let's see if we can get Johannes or Vighnesh, more familiar with this part of the code base, to review.

ahojnnes:
Looks good to me, thanks for catching the bug.

jni:
@AlexisMignon wow, sorry for the long wait! 😂 Thank you very much for the fix!

Issue #2695, opened 2017-07-11
Wording on order of iteration is confusing in user guide


stefanv:
"When the dimension you are iterating over is even larger, the speedup is even more dramatic. It is worth thinking about this data locality when writing algorithms. In particular, know that scikit-image uses C-contiguous arrays unless otherwise specified, so one should iterate along the last/rightmost dimension in the innermost loop of the computation."

This can be interpreted in many ways, and can be confusing, given the code above it at:

http://scikit-image.org/docs/stable/user_guide/numpy_images.html#notes-on-array-order

stefanv:
We may be able to expand this further in the future by showing a C-loop, but we don't currently have the ability to compile & display that in Sphinx.

PR #2836, opened 2017-10-12
reworded paragraph on order of iteration

anufern:
## Description
rewording as requested as in #2695 

## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/doc/source/user_guide/numpy_images.rst
+++ b/doc/source/user_guide/numpy_images.rst
@@ -91,7 +91,7 @@ Boolean arithmetic can be used to define more complex masks: ::
     >>> lower_half = row > cnt_row
     >>> lower_half_disk = np.logical_and(lower_half, outer_disk_mask)
     >>> camera = data.camera()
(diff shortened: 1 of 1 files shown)


stefanv:
I'm glad you caught this one!  I think this is a vast improvement; but can we explain it even better?  This seems like a great opportunity to educate, with an example.

anufern:
I agree how about a diagram of a c contiguous array with an arrow to show which way the iteration should occur
 "along the last/rightmost dimension in the innermost loop of the computation."

jni:
Yes, a diagram is sorely needed here. I would also add that scikit-image uses c-ordering *because it follows NumPy defaults*.

dpshelio:
@anufern - I'll help you to do that diagram this week, and we can also include @jni comment about NumPy.

jni:
@anufern I'm closing this in favour of #3127. We would still welcome a diagram to clarify these concepts!

PR #2838, opened 2017-10-13
Make `regionprops._slice` a user-facing property

jni:
Currently, `_slice` is hidden from the user, even though it would be
a useful property in many instances (for example, extracting objects
from one channel based on the regionprops of another). In fact there
was a discussion about this when we were talking about the bounding box
property, but that went nowhere.

## Checklist
- [ ] <del>Gallery example in `./doc/examples` (new features only)</del>
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_regionprops.py
+++ b/skimage/measure/_regionprops.py
@@ -58,6 +58,7 @@ PROPS = {
     'Perimeter': 'perimeter',
     # 'PixelIdxList',
     # 'PixelList',
+    'Slice': 'slice',
     'Solidity': 'solidity',
     # 'SubarrayIdx'
     'WeightedCentralMoments': 'weighted_moments_central',
@@ -110,6 +111,7 @@ class _RegionProperties(object):
         self.label = label
 
         self._slice = slice
+        self.slice = slice
         self._label_image = label_image
         self._intensity_image = intensity_image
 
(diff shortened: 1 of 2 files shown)


skimage/measure/_regionprops.py
```diff
@@ -101,6 +102,7 @@ def __init__(self, slice, label, label_image, intensity_image,
         self.label = label
 
         self._slice = slice
```
soupault: Shouldn't we remove `_slice` then?

stefanv: Also, isn't it confusing to use a reserved word here in several places?

jni: @stefanv I'm all ears for an alternative attribute name. Note that the `slice` in the `__init__` is old, and annoying to deprecate. `self.slice` is *not* very confusing, though, only a little bit confusing.

@soupault even though `._slice` is private, I'm concerned some of our users might have accessed it. So I thought we might as well keep it around, because it's not very costly to do so.

soupault: @jni I see.
As for the name, maybe something like `region_slice` or `slice_region`?

stefanv: self.slice is probably OK.  I'm more concerned with cases where the built-in gets clobbered.  `region_slice` or `slice_` work for me.

stefanv:
@jni Did you want to modify this further, or should we get it in?

jni:
@stefanv I'm happy with it as-is. Rebased.

PR #2846, opened 2017-10-14
OverflowError: Python int too large to convert to C long on win-amd64-py2.7

cgohlke:
Fixes two test errors for scikit-image-0.13.1.win-amd64-py2.7, where `xrange` is implemented using `C long` (32-bit). There's a related [CPython issue](https://bugs.python.org/issue26428), but it is "rejected".

```
======================================================================
ERROR: skimage.morphology.tests.test_skeletonize.TestThin.test_noiter
----------------------------------------------------------------------
  File "X:\Python27-x64\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "X:\Python27-x64\lib\site-packages\skimage\morphology\tests\test_skeletonize.py", line 144, in test_noiter
    result = thin(self.input_image).astype(np.uint8)
  File "X:\Python27-x64\lib\site-packages\skimage\morphology\_skeletonize.py", line 266, in thin
    for i in range(max_iter):
OverflowError: Python int too large to convert to C long

======================================================================
ERROR: skimage.morphology.tests.test_skeletonize.TestThin.test_zeros
----------------------------------------------------------------------
  File "X:\Python27-x64\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "X:\Python27-x64\lib\site-packages\skimage\morphology\tests\test_skeletonize.py", line 130, in test_zeros
    assert np.all(thin(np.zeros((10, 10))) == False)
  File "X:\Python27-x64\lib\site-packages\skimage\morphology\_skeletonize.py", line 266, in thin
    for i in range(max_iter):
OverflowError: Python int too large to convert to C long
```

### DIFF
--- a/skimage/morphology/_skeletonize.py
+++ b/skimage/morphology/_skeletonize.py
@@ -250,7 +250,7 @@ def thin(image, max_iter=None):
            [0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
     """
     # check parameters
-    max_iter = max_iter or sys.maxsize
+    max_iter = max_iter or 2**31-1
     # check that image is 2d
     assert_nD(image, 2)
 


cgohlke:
>         # Bug in CPython Win64

It's an issue in win-amd64-py2.x only. 

If the iteration limit is so important, use a `while` loop.

stefanv:
2**31 - 1 seems plenty.  Any reason not to apply this patch?  The extra complexity of getting it *exactly right* does not seem worth it.

jni:
:+1: to merging

stefanv:
@jni Chris closed this branch, so we need to reopen a PR. See #2865

PR #2847, opened 2017-10-14
Allow convex area calculation in 3D for regionprops

jni:
## Description
This uses an "expanded" convex hull calculation and volume to allow the definition of convex area and solidity for 3D objects. The current definition doesn't guarantee that solidity will be <= 1, but I think it's still useful. We could set the `square_pixels` argument to `True` to guarantee this, but that is a big overestimate in general, so I'm not super-excited about it. But, willing to discuss!

## Checklist
- [x] Unit tests

## For reviewers

  later.


skimage/measure/_volume.py
```diff
@@ -0,0 +1,68 @@
...
+def _offsets_cube(ndim):
+    offsets = np.zeros((2 ** ndim, ndim))
+    for vertex, offset in enumerate(product(*[[-0.5, 0.5]] * ndim)):
+        offsets[vertex] = offset
+    return offsets
+
+
+def expanded_convex_hull(points, reduce_inner_points=True,
+                         square_pixels=False):
+    """Return the ConvexHull of input points jiggled by 0.5 along all axes.
```
stefanv: convex hull?

skimage/measure/_volume.py
```diff
@@ -0,0 +1,68 @@
...
+
+    If we think of pixels as having zero area, then a normal convex hull on
```
stefanv: I don't understand why this works.

jni:
@scikit-image/core this is ready for review!

skimage/morphology/convex_hull.py
```diff
@@ -23,6 +27,14 @@ def convex_hull_image(image):
     ----------
     image : (M, N) array
```
soupault: The shape pattern should be updated (not (M, N) anymore)

skimage/morphology/convex_hull.py
```diff
@@ -34,48 +46,50 @@ def convex_hull_image(image):
...
-    mask = grid_points_in_poly(image.shape[:2], v)
+    hull = ConvexHull(coords)
+    vertices = hull.points[hull.vertices]
+
+    # If 2D, sort vertices clock-wise, and use Cython function to locate
+    # convex hull pixels
+    if ndim == 2:
+        offset = np.mean(vertices, axis=0)
+        v_centred = vertices - offset
+        angles = np.arctan2(v_centred[:, 0], v_centred[:, 1])
```
soupault: From https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.spatial.ConvexHull.html :
`We could also have directly used the vertices of the hull, which for 2-D are guaranteed to be in counterclockwise order`

skimage/morphology/convex_hull.py
```diff
@@ -34,48 +46,50 @@ def convex_hull_image(image):
...
+        vertices = vertices[np.argsort(angles)]
+        mask = grid_points_in_poly(image.shape, vertices)
+    else:
+        gridcoords = np.reshape(np.mgrid[tuple(map(slice, image.shape))],
+                                (ndim, -1))
+        # A point is in the hull if it satisfies all of the hull's inequalities
+        coords_in_hull = np.all(hull.equations[:, :ndim].dot(gridcoords) +
```
soupault: By any chance, have you checked how much this is slower compared to `grid_points_in_poly`?

stefanv:
Nicely done, @jni!  Can you verify whether the 2D and 3D routines calculate the same values for 2D input?

skimage/morphology/convex_hull.py
jni: This is old code, but I can update, sure!

skimage/morphology/convex_hull.py
jni: You mean in 2D? No, I haven't. You're right, worth a look.

jni: @soupault the hull equations are about 4.9x slower than `grid_points_in_poly`. There might be some optimization to be done with the array order of the coordinates, but it escapes me at present.

jni: Here is my test with `horse`, but the timing results were remarkably consistent when using `measure.tests.test_regionprops.SAMPLE`.

```python
In [26]: %timeit grid_points_in_poly2(image.shape, hull)
10.7 ms ± 199 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [27]: %timeit ch.grid_points_in_poly(image.shape, vertices)
2.16 ms ± 50.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

jni:
@stefanv yes, the output is the same!

```python
In [28]: np.all(grid_points_in_poly2(image.shape, hull) == ch.grid_points_in_poly(image.shape, vertices))
Out[28]: True
In [29]: horse3d = np.stack((horse, horse, horse))
In [30]: horse3d.shape
Out[30]: (3, 328, 400)
In [31]: convex_3d_horse = ch.convex_hull_image(horse3d)
In [32]: convex_horse = ch.convex_hull_image(horse)
In [33]: np.all(convex_horse == convex_3d_horse[1])
Out[33]: True
```

I'll add a test to this effect.

stefanv:
@jni How is that possible, with the dithering happen in 3D coords?!  Either way, that's great!

jni:
@soupault I'll work on a 3D example once my Python 3D viz skills are better. =)

@stefanv updated with better tests. =)

jni:
Ok the tests are a bit messier than I'd hoped but they are correct. The single missing point is exactly on the hull. I think the 3D version actually does a better job of identifying points that lie exactly on the hull, thanks to the tolerance.

skimage/morphology/convex_hull.py
stefanv: Remember this comment.

stefanv:
@jni Is the one point *on* the hull enough reason to modify our 2D implementation, or do you see them both as correct?

skimage/morphology/convex_hull.py
soupault: @jni Huge thanks for the benchmarking!

jni:
@stefanv regarding your question about the convex horse (great name for a band or a bar), I suspect that the probability of a point lying exactly on the hull decreases dramatically as the image size increases, except for perfectly horizontal or vertical lines, which the 2D Cython code apparently handles correctly.

I'm not sure that there is a single answer as to which version is more "correct". This is down to floating point arithmetic. You can say that checking `a == b` is "incorrect" when it comes to floating point comparisons, but then again, with `abs(a - b) < tol`, you open yourself to false positives below that tolerance. At any rate, I don't want to mess with the faster and existing 2D code, as other people probably depend on it. I don't consider these single-pixel "errors" to be a problem in real-world use. (Or someone would have complained by now. ;)

stefanv:
@jni Agreed, thanks.

PR #1444, opened 2015-03-22
Computing haar features

spark1729:
Implement Haar-like features as discussed in Issuee #1431 .

PR #1570, opened 2015-06-26
Object detect module

warmspringwinds:
I marked up the future module.
Changed the `mb_lbp` function to work faster.

PR #2848, opened 2017-10-16
Add Haar-like features

### LINKED ISSUE #1431: Implement Haar-like features
https://en.wikipedia.org/wiki/Haar-like_features

glemaitre:
## Description
This PR intend to implement the Haar-like feature as proposed in #1431 
Somehow it resurrects #1444

## Checklist

- [x] Unit tests

## References
closes #1431 

Viola, Paul, and Michael Jones. "Rapid object detection using a boosted cascade of simple features." Computer Vision and Pattern Recognition, 2001. CVPR 2001. Proceedings of the 2001 IEEE Computer Society Conference on. Vol. 1. IEEE, 2001.
[PDF](http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=501DBD52721A740AA6C6A7E4ECE28E1A?doi=10.1.1.10.6807&rep=rep1&type=pdf)

## For reviewers

  later.


glemaitre:
Minimum example for usage

```python
import numpy as np
from joblib import Parallel, delayed
from sklearn.feature_extraction.image import extract_patches_2d
from skimage.feature import haar_like_feature

img = np.random.random((100, 100))
patches = extract_patches_2d(img, patch_size=(24, 24), max_patches=10)

haar_features = Parallel(n_jobs=-1)(delayed(haar_like_feature)(roi, 'type-2-x')
                                    for roi in patches)
```

I would think that the Haar-like features are then used in an AdaBoost to make a cascade learner.

I also added a visualization function which is not necessary (it was just to check before to make real unit tests): 

```python
from skimage.io import imshow
from skimage.feature import haar_like_feature, haar_like_feature_coord, haar_like_feature_visualize

features = haar_like_feature_coord(feature_type='type-2-x', height=24, width=24)
output = haar_like_feature_visualize(features, 24, 24)

for img in output:
    imshow(img)
    plt.show()
```

glemaitre:
So before to go further, I would like some advice regarding:

- the unit tests;
- the performance of the integral computation.

### Unit tests

For the moment, I did not implement unit tests per se. I wanted to know what would be the best strategy.
I add in mind to generate mask using the visualization function and check if it corresponds to what is expected. We can also give `np.ones` input which should give a known output for the different type of features. However, it does not allow to check all possibilities, so WDYT?

### Performances

I wrote the code such that it fits with the `skimage.transform.integrate` function.
However, I saw that there is no cython version of the code and I would think that the benefit of such Haar-like feature is also the speed for which the feature can be computed (if I remember well my master course, it was supposed to solve some real-time problems at testing). I would agree that the code is quite user-friendly and readable but at the cost of speed limitation.

Therefore, is it wise to reuse the `integrate` function or implement a pure C (I mean Cython) version?

glemaitre:
ping @stefanv @jni @JDWarner @emmanuelle 

Here comes my couple of thoughts. Any suggestion?

skimage/feature/haar.py
```diff
@@ -0,0 +1,192 @@
...
+        width_feature = range(2, width, 2)
+        output = [[[], []], [[], []], [[], []], [[], []]]
+    else:
+        known_feature = ('type-2-x', 'type-2-y',
+                         'type-3-x', 'type-3-y',
+                         'type-4')
+        raise ValueError('The type of feature is unknown. Got {} instead of'
+                         ' one of {}'.format(feature_type, known_feature))
+
+    for y in range(height):
```
stefanv: The PR looks fine, except for this unpacked loop which gives me the heebie-jeebies.  Is there any way we can script this in a single, parameterized loop for all cases?

stefanv:
@glemaitre I just saw your extended comment:

Yes, I think hand calculating an expected Haar response on a minimal array is ideal, if that's feasible.

W.r.t. speed, we have fast integration code (see `skimage._shared.transform`) that can be re-used.

@warmspringwinds You may be interested in helping us review this PR, given your experience with multi-block feature extraction.

skimage/feature/haar.py
glemaitre: I have to think if a meshgrid or other type of grid could be used for that matter

stefanv: I should clarify that with "single", I meant "an additional".

glemaitre:
> I should clarify that with "single", I meant "an additional".

@stefanv I am not sure to get what you mean by "additional". If you can give the beginning of a snippet, I think that it might help.

Anyhow, I cynthonize the code. I would still need to profile to be sure that the remaining python calls are not the bottleneck but we already go from 22.9 s to 278 ms for one type of feature (10 patches of 24 x 24). So it seems much more useful in practice. I let make some function available at the python level in case that this useful

I am going to profile and start the unit testing. In the meanwhile, I would be happy to have some feedback on the coding style (I am not used to code in cython).

warmspringwinds:
@stefanv Sure, I would be glad to help to review it :)

@glemaitre Thank you for your contribution!
Haar-features and [Multiblock local binary patterns](http://scikit-image.org/docs/dev/auto_examples/features_detection/plot_multiblock_local_binary_pattern.html) (which is a hybrid of Local Binary Patterns and Haar features) are features that can be computed very fast and give good object detection and recognition performance while working in real-time on cpu.
That's a very interesting and important problem and, if solved, can add detection and recognition capability to scikit-image :)

Regarding your current efforts, I think you can reuse or customize a good amount of code from [here](http://scikit-image.org/docs/dev/auto_examples/features_detection/plot_multiblock_local_binary_pattern.html) (visualization and efficient computation).

Moreover, there is a [unfinished work](https://github.com/scikit-image/scikit-image/pull/1570) with adding an object detector module to scikit-image -- both training (adaboost classifier as you mentioned) and inference. You can also reuse code from there or, if you are interested, you can take over that work and I can guide you :)

Regarding your question about ```integrate``` function:
[there already exists and efficient implementation in scikit-image](https://github.com/scikit-image/scikit-image/blob/2f71b5787662ce7589bb878c55359cd076f84c83/skimage/_shared/transform.pyx#L7).
Feel free to ping me with more questions.

stefanv:
Thanks, @warmspringwinds!

stefanv:
Great to see the speed improvement using Cython, @glemaitre!

I think most of the time you should be able to get away without the manual memory allocations, but for the rest the code looks good to me.

glemaitre:
I made the code a bit more readable. However, I have a concern.

`skimage._shared.transform.integrate` is using a float (np.float32) memory view. However `integral_image` will keep the type of the data (at least integral and floating). I think that using fused type (at least for the input) could be nice.

WDYT?

glemaitre:
I did fused type the integral function which can be beneficial in other part of the code.
I also release the GIL is the cython code whenever is possible.

I have an API question. Right now, the `haar_like_feature` is taking a sub-ROI (integrated) to compute the feature. Should consider to change the API to accept a full-image, compute the integral of it, and then call the current implementation on a given window or we are letting the user managing this part.

stefanv:
> I have an API question. Right now, the haar_like_feature is taking a sub-ROI (integrated) to compute the feature. Should consider to change the API to accept a full-image, compute the integral of it, and then call the current implementation on a given window or we are letting the user managing this part.

Can you explain this @glemaitre ?

stefanv:
The fused dtypes look good!  Should the accumulator type not also be changed?

glemaitre:
> Can you explain this @glemaitre ?

### Case 1 -> current implementation

The current signature is

```python
haar_like_feature(roi_ii, feature_type)
```

with the output being a 1D ndarray

It means that a user will need to do something like:

```python
import numpy as np
from sklearn.feature_extraction.image import extract_patches_2d
from skimage.feature import haar_like_feature
from skimage.transform import integral_image

img = np.random.random((100, 100))
img_ii = integral_image(img)

size_detector = (24, 24)
patches = extract_patches_2d(img_ii, patch_size=size_detector)

haar_features = [haar_like_feature(roi, 'type-2-x') for roi in patches]

# ... further processing with labeling and training
```

### Case 2

It would be possible to extract the feature for a full image with the following signature

```python
haar_like_feature(img, size_detector, feature_type) # + some n_jobs
```

In this case, the call will be only:

```python
mport numpy as np
from skimage.feature import haar_like_feature

img = np.random.random((100, 100))

size_detector = (24, 24)
haar_features = haar_like_feature(img, size_detector, 'type-2-x')

# ... further processing with labeling and training
```

in this case you will get a 2D ndarray with (n_samples, n_features) since the patch extraction and integral computation will be included in the function itself.

### Summarize

I like the first implementation because it allows for more flexibility. The patch extraction is not integrated and you can make your own extractor (with custom overlapping, etc ...). The second version has the benefit of not exposing those mechanisms to the user.

@stefanv WDYT?

glemaitre:
> The fused dtypes look good! Should the accumulator type not also be changed?

I thought it would be more tricky to infer the type for the output but then it is possible.

glemaitre:
@stefanv @warmspringwinds I think that I am fine with the current version.

I would appreciate if you could review. I followed the multiblock LBP API to be consistent. I also introduce the possibility to extract feature using pre-computed rectangle. I think this is important since that the usual scenario is to use a feature-selection (AdaBoost) and then only compute those selected feature during inference. It is not as efficient when considering all feature because I need to create the C data structure, but I think that for a small number of feature it will be beneficial.

In the meanwhile I will just finish the gallery example in which I want to show an example of feature extracted similarly to other descriptors.

warmspringwinds:
@glemaitre Awesome! Thank you.

I will have a look and comment on parts of your code if I have questions/suggestions :)

stefanv:
@glemaitre This is looking great; excellent work, thanks.

Since you're still working on this, a small style issue: the docstrings become very long with blank lines between each item (both parameters, and references); so please remove those.

Otherwise, good style, perfect test coverage.  Looking forward to the gallery example!

glemaitre:
@stefanv I just pushed the gallery example and this is opened to be improved.

doc/examples/features_detection/plot_haar.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+from skimage.data import astronaut
+
+from skimage.color import rgb2gray
+
+from skimage.io import imshow
```
stefanv: Unused.

doc/examples/features_detection/plot_haar.py
```diff
@@ -0,0 +1,101 @@
+"""
+============================
+Haar-like feature descriptor
+============================
+
+Haar-like feature descriptors are a set of simple features which was used in
+the first real-time face detector [1]_. Indeed, these features can be
```
stefanv: Thanks for the example---I think a visual aid will really help readers to understand what these features are all about.

The 2nd sentence of this paragraph doesn't say much to me; could you clarify?

Also, can you try and give an intuitive description of what Haar features are, when they are typically detected, etc.?  This is where we get to educate and draw in the reader.

doc/examples/features_detection/plot_haar.py
```diff
@@ -0,0 +1,101 @@
...
+###############################################################################
+# The procedure to extract the Haar-like feature for an image is quite easy:
+# a region of interest (ROI) is defined for which all possible feature will
+# be extracted. In order to compute those features, an integral image needs to
+# be provided. Typically, for a ROI of 24 by 24 pixels, the total amount of
+# features which can be extracted is 160,000.
+
+image = rgb2gray(astronaut())
+plt.figure()
+plt.imshow(draw_haar_like_feature(image, 125, 200, 50, 50,
```
stefanv: I couldn't figure out, looking at this ROI, *why* the features got selected there.  Any insight?

skimage/feature/_haar.pyx
```diff
@@ -0,0 +1,318 @@
...
+
+    if feature_type == 0 or feature_type == 1:
+        local_n_rectangle = 2
+    elif feature_type == 2 or feature_type == 3:
+        local_n_rectangle = 3
+    else:
+        local_n_rectangle = 4
+    n_rectangle[0] = local_n_rectangle
+
+    rect_feat = <Rectangle**> malloc(local_n_rectangle * sizeof(Rectangle*))
```
stefanv: How much slower is it to use a Cython list, so that we don't have to do any custom memory allocation?

skimage/feature/_haar.pyx
```diff
@@ -0,0 +1,318 @@
...
+        - 'type-2-y': 2 rectangles varying along the y axis;
+        - 'type-3-x': 3 rectangles varying along the x axis;
+        - 'type-3-y': 3 rectangles varying along the y axis;
+        - 'type-4': 4 rectangles varying along x and y axis.
+
+    Returns
+    -------
+    haar_features : (n_features,) ndarray
+        Resulting Haar-like features.
+
```
stefanv: What is the dtype of this array, what does it contain?

skimage/feature/_haar.pyx
```diff
@@ -0,0 +1,318 @@
...
+        If using with ``feature_coord``, it should corresponds to the feature
+        type of each associated coordinate feature.
+    feature_coord : ndarray of list of tuples or None, optional
+        The array of coordinates to be extracted. This is useful when you want
+        to recompute only a subset of features. In this case ``feature_type``
```
stefanv: Single back-ticks around parameter names, double back-ticks around code.

skimage/feature/haar.py
```diff
@@ -0,0 +1,319 @@
...
+    if feature_type is None:
+        feature_type_ = FEATURE_TYPE
+    else:
+        if isinstance(feature_type, six.string_types):
+            feature_type_ = [feature_type]
+        else:
+            feature_type_ = feature_type
+        for feat_t in feature_type_:
+            if feat_t not in FEATURE_TYPE:
+                raise ValueError('The given feature type is unknown. Got {}'
```
stefanv: Can newline after opening bracket, then you have a bit more room to work with here

doc/examples/features_detection/plot_haar.py
glemaitre: Those features have not been selected. We don't make any selection, we are only generating them. I just plotted 3 random features over the 20,000 possible computed in the ROI of type-4.

skimage/feature/_haar.pyx
glemaitre: I'll try to check. Not sure what you mean by cython list. If actually you require a python interaction then I will not be able to release the GIL, which could be interesting in case you want to make Parallel with threading backend.

glemaitre: So you go from 150 ms to 240 ms to compute 43,000 features.

I put the code in a [gist](https://gist.github.com/glemaitre/c5f31b0669c1c11c8e2de272dcb4a665#file-list_haar-py). 
Appending in a python list will trigger a conversion of the C structure to a python dictionary.

Personally, I prefer a pure C allocation because I know how C will behave. To be honest, I was unsure of what would be the implicit conversion and I am not sure that mixing python and C in this specific piece of code will make it easier to understand while reading. But I am not an expert in Cython, so it could be due to make lack of experience there.

I still think that releasing the GIL could be nice as well.

WDYT?

doc/examples/features_detection/plot_haar.py
glemaitre: Seeing your question, it means that my example is not good (I was already questioning myself :))
What do you have in mind knowing that basically, all features are extracted.

glemaitre: FYI: I really like this: https://www.youtube.com/watch?v=hPCTwxF0qf4
For this case only the selected features are shown.

However, I think that it would be interesting to show something similar. take the example of the first extracted feature, and show the computation, then a second feature, ... etc. until the point that it is clear that all possible features are extracted for the ROI.

stefanv:
Oh, I see, so how are these typically used in practice? How to determine 
where and what type of features to compute at any location?

stefanv:
Using a Python list will definitely require the GIL, thanks for clarifying.

stefanv:
Yes, I think the code is good as is. I just always worry about being sure 
the appropriate free is done, but the logic here is not too complicated.

I try to use the heap wherever possible to work around that issue.

@ahojnnes @jni

stefanv:
Some text explaining what is usually calculated, vs what is shown here, 
should suffice.

stefanv:
That's a great video! We don't currently harte any animated examples in the 
gallery, but either matplotlib or moviepy should be able to produce 
animated gifs or MNGs.

This is an "above and beyond" milestone, but would be awesome!

glemaitre:
‎Then you would construct a 2d matrix where each line correspond to the feature extracted in a ROI and the target will be a bool representing the presence of the object. Therefore you will have multiple ROI for a given image and use an adaboost and roughly speaking make a feature selection (finding the X columns that are useful using selectfrommodel from scikit-learn)Once this done you will have your subset of feature to use. Then you can use these features using feature_coord and feature_type to recompute only the needed one.

stefanv:
Hrm, so it sounds like we could fairly easily build an object detector in a few lines of code.  That would be another fun gallery example.

glemaitre:
> Hrm, so it sounds like we could fairly easily build an object detector in a few lines of code. That would be another fun gallery example.

If you don't have issue requiring scikit-learn for example, it can be done quite easily I think.
Then the limitation could be the time to execute the example. We could maybe only limit to an image and see what feature we get.

I need to build something similar for some project, i could adapt that.

stefanv:
I think a dependence on scikit-learn only for the gallery is fine, yes.

stefanv:
I responded by email earlier, and now I see GH does not append them to the right comments; sorry about that :/  Hope it made sense still.

warmspringwinds:
@stefanv @glemaitre sorry for the delay guys -- I am having a look at your code this week

glemaitre:
Yes. Cython does not handle read only buffer when using memory view which is used for instance in joblib with Parallel. Then the trick is to use an numpy array and make a copy of the ROI and then we are fine with the memview. After profiling this is as fast and enable parallel processing for multipatch using joblib.

stefanv:
If this is functionality we explicitly envision, it would be good to add that to the gallery example (I don't think a dependency on sklearn in the gallery is an issue).

stefanv:
Also, I'm not quite sure I understand the joblib issue; do you mind clarifying a bit further?  Or perhaps I'll wait for the code example, and it'll make sense.

glemaitre:
To have a rough idea of what it will look at with joblib, you can check https://github.com/scikit-image/scikit-image/pull/2848#issuecomment-336941027

In the internal of joblib when using Parallel, the data which are shared are managed by a read-only memmap (I would assume that this is done to avoid the workers to modify some data that they are not supposed to).

Due to this read only buffer, you run into that issue: https://github.com/cython/cython/issues/1605
I think that in the future, cython should support read-only type memory-views.

So the trick is to copy the ROI into a numpy array read/write in the wrapper function which will not change the other part of the code (we can still use memory-views). Regarding performance, there is not changes.

glemaitre:
> If this is functionality we explicitly envision, it would be good to add that to the gallery example (I don't think a dependency on sklearn in the gallery is an issue).

I am still working to have a good gallery example close to a real use case. It will take a bit of time :)

stefanv:
I think the above code snippet you linked to is already a good example for the gallery.

Thank you for the additional information on the read-only views; do you mind adding some of that in as a comment in the code, so that we will know to revert to memory views in the future, once the Cython bug is closed?

skimage/feature/_haar.pyx
```diff
@@ -262,9 +262,10 @@ cpdef haar_like_feature_wrapper(cnp.ndarray[integral_floating, ndim=2] int_image
         Py_ssize_t n_rectangle, n_feature
         Py_ssize_t idx_rect, idx_feature
         integral_floating[:, ::1] rect_feature
-        # convert to an ndarray to a memory-view to be able to release the
-        # GIL. Furthermore, we need currently a ndarray to be able to use
-        # read-only memmap (for instance when using joblib).
+        # FIXME: convert to an ndarray to a memory-view to be able to release
```
stefanv: I'm struggling to parse this sentence.

doc/examples/features_detection/plot_haar.py
warmspringwinds: Looks good to me :)

I would also add that it can be computed in a constant time on any scale
which is a great advantage over other descriptors.

skimage/feature/_haar.pyx
warmspringwinds: @glemaitre You can also try C++ vector like here:
https://github.com/warmspringwinds/scikit-image/blob/object_detect_module/skimage/future/detect/cascade.pyx#L92

warmspringwinds:
@glemaitre I am still looking at your code and made a couple of comments. Would be cool to see if it's possible to use C++ vector structure in some parts of your code to make it more readable.

glemaitre:
> @glemaitre  I am still looking at your code and made a couple of comments. Would be  cool to see if it's possible to use C++ vector structure in some parts  of your code to make it more readable.

I used the C++ vector keeping the same structure since this is the most cache friendly. Inversing the indices (rectangle, feature) would be much less cache friendly and will be x10 slower (after testing it).

stefanv:
The code looks just about ready from my POV.  The gallery example can still be expanded to be more descriptive, but I also don't want to exhaust you on that problem.  If you feel you've done what you can there, I'm happy to merge this PR (thanks, great work!).

glemaitre:
I am still working on the example and I will change draw_haar_like_feature to have a more consistent API. I should have something ready by next week.

glemaitre:
@stefanv @warmspringwinds This is finished on my side. I would appreciate if you can give a look to the gallery example. I want to be sure that the narrative description is ok with you.

This example will required joblib (interactive session (typically ipython/sphinx-gallery) will required the master version while user can use the last release withtout issue) and scikit-learn.

glemaitre:
Since that I am not really sure how to integrate those requirements (doc vs. optional), I will need some advise on this.

warmspringwinds:
@glemaitre Great work! 
I will go through your updated code one more time this week and
I think we can merge it after that :)

glemaitre:
ping @stefanv @warmspringwinds Did you have time to look at the PR?

stefanv:
Thank you for the reminder @glemaitre; I am happy with this PR.  Unless @warmspringwinds responds with a comment, let's merge.

emmanuelle:
@stefanv the PR is great, but are you sure you want to merge with Travis failing? The issue is that joblib is not a required dependency, therefore Travis fails when building the doc.

emmanuelle:
@glemaitre we do have two files for requirements, requirements/default.txt and requirements/docs.txt. Please add joblib to docs.txt

glemaitre:
Thanks, that was one of my question: https://github.com/scikit-image/scikit-image/pull/2848#issuecomment-344091221

glemaitre:
I also might get into trouble with sphinx-gallery and joblib when trying pickling a function in `__main__`. I might need to trick a bit the example. Let see the CI (I am actually using the master version so I would like to be sure)

stefanv:
@emmanuelle Thanks for catching that; I didn't see the joblib dependency.  @glemaitre Can this not be done with dask, which is an existing dependency?

glemaitre:
@stefanv I need scikit-learn in which joblib is vendored anyway
I changed the code accordingly

glemaitre:
It might be nice to kill the job in travis of the commit 5b374ef since that it should fail

stefanv:
I guess that's okay; we don't have any other joblib examples, but we also don't have any other sklearn examples.  I presume the dask and joblib performances are about the same in this case.

glemaitre:
uhm, is docs.txt installed for each build in travis, because it is still failing?

glemaitre:
I am bit confused with the requirements. I assume that you would like not to have scikit-learn in the required dependencies. In the other hand the example is always executed. Would it make sense to skip running this example when the doc requirements were not installed.

warmspringwinds:
I am on it today guys. Sorry for delay.

doc/examples/features_detection/plot_haar.py
```diff
@@ -0,0 +1,169 @@
+"""
+============================
+Haar-like feature descriptor
+============================
+
+Haar-like feature descriptors are a set of simple features which was used in
+the first real-time face detector [1]_. Indeed, these features can be
+efficiently computed using a so-called integral image and by subtracting
```
warmspringwinds: Remove "so-called", just say "integral image [1]" with citation.

emmanuelle:
For requirements, the doc is built for every job of the build matrix, but the doc dependencies are only installed for one of these jobs (corresponding to BUILD_DOCS=1, see .travis.yml and tools/travis/script.sh). I'm not sure I understand the logic, shouldn't we use BUILD_DOCS=1 for all jobs?

glemaitre:
> BUILD_DOCS=1 for all jobs

It makes sense to me if the doc is always built.

doc/examples/features_detection/plot_haar.py
warmspringwinds: Remove "Indeed", just looks better without it :)

And please, address my comment that haar features can be computed on any scale
in constant time and cite the original Viola Jones paper next to that sentence.

warmspringwinds:
@glemaitre Amazing work, I really liked it, especially the feature selection pipeline
with random forest classifier from scikit-learn :+1: 

Let's polish some parts of your contribution and merge it.

These are my comments:

1. Your PR contains three separate contributions: (a) Haar feature computation module (b) Faces non-faces dataset (c) example of feature selection and training a classifier using haar features which depends on (b). Let's merge all of them gradually starting from (a). So, can you just leave the parts of the code that do feature computation and visualization for now and let's merge it if all build checks pass?

2. Please, submit (b) the faces dataset separately and let's make sure we don't violate any license agreements related to the dataset itself (other people from scikit-image can help us with that :)

3. We should totally add your example of feature selection from (c) -- it's very cool :) Create a separate PR for that after (b) is merged or maybe just run it on a couple of images that are already included in scikit image.

@stefanv I approve this PR if 1. is satisfied :) This is a great work.


Some other comments not related to the current PR:

1. I think what you can also do is take a .xml file from OpenCV that was trained for face detection using haar feautures and try to run it using your code? We did it hear but with LBP features:
https://github.com/scikit-image/scikit-image/pull/1570

2. If you are interested, you can have a look at our PR with object detection and borrow some parts or extend that work:
https://github.com/scikit-image/scikit-image/pull/1570
This will make a great impact as it will add a real-time object detection capability to scikit-image.

3. Following your great example for feature selection, you can extend your work and write a training module which will make it possible for users to write their own detectors/ classifiers.

glemaitre:
closed in favor of #2896 #2897 #2898

glemaitre:
Regarding the other suggestions, I might try to find some spare time but I cannot promise :)

PR #2851, opened 2017-10-19
[MRG] DOC fix documentation build

glemaitre:
There is no need for giving the `reference_url` for the sphinx-gallery since intersphinx is used.
It should solve the travis failure when building the doc.


jni:
Thanks @glemaitre!

Issue #2284, opened 2016-09-04
Cannot render `segmentation/plot_ncut.py` correctly


soupault:
## Description
See https://github.com/scikit-image/scikit-image/pull/2144#issuecomment-244626192 and https://github.com/scikit-image/scikit-image/pull/2144#issuecomment-244627673 for the details.

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [x] Code example: see 2nd link above
- [x] Linux x86_64, 4.4.19
- [x] Python version: 3.5.2
- [x] scikit-image version: 0.13dev, `master bd89d6df442cf69dc780cf19f80fa56fb8110f6b`

soupault:
Gallery example looks fine -> http://scikit-image.org/docs/dev/auto_examples/segmentation/plot_ncut.html.

sciunto:
@soupault What do you suspect? Problem with some versions?

soupault:
Dunno. I've checked both for `master` and `0.12.3`.

This is `label1`:
![figure_2](https://cloud.githubusercontent.com/assets/1315589/18256290/f74fff92-73bb-11e6-8caa-a79a85dc8722.png)

Returned `label2` is a matrix of zeros. So, there must be an issue in one of these calls:
```
g = graph.rag_mean_color(img, labels1, mode='similarity')
labels2 = graph.cut_normalized(labels1, g)
```

jni:
There could be some numerical instability in the implementation, which might affect the choice of threshold...

soupault:
@jni what do you see on your setup?

jni:
@soupault 

<img width="421" alt="screen shot 2016-09-08 at 10 43 10 pm" src="https://cloud.githubusercontent.com/assets/492549/18349638/a9483876-7615-11e6-953f-5513b3640188.png">

OSX El Capitan, Anaconda Python 3.5.2, scipy 0.18.0, numpy 0.11.1, matplotlib 1.5.1

sciunto:
Do we postpone to 0.14? Doesn't look critical from here?

soupault:
@sciunto absolutely! Too mystical to solve in a short time :).

soupault:
`graph.cut_normalized(..., max_edge=1.0-1e-12)` | `graph.cut_normalized(..., max_edge=1.0)` | `graph.cut_normalized(..., max_edge=1.0+1e-12)`
------------ | ------------- | -------------
[...]

PR #2856, opened 2017-10-24
Generator version of cut_normalized

magee256:
## Description
Adds a generator version of cut_normalized. When the user sends a threshold value to the generator a set of labels corresponding to that threshold is returned. The generator version reuses previously calculated optimal splits when a new threshold value is supplied. This allows the user to gauge the label quality and adjust the threshold accordingly with greatly reduced computational overhead. 

This pull request also provides minor corrections to the original cut_normalized implementation and improves on the method used to find subgraph partitions. 

## References
Jianbo Shi, & Malik, J. (2000). Normalized cuts and image segmentation. IEEE Transactions on Pattern Analysis and Machine Intelligence, 22(8), 888–905. http://doi.org/10.1109/34.868688

## For reviewers

  later.


skimage/future/graph/graph_cut.py
```diff
@@ -205,7 +284,7 @@ def get_min_ncut(ev, d, w, num_cuts):
     if np.allclose(mn, mx):
         return min_mask, mcut
 
-    # Refer Shi & Malik 2001, Section 3.1.3, Page 892
```
magee256: Reference is for the 2000 paper

skimage/future/graph/graph_cut.py
```diff
@@ -261,21 +340,20 @@ def _ncut_relabel(rag, thresh, num_cuts):
...
         d2 = d.copy()
         # Since d is diagonal, we can directly operate on its data
         # the inverse of the square root
         d2.data = np.reciprocal(np.sqrt(d2.data, out=d2.data), out=d2.data)
 
-        # Refer Shi & Malik 2001, Equation 7, Page 891
+        # Refer Shi & Malik 2000, Equation 7, Page 891
         vals, vectors = linalg.eigsh(d2 * (d - w) * d2, which='SM',
-                                     k=min(100, m - 2))
+                                     k=min(100, m - 1))
```
magee256: Ensures at least two eigenvectors are calculated

skimage/future/graph/graph_cut.py
```diff
@@ -261,21 +340,20 @@ def _ncut_relabel(rag, thresh, num_cuts):
...
 
         # Pick second smallest eigenvector.
-        # Refer Shi & Malik 2001, Section 3.2.3, Page 893
-        vals, vectors = np.real(vals), np.real(vectors)
-        index2 = _ncut_cy.argmin2(vals)
-        ev = vectors[:, index2]
```
magee256: Eigenpairs are returned in ascending order, so eigenvctor corresponding to second smallest eigenvalue will always be at index 1. 

See line 35 here: https://github.com/hpfem/arpack/blob/master/ARPACK/SRC/dseupd.f

skimage/future/graph/graph_cut.py
```diff
@@ -205,7 +284,7 @@ def get_min_ncut(ev, d, w, num_cuts):
     if np.allclose(mn, mx):
         return min_mask, mcut
 
-    # Refer Shi & Malik 2001, Section 3.1.3, Page 892
+    # Refer Shi & Malik 2000, Section 3.1.3, Page 892
```
magee256: Reference is for the 2000 paper

skimage/future/graph/graph_cut.py
```diff
@@ -261,21 +340,20 @@ def _ncut_relabel(rag, thresh, num_cuts):
...
-                                     k=min(100, m - 2))
+                                     k=min(100, m - 1))
 
         # Pick second smallest eigenvector.
-        # Refer Shi & Malik 2001, Section 3.2.3, Page 893
-        vals, vectors = np.real(vals), np.real(vectors)
-        index2 = _ncut_cy.argmin2(vals)
-        ev = vectors[:, index2]
+        # Refer Shi & Malik 2000, Section 3.2.3, Page 893
+        ev = vectors[:, 1]
```
magee256: Eigenpairs are returned in ascending order, so eigenvector corresponding to second smallest eigenvalue will always be at index 1.

See line 35 here: https://github.com/hpfem/arpack/blob/master/ARPACK/SRC/dseupd.f

jni:
@magee256 thanks! I haven't had a chance to look through the code in detail yet, but do you think you could write a small gallery example?

magee256:
Sure. Can I extend the Normalized Cut example already there or should I make my own? This is effectively a different interface for cut_normalized that makes searching for threshold values easier.

jni:
@magee256 whatever you think makes more sense. I have a slight preference for modifying the existing one, but either option is fine by me!

skimage/future/graph/graph_cut.py
```diff
@@ -138,6 +138,82 @@ def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
...
+
+    Examples
+    --------
+    >>> from skimage import data, segmentation
+    >>> from skimage.future import graph
+    >>> img = data.astronaut()
+    >>> labels = segmentation.slic(img, compactness=30, n_segments=400)
+    >>> rag = graph.rag_mean_color(img, labels, mode='similarity')
+    >>> 
+    >>> thresh = [1e-5]
```
stefanv: Do these values always get pushed one by one?  In that case, this may be a good case for https://docs.python.org/2/reference/expressions.html#generator.send

skimage/future/graph/graph_cut.py
```diff
@@ -292,3 +367,80 @@ def _ncut_relabel(rag, thresh, num_cuts):
     # Assign `ncut label` by picking any label from the existing nodes, since
     # `labels` are unique, `new_label` is also unique.
     _label_all(rag, 'ncut label')
+
+
+def _ncut_relabel_gen(rag, thresh, num_cuts):
+    """Perform Normalized Graph cut on the Region Adjacency Graph.
+
+    Recursively partition the graph into 2, until further subdivision
```
stefanv: 2 -> two

stefanv (commented):
Thanks for carefully combing over the normalized cut paper and implementation!

1) Did you see any mistakes in the existing skimage implentation
2) Can the existing implementation benefit from this version (i.e., can it be reused internally?)

skimage/future/graph/graph_cut.py
magee256: I don't think I've used that, but reading the docs I think this would be a good case. It felt a little hacky to have the user pass values by appending to a list and this should get around that. I'll make the changes.

magee256:
@stefanv No problem, being able to contribute to a software package is good motivation. 
1. No definite mistakes other than the ones I corrected. I'm not entirely sure the eigenpairs are being calculated the best way though. The implementation gets values for up to 100 eigenpairs, but the normalized cuts algorithm only needs the first two. When I tried reducing the number of eigenpairs calculated I got convergence issues though, so I think there may be a good reason for calculating that many. Directly calculating the eigenpairs instead of using the Lanczos algorithm for small m might speed things up, but I haven't checked. 
2. Yes, cut\_normalized and \_ncut\_relabel can both be rewritten in terms of the generator versions. Should \_ncut\_relabel be replaced entirely?

stefanv:
Strange, the eig convergence issue; it'd be good to get to the bottom of that.

magee256:
@stefanv I tried messing with eigsh some more. I can request 2 eigenpairs, but if I use any less than the max number of Lanczos vectors I get convergence issues when running the gallery example.

doc/examples/segmentation/plot_ncut.py
```diff
@@ -20,12 +20,26 @@
...
+
+# Getting the value for a single threshold
+labels1 = graph.cut_normalized(super_pixels, g, thresh=1e-3)
+
+# Or using a generator to try several
+label_gen = graph.cut_normalized_gen(super_pixels, g, thresh=1e-4)
+labels2 = next(label_gen)
+labels2 = label_gen.send(1e-3)
+
+# Either method yields the same labels, barring an unlucky random seed
```
stefanv: Let's fix the seed for this example then.

stefanv (commented):
I'd like for @jni and @vighneshbirodkar to take a look at this as well.

doc/examples/segmentation/plot_ncut.py
magee256: Unfortunately doing that isn't as easy as calling np.random.seed(0). The random initialization that seems to be causing the problem is the initial residual vector for the Lanczos algorithm. Reading through the source code for eigsh, it looks like that initialization gets performed by ARPACK in dsaupd.f. Do you know of any way to set that seed? 

If not, the best option I can think of is to generate a random vector in _ncut_relabel and use that to set eigsh's v0 argument. 

Should eigsh be modified to set a random initial residual vector itself? I probably wouldn't have realized np.random.seed wasn't fixing the output if I hadn't already been looking into the stability issues. 

As a side note, ARPACK's random initial vector seems to be reproducible - when I get a bad initial vector, it's bad every time I run the code. Other factors seem to affect the choice of initial vector though, since the stability test and gallery example assert statement sometimes break a Travis CI config and sometimes don't.

magee256: Just to back up the allegation that it's random initialization in ARPACK. When using commit 7168f5f, the labels produced by the cut_normalized_gen and cut_normalized in the gallery example are not the same for me. I was able to make them the same by calling cut_normalized three times in a row. This worked even when setting the numpy seed between each call.

stefanv: If you know the range within which the `v0` is that they generate, then we can certainly set it ourselves.

I suspect SciPy will reject a patch to have v0 be generated by NumPy simply to use the NumPy random seed.  But, it may be worth posting the issue on the mailing list and ask about it.  I can also do that, if you are not subscribed.

Our other option is simply not to compare the results above---i.e., show the API, but don't try to suggest that for different threshold we would get the same results.  That seems like an odd thing to expect anyway?

magee256: I don't see the v0 range mentioned in the docs for eigsh or dsaupd.f, but Wikipedia says the algorithm uses a norm 1 vector so that would probably work. 

I'm not subscribed to mailing list, so it might be better if you send something out. 

Not sure if I understood. I am comparing two separate results for thresh=1e-3 in the gallery example. For cut_normalized_gen I get results for 1e-4 first then overwrite them with results for 1e-3.

magee256:
The tests and gallery example both pass on my computer and the CI test fails at different points depending on version even when setting seed. Unless I messed up somewhere, there's probably another source of randomness not determined by numpy or random's seed, maybe in networkx somewhere. It would be nice to have reproducible region labeling, but I don't think it's that important. I'm leaning towards saying "results may vary" in the docs and just leaving it. Other opinions?

stefanv:
Au contraire, I think this is extremely important!  We should hunt for that source of randomness and pin it down.

magee256:
I should clarify what I mean by reproducible. As far as I can tell, every time I run a script that calls cut_normalized the same number of times with the same sequence of parameter sets I get the same labels. What changes are the labels when I call cut_normalized multiple times with the same parameters. If all those labels are qualitatively similar or "good" then I'd consider the algorithm to be doing its job. There are a few more tests I could run and I should make sure I didn't introduce this issue though, so I'll check that.

magee256:
The CI fails because the stability test hangs. It runs for me w/ Python 3.6 on both Mac and Ubuntu so I'm not sure what's going on. Is there some way I can change the verbosity of the CI messages or clone the environment to test myself?

skimage/_shared/test_rag.py
```diff
@@ -0,0 +1,321 @@
...
+
+@pytest.mark.skipif(not is_installed('networkx'),
+                    reason="networkx not installed")
+def test_cut_normalized_stability():
+    np.random.seed(349)
+    img = data.coffee()
+    labels = segmentation.slic(img, compactness=30, n_segments=400)
+    rag = graph.rag_mean_color(img, labels, mode='similarity')
+
+    raise(Exception('Initialized info'))
```
stefanv: What is happening here?

stefanv:
Travis-CI debugging can be a bit painful.  You can clone an Ubuntu Docker image and experiment, but of course you lose the ability to run the Travis script directly.

[Ignore comments I made on your debugging prints; sorry for the noise]

soupault:
@magee256 for convenience, you could debug everything locally. For instructions see - http://scikit-image.org/docs/dev/contribute.html?highlight=testing#testing

magee256:
@soupault I'd like to, but when I run locally everything works. Do you know if there are scikit-image specific instructions for cloning the Travis CI environments? I haven't used Docker too much and want to be sure I'm doing it right.

soupault:
@magee256 Interesting!
Maybe try this approach - https://github.com/scikit-image/scikit-image/wiki/Testing-in-32-bit-environment ?

magee256:
@soupault Thanks! That'll provide some direction in setting up

stefanv:
@bnaul Do you have a suitable Travis-CI clone image lying around?

bnaul:
I don't have anything special, I saw this recently which describes the steps for running Travis locally: https://medium.com/google-developers/how-to-run-travisci-locally-on-docker-822fc6b2db2e but it looks like the latest images at https://quay.io/repository/travisci/travis-python are 2.5 years old...

magee256:
I think I have it partially working. 

I followed the directions here: 
https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally
Using the Python image and until completing step 7. Then I opened up the `.travis.yml` file in the top directory of the scikit-image repo and executed the commands in the "before-install", "install" and "script" sections in that order. (I think that's what was meant in steps 8 and 9)

That seems to be most of what Travis does, but I wasn't able to reproduce the errors I'm seeing and I'm not sure if my configuration exactly matches. Might take a little more fine tuning.

magee256:
The timing results show the stability test takes longer than 10 minutes to run on Travis CI's servers. My laptop runs the test in a few seconds. Any idea what would cause that large a difference?

magee256:
Not sure why it happened, but it looks like the problem was calling np.linalg.inv and/or np.sqrt on a diagonal 2D array.

magee256:
I was able to get two runs producing different results to run locally and step through. It looks like there are two reasons for the discrepancy: 
- eigsh produces numerically different results with some (solidly non-zero) numbers being as much as double their counterpart's. Comparing the Mac and Linux CI runs, it looks like eigh has this problem as well. 
- RAG edges get weighted differently depending on package versions. Probably just the networkx package, but haven't checked. 

I can check to see if the difference between networkx versions is due to defaults changing, but if it's internal I don't think there's anything I can do. Any ideas on how to try fixing eigsh? Given eigh output differs too, I don't think it's poorly converged eigenpairs.

magee256:
The old implementation has the same problem. It also looks like the smallest eigenvalues are close to zero and fluctuate depending on platform which could change the second smallest eigenvector.

stefanv:
What are these thresholds?  It may make sense to set anything < 1e-15 or so to zero.

stefanv:
Alternatively, make a test that is insensitive to such minor fluctuations.

magee256:
Fluctuation insensitive tests are already there. I'll just remove the stability test. 

The values getting thresholded aren't typically near zero. It's just near-degeneracy causes them to fluctuate a lot. Taking another look at the paper it looks like the wrong values are getting thresholded too, they need to be elementwise divided by d2 first and should be made orthogonal to d. I'll make some changes.

magee256:
I think I'm done making changes. Anything people think should be added or modified?

skimage/future/graph/graph_cut.py
```diff
@@ -258,37 +325,67 @@ def _ncut_relabel(rag, thresh, num_cuts):
...
-            return
-
-    # The N-cut wasn't small enough, or could not be computed.
-    # The remaining graph is a region.
-    # Assign `ncut label` by picking any label from the existing nodes, since
-    # `labels` are unique, `new_label` is also unique.
-    _label_all(rag, 'ncut label')
+
+        # These variables no longer needed
+        del d, d2, w, vals, vectors, ev
```
jni: Do these `del` calls strictly affect performance? I would rather leave memory management to the interpreter if possible...

skimage/future/graph/graph_cut.py
```diff
@@ -122,20 +121,88 @@ def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
...
+    while True:
+        ncut_gen.send(thresh)
 
-    return map_array[labels]
+        map_array = np.zeros(labels.max() + 1, dtype=labels.dtype)
+        # Mapping from old labels to new
+        for n, d in rag.nodes(data=True):
+            map_array[d['labels']] = d['ncut label']
+
+        thresh = yield map_array[labels]
```
jni: If I'm not mistaken, calling `next` on an initialized generator will result in a nasty error, as `thresh` will be set to `None`. Perhaps we can add some kind of check to ensure that `thresh` is valid? I think most users won't be familiar with the `send` interface (it's actually the first time I see it used outside of a contrived tutorial!), so we need to be wary of misuse, more so than in normal functions.

jni (commented):
@magee256 I've pushed some nitpicky corrections to your branch, and have a couple of comments about the existing <del>values</del>**[edit]** code. I've run out of time today to check the math, but will do so in the next iteration. Regarding label consistency, was the concern that the exact integer value of the labels changes from run to run, or that the labelling itself is actually different, ie regions are shaped differently?

skimage/future/graph/graph_cut.py
magee256: The del calls might be premature optimization. `vectors` can get up to 100 x N_{super pixels} and the code generates two `vectors` arrays every time it subdivides a graph. It appears in the body of a generator, so I don't think the interpreter would know it can be deallocated. 

I could put everything after the if statement into a function that returns `mcut` and `cut_mask` and remove the first `del`. The second `del` isn't as important since `cut_mask` is smaller, but I don't think `cut_mask` can be treated the same way.

skimage/future/graph/graph_cut.py
magee256: Send is pretty unfamiliar to me too. Raising an error on None is probably a good idea.

magee256:
@jni The concern was that the regions are shaped differently. I don't think the difference is typically too huge, but it does make a difference.

magee256:
The newest commits should fix the issue referenced in #2284. The algorithm now checks all valid partitions to find the best ncut value. This change improved timings for some of the tests I ran. 

There may still be some numerical stability issues that affect the segmentation, but I don't think there's much that can be done to fix them. Hopefully they won't cause errors as noticeable as the one referenced in #2284.

matthew-brett:

<details>

## Rationale

</details>

Issue #2682, opened 2017-06-13
Add ability to calculate moments from coordinates (not only image)


stefanv:
Currently, scikit-image can only calculate moments (and, by implication, hu moments) from full images.  We should extend this to also work for arrays of coordinates.

- https://mail.python.org/pipermail/scikit-image/2017-June/005264.html
- http://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.moments

PR #2747, opened 2017-08-06
Added function for interpreting and expanding multidimensional parameters.

kne42:
## Description
Will be used in future pull requests for backwards-compatibility and standardized-parameter interpretation for properties of multidimensional functions.

Is additionally used in `filters._gaussian`, `segmentation.slic_superpixels`, and `measure._moments`.

    Provides an expected/standardized output of the parameter as
    an ndarray the size of times. Primarily used in n-dimensional
    image processing.
    
    An argument that is not what NumPy considers to be array-like
    will be repeated the number of times specified. Otherwise,
    the default value will be appended to the argument until
    it reaches a size equivalent to that of the number of times
    specified.
    
    All values of ``None`` within the argument will be replaced with
    the default value.
    
    Parameters
    ----------
    arg : any
        The argument to interpret and standardize.
    times : int
        The number of elements in the output matrix.
    arg_name : str, optional
        Name of the argument in the original function.
    default : any, optional (default: 0)
        The value to default to when None is provided.
    dtype: data-type, optional
        The desired data-type for the array.  If not given, then the type will
        be determined as the minimum type required to hold the objects in the
        sequence.  This argument can only be used to 'upcast' the array.  For
        downcasting, use the .astype(t) method.
    
    Returns
    -------
    expanded_arg : array, shape (``times``,)
        The standardized output of the argument.
    filled : int
        The number of arguments that have been filled in.

## Examples
```py
>>> expand_arg(None, 3)
(array([ 0.,  0.,  0.]), 3)

>>> expand_arg((None, 0, None, 42), 5, default=180)
(array([ 180.,    0.,  180.,   42.,  180.]), 1)

>>> expand_arg((2.6,  2.3), 2, dtype=int)
(array([2, 2]), 0)
```

## References
Will be used in #2739 

## Checklist
- [x] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [x] Unit tests

## For reviewers

(Don't remove the checklist below.)

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.

PR #2859, opened 2017-10-27
Add Computation of Image Moments to Coordinates

kne42:
## Description
Implements the computation of image moments by raw, unweighted points. This set of points takes the shape of `(N, D)`, where `N` is the number of points in the set and `D` is the dimensionality of the image they describe.

## Checklist
- [x] Unit tests

## References
Closes issue #2682.
1. [Johannes Kilian. Simple Image Analysis By Moments. Durham 
           University, version 0.2, Durham, 2001.](http://breckon.eu/toby/teaching/dip/opencv/SimpleImageAnalysisbyMoments.pdf)

## For reviewers

  later.


jni:
So, I realise now that you don't really mean `contours`, but rather, `coordinates`, right?

kne42:
@jni You’re right. I probably was thinking about specific applications while naming this feature. I’ll complete this refactorization when I have the opportunity.

On that note, @scikit-image/core what are your thoughts on how we should distinguish this enhancement from the original moments function? We could keep it as a separate function like it is or we could possibly add a `coords` booleab keyword argument with a default of `False` for whether or not the input array is a set of coordinates or an image.

skimage/measure/_moments.py
```diff
@@ -1,11 +1,137 @@
...
+           [ 20.,   0.,  25.,   0.],
+           [  0.,   0.,   0.,   0.]])
+    """
+    if isinstance(contour, tuple):
+        # Represent as array and rotate to standardize orientation.
+        contour = np.rollaxis(np.asarray(contour), axis=1)
+    assert_nD(contour, 2)
+    ndim = contour.shape[1]
+    if center is None:
+        M = moments_central(contour, center=0, order=order)
```
jni: (a) this should be `moments_contour`?

(b) This is overkill: the centroid is just `np.mean(contour, axis=0)`!

skimage/measure/_moments.py
```diff
@@ -1,11 +1,137 @@
...
+        center = M[tuple(np.eye(ndim, dtype=int))]
+    else:
+        # TODO: Handle broadcasting of `center` argument.
+        if center is int:
```
jni: This is probably not doing what you think it's doing:

```python
In [1]: x = 5

In [2]: x is int
Out[2]: False

In [3]: y = int

In [4]: y is int
Out[4]: True
```

jni: And at any rate, if we want to support scalar centers (which I'm not sure we do), we should allow floats, also.

The function you're after is the builtin `isinstance` or the NumPy function `np.isscalar`.

jni: Actually I just saw what you're doing with `center` below. That will work without modification for scalar centers.

skimage/measure/_moments.py
```diff
@@ -1,11 +1,137 @@
...
+            center = (center,) * ndim
+
+    calc = contour.astype(float)
+    calc -= center
```
jni: One liner: `calc = (contour - center)[..., np.newaxis] ** np.arange(order + 1, dtype=float)`

skimage/measure/_moments.py
```diff
@@ -1,11 +1,137 @@
...
+    >>> cc = M[0, 1] / M[0, 0]
+    >>> moments_contour_central(contour, (cr, cc))
+    array([[ 16.,   0.,  20.,   0.],
+           [  0.,   0.,   0.,   0.],
+           [ 20.,   0.,  25.,   0.],
+           [  0.,   0.,   0.,   0.]])
+    """
+    if isinstance(contour, tuple):
+        # Represent as array and rotate to standardize orientation.
+        contour = np.rollaxis(np.asarray(contour), axis=1)
```
jni: `contour = np.transpose(contour)`

skimage/measure/_moments.py
```diff
@@ -1,11 +1,137 @@
...
+
+    calc = contour.astype(float)
+    calc -= center
+
+    # generate all possible exponents for each axis in the given set of points
+    # produces a matrix of shape (N, D, order + 1)
+    calc = calc[..., np.newaxis] ** np.arange(order + 1)
+
+    # FIXME: Find a better way to do this than DIY `fold` function.
+    def _expand(current, mat):
```
jni: Yeah, as indicated by your comment, I don't understand what you're doing here...!

skimage/measure/_moments.py
```diff
@@ -18,7 +144,7 @@ def moments(image, order=3):
 
     Parameters
     ----------
-    image : 2D double or uint8 array
+    image : nD double or uint8 array
```
jni: Oops! Thanks for fixing these! =)

skimage/measure/_moments.py
```diff
@@ -1,11 +1,137 @@
...
+    >>> cr = M[1, 0] / M[0, 0]
+    >>> cc = M[0, 1] / M[0, 0]
+    >>> cr, cc
+    (14.5, 14.5)
+    """
+    return moments_contour_central(contour, 0, order=order)
+
+
+def moments_contour_central(contour, center=None, order=3):
+    """Calculate all raw image moments up to a certain order.
```
jni: raw -> central

skimage/measure/_moments.py
```diff
@@ -1,11 +1,137 @@
...
+    The following properties can be calculated from raw image moments:
+     * Area as: ``M[0, 0]``.
+     * Centroid as: {``M[1, 0] / M[0, 0]``, ``M[0, 1] / M[0, 0]``}.
+
+    Note that raw moments are neither translation, scale nor rotation
+    invariant.
+
+    Parameters
+    ----------
+    contour : (N, D) double or uint8 array
```
jni: Surely any non-complex data type works here?

skimage/measure/_moments.py
```diff
@@ -1,11 +1,137 @@
...
+           generating a matrix of all possible products of mat's elements."""
+        if mat.size == 0:
+            return current
+        first = mat[0]
+        rest = mat[1:]
+        current = current[..., np.newaxis] * first
+        return _expand(current, rest)
+
+    M = []
+    for point in calc:
```
jni: And this seems very slow...!

jni:
@kne42 ah, sorry, I only saw your comment after submitting my review.

Regarding flags, others will chime in, but I generally like them when the underlying implementations share most code. Currently, they share zero code, so imho it's not appropriate. I like these separate functions.

skimage/measure/_moments.py
kne42: Good catch – also ideally this would implement #2747 so that all similar nD functions that want to "broadcast" certain parameters, there is a single point of control.

skimage/measure/_moments.py
kne42: Think of it like this:
```python
In [1]: import numpy as np

In [2]: def expand(mat): _expand(mat[0], mat[1:])

In [3]: a = np.arange(1, 9).reshape(2, 4)

In [4]: a
Out[4]: 
array([[1, 2, 3, 4],
       [5, 6, 7, 8]])

In [5]: expand(a)
Out[5]: 
array([[ 5,  6,  7,  8],
       [10, 12, 14, 16],
       [15, 18, 21, 24],
       [20, 24, 28, 32]])

In [6]: b = np.arange(1, 13).reshape(3, 4)

In [6]: b
Out[6]: 
array([[ 1,  2,  3,  4],
       [ 5,  6,  7,  8],
       [ 9, 10, 11, 12]])

In [7]: (b[0][..., np.newaxis] * b[1])[..., np.newaxis] * b[2]
Out[7]: 
array([[[ 45,  50,  55,  60],
        [ 54,  60,  66,  72],
        [ 63,  70,  77,  84],
        [ 72,  80,  88,  96]],

       [[ 90, 100, 110, 120],
        [108, 120, 132, 144],
        [126, 140, 154, 168],
        [144, 160, 176, 192]],

       [[135, 150, 165, 180],
        [162, 180, 198, 216],
        [189, 210, 231, 252],
        [216, 240, 264, 288]],

       [[180, 200, 220, 240],
        [216, 240, 264, 288],
        [252, 280, 308, 336],
        [288, 320, 352, 384]]])

```

i.e., given an `(M, N)` matrix, `expand` will generate an `(N, ) * M` matrix of all possible products of the elements of each row multiplied by the elements of the other rows. I am not well-versed in the ways of numpy so if something like this already exists, please let me know!

jni: ah, so it's kind of like reducing the column vectors with `np.outer`...? =) Is M the dimensions or the number of points here???

kne42: @jni In this application of `expand`, `M` is the number of dimensions of the coordinates and `N` is the order of the moments to calculate.

Similar, but not quite the same. While `np.outer` will perform what we wish for two vectors, folding over our array for cases where M > 2 (our n-dimensional applications), it will produce an `(N**M)xN` array, which is the improper shape. However, the numbers are all correct – meaning that we can apply `reshape( (N, ) * M )` to our result. If you think that this might be faster, we can investigate it, but a few quick tests that I ran suggest otherwise.

kne42:
@jni Okay, I've changed the computation to make it as fast as I could manage while also keeping it as readable as possible.

kne42:
@scikit-image/core Still waiting on a review since the requested changes have been made.

ahojnnes:
LGTM, thanks for the contribution and sorry for the delay in reviewing!

stefanv:
Very nice!  Thanks, @kne42.  I notice a remaining "TODO" in the code; if that needs to be tracked, we should make an issue.

kne42:
@stefanv This `TODO` was resolved in #3023

Issue #2514, opened 2017-02-15
Add AppVeyor to wheel builder repository


stefanv:
And produce wheels for 3.4, 3.5, 3.6

https://github.com/scikit-image/scikit-image-wheels

andrewleech:
Now that scipy has official wheel for windows this would be really great!

http://wheels.scipy.org/scipy-1.0.0-cp35-none-win_amd64.whl
https://pypi.python.org/pypi/scipy/1.0.0
https://github.com/scipy/scipy/pull/7616

stefanv:
AppVeyor already builds wheels and saves them, FWIW.

andrewleech:
Ah whoops, I missed there is the appveyor definition file in the repo.  
Badges to link to the travis and appveyor projects would be great in the README.md

What would it take to get these builds onto http://wheels.scipy.org and pypi?

stefanv:
I think adding this to scikit-image-wheels, like you proposed, is the correct solution.  @matthew-brett ?

andrewleech:
Any chance of this getting a closer look?

matthew-brett:
Oh - whoops - sorry - I didn't see this before.  @andrewleech - any chance of a PR to scikit-image/scikit-image-wheels pulling across the appveyor file?

andrewleech:
@stefanv I've added a (possibly naive) AppVeyor config for the wheels builder project.
I haven't used appveyor before (I use gitlab ci for everything) so I expect there to be some basic mistakes... ? 
Does it make sense to add this to your existing AppVeyor account?

PR #2866, opened 2017-11-01
Add travis-ci badge to readme.md

andrewleech:
## Description
Minor update to README.md to add a badge to the current master build on travis-ci.  
This also provides a convenient link to the travis project page so new users know it's active.

It would also be great to get a badge to the appveyor build but it appears you need to be an admin of the appveyor project to get the url of the badge. https://github.com/scikit-image/scikit-image/issues/2514#issuecomment-340942721

## For reviewers

  later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Gitter](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/scikit-image/scikit-image?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 [![codecov.io](https://codecov.io/github/scikit-image/scikit-image/coverage.svg?branch=master)](https://codecov.io/github/scikit-image/scikit-image?branch=master)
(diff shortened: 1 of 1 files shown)


jni:
Hi @andrewleech! We actually had the badge once upon a time but removed it because Travis kept updating their environments and making it look like our project was broken when it wasn't. I think this has become a lot less frequent, so we could re-add it, but I'll let others on the team decide.

andrewleech:
Really, wow that really sucks.  
Gitlab have it so much better with their built in CI, just another reason I don't use github any more than absolutely necessary.

soupault:
>I think this has become a lot less frequent
Not enough. :) Despite the fact that Travis has been delivering more or less stable builds lately (yay, the environment is consistent), I'm still seeing occasional build timeouts quite a lot.
Above said, I don't see how we can benefit from having a badge :).

andrewleech:
Feel free to discard this then, I'm not too concerned either way.  
I guess I'm so used to 100% stable & repeatable builds on gitlab I just assumed travis would be comparable.

soupault:
Ok, then. Let's reloop in 1 year.

PR #2869, opened 2017-11-03
Fix `centroid` and add test coverage.

### LINKED ISSUE #2868: Improper computation of `centroid` in `measure._moments`
I noticed that the test suite `test_moments.py` did not properly cover cases where `center=None` (its default) in `moments_central`. As the adjusted test failed, I decided to create a test for `centroid` (called to compute the center of the image when none was provided) itself.

It failed:
```python
======================================== FAILURES =========================================
______________________________________ test_centroid ______________________________________

    def test_centroid():
        image = np.zeros((20, 20), dtype=np.double)
        image[14, 14] = 1
        image[15, 15] = 1
        image[14, 15] = 0.5
        image[15, 14] = 0.5
        im_centroid = centroid(image)
>       assert_equal(im_centroid, (14.5, 14.5))
E       AssertionError: 
E       Arrays are not equal
E       
E       (mismatch 100.0%)
E        x: array([ 43.5,  43.5])
E        y: array([ 14.5,  14.5])

skimage/measure/tests/test_moments.py:141: AssertionError
```

This bug was introduced in https://github.com/scikit-image/scikit-image/commit/a346268a461ef2a403ed8e8da6e9e446c741e75b.

kne42:
## Description
Fixes the computation of centroid in `centroid` by dividing the sum of each axes' weighted points by the "area" (sum of all weighted points).

## Checklist
- [x] Unit tests

## References
Closes #2868.

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_moments.py
+++ b/skimage/measure/_moments.py
@@ -235,7 +235,9 @@ def centroid(image):
         The centroid of the (nonzero) pixels in ``image``.
     """
     M = moments_central(image, center=(0,) * image.ndim, order=1)
-    center = M[tuple(np.eye(image.ndim, dtype=int))]
+    center = (M[tuple(np.eye(image.ndim, dtype=int))]  # array of weighted sums
+                                                       # for each axis
+              / M[(0,) * image.ndim])  # weighted sum of all points
     return center
 
 
--- a/skimage/measure/tests/test_moments.py
+++ b/skimage/measure/tests/test_moments.py
@@ -1,7 +1,7 @@
 import numpy as np
 from skimage import draw
 from skimage.measure import (moments, moments_central, moments_normalized,
-                             moments_hu)
+                             moments_hu, centroid)
 
 from skimage._shared import testing
 from skimage._shared.testing import assert_equal, assert_almost_equal
@@ -27,6 +27,10 @@ def test_moments_central():
     image[14, 15] = 0.5
     image[15, 14] = 0.5
     mu = moments_central(image, (14.5, 14.5))
+
+    # check for proper centroid computation
+    mu_calc_centroid = moments_central(image)
+    assert_equal(mu, mu_calc_centroid)
 
     # shift image by dx=2, dy=2
     image2 = np.zeros((20, 20), dtype=np.double)
--- a/skimage/measure/tests/test_moments.py
+++ b/skimage/measure/tests/test_moments.py
@@ -1,10 +1,12 @@
+from __future__ import division
 import numpy as np
 from skimage import draw
 from skimage.measure import (moments, moments_central, moments_normalized,
                              moments_hu, centroid)
 
 from skimage._shared import testing
-from skimage._shared.testing import assert_equal, assert_almost_equal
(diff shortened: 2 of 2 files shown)


skimage/measure/_moments.py
```diff
@@ -235,7 +235,9 @@ def centroid(image):
         The centroid of the (nonzero) pixels in ``image``.
     """
     M = moments_central(image, center=(0,) * image.ndim, order=1)
-    center = M[tuple(np.eye(image.ndim, dtype=int))]
+    center = (M[tuple(np.eye(image.ndim, dtype=int))]  # array of weighted sums
+                                                       # for each axis
+              / M[(0,) * image.ndim])  # weighted sum of all points
```
jni: Whoopsies! Thanks for spotting this @kne42! Yikes!

Could you fix the formatting to satisfy PEP8? (Operators should be at the end of the preceding line.)

```python
# weighted sums along each axis divided by weighted sum of all points
center = (M[tuple(np.eye(image.ndim, dtype=int))] /
          M[(0,) * image.ndim])
```

skimage/measure/tests/test_moments.py
```diff
@@ -96,3 +100,13 @@ def test_moments_hu():
...
+
+
+def test_centroid():
+    image = np.zeros((20, 20), dtype=np.double)
+    image[14, 14] = 1
+    image[15, 15] = 1
+    image[14, 15] = 0.5
+    image[15, 14] = 0.5
+    image_centroid = centroid(image)
+    assert_equal(image_centroid, (14.5, 14.5))
```
jni: I actually don't love this test because the result is the same with and without weighting. How about:

```python
image[14, 14:16] = 1
image[15, 14:16] = 1/3
image_centroid = centroid(image)
assert_allclose(image_centroid, (14.25, 14.5)
```

jni: You'll need to add `from __future__ import division` to the top of the file.

skimage/measure/_moments.py
kne42: See #2816

jni: @kne42 I'll be damned. Did PEP8 get updated or did I misread like 5y ago and have been doing it wrong ever since???

skimage/measure/tests/test_moments.py
jni: @kne42 what do you think of this suggestion for the test though? =)

skimage/measure/_moments.py
kne42: @jni Nah, whoever added code W503 to PEP8 probably just didn’t read the [PEP8 write-up](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) or the write-up was modified later but the code was not.

skimage/measure/tests/test_moments.py
kne42: @jni Good point – I’ll change the test to actually account for the weighting.

stefanv:
@kne42 It sounded like you wanted to make the changes @jni suggested, so I'll keep an eye out for more commits before merging.

PR #2877, opened 2017-11-13
Fix randomness and expected ranges for RGB in `test_random_shapes`.

kne42:
## Description
Replaces tests checking for `crop < 255` with `crop <= 255`.

## References
See https://travis-ci.org/scikit-image/scikit-image/jobs/301126609#L8671.
See https://github.com/scikit-image/scikit-image/pull/2879#issuecomment-344099426

## For reviewers

  later.


stefanv:
Why did these tests pass before?

kne42:
It’s a random function, meaning that the tests didn’t always run into the
fringe case where one of the values of a pixel was exactly `255`.

soupault:
Should we set a random state also for `test_can_generate_one_by_one_rectangle`?
Otherwise, LGTM.

kne42:
@jni Replying to https://github.com/scikit-image/scikit-image/pull/2879#issuecomment-344099426, [this line](https://github.com/scikit-image/scikit-image/blob/master/skimage/draw/_random_shapes.py#L196) specifies that the value will be in the interval `[min_pixel_intensity, 255]`. However, since in this test, we specified `min_pixel_intensity=1`, we should be checking that the values are in the interval `[1, 255]`, i.e. `np.shape(crop) == (1, 1, 3) and np.any(crop >= 1) and np.any(crop <= 255)`.

As such, I will update the PR.

stefanv:
@soupault @kne42 Yes, we should find a random seed that failed on the previous version, and use that in the test.

soupault:
@stefanv Sorry, I'd in no case test againt such cherry-picked example.
A proper solution to this issue I see in updating the `random_shapes` function to be more controllable in the sense of generated and background colors (see the last 5 comments in #2773, starting from https://github.com/scikit-image/scikit-image/pull/2773#issuecomment-344160842).

kne42:
@stefanv @soupault Agreed, I am definitely not a fan of testing via setting random seeds. I think the best way to go about doing this is to completely forgo random seeds and instead to take in a random generator. This random generator must comply with a certain abstractions so that it can be used in the function.

The advantages are twofold – not only would users have the option of using their preferred RNGs, but it allows us to pass in a dummy/mock RNG for testing. This grants us absolute control over the parameters, allowing us to make more specific tests, and test if the function itself handles our numbers properly.

However, this would have the downside of making it more difficult for someone to just set the random seed without customizing everything – although I can't imagine why someone would need such a narrow scope of control in the first place.

stefanv:
Tests are failing; missing numpy import.

kne42:
@stefanv Fixed.

soupault:
You are all very welcome to comment on https://github.com/scikit-image/scikit-image/issues/2884 :).

PR #2878, opened 2017-11-13
Nl means fixes for large datasets

blochl:
During a work on a large (600x1000x1000) dataset, I have noticed two problems with NL Means implementation in scikit-image:

1) At large xyz values (bottom right part of the volume, if viewed from above) a lot of noise was introduced. Not only that the data was not denoised there, it was corrupted and unusable.
2) It is extremely slow.

I address both of these problems in this series of patches. Firstly, now the data in NL Means is treated as double (float64) instead of float (float32). This solves the noise introduction issue completely. Secondly, a very substantial fraction of the computation time is being spent on calculating exponents (for determining the weight). An extremely precise result is not needed here, so a good and fast approximation would suffice. I have used the very fast and compact approximation described by Nic Schraudolph[1], to achieve a speedup of more than twice, after verifying that the deviation of this approximation from the math library exp function is not more than 3%.

## References
[1] https://schraudolph.org/pubs/Schraudolph99.pdf

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/fast_exp.h
+++ b/skimage/_shared/fast_exp.h
@@ -1,0 +1,40 @@
+/* A fast approximation of the exponential function.
+ * Reference [1]: https://schraudolph.org/pubs/Schraudolph99.pdf
+ * Reference [2]: http://dx.doi.org/10.1162/089976600300015033
+ * Additional improvements by Leonid Bloch. */
+
+/* use just EXP_A = 1512775 for integer version, to avoid FP calculations */
+#define EXP_A (1512775.3951951856938)  /* 2^20*ln2 */
+/* For min. RMS error */
+#define EXP_BC 1072632447              /* 1023*2^20 - 60801 */
+/* For min. max. relative error */
+/* #define EXP_BC 1072647449 */        /* 1023*2^20 - 45799 */
+/* For min. mean relative error */
+/* #define EXP_BC 1072625005 */        /* 1023*2^20 - 68243 */
+
+__inline double fast_exp (double y)
+{
+    union
+    {
+        double d;
+        struct { int i, j; } n;
+        char t[8];
+    } _eco;
+
+    _eco.n.i = 1;
+
+    switch(_eco.t[0]) {
+        case 1:
+            /* Little endian */
+            _eco.n.j = (int)(EXP_A*(y)) + EXP_BC;
+            _eco.n.i = 0;
+            break;
+        case 0:
+            /* Big endian */
+            _eco.n.i = (int)(EXP_A*(y)) + EXP_BC;
+            _eco.n.j = 0;
+            break;
+    }
+
+    return _eco.d;
+}
--- a/skimage/restoration/_nl_means_denoising.pyx
+++ b/skimage/restoration/_nl_means_denoising.pyx
@@ -493,7 +493,6 @@ cdef inline _integral_image_3d(IMGDTYPE [:, :, ::] padded,
     by avoiding copies of ``padded``.
     """
     cdef int pln, row, col
-    cdef float distance
     for pln in range(max(1, -t_pln), min(n_pln, n_pln - t_pln)):
         for row in range(max(1, -t_row), min(n_row, n_row - t_row)):
             for col in range(max(1, -t_col), min(n_col, n_col - t_col)):
--- a/skimage/restoration/_nl_means_denoising.pyx
+++ b/skimage/restoration/_nl_means_denoising.pyx
@@ -3,12 +3,12 @@ cimport numpy as np
 cimport cython
 from libc.math cimport exp
 
-ctypedef np.float32_t IMGDTYPE
+ctypedef np.float64_t IMGDTYPE
 
 cdef double DISTANCE_CUTOFF = 5.0
 
 @cython.boundscheck(False)
-cdef inline float patch_distance_2d(IMGDTYPE [:, :] p1,
+cdef inline double patch_distance_2d(IMGDTYPE [:, :] p1,
                                     IMGDTYPE [:, :] p2,
                                     IMGDTYPE [:, ::] w, int s):
     """
@@ -27,7 +27,7 @@ cdef inline float patch_distance_2d(IMGDTYPE [:, :] p1,
 
     Returns
     -------
-    distance : float
+    distance : double
         Gaussian distance between the two patches
 
     Notes
@@ -39,11 +39,11 @@ cdef inline float patch_distance_2d(IMGDTYPE [:, :] p1,
     cdef int i, j
     cdef int center = s / 2
     # Check if central pixel is too different in the 2 patches
-    cdef float tmp_diff = p1[center, center] - p2[center, center]
-    cdef float init = w[center, center] * tmp_diff * tmp_diff
+    cdef double tmp_diff = p1[center, center] - p2[center, center]
+    cdef double init = w[center, center] * tmp_diff * tmp_diff
     if init > 1:
         return 0.
-    cdef float distance = 0
+    cdef double distance = 0
     for i in range(s):
         # exp of large negative numbers will be 0, so we'd better stop
         if distance > DISTANCE_CUTOFF:
@@ -56,7 +56,7 @@ cdef inline float patch_distance_2d(IMGDTYPE [:, :] p1,
 
 
 @cython.boundscheck(False)
-cdef inline float patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
+cdef inline double patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
                                        IMGDTYPE [:, :, :] p2,
                                        IMGDTYPE [:, ::] w, int s):
     """
@@ -75,7 +75,7 @@ cdef inline float patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
 
     Returns
     -------
-    distance : float
+    distance : double
         Gaussian distance between the two patches
 
     Notes
@@ -87,8 +87,8 @@ cdef inline float patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
     cdef int i, j
     cdef int center = s / 2
     cdef int color
-    cdef float tmp_diff = 0
-    cdef float distance = 0
+    cdef double tmp_diff = 0
+    cdef double distance = 0
     for i in range(s):
         # exp of large negative numbers will be 0, so we'd better stop
         if distance > DISTANCE_CUTOFF:
@@ -102,7 +102,7 @@ cdef inline float patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
 
 
 @cython.boundscheck(False)
-cdef inline float patch_distance_3d(IMGDTYPE [:, :, :] p1,
+cdef inline double patch_distance_3d(IMGDTYPE [:, :, :] p1,
                                     IMGDTYPE [:, :, :] p2,
                                     IMGDTYPE [:, :, ::] w, int s):
     """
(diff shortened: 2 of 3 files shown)


stefanv:
Thank you for catching the precision issue, and for the optimization workaround.

1. It would be good to write a test to verify the improved accuracy of the new version.  The test should fail under the old float32 implementation.
2. I don't think we can rely on LITTLE_ENDIAN being defined at compile time.  See this page for more information: https://sourceforge.net/p/predef/wiki/Endianness/  Solutions may include populating a function table during runtime.
3. The fast exponent code should probably live under `skimage._shared`.

Thanks again!

soupault:
Would it be better to have `skimage/restoration/fast_exp.h` cythonized?

stefanv:
I don't think it makes much of a difference here; or do you have some specific benefit in mind?

soupault:
@stefanv no, just style consistency. :)

blochl:
Thanks Stefan,

On Mon, Nov 13, 2017 at 8:04 PM, Stefan van der Walt <
notifications@github.com> wrote:

> Thank you for catching the precision issue, and for the optimization
> workaround.
>
>    1. It would be good to write a test to verify the improved accuracy of
>    the new version. The test should fail under the old float32 implementation.
>
> I can attach samples of my dataset to illustrate the issue. Otherwise, a
very large dataset will be needed, and the test may take an unreasonably
long time to pass. The precision problem begins to appear only on quite
large datasets, so I don't see how a test can be made for this.

>
>    1. I don't think we can rely on LITTLE_ENDIAN being defined at compile
>    time. See this page for more information: https://sourceforge.net/p/
>    predef/wiki/Endianness/ Solutions may include populating a function
>    table during runtime.
>
> You are right, I'll think of a solution that won't sacrifice performance
towards this weekend.

>
>    1. The fast exponent code should probably live under skimage._shared.
>
> Yes, you're right.

Regards,
Leonid.

stefanv:
I see, I was hoping to find a small example that had the same problem, but if you can't think of a pathological case, that's fine.

Thanks for working on this!

grlee77:
Another approach for faster special function evaluation I have done in the past is to precompute a table of values over a suitable range and store the values in a look-up table.  Then during the patch comparisons, call a function that uses linear interpolation of the values from within the lookup table rather than `exp()` itself.  That approach could give better accuracy than the approximation proposed here and should be about the same amount of computation.  However, as pointed out by @blochl, high accuracy isn't really needed for the patch weights, so it may not be worth the trouble.

I already have Cython code that builds and retrieves interpolated values in this manner that I can track down if you are interested (it is not currently on-line in a public repository, but can be made available).

stefanv:
If it's not too much work, please do share.

blochl:
Hi Gregory,

On Tue, Nov 14, 2017 at 12:00 AM, Gregory R. Lee <notifications@github.com>
wrote:

> Another approach for faster special function evaluation I have done in the
> past is to precompute a table of values over a suitable range and store the
> values in a look-up table. Then during the patch comparisons, call a
> function that uses linear interpolation of the values from within the
> lookup table rather than exp() itself. That approach could give better
> accuracy than the approximation proposed here and should be about the same
> amount of computation. However, as pointed out by @blochl
> <https://github.com/blochl>, high accuracy isn't really needed for the
> patch weights, so it may not be worth the trouble.
>
If you'll read the paper that this approximation was taken from, you'll see
that it is, in a way, what it does. It actually makes linear interpolation
between lookup table values, without precomputing the actual lookup table.
It does it in a very clever way. One can see the linear interpolation
"signs" on the attached plot, in the form of sharp angles on the
approximated curve.

Regards,
Leonid.

blochl:
Didn't work well with the email... :)
The plot:
expapprox

Same plot on log scale, to illustrate the error behaviour better:
expapprox_log

grlee77:
> If you'll read the paper that this approximation was taken from, you'll see
that it is, in a way, what it does. It actually makes linear interpolation
between lookup table values, without precomputing the actual lookup table.
It does it in a very clever way.

I agree that this approach is reasonable based on the results you have posted (I have not read the paper).  For completeness I made a PR vs. your branch over at blochl/scikit-image#1 with an explicit lookup table implementation.  I did not do any testing of the relative performance vs. the `fast_exp` from this PR, but I would guess it should be pretty similar.

To be clear, I am not insisting that you use the explicit lookup table instead for this PR.  I am fine with either implementation.

grlee77:
@blochl 
I assume that the precision issue you encountered only occurs for the "fast" version of the algorithm?  I see that you have changed both variants to use double precision, but I think only the fast, integral image approach will have a precision that decreases with increasing image size.  Can the same precision fix be achieved using double precision only for the `integral` array used in `_integral_image_2d` and `_integral_image_3d`?  This would allow keeping other variables in float32 to preserve memory.

I think changing the precision in the Cython code will cause non-local means denoising to now always return a  `np.float64` result instead of a `np.float32` one?

Separate from this PR, it might be nice if these transforms had an option to preserve floating point precision of the input.  In current master, running denoising on a `np.float64` input gives a `np.float32` denoised output.  This can be annoying when using metrics such as `compare_psnr` or `compare_ssim` to compare the denoised and original because these functions currently raise errors if the dtypes of the inputs don't match.

blochl:
@grlee77 
No, the precision issues at large datasets were present in the old version, and that is what made me fiddle with it in the first place - not the performance enhancement.
With the new, fast, algorithm it is at all not possible to do single precision (not in the current version).

Also, the old version converted any data to single precision before processing, even if it was double precision to begin with. Now all is converted to double precision, like in several other skimage filters.

grlee77:
> No, the precision issues at large datasets were present in the old version, and that is what made me fiddle with it in the first place - not the performance enhancement.

sorry, I think I was not clear enough in the previous question.  by "fast" version I was not referring to the faster exponential evaluation you have added in this PR, but to running `denoise_nl_means` with the argument `fast_mode=True` (the default).  I don't think you would have had the same precision problem with `fast_mode=False` (but computation may be prohibitively slow for your large dataset in that case).  In `fast_mode=True`, integral images are formed which are likely the source of the precision loss (this is mentioned in the article by Froment et. al. as a drawback of this approach).

>  Now all is converted to double precision, like in several other skimage filters.

I agree that this may be more consistent with the rest of scikit-image, but it is still a behavior change so we should make note of it in the release notes.

blochl:
@grlee77 
Yes, you are right, sorry - I misunderstood your previous comment. It may very well be that the precision issue was affecting only the fast version. And indeed, making only the `integral` array double precision may solve the issue. I did not attempt the slow version, as it would be unreasonably slow. But I think that the behaviour of converting everything to double is more correct and consistent than converting everything to float32. This implementation is not memory intensive (just a small and constant number of arrays are being held in memory) so memory conservation is not the priority here, I think.

Also, I have noticed that the fast exponent approximation gives 280-290% speed boost only when compiled with the `-ffast-math` flag, which I use on my internal branch. When compiled without this flag, it gives only ~10% boost. I do think that it is still worth including though, because if the processing takes several hours, even 10% are substantial. In the future, maybe some parallel implementation may be considered.

blochl:
@stefanv 
>Thanks Stefan,
>>On Mon, Nov 13, 2017 at 8:04 PM, Stefan van der Walt < ***@***.***> wrote:
>>Thank you for catching the precision issue, and for the optimization
>>workaround.
>>
>>1. It would be good to write a test to verify the improved accuracy of
>>the new version. The test should fail under the old float32 implementation.
>
>I can attach samples of my dataset to illustrate the issue. Otherwise, a
>very large dataset will be needed, and the test may take an unreasonably
>long time to pass. The precision problem begins to appear only on quite
>large datasets, so I don't see how a test can be made for this.

Attaching planes with the introduced noise enhanced (*via* edge detection):

bad_denoise_example

>>2. I don't think we can rely on LITTLE_ENDIAN being defined at compile
>>time. See this page for more information: https://sourceforge.net/p/
>>predef/wiki/Endianness/ Solutions may include populating a function
>>table during runtime.
>
>You are right, I'll think of a solution that won't sacrifice performance
>towards this weekend.

Made a function instead of a macro, with runtime check for endianness. This actually performs a bit faster than the original macro in my benchmarks.

>>1. The fast exponent code should probably live under skimage._shared.
>Yes, you're right.

Done.

Regards,
Leonid.

grlee77:
Thanks for providing the image illustrating the problem.  I think that pattern looks consistent with `_integral_image_2d/_integral_image_3d` for `fast_mode=True` being the source of the instability.  These types of cumulative sums will become increasingly inaccurate as image size increases.

> In the future, maybe some parallel implementation may be considered.

I think the `fast_mode=True` variant is harder to implement in parallel than the slow variant at the Cython level.  I think you would need to use something like processing blocks of the image in parallel and then stitch the blocks back together (probably with overlapping blocks so edges of the blocks could be discarded to avoid edge effects).  This seems like something dask's ghost arrays would be ideally suited for.  see:
http://dask.pydata.org/en/latest/array-ghost.html
I think a PR with an option to run blocks in parallel would be well received.  As a secondary effect, this block-wise processing should also improve numerical stability as the integral images would be computed over smaller regions!  I may try this out if I get a chance, but I cannot guarantee a time-frame for completion.

blochl:
@grlee77 
Thanks for the tip regarding Dask's ghost arrays! I was just looking for a way to efficiently split to overlapping blocks.

grlee77:
@blochl:  I had forgotten that there is already an `apply_parallel` in `skimage/util/apply_parallel.py`.  This uses dask.ghost.map_overlap under the hood.

stefanv:
@grlee77 

> This can be annoying when using metrics such as compare_psnr or compare_ssim to compare the denoised and original because these functions currently raise errors if the dtypes of the inputs don't match.

Those functions should then be fixed!

stefanv:
@grlee77 Nice idea---breaking it up into smaller blocks!

stefanv:
@blochl Any reason not to use the `ffast-math` flag when we compile this library?

Otherwise, this looks good to go from my perspective.

grlee77:
@stefanv.  I agree that `compare_psnr`, etc. should be less picky about comparing inputs of different dtype.  I will plan to make a separate small PR for that.

blochl:
@stefanv 
>@blochl Any reason not to use the ffast-math flag when we compile this library?
>
>Otherwise, this looks good to go from my perspective.
>
Thanks.
`-ffast-math` is GCC specific. I can add a few words to the README section about building from source with how to use this flag when building with GCC.

stefanv:
I also ran into this interesting post: https://stackoverflow.com/a/22135559/214686

grlee77:
`-ffast-math` will also work for `clang`.  For `MSVC`, the desired flag is `/fp:fast`.  I think that we would have to set a compiler dependent flag, which may not be trivial.  

Maybe something along the lines of the solution proposed at the link below would work, but I think this is non-trivial and could be left to a separate, future PR:

https://stackoverflow.com/questions/30985862/how-to-identify-compiler-before-defining-cython-extensions

stefanv:
Also, as per the link I provided earlier on: `-ffast-math` comes at a price.  I think reproducibility is more important to us than speed.

blochl:
In this case there is not so much "price" for `-ffast-math`. It does not lose accuracy, at least not noticeably, but the gain is huge, because what takes time here are essentially a lot of very simple operations.
I don't think it's a good idea to include it though, for compatibility reasons.
An elegant, but difficult solution would be to rearrange the loops manually, to make better use of the processor's cache and SIMD. I just didn't wrap my head around it so far.

blochl:
Strange: Travis fails on Python 2.7 build for a reason that looks completely unrelated to these patches (old version of the Six package).

stefanv:
I've fixed that over here: https://github.com/scikit-image/scikit-image/pull/2909

blochl:
Another seemingly unrelated error from CodeCov...

stefanv:
@grlee77 Want to do the honors?

grlee77:
Thanks @blochl

blochl:
Tanks!

PR #2880, opened 2017-11-14
Extension of register_translation to enable subpixel precision in 3D and optionally disable error calculation

vilim:
## Description
I extended the register_translation function to:
- enable 3D subpixel-precision registration, by generalising the _upsampled_dft function
- optionally not calculate the error and phase shift

## Checklist

- [x] Unit tests

## For reviewers

  later.


skimage/feature/register_translation.py
```diff
@@ -59,21 +59,26 @@ def _upsampled_dft(data, upsampled_region_size,
...
+                   (np.arange(upsampled_region_size[0])[:, None] - ax_offset) @
+                   (np.fft.ifftshift(np.arange(n_items))[None,
+                    :] - n_items // 2))
 
+        )
+
+    if data.ndim == 2:
+        einsum_string = 'ij, ki, lj'
+    elif data.ndim == 3:
+        einsum_string = 'ijk, li, mj, nk'
```
jni: Can you add the output strings to this? Explicit is better than implicit. =) Also, could you time the einsum version of the kernel convolution compared to the original `np.dot` version? I suspect `.dot` will be much faster, in which case it would be worth keeping it for the 2D case, and only using einsum for 3D.

skimage/feature/register_translation.py
```diff
@@ -41,7 +41,7 @@ def _upsampled_dft(data, upsampled_region_size,
 
     Returns
     -------
-    output : 2D ndarray
+    output : ND ndarray
```
jni: This should be 2D or 3D, since 4D arrays are not supported.

skimage/feature/tests/test_register_translation.py
```diff
@@ -68,18 +68,14 @@ def test_3d_input():
                                                     shifted_image,
                                                     space="fourier")
     assert_allclose(result, -np.array(shift), atol=0.05)
-    # subpixel precision not available for 3-D data
-    subpixel_shift = (-2.3, 1., 5.)
+    # subpixel precision now available for 3-D data
```
jni: Ha! Nice. =D

vilim:
@jni thanks for the review! I benchmarked the einsum version and it is indeed almost twice slower in the 2D case, so I put the previous version back in. All other minor issues are resolved as well.

vilim:
@jni I think I fixed all the issues now, could it be merged so that it gets included in 0.14?

jni:
@vilim almost done! This needs a test now that 4D images with `upsampled=True` raise a `NotImplementedError`.

Sorry for the delay in reviewing — we've had a nice uptick in contributions and we are struggling to keep up!

Don't stress about it not getting caught in 0.14.0 — 0.14 will be a long-term support release that will get various PRs backported to it. This one would be in the mix. =)

skimage/feature/register_translation.py
```diff
@@ -59,20 +59,49 @@ def _upsampled_dft(data, upsampled_region_size,
...
 
-    return row_kernel.dot(data).dot(col_kernel)
+        return row_kernel.dot(data).dot(col_kernel)
+
+    elif data.ndim == 3:
+        im2pi = 1j * 2 * np.pi
+
+        dim_kernels = []
+
+        dim_kernels = []
```
gatagat: Twice empty?

vilim: It was a typo, sorry. Fixed now

gatagat:
I really like this PR and hope it gets merged soon:)

Here is a comparison of `dot` (+ `reshape` / `permute`) approach for the 3D case with the `einsum`. It does not seem there are too big differences in runtime. For small matrices (20x20x20), the first is up to 1.5x faster, but for larger (200x200x200) both approaches take about the same time, the latter being probably even faster.

Tested with the blobs:
```python
phantom = img_as_float(binary_blobs(length=200, n_dim=3))
reference_image = np.fft.fftn(phantom)
subpixel_shift = (-2.3, 1., 5.)
shifted_image = fourier_shift(reference_image, subpixel_shift)
```
and the the dot version of `_upsampled_dft` (the relevant part):

```python
    elif data.ndim == 3:
        upsampled_region_size = np.asarray(upsampled_region_size, dtype=int)
        im2pi = -1j * 2 * np.pi
        plane_kernel = np.exp(
            (-im2pi / (data.shape[0] * upsample_factor)) *
            (np.arange(upsampled_region_size[0])[:, None] - axis_offsets[0]).dot(
            (np.fft.ifftshift(np.arange(data.shape[0]))[None, :] - np.floor(data.shape[0] / 2)))
        )
        row_kernel = np.exp(
            (-im2pi / (data.shape[1] * upsample_factor)) *
            (np.arange(upsampled_region_size[1])[:, None] - axis_offsets[1]).dot(
            (np.fft.ifftshift(np.arange(data.shape[1]))[None, :] - np.floor(data.shape[1] / 2)))
        )
        col_kernel = np.exp(
            (-im2pi / (data.shape[2] * upsample_factor)) *
            (np.arange(upsampled_region_size[2])[:, None] - axis_offsets[2]).dot(
            (np.fft.ifftshift(np.arange(data.shape[2]))[None, :] - np.floor(data.shape[2] / 2)))
        )
        out = np.reshape(
            plane_kernel.dot(np.reshape(data, (data.shape[0], -1))),
            (upsampled_region_size[0], data.shape[1], data.shape[2]))
        out = np.transpose(out, (1, 0, 2))
        out = np.reshape(
            row_kernel.dot(np.reshape(out, (data.shape[1], -1))),
            (upsampled_region_size[1], upsampled_region_size[0], data.shape[2]))
        out = np.transpose(out, (1, 0, 2))
        out = np.transpose(out, (2, 1, 0))
        out = np.reshape(
            col_kernel.dot(np.reshape(out, (data.shape[2], -1))),
            upsampled_region_size)
        out = np.transpose(out, (2, 1, 0))
        return out
```

vilim:
@jni is there anything I should still do for this to be merged?

skimage/feature/register_translation.py
```diff
@@ -132,6 +159,9 @@ def register_translation(src_image, target_image, upsample_factor=1,
         Defines how the algorithm interprets input data.  "real" means data
         will be FFT'd to compute the correlation, while "fourier" data will
         bypass FFT of input data.  Case insensitive.
+    return_error : bool, optional
```
emmanuelle: would ``return_residual`` be clearer?

emmanuelle: I'm not sure whether we have other functions returning residual errors, if yes we should be consistent with names.

emmanuelle:
Having subpixel accuracy in 3D is awesome, we should really get this PR merged. My only concerns are
- the large memory footprint of einsum with ``optimize=True``: testing on a 500^3 float64 images, 16 Gb of RAM were used (for a 1 Gb image, it's a lot. 8x the size of input data - img and target - happens for other functions such as watershed, but I'm wondering if it's possible to do better?)
- maybe add a comment explaining the sums which are computed?

emmanuelle:
Hum, in fact the large memory footprint is the same for ``upsample_factor=1``, so it's not the ``einsum``. +1 for merging.

emmanuelle:
The issues with Travis could probably be solved by merging master (or rebasing if you dare).

hmaarrfk:
Emanualle, part of the huge memory usage is due to these 4 lines:
https://github.com/scikit-image/scikit-image/blob/9195bea8337625cdc721052cdf0b23c300fa5b62/skimage/feature/register_translation.py#L171-L174

There is no need to do the explicit cast as complex128, that is a "premature optimization" in my mind. Furthermore, `src_image` is always in scope which doesn't allow the python garbage collector to free it.

That is probably the **easiest** place to save money without sacrificing readability.

Finally, there are a few more things that are probably kept in scope longer than they need to. but that one is especially troubling since float32 -> complex128, which can cause a 4x increase in the memory usage right away!

emmanuelle:
@hmaarrfk good catch! I agree that these two lines are not necessary and are harmful in terms of memory consumption. I run a quick test without them and the memory footprint decreases by 1/3! @vilim if you don't mind removing these two lines that'd be awesome, we can also do it in another PR (I checked that tests still pass without them, which seemed obvious, but you never know :-)).

hmaarrfk:
@emmanuelle you can probably make a seperate PR for it. It doesn't seem like git will have a conflict for it. Since the changes are much less dramatic, it is probably a good idea.

In that PR, maybe also have a memory benchmark?

vilim:
@emmanuelle , @hmaarrfk thank you for looking into this. I removed the lines and merged the current master.

emmanuelle:
Thank you for the quick update @vilim. Travis fails with minimal dependency, because the ``optimize`` keyword was introduced in more recent versions of numpy. Maybe catch the ValueError here and run without the optimize keyword? (It's very ugly I know, if someone else has a better suggestion... but I'm reluctant to bump numpy dependency just because of this).

vilim:
I removed optimize=True, as it is the default since numpy 1.14, and I think in this case suboptimal performance is acceptable if old versions of libraries are used.

emmanuelle:
@jni can you review this pr once again and merge if you're happy?

jni:
Sorry for the (wow) long delay, @vilim!

I'm happy with the changes but I think there's something wrong with the test, since this is happening across all our builds:

```
skimage/feature/tests/test_register_translation.py ....
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

@vilim can you run the tests locally? Is this test actually taking 10min to run?

vilim:
For some reason removing optimize=True even on numpy 1.14 causes the very long delay. I put it back and caught the ValueError if using very old versions of numpy

vilim:
@jni so the case without optimize=True fails with numpy 1.11, which doesn't support the argument.
I don't know if this is acceptable missing functionality for an old version of numpy, or how it should be handled. I guess the other option is to provide the @gatagat's version of the algorithm with the transposes and reshapes.

jni:
@vilim is it just obscenely slow? Or does it hang indefinitely? Have you tried running it locally and measuring how long it takes? If it's a major slowdown, I'm happy to just add a skip to that test for numpy 1.11 (and perhaps a warning). If it's an indefinite hang, it would be good to get to the bottom of it...

hmaarrfk:
The `has_keyword ` function from dask might be useful
https://github.com/dask/dask/blob/master/dask/utils.py#L1020

vilim:
@jni on my computer, without optimize=True it takes a full 23 minutes, 24 seconds, but it passes

jni:
@vilim I'm gonna go ahead and feel smug about my earlier concerns that eigsum might have performance issues. :joy: 

Yes, if you skip + warn, I think this should be ready to go! Actually, could you add a super-tiny example to test that the warning is raised correctly? Search the codebase for `expected_warnings` for usage examples.

stefanv:
NumPy 1.11 is already two years old, so why don't we bump the dependency to 1.12?

skimage/feature/register_translation.py
```diff
@@ -58,20 +58,50 @@ def _upsampled_dft(data, upsampled_region_size,
...
-             np.arange(upsampled_region_size[1])[None, :] - axis_offsets[1])
-    )
-    row_kernel = np.exp(
-        (-1j * 2 * np.pi / (data.shape[0] * upsample_factor)) *
-        (np.arange(upsampled_region_size[0])[:, None] - axis_offsets[0]).dot(
-            np.fft.ifftshift(np.arange(data.shape[0]))[None, :] -
-            np.floor(data.shape[0] / 2))
-    )
+    # For the 2D case, it is more efficient to use the well-optimized
+    # dot product instead of einsum
```
stefanv: @mattip Could you double check this optimization for us?  I.e., is this the correct approach, giving optimizations that have recently been made inside of NumPy.

mattip: Benchmarking will give the most authoritative answer, and also will allow you to see if something changes in the future. 

In general, einsum has more parsing overhead, and depends on hand-written SSE routines to vectorize operations, where dot use BLAS-type calls, which could be OpenBLAS, Atlas, or MKL depending on the linalg backend.

vilim:
@jni everything should be ok now, however, Travis in one of the configurations outputs strange matplotlib-related errors, which seem to exists in the master branch as well.

emmanuelle:
Indeed Travis's failure is not related to your PR. It took us one year, but I'm merging! Thanks a lot!

jni:
@vilim thank you for this great contribution and for your patience! =)

PR #2882, opened 2017-11-14
DOC: use Gaussian noise in non-local means example

grlee77:
## Description
When looking at the non-local means example, I noticed that it is adding noise that is distributed uniformly in [0, 1.] .  This results in the "noisy" image being brighter than the original.  I think it was probably intended to use Gaussian noise instead.  This PR makes the following two modifications:

1.) use normally distributed noise
2.) adjust the `h` parmeter as needed to give a reasonable denoising result with the new noise settings

Here is the result I get prior to this PR:
result_master0

and here is after:
result_master

Also, there is a relatively simple modification to the weight calculation in non-local means that can further improve the denoising performance.  I hope to make a separate PR for that soon.  

## For reviewers

  later.

### DIFF
--- a/doc/examples/filters/plot_nonlocal_means.py
+++ b/doc/examples/filters/plot_nonlocal_means.py
@@ -21,10 +21,11 @@ from skimage.restoration import denoise_nl_means
 astro = img_as_float(data.astronaut())
 astro = astro[30:180, 150:300]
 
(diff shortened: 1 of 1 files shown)


stefanv:
We should check with the original author, @emmanuelle .

grlee77:
This fix is incorporated into the expanded example in #2890, so I am closing this one.

PR #2886, opened 2017-11-20
V0.13.x

biaojiang:
To reduce the image size using transform.resize. However, the result was different from scipy.misc.imresize, the latter was much smoother than yours. What's the problem? the two methods using the same algorithms of interpolation filters.

Skimage.transform.resize produced the similar results as Opencv.resize.

Pillow 4.3, Python 3.6.3, the codes are as following, both use the bilinear filter, then face1 is much smoother than face2.
`from scipy.misc import face, imresize

face = face(gray=True)

face1 = imresize(face, 0.10) / 255

face2 = resize(face, (face.shape[0]//10, face.shape[1]//10), mode= 'reflect')`


soupault:
@biaojiang what is `face` in your example? Could you also provide the processed images for the comparison?

biaojiang:
face is a function imported from scipy.misc, from which a gray-scale image of a fox face can be generated.
The following is the results comparison, the original face image was scaled to 10%, both of the resize functions used the bilinear interpolator:
scipy_misc_imresize_vs_skimage_resize

It can be seen that the result of `skimage.transform.resize` is more grained, not smooth. 

Finally, I realized it is because `skimage.transform.resize` and `opencv.resize` haven't included the anti-aliasing filter, but misc.imresize and the Matlab function `resize` use anti-aliasing filter before interpolating. So anti-aliasing filter is very important in resizing an image.
In the skimage document, it said the new released version will include the anti-aliasing filter. Then the output will be similar to Matlab's.

soupault:
@biaojiang yes, it has been added in https://github.com/scikit-image/scikit-image/pull/2802.
Could you install the latest `scikit-image` version from GitHub (`pip install git+git://github.com/jkbr/httpie.git`), and check if the results are as expected?

biaojiang:
Now I built the latest version of 0.14.dev0, then the results are as expected, the two methods are almost the same.
Good implementations!
scipy_misc_imresize_vs_skimage_resize_014

soupault:
Perfect!
FYI, we're having the release of 0.14 at the beginning of the next year. Stay tuned :)

PR #2890, opened 2017-11-22
noise-robust patch distance estimation for non-local means

grlee77:
## Description

This PR adds an additional optional parameter, `sigma`, to non-local means denoising.  This can be used to provide an estimate of the noise variance for use in a more robust determination of patch weights (i.e. the expected noise variance will be subtracted out of the estimates of patch distances).  This is the form discussed in section 2 of the paper given under References below (it was not proposed in the original non-local means paper).

The default of `sigma=0` matches the previously existing behavior.  The non-local means demo has been updated to show the relative PSNR achieved by both the fast and slow variants of non-local means when used with and without `sigma`.  The optimal choice of `h` is a bit different in each case and was hand-tuned in the demo to provide a best case for each variant tested.

Here is the output of the modified demo:

non_local_comparison

The quantitative PSNR numbers for each denoising case are:
PSNR (fast) = 28.38
PSNR (fast, using sigma) = 29.26
PSNR (slow) = 29.34
PSNR (slow, using sigma) = 29.68

So, providing `sigma` improves the PSNR of the denoised image, particularly for the fast variant of the algorithm.

## Checklist
- [x] Unit tests

## References

A. Buades, B. Coll, & J-M. Morel. Non-Local Means Denoising.
Image Processing On Line, 2011, vol. 1, pp. 208-212.
DOI: 10.5201/ipol.2011.bcm_nlm

## For reviewers

  later.

### DIFF
--- a/doc/examples/filters/plot_nonlocal_means.py
+++ b/doc/examples/filters/plot_nonlocal_means.py
@@ -54,19 +54,19 @@ patch_kw = dict(patch_size=5,      # 5x5 patches
                 multichannel=True)
 
 # slow algorithm
-denoise = denoise_nl_means(noisy, h=1.15*sigma_est, fast_mode=False,
+denoise = denoise_nl_means(noisy, h=1.15 * sigma_est, fast_mode=False,
                            **patch_kw)
 
 # slow algorithm, sigma provided
-denoise2 = denoise_nl_means(noisy, h=0.8*sigma_est, sigma=sigma_est,
+denoise2 = denoise_nl_means(noisy, h=0.8 * sigma_est, sigma=sigma_est,
                             fast_mode=False, **patch_kw)
 
 # fast algorithm
-denoise_fast = denoise_nl_means(noisy, h=0.8*sigma_est, fast_mode=True,
+denoise_fast = denoise_nl_means(noisy, h=0.8 * sigma_est, fast_mode=True,
                                 **patch_kw)
 
 # fast algorithm, sigma provided
-denoise2_fast = denoise_nl_means(noisy, h=0.6*sigma_est, sigma=sigma_est,
+denoise2_fast = denoise_nl_means(noisy, h=0.6 * sigma_est, sigma=sigma_est,
                                  fast_mode=True, **patch_kw)
 
 fig, ax = plt.subplots(nrows=2, ncols=3, figsize=(8, 6),
--- a/doc/examples/filters/plot_nonlocal_means.py
+++ b/doc/examples/filters/plot_nonlocal_means.py
@@ -22,6 +22,9 @@ distances.  This can lead to a modest improvement in image quality.
 
 The `estimate_sigma` function can provide a good starting point for setting
 the `h` (and optionally, `sigma`) parameters for the non-local means algorithm.
+`h` is a constant that controls the decay in patch weights as a function of the
+distance between patches.  Larger `h` allows more smoothing between disimilar
+patches.
 
 In this demo, `h`, was hand-tuned to give the approximate best-case performance
 of each variant.
--- a/doc/examples/filters/plot_nonlocal_means.py
+++ b/doc/examples/filters/plot_nonlocal_means.py
@@ -96,10 +96,10 @@ fig.tight_layout()
 
 # print PSNR metric for each case
 psnr_noisy = compare_psnr(astro, noisy)
-psnr = compare_psnr(astro, denoise.astype(astro.dtype))
-psnr2 = compare_psnr(astro, denoise2.astype(astro.dtype))
-psnr_fast = compare_psnr(astro, denoise_fast.astype(astro.dtype))
-psnr2_fast = compare_psnr(astro, denoise2_fast.astype(astro.dtype))
+psnr = compare_psnr(astro, denoise)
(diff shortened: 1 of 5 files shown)


grlee77:
If this PR is accepted it will become unnecessary to merge #2882 as the modifications to the demo here also use Gaussian noise.

doc/examples/filters/plot_nonlocal_means.py
```diff
@@ -10,32 +10,98 @@
...
+applied to the patches when computing patch distances.  When `fast_mode` is
+`True` a faster algorithm employing uniform spatial weighting on the patches
+is applied.
+
+For either of these cases, if the noise standard deviation, `sigma`, is
+provided, the expected noise variance is subtracted out when computing patch
+distances.  This can lead to a modest improvement in image quality.
+
+The `estimate_sigma` function can provide a good starting point for setting
+the `h` (and optionally, `sigma`) parameters for the non-local means algorithm.
```
stefanv: Can you add a word or two expanding on `h`, since a user looking at the gallery typically won't know what it is.

ahojnnes:
Looks fantastic, great job!

stefanv:
Minor tweak requested, then ping me and I'll merge.

emmanuelle:
Awesome, thanks :-)

grlee77:
ping @stefanv.  I think this should be ready now.

grlee77:
This was rebased using double precision and `fast_exp` now that #2878 has been merged. 

@blochl: you may also be interested in this PR for your application.  As shown in the example above, the sigma parameter seems to provide a substantial benefit in quality for the fast mode.  It does involve one additional subtraction during computation of the integral images, but I don't think this will cause a noticeable loss in performance.

blochl:
Thanks, @grlee77 .
I'll give it a try!

stefanv:
Any idea where the failure comes from?  It looks unrelated, but is also from the restoration module.

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -394,12 +388,12 @@ cdef inline double _integral_to_distance_2d(IMGDTYPE [:, ::] integral,
     distance = max(distance, 0.0) / h2s2
     return distance
 
-
+# @cython.wraparound(False)
 @cython.cdivision(True)
 @cython.boundscheck(False)
-cdef inline double _integral_to_distance_3d(IMGDTYPE [:, :, ::] integral,
```
blochl: Is there a reason why the decorators were left here, and the commented one added?

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -369,10 +365,8 @@ def _nl_means_denoising_3d(image, int s=7, int d=13, double h=0.1,
 #-------------- Accelerated algorithm of Froment 2015 ------------------
 
 
-@cython.cdivision(True)
-@cython.boundscheck(False)
-cdef inline double _integral_to_distance_2d(IMGDTYPE [:, ::] integral,
-                        int row, int col, int offset, double h2s2):
+cdef inline double _integral_to_distance_2d(
+        IMGDTYPE [:, ::] integral, int row, int col, int offset, double h2s2):
```
blochl: Won't the following be more PEP8 compliant, and also clearer (in one similar case below as well):
```
cdef inline double _integral_to_distance_2d(IMGDTYPE [:, ::] integral,
                                            int row, int col, int offset,
                                            double h2s2):

```

skimage/restoration/_nl_means_denoising.pyx
grlee77: good catch.  that was unintentional.

grlee77:
> Any idea where the failure comes from? It looks unrelated, but is also from the restoration module.

No, that worried me a bit too.  Adding those directives at the top of the file shoud not be doing anything to the directives used for the other files/modules.  I am not sure why it would only fail in that single test case, but I restarted that case on the CI yesterday and it still failed at the same point.

I am not sure what to try in terms of resolving it as there are not edits made to any files related to the phase unwrapping that failed.  I can roll back to using decorators on each individual function just to see if that makes any difference, but ideally it should not.

grlee77:
@stefanv
Well, I didn't really change anything, but it somehow passed all test cases on Travis this time around.  This should be good to merge now.

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -391,9 +394,10 @@ cdef inline double _integral_to_distance_2d(
 # @cython.wraparound(False)
 @cython.cdivision(True)
 @cython.boundscheck(False)
-cdef inline double _integral_to_distance_3d(
```
blochl: The decorators are still here - I guess they should be removed before merging.

blochl:
I think that cf2d636 and b48daed are redundant: the changes introduced there should be a part of the two previous commits (Cython and PEP8).

grlee77:
> I think that cf2d636 and b48daed are redundant: the changes introduced there should be a part of the two previous commits (Cython and PEP8).

I have now rebased, squashing thoset two commits into the prior PEP8 commit.

stefanv:
Thank you!

Issue #2853, opened 2017-10-22
Frequency response of `iradon()` filters differs greatly from the same filters in MATLAB


TheChymera:
## Description
The frequency response I get with different filters from `skimage.transform.iradon()` is not consistent with the results from MATLAB. 

As `scikit-image` does not return the frequency response like MATLAB does, I have simply isolated the code which produces the filter from [the relevant site in `iradon()`](https://github.com/scikit-image/scikit-image/blob/5a0c5a5f03438ae483b2fc9612c9d634c74c2786/skimage/transform/radon_transform.py#L222) in order to create the example below.

## Way to reproduce
Run:

```
import matplotlib.pylab as plt
import numpy as np
from scipy.fftpack import fftfreq
from skimage import data_dir
from skimage.io import imread
from skimage.transform import iradon, radon

def frequency_response(filter, radon_image):
	projection_size_padded = \
        max(64, int(2 ** np.ceil(np.log2(2 * radon_image.shape[0]))))
	f = fftfreq(projection_size_padded).reshape(-1, 1)
	omega = 2 * np.pi * f                                # angular frequency
	fourier_filter = 2 * np.abs(f)                       # ramp filter
	if filter == "ramp":
		pass
	elif filter == "shepp-logan":
		# Start from first element to avoid divide by zero
		fourier_filter[1:] = fourier_filter[1:] * np.sin(omega[1:]) / omega[1:]
	elif filter == "cosine":
		fourier_filter *= np.cos(omega)
	elif filter == "hamming":
		fourier_filter *= (0.54 + 0.46 * np.cos(omega / 2))
	elif filter == "hann":
		fourier_filter *= (1 + np.cos(omega / 2)) / 2
	elif filter is None:
		fourier_filter[:] = 1
	else:
		raise ValueError("Unknown filter: %s" % filter)

	return fourier_filter

image = imread(data_dir + "/phantom.png", as_grey=True)
theta = np.linspace(0., 180., max(image.shape), endpoint=False)
sinogram = radon(image, theta=theta, circle=True)

filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']
fig, ax = plt.subplots(1, len(filters), figsize=(10,2))

for ix, f in enumerate(filters):
	reconstruction_fbp = iradon(sinogram, theta=theta, circle=True, filter=f)
	response = frequency_response(f, sinogram)
	ax[ix].plot(response)
	ax[ix].set_title('{} FBP'.format(f))
[...]

PR #2895, opened 2017-11-23
Changing iradon filters to match those of MATLAB and Octave

ASchuberth:
Original issue: https://github.com/scikit-image/scikit-image/issues/2853

## Description
The filters currently being used do not match the MATLAB or Octave filters.  Output of the current program gives:

oldfilters

------

The proposed changes give an output of:

newfilters

which matches MATLAB and Octave.  The source code of the rho_filter function in Octave was used to generate them.

------ 

These changes pass all the tests.  However, as I was researching these filters, I noticed that they are shown in this format:
paperfilters

For example, [rice.edu notes](https://www.clear.rice.edu/elec431/projects96/DSP/filters.html) and in this [paper](https://www.researchgate.net/publication/251915833_A_prototype_system_for_infrared_computed_tomography_for_image_reconstruction).  These are the same filters, but with np.fft.fftshift performed on them.  I feel like these are what we should use, but they fail the test_radon_transform tests.  Does anyone know why we use the other shifted filters instead of these?  Octave appears to do the same thing.  I haven't tested these filters with test phantoms yet to check the difference in output.

------- 
The current filters in this branch all in the same graph for reference:
allfilters
 

## Checklist
- [ ] Unit tests

## References
This will close issue #2853.
[Octave rho_filter Source Code](https://sourceforge.net/p/octave/image/ci/default/tree/inst/rho_filter.m#l130)
[Rice.edu Notes](https://www.clear.rice.edu/elec431/projects96/DSP/filters.html)

## For reviewers

  later.


matthew-brett:
Hi - just checking - "The source code of the rho_filter function in Octave was used to generate them." - you mean generate the figures and not the code right?  Otherwise we'd have a license problem.

oachkatzlschwoaf123:
I referenced the rho_filter source code and edited the iradon function code in radon_transform.py to match the rho_filter code output.  Is that okay? Is there a reference in the code needed?  All of the figures here were generated from the original iradon code, the code from the OP of  issue #2853 , and edits that I made.

Thanks for catching that. I should have phrased that differently.

oachkatzlschwoaf123:
Disregard what I asked earlier about the filters.  The way I have it coded is the way it needs to be.

skimage/.cache/v/cache/lastfailed
```diff
@@ -0,0 +1,23 @@
+{
```
soupault: The cache files must be removed.

skimage/transform/.cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
soupault: This one also :)

stefanv:
Can we add a test to ensure that this doesn't break again?  E.g., generate an impulse response of the filter, and verify that Shepp-Logan has two peaks, or peaks at the expected locations, etc.

ASchuberth:
Yeah, I'll try to think of something.  Sorry for the late response.  I have been pretty busy, but I should have time next week to work on this.

kczimm:
Any update on this @ASchuberth? I've avoided using scikit-image due to the image bias incurred from the wrong recon kernels. I would love to start using it once the kernel bug is corrected. Thanks in advance for this effort.

stefanv:
@kczimm This PR is almost there, it just needs a test; is this something you can help with?

kczimm:
@stefanv I can try to help. I've never contributed to this project. At the risk of exposing my naïveté, how do I execute the tests in the test_radon_transform.py file?

soupault:
@kczimm see http://scikit-image.org/docs/dev/contribute.html#testing

kczimm:
After some inspection, the kernels still have a bias problem. But this is no surprise to me because they were modeled from Octave and that implementation has the same bias problem but MATLAB does not. The following script reconstructs a uniform circular phantom and plots the line profile through the center.

```
from __future__ import print_function, division

import numpy as np

from skimage.transform import iradon, radon

pixels = 512
xy = np.arange(-pixels/2, pixels/2) + 0.5
x, y = np.meshgrid(xy, xy)
image = x**2 + y**2 <= (pixels/4)**2

sinogram = radon(image, theta=theta, circle=True)

reconstruction_fbp = iradon(sinogram, theta=theta, circle=True)
error = reconstruction_fbp - image

plt.plot(reconstruction_fbp[int(pixels/2), :])
plt.plot(image[int(pixels/2), :])
plt.show()
```

If you zoom in on the uniform region of the phantom or the background, there is an approximate ~0.5% bias. While the reconstruction looks qualitatively correct, it is biased to a non-trivial degree. The zero frequency of the kernels needs to be adjusted to preserve the DC component of the image.

bias

The ramp filter should be implemented in the spatial domain first as in Kak and Slaney p.72 (https://engineering.purdue.edu/~malcolm/pct/CTI_Ch03.pdf)

PR #2897, opened 2017-11-23
Add CBCL face database subset to `skimage.data`

glemaitre:
## Description

## Checklist
- [x] Unit tests

## References

## For reviewers

  later.


glemaitre:
Added CBCL face dataset from: http://cbcl.mit.edu/software-datasets/FaceData2.html
Used in Haar-like feature example

ping @warmspringwinds @stefanv

glemaitre:
Uhm interesting failing test. I am not sure this is related. Is it a flaky test?

jni:
Yes this is getting fixed in #2877, sorry for the noise! =)

skimage/data/__init__.py
```diff
@@ -384,3 +385,46 @@ def stereo_motorcycle():
...
+    MIT. It should not be used for commercial purposes without specific
+    permission from the authors and MIT. MIT also requires written
+    authorization by the author(s) to publish results obtained with the data or
+    software and possibly citation of relevant CBCL reference papers.
+
+    We make no representation as to the suitability and operability of this
+    data or software for any purpose. It is provided "as is" without express or
+    implied warranty.
+
+    References
```
soupault: Please, add a link to the full database.

soupault: Please, add a link to the full database.

skimage/data/__init__.py
```diff
@@ -384,3 +385,46 @@ def stereo_motorcycle():
     return (load("motorcycle_left.png"),
             load("motorcycle_right.png"),
             _np.load(_os.path.join(data_dir, "motorcycle_disp.npz"))["arr_0"])
+
+
+def cbcl_database():
```
soupault: Maybe rename to something more explicit? For example, `cbcl_faces`, `cbcl_face_database`?

soupault:
I think you have occasionally removed the dataset itself.
Also, there are naming inconsistency: `cbcl_face_dataset`, `cbcl_face_database`.

glemaitre:
Ups ... Good catch

emmanuelle:
LGTM, merging. Thanks!

stefanv:
This dataset is not free for commercial use, so I don't think we can include it.

glemaitre:
I might be able to use http://vis-www.cs.umass.edu/lfw/ which is available in scikit-learn.
But I would need a non-face dataset.

Shall we revert this PR

warmspringwinds:
@glemaitre any patch from lfw that has an intersection over union < 0.5 with the face rectangle can be used as a negative example.

stefanv:
@glemaitre Would that work?  If so, I think it's a good solution.

glemaitre:
> @glemaitre Would that work? If so, I think it's a good solution.

After a try this is possible, I think. I used the following script to create a small dataset using LFW

```python
import numpy as np
from sklearn.datasets import fetch_lfw_people

dataset = fetch_lfw_people(resize=1.0,
                           slice_=(slice(0, 250),
                                   slice(0, 250)))
images = dataset.images
indexes = np.random.randint(0, images.shape[0], size=100)

faces = []
non_faces = []
selected = []
for im_idx in indexes:
    im = images[im_idx] / 255
    faces.append(resize(
        im[70:195, 70:195],
        output_shape=(25, 25),
        anti_aliasing=False))
    non_faces.append(resize(
        im[0:50, 0:50],
        output_shape=(25, 25),
        anti_aliasing=False))
    selected.append(im)

faces = np.array(faces)
non_faces = np.array(non_faces)

np.save('cbcl_database.npy', np.concatenate((faces, non_faces)), axis=0)
```

and I obtain the following results:

```
             precision    recall  f1-score   support

          0       0.96      0.96      0.96        27
          1       0.96      0.96      0.96        23

avg / total       0.96      0.96      0.96        50
```

whatever

However, I would be in favor of making a small npy file as it is now such that we don't have to load the full dataset as in scikit-learn. It is true that we could cache it but I would think that it would be enough for illustration.

warmspringwinds:
Looks good. Can you show us some examples of non-faces here?

@stefanv what do you think about the lfw and a subset of it?

@glemaitre I will have a final look at your feature PR these days -- sorry for delay.

stefanv:
My long term view on data is that we should package it externally, but for now I'm fine with including small datasets.  But, with small, I mean a few kilobytes or hundreds of kilobytes.

soupault:
>This dataset is not free for commercial use, so I don't think we can include it.

Right! I've completely missed the legal aspect.
I'm reverting this till the further decision.

PR #2898, opened 2017-11-23
EXA example for haar like features

glemaitre:
## Description

required #2896 #2897 

## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.


glemaitre:
ping @stefanv @emmanuelle @warmspringwinds

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
```diff
@@ -0,0 +1,147 @@
...
+
+Haar-like feature descriptors were successfully used to implement the first
+real-time face detector [1]_. Inspired by this application, we propose an
+example illustrating the extraction, selection, and classification of Haar-like
+features to detect faces vs. non-faces.
+
+Notes
+-----
+
+This example relies on scikit-learn regarding the feature selection and
```
stefanv: on scikit-learn for feature selection

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
```diff
@@ -0,0 +1,147 @@
...
+
+def extract_feature_image(img, feature_type, feature_coord=None):
+    """Extract the haar feature for the current image"""
+    ii = integral_image(img)
+    return haar_like_feature(ii, 0, 0, ii.shape[0], ii.shape[1],
+                             feature_type=feature_type,
+                             feature_coord=feature_coord)
+
+
+# trick such that we can pickle this function when building the doc with
```
stefanv: Is this really necessary?  I asked about Dask before; does it have the same issue?

glemaitre: I will check

glemaitre:
> The PR looks good; we should see if we can find a free face dataset, though.

The LFW could be fine in this regard. We need to resize the images on the fly.
What I would miss is a dataset of non-face.

emmanuelle:
You need to add the data file.

glemaitre:
@emmanuelle I think that we intended this with #2905.
Once #2905, merged I have to try using dask instead of joblib as @stefanv mentioned.

glemaitre:
So I used `dask` instead of `joblib`. However, it seems that only `dask[array]` is installed.
Using `joblib` we can use the part which is vendor in`sklearn`. So I am not sure what the best.

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
```diff
@@ -130,10 +128,13 @@ def extract_feature_image(img, feature_type, feature_coord=None):
 # Note: we could select those features from the
 # original matrix X but we would like to emphasize the usage of `feature_coord`
 # and `feature_type` to recompute a subset of desired features.
-X = np.array(Parallel(n_jobs=-1)(
-    delayed(extract_feature_image)(img, selected_feature_type,
-                                   selected_feature_coord)
-    for img in images))
+
+# Delayed the computation and build the graph using dask
+X = delayed(extract_feature_image(img, selected_feature_type,
```
jni: @glemaitre I think here you want instead:

```python
Xd = [delayed(extract_feature_image)(img, selected_feature_type, selected_feature_coord)
      for img in images]
X = np.array(dask.compute(*Xd))
```

Then you won't need dask.bag.

By the way, do we really need multiprocessing here? Aren't most operations gil-released C/Cython?

glemaitre:
‎The gil is released but for the processing of one ROI. For an image you will have a sliding window defining the ROIs. So there you acquired and free the GIL between each processing of a ROI. So this is more efficient as multiprocessing by image. Otherwise we would need an helper which given an image would compute every feature and release the GIL for the full operation. I did not do that since it was not following the API of the already available feature (I think the multiblock LBP)

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
glemaitre: Basically I don't need dask.bag. I need cloudpickle because it is imported when importing multiprocessing from dask. To avoid installing dask[complete] since I only need cloudpickle, I installed dask[array,bag] which install dask, numpy, and cloudpickle

Regarding the syntax, I think that they are equivalent.

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
```diff
@@ -0,0 +1,148 @@
...
+
+Haar-like feature descriptors were successfully used to implement the first
+real-time face detector [1]_. Inspired by this application, we propose an
+example illustrating the extraction, selection, and classification of Haar-like
+features to detect faces vs. non-faces.
+
+Notes
+-----
+
+This example relies on scikit-learn for feature selection regarding the feature
```
stefanv: Double check this sentence, please.  Everything else looks good to me!

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
stefanv: This is fine; another option would be to add cloudpickle to the requirements explicitly with a comment.

glemaitre:
The error seems to come from master also.

glemaitre:
Actually this is due to the rc of mpl #2965

glemaitre:
So good to be merged then ;)

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
```diff
@@ -0,0 +1,147 @@
...
+from sklearn.model_selection import train_test_split
+from sklearn.metrics import classification_report
+
+from skimage.data import lfw_subset
+from skimage.transform import integral_image
+from skimage.feature import haar_like_feature
+from skimage.feature import haar_like_feature_coord
+from skimage.feature import draw_haar_like_feature
+
+print(__doc__)
```
jni: Is this line needed? I don't see it in the other examples...

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
```diff
@@ -0,0 +1,147 @@
...
+# We will use a subset of the CBCL which is composed of 100 face images and 100
+# non-face images. Each image has been resized to a ROI of 19 by 19 pixels. We
+# will keep 75 images from each group to train a classifier and check which
+# extracted features are the most salient.
+
+images = lfw_subset()
+# For a gain of speed, only the two first types of features will be extracted.
+feature_types = ['type-2-x', 'type-2-y']
+
+# Delayed the computation and build the graph using dask
```
jni: Delay the computation

jni:
Hi @glemaitre! I love this example!

I've added a couple of minor comments. Additionally, I'm left wondering about the overall idea behind the example. It seems to me that you are doing a complete classification, followed by evaluating which features are most important, followed by a classification based only on those important features. Presumably the goal of the last two steps is to get fast classification, so you are implicitly doing a speed/accuracy tradeoff, but you don't say this anywhere. It would be cool to see the effect that subsetting the features has on both the test speed (including both feature extraction and classification) and the test accuracy. What do you think? I don't think these concerns should hold up the PR, but let me know if you want to include them.

glemaitre:
> Presumably the goal of the last two steps is to get fast classification, so you are implicitly doing a speed/accuracy tradeoff, but you don't say this anywhere.

It is one of the reason but not only. The main reason is that a lot of the feature will be redundant or not discriminative so you want to remove them. However, it is true that it will speed-up the processing as a side effect.

> It would be cool to see the effect that subsetting the features has on both the test speed (including both feature extraction and classification) and the test accuracy. 

We could do that and show this trade-off (in practice, I would do a grid-search and only care about classification performance). However, I have 2 concerns:

* it will make the example longer and might make it more clutter.
* the execution time will also be longer.

@jni Could you tell me what you think regarding the two last points.

jni:
@glemaitre I don't want a grid search, only a bit of a discussion to say *why* you are subsetting, ie just like you said above, many features are redundant. And since the reference paper is called "real-time face detection", presumably time is a minor concern at least. =) So I would:

- say we are removing the features because there is a lot of redundancy and so we can dramatically reduce computation cost while maintaining accuracy
- time and check the test accuracy, so that you can show that your motivation for doing this indeed lines up with reality. =)

But yeah let's ignore the grid search aspect!

stefanv:
@jni Is this ready to go?

jni:
@stefanv @glemaitre @soupault I've made a few changes, please review. (@glemaitre I would have made a PR against your branch but bizarrely GitHub wouldn't let me choose your fork as the base repo. =\)

I am unsure about b78543a, because it increases the complexity, but the importance in performance is dramatic: 13s of feature computation vs 0.3s. So I thought it worthwhile. Since this is an advanced example I think it's ok. Here's the output of the print on my machine (manually wrapped here for clarity):

```
724 features, or 0.7%, account for 70% of branch points in the random forest.
Computing the full feature set took 81.950s, plus 2.383s training, for an AUC of 1.00.
Computing the restricted feature set took 0.322s, plus 3.356s training, for an AUC of 1.00.
```

Note that the training time is quite variable, and appears to be independent of the number of features, which is interesting. Random Forests aren't very performant at all.

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
```diff
@@ -0,0 +1,176 @@
...
+t_start = time()
+X = np.array(X.compute(get=multiprocessing.get))
+time_full_feature_comp = time() - t_start
+y = np.array([1] * 100 + [0] * 100)
+X_train, X_test, y_train, y_test = train_test_split(X, y, train_size=150,
+                                                    random_state=0,
+                                                    stratify=y)
+
+# Extract all possible features to be able to select the most salient.
+feature_coord, feature_type = \
```
glemaitre: are using the escape `\` usually?

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
```diff
@@ -0,0 +1,176 @@
...
+
+# Train a random forest classifier and check performance
+clf = RandomForestClassifier(n_estimators=1000, max_depth=None,
+                             max_features=100, n_jobs=-1, random_state=0)
+t_start = time()
+clf.fit(X_train, y_train)
+time_full_train = time() - t_start
+auc_full_features = roc_auc_score(y_test, clf.predict_proba(X_test)[:, 1])
+
+# Sort features in order of importance, plot six most significant
```
glemaitre: significant features

jni:
@glemaitre yes I do think it makes sense to put it into the docstring, though I'm not exactly clear on the reasons for *such* a difference, in either step. Anyway it's your PR, I just wanted to make a small contribution to it. =) So feel free to push extra changes on. =)

doc/examples/xx_applications/plot_haar_extraction_selection_classification.py
glemaitre: Make sense.

glemaitre:
@jni I added the note. I am fine with the rest.
Is there anything else which need to be addressed before merging?

jni:
Yes, it's ready as far as I can tell. I'll let @stefanv review + merge, though. Thank you!

stefanv:
Excellent example, and also the first to use sklearn! Thanks!

PR #2902, opened 2017-11-26
Speedup 2D warping for affine transformations

ahojnnes:
## Description
Following up from https://twitter.com/jeremyphoward/status/931761385628778496, this should make non-projective transformations (any affine transformation) around 2x faster. In the end, we will not be able to reach the same speed as some of the highly optimized libraries, but this is a first step towards improving performance.

### DIFF
--- a/skimage/transform/_warps_cy.pyx
+++ b/skimage/transform/_warps_cy.pyx
@@ -128,9 +128,7 @@ def _warp_fast(cnp.ndarray image, cnp.ndarray H, output_shape=None,
     cdef Py_ssize_t cols = img.shape[1]
 
     cdef void (*transform_func)(double, double, double*, double*, double*) nogil
-    if (abs(M[2, 0]) < 1e-10 and
-            abs(M[2, 1]) < 1e-10 and
-            abs(M[2, 2] - 1) < 1e-10):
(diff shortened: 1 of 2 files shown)


skimage/transform/_warps_cy.pyx
```diff
@@ -110,6 +127,14 @@ def _warp_fast(cnp.ndarray image, cnp.ndarray H, output_shape=None,
     cdef Py_ssize_t rows = img.shape[0]
     cdef Py_ssize_t cols = img.shape[1]
 
+    cdef void (*transform_func)(double, double, double*, double*, double*) nogil
+    if (abs(M[2, 0]) < 1e-10 and
```
stefanv: This seems like a very dangerous approach; e.g., in my work we often optimize over the parameters of a transformation---don't want to be hitting a wall at some arbitrary value.  We should check for absolute zero here.

ahojnnes:
@stefanv Agreed, moved this to the resize/rotate functions.

stefanv:
Floating point exception; probably related.

ahojnnes:
@stefanv @soupault This should be working now, also sped up the warping code a little more for the `resize` case.

hmaarrfk:

PR #2910, opened 2017-12-01
Add Mock RNG for Testing

kne42:
## Description
Adds a mock RNG class for better control over testing functions with random properties.

Ideally, all random functions would be reworked to take in a random number generator that must have a function that returns a value of the specified data type.

Example:
```python
def foo(random):
    """Does something.

    Parameters
    ----------
    random : object with function `rand` which returns an int
        Random number generator.

    Returns
    -------
    bar : int
        Your result.
    """
    assert_attrs(random, ['rand'])
    return random.rand()
```
Then it would behave as follows:
```python
>>> random = MockRNG('rand', [42, 24, 5])
>>> foo(random)
42
>>> foo(random)
24
>>> foo(random)
5
>>> foo(random)
42
```

## Checklist
- [x] Unit tests

## References
https://github.com/scikit-image/scikit-image/pull/2877#issuecomment-345417868
#2773 

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/testing.py
+++ b/skimage/_shared/testing.py
@@ -31,6 +31,30 @@ raises = pytest.raises
 fixture = pytest.fixture
 
 
+class MockRNG:
+    """A mock (controlled) random number generator.
+
+    Parameters
+    ----------
+    name : String
+        The name of the function that will output the mock values.
+    mock_values : list
+        The mock values to return in order. Will loop if the end is reached.
+    """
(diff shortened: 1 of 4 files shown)


kne42:
@soupault @stefanv @jni

emmanuelle:
Hi @kne42 could you please explain why this addition would be needed?

kne42:
Closing as unnecessary.

PR #2922, opened 2017-12-14
do not assume 3 channels during non-local means denoising

### LINKED ISSUE #2913: bug in multichannel non-local means if channels != 3
## Description
The non-local means Cython code currently assumes 3 channels when computing multi-channel patch weights, but there is no check that the input images are actually 3 channel.

code that assumes three channels: [here](https://github.com/scikit-image/scikit-image/blob/cf3cdaf35797b0ebbaf155c91ecf5f8b761e0ea5/skimage/restoration/_nl_means_denoising.pyx#L108) and [here](https://github.com/scikit-image/scikit-image/blob/cf3cdaf35797b0ebbaf155c91ecf5f8b761e0ea5/skimage/restoration/_nl_means_denoising.pyx#L467-L472)
Other code outside of the weights computation [already supports arbitrary number of channels](https://github.com/scikit-image/scikit-image/blob/cf3cdaf35797b0ebbaf155c91ecf5f8b761e0ea5/skimage/restoration/_nl_means_denoising.pyx#L637)

If the input images have more than 3 channels, denoising will still occur, but the patch weights will be determined based on only the first 3 channels.  In the off chance that a two-channel image is used, out-of-bounds memory access can occur during computation of the patch weights.

The Cython code should be updated to fully support an arbitrary number of channels.  One real world use case for > 3 channels:  [hyperspectral imaging](https://en.wikipedia.org/wiki/Hyperspectral_imaging).  I volunteer to make a PR to do this.

## Way to reproduce
see the example code linked that demonstrates where the problem lies

alternatively, modify the top of the non-local means Cython code to set `boundscheck=True` and then try denoising only the first two channels of an RGB image.  You will see a lot of messages regarding out-of-bounds access on the memory-views.

grlee77:
See #2913 for a more detailed description of the issue.

## Description
After this PR, non-local means will consider all channels when computing patch weights (rather than always assuming 3 channels).

## Checklist
- [x] Unit tests

## References
closes #2913

## For reviewers

  later.

### DIFF
--- a/skimage/restoration/_nl_means_denoising.pyx
+++ b/skimage/restoration/_nl_means_denoising.pyx
@@ -67,7 +67,8 @@ cdef inline double patch_distance_2d(IMGDTYPE [:, :] p1,
 
 cdef inline double patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
                                         IMGDTYPE [:, :, :] p2,
-                                        IMGDTYPE [:, ::] w, int s, double var):
+                                        IMGDTYPE [:, ::] w, int s, double var,
+                                        int n_ch):
     """
     Compute a Gaussian distance between two image patches.
 
@@ -95,9 +96,7 @@ cdef inline double patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
 
     .. math::  \exp( -w ((p1 - p2)^2 - 2*var))
     """
-    cdef int i, j
-    cdef int center = s / 2
-    cdef int color
+    cdef int i, j, channel
     cdef double tmp_diff = 0
     cdef double distance = 0
     for i in range(s):
@@ -105,8 +104,8 @@ cdef inline double patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
         if distance > DISTANCE_CUTOFF:
             return 0.
         for j in range(s):
-            for color in range(3):
-                tmp_diff = p1[i, j, color] - p2[i, j, color]
+            for channel in range(n_ch):
+                tmp_diff = p1[i, j, channel] - p2[i, j, channel]
                 distance += w[i, j] * (tmp_diff * tmp_diff - 2 * var)
     distance = max(distance, 0)
     distance = fast_exp(-distance)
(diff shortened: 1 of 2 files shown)


skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -65,9 +65,10 @@ cdef inline double patch_distance_2d(IMGDTYPE [:, :] p1,
     return distance
 
 
-cdef inline double patch_distance_2drgb(IMGDTYPE [:, :, :] p1,
-                                        IMGDTYPE [:, :, :] p2,
-                                        IMGDTYPE [:, ::] w, int s, double var):
+cdef inline double patch_distance_2dmultichannel(IMGDTYPE [:, :, :] p1,
+                                                 IMGDTYPE [:, :, :] p2,
+                                                 IMGDTYPE [:, ::] w,
+                                                 int s, double var, int n_ch):
```
soupault: Add `n_ch` to the docstring,

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -634,10 +638,10 @@ def _fast_nl_means_denoising_2d(image, int s=7, int d=13, double h=0.1,
     # Normalize pixel values using sum of weights of contributing patches
     for row in range(offset, n_row - offset):
         for col in range(offset, n_col - offset):
-            for channel in range(n_ch):
+            for ch in range(n_ch):
```
soupault: Why this change? :)

grlee77: I should have probably left it as channels as that is more explicit.  I have now updated to use `channels` and `n_channels` rather than `ch`, `n_ch` everywhere.

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -353,7 +362,7 @@ def _nl_means_denoising_3d(image, int s=7, int d=13, double h=0.1,
                                     padded[pln_start_i:pln_end_i,
                                            row_start_j:row_end_j,
                                            col_start_k:col_end_k],
-                                    w, s, var*var)
+                                    w, s, var)
```
grlee77: The change at this line is a bug fix to my previously merged PR #2890  (I was originally squaring the standard deviation here, but later switched to using the variance and missed removing the square in this case)

PR #2923, opened 2017-12-15
add missing cdef in _integral_image_3d (non-local means)

grlee77:
## Description
Adding this missing Cython cdef to the fast_mode=True mode in 3D non-local means gave me an 8-fold speedup on an example that i tried (for shape 64x64x64, patch_size=3, patch_distance=3).  

and trivial to review as well!

cc:   @blochl 

p.s. I encountered this when looking into potentially adding `nogil` to the non-local means Cython functions so that they can be accelerated with multi-threading (e.g. via block-wise processing with dask). 

## For reviewers

  later.


stefanv:
Yikes! How did we miss that.  A reminder to liberally use `cython -a`!

PR #2932, opened 2018-01-05
Remove copyright notice with permission of the author (Thomas Lewiner)

stefanv:
/cc @almarklein 

> actually there is no licensing restriction on that piece of code (fully free source).
It could be governed by the general scikit-image one.
> -- Thomas Lewiner, via email, 4 January 2018


JDWarner:
LGTM, thanks to Thomas Lewiner for giving the green light!

Issue #2783, opened 2017-08-29
randomwalker error


Phlya:
## Description
I am trying to use the randomwalker segmentation method, which seems to be generally superior to watershed. However on some images or with certain pre-processing I get a strange error. When I apply the same code to the same image with different gaussian smoothing (i.e. different sigma) it can either work, or not! The same applies to identically processed z-slices of the same 3D-image, for example.
Here is a simplified version of my code:
```
def apply_randomwalker2D(image):
    distance = ndimage.distance_transform_edt(image)
    distance = np.array(distance)
    local_maxi = peak_local_max(distance, min_distance=10, num_peaks=2,
                                indices=False, labels=image)
    markers = label(local_maxi)
    markers[image==0] = -1
    labels_rw = skimage.segmentation.random_walker(image, markers, mode='bf')
    labels_rw[labels_rw==-1]=0
    return labels_rw

def segment_z(subim):
    a = gaussian(subim, 3)
    mask = a>threshold_otsu(a)
    regions = apply_randomwalker2D(ndimage.binary_fill_holes(mask))
    regions[~mask]=0
    regions = skimage.segmentation.relabel_sequential(regions)[0]
    return regions
```

Here is the traceback:
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-388-aafa10950978> in <module>()
      1 plt.imshow(im[61], cmap='Greys_r', vmax=im.max())
----> 2 regions = segment_z(im[61])
      3 plt.imshow(np.ma.masked_equal(regions, 0), vmin=0, vmax=regions.max(), cmap='viridis', alpha=0.15)
      4 plt.show()

<ipython-input-387-c57927be1e22> in segment_z(subim)
     17 
     18     mask = a>threshold_otsu(a)
---> 19     regions = apply_randomwalker2D(ndimage.binary_fill_holes(mask))
     20     regions[~mask]=0
     21     regions = skimage.segmentation.relabel_sequential(regions)[0]

<ipython-input-387-c57927be1e22> in apply_randomwalker2D(image)
      6     markers = label(local_maxi)
      7     markers[image==0] = -1
----> 8     labels_rw = skimage.segmentation.random_walker(image, markers, mode='bf')
      9     labels_rw[labels_rw==-1]=0
     10     return labels_rw

[...]

PR #2946, opened 2018-01-21
Correct bug in random walker when seed pixels are isolated inside pruned zones

emmanuelle:
Solves #2783.

### DIFF
--- a/skimage/segmentation/random_walker_segmentation.py
+++ b/skimage/segmentation/random_walker_segmentation.py
@@ -188,6 +188,20 @@ def _build_laplacian(data, spacing, mask=None, beta=50,
     return lap
 
 
+def _check_isolated_seeds(labels):
+    """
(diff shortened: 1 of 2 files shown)


emmanuelle:
@scikit-image/core I rebased on master for Travis' sake, just waiting for a review now :-)

emmanuelle:
thanks @stefanv ! @jni would you have the time to read it as well?

see--:
Ok. Thank you for the fix. Is there a similar way to resolve isolated unlabeled pixels (0-value)? (e.g. `_check_isolated_unlabeled`). I get an `ValueError` here:
```
import numpy as np
from skimage.segmentation import random_walker

img = -np.ones((10, 10))
img[5:6, 3:6] = 0
seg = random_walker(img, img)
```
--> `ValueError: zero-size array to reduction operation maximum which has no identity`

PR #2949, opened 2018-01-25
`plot_matches` function in `feature` can now show the images vertically

hmaarrfk:
## Description
Initially, you could only show features matched side by side:
image

Now you can show features matched vertically too:
image

## Checklist
- [X] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/util.py
+++ b/skimage/feature/util.py
@@ -41,7 +41,8 @@ class DescriptorExtractor(object):
 
 
 def plot_matches(ax, image1, image2, keypoints1, keypoints2, matches,
-                 keypoints_color='k', matches_color=None, only_matches=False):
+                 keypoints_color='k', matches_color=None, only_matches=False,
+                 left_right=True):
     """Plot matched features.
 
     Parameters
@@ -67,6 +68,9 @@ def plot_matches(ax, image1, image2, keypoints1, keypoints2, matches,
         color is chosen randomly.
     only_matches : bool, optional
         Whether to only plot matches and not plot the keypoint locations.
+    left_right : bool, optional
+        Whether to show images side by side (left and right) or one above
+        the other.
 
     """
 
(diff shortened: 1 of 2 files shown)


skimage/feature/util.py
```diff
@@ -41,7 +41,8 @@ def extract(self, image, keypoints):
 
 
 def plot_matches(ax, image1, image2, keypoints1, keypoints2, matches,
```
soupault: I feel that this is not the right approach. Do we have any alternatives?
I think this should be done by passing several axes via `ax` parameter or something, but, in that case, it won't be possible to draw the lines.

@ahojnnes would you like to comment?

jni: Since multiple axes are not possible, I think a kwarg is correct, but `left_right` doesn't work for me... I would suggest `alignment={'horizontal','vertical'}`

skimage/feature/util.py
```diff
@@ -96,18 +100,22 @@ def plot_matches(ax, image1, image2, keypoints1, keypoints2, matches,
         new_image2[:image2.shape[0], :image2.shape[1]] = image2
         image2 = new_image2
 
-    image = np.concatenate([image1, image2], axis=1)
-
-    offset = image1.shape
+    offset = np.array(image1.shape)
+    if left_right is True:
```
jni: As a side comment, the Pythonic way to use bool arguments is, for example, `if left_right:`. In which case a different renaming is suggested: `if horizontal_alignment:`. But as I mentioned before, I would prefer:

```python
if alignment == 'horizontal':
    ...
elif alignment == 'vertical':
    ...
else:
    raise ValueError('incorrect value for alignment: %s' % alignment)
```

hmaarrfk: Thanks for the tip.

I agree that `is True` is verbose.

Your suggestion keeps the best of both :D

PR #2964, opened 2018-02-14
Added functions to convert between data types.

hmaarrfk:
## Description
I'm trying to use skimage for my image analysis. Coming from Matlab, I found a few functions lacking.
I decided to take a hack at the easy ones listed on:
http://scikit-image.org/docs/0.11.x/coverage_table.html

It seems that coverage_table has been dropped in recent versions?

## Checklist
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/data_convert.py
+++ b/skimage/color/data_convert.py
@@ -47,7 +47,7 @@ def im2integer(image, dtype):
     else:
         raise TypeError("dtype must be a numpy integer type.")
 
-    if issubclass(image.dtype, np.floating):
+    if issubclass(image.dtype.type, np.floating):
         image_float = np.ldexp(image, bits)
         return image_float.astype(dtype)
     else:
@@ -61,7 +61,7 @@ def im2float(image, dtype):
 
     image_float = image.astype(dtype)
 
-    if issubclass(image.dtype, np.integer):
+    if issubclass(image.dtype.type, np.integer):
         max_val = np.iinfo(image.dtype.type).max
         image_float = image_float / max_val
 
--- a/skimage/color/data_convert.py
+++ b/skimage/color/data_convert.py
@@ -47,7 +47,7 @@ def im2integer(image, dtype):
     else:
         raise TypeError("dtype must be a numpy integer type.")
 
-    if issubclass(image.dtype.type, np.floating):
+    if issubclass(image.dtype, np.floating):
         image_float = np.ldexp(image, bits)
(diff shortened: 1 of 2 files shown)


hmaarrfk:
Sorry for submitting a messy (and plain wrong) PR at first. I wasn't able to get testing working on my personal machine (windows makes Python Dev complicated).

Ultimately, I was able to get my Linux install working and that helped alot.

I definitely learned alot of quirks about converting between `uint8` and `uint16`. I should really document the function's peculiarity a little more.

jni:
Hi @hmaarrfk 

Thank you for the PR! Normally this would count as a fantastic PR, given the full test coverage with zero prompting from us. However, unfortunately you have duplicated existing functionality:

http://scikit-image.org/docs/dev/user_guide/data_types.html

It looks like we need to work on the discoverability of that page, because many, many issues that people raise come from not finding it. If you have any ideas for improving it, we would be very receptive!

The table was out of date in 0.11 from what I can tell, so presumably someone made the call that removing it was less bothersome than maintaining it. I also think that there was a desire to follow our own roadmap instead of "chasing" Matlab. However, I actually think having something like that would be extremely useful, but quite a lot of work to create in the first place, and then to maintain. Still, if you're feeling brave, it would be good to compile.

Currently, much of the functionality is spread across NumPy, SciPy.ndimage, scikit-image, and Matplotlib. For other stuff (such as reading DICOM), you would need other libraries such as NiBabel. However, they certainly exist! So I would probably make a 4-column table, "Matlab", "Core SciPy (includes NumPy, SciPy, Matplotlib", "scikit-image", "third party suggestions". I would also store it as a csv and somehow parse that into HTML, as CSV would be way easier to maintain.

Anyway, sorry for the duplicated work here. I hope that there's other bits of the table that you would be interested in contributing! I made the Matlab -> Python jump myself once upon a time and it's totally worth it. ;) When in doubt, it's a good idea to ping us at scikit-image@python.org or gitter.im/scikit-image/scikit-image, where we can point you in the right direction or confirm that something isn't yet implemented.

Thanks! I'm going to unceremoniously close this for now. If there's functionality you need that we don't yet provide, I would suggest extending our `img_as_*` functions. Or if I'm way off the mark, just ping and we'll reopen this discussion!

hmaarrfk:
Funny that I didn't find that. It gave me a good chance to learn about the peculiarities of converting between datatypes.

All good :D

PR #2966, opened 2018-02-16
Support contours in future.manual_segmentation

aaronsnoswell:
## Description
This PR updates the functions [`future.manual_segmentation.manual_polygon_segmentation`](http://scikit-image.org/docs/dev/api/skimage.future.html#manual-polygon-segmentation) and [`future.manual_segmentation.manual_lasso_segmentation`](http://scikit-image.org/docs/dev/api/skimage.future.html#manual-lasso-segmentation). The changes add a second return parameter that is the actual list of contour points. This is sometimes more useful to an end user than the mask only.

NB: This will break previous APIs that depended on these functions. Old code might have looked like this (from the docstrings);

```
camera = data.camera()
mask = manual_segmentation.manual_lasso_segmentation(camera)
io.imshow(mask)  
io.show()
```

Now, users will have to add a temporary variable to catch the contour, even if they are not interested in it; viz.

```
mask, _ = manual_segmentation.manual_lasso_segmentation(camera)
```

Docstrings are updated and the changes were fairly minimal - I haven't read all of PEP8, but I believe it should still be compatible (unless multiple return statements are a no-no).

## Checklist
- [ ] Unit tests

##
Release notes message:
```
 ``future.manual_segmentation.manual_polygon_segmentation`` and
  ``future.manual_segmentation.manual_lasso_segmentation`` were updated to also
  return the contour points picked, not just a mask image.
```

### DIFF
--- a/skimage/future/manual_segmentation.py
+++ b/skimage/future/manual_segmentation.py
@@ -52,10 +52,9 @@ def manual_polygon_segmentation(image, alpha=0.4, return_all=False):
     labels : array of int, shape ([Q, ]M, N)
         The segmented regions. If mode is `'separate'`, the leading dimension
         of the array corresponds to the number of regions that the user drew.
-    contours : array of float, shape ([Q, ]O)
-        The contours of the segmented regions. If mode is `'separate'`, the
-        leading dimension of the array corresponds to the number of regions
-        that the user drew.
+    contours : array of float, shape (Q, 2)
+        The contours of the segmented regions. The leading dimension of the
+        array corresponds to the number of regions that the user drew.
 
     Notes
     -----
@@ -173,10 +172,9 @@ def manual_lasso_segmentation(image, alpha=0.4, return_all=False):
     labels : array of int, shape ([Q, ]M, N)
         The segmented regions. If mode is `'separate'`, the leading dimension
         of the array corresponds to the number of regions that the user drew.
-    contours : array of float, shape ([Q, ]O)
-        The contours of the segmented regions. If mode is `'separate'`, the
-        leading dimension of the array corresponds to the number of regions
-        that the user drew.
+    contours : array of float, shape (Q, 2)
+        The contours of the segmented regions. The leading dimension of the
+        array corresponds to the number of regions that the user drew.
 
     Notes
     -----
--- a/skimage/future/manual_segmentation.py
+++ b/skimage/future/manual_segmentation.py
@@ -32,7 +32,8 @@ def _draw_polygon(ax, vertices, alpha=0.4):
 
 @require("matplotlib", ">=3.3")
 def manual_polygon_segmentation(image, alpha=0.4, return_all=False):
-    """Return a label image and list of contours based on polygon selections made with the mouse.
+    """Return a label image and list of contours based on polygon selections
+    made with the mouse.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


skimage/future/manual_segmentation.py
```diff
@@ -49,6 +50,9 @@ def manual_polygon_segmentation(image, alpha=0.4, return_all=False):
     labels : array of int, shape ([Q, ]M, N)
         The segmented regions. If mode is `'separate'`, the leading dimension
         of the array corresponds to the number of regions that the user drew.
+    contours : array of float, shape (Q, 2)
```
soupault: The shape of `contours` is clearly not correct here. Has to be a list of lists or something similar.

skimage/future/manual_segmentation.py
```diff
@@ -134,13 +139,19 @@ def _extend_polygon(event):
...
+            np.maximum,
+            labels,
+            np.broadcast_to(0, image.shape[:2])
+        ),
+        list_of_vertex_lists
 
 
 def manual_lasso_segmentation(image, alpha=0.4, return_all=False):
-    """Return a label image based on freeform selections made with the mouse.
+    """Return a label image  and list of contours based on freeform selections
```
soupault: Keep it a one-liner. If more detailed description is required, put it 2 lines below.
Also, fix a double whitespace.

skimage/future/manual_segmentation.py
```diff
@@ -160,6 +171,9 @@ def manual_lasso_segmentation(image, alpha=0.4, return_all=False):
     labels : array of int, shape ([Q, ]M, N)
         The segmented regions. If mode is `'separate'`, the leading dimension
         of the array corresponds to the number of regions that the user drew.
+    contours : array of float, shape (Q, 2)
```
soupault: Same for the shape here.

skimage/future/manual_segmentation.py
```diff
@@ -212,6 +227,11 @@ def _on_lasso_selection(vertices):
     labels = (_mask_from_vertices(vertices, image.shape[:2], i)
               for i, vertices in enumerate(list_of_vertex_lists, start=1))
     if return_all:
```
soupault: Update `return_all` parameter description for both functions to include also info on contours.

soupault (changes requested):
Quite useful enhancement! Couple of remarks from me, otherwise ok

soupault:
While returning polygon vertices (PVs) / contour points (CPs) would be a nice addition to the manual segmentation functions, it comes here with a challenge. With `return_all` set to `False`, the polygons/masks are post-processed in a way that the polygons/masks drawn later zero out the areas of the ones drawn earlier when they overlap. In such cases, the returned PVs/CPs are not correct, since they are recorded from the regions before post-processing. The correct implementation would first post-process the polygons/masks and then recompute the PVs/CPs. For polygon vertices, this becomes quite complicated already, and I am not sure whether we want to go further in the direction of implementing advanced parts of image editing tools in `skimage`.

I would advocate for keeping the manual segmentaiton tools as simple as possible (or, perhaps, even removing them at some point) and encouraging users to learn and utilize the professional tools for image annotation (such as Napari, ITK-SNAP, Slicer3D, etc).

Since there has been no activity on this PR for a while and the implementation is somewhat early-stage, I propose to close the PR until further notice.

matthew-brett:

<details>

## Rationale

</details>

PR #2967, opened 2018-02-16
Watershed segmentation: make usable for large arrays

blochl:
The numbering of indexes is now 64-bit, to allow usage on arrays with more than 2<sup>31</sup>-1 elements. With volumes as large as (for example) 2048<sup>3</sup>, which is not an extraordinary size for 3D volumes anymore, 32-bit index numbering can not be used.

Additionally, some related quantities are kept 64-bit (and not converted to 32-bit, as before) to avoid redundant operations. These quantities are not parts of large arrays, so there is no noticeable effect on memory usage.

Also, some trivial refactoring is made: empty lines are inserted for readability and PEP8 compliance.

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/_watershed.pyx
+++ b/skimage/morphology/_watershed.pyx
@@ -17,6 +17,7 @@ cimport cython
 
 
 ctypedef cnp.int32_t DTYPE_INT32_t
+ctypedef cnp.int64_t DTYPE_INT64_t
 ctypedef cnp.int8_t DTYPE_BOOL_t
 
 
@@ -28,8 +29,8 @@ include "heap_watershed.pxi"
 @cython.cdivision(True)
 @cython.overflowcheck(False)
 @cython.unraisable_tracebacks(False)
-cdef inline double _euclid_dist(cnp.int32_t pt0, cnp.int32_t pt1,
-                                cnp.int32_t[::1] strides):
+cdef inline double _euclid_dist(cnp.int64_t pt0, cnp.int64_t pt1,
+                                cnp.int64_t[::1] strides):
     """Return the Euclidean distance between raveled points pt0 and pt1."""
     cdef double result = 0
     cdef double curr = 0
@@ -46,7 +47,7 @@ cdef inline double _euclid_dist(cnp.int32_t pt0, cnp.int32_t pt1,
 @cython.cdivision(True)
 @cython.unraisable_tracebacks(False)
 cdef inline DTYPE_BOOL_t _diff_neighbors(DTYPE_INT32_t[::1] output,
-                                         DTYPE_INT32_t[::1] structure,
+                                         DTYPE_INT64_t[::1] structure,
                                          DTYPE_BOOL_t[::1] mask,
                                          Py_ssize_t index):
     """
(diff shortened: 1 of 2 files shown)


blochl:
Ping?

(The Travis CI errors are unrelated - because of an incompatibility with the "future" Matplotlib)

jni:
@blochl sorry, I didn't read this carefully enough to realise that you weren't changing the type of the output. That was my main concern, as it would double the memory requirement for most use cases for the benefit of a smaller number of use cases. As it is I think it's a great PR. =) Thanks! We'll need one more maintainer to sign off on it.

emmanuelle:
My only concern is that the memory footprint of the watershed will increase, which is a problem for large images. Wouldn't it be possible to test if 64-bit is needed or not?

blochl:
@emmanuelle , What will take space is the `marker_locations` array, but if the volume size is more than 1290\*\*3 (and it's not extremely big nowadays), there *may* be greater marker locations than `int32` can handle.
`marker_locations` is only as big as how many markers there are. So let's take the very extreme theoretical scenario: there are 2**31-1 markers continuously, and none afterwards. In this case, using `int64` array will cause an unneeded increase of only 64 GB of RAM. And that's the maximal wasted RAM possible. Most likely it will never go near that. Just one marker after that - and it's already not a waste, cause 64-bit is needed.
One can test if 64-bit is needed, and make separate functions for 32-bit and 64-bit watershed, but it's a lot of added complexity to save a little RAM. Also, if there's not enough RAM for the 64-bit version, even the 32-bit version will not work, because `np.flatnonzero` returns an `int64` array anyway. My change just avoids its conversion to `int32`.

emmanuelle:
Thanks for your detailed answer @blochl. I don't agree that 64 GB is "a little RAM", but since the RAM is needed anyway because of the output of ``np.flatnonzero`` (although it could be freed after the conversion to int32), I'm OK with the change.

emmanuelle:
@scikit-image/core how about these failing tests from ``test_mpl_imshow``? Are they related to the PR or not?

blochl:
Well, 64 GB is the absolute theoretical maximum. :)
Actually with the conversion it required more RAM (albeit temporarily) because the conversion to 32-bit was not in-place.

The failing tests, as I have mentioned before, are a result of incompatibility with the future version of Matplotlib, which deprecated the `'box-forced'` option in this commit: matplotlib/matplotlib@4d6448ab5b392a943b7f3f704dfe5b18c0764aa4 .

stefanv:
Is there an easy way to test for regression, without eating up a huge amount of RAM during testing?

blochl:
@stefanv : I don't think so.

jni:
Regarding RAM concerns, I don't think they are major here. We are already using a double-precision float array as the image. In the vast majority of watershed use cases, the number of markers will be much, much smaller than the number of pixels. In general, a factor of 100 is a reasonable assumption here. Therefore, we are doubling the size of an array (`marker_locations`) about 100x smaller than the image array to begin with. So, without even taking into account @blochl's point that we are actually removing a copy (thus actually reducing peak RAM usage), we only expect a 1% increase in RAM use.

This is before taking into account the fact that our heap queue uses `Py_ssize_t` as the index type, and this is 64-bit in 64-bit platforms, so we were already using 64 bits in the larger part of the index memory footprint anyway! (We expect much more than one heap item per watershed marker.)

stefanv:
Thanks for the explanation, Juan, that's convincing.

PR #2979, opened 2018-03-07
Bugfix LineModelND.residuals does not use the optional parameter `params`

benbo:
## Description
Bugfix. Formerly, the argument `params` in `LineModelND.rediduals(self, data, params=None) ` was always overwritten by self.params. Now, assignment of `origin, direction` was changed from `self.params` to `params`. Formerly, `params` argument would have been ignored no matter if it was `None` or not. Now, if `params` is `None` it is set to `self.params` as defined in line 120 and 121 of fit.py. Otherwise, if params is not None, the `params` argument will be used to set origin, direction.

## For reviewers

  later.

### DIFF
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -123,7 +123,7 @@ class LineModelND(BaseModel):
         if len(params) != 2:
             raise ValueError('Parameters are defined by 2 sets.')
 
-        origin, direction = self.params
+        origin, direction = params
         res = (data - origin) - \
               np.dot(data - origin, direction)[..., np.newaxis] * direction
         return _norm_along_axis(res, axis=1)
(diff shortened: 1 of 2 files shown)


sciunto:
Thanks for the fix. This looks correct to me. Waiting for a second review.

sciunto:
Is it possible to add a testcase corresponding to this issue?

jni:
Holy moly! Here's a great example of test coverage not telling the whole story.

@benbo thank you for picking this up, but I second @sciunto: do you think you could add a test case that checks that this works? Let us know if not, though, and we'll pick it up.

benbo:
Thanks @jni and @sciunto for the quick response. I haven't interacted with this community previously. Just to double check, you want some test cases to be added to ensure that the params argument that is passed is of the right format/type?

benbo:
@jni @sciunto added a test case that fails without this fix.

jni:
Brilliant, thanks @benbo! For the record we normally would put it in a separate test function, say, `test_line_model_nd_residuals_params_arg`, so that it would be easy to tell when it failed what failed. However this is a minor detail and I won't let it hold up this PR. Thank you for your contribution!

benbo:
Thanks @jni, still learning the ropes :)

PR #2983, opened 2018-03-13
Use skimage gaussian filter to avoid integer rounding artifacts

stefanv:
scipy.ndimage.gaussian_filter tries to preserve dtypes, which leads to
the output of operations on integer input being less accurate than it
should be.


jni:
@scikit-image/core can anyone else review this?

emmanuelle:
I'm fine with the change and ready to merge; except that Travis is failing on something unrelated with the PR (opening an image over http). I'll first try to restart Travis.

jni:
Hmmm the 3.4 Travis failure *might* be related, since it's a canny error, albeit in the viewer. @stefanv could you investigate before we merge?

stefanv:
@jni well spotted

stefanv:
Fixed, but I had to adjust the test; I *presume* this is because the canny edge detector is now calculated on a more accurate floating point image, but I have no explicitly checked by hand.

jni:
Looks good. @emmanuelle? =)

Travis failures appear unrelated. (One is the new warning from SciPy 1.1)

stefanv:
Added documentation to notify users of change.

PR #2992, opened 2018-03-22
RANSAC: add parameter for initial inliers

Borda:
## Description
add a relative number of inliers and initial inliers...

## Checklist
- [x] Doc tests

## For reviewers
  later.

### DIFF
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -621,7 +621,7 @@ def _dynamic_max_trials(n_inliers, n_samples, min_samples, probability):
 def ransac(data, model_class, min_samples, residual_threshold,
            is_data_valid=None, is_model_valid=None,
            max_trials=100, stop_sample_num=np.inf, stop_residuals_sum=0,
-           stop_probability=1, random_state=None, init_inliers=None):
+           stop_probability=1, random_state=None, initial_inliers=None):
     """Fit a model to data with the RANSAC (random sample consensus) algorithm.
 
     RANSAC is an iterative algorithm for the robust estimation of parameters
@@ -666,7 +666,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
 
         where `success` indicates whether the model estimation succeeded
         (`True` or `None` for success, `False` for failure).
-    min_samples : int in range (0, <nb-samples>)
+    min_samples : int in range (0, N)
         The minimum number of data points to fit a model to.
     residual_threshold : float larger than 0
         Maximum distance for a data point to be classified as an inlier.
@@ -699,7 +699,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
         If RandomState instance, random_state is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`.
-    init_inliers : array-like of bool, shape (N,), optional
+    initial_inliers : array-like of bool, shape (N,), optional
         Initial samples selection for model estimation
 
 
@@ -772,7 +772,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
     >>> ratio = 0.5  # use half of the samples
     >>> min_samples = int(ratio * len(src))
     >>> model, inliers = ransac((src, dst), SimilarityTransform, min_samples, 10,
-    ...                         init_inliers=np.ones(len(src), dtype=bool))
+    ...                         initial_inliers=np.ones(len(src), dtype=bool))
     >>> inliers
     array([False, False, False,  True,  True,  True,  True,  True,  True,
             True,  True,  True,  True,  True,  True,  True,  True,  True,
@@ -796,7 +796,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
     num_samples = len(data[0])
 
     if not (0 < min_samples < num_samples):
-        raise ValueError("`min_samples` must be in range (0, <nb-samples>)")
+        raise ValueError("`min_samples` must be in range (0, <number-of-samples>)")
 
     if residual_threshold < 0:
         raise ValueError("`residual_threshold` must be greater than zero")
@@ -807,16 +807,16 @@ def ransac(data, model_class, min_samples, residual_threshold,
     if not (0 <= stop_probability <= 1):
         raise ValueError("`stop_probability` must be in range [0, 1]")
 
-    if init_inliers is not None and len(init_inliers) != num_samples:
+    if initial_inliers is not None and len(initial_inliers) != num_samples:
         raise ValueError("RANSAC received a vector of initial inliers (length %i)"
                          " that didn't match the number of samples (%i)."
                          " The vector of initial inliers should have the same length"
                          " as the number of samples and contain only True (this sample"
                          " is an initial inlier) and False (this one isn't) values."
-                         % (len(init_inliers), num_samples))
+                         % (len(initial_inliers), num_samples))
 
     # for the first run use initial guess of inliers
-    spl_idxs = (init_inliers if init_inliers is not None
+    spl_idxs = (initial_inliers if initial_inliers is not None
                 else random_state.choice(num_samples, min_samples, replace=False))
 
     for num_trials in range(max_trials):
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -787,10 +787,6 @@ def ransac(data, model_class, min_samples, residual_threshold,
     best_inliers = None
 
     random_state = check_random_state(random_state)
-    if isinstance(min_samples, float):
-        if not (0 < min_samples <= 1):
-            raise ValueError("`min_samples` as ratio must be in range [0, 1)")
-        min_samples = int(min_samples * len(data[0]))
     if min_samples < 0:
         raise ValueError("`min_samples` must be greater than zero")
 
@@ -810,6 +806,11 @@ def ransac(data, model_class, min_samples, residual_threshold,
     data = list(data)
     # number of samples
     num_samples = data[0].shape[0]
+
+    if isinstance(min_samples, float):
+        if not (0 < min_samples <= 1):
+            raise ValueError("`min_samples` as ratio must be in range [0, 1)")
+        min_samples = int(min_samples * num_samples)
 
     if init_inliers is not None and len(init_inliers) != num_samples:
         raise ValueError("RANSAC received a vector of initial inliers (length %i) "
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -668,7 +668,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
         (`True` or `None` for success, `False` for failure).
     min_samples : int or float in range [0, 1]
         The minimum number of data points to fit a model to.
-    residual_threshold : float
+    residual_threshold : float larger then 0
         Maximum distance for a data point to be classified as an inlier.
     is_data_valid : function, optional
         This function is called with the randomly selected data before the
@@ -789,15 +789,18 @@ def ransac(data, model_class, min_samples, residual_threshold,
     random_state = check_random_state(random_state)
     if isinstance(min_samples, float):
         if not (0 < min_samples <= 1):
-            raise ValueError("`min_samples` as ratio must be in range (0, 1)")
+            raise ValueError("`min_samples` as ratio must be in range [0, 1)")
         min_samples = int(min_samples * len(data[0]))
     if min_samples < 0:
         raise ValueError("`min_samples` must be greater than zero")
 
+    if residual_threshold < 0:
+        raise ValueError("`residual_threshold` must be greater than zero")
+
     if max_trials < 0:
         raise ValueError("`max_trials` must be greater than zero")
 
-    if stop_probability < 0 or stop_probability > 1:
+    if not (0 <= stop_probability <= 1):
         raise ValueError("`stop_probability` must be in range (0, 1)")
 
     if not isinstance(data, list) and not isinstance(data, tuple):
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -778,7 +778,6 @@ def ransac(data, model_class, min_samples, residual_threshold,
             True,  True,  True,  True,  True,  True,  True,  True,  True,
             True,  True,  True,  True,  True,  True,  True,  True,  True,
             True,  True,  True,  True,  True], dtype=bool)
-
     """
 
     best_model = None
@@ -790,7 +789,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
     if isinstance(min_samples, float):
         if not (0 < min_samples <= 1):
             raise ValueError("`min_samples` as ration must be in range (0, 1)")
-        min_samples = int(min_samples * len(data))
+        min_samples = int(min_samples * len(data[0]))
     if min_samples < 0:
         raise ValueError("`min_samples` must be greater than zero")
 
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -778,6 +778,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
             True,  True,  True,  True,  True,  True,  True,  True,  True,
             True,  True,  True,  True,  True,  True,  True,  True,  True,
             True,  True,  True,  True,  True], dtype=bool)
+
     """
 
     best_model = None
@@ -788,7 +789,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
     random_state = check_random_state(random_state)
     if isinstance(min_samples, float):
         if not (0 < min_samples <= 1):
-            raise ValueError("`min_samples` as ration must be in range (0, 1)")
+            raise ValueError("`min_samples` as ratio must be in range (0, 1)")
         min_samples = int(min_samples * len(data[0]))
     if min_samples < 0:
         raise ValueError("`min_samples` must be greater than zero")
@@ -808,22 +809,21 @@ def ransac(data, model_class, min_samples, residual_threshold,
     num_samples = data[0].shape[0]
 
     if init_inliers is not None and len(init_inliers) != num_samples:
-        raise ValueError("`number init samples %i are les then data %i "
+        raise ValueError("RANSAC received a vector of initial inliers (length %i) "
+                         "that didn't match the number of samples (%i). "
+                         "The vector of initial inliers should have the same length "
+                         "as the number of samples and contain only True (this sample "
+                         "is an initial inlier) and False (this one isn't) values."
                          % (len(init_inliers), num_samples))
 
-    for num_trials in range(max_trials):
+    # for the first run use initial guess of inliers
+    idxs = (init_inliers if init_inliers is not None
+            else random_state.randint(0, num_samples, min_samples))
 
-        # choose random sample set
-        samples = []
-        # for the first run use intial guess of inliers
-        if num_trials == 0 and init_inliers is not None:
-            for d in data:
-                samples.append(d[init_inliers])
-        else:  # for any further run...
-            random_idxs = random_state.choice(num_samples, min_samples,
-                                              replace=False)
-            for d in data:
-                samples.append(d[random_idxs])
+    for num_trials in range(max_trials):
+        samples = [d[idxs] for d in data]
+        # for next time choose random sample set
+        idxs = random_state.choice(num_samples, min_samples, replace=False)
 
         # check if random sample set is valid
         if is_data_valid is not None and not is_data_valid(*samples):
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -666,7 +666,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
 
         where `success` indicates whether the model estimation succeeded
         (`True` or `None` for success, `False` for failure).
-    min_samples : int or float in range [0, 1]
+    min_samples : int or float in range [0, 1)
         The minimum number of data points to fit a model to.
     residual_threshold : float larger than 0
         Maximum distance for a data point to be classified as an inlier.
@@ -779,6 +779,16 @@ def ransac(data, model_class, min_samples, residual_threshold,
             True,  True,  True,  True,  True,  True,  True,  True,  True,
             True,  True,  True,  True,  True], dtype=bool)
 
+    Les us present an example of transforming this function requesting mimimal number
+     of samples to passing a ration of minimal samples called `min_sample_ratio` defined
+     in range [0, 1) as portion of total dataset::
+
+        from skimage.measure import ransac as _ransac
+
+        def ransac(data, model_class, min_sample_ratio, residual_threshold, *args, **kwargs):
+            min_samples = int(min_sample_ratio * len(data))
+            return _ransac(data, model_class, min_samples, residual_threshold, *args, **kwargs)
+
     """
 
     best_model = None
@@ -787,8 +797,15 @@ def ransac(data, model_class, min_samples, residual_threshold,
     best_inliers = None
 
     random_state = check_random_state(random_state)
-    if min_samples < 0:
-        raise ValueError("`min_samples` must be greater than zero")
+
+    if np.asarray(data).ndim != 2:
+        raise ValueError("`data` must be (N, D) array, where N is the number of data points"
+                         " and D the dimensionality of the data")
+    # number of samples, since data array here is  (N, D)
+    num_samples = len(data)
+
+    if not (0 < min_samples <= num_samples):
+        raise ValueError("`min_samples` must be in range (0, <number-of-samples>]")
 
     if residual_threshold < 0:
         raise ValueError("`residual_threshold` must be greater than zero")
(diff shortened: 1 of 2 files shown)


skimage/measure/fit.py
```diff
@@ -697,6 +697,8 @@ def ransac(data, model_class, min_samples, residual_threshold,
         If RandomState instance, random_state is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`.
+    init_inliers : [list, tuple of] (N) array of bool or None, optional
++        Initial samples selection for model estimation
```
jni: Errant + running around here...

skimage/measure/fit.py
```diff
@@ -697,6 +697,8 @@ def ransac(data, model_class, min_samples, residual_threshold,
         If RandomState instance, random_state is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`.
+    init_inliers : [list, tuple of] (N) array of bool or None, optional
```
jni: Use "array-like of bool"

jni: Oh I just saw what you meant here. Ignore me. =P

jni:
@Borda I'm not sure why you ticked "unit tests", as there's a few new untested lines in this code.

Additionally, I'm worried about code bloat here. Can you make a case for why these modifications are needed? Do you have an example use case?

jni:
Oops, I just saw your older, now-closed PR. Hmm. I guess this explains the stray `+`! Good luck recovering your work! =\

Borda:
@jni thanks, I think that now it is mostly recovered...

skimage/measure/fit.py
```diff
@@ -782,15 +785,18 @@ def ransac(data, model_class, min_samples, residual_threshold,
     best_inliers = None
 
     random_state = check_random_state(random_state)
-
+    if isinstance(min_samples, float):
```
jni: I'm not a big fan of this change... It could be a breaking change for some users. From my reading of the code, there was nothing stopping someone from using 17.5 as `min_samples` before? And it could be convenient for some users. If we want to use some magic here, I would suggest that `min_samples` in the range (0, 1) is considered a fraction, and min_samples greater than or equal to 1 is considered an absolute value. Does that make sense?

skimage/measure/fit.py
```diff
@@ -782,15 +785,18 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
+        if not (0 < min_samples <= 1):
+            raise ValueError("`min_samples` as ration must be in range (0, 1)")
```
jni: ration -> ratio or fraction

skimage/measure/fit.py
```diff
@@ -782,15 +785,18 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
+        min_samples = int(min_samples * len(data))
     if min_samples < 0:
         raise ValueError("`min_samples` must be greater than zero")
 
     if max_trials < 0:
         raise ValueError("`max_trials` must be greater than zero")
 
     if stop_probability < 0 or stop_probability > 1:
```
jni: I would change this to `if not 0 < stop_probability < 1:` to match the error text. (Currently, `stop_probability=0` will not be caught here!)

skimage/measure/fit.py
```diff
@@ -800,13 +806,22 @@ def ransac(data, model_class, min_samples, residual_threshold,
     # number of samples
     num_samples = data[0].shape[0]
 
+    if init_inliers is not None and len(init_inliers) != num_samples:
+        raise ValueError("`number init samples %i are les then data %i "
```
jni: Typo (les -> less). Also, conceivably, the number could be more, not less!

As we are discussing data types in the mailing list, I'm becoming more aware that error messages should be more specific about the cause of the error *and* how to fix it. An example of what this message could be:

"RANSAC received a vector of initial inliers (length %i) that didn't match the number of samples (%i). The vector of initial inliers should have the same length as the number of samples and contain only True (this sample is an initial inlier) and False (this one isn't) values."

@stefanv care to comment? This would be a significant policy change but I think it would be very positive.

skimage/measure/fit.py
```diff
@@ -664,7 +664,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
 
         where `success` indicates whether the model estimation succeeded
         (`True` or `None` for success, `False` for failure).
-    min_samples : int
+    min_samples : int or float in range [0, 1]
```
hmaarrfk: this is the kind of magic we should avoid. make `min_samples=None`. `min_samples` should mean `min_samples`. 

@jni wanted to make these keyword only agruments. Lets do that! 
Add a new parameter for the new functionality.

hmaarrfk: I'm really sorry for asking so much back and forth, but I feel strongly against this kind of magic.

I'm not against convenience, but for this particular case, it is coming at the expense of readability, and maintainability in the future.

You are correct, it is a one-liner, but it has serious implication for adding new functionality in the future.

skimage/measure/fit.py
```diff
@@ -765,7 +767,8 @@ def ransac(data, model_class, min_samples, residual_threshold,
     >>> dst[0] = (10000, 10000)
     >>> dst[1] = (-100, 100)
     >>> dst[2] = (50, 50)
-    >>> model, inliers = ransac((src, dst), SimilarityTransform, 2, 10)
+    >>> model, inliers = ransac((src, dst), SimilarityTransform, 0.5, 10,
```
hmaarrfk: You should create a new example for new functionality.

It seems that `min_samples=2` would return the same thing in `0.14.1` as it would when your patch comes in. Changing the examples makes it seem like users need to change from 2, to 0.5

skimage/measure/tests/test_fit.py
```diff
@@ -339,12 +339,31 @@ def test_ransac_dynamic_max_trials():
...
                residual_threshold=0, max_trials=-1)
+    # "`stop_probability` must be in range (0, 1)"
     with testing.raises(ValueError):
         ransac(np.zeros((10, 2)), None, min_samples=2,
                residual_threshold=0, stop_probability=-1)
+    # "`stop_probability` must be in range (0, 1)"
     with testing.raises(ValueError):
         ransac(np.zeros((10, 2)), None, min_samples=2,
                residual_threshold=0, stop_probability=1.01)
+    # "`min_samples` as ration must be in range (0, 1)"
```
hmaarrfk: Style comment:

I don't think these kinds of comments belong in tests. If you want to assert something about the kind of message passed, I think it is better to be explicit about it:
https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

Try to use the `message` parameter
This would ensure that in the future, developers respect the spirit of your error messages (which are useful!!!! thanks for adding them in) if they ever refactor the code.

skimage/measure/fit.py
Borda: I do not understand how set `min_samples` to 2 is the same as using 50% of 50 points?

skimage/measure/fit.py
Borda: in such case, the variable should be `nim_num_sample`... as the variable name is now I feel comfortable to set `min_samples` as 20 of 50 or just 0.5~50% of 50

skimage/measure/fit.py
hmaarrfk: I was confused about that too. So in the usecase that you showed, `min_samples=2` and `min_samples_fraction=0.5` with `init_inliners=np.ones(....` gives the same result? 

That definitely warrants an *additional* example (even if it is just a partial example)

skimage/measure/tests/test_fit.py
Borda: these tests were there before I just add comments to be easier to understand what are the test cases related for...

skimage/measure/fit.py
Borda: no, `initial_inliers` is completely different use-case then `min_samples`
I am against introducing new `min_samples_fraction` which is duplicated to the `min_samples`, it would become confusing for users what the primary and/or secondary parameter

skimage/measure/tests/test_fit.py
hmaarrfk: My apologies, I thought you were adding it there for maintainers to understand the kind of error message that should be returned.
 
As I'm reading this, the comments are necessary, because of the dual functionality of `min_samples`.

I don't think this kind of comment should be necessary if the added functionality you are requesting had a better parameter name.

skimage/measure/fit.py
hmaarrfk: I don't understand what `nim` means. Is it shorthand for something?

jni: @hmaarrfk @Borda I'm on the fence for this one. Specifically, my definition of magic is that involves guessing. Is any guessing involved here? Could do [0, 1] -> fraction, (1, None) -> int.

But, earlier I argued that someone who meant 1 as the minimum might have used 0.5 because they were unsure whether 1 is included or not when we say `min_samples=1`.

I think my favourite option is something like what @hmaarrfk suggested, but with @Borda's renaming suggestion:

```python
ransac(data, ..., *, min_num_samples=None, min_fraction_samples=None):
```

and deprecate the existing `min_samples` argument. It would be hard to express that one or the other of the above would be a required keyword argument.

I dunno, it's a hard situation. @scikit-image/core comments welcome here! @Borda thanks for your patience with this! Since it's an API issue and, as you can see, changing the API later is hard, we want to make sure we get this right before merging.

jni: @hmaarrfk nim is a typo of min, I'm pretty sure. :joy:

skimage/measure/fit.py
hmaarrfk: I completely agree that it is a different usecase.

As it is currently, you have erased an old use-case from the examples. 
The author of that example thought that it was worthwhile to include it.

I think you can either:
1. Show a new example with the added parameters.
2. Explain how in this case, `min_samples=2` and (`min_samples_fraction=0.5, init_inliers=np.ones(...`) give the same result.

it is like changing between wavelength, optical frequency, optical frequency in radians depending on some arbitrary threshold. It is really useful for a particular restrictive usecase, but it gets very confusing when you have to accomodate many different users.

skimage/measure/fit.py
hmaarrfk: I really don't mean to discourage this kind of convenience.

It is certainly useful to be able to define parameters in terms of numbers, or fractions. Quite often, one way of thinking is better for a particular problem.

Python allows us to define many different ways of providing parameters to a function. That said, I don't think the `units` of a named parameter should change based on an arbitrary threshold.

FYI: previously, if a fraction was provided, it would fail:
```
TypeError: 'float' object cannot be interpreted as an integer
```

this would also fail for `min_samples = 0.99999999`.

I also think of it in terms of optimization problems, where it might be useful to try and optimize the result, based on some kind of relaxed integer program. With this hybrid interpretation of min_samples, it requires the user to **undo** the guessing.

skimage/measure/fit.py
hmaarrfk: is `iterate` a better word here?

skimage/measure/tests/test_fit.py
hmaarrfk: I guess I subscribe to the philosophy that a comment is sometimes indicative of:

1. Strange behaviour
2. Something that belongs in the documentation.
3. Something that can use a better variable name.

In this case, the old code didn't have comments, because in my opinion, the original authors felt that the tests were explicit enough not to need them. This is just my interpretation, and an other location where I feel that an additional parameter would better serve the new behaviour you are seeking.

hmaarrfk: Please take this solely as a suggestion. Maybe a different way to think about it would be:

Would you have submitted a PR solely to add the 3 comments to the existing tests? If yes, then by all means, keep the comments. But if you feel like the comments are only now necessary due to the new functionality, I think this indicates somewhere where the contribution you are making can improve.

skimage/measure/fit.py
Borda: do you suggest to rename all arrays to iterable?

skimage/measure/tests/test_fit.py
Borda: I feel confused about the contradiction "As I'm reading this, the comments are necessary..." and "I don't think this kind of comment should be necessary..."

skimage/measure/fit.py
jni: No, `iterable of arrays of bool, shape (N,)`

skimage/measure/fit.py
Borda: it is just combining case in one example...

skimage/measure/fit.py
Borda: do you think that distinguishing them as a floating-point number (float/double/...) and fix-point number (int) is not enough?

skimage/measure/tests/test_fit.py
Borda: well if you say that the ransac(...) as it is won't ever change or evolve that comments are not needed... otherwise I think that some kind of comments is needed for easier navigation

skimage/measure/fit.py
jni: I hadn't realised that the existing behaviour is that float raises an error. @Borda can you confirm that this is indeed the case *for all NumPy >= 1.11*? If so, then I'm sort of ok with type-based polymorphism. This is how polymorphism works in C++, for example.

skimage/measure/fit.py
hmaarrfk: Honestly, I'm not sure. I'm having a really hard time wrapping my head around what the dimensionality should be here (just by reading the doc string). I feel like I need an example of having a list of arrays???? Am I interpreting this right?

you say: `list or tuple` but then in the examples you show an array: `init_inliers=np.ones(len(src), dtype=bool)`

I think iterable is the wrong word because then you would be supporting generator expressions.

Maybe `sequence`?

skimage/measure/tests/test_fit.py
hmaarrfk: Should have read "As I'm reading the code with the new functionality, the comments are necessary".

Comments are tricky. In one case, they are adding to the total amount you have to read. You now have multiple places where you can have typos -- we wouldn't want to starve ransac by skipping meals ;).

In developing tests, I assume that the developer will have the docstring next to him, so duplicating it isn't always worthwhile.

Maybe it is just late for me, and I've looked at my screen too much.

skimage/measure/fit.py
Borda: I think that it should be `array-like of bool, shape (N,)`

skimage/measure/tests/test_fit.py
Borda: yes but still developer has to think twice what invalid input the author wanted to cover by an individual call while you do not test the error message...
I see the "ration" typo, but no other

skimage/measure/fit.py
Borda: do you mean float larger then 1 will raise error< yes, it is included in the tests

skimage/measure/fit.py
hmaarrfk: Ok. If you think that is best.

Borda: I think that these two cases are not in contradiction and they are still clear...

Borda:
@jni @hmaarrfk could you please have a look at last comments on the review?

skimage/measure/fit.py
```diff
@@ -664,9 +664,9 @@ def ransac(data, model_class, min_samples, residual_threshold,
 
         where `success` indicates whether the model estimation succeeded
         (`True` or `None` for success, `False` for failure).
-    min_samples : int
+    min_samples : int or float in range [0, 1]
         The minimum number of data points to fit a model to.
-    residual_threshold : float
+    residual_threshold : float larger then 0
```
hmaarrfk:
```suggestion
    residual_threshold : float larger than 0
```

hmaarrfk:
@Borda, thanks for following up. Unfortunately, I still strongly object to having different behaviour based on the type of `min_samples`. To re-iterate, (because github may hide my previous comments):


The following code
```python
    if isinstance(min_samples, float):
        # min samples is a fraction here, always smaller than 1
        min_samples = int(min_samples * num_samples) 
        # min samples is a number here. an integer, always larger than 1
```
is doing what I consider to be "unit conversion". I find implicit unit conversions such as this one to be one of the hardest parts of my code to teach and explain to others. I think it is really bad style to have a variable, with the same name change units in a single block of code. I do not personally think it should be present in modern code. A unit change should be clearly expressed in code either through a different variable name, or through a prefix.

I think we should set scikit-image on a path where we can eventually have:
```python
def ransac(data, *, min_samples=None, samples_fraction=None):
    if samples_fraction is None and min_samples is None:
         raise ValueError('Must provide either min_samples or sample_fraction to ransac')
    if samples_fraction is not None and min_samples is not None:
         raise ValueError('Cannot provide both min_samples and samples_fraction to ransac')
    if samples_fraction is not None:
         min_samples = int(samples_fraction * num_samples)
```

It seems that this is difficult for this particular function due to the use of historical positional arguments. 

Python is a very expressive programming language, that makes it easy to build on other's code. To obtain the functionality you need, I suggest you create your own ransac function

```python
from skimage.measure import ransac as _ransac

 def ransac(data, model_class, samples_fraction, residual_threshold, *args, **kwargs):
     if not isinstance(data, list) and not isinstance(data, tuple): 
        data = [data]
     return _ransac(data, model_class, int(samples_fraction*data[0].shape[0], residual_threshold, *args, **kwargs)
```

I consider this kind of implicit API behaviour to be difficult to unteach, and bad to impose on other.

As is, these suggestions, as a whole, won't be getting a :+1: from me.

soupault:
>do you think that distinguishing them as a floating-point number (float/double/...) and fix-point number (int) is not enough?

I agree with Mark, it will be better to have a dedicated `min_ratio` parameter if the value of `1.0` is permitted. If we allow it and keep the code as is, I foresee a lot of issues downstream.

For me, it would be fine to set valid range for floats to (0, 1) or [0,1) (if it is of any use). Please, notice that, currently, there is an inconsistency in the code - all intervals are different:

    min_samples : int or float in range [0, 1]
    ....
    if not (0 < min_samples <= 1):
        raise ValueError("`min_samples` as ratio must be in range [0, 1)")

Borda:
@hmaarrfk I like your example with an envelope around the raw RANSAC function...
@soupault what shall we do wit this PR? I would add @hmaarrfk example do documentation...

skimage/measure/fit.py
```diff
@@ -800,13 +805,27 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
 
-    for num_trials in range(max_trials):
+    if isinstance(min_samples, float):
+        if not (0 < min_samples <= 1):
+            raise ValueError("`min_samples` as ratio must be in range [0, 1)")
+        min_samples = int(min_samples * num_samples)
 
-        # choose random sample set
-        samples = []
-        random_idxs = random_state.randint(0, num_samples, min_samples)
```
Borda: this is weak, some indexes can repeat
```python
len(set(np.random.randint(0, 100, 99)))
Out[15]: 63
```

skimage/measure/fit.py
```diff
@@ -666,9 +666,9 @@ def ransac(data, model_class, min_samples, residual_threshold,
 
         where `success` indicates whether the model estimation succeeded
         (`True` or `None` for success, `False` for failure).
-    min_samples : int
+    min_samples : int or float in range [0, 1)
```
jni: This still needs fixing.

skimage/measure/fit.py
```diff
@@ -776,6 +779,16 @@ def ransac(data, model_class, min_samples, residual_threshold,
             True,  True,  True,  True,  True,  True,  True,  True,  True,
             True,  True,  True,  True,  True], dtype=bool)
 
+    Les us present an example of transforming this function requesting mimimal number
```
jni: I would suggest changing this example to use `ransac` directly. We don't need to teach users to write wrapper functions, but rather, we need to propose alternative ways to use our functions, and it is up to them whether they want to wrap. Here's my attempt:

```
The argument `min_samples` is an absolute number. If instead we want to use a proportion
of the full dataset, we can multiply by the length of the data and cast to int:

>>> model_class = SimilarityTransform
>>> min_sample_ratio = 0.5
>>> min_samples = int(min_sample_ratio * len(src))
>>> model, inliers = ransac((src, dst), model_class, min_samples, ...)
```

skimage/measure/fit.py
jni: @Borda this line still needs removing.

skimage/measure/fit.py
Borda: nice, I have changed the second example accordingly

Borda:
@jni any idea why it passes on my [branch](https://travis-ci.org/Borda/scikit-image/builds/563601647)  but not here?

sciunto:
It's related to #4052 No worries.

Borda:
@jni @hmaarrfk could you please have a look at changes according to your review? Thx

skimage/measure/fit.py
```diff
@@ -756,18 +758,21 @@ def ransac(data, model_class, min_samples, residual_threshold,
     >>> sum(inliers) > 40
     True
 
-    Robustly estimate geometric transformation:
+    Robustly estimate geometric transformation. The argument `min_samples`
+     is an absolute number. If instead we want to use a proportion of the full
```
jni: indentation is off here. 

```suggestion
    is an absolute number. If instead we want to use a proportion of the full
```

But, more critically, I don't think the new rephrasing works, because "Robustly estimate geometric transformation" works as a section title, not as a sentence. How about:

"""
RANSAC can be used to robustly estimate a geometric transformation. In this section, we also
show how to use a proportion of the total samples, rather than an absolute number.
"""

skimage/measure/fit.py
```diff
@@ -756,18 +758,21 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
+     dataset, we can multiply by the length of the data and cast to `int`:
```
jni:
```suggestion
    dataset, we can multiply by the length of the data and cast to `int`:
```

skimage/measure/fit.py
```diff
@@ -756,18 +758,21 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
     >>> src = 100 * np.random.rand(50, 2)
-    >>> model0 = SimilarityTransform(scale=0.5, rotation=1,
-    ...                              translation=(10, 20))
+    >>> model0 = SimilarityTransform(scale=0.5, rotation=1, translation=(10, 20))
     >>> dst = model0(src)
     >>> dst[0] = (10000, 10000)
     >>> dst[1] = (-100, 100)
     >>> dst[2] = (50, 50)
-    >>> model, inliers = ransac((src, dst), SimilarityTransform, 2, 10)
+    >>> get_min_samples = lambda ratio: max(1, int(ratio * len(src)))
```
jni: Ach. I'm as much a proponent of functional programming as the next guy (more, even), but I don't think this is friendly to Python newcomers. It's also overkill in this example since it is used only once. So I think this should just be:

```python
ratio = 0.5  # use half of the samples
sample_size = int(ratio * len(src))
```

Then use that sample size.

skimage/measure/fit.py
```diff
@@ -785,54 +790,64 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
     random_state = check_random_state(random_state)
 
-    if min_samples < 0:
-        raise ValueError("`min_samples` must be greater than zero")
+    # in case data is not pair of input and output, male it like it
+    if not isinstance(data, (tuple, list)):
+        data = (data, )
+    data_lens = list(map(len, data))
+    assert len(set(data_lens)) == 1, 'all data have to have the same number of samples,' \
+                                     ' provided: %r' % [d.shape for d in data]
```
jni: This seems like a big change in behaviour now, that is unrelated to the original point of this PR? Additionally, @sciunto made the point recently that we should not use assert statements in validation code, as these are removed by Python when using optimized mode. See #4006 for details.

skimage/measure/fit.py
```diff
@@ -785,54 +790,64 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
+    num_samples = data_lens[0]
+
+    if not (0 < min_samples <= num_samples):
+        raise ValueError("`min_samples` must be in range (0, <number-of-samples>]")
```
jni: This contrasts to the documentation, which states that it must be in `[0, <nb-samples>)`.

skimage/measure/fit.py
```diff
@@ -785,54 +790,64 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
+    if residual_threshold < 0:
+        raise ValueError("`residual_threshold` must be greater than zero")
 
     if max_trials < 0:
         raise ValueError("`max_trials` must be greater than zero")
 
-    if stop_probability < 0 or stop_probability > 1:
-        raise ValueError("`stop_probability` must be in range [0, 1]")
+    if not (0 <= stop_probability <= 1):
+        raise ValueError("`stop_probability` must be in range (0, 1)")
```
jni: The brackets should not be changed here?

skimage/measure/fit.py
Borda: it is in the documentation or what do you mean
```
min_samples : int in range [0, <nb-samples>)
        The minimum number of data points to fit a model to.
```

skimage/measure/fit.py
Borda: sorry, my confusion with open and close intervals... 
image

skimage/measure/fit.py
Borda: btw, does it make sense to have there all samples? then RANSAC does not have any sense, right? so just (0, <nb-samples>)... @jni

skimage/measure/fit.py
```diff
@@ -699,6 +699,8 @@ def ransac(data, model_class, min_samples, residual_threshold,
         If RandomState instance, random_state is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`.
+    init_inliers : array-like of bool, shape (N,), optional
```
hmaarrfk: init makes me think of the `__inti__` callable. Should this be better off as `initial_inliers`

skimage/measure/fit.py
```diff
@@ -785,54 +790,61 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
 
-    if min_samples < 0:
-        raise ValueError("`min_samples` must be greater than zero")
+    # in case data is not pair of input and output, male it like it
+    if not isinstance(data, (tuple, list)):
+        data = (data, )
+    num_samples = len(data[0])
+
+    if not (0 < min_samples < num_samples):
+        raise ValueError("`min_samples` must be in range (0, <nb-samples>)")
```
hmaarrfk: `nb` is very hard to understand. I know this makes sense to us now, but trust me when I say that people not use to the lingo will not understand very clearly.

Just use `len(data[0])` or `number of samples`.

hmaarrfk:
Generally I think this is fine. I would simply avoid shortening the name of parameters since it hides their meaning.

skimage/measure/fit.py
Borda: I had there `number-of-samples` then @jni proposed change...
I would not write `data[0]` since it may be confusing if you have jut single data, not input/output data pair...

hmaarrfk: I think it is probably best you change both to be the same thing, `number-of-samples` maybe?

Borda: I have changed it here, but I would keep the doc notation as N to be consistent with the rest of the text...

skimage/measure/fit.py
```diff
@@ -666,9 +666,9 @@ def ransac(data, model_class, min_samples, residual_threshold,
 
         where `success` indicates whether the model estimation succeeded
         (`True` or `None` for success, `False` for failure).
-    min_samples : int
+    min_samples : int in range (0, <nb-samples>)
```
hmaarrfk:
```suggestion
    min_samples : int in range (0, <number-of-samples>)
```

jni:
@Borda I think probably I might have said `num_samples` rather than `nb-samples`? =) nb is not as common an abbreviation. I'm not too fussed about any of the proposed values, except for len(data[0]), which as you rightly pointed out could be incorrect.

hmaarrfk:
nb is pretty common, once youse seen it more than once ;) I just think we are a little past the 70s, so we can make use of vowels in the english language.

The second challenge is that non-english speakers will not associate nb with number.

skimage/measure/fit.py
Borda: oh, sorry, I forget this one...

hmaarrfk:
Awesome! thanks for this.

Issue #2820, opened 2017-10-01
`transform.hough_ellipse` raises error if called with an array of zeros (all black image)


Eskapp:
When calling `transform.hough_ellipse` with a all black image, the following error is raised:
```
Traceback (most recent call last):
  File "C:\Anaconda3\lib\site-packages\IPython\core\interactiveshell.py", line 2881, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-8-4d14c0917102>", line 3, in <module>
    result = hough_ellipse(np.zeros((100, 100)))
  File "C:\Anaconda3\lib\site-packages\skimage\transform\hough_transform.py", line 163, in hough_ellipse
    min_size=min_size, max_size=max_size)
  File "skimage/transform/_hough_transform.pyx", line 150, in skimage.transform._hough_transform._hough_ellipse (skimage\transform\_hough_transform.c:3881)
ValueError: Buffer and memoryview are not contiguous in the same dimension.
```
Linked to the SO question: https://stackoverflow.com/questions/46470492/buffer-and-memoryview-are-not-contiguous-in-the-same-dimension-in-python/46513914#46513914 (`scikit-image 0.12.3`)

and reproduced with `scikit-image 0.13.0` running the following:

```
import numpy as np
from skimage.transform import hough_ellipse
result = hough_ellipse(np.zeros((100, 100)))
```

I believe that the appropriate behavior would be to return an empty list.

stefanv:
Great catch, thanks @Eskapp.  Do you have time to prepare a PR to fix this?

Eskapp:
Depends on the deadline. I defended my PhD thesis last week and hence have some sharp deadlines this week!
However this sounds like an easy bug to fix and could be great to start off working on this repo.
I will have a look at the link you gave me here: https://github.com/scikit-image/scikit-image/issues/2765
(You can also assign this to me btw)

If the deadline is not too sharp, assign this issue to me.

stefanv:
Congratulations!  That's a big milestone in your life.

There's no urgent deadline for this, so I'm happy to have you take care of it next week.

soupault:
@Eskapp Hi! Did you have a change to look at the issue?

Eskapp:
I just updated my fork to look at it. I ran into another similar issue today using `hough_line_peaks`. When the `threshold` parameter does not allow the detection of any peaks, the function raises an `IndexError`.
[...]

PR #2996, opened 2018-03-27
Return empty list on flat images with hough_ellipse #2820

Eskapp:
## Description
Previous behavior: Calling `hough_ellipse` on an array of zeros (all black image) caused a `ValueError`
Current behavior: Calling `hough_ellipse` on an array of zeros (all black image) returns an empty array.

Other change: Fixes a typo in the function's documentation.

## Checklist
- [x ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
- [x ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [x ] Gallery example in `./doc/examples` (new features only)
- [x ] Unit tests

I tested the modified function on numpy arrays of zeros as well as on images with artificially created ellipses.

## References
If this is a bug-fix, it closes issue #2820 

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_hough_transform.pyx
+++ b/skimage/transform/_hough_transform.pyx
@@ -118,11 +118,10 @@ def _hough_ellipse(cnp.ndarray img, int threshold=4, double accuracy=1,
 
     Returns
     -------
-    result : ndarray with fields [(accumulator, yc, xc, a, b, orientation)]
+    result : ndarray with fields [(accumulator, yc, xc, a, b, orientation)], or None
           Where ``(yc, xc)`` is the center, ``(a, b)`` the major and minor
           axes, respectively. The `orientation` value follows
           `skimage.draw.ellipse_perimeter` convention.
-          Returns None if no non-zero pixel in the image.
 
     Examples
     --------
@@ -147,6 +146,8 @@ def _hough_ellipse(cnp.ndarray img, int threshold=4, double accuracy=1,
     if img.ndim != 2:
             raise ValueError('The input image must be 2D.')
 
+    # The creation of the array `pixels` results in a rather nasty error when the image is empty.
+    # As discussed in GitHub #2820 and #2996, we opt instead to return None.
     if not np.any(img):
         return None
 
--- a/skimage/transform/_hough_transform.pyx
+++ b/skimage/transform/_hough_transform.pyx
@@ -118,7 +118,8 @@ def _hough_ellipse(cnp.ndarray img, int threshold=4, double accuracy=1,
 
     Returns
     -------
-    result : ndarray with fields [(accumulator, yc, xc, a, b, orientation)], or None
+    result : ndarray with fields [(accumulator, yc, xc, a, b, orientation)],
+          or None
           Where ``(yc, xc)`` is the center, ``(a, b)`` the major and minor
           axes, respectively. The `orientation` value follows
           `skimage.draw.ellipse_perimeter` convention.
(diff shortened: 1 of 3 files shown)


skimage/transform/_hough_transform.pyx
```diff
@@ -146,7 +146,12 @@ def _hough_ellipse(cnp.ndarray img, int threshold=4, double accuracy=1,
     if img.ndim != 2:
             raise ValueError('The input image must be 2D.')
 
-    cdef Py_ssize_t[:, ::1] pixels = np.row_stack(np.nonzero(img))
+    cdef cnp.ndarray[ndim=2, dtype=cnp.intp_t] noncontiguous_mem_pixels = np.row_stack(np.nonzero(img))
```
jni: Why isn't this just `if pixels.shape[1] == 0: return None`? 

You should also add a test in `skimage/transform/test_hough_transform.py`.

Eskapp: I created a new array because it is the type of the variable we want to create that causes the error when the image has no non-zero pixel. (`ValueError: Buffer and memoryview are not contiguous in the same dimension.`)

However, I think that the choice of using a contiguous array for a faster processing of all the calculations below this line should not be changed.

The only thing I could come up with was creating a non-contiguous array, testing for its shape, and transform this array into a continuous array if non-zero pixels are present.

I'm also going to add the test. As a side note, this is my first contribution and the first time I work with Cython :)

jni: @Eskapp Well, welcome! =) And thank you for this! I hope it won't be the last! 

Ah! I see now why. It still seems clumsy to me. I suggest the following:

```python
if not np.any(img):
    return None
```

And then create `pixels` afterwards. What do you think?

jni:
@Eskapp very nice, thank you! Some very minor additional comments from me and then this is good to go as far as I'm concerned!

sciunto:
A remark: do we prefer to return None or to return an empty list? An empty list has the advantage to avoid extra codes like `if blah is not None:` on the user side.

Eskapp:
Yes, I will modify this.
However, a solution with zeros everywhere will somehow give the impression that an ellipse has been found even when it has not. (That actually was my thought when I first chose to use `None`).
However, your solution seems more coherent with how `hough_circle` and `hough_line` which do not return `None` values. @jni

jni:
@Eskapp note that it's not an array of zeros, but an array of size zero. So downstream applications can get the exact same information: instead of `if result is None`, you would check `if result.size == 0` or `if len(result) == 0` (depending on your style preference).

Eskapp:
Yeah true I read a bit too fast :) Working on it in the forthcoming days!

sciunto:
@Eskapp What's the status of this PR? did you have a chance to have a look?

Eskapp:
@sciunto Yes, my last commit changes the returned value from None to an empty array as decided.
I do not know why the Travis tests are failing though. Things work on my side.
It is my first contribution and I do not know how to deal with this situation. I'll be happy if you can guide me through this :)

skimage/transform/_hough_transform.pyx
```diff
@@ -118,7 +118,8 @@ def _hough_ellipse(cnp.ndarray img, int threshold=4, double accuracy=1,
 
     Returns
     -------
-    result : ndarray with fields [(accumulator, y0, x0, a, b, orientation)]
+    result : ndarray with fields [(accumulator, yc, xc, a, b, orientation)],
+          or None
```
sciunto: Please, fix the docstring. (Remove or None)

skimage/transform/hough_transform.py
```diff
@@ -131,7 +131,8 @@ def hough_ellipse(image, threshold=4, accuracy=1, min_size=4, max_size=None):
 
     Returns
     -------
-    result : ndarray with fields [(accumulator, y0, x0, a, b, orientation)]
+    result : ndarray with fields [(accumulator, yc, xc, a, b, orientation)],
+          or None.
```
sciunto: same here.

sciunto:
Don't worry about travis, the failures are not due to your PR ;)

Just a minor thing to fix, and it will be good to me. Thanks!

jni:
Thanks @sciunto! =) and @Eskapp! This is a great usability improvement!

PR #2999, opened 2018-03-30
Add release notes for 0.13.1

jni:
It appears I never included these in any commit! Oops!


stefanv:
Agreed that the formatting is perhaps not ideal.  But merging until that discussion can happen.

PR #3000, opened 2018-03-30
WIP: Remove Python 2 compatibility

jni:
## Description

As we near release 0.14 (our last Python 2-compatible release) *and* issue #3000, I thought it would be appropriate to start this work. =D

## Checklist
- [ ] Unit tests

## For reviewers

  later.


JDWarner:
I sincerely hope we have airspeed velocity up and working before we merge this - also worth noting that despite what the release notes say, [`np.dot` and `@` are not drop-in equivalent](https://stackoverflow.com/questions/34142485/difference-between-numpy-dot-and-python-3-5-matrix-multiplication) if one argument is a scalar or the arrays have higher dimensionality.  Both cases we need to be concerned with.

jni:
@JDWarner yes, I am aware about `@` (came across this as I tried to implement it for dask!). In this case I tried to make sure that each use was between 2D and/or 1D arrays. Indeed there's one use of `np.dot` left in the codebase, in nD moments.

And very good point about airspeed. Let's make sure we have that up and running before merging this.

sciunto:
@jni, I think you missed this:
skimage/measure/_regionprops.py:639:            # For Python 2.x

:)

stefanv:
Well played @jni :)

jni:
:fireworks: 

CC @Carreau ;)

soupault:
This is HUGE! Thanks @jni :clap:

Issue #3007, opened 2018-04-08
threshold_niblack and threshold_sauvola sometimes return NaN when calling them with float64 images


gregsadetsky:
## Description

`threshold_niblack` and `threshold_sauvola` sometimes return NaN when calling them with float64 images

## Way to reproduce
- [X] Code example

```
# the code below will return NaN values only.
# depending on the image's pixel values,
# it's possible to get back a mix of NaN and float values

import numpy as np
from skimage import filters

# 16 pixels, single channel
src_img = np.array([0.03082192 + 2.19178082e-09] * 16).astype('float64')
src_img = src_img.reshape((4,4))

# will print array of 4x4 NaNs
print(filters.threshold_niblack(src_img))

# will print array of 4x4 NaNs
print(filters.threshold_sauvola(src_img))
```

the reason for the NaNs is that `sqrt` is being called in [_mean_std](https://github.com/scikit-image/scikit-image/blob/master/skimage/filters/thresholding.py#L775) with negative numbers.

- [X] Operating system and version

OS X 10.11.6

- [X] Python version

2.7.10

- [X] scikit-image version (run `skimage.__version__`)

0.14dev

I am using the last commit as of now, da0800a2a0c4abb51c621ff495c3297cc69d43a1

jni:
@gregsadetsky thanks very much for the minimal example — it makes our job so much easier! See #3008.

As an aside, I'll take this opportunity to gently prod you to adopt Python 3 if possible. Our upcoming 0.14 release will be the last one to support Python 2.7. (Though we will backport bug fixes to the 0.14 line until Python officially drops support in 2020.)

gregsadetsky:
wow, incredible to already see a fix! and "my" magical pathological values in the test codebase! haha :-)

thank you x 1000000!!

and yes on py3... old habits die hard, etc. ;-) I should be good by 2020! :-)

gregsadetsky:
would you prefer to keep this open while #3008 hasn't been merged, or is this good to close now considering that a fix has already been created?

jni:
Haha well, as I mentioned, your example made the fix easy! (As well as you finding the exact cause of the error!) You can see from our open issues that our response time has a long tail, sadly. =)

We typically keep things open until the fixes get merged. GitHub will automatically close this when #3008 is merged, so it should be all good right now. =)

soupault:
Closed via #3008

PR #3008, opened 2018-04-09
Ensure stdev is always nonnegative in _mean_std

jni:
For some values, floating point error can cause VarX = E(X^2) - (EX)^2
to be negative, resulting in NaNs when taking StdX = sqrt(VarX). Here
we ensure that VarX is nonnegative before taking the square root.

## Checklist
- [x] Unit tests

## References
Closes issue #3007 

## For reviewers

  later.


soupault:
Thank you @jni !

PR #3011, opened 2018-04-09
add Singh threshold

rodrigo2019:
I was working with characters segmentation and the Singh threshold got better results than Sauvola and Niblack, so I decided to share.

Here is the paper: https://arxiv.org/abs/1201.5227

### DIFF
--- a/doc/source/api_changes.rst
+++ b/doc/source/api_changes.rst
@@ -1,5 +1,6 @@
 Version 0.14
 ------------
+- ``skimage.filters.thresholding.threshold_singh`` added.
 - ``skimage.filters.gaussian_filter`` has been removed. Use
   ``skimage.filters.gaussian`` instead.
 - ``skimage.filters.gabor_filter`` has been removed. Use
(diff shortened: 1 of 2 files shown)


jni:
Hi @rodrigo2019, and thanks for the PR!

Unfortunately, it doesn't actually implement what they propose in the paper:

> In other techniques like Sauvola’s and Niblack’s technique, local mean m(x,y) and standard
> deviation δ(x,y) are required to determine the value of the threshold for each pixel. In this
> proposed technique, no local standard deviation is required and it requires to compute the
> local mean and mean deviation ∂(x,y) to determine the local threshold
> [...]
> where ∂(x,y)=I(x,y) - m(x,y) is the local mean deviation and k is a bias which can control the
> level of adaptation varying threshold value.

Generally I think the paper is dishonest — for example, the "computational time" figure (3) appears to show Sauvola thresholding *without* using the integral image. Sauvola requires only one more pass over the image (to compute the deviation), so surely it must be within a factor of 2 or 3 of the proposed method, not 20x as they purport to show.

So, in my opinion both the method in the paper and the method implemented are unproven, and shouldn't be added to scikit-image. (There is a cost to increasing complexity.) I'm 👎 for merging. I'll let others weigh in though.

rodrigo2019:
@jni I agree with you, using threshold_sauvola and threshold_singh the speed are almost the same.

sciunto:
I'm closing as WONTFIX. Feel free to comment / reopen if needed.

Thank you.

PR #3020, opened 2018-04-18
Add pyproject.toml, now supported by pip

stefanv:
This lets pip figure out that numpy is required to build skimage,
significantly simplifying the checks that need to be done in setup.py

For backward compatibility, we do not update setup.py here.  But we do
clean up AppVeyor and Travis where possible.

Documentation is also updated to simplify installation instructions, and
to include the requirements/* files, instead of copying their
information.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -41,7 +41,7 @@ install:
   # The --pre flag is necessary to grab a SciPy wheel, which is in
   # pre-release at the time of writing (03-10-2017)
   - pip install --pre -r requirements.txt
-  - pip install wheel
+  - pip install -r requirements/build.txt
   - pip wheel -w dist .
   - ps: "ls dist"
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -41,8 +41,8 @@ install:
   # The --pre flag is necessary to grab a SciPy wheel, which is in
   # pre-release at the time of writing (03-10-2017)
   - pip install --pre -r requirements.txt
-  - pip install -r requirements/build.txt
-  - python setup.py bdist_wheel bdist_wininst
+  - pip install wheel
+  - pip wheel -w dist .
   - ps: "ls dist"
 
   # Install the generated wheel package to test it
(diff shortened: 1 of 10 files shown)


.appveyor.yml
```diff
@@ -41,8 +41,7 @@ install:
   # The --pre flag is necessary to grab a SciPy wheel, which is in
   # pre-release at the time of writing (03-10-2017)
   - pip install --pre -r requirements.txt
-  - pip install -r requirements/build.txt
-  - python setup.py bdist_wheel bdist_wininst
+  - python wheel .
```
JDWarner: AppVeyor isn't happy with this line.

stefanv: Hehe

matthew-brett:
Only to say that Scipy just had to remove its `pyproject.toml` because it was messing up the by-hand specifications of numpy versions necessary for building (e.g.) wheels : https://github.com/scipy/scipy/pull/8735

stefanv:
Can you help us understand their issue? I am not sure what "environment 
markers" are.

matthew-brett:
I think environment markers are things you use to change the package version dependencies according to - say - the Python version - which is what many of us do, because later Pythons only work with more recent Numpy versions, whereas we want to build against earlier Numpy versions, for older Pythons, for greater compatibility.

stefanv:
I think we currently have only the requirement `numpy >= 1.11`, and pip does not reinstall `numpy` if it is already available.  So I *think* we should be fine, but maybe I am missing a subtlety.

stefanv:
OK, removed `pyproject.toml` for now; there are still quite a few cleanups that can be merged here.

emmanuelle:
Thanks a lot for all curating work. I'm glad you caught this random_walker problem, I couldn't figure out what was going on. Some weird Travis errors though, does it make sense for you?

soupault:
Travis failures seem to be related to the PR. I've restarted the build with an empty cache, and the result is still the same.

stefanv:
Looks like this PR was premature.  A couple of the fixes required to make tests pass were incorporated elsewhere already, so closing.

Issue #3016, opened 2018-04-13
Enhancement of peak detection features


sciunto:
Following this [thread on the mailing list](https://mail.python.org/pipermail/scikit-image/2018-April/005514.html), I initiate this issue to enhance the peak detection in scikit image.

Currently, we have two functions:
* [skimage.morphology.local_maxima](http://scikit-image.org/docs/dev/api/skimage.morphology.html#skimage.morphology.local_maxima)
* [skimage.feature.peak_local_max](http://scikit-image.org/docs/dev/api/skimage.feature.html#peak-local-max)

In particular, `peak_local_max` (and derivatives) is used in scikit-image at different places.

Our functions seem to be sub-optimal and lead to questionable peak detection, especially for plateaus as pointed out on the mailing list. It is also questioned to keep two separate functions.

Recently, @lagru committed to scipy https://github.com/scipy/scipy/pull/8264 a new function to detect peaks based on their prominence. This function is particularly easy to use and efficient for 1D data. Then, it would be worth to consider whether it can be generalized to 2D data. @lagru, would you like to share your expertise on this and eventually contribute to scikit-image if you wish and if it makes sense to you?

Thanks.

jni:
Hi @ThomasWalter, good to see you again! =)

I fully agree that multiple algorithms are totally ok, just like we have multiple segmentation or thresholding algorithms. However, we have more than once found that whatever algorithm `peak_local_maxima` implements, it's doesn't seem very good. If you follow @sciunto's link to the mailing list, you'll see an example I came up with that gave very surprising results, to say the least.

In short, I vote for multiple *functional* algorithms. =P

stefanv:
I'd like peak detection algorithms to at least do what is intuitively expected.  And there really aren't that many solutions that pass that test.  Flat peaks?  Oh, ok, it highlights all pixels. Or:  Flat peaks?  Oh, ok, it highlights the center of the peak.

ThomasWalter:
1. OK, let us start with this definition: a local maximum is a region of constant grey level, for which all neighbours inside the image have a strictly lower grey level. The only uncertainty in this definition is thus what happens with an entirely constant image (I would just say, that in this case, there is no maximum). This is the only unambiguous definition I know. As soon as you start filtering minima, reducing the flat zones, etc. you basically always run into trouble. This functionality is currently implemented in local_maxima, and it works for all data types. 
[...]

Issue #2876, opened 2017-11-12
Add flood fill algorithm


soupault:
## Description
https://en.wikipedia.org/wiki/Flood_fill

## References
Old mailing list thread - https://groups.google.com/forum/#!topic/scikit-image/n8lO2b_a3DQ.
Inspired by the discussion in https://stackoverflow.com/questions/47213246.

Please, share your thoughts.

JDWarner:
I've got a basic implementation of a stack-based flood fill for uint8 images I'll toss up as a PR, if nothing else as a starting point.

lagru:
I hope I'm not completely misjudging this but I basically implemented this already in #3022. It is essentially a queue based flood fill with a few additions to determine if a plateau is a maximum. I think that could be easily adapted to satisfy this request.

What do you think?

JDWarner:
@lagru looks like you're right, didn't see that as part of the peak changes.  I don't necessarily want to rope you into things out of the scope of your existing PR, but if there is the possibility to take some of that algorithm into, e.g., `skimage._shared` and re-use it in a directly exposed flood fill, that would definitely be better than redundancy!

lagru:
I would love to to that however I don't think reusing the same flood fill is possible in this case. The inner queue-consuming loop of my function has more logic than necessary for a basic flood-fill and I don't see an elegant way to move that logic outside. So I don't think a shared Cython function is the answer here. There are some Python parts that were adapted from the watershed algorithm so it might be possible to generalize and re-use those.

I mentioned this because it would be easy to take the basic skeleton of my code and with a few modifications use it as a basis for a n-dimensional flood-fill. I'm currently focused on #3022 but after that one is ready I'd be happy to do the work myself. Of course only if you agree and you haven't already gone ahead and solved this need with your own already working suggestion. :slightly_smiling_face:

jni:
@lagru 

> has more logic than necessary for a basic flood-fill and I don't see an elegant way to move that logic outside. So I don't think a shared Cython function is the answer here. 

Actually I think it *is*. Let's be ambitious here: what is a general floodfill? 

- initial position
- definition of neighbourhood
- predicate about each point (and its neighbours) defining whether to add it to the fill or stop

[...]

PR #3022, opened 2018-04-22
Replace morphology.local_maxima with faster flood-fill based Cython version

lagru:
This is a follow-up to #3016 that <strike>demonstrates a first draft of alternative algorithms</strike> proposes an alternative flood-fill based algorithm in Cython for maxima detection in images (and ND arrays). You can find a demo and benchmarks [here](http://nbviewer.jupyter.org/gist/lagru/4cd7ab94a101bbea15af216a3d44d1ca).

## Description
<strike>This PR is meant to be a place for discussion first and is not necessarily meant to be merged.

The current state is work in progress and</strike> I'm sure there is room for improvement. However I felt that I have reached a point where it would be useful to start a discussion and review process. There are some open points that will / could have a large impact on the functions performance and implementation which I list below or as review comments. Furthermore I felt a little out of my depth concerning some C/Cython related stuff. Especially while implementing the re-sizable buffer. So if you see any bad practices please point them out! I'm eager to improve.

For a description of the algorithms itself please have a look at the docstrings and code comments. Including it here wouldn't make much sense <strike>as these are still WIP</strike>.

Oh, and just because it isn't said enough: Thank you guys for your work and time. scikit-image is a great tool and library!

## Checklist
I'll try to add talking points and arguments you bring up here.

### General decision points
- [x] **(A)** Can plateaus that border the image edge be maxima? What if the entire image has only one value?
--> I'll just make this toggle-able. For now I'll set the default to exclude maxima on borders.

> Intuitively I would assume that local maxima are detected according to the ["mathematical" definition](https://en.wikipedia.org/wiki/Maxima_and_minima#Definition) which demands smaller values on all sides of a maximum. This would mean that maxima can't border the image edge and a constant image has no maximum.

> [...] don't return maxima on the edges. However, it would be easy to add an option to include them by using padding.

- [x] **(B)** Should the function return index positions (like `feature.peak_local_max`) or a "boolean" array with the same shape as the image (like `morphology.local_maxima`)?
--> Return array of 1's and 0's. Optionally add a parameter to change to indices.

> [...] we should return whatever is more natural for the algorithm (in this case, a boolean array), and provide utility functions to convert between them.

- [x] **(C)** Which variant of the algorithm should be used? [Benchmarks here...](http://nbviewer.jupyter.org/gist/lagru/4cd7ab94a101bbea15af216a3d44d1ca)
--> It seems that algorithm A (prefiltering the array in one direction) has a better performance for all test cases. So I'll prefer that one for now.

- [x] **(D)** Is this a candidate for SciPy's ndimage module or for scikit-image? If the latter, which module?
--> I'll post and ask on the SciPy mailing list when I'm confident with the progress.

> Even if it is integrated to scipy, it makes sense (IMO) to integrate temporarily to scikit-image to avoid bump the dependency version to the latest version and to do not wait a scipy release to use internally in scikit-image.

- [x] **(E)** Do you actually want this as an addition to the respective library? :wink: 

### Final tasks
  --> Already exists [here](http://scikit-image.org/docs/dev/auto_examples/segmentation/plot_extrema.html)
- [x] Unit tests

## References
- The [thread on the mailing list](https://mail.python.org/pipermail/scikit-image/2018-April/005514.html) that started it all.
- The follow-up issue #3016. 

## For reviewers

  later.


jni:
Thanks @lagru! Fantastic to get identical results to local_maxima with a 10-20x speed advantage!

To answer your decision points, from my perspective:

- A. This is my view: don't return maxima on the edges. However, it would be easy to add an option to include them by using padding.
- B. I think we should return whatever is more natural for the algorithm (in this case, a boolean array), and provide utility functions to convert between them.
- C. This I don't have a strong view on. Algo B is simpler, so I lean towards it, but algo A does appear to have a speed advantage in all cases. But, see my comments later.
- D. This is a good fit in scikit-image, either in feature or morphology. However it might make sense in scipy.ndimage, but I can't comment on that.
- E. Erm... Yes. =P

A major point in this review is that this function and these algorithms would make complete sense in n dimensions, rather than restricting to 2D, and also with varying levels of connectivity (4- or 8-connected in 2D, 6-, 18-, or 26-connected in 3D, though I prefer the terms 1-connected, 2-connected, 3-connected for the same). This would be easier for algo B than A, I think.

I've been meaning to write a guide about working in nD vs 2D, but in the meantime, I suggest looking at the morphology.watershed code for an example. The key is to think in indices and neighbours instead of rows/columns and -1/+1 on each.

One potential limiting factor is that, if I remember correctly, we have flood fill in 2D and 3D but not nD. But just 2D/3D is a big improvement over 2D-only.

lagru:
> A major point in this review is that this function and these algorithms would make complete sense in n dimensions, rather than restricting to 2D, and also with varying levels of connectivity (4- or 8-connected in 2D, 6-, 18-, or 26-connected in 3D, though I prefer the terms 1-connected, 2-connected, 3-connected for the same). This would be easier for algo B than A, I think.

I would generally agree. However I have the fear that that would incur noticeable speed penalties for the 2D case. 

> I've been meaning to write a guide about working in nD vs 2D, but in the meantime, I suggest looking at the morphology.watershed code for an example. The key is to think in indices and neighbours instead of rows/columns and -1/+1 on each.

That seems well documented but a little bit overwhelming to grasp at first glance. I think I get that you actually iterate each pixel on the flattened array. The neighbors are found by using offsets from the current index. I don't yet understand how these offsets are computed without using to much memory. E.g. for a connectivity of 2 in the 2D-case 8 offsets for each pixel (except for the edges) are required. So I would need memory at least 8 times the size of the provided image. This gets even worse for higher dimensions...

sciunto:
Awesome work. Many thanks! I reply to A, B and D

A. I agree with your intuition :)
B. +1 with @jni 
D. Even if it is integrated to scipy, it makes sense (IMO) to integrate temporarily to scikit-image to avoid bump the dependency version to the latest version and to do not wait a scipy release to use internally in scikit-image.

ThomasWalter:
Amazing. That's really cool. 

What I do not understand is: why do you get the same results as `extrema.local_maxima`. Because you assume that a plateau that is higher than any pixel inside the image but touching the border is not a maximum. But in `extrema.local_maxima` this would be a maximum. So, in principle you should get some differences, no? 

Another question I am asking myself: where does this enormous speed gain come from? Is this algorithmic or implementation?

lagru:
> What I do not understand is: why do you get the same results as extrema.local_maxima. Because you assume that a plateau that is higher than any pixel inside the image but touching the border is not a maximum. But in extrema.local_maxima this would be a maximum. So, in principle you should get some differences, no?

(Assuming you mean `morphology.local_maxima`) In the current implementation (b500723) touching the border is not a disqualifying criterion thus results is identical with `local_maxima`. According to the discussion around (A) I already have a modification in my local repo that prohibits maxima from touching the border. The performance is about the same.

> Another question I am asking myself: where does this enormous speed gain come from? Is this algorithmic or implementation?

I think the implementation in Cython plays definitely a role and `local_maxima` has a lot more Python around its core function `_greyreconstruction.reconstruction_loop`. Profiling would perhaps give more insight how much of that is due to the Python overhead and how much time is consumed.
At least in for algorithm A there are some algorithmic advantages because the first loop, which considers only one dimension, is relatively efficient and reduces the amount of evaluated pixels for following queue-based algorithm. For some images this can be really noticeable.

ThomasWalter:
OK cool, thanks a lot for your comment. 

By the way, I think it is better to have at least the option to also have maxima that touch the border. Actually, there is no mathematically sound reason for excluding or including them in the first place (I mean, it is just a choice), but for some morphological operators it is important to have all maxima to start with.

lagru:
@ThomasWalter I should have added that excluding maxima at borders simplifies the implementation of algorithm A and more importantly allows the first loop to exclude more pixels. Especially for "unnatural" images which have large plateaus which may get excluded early this can reduce the evaluation burden for the second queue-based part!

jni:
@ThomasWalter I'm going from memory/intuition here rather than a full overview of the code, but if I remember correctly, the morphological reconstruction will run a dilation on the full image multiple times, enough to cover any plateaus. On the other hand, this algorithm only needs one pass over the image + some flood filling, so it can potentially be much faster.

Regarding the borders, we can reduce borderless to border-ful by padding with the minimum value. (Though this may not be the fastest/most efficient approach).

@lagru 

> for a connectivity of 2 in the 2D-case 8 offsets for each pixel (except for the edges) are required. So I would need memory at least 8 times the size of the provided image.

No, you compute the offsets "just in time", when you are observing each pixel/voxel. So you compute an offset array (a 4-, 8-, 6-, 18-, or 26-element array of signed ints) and when you check a pixel you compute its neighbors by adding the pixel coordinate to that array. You do *not* broadcast all pixel coordinates against the offset array! That would be [crazy](https://github.com/jni/gala/blob/24290bc8e954ff38c9a2c1e1d3896367a9791c40/gala/morpho.py#L643). =P (In my defense, I wrote this code a very long time ago. =D)

lagru:
Alright, I got an ND-version of algorithm A to work and it seems like its even slightly faster than the 2D versions for the 2D case. :D I updated the [benchmark notebook](http://nbviewer.jupyter.org/urls/gist.githubusercontent.com/lagru/4cd7ab94a101bbea15af216a3d44d1ca/raw/5906c6e67d8144a564af27ada8cdebd657f4b350/benchmark_maxima.ipynb) as well so you can see for yourself.

skimage/morphology/extrema.py
```diff
@@ -356,3 +358,62 @@ def local_minima(image, selem=None):
...
+    # Possible states are:
+    #   3 - first or last value in a dimension
+    #   2 - potentially part of a maximum
+    #   1 - evaluated
+    flags = np.zeros(image.shape, dtype=np.uint8)
+
+    # Set edge values of flag-array to 3
+    without_borders = (slice(1, -1) for _ in range(flags.ndim))
+    flags = np.pad(flags[tuple(without_borders)], pad_width=1,
+                   mode="constant", constant_values=3)
```
lagru: There must be a better / faster way to do this right? But I can't think of a way to get a view or slice of only the "edge" values (the inverted of `without_borders`).

skimage/morphology/_extrema_cy.pyx
```diff
@@ -0,0 +1,191 @@
...
+            return 1
+        return 0
+
+    void _q_grow_buffer(Queue* self) nogil:
+        """Double the memory used for the buffer."""
+        cdef QueueItem* new_buffer
+
+        # TODO prevent integer overflow!
+        self._buffer_size *= 2
+        new_buffer_ptr = <QueueItem*>realloc(
```
lagru: I expect that there is a constant or variable that stores the maximal possible size of `Py_ssize_t` but I can't find it.

lagru:
Some additional question:

- Is the `morphology` module the right place for the ND-version?
- Should this replace the existing function `local_maxima` or be an alternative? There are several consequences:
  - If we replace the old function we would have to accept that peaks can border the array edge by default otherwise we would break the API.
  - If we don't replace the function... how do we name the new one to be distinct. Ideally there should be a clear recommendation for which function to prefer.
- The queue and many of the changes to make the algorithm n-dimensional were inspired by the watershed algorithm. So there is some overlap in functionality. 
  - E.g. do you desire to merge my queue and the functionality in `heap_general.pxi`. However there are some important differences between the two "data structures" (heap considers insertion time and my queue can be restored) so I'm hesitant to call this a good idea.
- `peak_local_max` features several arguments that can be used to restrict the set of found maxima. Should I keep my new function pure (do only one thing and do it good) or should I think about integrating arguments like a height condition, area, etc. into the function? 
  - I'm undecided on this. E.g. a height condition could be used to speed up the algorithm which is a plus in my book. But I think other things should be provided by separate well defined functions if so desired.

skimage/morphology/_extrema_cy.pyx
```diff
@@ -0,0 +1,190 @@
...
+    cnp.uint8_t
+    cnp.uint16_t
+    cnp.double_t
+
+
+ctypedef Py_ssize_t QueueItem
+
+
+cdef:
+    struct Queue:
```
emmanuelle: Any chance we could re-use a data structure already present somewhere else in order to limit the complexity of the code?

emmanuelle: Oh I see that you've mentioned the Queue of the watershed. A common codebase would be great if it makes sense.

stefanv:
I don't have anything to add at this point, but want to express my excitement at this conversation happening.  Thanks!

skimage/morphology/_extrema_cy.pyx
lagru: I agree, a common codebase would be ideal (see this comment as well https://github.com/scikit-image/scikit-image/issues/2876#issuecomment-385241507). However as already mentioned there are some problematic differences between the two queue implementations and its requirements which I will try to summarize below:

Heap (general_heap.pxi) is a priority queue used by watershed, Queue is used in my code:
- Multiple iterations:  Queue allows multiple iterations via `q_restore` and preserves a consumed (popped) item internally until cleared explicitly. Heap may overwrite consumed data. 
- Order: Queue uses insertion order while Heap sorts (using the array of pointers) by the return value of a function `smaller`.
- A consequence the internal data representation is different as well: Queue uses only one array to store its items while Heap uses an additional array to store pointers to its items. 

@emmanuelle You seem to be the original contributor of Heap so please point out any misunderstandings on my part.

I don't see much overlap here concerning the internal machinery of these "basic" containers. But it may be possible to implement a data container that fulfills the requirements of both Heap and Queue. But would that be wise? I think my code would be slower due to overhead introduced by the sorting machinery and watershed would have to suffer the larger memory footprint due to internal preservation of popped items.

Having said that, if you still think it might be worth it I'll try to make a quick draft of a queue that can be used as a replacement for both. That might be more useful to reach a more informed decision.

jni: No no, a FIFO Queue and a Heap are two different data structures, even if the latter is used to implement Priority Queues. I don't think there is much code to be shared between them and it makes sense to keep them separate.

lagru:
Okay, the new algorithm seems to pass the unit tests of the old `local_maxima` and `local_minima`.
Before I start writing tests and proper docstrings I'd like some decisions to be made regarding the API:

- [ ] **(F)** Should the new algorithm replace the old one in `local_maxima`? The output seems to be the same for all scenarios I tested (benchmarking and unit tests). If not how should this new function be named?
- [ ] **(G)** What's the scope of this function to be? Should this just find maxima? Or should additional options like for `peak_local_max` be integrated?
- [ ] **(H)** How do we handle code that shares functionality among each other, e.g. new `_offset_to_raveled_neighbours` was derived from `_compute_neighbors` in watershed.py.

To F: I would prefer to replace the old slower one. Mainly to avoid ambiguity between 3 maxima finding functions and because I can't think of a more fitting name than `local_maxima`. This wouldn't need a depreciation treatment because the new one supports all the features of the old one while being faster in all cases.  I would also recommend to depreciate `local_minima`. It seems trivial to me to invert an image and use `local_maxima`.

To G: I like the current scope: find local maxima depending on connectivity and definition at the edges. Convenient filter options could be provided with another function, e.g. `select_maxima`, that wraps this one and would provide the means to select the desired maxima out of all found ones. At least that is what I think I should have done with `find_peaks` in SciPy.

To H: Honestly, I would prefer to worry about that in another PR after the API and scope are clear.

jni:
@lagru my 2c:

- F: this function should replace the existing `local_maxima`.
- G: long term I would like to see this replace `peak_local_max` (this would require some processing of plateaus but otherwise is almost a straightforward `np.nonzero`!) But I agree with you that this PR is not the place to worry about this.
- H: that question is out of scope for this PR.

I want to keep `local_minima`. Many functions are trivial to obtain by composition (e.g. `rgb2lab` is a composition of `rgb2xyz` and `xyz2lab`), but they are there because they are common enough that people find it convenient.

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,358 @@ def h_minima(image, h, selem=None):
...
+
+
+def _fast_pad(image, value):
+    """Pad an array on all axes with one constant value.
+
+    Parameters
+    ----------
+    image : ndarray
+        Image to pad.
+    value : number
```
sciunto: Minor: in scikit-image, we usually write `value : scalar`

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,358 @@ def h_minima(image, h, selem=None):
...
+def _offset_to_raveled_neighbours(image_shape, selem):
+    """Compute offsets to a samples neighbors if the image would be raveled.
 
     Parameters
     ----------
-    image : ndarray
-        The input image for which the maxima are to be calculated.
+    image_shape : tuple
+        The image for which the offsets are computed.
     selem : ndarray, optional
```
sciunto: Not optional

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,358 @@ def h_minima(image, h, selem=None):
...
         The neighborhood expressed as an n-D array of 1's and 0's.
         Default is the ball of radius 1 according to the maximum norm
```
sciunto: default has to be removed as well...

sciunto: the `selem` parameter in this signature `_offset_to_raveled_neighbours(image_shape, selem)` is not optional.

lagru: Ah, sorry. :see_no_evil:  I thought this was a comment in local_maxima's docstring.

sciunto: no problem!

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,360 @@ def h_minima(image, h, selem=None):
...
 
-    We can calculate all local maxima:
+    For each candidate:
 
-    >>> maxima = extrema.local_maxima(f)
+    2. Perform a flood-fill to find all connected pixels that have the same
+       gray value and are part of the plateau.
+    3. Consider the connected neighborhood of a plateau: if any no bordering
+       sample has a higher gray level, mark the plateau as a definite local
+       maximum.
```
jni: Nice notes! =)

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,360 @@ def h_minima(image, h, selem=None):
...
+    array([[0, 0, 0, 0, 0, 0, 0],
+           [0, 1, 1, 0, 0, 0, 0],
+           [0, 1, 1, 0, 0, 0, 0],
+           [1, 0, 0, 0, 0, 0, 1]], dtype=uint8)
+    >>> local_maxima(image, indices=True)
+    (array([1, 1, 2, 2, 3, 3]), array([1, 2, 1, 2, 0, 6]))
+
+    Find local maxima without comparing to diagonal pixels (connectivity 1):
+
+    >>> local_maxima(image, selem=1)
```
jni: This is inconsistent with our existing APIs. `selem` should be a neighbourhood, like `footprint` for `ndimage` morphological functions; for integer connectivity, use `connectivity=1`. You should keep `selem` in for backwards compatibility, but otherwise please add `connectivity` to use integer input.

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,360 @@ def h_minima(image, h, selem=None):
...
+           [0, 1, 1, 0, 1, 1, 0],
+           [1, 0, 0, 0, 0, 0, 1]], dtype=uint8)
+
+    and exclude maxima that border the image edge:
+
+    >>> local_maxima(image, selem=1, include_border=False)
+    array([[0, 0, 0, 0, 0, 0, 0],
+           [0, 1, 1, 0, 1, 1, 0],
+           [0, 1, 1, 0, 1, 1, 0],
+           [0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
```
jni: The current `peak_local_max` implementation uses the keyword `exclude_border`. Probably we should use that here also.

jni: An alternative suggestion is `allow_borders=True`.

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,360 @@ def h_minima(image, h, selem=None):
...
+        if any(s != 3 for s in selem.shape):
+            raise ValueError("dimension size in selem is not 3")
+
+    neighbor_offsets = _offset_to_raveled_neighbours(image.shape, selem)
+
+    # Array of flags used to store the state of each pixel during evaluation.
+    # Possible states are:
+    #   3 - first or last value in a dimension
+    #   2 - potentially part of a maximum
+    #   1 - evaluated and part of a maximum
```
jni: As mentioned, let's define some constants for these, and ensure that they are the same across the full stack of functions.

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,360 @@ def h_minima(image, h, selem=None):
...
+    flags = np.zeros(image.shape, dtype=np.uint8)
+    _set_edge_values_inplace(flags, value=3)
+
+    # Up-cast dtype of image for wrapped Cython function
+    if np.issubdtype(image.dtype, np.unsignedinteger):
+        dtype = np.uint64
+    elif np.issubdtype(image.dtype, np.integer):
+        dtype = np.int64
```
jni: Is there a specific reason why you didn't include all numeric types in the fused types? ie why should I need to multiply my volume by 8x to find maxima in a uint8 image? (Which is probably *the* most common use case for this function?)

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,360 @@ def h_minima(image, h, selem=None):
...
+    >>> local_maxima(image, indices=True)
+    (array([1, 1, 2, 2, 3, 3]), array([1, 2, 1, 2, 0, 6]))
+
+    Find local maxima without comparing to diagonal pixels (connectivity 1):
+
+    >>> local_maxima(image, selem=1)
+    array([[0, 0, 0, 0, 0, 0, 0],
+           [0, 1, 1, 0, 1, 1, 0],
+           [0, 1, 1, 0, 1, 1, 0],
+           [1, 0, 0, 0, 0, 0, 1]], dtype=uint8)
```
jni: Was the return type of the previous local maxima uint8 or bool? uint8 is unintuitive for me.

skimage/morphology/extrema.py
```diff
@@ -207,152 +210,360 @@ def h_minima(image, h, selem=None):
...
+    if image.size == 0:
+        # Return early for empty input
+        return np.array([], dtype=(np.intp if indices else np.uint8))
+
+    if include_border:
+        # Ensure that local maxima are always at least one smaller sample away
+        # from the image border
+        image = _fast_pad(image, image.min())
+
+    if selem is None:
```
jni: Can you abstract all this logic into a private function? It's quite cluttery. Something like `selem = _validate_selem(image, selem, connectivity)`

skimage/morphology/tests/test_extrema.py
```diff
@@ -281,39 +171,268 @@ def test_extrema_float(self):
...
+        """Test results if selem is a scalar."""
+        # Connectivity 1: generates cross shaped structuring element
+        result_conn1 = extrema.local_maxima(self.image, selem=1)
+        assert_equal(result_conn1, self.expected_cross)
+
+        # Connectivity 2: generates square shaped structuring element
+        result_conn2 = extrema.local_maxima(self.image, selem=2)
+        assert_equal(result_conn2, self.expected_default)
+
+        # Connectivity 3: generates disk shaped structuring element
```
jni: Wait, what?

lagru:
> I've left a few comments of varying severity. They shouldn't be too hard to address I hope! -- @jni 

Thanks for the extensive feedback; you've raised some important points. I'll address these concerns in the next few days. :slightly_smiling_face:

skimage/morphology/_extrema_cy.pyx
```diff
@@ -0,0 +1,164 @@
...
+        Py_ssize_t i, i_max, i_ahead
+        unsigned char prefilter
+
+    # Current flag meanings:
+    # 3 - first or last index in a dimension
+    # 2 - potentially part of a maximum
+    # 1 - not used in first loop
+    # 0 - not evaluated or none of the above is true
+
+    # Prefilter candidates only if neighbors in last dimension are part of
```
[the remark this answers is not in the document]
lagru: I'll add a more thorough description to the docstring of a new `_prefilter` function as you suggest further below.

skimage/morphology/_extrema_cy.pyx
```diff
@@ -0,0 +1,164 @@
...
+                if flags[i] == 2:
+                    # Index is potentially part of a maximum:
+                    # Find all samples part of the plateau and fill with 0
+                    # or 1 depending on whether it's a true maximum
+                    _fill_plateau(image, flags, neighbor_offsets, &queue, i)
+        finally:
+            queue_exit(&queue)
+
+
+cdef inline void _fill_plateau(
```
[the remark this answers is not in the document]
lagru: Thank you. :blush:

skimage/morphology/_restorable_queue.pxi
```diff
@@ -0,0 +1,119 @@
+"""FIFO queue whose items can be iterated multiple times.
+
+.. warning:: Expects "QueueItem" to be defined before inclusion!
```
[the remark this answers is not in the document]
lagru: I don't understand what you mean by this. This is emulates the same implementation structure as `Heap` in `heap_general.pxi` which needs `HeapItem` to be defined before inclusion. This is meant to clearly communicate this to the user.

skimage/morphology/_restorable_queue.pxi
```diff
@@ -0,0 +1,119 @@
...
+
+This queue can be operated like a class. The structure stores the state of the
+instance and the contained functions act as instance methods. The important
+distinction compared to normal queues is that popping an element doesn't remove
+it internally. Thus unless the queue's internal buffer is explicitly cleared
+(see `queue_clear`) popped items can be restored at anytime using
+`queue_restore`.
```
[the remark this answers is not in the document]
jni: Ah, ok. I might rename this a "QueueWithHistory"? The main purpose is to ensure that we have a "list" at the end of everything that was ever in the queue, so that as we proceed with floodfill, if we need to undo the whole thing, we can? To clarify, this is not to "backtrack" individual elements, right?

In this case, then (a) "QueueWithHistory" makes more sense to me, and (b) please add these notes explaining what you want to do with it. Yes, it's a good solution. =)

skimage/morphology/_restorable_queue.pxi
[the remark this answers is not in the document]
jni: Yes yes, I just mean, in Python, I'm used to every needed element to be present in the file either directly or via an import. Having something "assumed to be defined" is scary. If it's done like this elsewhere in the library, that's (mostly) fine. =) If I remember correctly, sklearn had a similar issue, which happens the C/Cython abstraction leaks over.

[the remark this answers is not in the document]
lagru: Yeah, this is entirely due to the underlying C inclusion architecture. The advantage of requiring the definition elsewhere is that the queue could be included in many places at once each using its own unique version of `QueueItem`.

skimage/morphology/_restorable_queue.pxi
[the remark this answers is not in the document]
lagru: Sounds good. I don't have a strong opinion about the current name. 

> The main purpose is to ensure that we have a "list" at the end of everything that was ever in the queue, so that as we proceed with floodfill, if we need to undo the whole thing, we can?

Correct!

> To clarify, this is not to "backtrack" individual elements, right?

No, although the insertion order is maintained and backtracking (if you mean iterating over the plateau values in reverse insertion order) would be possible with a few tweaks, I don't need to maintain order. Its just a consequence of the current implementation. 

I'll add the explanation in the header of the file.

skimage/morphology/_restorable_queue.pxi
```diff
@@ -0,0 +1,119 @@
...
+    self._buffer_ptr = <QueueItem*>malloc(buffer_size * sizeof(QueueItem))
+    if not self._buffer_ptr:
+        with gil:
+            raise MemoryError("couldn't allocate buffer")
+    self._buffer_size = buffer_size
+    self._index_consumed = -1
+    self._index_valid = -1
+
+
+cdef inline void queue_push(RestorableQueue* self, QueueItem* item_ptr) nogil:
```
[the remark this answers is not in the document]
lagru: I'm certainly no expert either but I'll voice my thoughts on this anyway. Please keep that in mind. :)

My understanding is that the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock) is meant to protect Python objects from being accessed multiple times at once from different threads. This would be dangerous because CPython's memory management is not thread-safe. 

However (concerning the current PR) the queue doesn't operate on objects accessible from Python. Furthermore `_local_maxima`'s queue is not shared between multiple threads, even if we would try to call this function multiple times in parallel from different Python threads (desirable because it releases the GIL and thus allows "true parallelism") each call would have its own queue only accessible within the algorithm and thus single thread. So I'd argue that this is safe as long as one ensures that access from multiple threads is currently not possible.

I remember a slightly related discussion about [Memoryviews and the GIL](https://github.com/scipy/scipy/pull/8541#discussion_r175613803) in one of my previous PRs. Might be useful to understand the issue.

skimage/morphology/extrema.py
lagru: I don't have a strong opinion on this. So I'll use whatever the majority recommends.

skimage/morphology/_restorable_queue.pxi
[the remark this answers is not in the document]
jni: Ah, if you can't access the same instance of a Queue from Python, then all is well. Thanks for pointing that out! =)

With memoryviews, my (again, naive) understanding is that it's ok as long as you are only reading (not writing) and you know that no one is going to delete the memory from under you.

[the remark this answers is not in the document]
lagru: > as long as you are only reading (not writing)

Well I'm certainly writing to the `flags` memoryview! But I still think this is safe because Cython's doc only mentions that the GIL is needed if

> [... the dtype is object and an object element is read or written.](http://docs.cython.org/en/latest/src/userguide/memoryviews.html#memoryviews-and-the-gil)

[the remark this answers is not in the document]
jni: :+1:

skimage/morphology/extrema.py
lagru: I just noticed: `peak_local_max`'s argument `exclude_border` isn't a boolean switch but an integer that denotes how much of the border is excluded. So using the same name might actually be more confusing than useful in this case. 

I'd go with `allow_borders`. It sounds different enough an keeps the current behavior. What do you think?

skimage/morphology/extrema.py
lagru: I could do that. But I thought this would increase the source of the generated binary significantly. If that is acceptable I'd be happy to change this because I agree that upcasting is not a very elegant solution to cover all dtypes.

skimage/morphology/extrema.py
jni: Yeah go for it. I do intend to deprecate `peak_local_max` eventually but they will coexist for a while...

skimage/morphology/extrema.py
lagru: We are talking about 1.4 MiB (all dtypes) vs 1.2 MiB (3 dtypes) here. Furthermore I just noticed that the dtype numpy.float16 has no equivalent numpy.float16_t in Cythons PXD file for NumPy. I wonder why that is. This might be a problem if we want to support that type.

jni: I'm not worried about the 200KB. A 1000x1000 uint8 image upcasting will already end up using an extra 7MB of RAM if we don't include all the types.

float16 is rare so I reckon it can be ignored/upcasted if needed.

skimage/morphology/watershed.py
```diff
@@ -120,19 +120,36 @@ def _validate_connectivity(image_dim, connectivity, offset):
     return c_connectivity, offset
 
 
-def _compute_neighbors(image, structure, offset):
-    """Compute neighborhood as an array of linear offsets into the image.
+def _offsets_to_raveled_neighbors(image_shape, selem, center):
```
lagru: @jni I took the liberty to merge the old and new function. I think that the new names are a little clearer and more documentation shouldn't hurt either.

skimage/morphology/extrema.py
```diff
@@ -207,152 +211,365 @@ def h_minima(image, h, selem=None):
...
+    return selem
 
-def local_minima(image, selem=None):
-    """Determine all local minima of the image.
 
-    The local minima are defined as connected sets of pixels with equal
-    grey level strictly smaller than the grey levels of all pixels in direct
-    neighborhood of the set.
+def local_maxima(image, connectivity=None, selem=None, indices=False,
+                 allow_borders=True):
```
jni: Even though I prefer `connectivity` as an arg, `selem` was already there in 0.13, so it needs to come first so that we don't break that interface.

lagru: I guess this also means, that we have to keep using `selem` instead of `structure` here.

jni: Sadly, yes. In the future I'd love to deprecate it but not for this PR. Thanks!

lagru:
@jni Thanks for the constructive and useful reviews.

I just noticed while updating the profiling notebook that between commits 5ca5a17 and the current state (14e164a) the function got noticeably slower. In some cases up to 30%! I'm currently trying to find the reason and offending commit...

lagru:
> I just noticed while updating the profiling notebook that between commits 5ca5a17 and the current state (14e164a) the function got noticeably slower. In some cases up to 30%! I'm currently trying to find the reason and offending commit...

I have difficulties finding the offending commit. It seems like declaring the flag values as module variables in 08da763 caused a noticable slow down which I don't understand. I'd have thought that the compiler would optimize that change away.
Furthermore several other commits (e.g. 4452775) might have introduced small slow-ups. However its hard to pinpoint these because I can no longer reproduce the qualitative results in the current notebook and I have no idea why. 

These two commits serve only the purpose to make the code more readable; are they worth the performance hit? 

The current implementation however is still magnitudes faster than the old version and still in the same neighborhood as `peak_local_max`. So it might be enough to leave it at the current state unless somebody has further insights.

jni:
@lagru I'm running benchmarks on every commit in that range using the machinery in #3137. Generally readability counts a lot in this library, and anyway I think there are probably Cython optimisations to make sure that those variable assignments don't actually cost anything. Once I figure out exactly what broke, I'll report back to see whether there are any possible fixes.

jni:
@lagru which benchmark specifically are you concerned about? For the random image (1000,1000) it's actually a bit faster after "MAINT: Merge _compute_neighbors and _offsets_to_raveled_neighbors". Here's the results:
```
· Running 10 total benchmarks (10 commits * 1 environments * 1 benchmarks)
[  0.00%] · For scikit-image commit hash e55bec90:
[  0.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1................................................................................
[  0.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 10.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                         26.3±0.1ms
[ 10.00%] · For scikit-image commit hash 260efb8b:
[ 10.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1..............................................................................
[ 10.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 20.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                         26.2±0.1ms
[ 20.00%] · For scikit-image commit hash f0f40308:
[ 20.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1...........................................................................
[ 20.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 30.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                         26.2±0.1ms
[ 30.00%] · For scikit-image commit hash b8b29ed4:
[ 30.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1.............................................................................
[ 30.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 40.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                        26.4±0.09ms
[ 40.00%] · For scikit-image commit hash 0f750342:
[ 40.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1............................................................................
[ 40.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 50.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                         26.7±0.2ms
[ 50.00%] · For scikit-image commit hash 509b41bd:
[ 50.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1...........................................................................
[ 50.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 60.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                        20.2±0.09ms
[ 60.00%] · For scikit-image commit hash 119c8d7b:
[ 60.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1...........................................................................
[ 60.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 70.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                        20.2±0.08ms
[ 70.00%] · For scikit-image commit hash fecd7db3:
[ 70.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1.........................................................................
[ 70.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 80.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                         20.0±0.2ms
[ 80.00%] · For scikit-image commit hash 0b746e94:
[ 80.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1.........................................................................
[ 80.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[ 90.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                        20.3±0.04ms
[ 90.00%] · For scikit-image commit hash d0b7c80e:
[ 90.00%] ·· Building for conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1...........................................................................
[ 90.00%] ·· Benchmarking conda-py3.6-cython0.28-matplotlib2.2-numpy1.14-scipy1.1
[100.00%] ··· Running benchmarks.SegmentationSuite.time_local_maxima                                         20.0±0.1ms
```

This is on my branch `jni/maxima2d-bench`, which required a rebase so the commits don't match, but they start at the equivalent of 5ca5a17.

jni:
ping @lagru

lagru:
@jni Sorry that I took so long and thank you for investigating yourself!

> @lagru which benchmark specifically are you concerned about? For the random image (1000,1000) it's actually a bit faster after "MAINT: Merge _compute_neighbors and _offsets_to_raveled_neighbors". Here's the results:
[code snippet]

When rerunning the notebook with the benchmarks I get a qualitative difference between the new `local_maxima` and `peak_local_max`. This difference was previously negligible but is now very noticeable: e.g. `local_maxima` got  slower compared to `peak_local_max` for the first 3 benchmarks.

BUT I can't even reproduce the original results of the notebook any longer. My only explanation is that something in the background or an update to a related library changed how the benchmarks perform.

Considering that weird behavior, I'd leave it as it is. If I solve this mystery I can always come bag and apply it.

jni:
Ok, I agree, this is not worth worrying about at this moment. @scikit-image/core anyone care to review/merge?

@ThomasWalter maybe you would like to review also?

ThomasWalter:
@jni : please go ahead, I did not really follow. (actually I did not know that I **can** review, as I do not belong to the core)

jni:
@ThomasWalter well now you know. =) 

@soupault @stefanv @emmanuelle an extra ping for good measure. ;) This is a big PR but the code is very clean after several revisions, so it should not be too tricky to review. Have a look!

Note: I intend to backport to 0.14 so let's keep the Py2 compatibility stuff and take it out in a later PR.

stefanv:
@lagru This is an exceptional piece of work; I wish all PRs were of this quality!

stefanv:
@meeseeksbox backport to v0.14.x

stefanv:

lagru:
@stefanv That is a really nice and certainly motivating thing to hear.

 And in return I want to mention that without your helpful and constructive feedback (special thanks to @jni) this wouldn't have been possible. Thank you.

Issue #3028, opened 2018-04-24
Image Inversion for signed types


ThomasWalter:
## Description
Image inversion for signed types does not yield the expected result. 

`In [7]: img = np.array([[0, 1, -125], [1, 0, 125]], dtype=np.int8)`

`In [8]: invert(img)`
`Out[8]: `
`array([[127, 126,  -4],
       [126, 127,   2]], dtype=int8)`

I guess, that image inversion should do different things depending on the data type. If the data type is signed, inversion should more or less multiply with (-1) (with the problem that the range is not symmetric). 

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [x] Code example
- [ ] Relevant images (if any)
- [ ] Operating system and version
- [ ] Python version
- [ ] scikit-image version (run `skimage.__version__`)

ThomasWalter:
So, in principle, it would be: 

- for unsigned integer: `img_inv = dtype_limits(image, clip_negative=False)[1] - image`
- for float: `img_inv = -image`
- for signed integer types: `img_inv = -1-image`

soupault:
This one is tricky. I think implementing different behaviours for different dtypes is reasonable, but it makes the function so counter-intuitive...

>for float: img_inv = -image

I'm afraid it's more appropriate to distinguish [-1:1] and [0:1] representations of float ndarrays, performing `i = -x` for the former and `i = 1-x` for the latter.

ThomasWalter:
True ... it is a bit ugly. 

> I'm afraid it's more appropriate to distinguish [-1:1] and [0:1] representations of float ndarrays, performing i = -x for the former and i = 1-x for the latter.

You are certainly right ... but how can I know the allowed range? 

I mean, I cannot infer the allowed range from the actual range. An image could live in the space of signed float images, yet - by coincidence - have only positive values.

soupault:
>but how can I know the allowed range?

Unfortunately, this is currently not possible (see https://github.com/scikit-image/scikit-image/issues/2605).

I don't know... Maybe we should disable this function for float ndarrays for now?

ThomasWalter:
Or we have an option for the inversion of float ndarrays.

soupault:
Closed via #3030.

PR #3030, opened 2018-04-26
Type dependent inversion

ThomasWalter:
## Description
Depending on the image type we would expect the behaviour of inversion to be quite different. This was discussed in issue #3028. Concretely:
- for unsigned integer: `img_inv = dtype_limits(image, clip_negative=False)[1] - image`
- for signed integer types: `img_inv = -1-image`
- for float: `img_inv = -image` if the allowed range is `[-1,1]` and `img_inv = 1-image` if the allowed range is `[0,1]`
As something like the "allowed range" does not exist in scikit-image, the user should decide for the float type, how the behaviour of the function should be.

## Checklist
- [x] Unit tests

## References
This concerns issue #3028. 

## For reviewers

  later.

### DIFF
--- a/skimage/util/_invert.py
+++ b/skimage/util/_invert.py
@@ -5,37 +5,59 @@ from .dtype import dtype_limits
 def invert(image, signed_float=False):
     """Invert an image.
 
-    Image inversion. If the image is of type unsignedinteger, the image
-    is subtracted from the maximum value allowed by the dtype maximum.
-    If the image is of type signedinteger, the image is subtracted from -1.
-    If the image is of float type and signed_float is true, the result is
-    -image. If the image is of float type and signed_float is false (default)
-    the result is 1.0 - image.
+    Invert the intensity range of the input image, so that the maximum value is
+    now the minimum for the input dtype, and vice-versa. This operation is
+    slightly different depending on the input dtype:
+
+    - unsigned integers: subtract the image from the dtype maximum
+    - signed integers: subtract the image from -1 (see Notes)
+    - floats: subtract the image from 1 (if signed_float is False, so we
+      assume the image is unsigned), or from 0 (if signed_float is True).
+
+    See the examples for clarification.
 
     Parameters
     ----------
     image : ndarray
-        The input image.
-    signed_float : bool
+        Input image.
+    signed_float : bool, optional
         If True and the image is of type float, the range is assumed to
         be [-1, 1]. If False and the image is of type float, the range is
-        assumed to be [0, 1]. Default value is False (i.e. the assumed range is
-        [0, 1].
+        assumed to be [0, 1].
 
     Returns
     -------
-    invert : ndarray
+    inverted : ndarray
         Inverted image.
+
+    Notes
+    -----
+    Ideally, for signed integers we would simply multiply by -1. However,
+    signed integer ranges are asymmetric. For example, for np.int8, the range
+    of possible values is [-128, 127], so that -128 * -1 equals -128! By
+    subtracting from -1, we correctly map the maximum dtype value to the
+    minimum.
 
     Examples
     --------
-    >>> img = np.array([[100, 0, 200],
-    ...                 [0,  50, 0],
-    ...                 [30,  0, 255]], np.uint8)
+    >>> img = np.array([[100,  0, 200],
+    ...                 [  0, 50,   0],
+    ...                 [ 30,  0, 255]], np.uint8)
     >>> invert(img)
     array([[155, 255,  55],
            [255, 205, 255],
            [225, 255,   0]], dtype=uint8)
+    >>> img2 = np.array([[ -2, 0, -128],
+    ...                  [127, 0,    5]], np.int8)
+    >>> invert(img2)
+    array([[   1,   -1,  127],
+           [-128,   -1,   -6]], dtype=int8)
+    >>> img3 = np.array([[ 0., 1., 0.5, 0.75]])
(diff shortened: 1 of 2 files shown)


skimage/util/_invert.py
```diff
@@ -2,15 +2,25 @@
...
+    If the image is of type signedinteger, the image is subtracted from -1.
+    If the image is of float type and signed_float is true, the result is
+    -image. If the image is of float type and signed_float is false (default)
+    the result is 1.0 - image.
 
     Parameters
     ----------
     image : ndarray
         The input image.
+    signed_float : bool
```
sciunto: please, write: bool, optional

skimage/util/tests/test_invert.py
```diff
@@ -30,20 +30,39 @@ def test_invert_uint8():
...
+    image[1, :] = lower_dtype_limit
+    image[2, :] = upper_dtype_limit
+    expected = np.zeros((3, 3), dtype=dtype)
+    expected[2, :] = lower_dtype_limit
+    expected[1, :] = upper_dtype_limit
     result = invert(image)
     assert_array_equal(expected, result)
 
 
 def test_invert_float64():
```
sciunto: Please, could you change the name to mention the range, here and below.

skimage/util/_invert.py
```diff
@@ -2,15 +2,25 @@
 from .dtype import dtype_limits
 
 
-def invert(image):
+def invert(image, signed_float=False):
     """Invert an image.
 
-    Substract the image to the maximum value allowed by the dtype maximum.
+    Image inversion. If the image is of type unsignedinteger, the image
```
soupault: I'd make this as short and informative as possible:
- remove `Image inversion. `. It is already mentioned above;
- use ` ``image`` ` instead of `the image`;
- true -> True, false -> False;
- make type description consistent: `type signedinteger`, `float type` -> `signedinteger type`, `float type` or v.v..

Moreover, I think, this might look better in the description of the returned value (below `invert : ndarray`).

skimage/util/_invert.py
```diff
@@ -2,15 +2,25 @@
...
+    is subtracted from the maximum value allowed by the dtype maximum.
```
soupault: `by the dtype maximum.` -> `by the dtype.` maybe?

skimage/util/_invert.py
```diff
@@ -28,6 +38,13 @@ def invert(image):
            [225, 255,   0]], dtype=uint8)
     """
     if image.dtype == 'bool':
-        return ~image
+        return(~image)
```
soupault: I don't think we use parentheses in the return statements anywhere across the library. Please, avoid using them if possible.

skimage/util/tests/test_invert.py
```diff
@@ -30,20 +30,39 @@ def test_invert_uint8():
...
+
+def test_invert_float64_b():
+    dtype = 'float64'
+    image = np.zeros((3, 3), dtype=dtype)
+    lower_dtype_limit, upper_dtype_limit = \
+        dtype_limits(image, clip_negative=True)
+    image[2, :] = upper_dtype_limit
+    expected = np.zeros((3, 3), dtype=dtype)
+    expected[0, :] = upper_dtype_limit
+    expected[1, :] = upper_dtype_limit
```
soupault: Maybe use `np.fliplr` or `np.fliud` to simplify the test code?

jni:
@soupault I don't think flipud/fliplr would make the code clearer. Other than that I've addressed your comments, I think. (I also tried to explain the different behaviour for different dtypes in the output section but ultimately it felt out of place there, so I put it back in the heading section.) I expanded the doctests to include other dtypes, since examples are a very good way to clarify for new users.

Since @sciunto has already approved, feel free to merge if you like my changes. =)

soupault:
@jni The PR is beautiful now :). Thanks everyone for the work!

jni:
Ach! It wasn't a typo:

```python
In [1]: np.int8(-128) * np.int8(-1)
/Users/jni/conda/envs/36/bin/ipython:1: RuntimeWarning: overflow encountered in byte_scalars
  #!/Users/jni/conda/envs/36/bin/python
Out[1]: -128
```

jni:
Perhaps wrap the numbers in the dtype inside the docstring, just like I did above, to remove ambiguity?

soupault:
>Ach! It wasn't a typo

Oops. I've made a push to `master` reverting this line.

>Perhaps wrap the numbers in the dtype inside the docstring, just like I did above, to remove ambiguity?

I think, this could be done for 1 dtype just to show the idea. Although we have lots of examples in the docstring already. I'm 50/50.

jni:
@soupault 

> I've made a push to `master` reverting this line.

Gutsy, but the right call. =)

> I'm 50/50.

Me too. Note that I didn't mean an example: I meant inside the text, change `-1 * -128` to `int8(-1) * int8(-128)`. But I'm also 50/50. Let's see what @ThomasWalter and maybe others think.

soupault:
>Note that I didn't mean an example

Ok, sorry, I see now. The sentence looks pretty clear to me anyway. :)

ThomasWalter:
Arghh, so sorry. 

So, basically, what I meant is that as (-128) * (-1) = 128 which is outside the allowed range for the image, it is clear that we can not just multiply the image with (-1). 

I think that `int8(-1) * int8(-128)` is the best solution. 

Sorry for the mess.

PR #3032, opened 2018-04-27
Fix SLIC very slow on Windows

cgohlke:
Fixes #3026

Tested on Windows 10, Python 3.6.5 and 3.7.0b3 64-bit, Cython 0.28.2, Visual Studio 14 Update 3 and 15.6.7.  Speed increase is 10x for the example in #3026.

PR #3035, opened 2018-04-28
Replace pow function calls in Cython modules to fix performance issues on Windows

### LINKED ISSUE #2657: High execution time of several examples
## Description
Should speedup the functions (especially, denoising) to facilitate the testing process.
```
Computation time summary (top 5):
	- plot_denoise.py : 1.2e+02 sec
	- plot_circular_elliptical_hough_transform.py : 48 sec
	- plot_rank_filters.py : 40 sec
	- plot_join_segmentations.py : 29 sec
	- plot_chan_vese.py : 14 sec
```

Also, from comment below:

```
skimage/future/graph/graph_cut.py::skimage.future.graph.graph_cut.cut_normalized 

The job exceeded the maximum time limit for jobs, and has been terminated.
```

cgohlke:
## Description
Fixes/improves several performance issues on Windows by replacing `x**2`  function calls with `x*x` statements in Cython extension modules.

Tested on Windows 10, Python 3.6.5 64-bit, Cython 0.28.2, Visual Studio 14 Update 3.

FWIW, some Cython modules could be improved by releasing the GIL, not accessing Python objects in inner loops, and consistent use of `Py_ssize_t` for sizes and indices.

## References
Closes #3032. 
Partly fixes #2657
Supersedes #3032

## For reviewers

  later.

### DIFF
--- a/skimage/feature/corner_cy.pyx
+++ b/skimage/feature/corner_cy.pyx
@@ -65,8 +65,8 @@ def _corner_moravec(image, Py_ssize_t window_size=1):
     cdef double[:, ::1] cimage = np.ascontiguousarray(img_as_float(image))
     cdef double[:, ::1] out = np.zeros(image.shape, dtype=np.double)
 
-    cdef double msum, min_msum
-    cdef Py_ssize_t r, c, br, bc, mr, mc, a, b, t
+    cdef double msum, min_msum, t
+    cdef Py_ssize_t r, c, br, bc, mr, mc, a, b
 
     with nogil:
         for r in range(2 * window_size, rows - 2 * window_size):
--- a/skimage/feature/corner_cy.pyx
+++ b/skimage/feature/corner_cy.pyx
@@ -66,7 +66,7 @@ def _corner_moravec(image, Py_ssize_t window_size=1):
     cdef double[:, ::1] out = np.zeros(image.shape, dtype=np.double)
 
     cdef double msum, min_msum
-    cdef Py_ssize_t r, c, br, bc, mr, mc, a, b
+    cdef Py_ssize_t r, c, br, bc, mr, mc, a, b, t
 
     with nogil:
         for r in range(2 * window_size, rows - 2 * window_size):
@@ -78,8 +78,8 @@ def _corner_moravec(image, Py_ssize_t window_size=1):
                             msum = 0
                             for mr in range(- window_size, window_size + 1):
                                 for mc in range(- window_size, window_size + 1):
-                                    msum += (cimage[r + mr, c + mc]
-                                             - cimage[br + mr, bc + mc]) ** 2
+                                    t = cimage[r + mr, c + mc] - cimage[br + mr, bc + mc]
+                                    msum += t * t
                             min_msum = min(msum, min_msum)
 
                 out[r, c] = min_msum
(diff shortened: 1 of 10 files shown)


skimage/feature/corner_cy.pyx
```diff
@@ -78,8 +78,8 @@ def _corner_moravec(image, Py_ssize_t window_size=1):
                             msum = 0
                             for mr in range(- window_size, window_size + 1):
                                 for mc in range(- window_size, window_size + 1):
-                                    msum += (cimage[r + mr, c + mc]
-                                             - cimage[br + mr, bc + mc]) ** 2
+                                    t = cimage[r + mr, c + mc] - cimage[br + mr, bc + mc]
```
JDWarner: Cython compilation is failing here, because `cimage` is a `double` while new variable `t` is `Py_ssize_t` in this file.

skimage/filters/rank/generic_cy.pyx
```diff
@@ -377,7 +377,7 @@ cdef inline void _kernel_otsu(dtype_t_out* out, Py_ssize_t odepth,
                               Py_ssize_t s0, Py_ssize_t s1) nogil:
     cdef Py_ssize_t i
     cdef Py_ssize_t max_i
-    cdef double P, mu1, mu2, q1, new_q1, sigma_b, max_sigma_b
+    cdef double P, mu1, mu2, q1, new_q1, sigma_b, max_sigma_b, t
```
JDWarner: This `t` also needs to be a `double`.

JDWarner: Whoops, my mistake.  I think it's only the prior one.

cgohlke: Hmm, I thought I fixed those. May have pushed a wrong branch...

cgohlke: Indeed, this is not what I was testing locally.

JDWarner:
I approve.  The failures are now unrelated, common to all current builds due to a difference in calls to a particular SciPy function.  Thanks, @cgohlke!

JDWarner:
Going ahead and merging this.

lagru:
> consistent use of Py_ssize_t for sizes and indices. -- @cgohlke

If I may hijack this for a slightly related question: Is `Py_ssize_t` preferred over `np.inpt_t`? The latter is used internally by NumPy's API and returned by things such as `ndarray.shape[0]`. Furthermore when initializing a new array within Cython (e.g. by calling `np.empty(..., dtype=...)` there is no Python equivalent to Py_ssize_t but `np.intp` exists... So I would prefer `np.intp_t` or are there more reasons to use `Py_ssize_t`?

cgohlke:
Re `Py_ssize_t` vs `ssize_t`: I prefer to use C `ssize_t` for sizes and indices in Cython code (to be consistent with C/C++) and use `numpy.intp_t` for numpy dtypes. Skimage seems to use `Py_ssize_t` throughout instead of `ssize_t`. Doesn't really matter.
In my comment I was referring to the use of C `int` and `long` types for sizes and indices. But that is not a performance issue so I left it out of this PR.

jni:
@lagru @cgohlke [Cython guidelines for NumPy](https://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html) suggest using `Py_ssize_t` for array sizes and indexing:

> Py_ssize_t is the proper C type for Python array indices.

It appears to be equivalent to `ssize_t` when available. From the [PEP](https://www.python.org/dev/peps/pep-0353/) introducing it:

> A new type Py_ssize_t is introduced, which has the same size as the compiler's size_t type, but is signed. It will be a typedef for ssize_t where available.

The Cython guide is the reference behind the use of `Py_ssize_t` in skimage. `ssize_t` seems like a reasonable alternative, but the PEP suggests that `Py_ssize_t` is strictly more general.

PR #3041, opened 2018-04-30
Optimize _probabilistic_hough_line function

cgohlke:
## Description

Release the GIL.
Use `ssize_t` for sizes and indices.
Replace weird `point` list with numpy arrays.
Store lines in a pre-allocated numpy array.

Tested on win-amd64-py3.7 only.

This change is not backwards compatible if seed is specified because of the use of `random.shuffle` instead of `random_state.randint`. One test needed to be adjusted.

Why does the `_probabilistic_hough_line` function return a "list of tuples of tuples" instead of a 3D numpy array?

## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_hough_transform.pyx
+++ b/skimage/transform/_hough_transform.pyx
@@ -7,6 +7,7 @@ import numpy as np
 cimport numpy as cnp
 cimport cython
 
+from cpython.mem cimport PyMem_Malloc, PyMem_Free
 from libc.stdlib cimport abs
 from libc.math cimport fabs, sqrt, ceil, atan2, M_PI
 
@@ -365,8 +366,11 @@ def _probabilistic_hough_line(cnp.ndarray img, Py_ssize_t threshold,
 
     # compute the bins and allocate the accumulator array
     cdef cnp.ndarray[ndim=2, dtype=cnp.uint8_t] mask = \
-         np.zeros((height, width), dtype=np.uint8)
-    cdef cnp.intp_t[:, ::1] line_end = np.zeros((2, 2), dtype=np.intp)
+        np.zeros((height, width), dtype=np.uint8)
+    cdef Py_ssize_t *line_end = \
+        <Py_ssize_t *>PyMem_Malloc(4 * sizeof(Py_ssize_t))
+    if not line_end:
+        raise MemoryError('could not allocate line_end')
     cdef Py_ssize_t max_distance, offset, num_indexes, index
     cdef double a, b
     cdef Py_ssize_t nidxs, i, j, k, x, y, px, py, accum_idx, max_theta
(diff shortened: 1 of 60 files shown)


skimage/transform/_hough_transform.pyx
```diff
@@ -97,8 +98,8 @@ def _hough_circle(cnp.ndarray img,
     return acc
 
 
-def _hough_ellipse(cnp.ndarray img, int threshold=4, double accuracy=1,
-                   int min_size=4, max_size=None):
+def _hough_ellipse(cnp.ndarray img, Py_ssize_t threshold=4, double accuracy=1,
+                   Py_ssize_t min_size=4, max_size=None):
```
jni: Given the use of `min_size` below, I think this should actually be a double? The use of `Py_ssize_t` for `threshold` is also dubious...? These do not refer to array indices or sizes...

cgohlke: `min_size` is compared to a C `double` but defined/documented as an integer, which might be intentional. It is a size/length, so I used `Py_ssize_t` for consistency but I'm OK with reverting to int.

`threshold` is compared to the max of a histogram count so `Py_ssize_t` should be correct.

stefanv:
@cgohlke Thanks for these changes; your changes look good, and I agree that `points` should have been an ndarray. What I don't quite understand in your PR is `lines` being changed from `list` to `array`.  It seems like lists are a good idea when building up arrays of unknown lengths?  Also, in this case, it looks like an array is returned in the end anyway.  I may well have missed something obvious.

mcs07:
Does `np.random.shuffle` return consistent results across platforms given the same seed? Also, does it need to be something like `random_state.shuffle` in order to respect the seed?

cgohlke:
Good catch. Of course it should be `random_state.shuffle`.

cgohlke:
Running the example multi-threaded on 6 cores is 12 times faster when using this PR:

```Python
from concurrent.futures import ThreadPoolExecutor
from multiprocessing import cpu_count

max_workers = cpu_count() // 2
print('max_workers =', max_workers)

def map_threaded(function, *iterables, max_workers=max_workers):
    with ThreadPoolExecutor(max_workers) as executor:
        return list(executor.map(function, *iterables))
    
def function(img):
    probabilistic_hough_line(img, threshold=10, line_length=5, line_gap=3, seed=1234)

%timeit map_threaded(function, [edges] * max_workers)
```

skimage/transform/_hough_transform.pyx
jni: Well, in my interpretation, `Py_ssize_t` is specifically for sizes of collections, arrays, etc., and corresponding indices, rather than any conceptual count. However I'm not super-fussed about this.

cgohlke: The maximum count in a histogram is the size of the input array. It's not any conceptual count.

soupault:
@cgohlke could you, please, rebase the branch onto `master`? Just for us to ensure that the tests are passing.

cgohlke:
Something went wrong. Closing.

PR #3055, opened 2018-05-05
WIP: do not merge, Feature expose precision loss logic

hmaarrfk:
## Description
I'm working to convert images between raw with bayer pattern and rgb or grey scale. It would be good to have this logic exposed to avoid unnecessary image copies.

This logic will be reused there to catch most cases.

In writing the tests for this, I found Issue #3080. This PR fixes Issue #3080.

## Checklist
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/util/dtype.py
+++ b/skimage/util/dtype.py
@@ -56,45 +56,38 @@ def _check_precision_loss(dtypeobj_in, dtypeobj_out):
         be considered lossy.
 
     """
-    dtypeobj_in = np.dtype(dtypeobj_in)
-    dtypeobj_out = np.dtype(dtypeobj_out)
-    kind_in = dtypeobj_in.kind
-    kind_out = dtypeobj_out.kind
-    itemsize_in = dtypeobj_in.itemsize
-    itemsize_out = dtypeobj_out.itemsize
+    def n_significant_bits(t):
+        dtypeobj = np.dtype(t)
+        kind = dtypeobj.kind
+        if kind == 'f':
+            return np.finfo(dtypeobj).nmant
+        elif kind == 'b':
+            return 1
+        elif kind == 'u':
+            # 8 bits per byte
+            return dtypeobj.itemsize * 8
+        elif kind == 'i':
+            # 1 bit used for sign
+            return dtypeobj.itemsize * 8 - 1
+        else:
+            raise ValueError('Unknown image dtype.')
 
-    has_loss = False
+    kind_in = np.dtype(dtypeobj_in).kind
+    kind_out = np.dtype(dtypeobj_out).kind
+    significant_bits_in = n_significant_bits(dtypeobj_in)
+    significant_bits_out = n_significant_bits(dtypeobj_out)
 
-    if kind_in != 'b' and kind_out == 'b':
+    if kind_in == 'f' and kind_out != 'f':
         has_loss = True
-    elif kind_in == 'f' and kind_out == 'f':
-        # float->float
-        if itemsize_out < itemsize_in:
-            has_loss = True
-    elif kind_in == 'f':
-        # float -> other (integer or bool)
+    elif significant_bits_in > significant_bits_out:
         has_loss = True
-    elif kind_out == 'f':
-        # signed/unsigned int -> float
-        if itemsize_in >= itemsize_out:
-            has_loss = True
-    elif ((kind_in == 'i' and kind_out == 'i') or
-            (kind_in == 'u' and kind_out == 'u')):
-        # signed/unsigned to same kind
-        if itemsize_out < itemsize_in:
-            has_loss = True
-    elif kind_in == 'u' and kind_out == 'i':
-        # usigned -> signed int
-        if itemsize_in >= itemsize_out:
-            has_loss = True
-    elif kind_in == 'i' and kind_out == 'u':
-        # signed int to unsigned
-        if itemsize_out <= itemsize_in:
-            has_loss = True
+    else:
+        has_loss = False
 
     if has_loss:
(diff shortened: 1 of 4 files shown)


hmaarrfk:
I also just added optional warnings.

I guess warnings are nice the first time you run your code, but after a while you realize that you are OK with the precision loss and therefore they shouldn't add to visual noise if you don't need them.

stefanv:
Decision: refactor dtype range checking; will not include warning handling.

hmaarrfk:
@scikit-image/core bump

skimage/util/dtype.py
```diff
@@ -32,7 +32,73 @@
...
+    dtypeobj_in = np.dtype(dtypeobj_in)
+    dtypeobj_out = np.dtype(dtypeobj_out)
+    kind_in = dtypeobj_in.kind
+    kind_out = dtypeobj_out.kind
+    itemsize_in = dtypeobj_in.itemsize
+    itemsize_out = dtypeobj_out.itemsize
+
+    has_loss = False
+
+    if kind_in != 'b' and kind_out == 'b':
```
stefanv: There seems to be some more generic pattern here.

Can we check only the itemsize_in vs itemsize_out?

stefanv: plus minor logic for signed ranges

hmaarrfk: seems like i can just treat it as 'u' and do something like `kind in ['b', 'u']`

hmaarrfk: Seems like I cannot.
```
In [1]: import numpy as np

In [2]: np.dtype(np.uint8).itemsize
Out[2]: 1
```

stefanv: So `bool` is the exception that needs to be checked for, because it is atypically sized.

hmaarrfk:
@stefanv commenting here so this discussion doesn't get lost

right, so that exception checking is done in the line you pointed to right?

the last 3 if statements do the boundary checking for the different int conversions. A uint8->int8 actually has loss, so that is why I use the appropriate comparison operator.

I think the float logic was written assuming 32 bit floats. In which case, they are more expressive than the other types we support. Can I access things like the number of bits in the mantisa? that is the correct thing to compare.


The logic for the

hmaarrfk:
I think I have an idea. Let me think about it later tonight.

stefanv:
Yes!  That looks great.

hmaarrfk:
merger: please squash merge

hmaarrfk:
@stefanv I think I was off by one in the last version you saw:

the current version uses:
```
 59     def n_significant_bits(t):                                                  
 60         dtypeobj = np.dtype(t)                                                  
 61         kind = dtypeobj.kind                                                    
 62         if kind == 'f':                                                         
 63             # IEEE floating point point has an implied 1, increasing the number 
 64             # of significant bits given a mantisa of a certain size                                                                                                      
 65             return np.finfo(dtypeobj).nmant + 1
```
which i think is more correct. though, if you are relying on 1 very small bit, maybe your algorithm is too sensitive.

skimage/util/dtype.py
```diff
@@ -32,7 +32,68 @@
...
+        float -> signed/unsigned ints will always return true.
+        signed/unsigned -> float of same number of bytes will return true.
+
+    Parameters
+    ----------
+    dtypeobj_in: np.dtype
+        dtype of the input image.
+    dtypeobj_out: np.dtype
+        dtype of the output image.
+    output_warning: bool, optional
```
sciunto: This and the option below do not appear in the signature

skimage/util/dtype.py
```diff
@@ -32,7 +32,68 @@
...
+        has_loss = False
+
+    if has_loss:
+        warn("Possible precision loss when converting from {} to {}"
+             .format(dtypeobj_in, dtypeobj_out))
+
+    return has_loss
+
+
+def dtype_limits(image, clip_negative=None):
```
sciunto: This is an API change, isn't it?

.gitignore
```diff
@@ -32,3 +32,6 @@ doc/release/_release_notes_for_docs.txt
 doc/gh-pages
 wheels
 skimage/morphology/_skeletonize_3d_cy.pyx
+
+.ropeproject
```
sciunto: What's that?

skimage/util/dtype.py
sciunto: No I read it changed from ` clip_negative=False)` to ` clip_negative=None)` The original code is `=False` as far as I can tell.

hmaarrfk:
@sciunto i guess this one wont make 0.14 easily

https://github.com/scikit-image/scikit-image/blob/v0.14.x/skimage/util/dtype.py#L36

sciunto:
yes, we changed the API because the former default value was dangerous (and if I remember well, caused a bug in the library).

sciunto:
In test_downcast(), the warning differs from the one expected in the test.

hmaarrfk:
@sciunto the `_scale` function is scary. I didn't realize it did [**magic**](https://github.com/scikit-image/scikit-image/blob/35ef2171e0da60e5d8f68c6c29b8e183a5c35a19/skimage/util/dtype.py#L176).

Although I really think this kind of refactoring pushes things in the right direction, as it stands, it isn't actually issuing the right error messages `img_as` is doing too much magic.

Lets wait until: https://github.com/scikit-image/scikit-image/issues/3373
this is resolved. Maybe after that one, I can make an audit of `img_as` for the int types.

hmaarrfk:
I guess this is no longer necessary.

Issue #1737, opened 2015-10-03
Consider not converting images when using `img_as_float` on a float32 image


jni:
See e.g. #1627

stefanv:
Yes, `img_as_float` should not convert the range of images that are already floating point.  We may rely on the type being float64 in some Cython code, so we'll probably still have to change the dtype.

stefanv:
Given that the underlying Cython code requires doubles, what do you say?  Should we leave ``img_as_float`` as is?

jni:
@stefanv One word: fused types. =P

stefanv:
That seems like a perfectly reasonable suggestion, although it would require changing *all* Cython functions that expect doubles.

taion:
It would be great if `img_as_float` didn't generally convert float32 to float64. We're seeing that when running something like `warp` on a `float32` image, a significant fraction of the time is spent unnecessarily converting the image into `float64`, which does little to really improve the output quality.

jni:
Pinging this, as I was reminded of it while writing #3052.

jni:
We should have some prioritisation tags I think... Is there a reason we don't have them?

Anyway I would call this high priority.

jni:
Well, the problem is then you end up switching a massive number of `0.14.0optional` tags to `0.15.0optional` tags...

soupault:
@jni This is why we don't use such notation, and all of our current milestones are the `must` ones. :trollface:

stefanv:
Also addressed by #3110.  Also note https://github.com/scikit-image/scikit-image/issues/3111

jni:
Closed by #3110

Issue #1234, opened 2014-11-25
Rename conversion functions rescale_to_*


jni:
Many users have been confused by our image conversion functions before. Renaming them to rescale_to_{ubyte,uint,int,float} will more accurately reflect what they do. See #1229 for discussion.

sciunto:
@scikit-image/core This is a ~ 5 years old proposal. Do we still agree that we want these rescale_to_* functions in scikit-image?

jni:
@sciunto thanks for the ping! imho this should remain on our radar. It's just not a high priority because things are working fine...

sciunto:
Thanks! I plan it for 1.0.

alexandrejaguar:
Hey everyone,
I really like the idea, but thinking on the "beginner", I was wondering if `rescale_to...` wouldn't be confused with the `rescale` operation (ours on `skimage.transform.rescale`).
How about `convert_to_*`?

hmaarrfk:
I seriously wonder what the root cause of the confusion is.

Most people have trouble understand why the scales are sometimes between 255 and sometimes between 0 and 1.

People generally are unformtable with image saving formats and scientific precision. Not sure if renaming stuff is the best way to reach that.

alexandrejaguar:
@hmaarrfk I get your point, but being explicit in this case would be part of the solution. On the other hand, we could put an extended section on our docs explaining how to deal with image formats/precision; that could help solving the other issue you exposed.

sciunto:
We have this: https://scikit-image.org/docs/stable/user_guide/data_types.html

alexandrejaguar:
@sciunto yes, I remember this one. I was thinking on maybe extending it with some bells and whistles :)

mkcor:
> Hey everyone,
> I really like the idea, but thinking on the "beginner", I was wondering if `rescale_to...` wouldn't be confused with the `rescale` operation (ours on `skimage.transform.rescale`).
> How about `convert_to_*`?

I changed my mind and prefer `rescale_`; I think that `convert_` refers to conversions between colour spaces.

lagru:
We, @stefanv and @matthew-brett, had an extensive discussion about this yesterday. I'll attempt to summarize my understanding of it.

[...]

PR #3062, opened 2018-05-07
Make `img_as_float` act as `ensure_img_is_float` of any precision

hmaarrfk:
## Description
Currently if you provide a `float32` image, it would get converted to `float64` by almost any function.
This is because `img_as_float` is an alias for `img_as_float64`.

I would like `img_as_float` to leave `float32` (or `float16`) to whatever precision they have.
It is only important if you are providing integer images (and in that case, you shouldn't be going from `uint8->float64`, that is a little excessive, but maybe not doing that is "premature optimization")

Image as float now takes an additional parameter that allows the user to specify the desired precision if the image is not already a floating point.

The function should probably be called `ensure_img_is_float`, but I figured many people were using this function as a "make my image float".

If this strategy is accepted, then PR #3060 is not necessary.
I would also have to add some tests.

## Checklist
- [x] Unit tests
- [ ] If #3055  is accepted, then we also need to allow img_as_float to take the issue warning parameter.

## References
xref: #1234, #1737
PR #3060 

## For reviewers

  later.

### DIFF
--- a/skimage/util/dtype.py
+++ b/skimage/util/dtype.py
@@ -370,7 +370,36 @@ def img_as_float64(image, force_copy=False):
     return convert(image, np.float64, force_copy)
 
 
-img_as_float = img_as_float64
+def img_as_float(image, force_copy=False, default_dtype=np.float64):
+    """Ensure that an image is of floating point type.
+
+    First checks if the image is a floating point (any precision). If not
+    it will convert it to the desired precision (by default, 64-bit).
+
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


soupault:
xref: https://github.com/scikit-image/scikit-image/issues/1234, https://github.com/scikit-image/scikit-image/issues/1737

stefanv:
Superseded by https://github.com/scikit-image/scikit-image/pull/3110

Thanks for your work on this @hmaarrfk!

PR #3067, opened 2018-05-09
transform: fix filter bias

kczimm:
## Description
There is a bias issue in the reconstruction images produced with function `iradon` because of the kernel implementation. This issue is fixed using the ramp filter equation in the spatial domain found in the referenced material.

This is a continuation of the work done by @ASchuberth.

## Checklist
- [ ] Unit tests

## References
Addresses issues: #2895, #2853 
Implements ramp filter equation from Kak and Slaney p.72 (https://engineering.purdue.edu/~malcolm/pct/CTI_Ch03.pdf)

## For reviewers

  later.


kczimm:
You can see that the reconstruction is unbiased in the uniform region of the phantom now.

unbiased

hmaarrfk:
@kczimm Can you rebase onto Master, that will get your builds to pass.

kczimm:
Rebase done. Thanks @hmaarrfk

skimage/transform/radon_transform.py
```diff
@@ -216,20 +216,31 @@ def iradon(radon_image, theta=None, output_size=None,
...
-    omega = 2 * np.pi * f                                # angular frequency
-    fourier_filter = 2 * np.abs(f)                       # ramp filter
+    n1 = np.arange(0, projection_size_padded / 2 + 1, dtype=np.int)
+    n2 = np.arange(projection_size_padded / 2 - 1, 0, -1, dtype=np.int)
+    n = np.concatenate((n1, n2))
+    f = np.zeros(projection_size_padded)
+    f[0] = 0.25
+    f[1::2] = -1 / (np.pi * n[1::2])**2
+
+    omega = 2 * np.pi * fftfreq(projection_size_padded).reshape(-1, 1)
```
hmaarrfk: it seems you you using reshape and slicing on a 1D array so assign a 2D array. is it true?

kczimm:
The code and before screenshot is in #2895. I can copy here if required for completeness.

hmaarrfk:
No it is fine. my mistake.

hmaarrfk:
It seems that the [previous discussion](https://github.com/scikit-image/scikit-image/pull/2895#issuecomment-366070852) indicated that they really wanted a test to help catch the underlying problem. Can you work on that?

kczimm:
Yes, I will work on that.

skimage/transform/radon_transform.py
kczimm: I was just adding a singleton dimension because a 2D np.array is required later but I just cleaned it up in a new commit.

kczimm:
The test passes on my system. Assuming it passes PEP8 and CI, are there any other suggestions?

hmaarrfk:
Great! And I think they wanted a test to ensure that the filters have 2 peaks. Quoting @stefanv 

> Can we add a test to ensure that this doesn't break again? E.g., generate an impulse response of the filter, and verify that Shepp-Logan has two peaks, or peaks at the expected locations, etc.

kczimm:
I agree that having tests on the filters themselves is a good idea. Perhaps I should not have extended the previous PR because I believe we were solving slightly different issues. The problem I solved here is the ramp filter bias. The previous issues were with the filter windows. 

I propose that we make a separate issue/PR moving the `iradon` kernel creation to a separate function (e.g. `iradon_filter(filter_name: str)` so that testing/validation of filter creation can be easily included. Right now, iradon creates the filters from within and so I don't know of a good way to test the characteristics of the filters without assessing the image quality of `iradon`.

Thoughts, @hmaarrfk, @stefanv ?

hmaarrfk:
@kczimm that is a very good point.

stefanv:
Sounds like a good plan to me.

philipstarkey:
I've just been bitten by this issue when trying to port some code from MATLAB to Python.

Is there any chance this could be merged in soon? It seems like everything is done... If not, I'm happy to try and help out to get this over the line.

skimage/transform/radon_transform.py
```diff
@@ -216,26 +216,37 @@ def iradon(radon_image, theta=None, output_size=None,
     img = np.pad(radon_image, pad_width, mode='constant', constant_values=0)
 
     # Construct the Fourier filter
-    f = fftfreq(projection_size_padded).reshape(-1, 1)   # digital frequency
-    omega = 2 * np.pi * f                                # angular frequency
-    fourier_filter = 2 * np.abs(f)                       # ramp filter
+    n1 = np.arange(0, projection_size_padded / 2 + 1, dtype=np.int)
```
hmaarrfk: Lets use floor divide here
```suggestion
    n1 = np.arange(0, projection_size_padded // 2 + 1, dtype=np.int)
```

skimage/transform/radon_transform.py
```diff
@@ -216,26 +216,37 @@ def iradon(radon_image, theta=None, output_size=None,
...
+    n2 = np.arange(projection_size_padded / 2 - 1, 0, -1, dtype=np.int)
```
hmaarrfk: Floor divide here as well.
```suggestion
    n2 = np.arange(projection_size_padded // 2 - 1, 0, -1, dtype=np.int)
```

skimage/transform/radon_transform.py
```diff
@@ -216,26 +216,37 @@ def iradon(radon_image, theta=None, output_size=None,
...
+    f[1::2] = -1 / (np.pi * n[1::2])**2
+
+    omega = 2 * np.pi * fftfreq(projection_size_padded)
+    fourier_filter = 2 * np.real(fft(f))         # ramp filter
     if filter == "ramp":
         pass
     elif filter == "shepp-logan":
         # Start from first element to avoid divide by zero
-        fourier_filter[1:] = fourier_filter[1:] * np.sin(omega[1:]) / omega[1:]
+        fourier_filter[1:] *= np.sin(omega[1:] / 2) / (omega[1:] / 2)
```
hmaarrfk: Why did we need to divide by 2 here again?

skimage/transform/radon_transform.py
hmaarrfk: Seems like my suggestion doesn't give the same results
image

skimage/transform/radon_transform.py
hmaarrfk: image
Left is the original fourier_filter at the endof this line, Right is the one you are proposing. Pretty drastically different. is this something you meant to change?

philipstarkey: @hmaarrfk I believe that change was one of the main points of the PR. See #2895

skimage/transform/radon_transform.py
kczimm: projection_size_padded will always be even.

kczimm:
This PR is pretty complete. I just looked it over for the first time in nearly 10 months. It fixes a major issue with the function that prevents anyone from using this function for real analysis. Let's streamline this process and get this merged so the community can rely on scikit-image for quantitative reconstructions.

skimage/transform/radon_transform.py
hmaarrfk: Alright, good point, I would still prefer if we releaved our intentions on making this an int. but whatever here.

skimage/transform/radon_transform.py
hmaarrfk: @philipstarkey @kczimm you are both right..... yeah, looking at this PR again after so long introduces makes merging more difficult. It seems there is a small artifact at point 33. I think it is from the way we are suggesting to compute the ramp filter.

image

hmaarrfk:
```suggestion
        fourier_filter = np.abs(np.sin(omega / 2)) / np.pi * 2
```

skimage/transform/radon_transform.py
```diff
@@ -216,26 +216,37 @@ def iradon(radon_image, theta=None, output_size=None,
...
+    f = np.zeros(projection_size_padded)
+    f[0] = 0.25
+    f[1::2] = -1 / (np.pi * n[1::2])**2
+
+    omega = 2 * np.pi * fftfreq(projection_size_padded)
+    fourier_filter = 2 * np.real(fft(f))         # ramp filter
     if filter == "ramp":
         pass
     elif filter == "shepp-logan":
         # Start from first element to avoid divide by zero
```
hmaarrfk:
```suggestion

```

skimage/transform/radon_transform.py
hmaarrfk: This is probably the desired artifact...

kczimm: what parameters are you using to create this filter? I would like to reproduce it.

skimage/transform/radon_transform.py
kczimm: Sure, actually we can remove the `dtype=np.int`. They are going to be promoted to floats later anyway. It was an unnecessary conversion on my part.

skimage/transform/radon_transform.py
```diff
@@ -216,26 +216,37 @@ def iradon(radon_image, theta=None, output_size=None,
...
-    fourier_filter = 2 * np.abs(f)                       # ramp filter
+    n1 = np.arange(0, projection_size_padded / 2 + 1, dtype=np.int)
+    n2 = np.arange(projection_size_padded / 2 - 1, 0, -1, dtype=np.int)
+    n = np.concatenate((n1, n2))
+    f = np.zeros(projection_size_padded)
+    f[0] = 0.25
+    f[1::2] = -1 / (np.pi * n[1::2])**2
+
+    omega = 2 * np.pi * fftfreq(projection_size_padded)
+    fourier_filter = 2 * np.real(fft(f))         # ramp filter
```
hmaarrfk:
```suggestion
    # Computing the ramp filter from the fourier transform of is frequency domain representation
    # lessens artifacts and removes a small bias as explained in [1]
    fourier_filter = 2 * np.real(fft(f))         # ramp filter
```

hmaarrfk:
Anyway, most things are nits. I would like to see a small comment added about your computation method. Other than that, it is off to somebody else to merge.

skimage/transform/radon_transform.py
hmaarrfk: Yeah, I just realized that this is such a small point and there are 1 million ways to compute the same thing, just leave it.

I guess I got confused looking at this again after so long since there were 2 key, and very important changes here.

hmaarrfk:
I added my little comment, and retriggered the builds. Hopefully this should go through without delay. Now off to releasing scikit-image.

hmaarrfk:
@scikit-image/core  if anybody has the chance to review this it would be greatly appreciated. It looks like a really good PR that has taken too long to get mainlined.

stefanv:
@hmaarrfk I like your recommended comment change where you explain why this is a viable way of calculating the filter.  Why don't you add that, and then we merge?

@kczimm Thank you for this important change, and sorry that it waited for so long.

hmaarrfk:
Could be a bug on github, or just lack permissions from @kczimm I'll try again later and then create a new PR if needed

image

kczimm:
Hmm. How do I change the permissions?

kczimm:
@hmaarrfk I added you as a collaborator to the repository. I'm not sure if that will do it. Let me know.

hmaarrfk:
Well that worked, but typically there is a little checkbox next to the top of the PR, next to the labels and stuff.

It allows maintainers to push to your branch if you allow it. We typically use that sparingly since it can cause conflicts, which is especially troublesome to those new to git.

hmaarrfk:
Cool. Thanks for getting this in kczimm.

I'm going to merge this by taking Stefan's comment as the second approval. 

This should be in 0.16.0 but we should consider backporting this too.

hmaarrfk:
Whoa, Stefan beat me to the clock.

stefanv:
:zap:

kczimm:
Whoa. Feels good to be a contributor.

stefanv:
We are very appreciative of your efforts; thank you!

jni:
@hmaarrfk as far as I can tell (I've run into this problem myself), push rights don't extend to the web interface, for some weird reason.

PR #3069, opened 2018-05-10
Feature rectangle perimeter

hmaarrfk:
## Description
These are useful!

## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/draw/draw.py
+++ b/skimage/draw/draw.py
@@ -793,6 +793,34 @@ def rectangle_perimeter(start, end=None, extent=None, shape=None, clip=False):
     -------
     coords : array of int, shape (Ndim, Npoints)
         The coordinates of all pixels in the rectangle.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from skimage.draw import rectangle_perimeter
+    >>> img = np.zeros((5, 5), dtype=np.uint8)
+    >>> start = (2, 2)
+    >>> end = (3, 3)
+    >>> rr, cc = rectangle_perimeter(start, end=end, shape=img.shape)
+    >>> img[rr, cc] = 1
+    >>> img
+    array([[0, 0, 0, 0, 0],
+           [0, 1, 1, 1, 0],
+           [0, 1, 0, 1, 0],
+           [0, 1, 1, 1, 0],
+           [0, 0, 0, 0, 0]], dtype=uint8)
+
+    >>> img = np.zeros((5, 5), dtype=np.uint8)
+    >>> rr, cc = rectangle_perimeter(start, end=(10, 10), shape=img.shape,
+                                     clip=True)
+    >>> img[rr, cc] = 1
+    >>> img
+    array([[0, 0, 0, 0, 0],
+           [0, 1, 1, 1, 1],
+           [0, 1, 0, 0, 1],
+           [0, 1, 0, 0, 1],
+           [0, 1, 1, 1, 1]], dtype=uint8)
+
     """
 
     if extent is not None:
--- a/skimage/draw/draw.py
+++ b/skimage/draw/draw.py
@@ -798,17 +798,17 @@ def rectangle_perimeter(start, end=None, extent=None, shape=None, clip=False):
     --------
     >>> import numpy as np
     >>> from skimage.draw import rectangle_perimeter
-    >>> img = np.zeros((5, 5), dtype=np.uint8)
-    >>> start = (2, 2)
-    >>> end = (3, 3)
+    >>> img = np.zeros((5, 6), dtype=np.uint8)
+    >>> start = (2, 3)
+    >>> end = (3, 4)
     >>> rr, cc = rectangle_perimeter(start, end=end, shape=img.shape)
     >>> img[rr, cc] = 1
     >>> img
-    array([[0, 0, 0, 0, 0],
-           [0, 1, 1, 1, 0],
-           [0, 1, 0, 1, 0],
-           [0, 1, 1, 1, 0],
-           [0, 0, 0, 0, 0]], dtype=uint8)
+    array([[0, 0, 0, 0, 0, 0],
+           [0, 0, 1, 1, 1, 0],
+           [0, 0, 1, 0, 1, 0],
+           [0, 0, 1, 1, 1, 0],
+           [0, 0, 0, 0, 0, 0]], dtype=uint8)
 
     >>> img = np.zeros((5, 5), dtype=np.uint8)
     >>> rr, cc = rectangle_perimeter(start, end=(10, 10), shape=img.shape,
@@ -829,6 +829,6 @@ def rectangle_perimeter(start, end=None, extent=None, shape=None, clip=False):
         raise ValueError("Either `end` or `extent` must be given")
 
     start = (start[0] - 1, start[1] - 1)
-    r = [start[1], end[1], end[1], start[1], start[1]]
-    c = [start[0], start[0], end[0], end[0], start[0]]
+    r = [start[0], start[0], end[0], end[0], start[0]]
+    c = [start[1], end[1], end[1], start[1], start[1]]
     return polygon_perimeter(r, c, shape=shape, clip=clip)
--- a/skimage/draw/draw.py
+++ b/skimage/draw/draw.py
@@ -612,7 +612,7 @@ def ellipse_perimeter(r, c, r_radius, c_radius, orientation=0, shape=None):
            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
 
 
-    Note that the positions of `ellipse` without specified `shape` can have 
+    Note that the positions of `ellipse` without specified `shape` can have
     also, negative values, as this is correct on the plane. On the other hand
     using these ellipse positions for an image afterwards may lead to appearing
     on the other side of image, because ``image[-1, -1] = image[end-1, end-1]``
@@ -764,3 +764,43 @@ def rectangle(start, end=None, extent=None, shape=None):
     coords = np.meshgrid(*[np.arange(st, en) for st, en in zip(tuple(tl),
                                                                tuple(br))])
     return coords
+
+
+def rectangle_perimeter(start, end=None, extent=None, shape=None, clip=False):
+    """Generate coordinates of pixels that are exactly around a rectangle.
+
+    Parameters
+    ----------
+    start : tuple
+        Origin point of the inner rectangle, e.g., ``(row, column)``.
+    end : tuple
+        End point of the inner rectangle ``(row, column)``.
+        Either `end` or `extent` must be specified.
+    extent : tuple
+        The extent (size) of the inner rectangle.  E.g.,
+        ``(num_rows, num_cols)``.
+        Either `end` or `extent` must be specified.
+    shape : tuple, optional
+        Image shape used to determine the maximum bounds of the output
+        coordinates. This is useful for clipping perimeters that exceed
+        the image size. By default, no clipping is done.
+    clip : bool, optional
+        Whether to clip the perimeter to the provided shape. If this is set
+        to True, the drawn figure will always be a closed polygon with all
+        edges visible.
+
+    Returns
+    -------
+    coords : array of int, shape (Ndim, Npoints)
+        The coordinates of all pixels in the rectangle.
+    """
+
+    if extent is not None:
+        end = np.array(start) + np.array(extent)
+    elif end is None:
+        raise ValueError("Either `end` or `extent` must be given")
+
+    start = (start[0] - 1, start[1] - 1)
+    r = [start[1], end[1], end[1], start[1], start[1]]
+    c = [start[0], start[0], end[0], end[0], start[0]]
+    return polygon_perimeter(r, c, shape=shape, clip=clip)
--- a/skimage/draw/tests/test_draw.py
+++ b/skimage/draw/tests/test_draw.py
@@ -7,7 +7,8 @@ from skimage._shared.testing import assert_almost_equal
 from skimage.draw import (set_color, line, line_aa, polygon, polygon_perimeter,
                           circle, circle_perimeter, circle_perimeter_aa,
                           ellipse, ellipse_perimeter,
-                          _bezier_segment, bezier_curve, rectangle)
+                          _bezier_segment, bezier_curve, rectangle,
+                          rectangle_perimeter)
 from skimage.measure import regionprops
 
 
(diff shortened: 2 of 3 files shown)


skimage/draw/draw.py
```diff
@@ -764,3 +764,71 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+    """
+
+    if extent is not None:
+        end = np.array(start) + np.array(extent)
+    elif end is None:
+        raise ValueError("Either `end` or `extent` must be given")
+
+    start = (start[0] - 1, start[1] - 1)
+    r = [start[0], start[0], end[0], end[0], start[0]]
+    c = [start[1], end[1], end[1], start[1], start[1]]
```
jni: I presume this is to draw bounding boxes? Why are the start coordinates offset but not the end coordinates?

skimage/draw/draw.py
```diff
@@ -764,3 +764,71 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+           [0, 0, 1, 1, 1, 0],
+           [0, 0, 0, 0, 0, 0]], dtype=uint8)
+
+    >>> img = np.zeros((5, 5), dtype=np.uint8)
+    >>> rr, cc = rectangle_perimeter(start, end=(10, 10), shape=img.shape,
+                                     clip=True)
+    >>> img[rr, cc] = 1
+    >>> img
+    array([[0, 0, 0, 0, 0],
+           [0, 1, 1, 1, 1],
```
soupault: From the above, `start = (2, 3)`. This array doesn't seem correct.

skimage/draw/draw.py
```diff
@@ -764,3 +764,71 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+           [0, 1, 1, 1, 1]], dtype=uint8)
+
+    """
+
+    if extent is not None:
+        end = np.array(start) + np.array(extent)
+    elif end is None:
+        raise ValueError("Either `end` or `extent` must be given")
+
+    start = (start[0] - 1, start[1] - 1)
```
soupault: Why do you subtract 1 here? I'd expect this function to use the coordinates "as is".

hmaarrfk: I want it to draw a box AROUND the pixels that would be filled in by the `rectangle function`

skimage/draw/draw.py
hmaarrfk: I think I was confused by the different features of end in rectangle.
I'm updating it now to include  more logic from the rectangle function itself.

I was only using extent.

hmaarrfk:
Looking at the other functions, it seemed like the perimeter was always drawn AROUND the object itself.
See notebook [example](https://gist.github.com/hmaarrfk/0c096a2ed2aaaf38b8bfccfe6300cd82)

skimage/draw/draw.py
hmaarrfk: Thanks, I reran the code and fixed it.

skimage/draw/draw.py
```diff
@@ -764,3 +764,76 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+    array([[0, 0, 0, 0, 0],
+           [0, 0, 1, 1, 1],
+           [0, 0, 1, 0, 1],
+           [0, 0, 1, 0, 1],
+           [0, 0, 1, 1, 1]], dtype=uint8)
+
+    """
+
+    if extent is not None:
+        end = np.array(start) + np.array(extent)
```
soupault: Rather `np.asarray` here.

skimage/draw/tests/test_draw.py
```diff
@@ -891,3 +892,102 @@ def test_rectangle_extent():
...
+    assert_array_equal(img, expected)
+
+    img = np.zeros((5, 5), dtype=np.uint8)
+    start = (2, 2)
+    end = (2, 2)
+    rr, cc = rectangle_perimeter(start, end=end, shape=img.shape)
+    img[rr, cc] = 1
+    assert_array_equal(img, expected)
+
+    expected = np.array([[0, 0, 0, 0, 0, 0],
```
soupault: This test doesn't seem to add anything.

skimage/draw/tests/test_draw.py
```diff
@@ -891,3 +892,102 @@ def test_rectangle_extent():
...
+                         [1, 0, 0, 1, 0],
+                         [1, 1, 1, 1, 0],
+                         [0, 0, 0, 0, 0]], dtype=np.uint8)
+    img = np.zeros((5, 5), dtype=np.uint8)
+    start = (-5, -5)
+    end = (2, 2)
+    rr, cc = rectangle_perimeter(start, end=end, shape=img.shape,
+                                 clip=True)
+    img[rr, cc] = 1
+    assert_array_equal(img, expected)
```
soupault: I'd add a test with upper right start and lower left end / (+, -) extent.

Also, a very simple 3D test would be nice.

skimage/draw/draw.py
```diff
@@ -764,3 +764,76 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+    elif end is None:
+        raise ValueError("Either `end` or `extent` must be given")
+    top_left = np.minimum(start, end)
+    bottom_right = np.maximum(start, end)
+    if extent is None:
+        bottom_right += 1
+    top_left -= 1
+    r = [top_left[0], top_left[0], bottom_right[0], bottom_right[0],
```
soupault: This is not a valid perimiter for a 3D+ box, is it? I mean `Ndim` in the docstring doesn't actually represent arbitrary number of dimensions, as we only support 2D in this implementation.

skimage/draw/draw.py
soupault: Just a side note: I think we have to restrict usage of negative `extent` values. Here and also in `rectangle` function. This could be very confusing.

hmaarrfk: `asarray` ok.

I could add logic like:
```python
        extent = np.asarray(extent)
        if (extent < 0).any():
            raise ValueError("Extent should be positive.")
        end = np.array(start) + extent
```
but then that would duplicate logic like 
```
    tl = np.minimum(start, end)
    br = np.maximum(start, end)
```

skimage/draw/draw.py
hmaarrfk: Yes. I did not worry about 3D boxes. Should I? All the other functions except for `rectangle` support only 2D.

hmaarrfk: Fixed docstring to specify only 2 dimensions.

hmaarrfk:
Todo list from @soupault

[ ] consider stopping extent from being negative
[ ] Remove duplicate tests (keep only the "rectangle" test)
[ ] Use `np.asarray` instead of `np.array`
[ ] Add tests about top right and bottom left clipping.
[ ] Consider 3D tests
[ ] Add test for negative extent

skimage/draw/draw.py
soupault: What I meant here is that negative `extend` values are hard to understand (at least, for myself). For example, `extent (1,1)` returns a perimeter around 1x1 area. But for `extent (-1, -1)` it is not so intuitively to predict what will be the output. A simple solution to this issue would be to disallow negative values of `extent`.
If you think, however, that negative `extent`'s are useful, and it is not confusing at all, we can consider leaving this part as it is, and improving the docstring.

hmaarrfk: Got it. I'll see what negative extents actually do. It is not something I had considered.

hmaarrfk: Negative extents seem to allow you to specify a corner other than the top left one. Could be pretty useful.

```python
from skimage.draw import rectangle_perimeter
img = np.zeros(shape=(6, 6), dtype=int)
rr, cc = rectangle_perimeter(start=(3, 3), extent=(0, 0))
img[rr, cc] = 2
print(img)


img = np.zeros(shape=(6, 6), dtype=int)
rr, cc = rectangle_perimeter(start=(3, 3), extent=(-1, -1))
img[rr, cc] = 2
print(img)

img = np.zeros(shape=(6, 6), dtype=int)
rr, cc = rectangle_perimeter(start=(3, 3), extent=(-1, 2))
img[rr, cc] = 2
print(img)
```

skimage/draw/draw.py
soupault: >I did not worry about 3D boxes. Should I?

I wouldn't :). 2D is arguably the most common case. However, I'm pretty sure that some users will find 3D box perimeter drawing function useful (@jni ?).
I suppose it shouldn't be too complex implementing this (we have only to decide on the order of iteration for edges).

hmaarrfk: Personally, my personal feel is that I shouldn't implement the 3D boxes if I'm not going to put the effort in thinking about it (which I have little motivation for at the moment).

I'll leave it to the next person to take on the torch.

hmaarrfk:
@jni @soupault I think I finally figured out why the top left coordinate is offset but not the bottom right. The reason has to do with python indexing starting at 0. If you slice something starting at `0:2`, you are in fact selecting `0` and `1`. The polygon function seems to include both extremeties. Therefore both the top left and bottom right are actually shifted by 1. The top left needs to be specified, but the bottom right is done because of the way things are sliced in python. Maybe this is a bug in polygon though?

skimage/draw/draw.py
```diff
@@ -749,18 +752,119 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+    >>> start = (2, 3)
+    >>> end = (3, 4)
+    >>> rr, cc = rectangle_perimeter(start, end=end, shape=img.shape)
+    >>> img[rr, cc] = 1
+    >>> img
+    array([[0, 0, 0, 0, 0, 0],
+           [0, 0, 1, 1, 1, 1],
+           [0, 0, 1, 0, 0, 1],
+           [0, 0, 1, 0, 0, 1],
+           [0, 0, 1, 1, 1, 1]], dtype=uint8)
```
jni: @hmaarrfk this example seems wrong to me — didn't you offset the end coords as well? I think I would expect the coordinates to surround a rectangle defined by the slice [start0:end0, start1:end1]. This means that in this case, they should surround a rectangle of shape (1, 1), not (2, 2). Thanks for figuring out the offsets thing btw. =)

For the record, I don't think drawing boxes needs to happen in nD, at least initially, but thanks for considering it. =)

hmaarrfk: Hi @jni , Thanks for looking through the examples. The rectangle function includes the `end` points. Therefore, it is defined by the rectangle is defined by the pythonic slice  `[start:end+1]`.

I'll update the comments in both rectangle functions.

1. 
+    >>> start = (2, 3)
+    >>> end = (3, 4)
+    >>> rr, cc = rectangle_perimeter(start, end=end, shape=img.shape)

skimage/draw/draw.py
```diff
@@ -749,18 +754,121 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+
+    top_left -= 1
+    r = [top_left[0], top_left[0], bottom_right[0], bottom_right[0],
+         top_left[0]]
+    c = [top_left[1], bottom_right[1], bottom_right[1], top_left[1],
+         top_left[1]]
+    return polygon_perimeter(r, c, shape=shape, clip=clip)
+
+
+def _rectangle_bounds(start, end=None, extent=None):
```
soupault: What do you think about changing the name of the function to `_rectangle_slice` or something similar? I think `bounds` can be understood in different ways.

skimage/draw/draw.py
```diff
@@ -749,18 +754,121 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+
+    if extent is not None:
+        end = np.asarray(start) + np.asarray(extent)
+    top_left = np.minimum(start, end)
+    bottom_right = np.maximum(start, end)
+
+    if extent is None:
+        bottom_right += 1
+
+    return (bottom_right, top_left)
```
soupault: I'd rather return the results in ndarray, not tuple.

skimage/draw/draw.py
hmaarrfk: It isn't a python `slice` though. So I'm hesitant to call it a `slice`.

`_rectangle_corners` ?
`_rectangle_bottom_right_top_left` ?

skimage/draw/draw.py
hmaarrfk: I guess for me I see `bottom_right` and `top_left` as two distinct objects, that probably shouldn't be acted on together.

This also made it a drop in replacement for the previous code.
If I make `ndarray`, then we have to find a convention on how to order the array and whatnot.

I'll make the change if you insist, but since there are only 2 corners that will ever exist in this function, I'm inclined to keep it as a tuple.

skimage/draw/draw.py
soupault: My point is that the function doesn't actually return _corners_ (see the example below: upper left point is inclusive, but the lower right isn't). So the output is conceptually more close to Python `slice`.
```
In [5]: _rectangle_bounds((1, 1), (2, 2))
Out[5]: (array([3, 3]), array([1, 1]))
```

skimage/draw/draw.py
soupault: Yes, sorry, you are right. Ignore this point.

skimage/draw/draw.py
hmaarrfk: Ok, I'll change it to slice. And swap the return values around....

soupault:
@jni any last remarks?

soupault:
@jni Kind extra ping :wink:

hmaarrfk:
@jni hi!

skimage/draw/draw.py
```diff
@@ -749,18 +754,132 @@ def rectangle(start, end=None, extent=None, shape=None):
...
+    >>> img[rr, cc] = 3
+    >>> rr, cc = rectangle(start, extent=(2, -2))
+    >>> img[rr, cc] = 4
+    >>> print(img)
+    [[0 0 0 0 0 0]
+     [0 3 3 2 2 0]
+     [0 3 3 2 2 0]
+     [0 4 4 1 1 0]
+     [0 4 4 1 1 0]
+     [0 0 0 0 0 0]]
```
jni: This example is :fire: :fire: :fire: 

Thanks @hmaarrfk!

hmaarrfk:
@jni, hi!

hmaarrfk:
Wooooooo. This finally got in!!!

That you all for your reviews.

stefanv:
Great, now we can start working on `rectangle_perimeter_aa` ;)

Thanks, @hmaarrfk.

PR #3070, opened 2018-05-11
Allow passing callable for high and low threshold of Canny

### LINKED ISSUE #3054: Proposal: Add option of using threshold algorithm in Canny
## Description

As explained in the "Improvement" section of the [wiki page on the Canny edge detector](https://en.wikipedia.org/wiki/Canny_edge_detector#Robust_method_to_determine_the_dual-threshold_value), can we add in an option to use e.g. `threshold_otsu` (or any other?) of the threshold algorithms to determine the high threshold, instead of requiring it to be passed in or using the image `dtype`. 

As suggested in the wiki page we could start with 

```
high_thresh = threshold_otsu(magnitude) 
low_thresh = 0.5*high_thresh
```
 (or similar), and expand to allowing any of the `skimage` threshold algorithms to be used.

jmetz:
## Description
Adds option of passing callables for `high_threshold` and `low_threshold`, which 
take the magnitude image as inputs. This allows more general threshold strategies. 

## Checklist
- [x] Unit tests

## References

Closes #3054, a feature request to add this option.

## For reviewers

  later.

### DIFF
--- a/doc/examples/edges/plot_canny.py
+++ b/doc/examples/edges/plot_canny.py
@@ -20,7 +20,7 @@ import matplotlib.pyplot as plt
 from scipy import ndimage as ndi
 
 from skimage import feature
-from skimage.filters import threshold_li, threshold_otsu
+from skimage.filters import threshold_otsu
 
 
 # Generate noisy image of a square
@@ -74,7 +74,7 @@ axes[0][3].set_title('Canny filter, $\sigma=3$\nusing quantiles')
 
 axes[0][4].imshow(edges1_4,  cmap=plt.cm.gray)
 axes[0][4].axis('off')
-axes[0][4].set_title('Canny filter, $\sigma=3$\nLi thresholding')
+axes[0][4].set_title('Canny filter, $\sigma=3$\nOtsu threshold (callable)')
 
 axes[1][0].imshow(im2, cmap=plt.cm.gray)
 axes[1][0].axis('off')
@@ -94,7 +94,7 @@ axes[1][3].set_title('Canny filter, $\sigma=3$\nquantiles')
 
 axes[1][4].imshow(edges2_4,  cmap=plt.cm.gray)
 axes[1][4].axis('off')
-axes[1][4].set_title('Canny filter, $\sigma=3$\nLi thresholding')
+axes[1][4].set_title('Canny filter, $\sigma=3$\nOtsu threshold (callable)')
 
 fig.tight_layout()
 
--- a/doc/examples/edges/plot_canny.py
+++ b/doc/examples/edges/plot_canny.py
@@ -20,6 +20,7 @@ import matplotlib.pyplot as plt
 from scipy import ndimage as ndi
 
 from skimage import feature
+from skimage.filters import threshold_li, threshold_otsu
 
 
 # Generate noisy image of a square
@@ -30,25 +31,70 @@ im = ndi.rotate(im, 15, mode='constant')
 im = ndi.gaussian_filter(im, 4)
 im += 0.2 * np.random.random(im.shape)
 
+# Generate lower SNR noisy image of a square
+im2 = np.zeros((128, 128))
+im2[32:-32, 32:-32] = 1
+
+im2 = ndi.rotate(im2, 15, mode='constant')
+im2 = ndi.gaussian_filter(im2, 4)
+im2 += 0.4 * np.random.randn(*(im2.shape))
+
 # Compute the Canny filter for two values of sigma
-edges1 = feature.canny(im)
-edges2 = feature.canny(im, sigma=3)
+edges1_1 = feature.canny(im)
+edges1_2 = feature.canny(im, sigma=3)
+edges1_3 = feature.canny(im, use_quantiles=True, sigma=3)
+edges1_4 = feature.canny(im, sigma=3, high_threshold=threshold_otsu,
+                         low_threshold=threshold_otsu)
+
+edges2_1 = feature.canny(im2)
+edges2_2 = feature.canny(im2, sigma=3)
+edges2_3 = feature.canny(im2, sigma=3, use_quantiles=True)
+edges2_4 = feature.canny(im2, sigma=3, low_threshold=threshold_otsu,
+                         high_threshold=threshold_otsu)
 
 # display results
-fig, (ax1, ax2, ax3) = plt.subplots(nrows=1, ncols=3, figsize=(8, 3),
+fig, axes = plt.subplots(nrows=2, ncols=5, figsize=(12, 6),
                                     sharex=True, sharey=True)
 
-ax1.imshow(im, cmap=plt.cm.gray)
-ax1.axis('off')
-ax1.set_title('noisy image', fontsize=20)
+axes[0][0].imshow(im, cmap=plt.cm.gray)
+axes[0][0].axis('off')
+axes[0][0].set_title('Low noise image')
 
-ax2.imshow(edges1, cmap=plt.cm.gray)
-ax2.axis('off')
-ax2.set_title('Canny filter, $\sigma=1$', fontsize=20)
+axes[0][1].imshow(edges1_1, cmap=plt.cm.gray)
+axes[0][1].axis('off')
+axes[0][1].set_title('Canny filter, $\sigma=1$')
 
-ax3.imshow(edges2, cmap=plt.cm.gray)
-ax3.axis('off')
-ax3.set_title('Canny filter, $\sigma=3$', fontsize=20)
+axes[0][2].imshow(edges1_2, cmap=plt.cm.gray)
+axes[0][2].axis('off')
+axes[0][2].set_title('Canny filter, $\sigma=3$')
+
+axes[0][3].imshow(edges1_3, cmap=plt.cm.gray)
+axes[0][3].axis('off')
+axes[0][3].set_title('Canny filter, $\sigma=3$\nusing quantiles')
+
+axes[0][4].imshow(edges1_4,  cmap=plt.cm.gray)
+axes[0][4].axis('off')
+axes[0][4].set_title('Canny filter, $\sigma=3$\nLi thresholding')
+
+axes[1][0].imshow(im2, cmap=plt.cm.gray)
+axes[1][0].axis('off')
+axes[1][0].set_title('High noise image')
+
+axes[1][1].imshow(edges2_1, cmap=plt.cm.gray)
+axes[1][1].axis('off')
+axes[1][1].set_title('Canny filter, $\sigma=1$\ndefault threshold')
+
+axes[1][2].imshow(edges2_2, cmap=plt.cm.gray)
+axes[1][2].axis('off')
+axes[1][2].set_title('Canny filter, $\sigma=3$\ndefault threshold')
+
+axes[1][3].imshow(edges2_3, cmap=plt.cm.gray)
+axes[1][3].axis('off')
+axes[1][3].set_title('Canny filter, $\sigma=3$\nquantiles')
+
+axes[1][4].imshow(edges2_4,  cmap=plt.cm.gray)
+axes[1][4].axis('off')
+axes[1][4].set_title('Canny filter, $\sigma=3$\nLi thresholding')
 
 fig.tight_layout()
 
--- a/doc/examples/edges/plot_canny.py
+++ b/doc/examples/edges/plot_canny.py
@@ -56,45 +56,38 @@ edges2_4 = feature.canny(im2, sigma=3, low_threshold=threshold_otsu,
 fig, axes = plt.subplots(nrows=2, ncols=5, figsize=(12, 6),
                                     sharex=True, sharey=True)
 
-axes[0][0].imshow(im, cmap=plt.cm.gray)
-axes[0][0].axis('off')
-axes[0][0].set_title('Low noise image')
+axes[0, 0].imshow(im, cmap=plt.cm.gray)
+axes[0, 0].set_title('Low noise image')
 
-axes[0][1].imshow(edges1_1, cmap=plt.cm.gray)
-axes[0][1].axis('off')
-axes[0][1].set_title('Canny filter, $\sigma=1$')
+axes[0, 1].imshow(edges1_1, cmap=plt.cm.gray)
+axes[0, 1].set_title('Canny filter, $\sigma=1$')
 
-axes[0][2].imshow(edges1_2, cmap=plt.cm.gray)
-axes[0][2].axis('off')
-axes[0][2].set_title('Canny filter, $\sigma=3$')
+axes[0, 2].imshow(edges1_2, cmap=plt.cm.gray)
+axes[0, 2].set_title('Canny filter, $\sigma=3$')
 
-axes[0][3].imshow(edges1_3, cmap=plt.cm.gray)
-axes[0][3].axis('off')
-axes[0][3].set_title('Canny filter, $\sigma=3$\nusing quantiles')
+axes[0, 3].imshow(edges1_3, cmap=plt.cm.gray)
+axes[0, 3].set_title('Canny filter, $\sigma=3$\nusing quantiles')
 
-axes[0][4].imshow(edges1_4,  cmap=plt.cm.gray)
-axes[0][4].axis('off')
-axes[0][4].set_title('Canny filter, $\sigma=3$\nOtsu threshold (callable)')
+axes[0, 4].imshow(edges1_4,  cmap=plt.cm.gray)
+axes[0, 4].set_title('Canny filter, $\sigma=3$\nOtsu threshold (callable)')
 
-axes[1][0].imshow(im2, cmap=plt.cm.gray)
-axes[1][0].axis('off')
-axes[1][0].set_title('High noise image')
+axes[1, 0].imshow(im2, cmap=plt.cm.gray)
+axes[1, 0].set_title('High noise image')
 
-axes[1][1].imshow(edges2_1, cmap=plt.cm.gray)
-axes[1][1].axis('off')
-axes[1][1].set_title('Canny filter, $\sigma=1$\ndefault threshold')
+axes[1, 1].imshow(edges2_1, cmap=plt.cm.gray)
+axes[1, 1].set_title('Canny filter, $\sigma=1$\ndefault threshold')
 
-axes[1][2].imshow(edges2_2, cmap=plt.cm.gray)
-axes[1][2].axis('off')
-axes[1][2].set_title('Canny filter, $\sigma=3$\ndefault threshold')
+axes[1, 2].imshow(edges2_2, cmap=plt.cm.gray)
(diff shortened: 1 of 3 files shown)


skimage/feature/_canny.py
```diff
@@ -60,11 +60,15 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
         Greyscale input image to detect edges on; can be of any dtype.
     sigma : float
         Standard deviation of the Gaussian filter.
-    low_threshold : float
+    low_threshold : float or callable
         Lower bound for hysteresis thresholding (linking edges).
+        If callable is given, it is applied to the gradient image
+        to generate the low_threshold.
```
emmanuelle: Maybe you could extend the Examples section to illustrate this possibility, maybe with percentiles?

emmanuelle:
Hi @jmetz thank you for your PR. What kind of functions would typically be used to compute thresholds in your experience, apart from percentiles? If it's only percentiles, we can also think of a keyword argument ``threshold_mode`` (absolute, percentile etc.) which could be easier to use for non-expert users.

jmetz:
Hi @emmanuelle - I've used skimage's own `threshold_li` and `threshold_otsu` in the past... in principal though it can be any function that takes an image (2d ndarray) and returns a single value representing a threshold to use. 

Because of this I'd suggest at least having the option to pass in a `callable`... if need be we could possibly have two canny interfaces, something like `canny` and `canny_advanced` or similar, where 

* `canny` implements the original simple function which takes scalar threshold values (and possibly the quartile option as it currently does). Internally this could simply wrap `canny_advanced`. 
* `canny_advanced` would then take the callables as shown in this PR. 

Thoughts?

jni:
Hi @emmanuelle and @jmetz 

I am -1 for doing separate functions. We already support int/float thresholds *and* quantiles. From the users' perspective, this simplicity never goes away. Advanced users would then have the complex use at their disposal if they so desire, without the need to clutter the API.

Another example where the advanced API could be used: using `threshold_sauvola` with different values of k.

jmetz:
I tend to agree @jni - are you happy with the PR as is then?

skimage/feature/_canny.py
jni: @jmetz, this docstring statement isn't bad but it's not great either. Specifically, I want it to state the type of the callable: takes in `image` as its only input and returns either a float or an array of the same shape as `image` as output.

skimage/feature/_canny.py
```diff
@@ -265,6 +269,16 @@ def fsmooth(x):
 
         high_threshold = np.percentile(magnitude, 100.0 * high_threshold)
         low_threshold = np.percentile(magnitude, 100.0 * low_threshold)
+    else:
+        #
+        # Else if high_threshold and/or low_threshold are callables
+        # call them to determine the threshold value / image
+        #
```
jni: Do you mind removing the whitespace comments? I would also change `call them` to `apply them to the input image`.

skimage/feature/_canny.py
```diff
@@ -265,6 +269,16 @@ def fsmooth(x):
...
+        if callable(high_threshold):
+            high_threshold = high_threshold(magnitude)
+        if callable(low_threshold):
+            low_threshold = low_threshold(magnitude)
```
jni: I'm not going to hold up the PR for this, but ideally we would do some sanity checking here: use try/except so that if there's an error, *we* raise a TypeError with the appropriate error message (the input type of the callable passed to Canny is wrong; and reproduce the traceback that came with the call), and if the type of high_threshold and low_threshold are wrong, then we raise a TypeError saying that the *output* type of the callable is wrong.

If you can add this we would be eternally grateful! =)

jni:
@jmetz I won't hold the PR for the example though if you don't have time right now. Let us know! And thanks again! =)

skimage/feature/_canny.py
jmetz: Just working on the example quickly over lunch ;)

jmetz:
@jni - Added callable usage to example. 

Even for simple data (in this case just making a lower SNR square image) shows the strength of using a callable (here I used `threshold_otsu`).

doc/examples/edges/plot_canny.py
```diff
@@ -30,25 +31,70 @@
...
-fig, (ax1, ax2, ax3) = plt.subplots(nrows=1, ncols=3, figsize=(8, 3),
+fig, axes = plt.subplots(nrows=2, ncols=5, figsize=(12, 6),
                                     sharex=True, sharey=True)
 
-ax1.imshow(im, cmap=plt.cm.gray)
-ax1.axis('off')
-ax1.set_title('noisy image', fontsize=20)
+axes[0][0].imshow(im, cmap=plt.cm.gray)
+axes[0][0].axis('off')
+axes[0][0].set_title('Low noise image')
```
jni: Can you change this to NumPy indexing? [0, 0], etc. Much neater imho. 

I also feel like some of this repetition can be removed: instead of individual calls to `axis('off')`, remove them all and at the end say, `for ax in axes.ravel(): ax.axis('off')`.

jmetz:
@jni - thanks, I should be able to add those last changes in later today - though your link to the comment that I didn't address doesn't seem to take me to any specific comments, would you be able to let me know which one it is?

jni:
Ah it’s the thing about producing a meaningful error message when the callables don’t produce the right type of output or fail with the given input

jmetz:
@jni - just taking a look at this now... so cleaning up the example to use a loop is straight-forward. 
Only question there would be how far to go. E.g. I could conceivably use: 

```python
names_and_ims = (
    ('Input Image', im), 
    ('Default', edges_1), 
    ...
)
for ax, (name, im) in zip(axes.ravel(), names_and_ims):
    ax.imshow(im)
    ax.set_title(name)
```

In terms of analysing the callables, it's a little less straight-forward how I should check that they have the correct call signature and return the correct things. 
As far as I can tell there are two main groups of options: 

a) Put the callable-related code in a try-except and then raise an error on an except
b) Check the output (e.g. `isinstance(result, numbers.Number) or (isinstance(result, np.array) and (result.ndim == 2))`) as well as checking the correct number of inputs with `inspect`. 

I'd suggest we go with a), as that's more Pythonic, and cleaner to code up - what do you guys thing?

jni:
@jmetz there are two types of errors to catch:

(1) the callable fails with a typeerror, which probably means that the callable passed has the wrong input type. We should put this in a try/except clause and raise the appropriate error in the except (a TypeError with an explanation of how to use callables here, and the full traceback to help diagnose things).
(2) the callable works but returns an output that is the wrong type (not float or ndarray). This should also be caught (this time with an if statement) and an error raised.

Regarding the example, I would only go for removing the `axis('off')` calls. The rest of the stuff, I think, would be *less* clear if drawn into a loop, except for expert users. Would you agree?

jmetz:
@jni - I've implemented the changes outlined, but just before I commit, final detail: I check that the output of the thresholds is either a 2d array or a `numbers.Number` (i.e. allow for ints etc, so in case the chosen function returns that...) - you mentioned checking for `float`... just wanted to clarify whether you're happy accepting `numbers.Number` or want to stick specifically to `float`?

jni:
@jmetz when you say "2D array" I hope you mean "array of the same shape as the input image"? =)

Yes `numbers.Number` is exactly correct, thank you for pointing that out!

skimage/feature/_canny.py
```diff
@@ -275,11 +277,27 @@ def fsmooth(x):
...
+            try:
+                high_threshold = high_threshold(magnitude)
+            except:
+                traceback.print_exc()
+                raise ValueError("Callable `high_threshold` raised above exception.\nIt must take one input (image array) and return a scalar value or image array of the same shape as input image")
+
+            if not( isinstance(high_threshold, numbers.Number) or (
+                    isinstance(high_threshold, np.ndarray) and
+                    (high_threshold.shape == magnitude.shape))):
+                raise ValueError("Callable `high_threshold` must take one input (image array) and return a scalar value or image array of the same shape as input image")
```
jni: Thanks @jmetz! These are great, except they should be `TypeError`s. (the same error as when a function is called with the wrong arguments.) And incidentally the `except` above should not be bare — it should only catch `TypeError`s!

jmetz: Ah... So I checked about TypeError Vs ValueError, and because the issue is with the "value" of the callables, I thought ValueError would be more appropriate?
I.e. to quote the docs for ValueError:
"Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError."

Also as for not having a bare except (I know that's generally quite bad), if there is any issue with a called function, I thought it might still be useful to have the info from canny also, not just have it raise the internal exception... Happy to only catch ValueError though if that's preferred.

jmetz: (put another way, as we're in the case where eg high_threshold *is* a callable, it's type is technically already correct... Just that it's specific value is wrong... Right?)

stefanv:
One confusing thing about this example: why can you provide the same function (otsu) and get different high/low threshold values out?

jmetz:
@stefanv : the code for using the high and low threshold callables was inspired by this (and my own uses in the past): https://en.wikipedia.org/wiki/Canny_edge_detector#Robust_method_to_determine_the_dual-threshold_value

There, the lower threshold value is set to half of the higher value (ie requires calling the higher threshold callable). In general then, the low threshold callable gets fed the output of calling the high threshold callable, ie `low_threshold(magnitude [magnitude<high_threshold])`.

Otherwise, to obtain the same result, the high_threshold function would have to be passed in as part of the low_threshold callable.

skimage/feature/_canny.py
jni: I once thought the same as you, but I've learned a bit more about type theory since then. =) Specifically, the type of the callable is not just "callable", but `Union[(array -> float), (array -> array)]`. That is: the inputs and outputs of the callable are part of the type. (Whereas a ValueError would stem from, e.g. returning a float that is outside some allowed range, for example.)

And, as I mentioned, you should catch *TypeError*, because that is the error we are detecting: when the function is unhappy with the inputs as provided.

The errors will propagate through if they are of a different type, and this is what we want: our error message that "you have provided the wrong type of function" will not necessarily be accurate if we catch all errors. If someone, for example, provides a threshold function that sends the data to some Google Compute Engine to process on their TPUs (to get back a threshold 😂), and there is a NetworkError, we don't want to say that it's the wrong kind of function: it is the right kind, but it failed for extraneous reasons.

jmetz: @jni - many thanks for the explanation, glad I know more about type theory wrt functions now too :) - I'll amend  in the last commit in the morning.

stefanv:
@jmetz That's a neat method, but all this information should be in the docs.

jmetz:
@stefanv - good point, I'll add that's this morning's amendments also

jmetz:
Actually on reflection I think @stefanv has pointed out a good issue. 

Essentially, broadly speaking there were 3 types of options in making the choice about what to pass in to `low_threshold` when it is a callable (here `magnitude` is the gradient magnitude image): 

a) only pass in the original `magnitude` image - i.e. exactly the same input as for `high_threshold` as a callable
b) Pass in only the data in `magnitude` after applying `high_threshold` (this is what I currently have) 
c) A hybrid approach, e.g. pass in two inputs - original `magnitude` and the `high_threshold` mask image (set to True where `magnitude` is less than result from `high_threshold` - a general way to represent the action of `high_threshold`) 

The reason I chose **a)** is that it's what I've done in the past and it's worked well, as well it avoiding e.g. having to apply `high_threshold` twice as would happen with **b)**. Beyond that I didn't dwell too long on this question. 

Taking for example if we wished to exactly replicate the suggestion in https://en.wikipedia.org/wiki/Canny_edge_detector#Robust_method_to_determine_the_dual-threshold_value, i.e. using **Otsu** for `high_threshold` and then `0.5*high_threshold` value for `low_threshold`, then the three methods outlines would require a `low_threshold` similar to: 

a) `low_threshold = lambda im: 0.5 * np.max(im)` 
b) `low_threshold = lambda im: 0.5 * high_threshold(im)`
c) `low_threshold = lambda im, mask: 0.5 * np.max(im[mask])`

i.e. **b)** clearly requires re-calling whatever callable is passed for `high_threshold`. 

*However* on the downside to **a)**, it does lock-in the use of `high_threshold` in the calculation of `low_threshold`, so perhaps in fact something like **b)** (or **c)**) would be best in terms of generality. 

@stefanv & @jni - your input would be much appreciated in confirming / changing the choice of what is passed in to the `low_threshold` callable.

stefanv:
It's hard to make this general, because all existing methods imply a certain order of thresholding (high, then low).  What if some method comes along that does it the opposite way?

Why don't we add a flag to canny to enable only  what you want to achieve here specifically, instead of trying to expose a more general API?

E.g.

```
out = canny(image, threshold='hist')
```

jmetz:
@stefanv - I can't see how this can work... setting, for example `high_threshold` to `"robust"`, but not `low_threshold` could easily result in nonsensical thresholds (e.g. `low_threshold` above `high_threshold`). 

As you point out, the whole logic of the low and high threshold does, to some extent, imply a dependency, i.e. `low_threshold` should always depend on `high_threshold` for Canny to work. 

My vote would be to leave it as is and accept that this dependency is a current limitation, but I'm happy to switch to having one or several hard-coded explicit implementations (such as what I used for the example).

skimage/feature/tests/test_canny.py
```diff
@@ -104,3 +104,38 @@ def test_invalid_use_quantiles(self):
 
         self.assertRaises(ValueError, F.canny, image, use_quantiles=True,
                           low_threshold=0.5, high_threshold=-100)
+
+    def test_use_callables(self):
+        image = img_as_float(data.camera()[::50, ::50])
+
+        # Correct output produced manually with quantiles
+        # of 0.8 and 0.6 for high and low respectively
+        correct_output = np.array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
```
Borda: use a new line...

skimage/feature/_canny.py
```diff
@@ -265,6 +273,31 @@ def fsmooth(x):
...
+        if callable(low_threshold):
+            mask_threshold = magnitude < high_threshold
+            try:
+                low_threshold = low_threshold(magnitude[mask_threshold])
+            except:
+                traceback.print_exc()
+                raise ValueError("Callable `low_threshold` raised above exception.\nIt must take one input (image array) and return a scalar value or image array of the same shape as input image")
```
Borda: I think this is too long line...

stefanv:
@jmetz Do you think the robust thresholding method can be implemented (perhaps not with ideal efficiently, but still) by calling the existing threshold method multiple times?  If so, the complexity this PR generates may be worth avoiding, and either adding a new method that does it that way (for now) or documenting that a robust method is possible and giving appropriate pointers.

jni:
I agree. This comment comes from a discussion in the group meeting about the PR. Unless I'm mistaken, these things could be achieved outside of the function of interest, so perhaps this PR is best replaced by a gallery example, which I think would actually be a very valuable addition to our gallery!

jmetz:
Hi @stefanv & @jni  - tbh I'm a little lost now... So to recap, my original PR was to add in extended functionality in canny by allowing the user to pass in callables as `high_threshold` and `low_threshold`. 

I can see how this would be deemed quite an advanced functionality and therefore there's reluctance to accept this PR, but I'm confused what's being proposed now instead?

jni:
You linked to a part of the wikipedia article that mentions how to find a robust threshold: set the high threshold to be something principled (e.g. Otsu but I guess anything will do), then, set the low threshold to be the same principled method *limited to the pixels below the original threshold*. I think this would make a very nice example in the gallery.

In fact, I might be inclined to make this be the default going forward. The old defaults seem pretty random!

jmetz:
There is a problem with that idea - using Otsu to calculate the threshold values can only be done either
- within the `canny` function as the array the threshold values are applied to are not the input image, but the sobel filtered values (which happens within `canny`). 
- or on a filtered image which has been filtered in _exactly the same way_ as the data will be filtered within `canny`. 

This was the reason I first suggested passing _callables_ for the threshold values. 
We could though as you suggest simply replace the implementation to use the enhanced threshold determination (i.e. Otsu) within `canny`. 

Either way, IMO it makes most sense to close this PR and create one or more issues to cover these proposals.

jmetz:
On a related but separate note, I would also suggest that the `canny` implementation be broken up into it's constituent subfunctions, namely (roughly): 

- edge filtering (sobel magnitude) 
- non-maxima suppression
- threshold determination 
- hysteresis thresholding 

such that `canny` itself becomes much shorter (and passes PEP8 wrt function length :sweat_smile: ) and advanced users can then stitch together the relevant functions, but replacing e.g. the threshold determination step to much more easily create their own canny variant functions.

rfezzani:
Sorry for coming late to the discussion.
I agree with your last comment @jmetz, I refactored the canny function in #4342 ;-).

I see two other options concerning the discussion above:
- if we want to allow passing callable for high and low threshold with more flexibility, we can define a new argument `thresholds` defining both `low_threshold` and `high_threshold` (mainly `thresholds=(low_threshold, high_threshold)`). Doing so, the robust thresholds selection can be implemented if threshold is a callable...
- we can also modify the default thresholds selection (when `threshold_low` and `threshold_high` are None) to be the robust selection method.

rfezzani:
@jmetz I propose in #5114 to deprecate `low_threshold` and `high_threshold` in favor of the unique parameter `thresholds`, this may ease the implementation of this PR :wink:

stefanv:
@rfezzani How would this work with other thresholding methods that only have one threshold?  Perhaps the keyword should be `threshold`, and for some algorithms it should be allowed to be a tuple?

rfezzani:
As @jni [mentioned](https://github.com/scikit-image/scikit-image/pull/5114#issuecomment-741447266), simply adding a `thresholds` argument that supplant `low_threshold` and `high_threshold` is an excellent solution to allow passing callable. What do you think @jmetz?

jmetz:
> As @jni [mentioned](https://github.com/scikit-image/scikit-image/pull/5114#issuecomment-741447266), simply adding a `thresholds` argument that supplant `low_threshold` and `high_threshold` is an excellent solution to allow passing callable. What do you think @jmetz?

Thanks for the updates to this question @rfezzani, and I agree - switching to a more generic `thresholds` parameter could be a nice way to achieve additional versatility (and the modification to the current PR should be minimal).

stefanv:
@jmetz Thank you for this PR. It was a great thought experiment in how we can improve the canny implementation in scikit-image, and several concrete ideas have been proposed that we should follow-up on.

This specific PR, then, is probably not the right fit, but I hope to see some follow-up PRs that take on the rest.

Thanks again for going through the whole review cycle and engaging in such constructive conversation; I know it's not ideal to have a PR closed after all that work, but I think we all learned a lot.

PR #2935, opened 2018-01-07
Handle NumPy 1.14 API changes

oleksandr-pavlyk:
A number of test errors result when testing scikit-image with just released NumPy 1.14. The following change fixes the following errors:

```
'test_rag.test_rag_merge',
'test_rag.test_rag_hierarchical',
'test_rag.test_cut_normalized',
'test_rag.test_threshold_cut',
'skimage.io.tests.test_mpl_imshow.test_signed_image',
'skimage.io.tests.test_mpl_imshow.test_uint8',
'skimage.io.tests.test_mpl_imshow.test_uint16',
'skimage.io.tests.test_mpl_imshow.test_float',
'skimage.io.tests.test_mpl_imshow.test_nonstandard_type',
'skimage.io.tests.test_mpl_imshow.test_outside_standard_range',
'skimage.io.tests.test_mpl_imshow.test_low_data_range',
```

The fix is the same as [was applied](https://github.com/scikit-learn/scikit-learn/pull/9683/files#diff-904f5f849e36c648b8b05de140d7fad8R1089) to scikit-learn.

PR #3072, opened 2018-05-12
Update dependencies and deprecations to fix Travis builds

### LINKED ISSUE #3034: Weird matplotlib infinite loop Travis failure
I'm not sure yet whether this is reproducible, but on Py2.7 with minimal requirements (MPL 1.3.1), we are getting a maximum recursion depth error:

https://travis-ci.org/scikit-image/scikit-image/jobs/372237528#L5768

It doesn't appear to be random OR due to a new Matplotlib release, so I really don't understand why it's popping up now. 

Environment variables:  
WITH_PYAMG=1  
MINIMUM_REQUIREMENTS=1  
WITH_QT=1

The failing example is `doc/examples/color_exposure/plot_ihc_color_separation.py`, line 72, `axis.imshow(zdh)`.

hmaarrfk:
## Description
Travis now passes! 
I'm no git fu master, so maybe just a squash merge is OK?

Relevant changes are:
- [x] A bunch of warnings for deprecated default values (I set the values to the old default values)
- [x] Matplotlib minimum now 2.0.0 to pass some tests. Fixes #3034
- [x] Pillow minimum now 4.3.0 to have support for jpeg decoder (4.2.1 had bugs and was vitoed by @jni)
- [x] Building docs with a modern version of python. Now using 3.6.
- [x] All tests run on Linux + OSX
- [x] All tests run with `pytest -v` do not issue warnings (on 3.6) (Doctests still issue warnings, i think that is ok)
- [x] Doctests still have some warnings. (Catching them would make the docs less readable sometimes)
- [x] Optional dependencies installed in the `install` section of travis
- [x] Single shot pytests. No running once, installing more dependencies, then running again.
- [x] `--pre` builds for 2.7 and 3.6
- [x] Qt4 for 2.7. <- Thought it has been said that qt4 is no longer on pypi
- [x] Qt5 for 3.6

Things that were passed off to future issues:
- [x] Find a solution for GUI tests that I currently skip. #3081
- [x] 3.4 sometimes fails on wrap around test. #3079 
- [x] OSX cannot install #3084

## Checklist
- [x] Unit tests

## For reviewers

  later.


hmaarrfk:
Other things I noticed:
# Running the tests twice
The tests are effectively being run twice:
1. With minimum dependencies
2. With more dependencies.

This makes it take twice the amount of time, and is likely overkill. Also, the results from travis come later. I would be for faster initial results from travis so that when you are still in "development" mode, you can get quick feedback, even if it means creating more configurations.

# OSX Failing
OSX  tests seem to fail on some documentation dependency requirements that it has to build from source. Since OSX isn't being used to generate the dependencies, I simply didn't pull the dependencies on OSX.


# Pythong 2.7 failing
Some combination of PYAMG, QT, MINIMUM_REQUIREMENTS is causing 2.7 build to fail. Will this cause 3.X build to fail too when PYAMG or QT is tested with them?

.travis.yml
```diff
@@ -36,26 +40,28 @@ notifications:
 
 matrix:
   include:
+    # 2.7 build (until 0.14 is release)
     - os: linux
       python: 2.7
-      env: PIP_FLAGS="--pre"
+      env: WITH_PYAMG=1 QT=PyQt4
+      # Examples need strange versions of Matplotlib and Qt
+      # I don't think we should be increasing the minimum version requirement
```
jni: @hmaarrfk can you elaborate on this? Given [this comment](https://github.com/scikit-image/scikit-image/issues/3034#issuecomment-389424649), I'm inclined to increase minimum mpl to 2.0 (over a year old now), and minimum Qt to 5.

.travis.yml
```diff
@@ -72,12 +78,60 @@ before_install:
...
+        cp $MPL_DIR/matplotlibrc $MPL_DIR/matplotlibrc_backup
+        echo 'backend : Template' > $MPL_DIR/matplotlibrc
+        for f in doc/examples/*/*.py; do
+            python "${f}"
+            if [ $? -ne 0 ]; then
+                exit 1
+            fi
+        done
+        mv $MPL_DIR/matplotlibrc_backup $MPL_DIR/matplotlibrc
+      fi
```
jni: Why did you change back from a script to travis.yml lines?

skimage/io/tests/test_pil.py
```diff
@@ -45,18 +45,19 @@ def test_png_round_trip():
...
+        imsave(fname, I)
     Ip = img_as_float(imread(fname))
     os.remove(fname)
     assert np.sum(np.abs(Ip-I)) < 1e-3
 
 
 def test_imread_flatten():
     # a color image is flattened
-    img = imread(os.path.join(data_dir, 'color.png'), flatten=True)
+    img = imread(os.path.join(data_dir, 'color.png'), as_gray=True)
```
jni: If I'm not mistaken, this test should remain until `flatten` is removed, but with `expected_warnings`.

skimage/morphology/tests/test_misc.py
```diff
@@ -163,7 +164,8 @@ def test_uint_image_holes():
                          [0, 0, 0, 0, 0, 0, 0, 1, 1, 1],
                          [0, 0, 0, 0, 0, 0, 0, 1, 1, 1],
                          [0, 0, 0, 0, 0, 0, 0, 1, 1, 1]], dtype=np.bool_)
-    observed = remove_small_holes(labeled_holes_image, area_threshold=3)
+    with expected_warnings(['Any labeled images will be returned as a boolean array.']):  # noqa
```
jni: You don't need the full string here.

skimage/morphology/tests/test_skeletonize_3d.py
```diff
@@ -126,7 +126,10 @@ def test_skeletonize_num_neighbours():
     circle2 = (ic - 135)**2 + (ir - 150)**2 < 20**2
     image[circle1] = 1
     image[circle2] = 0
-    result = skeletonize_3d(image)
+    with warnings.catch_warnings():
+        # UserWarning for possible precision loss, expected
+        warnings.simplefilter('ignore', UserWarning)
+        result = skeletonize_3d(image)
```
jni: Why not use `expected_warnings` here also?

skimage/restoration/tests/test_denoise.py
```diff
@@ -224,7 +224,9 @@ def test_denoise_bilateral_multidimensional():
 
 def test_denoise_bilateral_nan():
     img = np.full((50, 50), np.NaN)
-    out = restoration.denoise_bilateral(img, multichannel=False)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        out = restoration.denoise_bilateral(img, multichannel=False)
```
jni: Here also, why not `expected_warnings`?

skimage/segmentation/tests/test_random_walker.py
```diff
@@ -78,6 +79,7 @@ def test_2d_bf():
     assert data.shape == labels.shape
 
 
+@pytest.mark.skip(reason="This test keeps failing on my computer")
```
jni: But does this fail on Travis? Isn't this the test that was fixed by https://github.com/scikit-image/scikit-image/pull/3063/files ? Are you sure you are on the latest version of master?

.travis.yml
hmaarrfk: Well, I guess that relates to the philosophy of what `scikit-image` is. Is it a plotting and visualization guide or an image processing library?

Personally, I don't think that matplotlib should even be a dependency. But I see where things like "viewers" and other things might be useful.

I'm ok with increasing minimum version requirements, but I feel like that is because I am inexperience and haven't had the (mis)forture of dealing with legacy software that cannot be updated.

.travis.yml
hmaarrfk: I was trying to get little helpful timings to show up:

image
I"m not sure if it is a bug with travis or not. but the "script" didn't have those cool folding sections.

hmaarrfk: They do some some kind of timing per command, but it isn't nearly as helpful.

I'll change it back to the script.

https://travis-ci.org/scikit-image/scikit-image/jobs/379492893#L4072

.travis.yml
jni: I think @efiring is right: it's hard to imagine a situation where you can't update MPL past 1.3 but you can update skimage to 0.14. In my view, conda environments have pretty much solved the update problem, because any user can install up-to-date dependencies in user-space.

More generally, we could have a calendar-based guide: we will support dependency versions up to three years old, for example.

In either view, 1.3 should be updated to either 1.5 or 2.0.

Regarding removing the dependency altogether... Well, now that we are doing separate dependencies for documentation/examples and the library, this is a possibility. But that's a big job, as we would need to extricate the viewer from the library. I think this is a worthwhile goal... For skimage 1.0. For now, let's just bump up the dependency. =D

skimage/io/tests/test_pil.py
hmaarrfk: Ok, I added this test:
```python
def test_img_as_grey_flatten():
    img = imread(os.path.join(data_dir, 'color.png'), as_gray=True)
    with expected_warnings(['deprecated']):
        img_flat = imread(os.path.join(data_dir, 'color.png'), flatten=True)
    assert_array_equal(img, img_flat)
```

skimage/viewer/tests/test_widgets.py
```diff
@@ -88,6 +88,7 @@ def test_slider_float():
     assert_almost_equal(sld.val, 2.5, 2)
 
 
+@testing.skipif(True, reason="Don't know how to automate skipping.")
```
jni: Can you elaborate on this? LOL I don't understand why the previous skipif was insufficient?

skimage/morphology/tests/test_misc.py
hmaarrfk: Was still learning to use `expected_warnings`.

tools/travis/install_qt.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+set -ex
+
+if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
+    echo "backend : Template" > $MPL_DIR/matplotlibrc
+fi
+# Now configure Matplotlib to use Qt4
+if [[ "${QT}" == "PyQt4" ]]; then
```
jni: I vote for deprecating Qt4 altogether

jni: Actually deprecating is the wrong word. Removing. Nuking. Obliterating. =P

skimage/morphology/tests/test_skeletonize_3d.py
hmaarrfk: You know, I really don't know why.

I guess for that file I copy pasted the warning statement from an other function that ignored warnings that way.


Fixed.

jni:
@hmaarrfk thanks for a gargantuan effort! When these commits first started coming in, I was wondering whether a lot of the work was even necessary. Having read through the changes, I think almost all of them are needed. 

Regarding NumPy printing: are you sure you are not on an outdated master? That should have been fixed in #2935 and #2975.

I've made a few comments to address.

To finish your task, [this comment](https://github.com/scikit-image/scikit-image/issues/3034#issuecomment-385237656) will probably be very useful.

How good is your git-fu? There are many changes here, which ideally would be grouped together logically and organised into a nice commit history:
- handle all the warnings by either wrapping in `expected_warnings` or passing the correct kwarg explicitly
- remove Qt
- bump matplotlib

However I'm happy to squash if that becomes too hard. But you should rebase on the latest master, for sure.

Where are you based? Getting this in is a high priority and I would like to work with you to get it merged. It might be easier if we chat over gitter. My timezone is Melbourne, Australia.

Thanks again!

skimage/restoration/tests/test_denoise.py
hmaarrfk: I'm pretty sure for that one, it gives a warning on 3.X and not on 2.7. Probably because of a divide by 0 or Nan rule.

hmaarrfk:
Writing it here so I don't forget.

What do you want "novice" to do when you resize an image. `mode='constant'` `anti_aliasing=False`

skimage/segmentation/tests/test_random_walker.py
hmaarrfk: Sorry, I was between flights, my internet access was slow. good catch. that comment was there for you to find =D

skimage/viewer/tests/test_widgets.py
hmaarrfk: This is issue #3081 

The previous timer would close the QT window after 100 ms. But it doesn't seem to be working on my computer. Testing on travis is slow. I'll keep this on my todo after addressing your other comments.

tools/travis/install_qt.sh
hmaarrfk: Yeah probably best. I just didn't want to impose things you know.

jni: Absolutely. For giant changes like this one, sometimes you can ask the mailing list or on the PR directly about what to do. It sometimes helps to avoid unnecessary work. But you have pushed things along very quickly by just forging ahead, so I very much appreciate that! =)

.travis.yml
hmaarrfk: 1.5 didn't work if i recall. I'll try 2.0. I could setup docker to test. But lets see how many commits it takes me.

jni:
@hmaarrfk was that last question a note to self or a question for me? At any rate `novice` altogether is deprecated so just choose whatever makes the tests pass silently!

.travis.yml
hmaarrfk: But it doesn't work really. The script one doesn't fold as nicely. I don't know why.

hmaarrfk:
OK. tests pass silently for novice. I just kept the old behaviour.

I actually went in an made the resize warnings raise exceptions (temporarily) and found a few tests that had put the "expected warning" to silently suppress a default behaviour change. Because they are test cases and were probably written with a specific behaviour, I explicitly set them to the old default.

.gitignore
```diff
@@ -9,9 +9,13 @@ __pycache__
 *.pyd
 *.bak
 *.c
+skimage/feature/_haar.cpp
 *.new
 *.md5
 *.old
+.pytest_cache
+temp.tif
+.ropeproject
```
jni: What's going on here?

.gitignore
```diff
@@ -9,9 +9,13 @@ __pycache__
 *.pyd
 *.bak
 *.c
+skimage/feature/_haar.cpp
```
jni: Did you also have to build this file manually? I had to do so on my machine, but it appeared to be working on Travis so I thought it was just a quirk of my own install...!

.travis.yml
```diff
@@ -21,6 +21,10 @@ addons:
     - texlive-latex-extra
     - dvipng
     - python-qt4
+    # - python3-qt4
+    # - python3-qt5
+    # - python3-pyside
```
jni: What's the deal here? We don't leave comments except in exceptional circumstances.

.travis.yml
```diff
@@ -36,26 +40,25 @@ notifications:
 
 matrix:
   include:
+    # 2.7 build (until 0.14 is release)
```
jni: Rebasing #3000 on this is going to be interesting LOL

.gitignore
hmaarrfk: no, i just did `python setup.py develop`. I didn't know if we had other cpp code. so I didn't want to blanket ignore them.

.travis.yml
hmaarrfk: That was a WIP. will remove. I was trying to install pyqt, old pyqt needed to be installed on the system. New pyqt can just be in a virtualenv.

.gitignore
hmaarrfk: `.ropeproject`

 is some atom thing. I'll remove it if you don't want it.
`.pytest_cache` is for running tests on your own PC. gets created I guess.

`temp.tif` is left behind by some test.

.travis.yml
```diff
@@ -72,12 +75,61 @@ before_install:
     - tools/check_sdist.py
 
 install:
-    - section build
     - python setup.py develop
-    - section_end build
+    # Matplotlib settings - do not show figures during doc examples
+    # Where is the matplotlib home on a mac?
```
jni: It appears to be in .matplotlib

.travis.yml
```diff
@@ -72,12 +75,61 @@ before_install:
...
+                exit 1
+            fi
+        done
+        mv $MPL_DIR/matplotlibrc_backup $MPL_DIR/matplotlibrc
+      fi
 
 after_success:
     - codecov
+    # Prepare.release
+    - doc/release/contribs.py HEAD~10
```
jni: What's this for?

skimage/color/colorconv.py
```diff
@@ -94,15 +94,16 @@ def guess_spatial_dimensions(image):
 def convert_colorspace(arr, fromspace, tospace):
     """Convert an image array to a new color space.
 
+    Valid color spaces include:
+        'RGB', 'HSV', 'RGB CIE', 'XYZ', 'YUV', 'YIQ', 'YPbPr', 'YCbCr', 'YDbDr'
```
jni: include -> are

wrap this line to 80 chars since we aren't in a specially-formatted section now

.travis.yml
hmaarrfk: Was in the script

https://github.com/scikit-image/scikit-image/blob/51f598aaedc73ef180913c670d2c20a8032aaf1e/tools/travis/script.sh#L121

skimage/io/tests/test_pil.py
```diff
@@ -45,18 +45,26 @@ def test_png_round_trip():
...
     Ip = img_as_float(imread(fname))
     os.remove(fname)
     assert np.sum(np.abs(Ip-I)) < 1e-3
 
 
+def test_img_as_grey_flatten():
+    img = imread(os.path.join(data_dir, 'color.png'), as_gray=True)
+    with expected_warnings(['deprecated']):
+        img_flat = imread(os.path.join(data_dir, 'color.png'), flatten=True)
+    assert_array_equal(img, img_flat)
```
jni: very nice! Although rename `as_grey_flatten` to `as_gray_flatten`. ;)

.travis.yml
hmaarrfk: Removed question.

skimage/morphology/tests/test_misc.py
jni: Sure! But please update so that it's PEP8 friendly and the noqa is unnecessary. =)

skimage/restoration/tests/test_denoise.py
```diff
@@ -223,8 +223,13 @@ def test_denoise_bilateral_multidimensional():
 
 
 def test_denoise_bilateral_nan():
+    import sys
     img = np.full((50, 50), np.NaN)
-    out = restoration.denoise_bilateral(img, multichannel=False)
+
+    # TODO: This warning is not optional in python3. This should be
+    # made a strict warning when we get to 0.15
```
jni: Add this to `TODO.txt` under 0.15.

skimage/restoration/tests/test_unwrap.py
```diff
@@ -116,6 +117,8 @@ def check_wrap_around(ndim, axis):
 dim_axis = [(ndim, axis) for ndim in (2, 3) for axis in range(ndim)]
 
 
+@skipif(sys.version_info[:2] == (3, 4),
+        reason="Doesn't with python 3.4. See issue #3079")
```
jni: doesn't work with

skimage/color/colorconv.py
hmaarrfk: I'll have you know, that it is exactly 79 characters long ;)
image

skimage/io/tests/test_pil.py
hmaarrfk: done

skimage/restoration/tests/test_denoise.py
hmaarrfk: done.

skimage/viewer/tests/test_widgets.py
jni: ping. ;) If you can't figure this out quickly, then write a better error message, something like "see issue #3081" ;)

skimage/restoration/tests/test_unwrap.py
hmaarrfk: done.

skimage/color/colorconv.py
jni: Touché!!! LOL!

hmaarrfk:
Thanks! This was a long one, but I really wanted tests to start passing with examples on OSX. I hacked the install of `scikit-learn`

jni:
Fascinating! Just saw that commit, works for me. Probably the right way is to have another requirements: examples.txt, and docs.txt is just sphinx, numpydoc, and sphinx-gallery and implies examples.txt. But we can leave that till another PR.

Don't forget to push your latest changes... ;) Maybe you're waiting for the present build to finish, which is fine. =)

hmaarrfk:
Yeah waiting for the results of OSX. The changes were to comments.

.travis.yml
stefanv: We have this in a script because editing it is much less painful than updating a YAML list of commands.

.travis.yml
```diff
@@ -36,26 +37,25 @@ notifications:
...
-      env: WITH_PYAMG=1 MINIMUM_REQUIREMENTS=1 WITH_QT=1
+      env: MINIMUM_REQUIREMENTS=1
     - os: linux
       python: 3.4
-      env: OPTIONAL_DEPS=1 WITH_PYSIDE=1 BUILD_DOCS=1 DEPLOY_DOCS=1
-    - os: linux
-      python: 3.5
     - os: linux
       python: 3.5
-      env: PIP_FLAGS="--pre"
```
jni: @hmaarrfk, @stefanv makes a good point: at least one of our builds should use `--pre`... I suggest 3.6.

.travis.yml
hmaarrfk: @stefanv yes editing the YAML was painful. I changed it back.

skimage/morphology/tests/test_misc.py
hmaarrfk: I guess I had the same line in two locations. Thanks for spotting this one.

tools/travis/install_qt.sh
hmaarrfk: Do you mean use something like `qtpy`?

requirements/default.txt
```diff
@@ -1,9 +1,9 @@
 numpy>=1.11
 scipy>=0.17.0
-matplotlib>=1.3.1
+matplotlib>=2.0.0
```
tacaswell: :tada: I strongly approve of this change!

hmaarrfk:
@stefanv @jni I want to bring to your attention that in the proposed version, Qt and all optional dependencies are installed before any tests are run (with the exception of the doc dependencies, but now that I think those should also be moved to the install section).

The old strategy of
1. Test with minimum dependencies
2. Install more things
3. Test again

has been removed in favour of: Just ask travis for two builds.

tools/travis/install_qt.sh
tacaswell: One data point on this is that riverbank have pulled all the pyqt4 files from pypi (https://pypi.org/project/PyQt4/).

.travis.yml
```diff
@@ -36,26 +37,25 @@ notifications:
 
 matrix:
   include:
+    # 2.7 build (until 0.14 is release)
     - os: linux
       python: 2.7
-      env: PIP_FLAGS="--pre"
+      env: WITH_PYAMG=1 QT=PyQt4
```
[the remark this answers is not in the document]
hmaarrfk: I'll add a 2.7 and 3.6 build that use pre. The only reference to PIP_FLAGS I found was in the [before_install.sh](https://github.com/scikit-image/scikit-image/blob/51f598aaedc73ef180913c670d2c20a8032aaf1e/tools/travis/before_install.sh#L68)  is that correct?

.travis.yml
hmaarrfk: Yeah, I didn't understand why were were testing for `--pre`.

I was mostly trimming builds to test quickly.

For now, I also added it to 2.7

hmaarrfk:
FYI, appveyor file states the following:

```
  # Install the build and runtime dependencies of the project.
  # The --pre flag is necessary to grab a SciPy wheel, which is in
  # pre-release at the time of writing (03-10-2017)
```
Therefore do we still need `--pre` builds?

matthew-brett:
The `--pre` builds are to test against not-yet released versions of the dependencies, to make it less likely that the scikit-image code will break with new releases of say numpy or scipy, and to give those projects feedback when they break scikit-image code.

Just for example, statsmodels didn't have a `--pre` entry, released, and the next day pandas did a release which broke stuff in statsmodels.

hmaarrfk:
@matthew-brett Thanks for the explination. 

Ok, I don't understand what is happening with Appveyor.
Half the time it refuses to find pip packages.

tools/travis/install_qt.sh
hmaarrfk: @tacaswell , it seems that conda shipped pyqt5.6 for 2.7 but not pyqt5.9.
Seeing as 0.14 is the last release to officially support 2.7, I think we are good :D.

stefanv:
Thank you @hmaarrfk! Not having a working Travis is very disruptive to development.

tools/travis/install_qt.sh
stefanv: If Riverbank pulled 4, then 5-only seems very reasonable.

hmaarrfk:
@stefanv yes, it was, I never knew what was actually failing.

Taught alot about how to use pytest in the meantime :dango:

Issue #3073, opened 2018-05-14
skimage.measure.marching_cubes_lewiner, spacing expects tuple


kmader:
## Description

The `marching_cubes_lewiner` spacing argument is expected to be a tuple, but it is never explicitly checked and returns an unhelpful error message. This is particularly an issue because when dealing with other libraries like SimpleITK the array dimensions are often returned as a ndarray and ndarrays are much easier to scale if using functions like ```scipy.ndimage.zoom```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-74-b53599eae95f> in <module>()
      2 import numpy as np
      3 
----> 4 verts, faces, normals, values = measure.marching_cubes_lewiner(np.random.uniform(0, 1, size = (5,5,5)), 0.5, spacing = np.ones(3))

/opt/conda/lib/python3.6/site-packages/skimage/measure/_marching_cubes_lewiner.py in marching_cubes_lewiner(volume, level, spacing, gradient_direction, step_size, allow_degenerate, use_classic)
    230         raise ValueError("Incorrect input %s in `gradient_direction`, see "
    231                          "docstring." % (gradient_direction))
--> 232     if spacing != (1, 1, 1):
    233         vertices = vertices * np.r_[spacing]
    234 

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
## Way to reproduce
```
from skimage import measure
import numpy as np
verts, faces, normals, values = measure.marching_cubes_lewiner(np.random.uniform(0, 1, size = (5,5,5)), 0.5, spacing = np.ones(3))
```

kmader:
classic works fine
```
from skimage import measure
import numpy as np
verts, faces = measure.marching_cubes_classic(np.random.uniform(0, 1, size = (5,5,5)), 0.5, spacing = np.ones(3))
```

PR #3074, opened 2018-05-14
Sanitizing marching_cubes_lewiner spacing input argument

kmader:
convert an np.array to a tuple

## Description
Fixing the problem if you feed a np.ndarray to the spacing argument of `skimage.measure.marching_cubes_lewiner` https://github.com/scikit-image/scikit-image/issues/3073

## Checklist
- [x] Unit tests

## References
Fixes https://github.com/scikit-image/scikit-image/issues/3073

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_marching_cubes_lewiner.py
+++ b/skimage/measure/_marching_cubes_lewiner.py
@@ -134,7 +134,7 @@ def marching_cubes_lewiner(volume, level=None, spacing=(1., 1., 1.),
         if level < volume.min() or level > volume.max():
             raise ValueError("Surface level must be within volume data range.")
     # spacing
-    if isinstance(spacing, np.ndarray): # it should be a tuple
+    if isinstance(spacing, np.ndarray):  # it should be a tuple
         spacing = tuple(spacing)
     if len(spacing) != 3:
         raise ValueError("`spacing` must consist of three floats.")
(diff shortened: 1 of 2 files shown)


skimage/measure/_marching_cubes_lewiner.py
```diff
@@ -134,6 +134,8 @@ def marching_cubes_lewiner(volume, level=None, spacing=(1., 1., 1.),
         if level < volume.min() or level > volume.max():
             raise ValueError("Surface level must be within volume data range.")
     # spacing
+    if isinstance(spacing, np.ndarray):  # it should be a tuple
```
soupault: Isn't it simpler to case to ndarray instead?

kmader: possible, I was just looking for the smallest change to get the functionality

soupault: I mean `spacing` could be a list as well. I think it is better to be more generic. Good catch, anyway!

kmader: sure, it did work with lists before (but for the wrong reasons), I'll fix it

JDWarner:
Thanks @kmader, simple and straightforward!  Happy to merge.

There may be opportunity for very similar PRs in other places in the package - take a look at other 3D capable functions which accept a `spacing` kwarg.  If the sanitization of these varies, the framework may be able to be exported to a shared, non-public function in `skimage._shared` or similar.

Issue #3075, opened 2018-05-14
convex_hull_image of empty image returns error


kmader:
## Description
Often it is necessary to run convex_hull on every slice of a 3D image, this does not work since if the image is empty skimage throws an error message about invalid division. Ideally it would return an empty boolean image as the convex hull of nothing is nothing?

## Way to reproduce
```py
from skimage.morphology import convex_hull_image
convex_hull_image(np.zeros((5,5)))
```

```
/opt/conda/lib/python3.6/site-packages/skimage/morphology/convex_hull.py:61: RuntimeWarning:

Mean of empty slice.

/opt/conda/lib/python3.6/site-packages/numpy/core/_methods.py:73: RuntimeWarning:

invalid value encountered in true_divide

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-86-1a3c85daeeaa> in <module>()
      1 from skimage.morphology import convex_hull_image
----> 2 convex_hull_image(np.zeros((5,5)))

/opt/conda/lib/python3.6/site-packages/skimage/morphology/convex_hull.py in convex_hull_image(image)
     63 
     64     # Find the convex hull
---> 65     chull = Delaunay(coords).convex_hull
     66     v = coords[np.unique(chull)]
     67 

qhull.pyx in scipy.spatial.qhull.Delaunay.__init__()

qhull.pyx in scipy.spatial.qhull._Qhull.__init__()

ValueError: No points given
```

soupault:
Closed via #3076

PR #3076, opened 2018-05-14
Allow convex_hull_image on empty images

kmader:
## Description
changing convex_hull_image behavior to show a warning and return an empty image if given an empty image

## Checklist
- [x] Unit tests

## References
Closes issue: https://github.com/scikit-image/scikit-image/issues/3075

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/convex_hull.py
+++ b/skimage/morphology/convex_hull.py
@@ -49,10 +49,10 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
     """
     ndim = image.ndim
     if not np.max(image):
-        # if there are no non-zero points print a warning and return an empty image
-        warn("Input image is entirely zero and does not have a valid convex hull"
+        # if image is empty print a warning and return an empty image
+        warn("Input image is entirely zero, no valid convex hull"
              "Returning empty image", UserWarning)
-        return np.zeros(image.shape, dtype = bool)
+        return np.zeros(image.shape, dtype=bool)
     # In 2D, we do an optimisation by choosing only pixels that are
     # the starting or ending pixel of a row or column.  This vastly
     # limits the number of coordinates to examine for the virtual hull.
--- a/skimage/morphology/convex_hull.py
+++ b/skimage/morphology/convex_hull.py
@@ -6,6 +6,7 @@ from ..measure.pnpoly import grid_points_in_poly
 from ._convex_hull import possible_hull
 from ..measure._label import label
 from ..util import unique_rows
+from .._shared.utils import warn
 
 __all__ = ['convex_hull_image', 'convex_hull_object']
 
@@ -47,7 +48,11 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
 
     """
     ndim = image.ndim
-
+    if not np.max(image):
+        # if there are no non-zero points print a warning and return an empty image
+        warn("Input image is entirely zero and does not have a valid convex hull"
+             "Returning empty image", UserWarning)
+        return np.zeros(image.shape, dtype = bool)
     # In 2D, we do an optimisation by choosing only pixels that are
     # the starting or ending pixel of a row or column.  This vastly
     # limits the number of coordinates to examine for the virtual hull.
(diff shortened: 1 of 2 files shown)


skimage/morphology/convex_hull.py
```diff
@@ -47,7 +48,11 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
 
     """
     ndim = image.ndim
-
+    if not np.max(image):
+        # if image is empty print a warning and return an empty image
```
soupault: This comment looks redundant to me. Warning message explains everything well.

skimage/morphology/convex_hull.py
```diff
@@ -47,7 +48,11 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
...
+        warn("Input image is entirely zero, no valid convex hull"
```
soupault: `hull"` -> `hull. "`

skimage/morphology/convex_hull.py
```diff
@@ -47,7 +48,11 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
...
+             "Returning empty image", UserWarning)
+        return np.zeros(image.shape, dtype=bool)
```
soupault: I'd use `np.bool_` here.

skimage/morphology/convex_hull.py
```diff
@@ -47,7 +48,11 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
 
     """
     ndim = image.ndim
-
+    if not np.max(image):
```
soupault: Could you add a test case covering this change?

kmader: done

soupault: Hmm... On a second though this could work wrong with intXX or floatXX images with non-positive values. Should we have a stricter check here? Otherwise, LGTM.

/offtopic LOL, I've just noticed that we have a series of lies in the docstrings: https://github.com/scikit-image/scikit-image/blob/master/skimage/morphology/convex_hull.py#L29 (actually, isn't cast);  https://github.com/scikit-image/scikit-image/blob/master/skimage/morphology/_convex_hull.pyx#L10, https://github.com/scikit-image/scikit-image/blob/master/skimage/morphology/_convex_hull.pyx#L15 (not boolean).

skimage/morphology/convex_hull.py
```diff
@@ -47,7 +48,10 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
 
     """
     ndim = image.ndim
-
+    if np.max(image)<1:
```
soupault: Sorry, I wasn't clear enough in my previous comment. I was saying that this statement will not work properly if user provides, for example, an array of `int8` with values `-1` and `0` only. I'd expect the function to draw a convex hull around `-1` elements, but in this case it'll say that the image is zero.
I don't like that we actually limit input dtypes to unsigned for 2D arrays several lines below, but this is another story. For 1D/3+D negatives values should matter.

How about `if np.count_nonzero(image) > 0:`?

soupault: >I don't like that we actually limit input dtypes to unsigned for 2D arrays several lines below

Ignore this :) .
```
In [6]: x = np.array([-1, 0], dtype=np.int8)

In [7]: x.astype(np.uint8)
Out[7]: array([255,   0], dtype=uint8)
```

kmader:
Ah I grew up in a world of unsigned ints for masks and labels and never dreamed of using -1 as a label id but count_nonzero definitely handles that better

soupault:
@kmader everything is possible in Python :).

skimage/morphology/tests/test_convex_hull.py
```diff
@@ -26,6 +26,26 @@ def test_basic():
     assert_array_equal(convex_hull_image(image), expected)
 
 
+def test_empty_image():
+    image = np.array(
```
stefanv: `image = np.zeros((6, 6), dtype=bool)`

kmader: I had just copied the format before but zeros is easier

jni:
@kmader nice and clean! Thank you, this is a great usability improvement!

PR #3077, opened 2018-05-14
Proposed changes to `skimage.transform._geometry`

hmaarrfk:
# Proposed changes to `skimage.transform._geometry`

I would like to propose a few small breaking changes to the transform API.
Theses proposals turned into an implementation during a long flight.

1. I would like to make `inverse` a property that returns an instance of the same class (if possible).
2. I would like to deprecate the `__add__` API in favour of the `__matmul__` api. I guess we could use `dot` for 2.7, but I find `dot` confusing because there are so many `dot` products out there.
3. Further, I would like to reverse the order of the operands in `__matmul__` compared to the current implementation to match the linear algebraic language.

Since this requires the `@` operator which is python 3 only, I think this should just wait until python 2.7 support is dropped.

## 1. Changes to the behaviour of `inverse`
Currently, calling `.inverse` doesn't actually return a transform. It returns a function that maps the outputs to the inputs. This is fine, but I think it leads to a little bit of confusion. A lot of the notation in the `transform` API uses matrix multiplication language. You would think that things like the identity would be well defined:

For example, if the matrix $A$ is the mathematical representation of the `SimilarityTransform` we are working with, we would expect the following two statements to hold for the identity

In math
$$
I = A^{-1}A
I = A A^{-1}
$$
And in python, you would wish that
```python
I = A.inverse @ A
I = A @ A.inverse
```
Currently, only one of them actually works (due to some hacky code IMO see below)

```python
    def __add__(self, other):
        [...]
        elif (hasattr(other, '__name__')
               and other.__name__ == 'inverse'
               and hasattr(get_bound_method_class(other), '_inv_matrix')):
           return ProjectiveTransform(other.__self__._inv_matrix.dot(self.params))
        [...]
```

~~I understand that this would require a small change to the way `warp` works too.~~ I implemented the necessary changes. There might be a few more lines to delete in `warp` that I marked with a TODO

I suggest:

  1. For `GeometricTransforms` `inverse` returns the mathematical inverse of the original `GeometricTransform`
  2. ~~The old function can be retained as `inverse_map` matching the naming convention established in the `warp` function.~~ This isn't necessary since you can simply call `t.inverse(coords)` just like you used to since the `GeometricTransforms` are callable themselves.

## 2. Deprecating `__add__` in favour of `__matmult__`
The language used to describe most of these transformations seems to be that of linear algebra. My limited knowledge of linearly algebra states that addition is commutative. That said, in this case, addition is **not** commutative.

In scikit-image, if `B` and `A` are `SimilarityTransform`, `B + A` and `A + B` lead to different results which is not what linear algebra tells us.

Using `__matmul__` makes it much cleaner that the `combination` of the transforms is not commutative since it is known that matrix multiply is not commutative.

I suggest
a. We cause the `__add__` function to raise a Warning (if people want to shoot themselves in the foot, that is their choice for at least for for the first version this is accepted in).
b. We implement the `__matmut__` function as described below.

## 3. Reverse the order of the operands in `__matmult__` compared to `__add__`

Current, if `A` and `B` are `SimilarityTransform`, then
```python
C = A + B
# or as 2 suggest
C = A @ B
```
leads to the mathematical operation

$$
x_prime = Cx = BAx
$$
and not
$$
x_prime = Cx = ABx = A(B(x))
$$

I don't think this it makes sense to swap the order of the matricies. Therefore, I suggest that the python operation, matches the linear algebraic one too.

## Checklist
- [x] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -163,7 +163,11 @@ class GeometricTransform(object):
         """
         raise NotImplementedError()
 
-    def inverse(self, coords):
+    def forward_map(self, coords):
+        __doc__ = self.__call__.__doc__
+        return self(coords)
+
+    def inverse_map(self, coords):
         """Apply inverse transformation.
 
         Parameters
@@ -245,6 +249,10 @@ class FundamentalMatrixTransform(GeometricTransform):
             raise ValueError("Invalid shape of transformation matrix")
         self.params = matrix
 
+    @property
+    def inverse_params(self):
+        return self.params.T
+
     def __call__(self, coords):
         """Apply forward transformation.
 
@@ -262,7 +270,7 @@ class FundamentalMatrixTransform(GeometricTransform):
         coords_homogeneous = np.column_stack([coords, np.ones(coords.shape[0])])
         return coords_homogeneous @ self.params.T
 
-    def inverse(self, coords):
+    def inverse_map(self, coords):
         """Apply inverse transformation.
 
         Parameters
@@ -277,7 +285,7 @@ class FundamentalMatrixTransform(GeometricTransform):
 
         """
         coords_homogeneous = np.column_stack([coords, np.ones(coords.shape[0])])
-        return coords_homogeneous @ self.params
+        return np.dot(coords_homogeneous, self.inverse_params.T)
 
     def _setup_constraint_matrix(self, src, dst):
         """Setup and solve the homogeneous epipolar constraint matrix::
@@ -539,7 +547,7 @@ class ProjectiveTransform(GeometricTransform):
         self.params = matrix
 
     @property
-    def _inv_matrix(self):
+    def inverse_params(self):
         return np.linalg.inv(self.params)
 
     def _apply_mat(self, coords, matrix):
@@ -571,7 +579,7 @@ class ProjectiveTransform(GeometricTransform):
         """
         return self._apply_mat(coords, self.params)
 
-    def inverse(self, coords):
+    def inverse_map(self, coords):
         """Apply inverse transformation.
 
         Parameters
@@ -585,7 +593,7 @@ class ProjectiveTransform(GeometricTransform):
             Source coordinates.
 
         """
-        return self._apply_mat(coords, self._inv_matrix)
+        return self._apply_mat(coords, self.inverse_params)
 
     def estimate(self, src, dst):
         """Estimate the transformation from a set of corresponding points.
@@ -703,9 +711,9 @@ class ProjectiveTransform(GeometricTransform):
                 tform = ProjectiveTransform
             return tform(other.params @ self.params)
         elif (hasattr(other, '__name__')
-                and other.__name__ == 'inverse'
-                and hasattr(get_bound_method_class(other), '_inv_matrix')):
-            return ProjectiveTransform(other.__self__._inv_matrix @ self.params)
+                and other.__name__ == 'inverse_map'
+                and hasattr(get_bound_method_class(other), 'inverse_params')):
+            return ProjectiveTransform(other.__self__.inverse_params.dot(self.params))
         else:
             raise TypeError("Cannot combine transformations of differing "
                             "types.")
@@ -902,7 +910,7 @@ class PiecewiseAffineTransform(GeometricTransform):
 
         return out
 
-    def inverse(self, coords):
+    def inverse_map(self, coords):
         """Apply inverse transformation.
 
         Coordinates outside of the mesh will be set to `- 1`.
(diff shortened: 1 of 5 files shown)


skimage/transform/_geometric.py
```diff
@@ -903,7 +876,11 @@ def __call__(self, coords):
 
         return out
 
-    def inverse(self, coords):
+    @property
+    def inverse(self):
+        raise NotImplementedError()
```
kne42: Please include a message that specifies that `inverse_map` has replaced this functionality.

skimage/transform/_warps.py
```diff
@@ -831,8 +834,9 @@ def warp(image, inverse_map, map_args={}, output_shape=None, order=1,
             matrix = inverse_map.params
 
         elif (hasattr(inverse_map, '__name__') and
-              inverse_map.__name__ == 'inverse' and
+              inverse_map.__name__ == 'inverse_map' and
               get_bound_method_class(inverse_map) in HOMOGRAPHY_TRANSFORMS):
+            # TODO: do we still need this introspection hack?
```
kne42: This case is no longer needed since both `inverse` and `inverse_map` are properties that return `GeometricTransform` and `np.ndarray` respectively, which are covered by the two previous cases.

skimage/transform/_geometric.py
```diff
@@ -164,22 +164,18 @@ def __call__(self, coords):
...
-
-        """
+    @property
+    def inverse_params(self):
         raise NotImplementedError()
 
+    @property
+    def inverse(self):
+        """Return the inverse transformation."""
+        # ASK: Can you make a property a class method too?
```
kne42: If you're referring to [_warps.py:830](https://github.com/scikit-image/scikit-image/pull/3077/files?utf8=%E2%9C%93&diff=unified#diff-c0f33f5600401be736e79b65164b4c81R839), then as explained, this shouldn't be an issue.

kne42 (commented):
@hmaarrfk Excellent work! I definitely agree with a lot of your changes, but am hesitant with respect to jumping on board with the `__matmult__`/`@` train for our operator just yet.

This is due to its incompatibility with Python 2.7, which would require users to write out the full `__matmult__` function resulting in messy//inelegant code.

Instead, I propose that we use the Python 2-compatible `__mul__`/`*` operator, or just wait a week or two until `0.14` is released, after which Python 2 support will be dropped (see #3000).

P.S. LaTeX doesn't work in Github Markdown. I suggest using [this tool](https://www.codecogs.com/latex/eqneditor.php) instead.

hmaarrfk:
@kne42, I am more than OK to wait for 2.7 to be dropped so that the changes can follow the lines of linear algebra more closely. I'm already on my own fork of scikit-image for my work, so I don't think this is going to affect me.

I think `*` would lead to more confusion since for numpy, `*` means element wise multiplication, and I'm sure somebody thought of what the Hadamard product means for geometric transforms. I'll leave that part to somebody that understand more linear algebra!

skimage/transform/_warps.py
kne42: @hmaarrfk Disregard what I said about removing this case; I forgot that `inverse_map` is actually a function so this still applies.

However, to answer your question, you shouldn't need this introspection hack any more. I'll explain in additional comments.

skimage/transform/_warps.py
```diff
@@ -831,8 +834,9 @@ def warp(image, inverse_map, map_args={}, output_shape=None, order=1,
...
             # inverse_map is the inverse of a homography
             matrix = np.linalg.inv(six.get_method_self(inverse_map).params)
```
kne42: In Python 3, to get the bound method's instance, use `method.__self__`. Additionally, call `inverse_map.__self__.inverse` here instead. Don't forget to remove the `six` import up top!

kne42 (approved):
LGTM! Don't forget to remove Python 2 compatibility code if you do opt to wait until Py2 support is dropped.

skimage/transform/_warps.py
hmaarrfk: Ok, I think I get it. Warping is trying to get at the mathematical representation of the transform and not the functional one to speed things up.

This code wouldn't even work with PeicewiseAffine, which is the last case I was considering.

hmaarrfk:
What is the consensus on this PR? 

I think the change to the behaviour of inverse is legitimately an improvement.
The notation benefits of the `@` operator are good on a mathematical level.

I can remove the deprecation warning on `+`, and I think then the main objections are no longer an issue.

matthew-brett:

<details>

## Rationale

</details>

PR #3087, opened 2018-05-22
0.13.2: Version bump for new build

hmaarrfk:
## Description
@stefanv @jni could you also please create a tag for this? That way we can test the recipe?

## References
[feedstock](https://github.com/conda-forge/scikit-image-feedstock/pull/17)
Change [this](https://github.com/conda-forge/scikit-image-feedstock/blob/7f2190d3940f7607b30fe432dee4925a66f4a7e8/recipe/meta.yaml#L1) line in the recipe after we create the tag.

## For reviewers

  later.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -64,7 +64,7 @@ test_script:
   -  echo backend:Agg > matplotlibrc
 
   # Run unit tests with pytest
-  - "python -c \"import pytest; import skimage; pytest.main([skimage.pkg_dir])\""
+  - pytest -v --pyargs skimage
 
 artifacts:
   # Archive the generated wheel package in the ci.appveyor.com build report.
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -65,7 +65,7 @@ import sys
 pkg_dir = osp.abspath(osp.dirname(__file__))
 data_dir = osp.join(pkg_dir, 'data')
 
-__version__ = '0.13.1'
+__version__ = '0.13.2'
 
 try:
     imp.find_module('nose')


jni:
Hmm, dang, the version number on the branch should have been v0.13.2-dev. Oh well. =)

I prefer to upload a proper release to PyPI before changing the conda-forge release. This will take a bit of time, but I can do that later today. Hold please. :phone:  :notes:

hmaarrfk:
I vote for making versions python compatible `v0.13.2.dev0` I forget if you need a `.` between `dev` and `0` too.

hmaarrfk:
Also, I was once able to automate uploading tags to PyPi upon sucessful build.

This is a really simple, python-only package though. [Travis Deployment to PyPi](https://github.com/colinoflynn/pico-python/blob/f83594f9c564fc89653e29bd3742dd64d111773a/.travis.yml#L14)

jni:
@hmaarrfk yeah we haven't set up automatic deployment, I believe purely for historical reasons. @stefanv may chime in here as to whether we want this. It would require changing our release procedure notes.

I didn't realise that Python had a prescriptive [release numbers PEP](https://www.python.org/dev/peps/pep-0440/)! Having said that, it looks like both `-` and `.` are acceptable *but* the canonical version is with no separator at all: `0.13.2dev`. See ["pre-release separators"](https://www.python.org/dev/peps/pep-0440/#id32). Personally I quite dislike this version...

hmaarrfk:
The Python BDFL have rules for everything!

stefanv:
Should that be `0.13.2dev0`?

W.r.t. uploading packages automatically, I think the fear is that PyPi does not allow re-uploading of a file ever again, so we want to double check that everything worked fine.  That said, if we automate it, and something goes wrong, we could always just increase the patch version.  We've never reached beyond `x.y.3`.

stefanv:
Woah, so much version confusion.  Looks like various options are allowed.  Best, to my eye, seems to be `0.13.2.dev0` (see https://www.python.org/dev/peps/pep-0440/).

jni:
@stefanv sounds good.

stefanv:
It looks like the test decorators like `skipif` will also have to be backported to make AppVeyor pass.

hmaarrfk:
Yeah. There are too many changes now to make this a “small upgrade”. I’m
out on the .13.2 front. I would rather devote my time to fixing the let few
bugs with .14.0

stefanv:
Agreed.  Let's do that.

PR #2165, opened 2016-06-21
Update radon_transform.py

jcesardasilva:
I have notice some small mistakes in the filters of the iradon in scikit-image/skimage/transform/radon_transform.py, in lines 208 up to 216. In some part where the division by 2 should exist, it does not, and vice-versa. See below the code lines with the corrections. 
# in scikit-image/skimage/transform/radon_transform.py, in lines 208 up to 216

```
elif filter == "shepp-logan":
    # Start from first element to avoid divide by zero
    fourier_filter[1:] = fourier_filter[1:] * np.sin(omega[1:]/2) / (omega[1:]/2)
elif filter == "cosine":
    fourier_filter *= np.cos(omega/2)
elif filter == "hamming":
    fourier_filter *= (0.54 + 0.46 * np.cos(omega))
elif filter == "hann":
    fourier_filter *= (1 + np.cos(omega)) / 2
elif filter is None:
    fourier_filter[:] = 1
```

This becomes more evident when using frequency cutoffs. 
# code

```
elif filter_type == "shepp-logan":
    # Start from first element to avoid divide by zero
    fourier_filter[1:] = fourier_filter[1:] * (np.sin(omega[1:]/(2*freqcutoff)) / (omega[1:]/(2*freqcutoff)))
elif filter_type == "cosine":
    fourier_filter *= np.cos(omega/(2*freqcutoff))
elif filter_type == "hamming":
    fourier_filter *= (0.54 + 0.46 * np.cos(omega / (freqcutoff)))
elif filter_type == "hann":
    fourier_filter *= (1 + np.cos(omega / (freqcutoff))) / 2
```

I also change the code to enable the reconstruction from differential projections like the ones taken with grating interferometer or to be insensitive to the big phase-wraps of ptychographic X-ray computed tomography dataset

PR #3099, opened 2018-05-23
iradon filter function

kczimm:
## Description
A continuation of #3067 to split the iradon filter creation into a separate function and provide tests for the filters.

## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.


skimage/transform/radon_transform.py
```diff
@@ -124,6 +124,64 @@ def _sinogram_circle_to_square(sinogram):
...
+    f[0] = 0.25
+    f[1::2] = -1 / (np.pi * n[1::2])**2
+
+    omega = 2 * np.pi * fftfreq(size)
+    kernel = 2 * np.real(fft(f))         # ramp filter
+    if filter_name == "ramp":
+        pass
+    elif filter_name == "shepp-logan":
+        # Start from first element to avoid divide by zero
+        kernel[1:] *= np.sin(omega[1:] / 2) / (omega[1:] / 2)
```
hmaarrfk: Should be [`sinc`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.sinc.html).

Since `sinc` is defined as sin(pi x) / (pi x), I suggest using `f` instead of omega. and doing
```
f = fftfreq(size)
kernel *= np.sinc(f)
```

emmanuelle:
Thanks, LGTM. With your improvements, is it possible to decrease the tolerance in the ``test_radon_transform`` module?

emmanuelle:
I suggest that we merge your PR first, and then see if we must also include changes from #2165.

kczimm:
@emmanuelle I created that tolerance to have a passing test. There is still some bias in the reconstruction but with this PR, it is significantly reduced. I cannot decrease the tolerance by an order of magnitude without the test failing. However, the error is small enough that after a user converts to Hounsfield units (which is a common use case in CT applications) the error will still be less than 0.1 HU bias using the ramp filter.

stefanv:
This is *so close* with only minor modifications needed before merge. @kczimm would you like to take a look, or should @scikit-image/core polish it up before merging?

kczimm:
@stefanv if @scikit-image/core can polish it up, I'm OK with that. This was so long ago that I did this that it might take me more effort to get back up to speed than the core team would need to finish it off. If it won't go through otherwise, I'll finish it up. It shouldn't stay in scikit-image broken.

stefanv:
@kczimm I apologize for letting this issue slide; thank you again for your work.

alexdesiqueira:
Hey @hmaarrfk, could you help me close this one? As @stefanv said, we're almost there (but I don't know what's happening here :sweat_smile:). Thanks!

hmaarrfk:
Because digging this one out took alot of merging, I decided to revert my changes to this PR (sorry for the force push), and move them to: https://github.com/scikit-image/scikit-image/pull/4950 so we can visually compare.

I think there were two goals for this PR:

1. exporting a public API function for iradon filters.
2. Reducing the error.

rfezzani:
@kczimm, the `iradon_filter` function that you defined is now implemented as the hidden function `_get_fourier_filter`. May be can you update your PR to add the `test_iradon_rampfilter_bias_circular_phantom` test (or it is may be easier to start a new PR)?
I would be very happy to accept it :wink:.
Thank you for your time!

stefanv:
I think this was quite an elegant refactoring, and surprisingly it has aged well. @kczimm is there any chance you want to pick it back up?

stefanv:
I missed that this work has already been included by @rfezzani. Thank you again @kczimm. I will grab the test from here and add that as a new PR.

stefanv:
See https://github.com/scikit-image/scikit-image/pull/6873

kczimm:
@stefanv thanks for doing this and sorry I'm just seeing this now.

Issue #3091, opened 2018-05-22
Multiple failures in 32-bit Linux builds


matthew-brett:
I'm very sorry not to be more detailed, but this is just to flag up that all scikit-image-wheels matrix entries have multiple test failures.  Here's a build that I think is correctly updated:

https://travis-ci.org/scikit-image/scikit-image-wheels/jobs/382142313

soupault:
@matthew-brett there are some failing I've never seen before, but the used codebase is, most likely, not the latest - `random_walker` tests should be passing since https://github.com/scikit-image/scikit-image/pull/3063. Could you try to clear the build cache on Travis and restart?

hmaarrfk:
i think the issue is that those wheels are using the tag [0.13.1](https://github.com/scikit-image/scikit-image-wheels/blob/be1556fcff1581f23c16a91a82f754ddf478f3cc/.travis.yml#L4), and daily seems to be on [0.13.0](https://github.com/scikit-image/scikit-image-wheels/blob/4aaf20bfb8db67eab519339a77718de0b58f88c0/.travis.yml#L5) which was broken a few days ago. 
We need a version bump and tag to allow 

See PR #3087
and #3085

matthew-brett:
Yes, sorry, my mistake.  Current build matrix is for scikit-image master:

https://travis-ci.org/scikit-image/scikit-image-wheels/builds/382260738

New errors, on Python 2 and 3, 32-bit : 

* watershed errors on 32-bit : #3089
* test_spacing_1 : #3092
* test_descriptor_orb : https://travis-ci.org/scikit-image/scikit-image-wheels/jobs/382260741#L3102, https://travis-ci.org/scikit-image/scikit-image-wheels/jobs/382260742#L2305
* TestRank.test_all : https://travis-ci.org/scikit-image/scikit-image-wheels/jobs/382260744#L2392, https://travis-ci.org/scikit-image/scikit-image-wheels/jobs/382260742#L2425
* test_ellipse_model_estimate_failers : https://travis-ci.org/scikit-image/scikit-image-wheels/jobs/382260744#L2589, https://travis-ci.org/scikit-image/scikit-image-wheels/jobs/382260742#L2596

hmaarrfk:
of course 32 bit has not been tested......

matthew-brett:
[...]

PR #3103, opened 2018-05-24
Ellipsis fitting: Check the condition number before computiing eigen vectors

hmaarrfk:
Under certain conditions, the result of a matrix we have to invert becomes ill conditioneed. The 32 bit solver and 64 bit solver converge on different answers.
We first check if that matrix is ill conditioned before we move forward with the computation.

See #3091

## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -434,6 +434,12 @@ class EllipseModel(BaseModel):
         except np.linalg.LinAlgError:  # LinAlgError: Singular matrix
             return False
 
+        # While S3 might be well conditioned enough to take an inverse
+        # M might not be well conditioned.
+        # Wikipedia says you loose log2(cond(M)) digits of precision
+        if math.log2(np.linalg.cond(M)) > 12:
+            return False
+
         # M*|a b c >=l|a b c >. Find eigenvalues and eigenvectors
         # from this equation [eqn. 28]
         eig_vals, eig_vecs = np.linalg.eig(M)
--- a/skimage/measure/tests/test_fit.py
+++ b/skimage/measure/tests/test_fit.py
@@ -225,11 +225,20 @@ def test_ellipse_model_estimate_from_data():
                'details: '
                'https://github.com/scikit-image/scikit-image/issues/3091 '
                'https://github.com/scikit-image/scikit-image/issues/2670'))
-def test_ellipse_model_estimate_failers():
+@testing.parametrize('epsilon', 10.**np.arange(-4, -20, -1))
+def test_ellipse_model_estimate_failers(epsilon):
+    # This test is known to return true on 32 bit machines with epsilon = 0
+    # Setting the value of espilon to a small number causes it return true
+    # on 64 machiens too.
+    # Unfortunately, it seems to depend on the version of python that is used
+    # Therefore, test for many cases for different values of epsilon.
+
     # estimate parameters of real data
     model = EllipseModel()
     assert not model.estimate(np.ones((5, 2)))
-    assert not model.estimate(np.array([[50, 80], [51, 81], [52, 80]]))
+    assert not model.estimate(np.array([[50, 80],
+                                        [51, 81],
+                                        [52, 80 + epsilon]]))
 
 
 def test_ellipse_model_residuals():


skimage/measure/fit.py
```diff
@@ -434,6 +434,11 @@ def estimate(self, data):
         except np.linalg.LinAlgError:  # LinAlgError: Singular matrix
             return False
 
+        # For some reason the above doesn't catch the error on linux 32 bit
+        # https://stackoverflow.com/questions/13249108/efficient-pythonic-check-for-singular-matrix
+        if np.linalg.cond(M) > 1 / np.finfo(np.double).eps:
```
stefanv: This check can be done before the inversion is attempted?  It simplifies the logic a bit.

stefanv:
It seems like the test suite already fails, but it may be worth inserting an explicit test for this, potentially one that also fails on 64-bit.

skimage/measure/fit.py
hmaarrfk: I'm not sure. in fact, when I computed `cond(M)` in the debugger with `import pdb; pdb.set_trace()`, the condition number was already `Inf`. I don't really want to cause other overflow errors or "old numpy specific errors"

From the docs
> eps | (float) The smallest representable positive number such that 1.0 + eps != 1.0.  Type of eps is an appropriate floating point type.

They don't seem to have a eps^-1. Thought I could do
`eps**-1` if you like.

matthew-brett:
Should this be a bug report to numpy too?

stefanv:
@matthew-brett I've raised this with the NumPy folks; they think it's not NumPy's responsibility.

Overall, it is a bad idea to use inverse if you can use solve.  Could we rewrite the solution here to do solves instead?

hmaarrfk:
Just for you two, I actually stepped through to figure out what was happening.

`S1` and `C1` are well conditioned. It is the result of the matrix multiplications and addition that is illconditionned. Typically this is caught a little lower

```python
        # M*|a b c >=l|a b c >. Find eigenvalues and eigenvectors
        # from this equation [eqn. 28]
        eig_vals, eig_vecs = np.linalg.eig(M)

        # eigenvector must meet constraint 4ac - b^2 to be valid.
        cond = 4 * np.multiply(eig_vecs[0, :], eig_vecs[2, :]) \
               - np.power(eig_vecs[1, :], 2)
        a1 = eig_vecs[:, (cond > 0)]
        # seeks for empty matrix
        if 0 in a1.shape or len(a1.ravel()) != 3:
            return False
```
On 64 bit:
```
(Pdb) cond
array([-0.28717554, -0.71271829, -0.00108459])
```
On 32 bit
```
(Pdb) cond
array([-1.38708836,  0.84396406, -0.03172746])
```

It just happened that on 64 bit the error didn't show up. But I don't think that the `eig` algorithm will complain if you have a ill-conditionned matrix.

Therefore, the tests both in *64 bit* and *32 bit* are caught by my added logic.

hmaarrfk:
Technically, since the matrix is so small, I think we can simply check that if one `eig_vals` is small and also return `False`.

Anyway, I think checking the condition number is more "mathematical"

hmaarrfk:
@stefanv. I probably agree that doing something like finding tge best eigen value is probably verbose. 

From ‘numpy.linalg.solve’

> a must be square and of full-rank 

We would need to check the condition number of the matrix anyway otherwise risk having an other linalg error

hmaarrfk:
Finally @stefanv. I quickly skimmed the paper and the code seems to be following the formalism the paper is using.

I suggest that we leave the reformulation to somebody that wants to derive it. (It May be a paper for itself)

emmanuelle:
So, should we merge this PR?

hmaarrfk:
I think so! I can rebase if you want.

jni:
@hmaarrfk I can't quite tell from the conversation whether you have a test case that will fail on 64 bit? It would be nice for this to be tested.

hmaarrfk:
Sorry, I guess I coudln't think of one. I'll try to bring out my pen and paper this weekend. Remind me.

hmaarrfk:
@jni 

```
=============================================== FAILURES ================================================
__________________________________ test_ellipse_model_estimate_failers __________________________________

    def test_ellipse_model_estimate_failers():
        # estimate parameters of real data
        model = EllipseModel()
        assert not model.estimate(np.ones((5, 2)))
        # Before PR https://github.com/scikit-image/scikit-image/pull/3103
        # This next line would return True on 32bit
        assert not model.estimate(np.array([[50, 80], [51, 81], [52, 80]]))
        # This is the same test that causes the model to
        # return true on 64 bit prooving that this was also an issue on
        # 64 bit arch
>       assert not model.estimate(np.array([[50, 80], [51, 81], [52, 80.00000000001]]))
E       assert not True
E        +  where True = <bound method EllipseModel.estimate of <skimage.measure.fit.EllipseModel object at 0x7f422d07e5c0>>(array([[ 50.,  80.],\n       [ 51.,  81.],\n       [ 52.,  80.]]))
E        +    where <bound method EllipseModel.estimate of <skimage.measure.fit.EllipseModel object at 0x7f422d07e5c0>> = <skimage.measure.fit.EllipseModel object at 0x7f422d07e5c0>.estimate
E        +    and   array([[ 50.,  80.],\n       [ 51.,  81.],\n       [ 52.,  80.]]) = <built-in function array>([[50, 80], [51, 81], [52, 80.00000000001]])
E        +      where <built-in function array> = np.array

skimage/measure/tests/test_fit.py:233: AssertionError
================================== 1 failed, 22 passed in 0.72 seconds ==================================
```

You can go ahead and force push to remove my "Reverting" of the fix once travis shows that the tests fail. I'm not really a fan of using `eps`. Apparently it is pretty slow the first time you call it.

Also, you can precompute the inv of `C1`, and paste it in instead. It is nice round numbers so it stays readable. I figured that adding that in would delay this PR getting merged.

hmaarrfk:
Failed on Appveyor: https://ci.appveyor.com/project/scikit-image/scikit-image/build/1.0.1014/job/qqf67dhmccr8f0so#L8458
Failed on Travis python 3.6: https://travis-ci.org/scikit-image/scikit-image/jobs/430813979#L2803
But passed on python 3.7: https://travis-ci.org/scikit-image/scikit-image/jobs/430813978

hmaarrfk:
@jni Here is what recreates the issue:

```python
from skimage.measure import EllipseModel
import numpy as np
model = EllipseModel()
print(model.estimate(np.array([[50, 80], [51, 81], [52, 80.00000000001]])))
print(model.estimate(np.array([[50, 80], [51, 81], [52, 80.0001000001]])))
print(model.estimate(np.array([[50, 80], [51, 81], [52, 80.000000001]])))
print(model.estimate(np.array([[50, 80], [51, 81], [52, 80.00000001]])))
print(model.estimate(np.array([[50, 80], [51, 81], [52, 80.0000001]])))
print(model.estimate(np.array([[50, 80], [51, 81], [52, 80.000001]])))
print(model.estimate(np.array([[50, 80], [51, 81], [52, 80.00001]])))
```

```
False
False
True
True
True
True
False
```
I thought you could simply check the condition number, but it fails the doctest, which should clearly pass....

hmaarrfk:
part of me thinks we should chek the number of input points. The problem of fitting an ellipse is ilposed with less than 5 points

skimage/measure/tests/test_fit.py
```diff
@@ -229,7 +229,22 @@ def test_ellipse_model_estimate_failers():
...
     assert not model.estimate(np.ones((5, 2)))
+    # Before PR https://github.com/scikit-image/scikit-image/pull/3103
+    # This next line would return True on 32bit
     assert not model.estimate(np.array([[50, 80], [51, 81], [52, 80]]))
+    # This is the same test that causes the model to
+    # return true on 64 bit prooving that this was also an issue on
+    # 64 bit arch
+    assert not model.estimate(np.array([[50, 80], [51, 81], [52, 80.00000000001]]))
+    # Unfortunately, it seems to depend on the version of python that is used
+    # Therefore, test for many cases
```
jni: @hmaarrfk based on your comment, shouldn't this test be:

```python
assert not all(model.estimate(np.array([[50, 80], [51, 81], [52, 80 + 10**(-n)]]))
               for n in range(4, 12))
```
?

hmaarrfk: @jni, refactored as per your suggestion. problem remains :/

jni: Which problem? Do they all pass for some Python versions?

hmaarrfk: Well my **fix** causes the doctest to fail. But the doctest seems to make sense. https://travis-ci.org/scikit-image/scikit-image/jobs/432304643#L2823

I'm glad to make the test xfail for code self-documentation.

jni: OOOOoooooOOOOhhhh. Very interesting. Hmm, I think ellipse-fitting to a line should work, actually. =\

skimage/measure/tests/test_fit.py
```diff
@@ -225,11 +225,20 @@ def test_ellipse_model_estimate_from_data():
...
+    # Unfortunately, it seems to depend on the version of python that is used
+    # Therefore, test for many cases for different values of epsilon.
+
     # estimate parameters of real data
     model = EllipseModel()
     assert not model.estimate(np.ones((5, 2)))
-    assert not model.estimate(np.array([[50, 80], [51, 81], [52, 80]]))
+    assert not model.estimate(np.array([[50, 80],
+                                        [51, 81],
+                                        [52, 80 + epsilon]]))
```
Borda: according to the comment, what is the epsilon number (it is not clear from the review scope)

hmaarrfk: sorry, quite often, epsilon is a small number in math. I could add a comment, but this wasn't the right approach anyway, and gave the same results as before.

matthew-brett:

<details>

## Rationale

</details>

PR #3109, opened 2018-05-28
Linking preserve_range parameter calls to docs

Jathrone:
If committed, this will add links to http://scikit-image.org/docs/dev/user_guide/data_types.html wherever preserve_range is used.

### DIFF
--- a/skimage/filters/_gaussian.py
+++ b/skimage/filters/_gaussian.py
@@ -15,10 +15,9 @@ def _convert_input(image, preserve_range):
     Parameters
     ----------
     preserve_range : bool, optional
-    Whether to keep the original range of values. Otherwise, the input
-    image is converted according to the conventions of `img_as_float`.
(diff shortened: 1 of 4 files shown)


skimage/filters/_gaussian.py
```diff
@@ -11,6 +11,14 @@
 
 
 def _convert_input(image, preserve_range):
+    """
+    Parameters
+    ----------
+    preserve_range : bool, optional
+    Whether to keep the original range of values. Otherwise, the input
+    image is converted according to the conventions of `img_as_float`.
+    See more http://scikit-image.org/docs/dev/user_guide/data_types.html.
```
jni: @Jathrone 

Either "See URL for more" or "See more at URL". Sorry. =P Probably the former is nicer for line length.

jni:
@Jathrone thanks! =D

PR #3110, opened 2018-05-28
Let lower precision float arrays pass through `img_as_float`

stefanv:
This is an alternative approach to https://github.com/scikit-image/scikit-image/pull/3062/files

It also led me to identify a precision issue in the conversion from signed int -> float.

### DIFF
--- a/skimage/util/dtype.py
+++ b/skimage/util/dtype.py
@@ -111,7 +111,18 @@ def convert(image, dtype, force_copy=False, uniform=False):
     itemsize_in = dtypeobj_in.itemsize
     itemsize_out = dtypeobj_out.itemsize
 
-    if dtype_in == dtype_out:
+    # Below, we do an `issubdtype` check.  Its purpose is to find out
+    # whether we can get away without doing any image conversion.  This happens
+    # when:
+    #
+    # - the output and input dtypes are the same or
+    # - when the output is specified as a type, and the input dtype
+    #   is a subclass of that type (e.g. `np.floating` will allow
+    #   `float32` and `float64` arrays through)
+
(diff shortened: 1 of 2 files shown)


skimage/util/dtype.py
```diff
@@ -378,7 +390,33 @@ def img_as_float64(image, force_copy=False):
...
+
+    Notes
+    -----
+    The range of a floating point image is [0.0, 1.0] or [-1.0, 1.0] when
+    converting from unsigned or signed datatypes, respectively.
+    If the input image has a float type, intensity values are not modified
+    and can be outside the ranges [0.0, 1.0] or [-1.0, 1.0].
+
+    """
+    return convert(image, np.floating, force_copy)
```
hmaarrfk: Interesting, good call on the subclass!

skimage/util/dtype.py
```diff
@@ -292,9 +292,9 @@ def _scale(a, n, m, copy=True):
             # if imin_in:
             #     np.maximum(image, -1.0, out=image)
         else:
-            image = np.multiply(image, 2. / (imax_in - imin_in),
-                                dtype=computation_type)
-            image += 1.0 / (imax_in - imin_in)
+            image = np.add(image, 0.5, dtype=computation_type)
```
hmaarrfk: Did this really lead to an inacurracy?

Interesting.

jni: Yes, it did. See the `np.int8` test case that @stefanv added. It fails without this change, due to floating point arithmetic error.

hmaarrfk:
@jni is there a chance this can be backported to 0.14.1?

jni:

jni:
@hmaarrfk apparently it can. =D

PR #3113, opened 2018-05-28
Update installation instructions to reference existing requirements s…

stefanv:


INSTALL.rst
```diff
@@ -16,68 +16,46 @@ environment.
...
-.. _WinPython: https://winpython.github.io/
-
-b. Debian and Ubuntu
-````````````````````
+  pip install scikit-image
 
-On Debian and Ubuntu, install ``scikit-image`` with::
+If you are running Anaconda or miniconda, use::
 
-  sudo apt-get install python-skimage
```
soupault: Is this no longer the case for Debian? Did they remove the package from the repositories?

stefanv: Yes, I suspect that is what the greater majority of Debian users want.  Those who don't probably also know how to apt-get.

INSTALL.rst
```diff
@@ -16,68 +16,46 @@ environment.
...
-or using conda (for Anaconda users) as::
+or, on conda-based systems::
 
   conda uninstall scikit-image
 
-Now clone scikit-image on your local computer::
+Now, clone scikit-image on your local computer, and install::
 
   git clone https://github.com/scikit-image/scikit-image.git
+  pip install -e .
```
soupault: I'd add `cd scikit-image` in between

stefanv:
Good catch, please go ahead.

Issue #3117, opened 2018-05-29
Finish Python 2 removal


Carreau:
Followup on @jni pining me on #3000 (really good PR number for this BTW).

One of the critical things to do is to add the `python_requires='>=3.4'` to setup.py and add some extra failsafes in `setup.py` to ensure users get a good error message if they have an old pip version. Read http://python3statement.org/practicalities/ or more info.

Do you want to add scikit image to https://github.com/python3statement/python3statement.github.io ?

jni:
@Carreau Yes, as noted in the PR description, I "nabbed" the # when I saw it coming. =D

Thanks for the link — I was unaware of the `python_requires` option. I knew pinging you was a good idea. =D

It feels a bit redundant to sign the statement now. I would have liked to sign it in the first wave but there was some resistance. However it is perfectly aligned with our roadmap (which is not yet written down but I intend to solve that this week), so yes, we can add scikit-image to it.

stefanv:
LTS support for 2.7 on the 0.14.x branch will end 1 January 2020.  The first 3-only release will be 0.15.0, around November 2018–January 2019.

stefanv:
@Carreau The instructions on python3statement are a bit confusing: if you specify `python_requires`, that would stop newer pips also on 2.7 from installing.  If you throw an error in `setup.py` if 2.7 is detected, what happens to pip?  So, an example of how to fail in setup.py would also be useful.  For now, I'll add the `python_requires` and throw an error on skimage import if on 2.

Carreau:
Thanks for your question, 

It is a really good one, and the crux of why it took us about a year to get things right and patch upstream PyPI. We need to take a step back and realize that setup.py has 2 functions:

 - Build the package on the maintainer machine, and release it
 - Build the package on the user machine and install it (whl, or tgz). 

The `requires_python` set a metadata on the tgz/whl so that pip 9+ will not _attempt_ to download if the current Python is Python 2.7 (or 3.0, 3.1...). 

Though if you do _not_ have pip 9+ installed, then:
 1. pip will see the `py3.whl` extension
 2. decide there is no wheels for Py2, thus fallback on the `.tgz`.
 3. It will then unpack the `.tgz` and try to install the package; in which case the `python_requires` is ignored by setuptools. 

Now you have two choice. 
[...]

PR #3119, opened 2018-05-29
Ensure we are installing with / running on Python 3

stefanv:
Closes gh-3117

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -136,6 +136,7 @@ if __name__ == "__main__":
         ],
         install_requires=INSTALL_REQUIRES,
         requires=REQUIRES,
+        python_requires='>=3.5',
         packages=setuptools.find_packages(exclude=['doc']),
         include_package_data=True,
         zip_safe=False,  # the package can run out of an .egg file
--- a/setup.py
+++ b/setup.py
@@ -25,10 +25,21 @@ import setuptools
 from distutils.command.build_py import build_py
 from distutils.command.sdist import sdist
 
-if sys.version_info[0] < 3:
-    import __builtin__ as builtins
-else:
-    import builtins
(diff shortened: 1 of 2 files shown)


sciunto:
LGTM.

stefanv:
@sciunto You have merge privileges. Go for it!

sciunto:
Bang! :)

setup.py
```diff
@@ -25,10 +25,21 @@
 from distutils.command.build_py import build_py
 from distutils.command.sdist import sdist
 
-if sys.version_info[0] < 3:
-    import __builtin__ as builtins
-else:
-    import builtins
+if sys.version_info < (3, 4):
```
soupault: Shouldn't this be (3, 5)?

setup.py
```diff
@@ -25,10 +25,21 @@
...
+
+    error = """Python {py} detected.
+
+scikit-image 0.15+ support only Python 3.5 and above.
```
soupault: Or 3.4 here?

stefanv:
Say da da da da.

I wonder who can decode that one :)

stefanv:
Yes, probably! You can go ahead and push that fix to master.

soupault:
Having Python 3.4 EOL in 2019 (even before Python 2.7, https://devguide.python.org/#status-of-python-branches), the comparison above should be to 3.5.
Fixed in https://github.com/scikit-image/scikit-image/commit/1b379580782180f4e506864949b6257d9c43a94c.

jni:
Thanks @soupault! 3.4 was never the right choice. I might even argue for 3.6, but this debate can wait a bit. Thank you for the push and best of luck with your move!

PR #3120, opened 2018-05-29
Trying to accelerate slic

emmanuelle:
This PR is here to share some work I've been doing during the BIDS spring to accelerate cython for loops using prange. I do get a nice acceleration (factor of 4 using 8 threads, so it's quite good). 

Some remaining issues:
- ~~as I understand, prange only works with gcc (not with clang). I haven't tried compiling this code on a Mac.~~ compilation with openmp can be made optional
- ~~how to choose the number of threads? Should we leave it to the user? Can it be an argument of the function~~ that's what I did.


emmanuelle:
@scoder comments or ideas very welcome :-)

stefanv:
Compilation with OpenMP see also https://github.com/scikit-image/scikit-image/pull/1570/files

emmanuelle:
On my machine tests fail randomly (half of the time). With ``test_gray_3d`` for example, there seems to be some boundary effects (values are the right ones, except at the border of expected superpixels). Do idea where this comes from... @jni

skimage/segmentation/_slic.pyx
```diff
@@ -157,15 +158,15 @@ def _slic_cython(double[:, :, :, ::1] image_zyx,
             n_segment_elems[:] = 0
             segments[:, :] = 0
             for z in range(depth):
-                for y in range(height):
-                    for x in range(width):
+                for y in prange(height):
+                    for x in prange(width):
```
scoder: Why the nested prange here? Wouldn't the middle loop generate enough tasks?
There's some overhead involved in setting up the loop, so doing it for the inner loop could easily slow things down.
My (pure) guess is that `depth` (probably something like 2-3?) would be too small to provide enough tasks for multi-core systems, but the `height` is probably large enough.

Normally, it should perform best to apply parallelism to larger memory blocks (e.g. rows, which then get processed in larger chunks by OpenMP), and have each thread run over the entire block to exploit the memory locality through data streaming and prefetching. Basically, the larger the blocks, the smaller the threading overhead.

You probably already did that, but I'd just swap things around and try prange in different places to benchmark what performs best. It's a bit difficult to predict because I can't judge the locality impact of the `segments` vs. the `image_zyx` arrays.

skimage/segmentation/_slic.pyx
```diff
@@ -4,6 +4,7 @@
 #cython: wraparound=False
 from libc.float cimport DBL_MAX
 from cpython cimport bool
+from cython.parallel import prange
```
scoder: This should be a `cimport`, although I think it makes no difference.

emmanuelle:
Thank you for your comments @scoder. I actually had to revert to range in the part you commented on because of concurrent writing by different threads (which was the reason why tests are failing). However, there is another set of nested for loops on pixels where we can also decide which one is prange. My idea was that the set of array elements considered inside the prange loop would fit on cache, but maybe this is not needed?

emmanuelle:
The scalability I get seems to depend quite a lot on the machine: on my laptop I get only a 1.5 speedup for 4 threads, while I get a factor of 3 for 4 threads on another machine...

emmanuelle:
I removed the "do not merge" in the PR title because maybe this is something we want to merge :-).

jni:
@emmanuelle I think it needs to be rebased on master and the asv benchmark function there adapted to test n_jobs=1, n_jobs=2, and n_jobs=4. =D (these should be different methods on the same class, I think.)

emmanuelle:
@jni ok I rebased :-)

emmanuelle:
@jni I rebased and added the asv benchmarks (is it the right way to do it?).

emmanuelle:
The code in the setup.py to add compiler flags is basically the same as in #3267, so we'll see which one gets merged first.

jni:
@emmanuelle I'm not sure whether I'm not compiling it right (using `make clean; pip install -U --no-deps -e .`), but currently the benchmark shows a decrease in performance on my machine:

```
 $ asv dev --bench time_slic_basic.*
· Discovering benchmarks
· Running 3 total benchmarks (1 commits * 1 environments * 3 benchmarks)
[  0.00%] ·· Building for existing-py_home_jni_miniconda3_envs_cf_bin_python
[  0.00%] ·· Benchmarking existing-py_home_jni_miniconda3_envs_cf_bin_python
[ 33.33%] ··· Running benchmark_segmentation.SegmentationSuite.time_slic_basic        2.36s
[ 66.67%] ··· Running benchmark_segmentation.SegmentationSuite.time_slic_basic_2jobs  2.52s
[100.00%] ··· Running benchmark_segmentation.SegmentationSuite.time_slic_basic_4jobs  3.24s
```

Have you tried it?

```
 $ gcc --version
gcc (Ubuntu 7.3.0-16ubuntu3) 7.3.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

emmanuelle:
@jni hum, this is interesting because with asv I get the same result as you, but when using %timeit I see the improvement
```
In [4]: %timeit b = segmentation.slic(a, n_jobs=1)
3.47 s ± 4.44 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: %timeit b = segmentation.slic(a, n_jobs=2)
1.9 s ± 5.73 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit b = segmentation.slic(a, n_jobs=4)
1.48 s ± 22.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

About gcc:
```
$ gcc --version
gcc (Ubuntu 7.2.0-8ubuntu3.2) 7.2.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

jni:
Ah! This is very interesting and annoying! Looks like asv measures *CPU time*, not wall time:

> The default timing function is the POSIX `CLOCK_PROCESS_CPUTIME`, which measures the CPU time used only by the current process.

https://github.com/airspeed-velocity/asv/blob/master/docs/source/writing_benchmarks.rst

I’m not sure what the right way to handle this is... perhaps to mentally divide by the number of cores? I’ll investigate.

jni (dismissed):
@emmanuelle ideally we would "fix" the benchmarks before merging, but I can see that you are eager to see this in so I'll approve it. =) I definitely approve of the code changes, and having this example opens up other Cython functions for optimisation. Thanks! =)

sciunto:
I ignored the existence of `prange`. I'm curious to check if it could improve our functions on hough transforms.

Naive question. If `with nogil` is used, can we still expect better performances?

jni:
@sciunto, GIL is the Python Global Interpreter Lock, which prevents two Python statements from being run at the same time on two different threads. It is there so that Python is memory-safe. (For example, you won't get two different functions simultaneously increasing the reference count to an existing object, which could cause the count to become incorrect.)

`nogil` means that if you run this function concurrently with another `nogil` function in another thread, they can run at the same time on two different processors. For example, *without* `nogil`, if you do `threadpool.map(function, image_collection)`, it doesn't matter how many threads or compute cores you have, it will take the same time to run as `list(map(function, image_collection))`. However, if the bulk of your function's work is done within a `nogil` block, then the threadpool will be faster.

The limitation of this is that users of scikit-image need to be writing things with `dask`, `threading`, or some other concurrency library to see any multi-core benefits. I think probably 95%+ of our users are not doing this, and they are writing simple serial code. (I myself rarely bother writing threaded code!)

In contrast, all SLIC users can easily benefit from `prange` with this PR. (Well, those using gcc-compiled scikit-image anyway. ;)

The reason that we don't want to use `n_jobs=-1` always (use as many cores as possible) is that if you use *both* `threaded.map` or `dask` *and* prange, you now get something called thread oversubscription, where you are trying to execute more things in parallel than you have cores, and things become very slow, even slower than serial execution. Intel is doing [fancy things](https://www.threadingbuildingblocks.org) to fix this, including releasing their own Python distribution, but that is beyond the scope of scikit-image to worry about, I think. =) (For one, only Intel processors would benefit from fixes using these techniques.)

Hope this helps!

@hmaarrfk feel free to use some, all, or none of the above in your parallel scikit-image guide. =)

emmanuelle:
I solved the conflicts coming from the merge of #3267, so I think this one is ready @stefanv @soupault

hmaarrfk:
Does this slow down the old function, the one that didn't have the parameter n_jobs?

emmanuelle:
@hmaarrfk good question... I ran a test on a (1500, 2555) RGB image (tiled chelsea  :-)) and actually there is a significant slow down with the new version for ``n_jobs = 1`` :-(. 
Old function: 3.51 s
New function with n_job1=1: 4.32 s
New function with n_jobs=2: 3.21 s
I don't know where the slow down comes from, I need to think it over. Anyway, this kind of behavior proves that having the benchmarks suite is useful...

hmaarrfk:
Interesting.

Maybe write the tests so that they can be run backward. I was thinking with a 
```
try:
    new_function_sig
except:
    old_function_sig
```

jni:
@hmaarrfk try/except are actually quite expensive to run when hitting the except. So you would artificially inflate the numbers. I would just run the single core benchmark without njobs, since it defaults to 1.

setup.py
```diff
@@ -121,6 +121,55 @@ def build_extensions(self):
 REQUIRES = [r.replace('[array]', '') for r in REQUIRES]
 
 
+compile_flags = ['-fopenmp']
```
soupault: https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html#compiling says:
```
For Microsoft Visual C++ compiler, use '/openmp' instead of '-fopenmp'.
```

Should we consider also Win?

setup.py
```diff
@@ -121,6 +121,55 @@ def build_extensions(self):
...
+            except CompileError:
+                return False
+            try:
+                # Link shared lib rather then executable to avoid
+                # http://bugs.python.org/issue4431 with MSVC 10+
+                cc.link_shared_lib(objects, "testlib",
+                                   extra_postargs=link_flags)
+            except (LinkError, TypeError):
+                return False
+        finally:
```
soupault: FYI, this `try-finally` could we rewritten using `with tempfile.TemporaryDirectory() as tmpdir:` context manager.

skimage/segmentation/slic_superpixels.py
```diff
@@ -11,7 +11,7 @@
 def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
          spacing=None, multichannel=True, convert2lab=None,
          enforce_connectivity=True, min_size_factor=0.5, max_size_factor=3,
-         slic_zero=False):
+         slic_zero=False, n_jobs=1):
```
soupault: I think we should mention this in https://github.com/scikit-image/scikit-image/blob/master/doc/release/release_dev.rst .

soupault (commented):
Thanks @emmanuelle ! Few remarks, otherwise looks great!

jni:
Regarding single-thread performance, here's a page that *might* be useful in exploring what's going wrong:

https://software.intel.com/en-us/articles/performance-obstacles-for-threading-how-do-they-affect-openmp-code

Regarding measuring the multithreaded performance correctly on asv, apparently we can set the timer on a benchmark class, so we can change from CPU time to Wall time, or even from CPU time to CPU time divided by the number of cores, if we write our own timer function. See:

https://asv.readthedocs.io/en/stable/writing_benchmarks.html#timing
and
https://asv.readthedocs.io/en/stable/writing_benchmarks.html#benchmark-attributes

sciunto:
@emmanuelle I modify the milestone to 0.17.

emmanuelle:
```
emma@plotty:~/code/scikit-image$ asv dev -b SegmentationSuite
· Discovering benchmarks
· Running 8 total benchmarks (1 commits * 1 environments * 8 benchmarks)
[  0.00%] ·· Benchmarking existing-py_usr_bin_python3
[  6.25%] ··· benchmark_segmentation.SegmentationSuite.peakmem_setup                                                        306M
[ 12.50%] ··· benchmark_segmentation.SegmentationSuite.peakmem_slic_basic                                                   820M
[ 12.50%] ···· /home/emma/code/scikit-image/benchmarks/benchmark_segmentation.py:50: FutureWarning: skimage.measure.label's indexing starts from 0. In future version it will start from 1. To disable this warning, explicitely set the `start_label` parameter to 1.
               multichannel=False)

[ 18.75%] ··· benchmark_segmentation.SegmentationSuite.peakmem_slic_basic_2jobs                                             820M
[ 18.75%] ···· /home/emma/code/scikit-image/benchmarks/benchmark_segmentation.py:54: FutureWarning: skimage.measure.label's indexing starts from 0. In future version it will start from 1. To disable this warning, explicitely set the `start_label` parameter to 1.
               multichannel=False)

[ 25.00%] ··· benchmark_segmentation.SegmentationSuite.peakmem_slic_basic_4jobs                                             820M
[ 25.00%] ···· /home/emma/code/scikit-image/benchmarks/benchmark_segmentation.py:58: FutureWarning: skimage.measure.label's indexing starts from 0. In future version it will start from 1. To disable this warning, explicitely set the `start_label` parameter to 1.
               multichannel=False)

[ 31.25%] ··· benchmark_segmentation.SegmentationSuite.time_mask_slic                                                     failed
[ 31.25%] ···· asv: benchmark timed out (timeout 60.0s)
[ 37.50%] ··· benchmark_segmentation.SegmentationSuite.time_slic_basic                                                   21.0±0s
[ 37.50%] ···· /home/emma/code/scikit-image/benchmarks/benchmark_segmentation.py:17: FutureWarning: skimage.measure.label's indexing starts from 0. In future version it will start from 1. To disable this warning, explicitely set the `start_label` parameter to 1.
               multichannel=False)

[ 43.75%] ··· benchmark_segmentation.SegmentationSuite.time_slic_basic_2jobs                                             12.5±0s
[ 43.75%] ···· /home/emma/code/scikit-image/benchmarks/benchmark_segmentation.py:21: FutureWarning: skimage.measure.label's indexing starts from 0. In future version it will start from 1. To disable this warning, explicitely set the `start_label` parameter to 1.
               multichannel=False)

[ 50.00%] ··· benchmark_segmentation.SegmentationSuite.time_slic_basic_4jobs                                             7.88±0s
[ 50.00%] ···· /home/emma/code/scikit-image/benchmarks/benchmark_segmentation.py:25: FutureWarning: skimage.measure.label's indexing starts from 0. In future version it will start from 1. To disable this warning, explicitely set the `start_label` parameter to 1.
               multichannel=False)
```
asv benchmark on this branch

emmanuelle:
on master
```
emma@plotty:~/code/scikit-image$ asv dev -b SegmentationSuite
· Discovering benchmarks
· Running 4 total benchmarks (1 commits * 1 environments * 4 benchmarks)
[  0.00%] ·· Benchmarking existing-py_usr_bin_python3
[ 12.50%] ··· benchmark_segmentation.SegmentationSuite.peakmem_setup                                                        307M
[ 25.00%] ··· benchmark_segmentation.SegmentationSuite.peakmem_slic_basic                                                   820M
[ 25.00%] ···· /home/emma/code/scikit-image/benchmarks/benchmark_segmentation.py:40: FutureWarning: skimage.measure.label's indexing starts from 0. In future version it will start from 1. To disable this warning, explicitely set the `start_label` parameter to 1.
               segmentation.slic(self.image, enforce_connectivity=False, multichannel=False)

[ 37.50%] ··· benchmark_segmentation.SegmentationSuite.time_mask_slic                                                     failed
[ 37.50%] ···· asv: benchmark timed out (timeout 60.0s)
[ 50.00%] ··· benchmark_segmentation.SegmentationSuite.time_slic_basic                                                   12.6±0s
[ 50.00%] ···· /home/emma/code/scikit-image/benchmarks/benchmark_segmentation.py:16: FutureWarning: skimage.measure.label's indexing starts from 0. In future version it will start from 1. To disable this warning, explicitely set the `start_label` parameter to 1.
               segmentation.slic(self.image, enforce_connectivity=False, multichannel=False)

```

grlee77:
```
[ 37.50%] ··· benchmark_segmentation.SegmentationSuite.time_mask_slic                                                     failed
[ 37.50%] ···· asv: benchmark timed out (timeout 60.0s)
```
This timeout is why I had reduced the size of the data for this benchmark in #4891

emmanuelle:
I'm probably going to close this PR soon since I can't get any satisfying performance improvement. I tried having the prange for one of the loops on coordinates (z, y, and x) and for these three trials it's a little bit better when the prange is on the outer loop, but for `num_threads=2` I always get a result worse or equal than 1 thread on master, and with more threads the best I can reach is 50% acceleration with respect to master. Unless someone else has an idea for a better implementation? @grlee77 maybe?

emmanuelle:
Probably slic is not a good candidate for using `prange` since in fact the chunks are very small since the x y z indices are just for a grid tile. Ideally one would parallelize over segments instead (the `k` loop) but there are some problems of concurrent access for the part below
```
if distance[z, y, x] > dist_center:
     nearest_segments[z, y, x] = k + start_label
     distance[z, y, x] = dist_center
```
but maybe we can iterate over segments in a more clever way than consecutive ones, so that there is no risk of concurrent access.

grlee77:
> Probably slic is not a good candidate for using prange since in fact the chunks are very small since the x y z indices are just for a grid tile

Yeah, I didn't try compiling and running the benchmarks, but it looks like there are only basic multiply/add operations in the inner loop, so we may be limited by memory access times more than computation in the multithreaded case.

scoder:
@emmanuelle I agree about the problem being here:
```cython
if distance[z, y, x] > dist_center:
     nearest_segments[z, y, x] = k + start_label
     distance[z, y, x] = dist_center
```
Could you use separate `nearest_segments` and `distance` arrays (or ones with an additional `k` dimension) to collect the "k local" changes in parallel, and then merge the `k` sub-arrays back together in a final sequential loop in the end? That's obvious trading `k` times the space for a bit of time, not sure if that is acceptable. Images can be quite large sometimes.

There's also an [open issue for Cython](https://github.com/cython/cython/issues/3585) to add `min()` and `max()` as OpenMP regressions. Would apply the the `distance` array but probably not to the `nearest_segments`.

scoder:
> Could you use separate `nearest_segments` and `distance` arrays …

Or, rather than allowing me to make up some fishy "improvement" to an established algorithm, it's probably better to take a look through existing approaches to parallelise SLIC. Searching for [parallel slic implementation](https://duckduckgo.com/?q=parallel+slic+implementation) gives me a few hits, at least.

I understand that this was an approach to get something out of it quickly, but it seems that the algorithm itself resists a little bit.

emmanuelle:
@scoder thanks for your comments! Indeed, yesterday I was a bit vexed that I could not get any reasonable speed-up so I started looking at the literature on parallel slic. For example, in [this paper proposing a GPU implementation](https://www.researchgate.net/profile/Carl_Ren/publication/265890194_gSLIC_a_real-time_implementation_of_SLIC_superpixel_segmentation/links/56d8aedd08aee1aa5f802ceb/gSLIC-a-real-time-implementation-of-SLIC-superpixel-segmentation.pdf), the outer loop is on image pixels, and each pixel has a list of k indices corresponding to the possible centroids it can belong to (9 such indices in 2D). Distances to these centroids is computed for each pixel, a new label is attributed to the pixel, and after looping over all pixels, the new position of centroids is computed. This way, there is no problem of concurrent access. You are right that keeping the lists of indices for each pixel would have an important memory cost, unless there is a way to compute the indices (using modulo and division operations).

It could be worth giving it a try with this other way of looping. However, I'm not overly optimistic since I brutally tried to turn the loop over segments to a `prange` (ignoring the risk of concurrent access), and I could not see an improvement. Probably @grlee77 is right that

> it looks like there are only basic multiply/add operations in the inner loop, so we may be limited by memory access times more than computation in the multithreaded case.

Thanks also for pointing at the open issue on min/max reduction!

alexdesiqueira:
Hey @scikit-image/core,
how are we on this one? If we are happy, I'll fix the conflicts and merge it. Thanks!

skimage/segmentation/_slic.pyx
```diff
@@ -3,6 +3,8 @@
 #cython: nonecheck=False
 #cython: wraparound=False
 from libc.float cimport DBL_MAX
+from cpython cimport bool
```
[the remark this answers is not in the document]
mkcor: looks like it indeed

mkcor (commented):
@alexdesiqueira +1 let's not lose this contribution, thanks for taking care of the merge conflicts :pray:

alexdesiqueira:
Alright, let's see what we got. :slightly_smiling_face:

alexdesiqueira:
Azure warning:
```
022-08-25T19:27:47.3102697Z setup.py:9: DeprecationWarning: 
2022-08-25T19:27:47.3104273Z 
2022-08-25T19:27:47.3106742Z   `numpy.distutils` is deprecated since NumPy 1.23.0, as a result
2022-08-25T19:27:47.3108078Z   of the deprecation of `distutils` itself. It will be removed for
2022-08-25T19:27:47.3110438Z   Python >= 3.12. For older Python versions it will remain present.
2022-08-25T19:27:47.3111790Z   It is recommended to use `setuptools < 60.0` for those Python versions.
2022-08-25T19:27:47.3113069Z   For more details, see:
2022-08-25T19:27:47.3114456Z     https://numpy.org/devdocs/reference/distutils_status_migration.html 
2022-08-25T19:27:47.3115307Z 
2022-08-25T19:27:47.3116039Z 
2022-08-25T19:27:47.3117461Z   from numpy.distutils.command.build_ext import build_ext as npy_build_ext
2022-08-25T19:27:47.3119344Z C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\config\setupcfg.py:508: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.
2022-08-25T19:27:47.3120776Z   warnings.warn(msg, warning_class)
2022-08-25T19:27:47.3122055Z Partial import of skimage during the build process.
```

Azure error:
```
22-08-25T19:27:52.4272028Z --- Logging error ---
2022-08-25T19:27:52.4274135Z Traceback (most recent call last):
2022-08-25T19:27:52.4275547Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\logging\__init__.py", line 1088, in emit
2022-08-25T19:27:52.4276925Z     stream.write(msg + self.terminator)
2022-08-25T19:27:52.4278473Z ValueError: underlying buffer has been detached
2022-08-25T19:27:52.4279856Z Call stack:
2022-08-25T19:27:52.4281015Z   File "setup.py", line 263, in <module>
2022-08-25T19:27:52.4282048Z     setup(
2022-08-25T19:27:52.4283346Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\numpy\distutils\core.py", line 169, in setup
2022-08-25T19:27:52.4284637Z     return old_setup(**new_attr)
2022-08-25T19:27:52.4285939Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\__init__.py", line 87, in setup
2022-08-25T19:27:52.4287747Z     return distutils.core.setup(**attrs)
2022-08-25T19:27:52.4289555Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\core.py", line 185, in setup
2022-08-25T19:27:52.4291001Z     return run_commands(dist)
2022-08-25T19:27:52.4292778Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\core.py", line 201, in run_commands
2022-08-25T19:27:52.4294445Z     dist.run_commands()
2022-08-25T19:27:52.4296190Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\dist.py", line 973, in run_commands
2022-08-25T19:27:52.4298062Z     self.run_command(cmd)
2022-08-25T19:27:52.4299787Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\dist.py", line 1217, in run_command
2022-08-25T19:27:52.4301544Z     super().run_command(command)
2022-08-25T19:27:52.4303365Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\dist.py", line 992, in run_command
2022-08-25T19:27:52.4304766Z     cmd_obj.run()
2022-08-25T19:27:52.4306331Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\wheel\bdist_wheel.py", line 299, in run
2022-08-25T19:27:52.4307740Z     self.run_command('build')
2022-08-25T19:27:52.4309538Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\cmd.py", line 319, in run_command
2022-08-25T19:27:52.4311094Z     self.distribution.run_command(command)
2022-08-25T19:27:52.4312829Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\dist.py", line 1217, in run_command
2022-08-25T19:27:52.4314693Z     super().run_command(command)
2022-08-25T19:27:52.4316544Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\dist.py", line 992, in run_command
2022-08-25T19:27:52.4318510Z     cmd_obj.run()
2022-08-25T19:27:52.4319986Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\numpy\distutils\command\build.py", line 62, in run
2022-08-25T19:27:52.4321784Z     old_build.run(self)
2022-08-25T19:27:52.4323362Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\command\build.py", line 132, in run
2022-08-25T19:27:52.4325143Z     self.run_command(cmd_name)
2022-08-25T19:27:52.4326796Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\cmd.py", line 319, in run_command
2022-08-25T19:27:52.4328763Z     self.distribution.run_command(command)
2022-08-25T19:27:52.4330441Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\dist.py", line 1217, in run_command
2022-08-25T19:27:52.4331789Z     super().run_command(command)
2022-08-25T19:27:52.4333525Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\setuptools\_distutils\dist.py", line 992, in run_command
2022-08-25T19:27:52.4334749Z     cmd_obj.run()
2022-08-25T19:27:52.4336415Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\numpy\distutils\command\build_src.py", line 144, in run
2022-08-25T19:27:52.4337788Z     self.build_sources()
2022-08-25T19:27:52.4339470Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\numpy\distutils\command\build_src.py", line 161, in build_sources
2022-08-25T19:27:52.4340876Z     self.build_extension_sources(ext)
2022-08-25T19:27:52.4342754Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\numpy\distutils\command\build_src.py", line 306, in build_extension_sources
2022-08-25T19:27:52.4344311Z     sources = list(ext.sources)
2022-08-25T19:27:52.4345932Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\pythran\dist.py", line 170, in sources
2022-08-25T19:27:52.4347876Z     tc.compile_pythranfile(source, output_file,
2022-08-25T19:27:52.4351427Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\pythran\toolchain.py", line 477, in compile_pythranfile
2022-08-25T19:27:52.4354806Z     output_file = compile_pythrancode(module_name, fd.read(),
2022-08-25T19:27:52.4356764Z   File "C:\hostedtoolcache\windows\Python\3.8.10\x86\lib\site-packages\pythran\toolchain.py", line 411, in compile_pythrancode
```

mkcor:
And this error https://github.com/scikit-image/scikit-image/pull/6453#issuecomment-1228957405 as well.

stefanv:
I suspect we'd be able to get this built fairly easily on Meson now. And the changes are not very invasive.

stefanv:
Examining the PR more closely, it looks like it uses `prange` only on the height of the image. Ideally, I suppose, one would have to determine the best dimension to prange over.  Perhaps something like:

```
                for z in range(z_min, z_max, num_threads=z_jobs):
                    for y in prange(y_min, y_max, num_threads=y_jobs):
                        for x in range(x_min, x_max, num_threads=x_jobs):
```

Although, I'm not even sure whether Cython would allow you to dynamically compute num_threads on prange.

Anyway, this PR isn't quite ready for merging, and requires some further investigation and benchmarking.  So, I'm going to leave it here to note the idea, and remove the "OK to Merge" label.

scoder:
> uses `prange` only on the height of the image. Ideally, I suppose, one would have to determine the best dimension to prange over

Not sure. What matters is a straight memory layout, so the height (i.e. chunks of rows of consecutive points in memory that benefit from RAM prefetching) seems a good thread partitioning scheme (for a 2D image). Large chunks of consecutive memory are usually as good as it gets.

stefanv:
That's a great point, thanks for the reminder. We operate over z, y, x on a pixel level, so batching on `y` was probably sensible.

matthew-brett:

<details>

## Rationale

</details>

PR #3125, opened 2018-05-30
Add nD rotation matrix generation

kne42:
## Description
Add function for the generation of a matrix to rotate a vector to coincide with another.

## Checklist
- [x] Unit tests

## References
- taken from #2739 
- **Zhelezov OI. 2018.** One Modification which Increases Performance of N-Dimensional rotation Matrix Generation Algorithm. **International Journal of Chemistry, Mathematics, and Physics** **2**(2): _pp. 13-18_. https://dx.doi.org/10.22161/ijcmp.2.2.1
- **ter Haar Romeny BM. 2003.** Front-End Vision and Multi-Scale Image Analysis. **Computation Imaging and Vision Vol. 27**. Springer: Dordrecht, Netherlands: _pp. 37-51_. https://doi.org/10.1007/978-1-4020-8840-7_3
- https://en.wikipedia.org/wiki/Rotation_matrix#In_three_dimensions

## For reviewers

  later.


emmanuelle:
About 3D rotations, did you see that a GSoC student is working on this for Scipy? https://mail.python.org/pipermail/scipy-dev/2018-May/022831.html
@kne42 @jni

jni:
I did not see that! @kne42 the blog linked there looks like a good resource!

Thanks @emmanuelle!

kne42:
@emmanuelle @jni @soupault @stefanv This PR is (finally) ready for review! :)

jni:
@kne42 Wow! I will do a proper review tomorrow as part of the scikit-image Paris sprint, but on first glance: :heart_eyes::heart_eyes::heart_eyes:

stefanv:
This looks great, thanks @kne42.  Questions: what is the situation around rotations in SciPy after their GSoC?  How would this typically be used in image processing, i.e. can we come up with a good gallery example?

kne42:
Sorry for the late reply, I keep forgetting about this 😅

So briefly, there are two rotation-centric utilities in SciPy: an nD angle-plane rotation in ndimage and the new Rotation object from GSoC. This new object offers conversion between angle-axis, vector-vector, matrix, quarternion, and (I think) euler angle rotations in 3 dimensions. Internally, everything is stored as a quaternion.

This proposed addition is an nD implementation of a vector-vector -> rotation matrix conversion. E.g. generate the matrix to rotate one vector to point in the same direction as the other.

This is probably most useful internally as a building block for nD algorithms that require rotation, but I’m sure users will be able to find direct use cases. Since I’m largely a beginner when it comes to image processing, the first example that comes to my mind would be with a rotation variant feature detector that outputs features in a plottable format. In this case, one could first rotate the image using the rotation matrix to the desired direction, then rotate the output features using the matrix transpose and plot them on the original image.

skimage/transform/tests/test_orientation.py
```diff
@@ -0,0 +1,139 @@
+import numpy as np
+from numpy import pi, sin, cos
+from numpy.testing import (assert_equal, assert_almost_equal,
+                           assert_array_almost_equal)
+
+from skimage.transform._orientation import (convert_quasipolar_coords,
+                                            compute_rotation_matrix,
+                                            _normalize, _axis_0_rotation_matrix)
+
```
alexandrejaguar: Could you please document the tests? What we are testing and the expected output would be great.

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,265 @@
...
+    True
+    """
+    u = np.asarray(x)
+
+    norm = np.linalg.norm(u)
+
+    return u / norm
+
+
+def _axis_0_rotation_matrix(u, indices=None):
```
alexandrejaguar: I think we need better names for the variables; what about `array_u`, `matrix_u` or something close to that?

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,265 @@
...
+    >>> R = _axis_0_rotation_matrix([0, 1, 0])
+    >>> R @ [0, 1, 0]
+    array([ 1.,  0.,  0.])
+    """
+    ndim = len(u)
+
+    if indices is None:
+        indices = range(ndim)
+
+    x = u
```
alexandrejaguar: Do we need this for not changing `u`? Could we explicit say this is a copy, with `x = u.copy()` or something close? `x = np.copy(u)` would be better when dealing with lists.

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,265 @@
...
+    w = indices
```
alexandrejaguar: Maybe it would be better only using `indices` and `u` as the input anyway. I would recommend you giving better names to the following variables: `R`, `A`, `r`, and so on.

alexdesiqueira (changes requested):
Hey @kne42, thank you very much for your contribution! I am reading the code and for now I have some small changes to ask.

kne42:
this is ready for re-review @alexdesiqueira

alexdesiqueira:
Thank you for your work @kne42, and sorry for not checking this before.
@scikit-image/core (especially the Guardian of the Dimensions, @jni) LGTM right now! Where are we on this one?

alexdesiqueira (approved):

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
+import numpy as np
+
+
+def _normalize(x):
+    """Normalizes an array.
```
jni:
```suggestion
    """Scale an array to have norm=1.
```

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+        Unitary array.
+
+    Examples
+    --------
+    >>> x = np.arange(5)
+    >>> uX = _normalize(x)
+    >>> np.isclose(np.linalg.norm(uX), 1)
+    True
+    """
+    u = np.asarray(x)
```
jni: Since `u` is the return value, I would call this something else, `arr` or `v`, for example.

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+
+    norm = np.linalg.norm(u)
+
+    return u / norm
+
+
+def _axis_0_rotation_matrix(unit_vector, indices=None):
+    """Generate a matrix that rotates a vector to coincide with the 0th (y-)
```
jni:
```suggestion
    """Generate a matrix that rotates a vector to be collinear with axis 0.
```

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+       coordinate axis.
```
jni:
```suggestion
```

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Parameters
+    ----------
+    unit_vector : (N, ) array
+        Unit vector.
+    indices : sequence of int, optional
+        Indices of the components of `axis` that should be transformed.
+        If `None`, defaults to all of the indices of `axis`.
```
jni:
```suggestion
        If `None`, defaults to all of the indices of `unit_vector`.
```

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+    rotation_matrix = np.eye(ndim)
+
+    # loop to create stages of 2D rotations around fixed axes
+    # that are multiplied to form our nD matrix; see: [2]_
+    for step in np.round(2 ** np.arange(np.log2(ndim))).astype(int):
+        plane_rotation_matrix = np.eye(ndim)
+
+        for n in range(0, ndim - step, step * 2):
+            if n + step >= len(indices):
+                break
```
jni: These two lines seem unnecessary? The `range` expression should ensure that this never happens?

If that is not the case for non-default indices, then maybe the range expression should be changed to stop at `len(indices) - step`?

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+                unit_vector[j] = 0
+
+        # compound current plane's rotation with previous ones'
+        rotation_matrix = plane_rotation_matrix @ rotation_matrix
+
+    return rotation_matrix
+
+
+def compute_rotation_matrix(src, dst, use_homogeneous_coords=False):
+    """Generate a matrix for the rotation of one vector to the direction
```
jni: PEP257 stipulates that the short description should be a one-liner. So:
```suggestion
    """Generate a matrix to rotate one vector onto another.
```

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+    of another.
+
+    The MNMRG algorithm [1]_ can be summarized as follows:
```
jni:
```suggestion
    The MNMRG algorithm [1]_, implemented here, is summarized as:
```

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+        1. directional vectors ``X`` and ``Y`` are normalized
+        2. a vector ``w`` is initialized containing the
+           indices of the differences between ``X`` and ``Y``
+        3. matrices ``Mx`` and ``My`` are generated for the rotation
+           of ``X`` and ``Y`` to the same axis for all indices
+           in ``w``
+        4. the inverse of ``My`` is mulitplied by ``Mx`` to form
+           the rotation matrix ``M`` which rotates vector ``X`` to the
+           direction of vector ``Y``
```
jni: Can we change all these to use active voice? "vectors are normalized" -> "normalize vectors"

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Parameters
+    ----------
+    src : (N, ) array
+        Vector to rotate.
+    dst : (N, ) array
+        Vector of desired direction.
+    use_homogeneous_coords : bool, optional
```
jni: I would rename this argument to just `homogeneous_coords`, and I would link to the wikipedia entry on the topic, which is excellent.

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+    True
+    """
+    # step 1: vectors are normalized
+    homogeneous_slice = -use_homogeneous_coords or None
+    src = _normalize(src[:homogeneous_slice])
+    dst = _normalize(dst[:homogeneous_slice])
+
+    if use_homogeneous_coords:
+        src = np.append(src, 1)
+        dst = np.append(dst, 1)
```
jni: TIL about `np.append`! =D Sheesh. LOL

skimage/transform/tests/test_orientation.py
```diff
@@ -0,0 +1,79 @@
...
+    assert_almost_equal(np.linalg.norm(uX), 1)
+
+
+def test_axis_0_rotation_matrix():
+    """test that rotation matrix properly rotates vector to face axis 0"""
+    # trivial case
+    M = _axis_0_rotation_matrix([0, 1, 0])
+    assert_almost_equal(M @ [0, 1, 0], [1, 0, 0])
+
+    # non-trivial case
```
jni: Maybe separate these into two tests?

skimage/transform/tests/test_orientation.py
```diff
@@ -0,0 +1,79 @@
...
+    Y = np.arange(5, 10)
+    uY = np.linalg.norm(X) / np.linalg.norm(Y) * Y
+
+    M = compute_rotation_matrix(X, Y)
+
+    Z = M @ X
+
+    assert_almost_equal(Z, uY)
+
+    # ensure preserves expected length
```
jni: As above: best if these tests are separated.

jni (changes requested):
Damn, this is great work, @kne42! It's crazy what a bit of mulling over time will do, this went from being quite arcane to me to being crystal clear! I've made some very minor suggestions that should be easy to fix, if you can get them done soon I can push that green button! :tada:

skimage/transform/_orientation.py
kne42: ah, thanks for informing of the correct mathy term! I was unsure of what to call this lol

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+    Examples
+    --------
+    >>> rotation_matrix = _axis_0_rotation_matrix([0, 1, 0])
+    >>> rotation_matrix @ [0, 1, 0]
+    array([ 1.,  0.,  0.])
+    """
+    ndim = len(unit_vector)
+
+    if indices is None:
+        indices = range(ndim)
```
kne42: note to self: this should be `list(range(ndim))`

carterbox (commented):
I previously implemented an ND rotation algorithm for [`polytope`](https://github.com/tulip-control/polytope), so I'd like to point out that there is _more than one rotation_ that aligns start and end vectors. For N > 2, not only are there multiple rotations, but these rotations do not all leave the space in the same orientation. Therefore, I'm not sure that this API allows sufficient control over the returned rotation matrix for the desired application.

As an example, if you want to rotate a pencil from pointing up to pointing down, you can flip the pencil way from you or rotate it to the right. These two rotations align the start vector with the end vector, so they could be valid matrices returned by `compute_rotation_matrix`. However, the pencil ends in different orientations, so which one will you get?

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+    src : (N, ) array
+        Vector to rotate.
+    dst : (N, ) array
+        Vector of desired direction.
+    use_homogeneous_coords : bool, optional
+        Whether the input vectors should be treated as homogeneous coordinates.
+
+    Returns
+    -------
+    rotation_matrix : (N, N) array
```
rfezzani:
```suggestion
    rotation_matrix : (N, N) ndarray
```

rfezzani (commented):
LGTM, despite some comments. Thank you for this contribution @kne42 ;-)

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+           in ``w``
+        4. the inverse of ``My`` is mulitplied by ``Mx`` to form
+           the rotation matrix ``M`` which rotates vector ``X`` to the
+           direction of vector ``Y``
+
+    Parameters
+    ----------
+    src : (N, ) array
+        Vector to rotate.
+    dst : (N, ) array
```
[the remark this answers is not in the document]
kne42: changed to `array-like`

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+        3. matrices ``Mx`` and ``My`` are generated for the rotation
+           of ``X`` and ``Y`` to the same axis for all indices
+           in ``w``
+        4. the inverse of ``My`` is mulitplied by ``Mx`` to form
+           the rotation matrix ``M`` which rotates vector ``X`` to the
+           direction of vector ``Y``
+
+    Parameters
+    ----------
+    src : (N, ) array
```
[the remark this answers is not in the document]
kne42: changed to `array-like`

skimage/transform/_orientation.py
```diff
@@ -0,0 +1,179 @@
...
+    return u / norm
+
+
+def _axis_0_rotation_matrix(unit_vector, indices=None):
+    """Generate a matrix that rotates a vector to coincide with the 0th (y-)
+       coordinate axis.
+
+    Parameters
+    ----------
+    unit_vector : (N, ) array
```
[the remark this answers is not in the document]
kne42: changed to `array-like`

skimage/transform/_orientation.py
kne42: > These two lines seem unnecessary? The `range` expression should ensure that this never happens?

Sometimes `len(indices) < ndim - 2` in non-default indices

> If that is not the case for non-default indices, then maybe the range expression should be changed to stop at len(indices) - step?

I'm not 100% sure. If you can convince me then I'll change it, but I don't want to risk breaking anything.

kne42:
This is ready for re-review.

@carterbox Honestly, I'm not a mathy person so idk. I just took an algorithm from some paper I kinda understood and implemented it here.

jni:
@carterbox thank you so much for chiming in! nD rotations have been a headache for us for a long time. 😅 Looking at polytope might be very useful for us.

@kne42 does the paper not discuss this problem at all?

kne42:
@jni Well here's how I understood @carterbox's comment:

```
|
v
-->
```
These vectors are **not** aligned **nor** face the same direction

```
-->
<--
```
These vectors **are** aligned **but do not** face the same direction

```
-->
-->
```
These vectors **are** aligned **and** face the same direction.

---

The paper _does_ say that it makes both vectors face the same direction (see [section II](http://aipublications.com/ijcmp/detail/one-modification-which-increases-performance-of-n-dimensional-rotation-matrix-generation-algorithm/)):

for any vectors of the same dimension `X` and `Y`, the matrix `M = compute_rotation_matrix(X, Y)` can be computed such that the vector `Y2 = M @ X` has the same norm as `X` and the same direction as `Y`

so assuming that the algorithm is right (which I do not have the skills to verify) and I implemented it correctly, yes, it should handle direction correctly

---

sidenote: I'm pretty sure using the term "coincide" is wrong (as I have been putting it in some of the docs). what is the correct mathematical term for when a vector has the same direction as another?

jni:
@kne42 no, the comment is more subtle than that, and you can't illustrate it with "point" vectors. Think now about rotating not a vector, but a human, 180 degrees. If you flip forward, your head will point towards the ground, but you will be facing backwards. If you flip right, your head will still point down, but you will be facing forward. So there is actually an infinite number of rotation matrices that will make your head point down, but the algorithm only returns one.

kne42:
@jni Ohhhhhhhhhhh I see. I just skimmed over the paper again and I do not see anything mentioning this.

@carterbox what are your suggestions? I can implement a different algorithm or maybe someone who understands math can help me out...

carterbox:
@jni, Yes! That's what I meant. Thanks for clarifying.

@kne42, My suggestions are to address the following edge cases:
1. Check that none of the vectors `src`, `dst` are the zero-vector because you can't align a vector with the zero-vector. It looks like your implementation would fail during normalization due to a division by zero. Not sure if you want explicitly check for zero-vectors to prevent that error.
2. Check that your implementation returns the identity matrix when `src` and `dst` are parallel.
3. Warn when `src`, `dst` are anti-parallel. In this case, there are many solutions, but you only return one. Alternatively, as in `polytope`, you can define an interface which takes a `dst` vector half-way along the desired rotation. This approach removes the ambiguity about how to rotate by 180 degrees, but it also might require the user to do more work when most of the time they are not rotating by 180 degrees? You could also just explain this alternative approach in the warning. Choose the approach that best fits the application here.

kne42:
@jni @carterbox How often do you think this would be an issue for someone? Does this warrant a warning over including something about this in the notes instead?

@carterbox So would the instructions be:
```python
M = compute_rotation_matrix(X, halfway_of_Y)
Y2 = M @ M @ X
```
?

carterbox:
From the [python docs:](https://docs.python.org/3/howto/logging.html)

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning

I think a warning is appropriate because I'm not sure that the half-angle approach is compatible with homogeneous coordinates. If `homogeneous_coords=True`, then the space would also be translated/scaled twice? I'm actually a little rusty when it comes to homogeneous coordinates. I just remember that the order of operations matters when you mix translations and rotations.

jni:
@carterbox the thing is, if I read the implementation correctly, it is *almost never* the shortest rotation: all rotations go via (1,0, ..., 0). I don't know what happens when you try to rotate (1, 0, 0) to (-1, 0, 0), but that would be an interesting test case, @kne42! =)

carterbox:
@jni I don't understand what you mean by "never the shortest rotation". All of the rotations are expressed as matrices of the same size, so they are all 'equally short' (computationally speaking).

carterbox:
I didn't mean to kill the momentum of this pull request! :sweat_smile: I just wanted to point out that there should be some extra checking of the inputs and because the solution to the rotation problem is non-unique when the start and end vectors are parallel.

jni:
@carterbox :joy: sometimes, complete discussion can have this effect. =D

The problem is that the issue you identified is in no way unique to anti-parallel rotations. For example, to point my head to the right, I can simply roll right, then my head will point right and I will face forward, or I can pitch forward and *then* roll (my reference frame) / yaw (external reference frame) right. Then I will be facing down, but head pointing right as before.

Unless I'm mistaken, for *any* rotation, there is an infinite number of ways to get there, and they will all leave the object/space in a different orientation. This PR always rotates from the source vector to the 0th axis, then to the target vector, which is different from going from source to target directly.

carterbox:
@kne42 @jni Let's back up to my earliest question about the desired application. This function claims to align two vectors in an ND space. If the final orientation of the space doesn't matter, and it only matters that the vectors are aligned. Then, just add some documentation for people not familiar with the non-uniqueness of this problem, and that's it. The function does what it claims. Done.

If the purpose of this function is to provide a general API for rotating ND spaces, then have a look at the docstring I wrote for [polytope](https://github.com/tulip-control/polytope/blob/b478461891c2bfc4eebbfc098628f57ee0752dea/polytope/polytope.py#L450-L478). In those docs, I note that for N>3 dimensions and when using only two vectors, it is only possible to define *simple* rotations (rotations whose motion can be projected into a single plane). Because the polytope API only promises *simple* rotations, using the half-rotation end-vector is enough to avoid the only ambiguous case for defining a plane (the case when the start and end vectors are co-linear). For *simple* rotations, the rotation matrix generating algorithm should be generating a rotation matrix in which all motion can be projected into a single plane of rotation. (I feel like there may be an easy test for this which involves row reduction or principal component analysis.)

If the user has a fully defined mapping from one orientation to another, a rotation function isn't even necessary because one can trivially construct the reorientation matrix by [mapping the basis vectors the starting space to the ending space](https://www.youtube.com/watch?v=kYB8IZa5AuE).

matthew-brett:

<details>

## Rationale

</details>

PR #3127, opened 2018-05-30
Update the documentation regarding datalocality

### LINKED ISSUE #2695: Wording on order of iteration is confusing in user guide

hmaarrfk:
Fixes #2695

Just my attempt at improving documentation.

## For reviewers

  later.


hmaarrfk:
@scikit-image/core backport?

PR #3137, opened 2018-05-31
Add initial airspeed velocity (asv) framework

### LINKED ISSUE #1061: Add Airspeed Velocity monitoring
I just watched Michael Droettboom's talk on his excellent and excellently named project, Airspeed Velocity, which tracks a project's benchmarks over time to ensure that benchmarks don't degrade. It seems pretty bloody awesome and I definitely think we should use it, though the logistics will be a bit tricky.

Talk link:
http://youtu.be/OsxJ5O6h8s0

Once we figure out where to run this, we can run 0.7 - 0.10 and then run master or each PR. (As Michael mentions in his talk, the PR stuff needs to be manual after code vetting or it will be a security risk. Running on a VM messes with the benchmarks.) (Obviously it will be a security risk even after vetting, but a much smaller one. =)

jni:
## Description
In order to prevent performance regressions, we should set up asv to benchmark our slowest functions. This branch exists to work on that goal.

Closes #1061

## For reviewers

  later.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,6 +7,8 @@
 - [ ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
 - [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
 - [ ] Gallery example in `./doc/examples` (new features only)
+- [ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
+  existing benchmark
 - [ ] Unit tests
 
 [For detailed information on these and other aspects see [scikit-image contribution guidelines](http://scikit-image.org/docs/dev/contribute.html)]
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -52,6 +52,7 @@
     // followed by the pip installed packages).
     //
     "matrix": {
+         "cython": ["0.27", "0.28"],
          "numpy": ["1.13", "1.14"],
          "scipy": ["1.0", "1.1"],
          "matplotlib": ["2.2"]
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -1,0 +1,140 @@
+{
+    // The version of the config file format.  Do not change, unless
+    // you know what you are doing.
+    "version": 1,
+
+    // The name of the project being benchmarked
+    "project": "scikit-image",
+
+    // The project's homepage
+    "project_url": "http://scikit-image.org/",
+
+    // The URL or local path of the source code repository for the
+    // project being benchmarked
+    "repo": ".",
+
+    // List of branches to benchmark. If not provided, defaults to "master"
+    // (for git) or "default" (for mercurial).
+    "branches": ["master"], // for git
+
+    // The DVCS being used.  If not set, it will be automatically
+    // determined from "repo" by looking at the protocol in the URL
+    // (if remote), or by looking for special directories, such as
+    // ".git" (if local).
+    "dvcs": "git",
+
+    // The tool to use to create environments.  May be "conda",
+    // "virtualenv" or other value depending on the plugins in use.
+    // If missing or the empty string, the tool will be automatically
+    // determined by looking for tools on the PATH environment
+    // variable.
+    "environment_type": "conda",
+
+    // timeout in seconds for installing any dependencies in environment
+    // defaults to 10 min
+    "install_timeout": 1200,
+
+    // the base URL to show a commit for the project.
+    "show_commit_url": "http://github.com/scikit-image/scikit-image/commit/",
+
+    // The Pythons you'd like to test against.  If not provided, defaults
+    // to the current version of Python used to run `asv`.
+    "pythons": ["3.6"],
+
+    // The matrix of dependencies to test.  Each key is the name of a
+    // package (in PyPI) and the values are version numbers.  An empty
+    // list or empty string indicates to just test against the default
+    // (latest) version. null indicates that the package is to not be
+    // installed. If the package to be tested is only available from
+    // PyPi, and the 'environment_type' is conda, then you can preface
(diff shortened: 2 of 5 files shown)


emmanuelle:
So I'm thinking about generating the benchmarks instead of writing each of them. I see several possibilities for that:

- turning every gallery example into a benchmark. The nice thing is that examples correspond to relevant usecases. The downside is that we probably don't want the plotting stuff in the benchmark.
- do some inference on input type and run functions on random values, like in the benchmarking scripts I wrote for the sprint.

What do you think?

jni:
@emmanuelle yes this is a very good plan, but now the autogeneration should somehow fit into the asv benchmark format, this is why I pushed things here. =)

TomAugspurger:
Once this is merged, we'll add an entry to https://github.com/TomAugspurger/asv-runner/blob/master/tests/full.yml and it'll start running automatically.

jni:
@TomAugspurger what happens when benchmarks themselves are added/removed/changed? Can one re-run a whole range of commits on benchmarks committed at the end?

TomAugspurger:
I'm not sure offhand, but I think ASV tracks the most recent commit, and
only runs newer ones.

It's not ideal, but I can manually run additional commits if you want to
build up a history.

hmaarrfk:
@TomAugspurger You are asking for alot. I'm not sure it is possible to build anything older than 0.14 and have the tests pass ;)

jni:
@hmaarrfk tests don't need to pass, only benchmarks.

TomAugspurger:
They'll show up at pandas.pydata.org/speed/

They're run nightly (Eastern time), but if this is merged in the next
couple hours I'll kick off a build now :)

soupault:
@jni could you, please, add a link from the comment above to somewhere (including the `release_dev.rst`)? I assume, both the contributors and the users might find it useful. Otherwise, lgtm. Thanks!

emmanuelle:
@jni if you add the comment requested by @soupault I will merge :-)

jni:
@emmanuelle @soupault done! =) 

@TomAugspurger how does asv work with PRs? The specific feature that I'm interested in is the ability to run even a new benchmark (provided by the PR) on the master branch and the PR tip (or the merge commit) *before* merging. Is this easy enough to do?

jni:
@TomAugspurger that sounds like a fun challenge. And @Carreau has done the hard work with https://github.com/MeeseeksBox/MeeseeksDev. Can you give me a bit of background about what the bot would need to do? ie if I have a head commit and the master branch it split from, how do I:

- grab the benchmarks and asv conf from the head commit
- run the benchmarks on both
- publish the results somewhere

?

TomAugspurger:
@jni As far as running goes, yes I think you've got it. You should be able to clone the PR and run `asv continuous -f 1.1 HEAD master` and that'll give you a text output that can be published on the issue. The bot could optionally take a `-b` flag to limit it to a certain set of benchmarks.

I think the main concern is scheduling these so that we aren't running these on-demand runs with

a.) eachother
b.) the nightly benchmarks

The benchmark machine is using Airflow right now to do the scheduling. Airflow supports external triggers, so this should be possible. I can help out with the airflow stuff if you want.

jni:
@emmanuelle if we merge this soon we can beat the 4th anniversary of #1061. =P

hmaarrfk:
@jni 
image

jni:
@hmaarrfk I know :joy: :sob: :joy:

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -7,6 +7,8 @@
 - [ ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
 - [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
 - [ ] Gallery example in `./doc/examples` (new features only)
+- [ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
```
stefanv: Do we really want a benchmark for every new piece of functionality?

stefanv:
Thank you, Juan!

.github/PULL_REQUEST_TEMPLATE.md
jni: @stefanv as with all of these tick marks, they are at the reviewers' discretion. =) The idea should certainly be considered for all PRs. And the idea is to build a suite over time. So anything that touches e.g. dilation, gaussian filters, watershed, etc should add a benchmark, and that benchmark should be run before and after the change, so that skimage moves towards speed as a whole.

hmaarrfk:
@jni good job on the asv PR.

For those looking to install asv locally, we need to wait either for 0.2.2 or install with pip and git

```
pip install git+https://github.com/airspeed-velocity/asv.git#egg=asv
```

jni:
@TomAugspurger none of the benchmarks added after this PR appear on the ASV server! Any ideas why? I looked at `benchmarks/__init__.py` and `asv.conf.json` but there's nothing there to suggest that new benchmarks shouldn't be run. Is there some asv cache that needs clearing when new benchmarks are added?

TomAugspurger:
I think my 2 year old hit the power button on the benchmark machine, and I didn't notice till last night 😆 It should be catching up now, and we've had a talk about touching Daddy's stuff.

jni:
@TomAugspurger I feel like this is a longer term problem of not downloading from master. New benchmarks were added weeks ago...

jni:
Oh although I just reread your statement that you think it's a longer term problem on your end also. =P Ok let's see how we go... =)

TomAugspurger:
Found the problem. When we initially set this up, I manually made the change adding a `benchmarks/__init__.py` on the ASV machine. But I never cleared that change, so the `git pull` wasn't working for scikit-image. Fixed that, and am re-running.

TomAugspurger:
http://pandas.pydata.org/speed/scikit-image/ seems to be up to date now.

jni:
:tada: thanks @TomAugspurger!

PR #3145, opened 2018-06-02
Proposal: use python-versioneer for version number.

### LINKED ISSUE #3144: Python versioneer?
I wanted to ask if you all thought it was applicable to use [python-versioneer](https://github.com/warner/python-versioneer) to simplify the management of our versions.

It will automatically increment our dev versions the version according to the latest release tag as well as "how many commits we are ahead of it".

Furthermore, we get the skip the "increment the version number in the source".

This suggestion came from @jakirkham who uses it for his own projects and suggest that I use it for some of mine.

If you are all for it, I will try to create a PR to implement this feature.

hmaarrfk:
Fixes #3144. Please follow the discussion on the issue itself.

Needs the addition of a tag on the master branch.
I think the appropriate commit for the `0.15.0.dev` tag is be42f61150cc5d16c6618285d908d24f904b949e (the last 0.14.0.dev commit).

I made this PR while I remembered how to install versioneer.

## For reviewers

  later.


skimage/__init__.py
```diff
@@ -62,12 +62,13 @@
 import warnings
 import sys
 
+from ._version import get_versions
```
jni: I'm not a fan of the `del`... maybe `from ._version import get_versions as _get_versions`?

requirements/build.txt
```diff
@@ -1,2 +1,3 @@
 Cython>=0.23
 wheel
+versioneer
```
jni: It appears you are vendoring versioneer while also adding it as a dependency?

RELEASE.txt
```diff
@@ -41,11 +41,22 @@ make a PR to update these notes after you are done with the release. ;-)
   - Discuss with others, and make the changes directly to v<major>.<minor>.x branch.
   - Once the consensus is found, ask the project maintainers to merge the PR.
 
-- In the master branch, update the version number in ``skimage/__init__.py``
-  to the next ``-dev`` version, commit, and push.
+  - In the master branch, create a new tag with an updated version number and
+    append ``.dev`` to it. python-versioneer will take care of updating the
+    version number in the source file.
```
jni: What exactly is happening here? I'm specifically confused about why there is no version number anywhere in the changes... Is the version number always determined "live"? What about when we distribute the source without the git history?

requirements/build.txt
hmaarrfk: hmm, I think you are right. We don't need it as a dependency. Good catch.

skimage/__init__.py
hmaarrfk: Ok. That was the boilerplate code from versioneer.

jni:
:joy: great commit message

RELEASE.txt
hmaarrfk: sdist seems ok, see issue for more [details](https://github.com/scikit-image/scikit-image/issues/3144#issuecomment-394112573).

Review comments seem to get hidden eventually and this is a real concern.

jni: @hmaarrfk yes it is, there are quite a few rants about GitHub's review process floating around. But it's simple and very useful in the simple cases so I cut them a bit of slack.

stefanv:
I am concerned that versioneer adds complexity without enough of a benefit.  At the moment, our versioning is explicit and simple.  It is all in one line of code in `skimage/__init__.py`.  It's not difficult to update, and there's zero magic.  So I am leaning towards -1 on this proposal.

hmaarrfk:
yeah, so am i. Versioneer just added like 2k * 2 lines of code for marginal value.
Eventually, it will have to be maintained and will cause annoyances.

If this issue is ever brought up again, [miniver](https://github.com/jbweston/miniver) is a project that does similar things, but with less lines of code (complexity).

jni:
That's too bad. It's nice to have git hashes in the version. But I have to agree that 4K lines is at least one too many. =P

@hmaarrfk I must commend you on your ability to throw away dead ends at the drop of a hat. This is a very useful skill that I don't quite have: I tend to flog at the dead horse for quite a while before accepting its death. =P

PR #3147, opened 2018-06-03
Maintenance: cleanup build process and `__init__.py`

### LINKED ISSUE #3155: Maintenance: More simplification of `__init__.py`
## Description
In my mind, `__init__.py` is very cluttered. Here is a breakdown of the lines of code in `__init__.py`.

Happy to do PRs for any of the suggestions that get approved.

| Lines | Purpose | Fraction of file | Alternative |
|-------|---------|--------|--------|
| 1-57  | Documentation   | large| Keep, this is great!  |
| 59, 65-66 | data_dir  | small  | ☑ [data_dir] Fixed in  #3161 |
| 68   | version  |   | Keep, this is great!  |
| 71-88   | python version  |  medium  | ☑ [python version] Fixed in #3160 |
| 91-131 | Test  |  large | ☑ [pytest] Fixed in #3164 |
| 134-178 | `__SKIMAGE_SETUP__` hack |  large | ☐ Addressed in #3158  |
| 181-192 | Numpy cross/docs | medium | ☑ [lookfor] Fixed in #3162 |

With refactoring, the file would look like:

```python
"""Docs
[...]
"""
__version__ = "0.15.dev0"

from .util.dtype import *
from .data import data_dir

from .util import lookfor
```

### [datadir] 
I suggest:
  - `from .data import data_dir`
  - Or just tell users to `from skimage.data import data_dir as skimge_datadir` as they are only likely to use this in the initial tutorial phase of their work.

### [python version] 
I think this should be removed.
Also, this problem shouldn't happen much most users will be installing scikit-image from pip/conda where 0.15 won't even exist in python2 repos.

Alternatively, it should point to `http://scikit-image.org/docs/dev/install.html` and not duplicate the instructions IMO.

### [pytest] 
Instead of having our own wrapper for pytest, just tell users to run the pytest command themselves.

From the terminal: `python -m pytest --pyargs skimage`
From a python console: `import pytest; pytest.main('--pyargs skimage')`

The reasoning is:
  1. It never worked before. See PR #3152.
  2. It was never tested before #3152.
  3. I don't think you should have wrappers for 1 liners.

### [lookfor] 
This one is tricky because it uses introspection.

Maybe move to `skimage.utils.docs` but it uses introspection, so the code would have to be rewritten too.

hmaarrfk:
I cleaned up the build process quite a bit.

Subsequently, I continued to cleanup the `__init__.py` file quite a bit in accordance with @stefanv's comments in Issue #3155.

Here is a summary of the changes:

### No use of the `WHEELHOUSE`. 
This seemed like a lot of hard work from @matthew-brett. The last dependency that used it was pyside (not pyside 2) and that was only used in in python <=3.4. We now support only 3.5 therefore this isn't necessary.
Maybe we should point to some of matthew-brett's instructions in case future maintainers need this? I fear that if we leave the instructions in but don't really use them, then the instructions kept on scikit-image will become too old to be useful.

### No more `__SKIMAGE_SETUP__`
This required me to move the cython helper untils to the `tools` directory. I think this is an equally good solution for the cython helper utilities, and a much better solution in terms of poluting the install procedure with the use of global variables.
  * Reasoning: the choice seems to not want to have cython as a build time dependency for users. As such, this cython helper script has no place in the source distribution and is only necessary for those cloning from git.
  * Also added the script to the MANIFEST.in

### No more random `pip install` commands in the travis scripts
All packages that need to be installed by pip are now in some of the requirement files. I had to duplicate `numpy` in both build and defaults. Thought I think this is OK.

### Consistent cython version
Sometimes cython was specified as `0.23` and othertimes as `0.23.4`. Everything now specifies `0.23.4`. This version is sprinkled in many locations, so I have a feeling this will be a problem in the future too.

### Build procedure happens before we install `defaults.txt`
This now checks that we really only require cython and numpy to build the system.

### I removed the `retry` bash helper function
It was never used. I'll add it back on request.

### Minimum requirements now locks down all listed requirements
Previously, this was only applied to defaults.txt

### Normalized the package number 
Now set to `0.15.dev0` instead of `0.15dev` since you keep getting this warning otherwise:
```
[...]/site-packages/setuptools/dist.py:398: UserWarning: Normalizing '0.15dev' to '0.15.dev0'
```

### Fix the build process from `sdist` without `cython`
Previously if you didn't have cython installed, the build process from the `sdist` would fail because it wouldn't look for the existance of `_haar.cpp` as it only checked for `_haar.c`.

### `data_dir` now defined in `data/__init__.py`
and import via
```python
from .data import data_dir
```

### Python version checking removed
This should never happen as it will not be uploaded to pypi nor conda with for Python 2.7.

### Test
Until  PR #3152 was merged, this function didn't actually work. Many of the build scripts just called `pytest` directly.
The only one that I found that didn't call `pytest` directly was the `conda-forge` recipe, which will be fixed when [this PR](https://github.com/conda-forge/scikit-image-feedstock/pull/25) is merged.

### `lookfor`
The code for this has been moved to `skimage.util.lookfor`. I don't think code should be in the `__init__.py` file, let alone the top level one.

Fixes #3155

## For reviewers

  later.


hmaarrfk:
I don't understand codecov. what did I miss in my patch?

JDWarner:
Coverage percent reduction appears to simply be due to a high-ish number of covered lines being removed, so proportionally the package level coverage decreased.  I don't think this should be considered blocking.

This looks like a dramatic simplification/cleanup.

hmaarrfk:
Question: Why is `requirements/tests.text` inside the base `requirements.txt`

doc/release/release_dev.rst
```diff
@@ -47,6 +47,8 @@ API Changes
   and ``skimage.transform.rescale`` has been set to ``reflect``.
 - Default value of ``anti_aliasing`` parameter in ``skimage.transform.resize``
   and ``skimage.transform.rescale`` has been set to ``True``.
+- Removed the ``skimage.test`` function. This functionality can be acheived
+  by calling ``pytest`` directly.
```
jni: I understand the sentiment, but I think we don't want to do this. The reason this exists is that we used to use nosetests and were bitten when it reached end-of-life. `skimage.test` allows us to keep the same testing syntax even if we decide to switch frameworks later on.

skimage/__init__.py
```diff
@@ -56,140 +56,7 @@
...
-
-    """
-    import numpy as np
-    import sys
-    return np.lookfor(what, sys.modules[__name__])
-
-
-del warnings, functools, osp, imp, sys
+__version__ = '0.15.dev0'
+from .util.dtype import *
```
jni: NEVER import *

jni:
@hmaarrfk as a general rule, could you try to keep your PRs more modular? This is very hard to review because it's hard to know what each change is doing, relative to your long list of proposed changes. Many changes are completely unrelated. For example, the `skimage.lookfor` fix is a slam dunk and could get in with minimal review, but it is held up by the much more complex remaining changes.

I feel like develop doing `--no-deps` by default is probably an anti-pattern. I've been bitten by it updating a buttload of dependencies but also by running `no-deps` too eagerly and missing out on e.g. PyWavelets. =\

doc/release/release_dev.rst
jni: also, "achieved". =P

I'm neither +1 or -1 on this btw. Running individual tests is more natural if we normalise with pytest. Just wanted to flag it since others (@soupault) have in the past preferred strongly hiding pytest (using skimage.testing.raises instead of pytest.raises, for example).

jni:
The data_dir change is also a slam dunk.

doc/release/release_dev.rst
hmaarrfk: @jni, I understand that the idea was to have the testing localized, but after having going through a few of the other build procedures, it seems that `skimage.test` was never used in the first in place.

When I tried `skimage.test` myself a few months back, I thought I wasn't using the function correctly, but a recent PR showed me that it wasn't just me.

I'm glad to go back to the old functionality, but I feel like the ship has sailed on that, even before my PR. I didn't have to change any of the other tests when I removed this function.

skimage/__init__.py
hmaarrfk: I'm with you on that one, but that line of code was buried so deep in the past that it was impossible to find it.

https://github.com/scikit-image/scikit-image/blob/master/skimage/__init__.py#L178

hmaarrfk: There remains the question on what is the expected functionality. I left that one as a "Good first issue" in Issue #3156

hmaarrfk:
@jni, I agree that many of the changes are unrelated, unfortunately, they mostly touch 3 files:
`travis.yml`
`setup.py`
`__init__.py`

If i broke them up into different PRs, then during merging/rebasing, it would be rather impossible to tell what is going on as there would be many conflicts. The changes would have to be tripple checked to ensure that we don't undo some of the previous ones.

In fact, I had first broken up the [`skimage.test`](https://github.com/hmaarrfk/scikit-image/pull/1) and tested the diffs on my own conda recipe, but I was unable to merge it since many of the lines were overlapping some previous diffs here.

To me, these changes all fall into: `removing lines of code that are never touched, either by tests, documentation, or by sensible end users`.

In either case, let me give it a shot in a few hours.

PR #3148, opened 2018-06-03
WIP: Overhaul `transform` module

kne42:
## Refactorizations & Deprecations
- [ ] refactor `transform` module to `transform_xy`
- [ ] create new `transform_rc` module
- [ ] create wrapper `transform` module around `transform_xy` with automatic deprecation messages

## Changes (implemented in `transform_rc`)
- [ ] switch to numpy-style coordinates (row, column)
- [ ] add nD support
- [ ] improve transformation machinery using `scipy`'s `LowLevelCallable`s

## Other
- Implementation notes/benchmarks may be found in [this notebook](https://gist.github.com/kne42/b487af595332db468654a5c5b1ce416c).
- Core `LowLevelCallable` notes may be found in [this notebook](https://gist.github.com/kne42/6271a42967d8232b32b0b3e918761921).


kne42:
cc @jni

jni:
@kne42 Since this is going to touch so many files, it's important for the commit messages to be extremely thorough, and for the commits to be nicely separated. Specifically, I suggest making one commit for moving transform/ to transform_xy/, another for creating the alias for transform -> transform_xy, together with a deprecation message, and another for creating transform_rc.

Also, for the initial transform_rc "port", you could simply use the existing warps functions, but transpose the coordinates. This would let us actually merge this and then add nD capabilities little by little.

What do you think?

kne42:
@jni
> Since this is going to touch so many files, it's important for the commit messages to be extremely thorough, and for the commits to be nicely separated. Specifically, I suggest making one commit for moving transform/ to transform_xy/, another for creating the alias for transform -> transform_xy, together with a deprecation message, and another for creating transform_rc.

While I do agree on making the commit messages thorough, I typically prefer to group commits together with message prefixes and then squash them prior to a merge. However, I suppose that this, being the massive change that it is, warrants individual commit messages for each task at hand.

> Also, for the initial transform_rc "port", you could simply use the existing warps functions, but transpose the coordinates. This would let us actually merge this and then add nD capabilities little by little.

Oooh yes I like this idea! Cleans up PR history too :P

kne42:
Oopsies, that was an accident!

Anyways, @jni lmk if you spot me missing anything! This is a pretty large change so I'm not quite sure if I covered everything necessary/went too far.

Also, for the functions that are already nD/take in full-on images instead of just coordinates, do you think that we should put them in `skimage.transform` and just import them into `skimage.transform_xy` and `skimage.transform_rc`?

skimage/transform_xy/_warps.py
```diff
@@ -208,7 +208,7 @@ def rescale(image, scale, order=1, mode='reflect', cval=0, clip=True,
     ----------------
     order : int, optional
         The order of the spline interpolation, default is 1. The order has to
-        be in the range 0-5. See `skimage.transform.warp` for detail.
+        be in the range 0-5. See `skimage.transform_xy.warp` for detail.
```
jni: I'm concerned about the documentation. Generally, when we deprecate, we want to point to the *new* way of doing things. The only reason to keep the old way hanging around is for backwards compatibility. But anyone new who checks the docs should be pointed to the One True Way of doing things, and that's the new way.

lagru:
Hey everyone, what is the status of this PR? Is this something you or we as a team want to pick at some point? I'm asking because I'm preparing a proposal for an Outreachy project that would focus on our `transform` module.

Issue #2178, opened 2016-07-03
The output of adaptive methods in try_all_threshold is a bit misleading


emmanuelle:
In `filters.try_all_threshold`, two local methods are presented: `filters.threshold_adaptive` and "local otsu", which is defined internally in `try_all_threshold`. When the user looks at the graphical output, he/she does not known how to use the local otsu method. Also, `threshold_adaptive` uses as default parameter `offset=10`, which works well for the page image, but is quite for arbitrary for other images.

I was the one to push for showing local methods in `try_all_threshold`, so I guess I'm partly responsible for this problem :-). Should we separate local algorithms in a `try_all_local_threshold` that would test more parameters such as two block sizes and three values of offset? And/or should we define a function that would transform any thresholding function into a local thresholding method, like the `_include_selem` that is currently an internal function?

sciunto:
No problem... until we release a new version :)

If I sum up, we need to remove local algo. That's for sure. 

For the rest, I'm not 100% convinced so far.
For `try_all_local_threshold`, what would happen the day we add new algorithms (and maybe additional parameters)? I'm afraid it's going to be out of control.
For your second option, it sounds like a new feature to me.

sciunto:
@emmanuelle I started a PR.

sciunto:
@emmanuelle I let you choose if you want to open another issue (or keep this one) about local thresholding. My PR solves the problem partially.

sciunto:
given the current output, I'm wondering if it's not better to rename the function as `try_all_global_threshold`. Otherwise, it's confusing.

jni:
Hmm, I quite like having both local and global in the same function... I suspect there would be a way to achieve this consistently...

sciunto:
If I sumarize. For now, it's not clear how to support local thresholding.

Do we vote pro or con a renaming `try_all_global_threshold`?

sciunto:
@scikit-image/core needs decision here as well.

blink1073:
Why just allow additional threshold methods to be passed in as a list of `(name, function)` tuples?

jni:
@blink1073 I quite like that proposal!

jni:
Given that we merged #2180, can we remove the 0.13 tag on this one? Adding a new fancy generic API for `try_all_threshold` can wait till the next release, I believe.

sciunto:
I'm +1.

blink1073:
Moved to 0.14.

stefanv:
[...]

PR #3149, opened 2018-06-04
Recover from exceptions in filters.try_all_threshold()

GenevieveBuckley:
## Description
Currently if filters.try_all_threshold() encounters an exception in one of the threshold methods it fails to complete. I suggest instead the output should note which kind of exception was encountered and continue. 
Here's an example to illustrate:
```
%matplotlib inline
import numpy as np
from skimage.filters import try_all_threshold

simple_image = np.array([[0, 0, 1, 3, 5],
                         [0, 1, 4, 3, 4],
                         [1, 2, 5, 4, 1],
                         [2, 4, 5, 2, 1],
                         [4, 5, 1, 0, 0]], dtype=int)
flat_image = np.ones(simple_image.shape).astype(np.uint8)

try_all_threshold(simple_image)
try_all_threshold(flat_image)
```

### Proposed behaviour:
Output of `try_all_threshold(simple_image)`:
proposed_tryallthreshold_flatimage
Output_of `try_all_threshold(flat_image)`:
proposed_tryallthreshold_simpleimage

### Current behaviour:
Output of `try_all_threshold(simple_image)`:
current_tryallthreshold_flatimage
Output_of `try_all_threshold(flat_image)`:
current_tryallthreshold_simpleimage

## Checklist
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/tests/test_thresholding.py
+++ b/skimage/filters/tests/test_thresholding.py
@@ -1,3 +1,4 @@
+import pytest
 import numpy as np
 from scipy import ndimage as ndi
 
@@ -15,6 +16,7 @@ from skimage.filters.thresholding import (threshold_local,
                                           threshold_mean,
                                           threshold_triangle,
                                           threshold_minimum,
+                                          try_all_threshold,
                                           _mean_std)
 from skimage._shared import testing
 from skimage._shared.testing import assert_equal, assert_almost_equal
@@ -27,6 +29,16 @@ class TestSimpleImage():
                                [1, 2, 5, 4, 1],
                                [2, 4, 5, 2, 1],
                                [4, 5, 1, 0, 0]], dtype=int)
+
+    def test_minimum(self):
+        with pytest.raises(RuntimeError):
+            threshold_minimum(self.image)
+
+    def test_try_all_threshold(self):
+        fig, ax = try_all_threshold(self.image)
+        all_texts = [axis.texts for axis in ax if axis.texts != []]
+        text_content = [text.get_text() for x in all_texts for text in x]
+        assert 'RuntimeError' in text_content
 
     def test_otsu(self):
         assert threshold_otsu(self.image) == 2
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -59,8 +59,11 @@ def _try_all(image, methods=None, figsize=None, num_cols=2, verbose=True):
 
     i = 1
     for name, func in methods.items():
-        ax[i].imshow(func(image), cmap=plt.cm.gray)
         ax[i].set_title(name)
+        try:
+            ax[i].imshow(func(image), cmap=plt.cm.gray)
+        except Exception as e:
+            ax[i].text(0.5, 0.5, "%s" % type(e).__name__, ha="center", va="center", transform=ax[i].transAxes)
         i += 1
         if verbose:
             print(func.__orifunc__)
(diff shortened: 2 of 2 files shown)


jni:
@GenevieveBuckley I like it! But I think you probably need to add a test that triggers this, like your flat image. See: https://codecov.io/gh/scikit-image/scikit-image/pull/3149/diff

GenevieveBuckley:
@jni - a test is a good idea, especially since there currently isn't one for try_all_threshold().
I know how to test for an exception we expect will be raised. What I don't know how to do is test an exception has been raised **and handled appropriately**. Without the second half I get an improperly passing test all the time.

Is there a resource you can point me towards so I can work it out? My googling hasn't been especially successful so far.

emmanuelle:
You can put the above code in a test function: with master the test would fail because of an Error, and with your changes it will not fail. You do not always need to test an equality, just the fact that the function does not raise an error is already a good test IMHO (maybe just leave a comment inside the function).

emmanuelle:
It was mentioned in #2178 that ``try_all_threshold`` should return a dictionary and not a plot, and I agree. However, this PR is already an improvement so I'm in favor of merging it once a test has been added. Thanks @GenevieveBuckley !

stefanv:
I like it! Approved conditional on a running test.

jni:
@GenevieveBuckley since it returns the axes object, you can look for specific text in `ax.texts`:

```python
In [1]: import matplotlib.pyplot as plt
In [2]: fig, ax = plt.subplots()
In [3]: ax.text(0.5, 0.5, 'hello')
Out[3]: Text(0.5,0.5,'hello')
In [4]: ax.texts
Out[4]: [Text(0.5,0.5,'hello')]
```

I've found that introspection of Matplotlib objects is extremely powerful!

GenevieveBuckley:
Test case implemented using @jni's idea to inspect the matplotlib text (thanks!)

@emmanuelle I hadn't heard that conversation about returning a dictionary (https://github.com/scikit-image/scikit-image/issues/2178). I could look into that, but it won't be today. Do you have a preference for @stefanv's suggestion of `visiualize=True` flag, or having them be two separate functions?

jni:
As we are aiming to remove our matplotlib dependency in the medium term, probably the right approach is to have separate functions for visualising, rather than a keyword argument. This will make it easier to contain all mpl functionality in a single sub-package.

skimage/filters/tests/test_thresholding.py
```diff
@@ -1,3 +1,4 @@
+import pytest
```
soupault: Sorry, I'm a bit late...
A while ago we decided not to use direct testing framework imports in the library, and to wrap those into `skimage._shared.testing`.

jni: @soupault sure, but as you know, this is not set in stone:

https://github.com/scikit-image/scikit-image/pull/3164#issuecomment-395649448

This might be a decision we revisit soon.

soupault: @jni Hmm... Do you have in mind a better option? My thinking is that for that particular case (`skimage.test`) it is more convenient to use `pytest` directly because it allows one to run the tests for (1) the full package, (2) modules, (3) single tests, and (4) documentation using a similar syntax (also, the usage is already well-described in the guidelines - http://scikit-image.org/docs/stable/contribute.html#testing). The legacy solution only implemented (1) and (4), and had to be tested itself.

soupault:
@GenevieveBuckley thanks a lot for the contribution, in any case!

PR #3158, opened 2018-06-08
Move skimage._build.py to tools/skimage_build_helpers.py

hmaarrfk:
This allows us to remove a lot of the convoluted logic in the `__init__.py`

in doing this, I also

### Fix the build process from `sdist` without `cython` 
Previously if you didn't have cython installed, the build process from the `sdist` would fail because it wouldn't look for the existance of `_haar.cpp` as it only checked for `_haar.c`.

### Todo
If #3221 is accepted before hand, there is a comment in the `before_install.sh` script that should be updated.
## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -56,12 +56,6 @@ img_as_ubyte
 
 """
 
-import imp
-import functools
-import warnings
-import sys
-
-
 __version__ = '0.15.dev0'
 
 from ._shared.version_requirements import ensure_python_version
@@ -80,8 +74,10 @@ Your install of scikit-image appears to be broken.
 Try re-installing the package following the instructions at:
 http://scikit-image.org/docs/stable/install.html """
 
-
-def _raise_build_error(e):
+try:
+    from ._shared import geometry
+    del geometry
+except ImportError as e:
     # Raise a comprehensible error
     local_dir = osp.split(__file__)[0]
     msg = _STANDARD_MSG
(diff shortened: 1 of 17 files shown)


hmaarrfk:
@jni, could you re-trigger this build please. Maybe I forced pushed.

skimage/__init__.py
```diff
@@ -131,51 +131,7 @@ def _test(doctest=False, verbose=False):
...
-    # Raise a comprehensible error
-    local_dir = osp.split(__file__)[0]
-    msg = _STANDARD_MSG
-    if local_dir == "skimage":
-        # Picking up the local install: this will work only if the
-        # install is an 'inplace build'
-        msg = _INPLACE_MSG
-    raise ImportError("""%s
-It seems that scikit-image has not been built correctly.
-%s""" % (e, msg))
```
jni: I don't think it's wise to get rid of this completely. These messages were added so that users trying to run skimage from source without compiling didn't get an opaque import error like `no module _foo_cy`.

hmaarrfk: Hmm ok, maybe we should refractor this more like [scikit-learn's](https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/__check_build) which has it in a fully seperate directory that compiles a specific dummy pyx file.

hmaarrfk: So am I allowed to copy [90% of scikit-learn's code for this check](https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/__check_build)? Would I have to include their "COPYING" notice in the specific folder?

They use the [New BSD License](https://github.com/scikit-learn/scikit-learn/blob/master/COPYING)
```
  a. Redistributions of source code must retain the above copyright notice,
this list of conditions and the following disclaimer.
```

hmaarrfk:
@jni, regarding your previous comment, I think it is best to keep that refractoring for a different PR where we can think of the best way to address an improper build. Scikit-Learn actually builds a module that is guaranteed to ALWAYS be in Cython just for checking if the package as a whole was compiled.

The check now isn't so hacky, as much as it is visual noise, and I think it is better to get the hacky code (with `builtins` and `__SKIMAGE_SETUP__`) out first as was the original intent.

stefanv:
This looks fishy to me: tools is not part of the distributed package, but now we have imports in `_shared` that rely on it.  This may technically work but doesn't feel right to me.  What was wrong with using `_build` before?

hmaarrfk:
The build scripts shouldn't be part of the package themselves.

Python was having a hard time because half the package wasn't built yet. When it crawleded through the namespace to find things, it would raise errors. That is why there were about 10 lines of comments explaining that this was a "hacky solution". We could put it somewhere else.

stefanv:
Maybe I didn't look carefully enough. If this is used only in the various 
setup.py files, you are probably right.

hmaarrfk:
As a note, this "hacky" (not my words) solution is also in scikit-learn. 

My main goal is to get the hacky comments and partial loading out of the package, because it really doesn't belong there, and is definitely not pythonic.

My proposal also includes hacky code to check for `.c` and `.cpp` files.
So if you accept a proposal along those lines, I would have to clean that code up a little bit too.

stefanv:
@hmaarrfk This PR is still mentioned in `before_install.sh`. Should that be updated?

PR #3159, opened 2018-06-08
Modernize build process for use of pypi wheels

hmaarrfk:
### Removed wheelhouse
This seemed like a lot of hard work from @matthew-brett. The last
dependency that used it was pyside (not pyside 2) and that was only used
in in python <=3.4. We now support only 3.5 therefore this isn't
necessary.
Maybe we should point to some of matthew-brett's instructions in case
future maintainers need this? I fear that if we leave the instructions
in but don't really use them, then the instructions kept on scikit-image
will become too old to be useful.

### No more random pip install commands
All packages that need to be installed by pip are now in some of the
requirement files. I had to duplicate `numpy` in both build and
defaults. Thought I think this is OK.

### Constistent cython versioning
Sometimes cython was specified as `0.23` and othertimes as `0.23.4`.
Everything now specifies `0.23.4`. This version is sprinkled in many
locations, so I have a feeling this will be a problem in the future too.

### Build happens before we install `requirements/default.txt`
This now checks that we really only require cython and numpy to build
the system.

### Pin all versions in `requirements` when MINIMUM = 1
Previously, this was only applied to defaults.txt

## Checklist
- [ ] Unit tests

## For reviewers

  later.


soupault:
RIP coverage, apparently.

hmaarrfk:
I just triggered a new build, lets see what happens.

hmaarrfk:
This makes no sense.

It claims that `osp` is still in `skimage.__init__.py`
```
[33mDuring handling of the above exception, another exception occurred:[0m
[1m[31m../../../venv/lib/python3.6/site-packages/py/_path/local.py[0m:668: in pyimport
[1m    __import__(modname)[0m
[1m[31mskimage/__init__.py[0m:113: in <module>
[1m    _raise_build_error(e)[0m
[1m[31mskimage/__init__.py[0m:86: in _raise_build_error
[1m    local_dir = osp.split(__file__)[0][0m
[1m[31mE   NameError: name 'osp' is not defined[0m
```

hmaarrfk:

```
set -ex
```

tools/travis/before_install.sh
```diff
@@ -48,26 +23,17 @@ retry () {
     return 0
 }
 
-# add build dependencies
-echo "cython>=0.23.4" >> requirements/default.txt
-echo "numpydoc>=0.6" >> requirements/default.txt
```
hmaarrfk: Is numpydoc really a build dependency?
It pulls in ALOT of things.

hmaarrfk:
Seems like this is getting too "omnibus" like. I'm going to break this down so as to reduce the chances of a regression.

PR #3160, opened 2018-06-08
Refractor python version checking.

hmaarrfk:
~~Developers will be knowledgeable enough to know that it isn't python2
compatible.~~

~~The wheels won't be available in pypi or conda for python 2~~

This removes a bunch of visual noise from the `__init__.py` file.

## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -87,27 +87,6 @@ Newer versions of these tools avoid installing packages incompatible
 with your version of Python.
 """)
 
-if sys.version_info < (3, 5):
-    raise ImportError("""
-
-You are running scikit-image on an unsupported version of Python.
-Likely Python 2.
-
-Unfortunately, scikit-image 0.15 and above no longer work on this
-version of Python.  You therefore have two options: either upgrade to
-Python 3.5, or install an older version of scikit-image. For Python 2.7, use
-
- $ pip install 'scikit-image<0.15'
-
-Please also consider updating `pip` and `setuptools`:
-
- $ pip install pip setuptools --upgrade
-
-Newer versions of these tools avoid installing packages incompatible
-with your version of Python.
-""")
-
-
 # Logic for checking for improper install and importing while in the source
 # tree when package has not been installed inplace.
 # Code adapted from scikit-learn's __check_build module.
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -87,14 +87,15 @@ Newer versions of these tools avoid installing packages incompatible
 with your version of Python.
 """)
 
-if sys.version_info < (3,):
+if sys.version_info < (3, 5):
     raise ImportError("""
 
-You are running scikit-image on Python 2.
+You are running scikit-image on an unsupported version of Python.
+Likely Python 2.
 
 Unfortunately, scikit-image 0.15 and above no longer work on this
 version of Python.  You therefore have two options: either upgrade to
-Python 3, or install an older version of scikit-image using
+Python 3.5, or install an older version of scikit-image. For Python 2.7, use
 
  $ pip install 'scikit-image<0.15'
 
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -67,8 +67,8 @@ data_dir = osp.join(pkg_dir, 'data')
 
 __version__ = '0.15dev'
 
-from ._shared.version_requirements import check_python_version
-check_python_version((3, 5))
+from ._shared.version_requirements import ensure_python_version
+ensure_python_version((3, 5))
 
 # Logic for checking for improper install and importing while in the source
 # tree when package has not been installed inplace.
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -75,7 +75,7 @@ Likely Python 2.
 
 Unfortunately, scikit-image 0.15 and above no longer work on this
 version of Python.  You therefore have two options: either upgrade to
-Python 3.5, or install an older version of scikit-image using
+Python 3.5, or install an older version of scikit-image. For Python 2.7, use
 
  $ pip install 'scikit-image<0.15'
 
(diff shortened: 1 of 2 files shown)


jni:
I'm -1 on this change. This will lead to a bunch of confusion for users. See [Python 3 Statement: Practicalities](https://python3statement.org/practicalities/) under "Fail early at import time". Feel free to move this into a utility function though, `assert_valid_python3_version()`. Actually we should check for Python >= 3.5, not just 3...

hmaarrfk:
Good point.

I guess I stopped reading at "Fail early at install time".

I'm just really against the 20 lines of code this adds to what is supposed to be the starting point of the code. I know people don't like looking at source, but the source is really the only reliable form of information, and python makes it really to document the source.

Maybe a 
```
from .util.check_python_version import ensure_python3
```

hmaarrfk:
@jni, see if you like this. If not, just close this PR.

You convinced me to keep this.

The other solutions would create more file pollution.

skimage/__init__.py
```diff
@@ -68,14 +68,15 @@
 __version__ = '0.15dev'
 
 
-if sys.version_info < (3,):
+if sys.version_info < (3, 5):
```
soupault: JFYI, there was a discussion on the best approach for version comparison - https://github.com/scikit-image/scikit-image/pull/2975

jni: @soupault I think that discussion is more relevant to third-party packages where the version string is less predictable than Python system versions

hmaarrfk:
@jni this is what the output looks like on a conda python2 environment:

```python
(py2) \git\scikit-image>python
Python 2.7.15 | packaged by conda-forge | (default, May  8 2018, 14:46:56) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import skimage
  File "skimage\__init__.py", line 71, in <module>
    check_python_version((3, 5))
  File "skimage\_shared\version_requirements.py", line 28, in check_python_version
    """ % '.'.join([str(v) for v in min_version]))
ImportError:

You are running scikit-image on an unsupported version of Python.
Likely Python 2.

Unfortunately, scikit-image 0.15 and above no longer work on this
version of Python.  You therefore have two options: either upgrade to
Python 3.5, or install an older version of scikit-image. For Python 2.7, use

 $ pip install 'scikit-image<0.15'

Please also consider updating `pip` and `setuptools`:

 $ pip install pip setuptools --upgrade

Newer versions of these tools avoid installing packages incompatible
with your version of Python.

>>>
```

skimage/_shared/version_requirements.py
```diff
@@ -4,6 +4,30 @@
 import sys
 
 
+def check_python_version(min_version):
+    if not isinstance(min_version, tuple):
```
hmaarrfk: I'm impartial to this, but it helps catch the case where you don't specify the python version as a tuple, but just as an int

hmaarrfk:
Man, I didn't know that older pip didn't even check for version compatibility between python when pulling in sdist.

I had thought that this was caused by an adventurous user that just didn't listen to pip.

jni:
> this is what the output looks like

Nice! The only change that I would suggest is not to guess their version ("Likely Python 2"), and instead spell it out: "You are using Python x.x, but the latest scikit-image requires version 3.5 or higher." Python 3.4 users are unlikely but plausible.

hmaarrfk:
Yeah, I can do 

```
from platform import python_version
python_version()
```

should create a nicely formatted string.

The issue then is that the error message needs to branch:
  1. For 2.7/3/4, the last supported version is 0.14.x
  2. For other versions, please check our documentation on http://scikit-image.org/ to find the latest supported version.

jni:
@hmaarrfk oh no, no branching, just ignore <= 2.6. We dropped support for that years ago, and haven't had any complaints.

hmaarrfk:
New wording:

```
(py2) C:\Users\mark\git\scikit-image>python
Python 2.7.15 | packaged by conda-forge | (default, May  8 2018, 14:46:56) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import skimage
  File "skimage\__init__.py", line 71, in <module>
    check_python_version((3, 5))
  File "skimage\_shared\version_requirements.py", line 30, in check_python_version
    """ % (python_version(), '.'.join([str(v) for v in min_version])))
ImportError:

You are running scikit-image on an unsupported version of Python.
Likely Python 2.

Unfortunately, scikit-image 0.15 and above no longer work with your installed
version of Python (2.7.15).  You therefore have two options: either upgrade to
Python 3.5, or install an older version of scikit-image.

For Python 2.7 or Python 3.4, use

 $ pip install 'scikit-image<0.15'

Please also consider updating `pip` and `setuptools`:

 $ pip install pip setuptools --upgrade

Newer versions of these tools avoid installing packages incompatible
with your version of Python.
```

jni:
:+1: !

@scikit-image/core just need one more OK! ;)

skimage/_shared/version_requirements.py
```diff
@@ -1,9 +1,36 @@
 from distutils.version import LooseVersion
+from platform import python_version
 import functools
 import re
 import sys
 
 
+def check_python_version(min_version):
```
soupault: How about `ensure_python_version`?

skimage/_shared/version_requirements.py
```diff
@@ -1,9 +1,36 @@
...
+    if not isinstance(min_version, tuple):
+        min_version = (min_version, )
+    if sys.version_info < min_version:
+        raise ImportError("""
+
+You are running scikit-image on an unsupported version of Python.
+Likely Python 2.
```
soupault: `Likely Python 2.` can be safely removed now, I suppose.

PR #3164, opened 2018-06-08
Remove skimage.test as it was never used.

hmaarrfk:
Until  PR #3152 was merged, this function didn't actually work. Many of the build scripts just called `pytest` directly. The only one that I found that didn't call `pytest` directly was the`conda-forge` recipe,
which will be fixed when [thisPR](https://github.com/conda-forge/scikit-image-feedstock/pull/25)
is merged.

Previous comments from the omnibus PR #3147

@jni [mentionned](https://github.com/scikit-image/scikit-image/pull/3147#discussion_r193919183)
> I understand the sentiment, but I think we don't want to do this. The reason this exists is that we used to use nosetests and were bitten when it reached end-of-life. skimage.test allows us to keep the same testing syntax even if we decide to switch frameworks later on.

@jni [said](https://github.com/scikit-image/scikit-image/pull/3147#discussion_r193921324):
> also, "achieved". =P
> 
> I'm neither +1 or -1 on this btw. Running individual tests is more natural if we normalise with pytest. Just wanted to flag it since others (@soupault) have in the past preferred strongly hiding pytest (using skimage.testing.raises instead of pytest.raises, for example).

@hmaarrfk [replied]:
> @jni, I understand that the idea was to have the testing localized, but after having going through a few of the other build procedures, it seems that skimage.test was never used in the first in place.
>
> When I tried skimage.test myself a few months back, I thought I wasn't using the function correctly, but a recent PR showed me that it wasn't just me.
>
> I'm glad to go back to the old functionality, but I feel like the ship has sailed on that, even before my PR. I didn't have to change any of the other tests when I removed this function.

## Checklist
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -47,6 +47,8 @@ API Changes
   and ``skimage.transform.rescale`` has been set to ``reflect``.
 - Default value of ``anti_aliasing`` parameter in ``skimage.transform.resize``
   and ``skimage.transform.rescale`` has been set to ``True``.
+- Removed the ``skimage.test`` function. This functionality can be achieved
+  by calling ``pytest`` directly.
 
 
 Bugfixes
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -88,49 +88,6 @@ with your version of Python.
 """)
 
 
-try:
-    imp.find_module('pytest')
-except ImportError:
-    def _test(doctest=False, verbose=False):
-        """This would run all unit tests, but pytest couldn't be
-        imported so the test suite can not run.
-        """
-        raise ImportError("Could not load pytest. Unit tests not available.")
-
-else:
-    def _test(doctest=False, verbose=False):
-        """Run all unit tests."""
-        import pytest
-        import warnings
-        args = ['--pyargs', 'skimage']
-        if verbose:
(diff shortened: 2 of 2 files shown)


jni:
Thanks for the summary @hmaarrfk!

I believe @NelleV, @stefanv, and @soupault were the main architects of `skimage._shared.testing`, so I will let them comment. I myself am happy to go full pytest.

hmaarrfk:
As an extra remark:

I understand that `skimage._shared.testing` served a purpose to ensure that future testing backends wouldn't require 100+ changes in the future.

This testing function does not serve that purpose though.
It only wraps one function, and the code for wrapping 1 line of code is quite literally 43 lines long.

The documentation either reads:
> To test `scikit-image` please run:
```python
import skimage; skimge.test()
```

or

> To test scikit-image, please run
```python
import pytest; pytest.main(['--pyargs', 'skimage'])
```

For developers, this adds yet an other layer of "just for scikit-image" that they must learn in order to jump on board.

emmanuelle:
I'm also in favor of removing this code.

emmanuelle:
Merging, thanks.

hmaarrfk:

PR #3169, opened 2018-06-09
Backport: Travis should fail on failed tests

hmaarrfk:

```
set -ex
```

## For reviewers

  later.


hmaarrfk:
One change to an exception was never made in this branch.
The change to  INSTALL.rst was never made in this branch.

For some reason, the docstring test still works here. I have to investigate this. We should open an issue and figure out why the docstring is different between 0.14 and 0.15

PR #3170, opened 2018-06-09
Regression fix: Travis should fail if tests fail

hmaarrfk:
This is the PR to merge this PR #3169 into master.

```
set -ex
```

filter._unsharp mask is also given troubles. I can't pinpoint the source the bug just yet.

## For reviewers

  later.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -16,8 +16,10 @@ environment.
 -------------------------
 
 ``scikit-image`` comes pre-installed with several Python distributions,
-including Anaconda_, `Enthought Canopy`_, `Python(x,y)`_ and
-`WinPython`_.
+including `Anaconda <https://www.anaconda.com/download/>`_,
+`Enthought Canopy <https://www.enthought.com/product/canopy/>`_,
+`Python(x,y) <https://python-xy.github.io/>`_ and
+`WinPython <https://winpython.github.io/>`_.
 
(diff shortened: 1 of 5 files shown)


hmaarrfk:
@dvolgyes do you know why I had to change the docstrings to get the tests to pass `_unsharp_mask`?

INSTALL.rst
```diff
@@ -16,8 +16,10 @@ environment.
 -------------------------
 
 ``scikit-image`` comes pre-installed with several Python distributions,
-including Anaconda_, `Enthought Canopy`_, `Python(x,y)`_ and
-`WinPython`_.
+including `Anaconda <https://www.anaconda.com/download/>`_,
+`Enthought Canopy <https://www.enthought.com/product/canopy/>`_,
+`Python(x,y) <https://python-xy.github.io/>`_ and
+`WinPython <https://winpython.github.io/>`_.
```
jni: Sheesh, nice catch

jni:
@MeeseeksDev backport to v0.14.x

INSTALL.rst
stefanv: That was the intent with the original PR, fwiw.

hmaarrfk: Yeah, the tests would have caught it too, if they were actually enabled...... sorry for that

Issue #3071, opened 2018-05-12
rotating 90 degrees changes image size slightly


axu2:
## Description
When I rotate an image by 90 degrees, the size changes by one pixel with certain images, which is not the behavior I want.

Is there a way to do it in scikit? 

Otherwise, I'll just Transpose and reflect like this post says:

https://stackoverflow.com/questions/42519/how-do-you-rotate-a-two-dimensional-array

axu2:
```
from skimage import data
from skimage import transform

camera = data.camera()
camera = transform.resize(camera, (342, 470, 3))
print(camera.shape)

rot = transform.rotate(camera, 90, resize=True)
print(rot.shape)
```

gives me

```
(342, 470, 3)
(470, 343, 3)
```

I ended up transposing and then flipping the image using np.fliplr and np.transpose. Do you think this feature should be added to scikit-image?

soupault:
@axu2 thank you for the provided example. Looks like a bug.

>Do you think this feature should be added to scikit-image?

We'll dig into the current implementation first (as it is more generic).

axu2:
```
np.fliplr(np.transpose(img, (1,0,2)))  # 90 degree rotation
np.transpose(np.fliplr(rgby), (1,0,2))  # -90 degree rotation
```

I'll just post my workaround in the meantime.

ahojnnes:
See https://github.com/scikit-image/scikit-image/pull/3173

rfezzani:
Isn't this issue solved by #3173? At least, the `data.camera` example gives now the expected shape...

soupault:
Yes, thanks!

PR #3173, opened 2018-06-11
Round instead of ceil output shape when rotating image

ahojnnes:
Addresses issue https://github.com/scikit-image/scikit-image/issues/3071

Previously, we ensured that the output contained every subpixel information
from the input when rotating by ceiling the shape. This lead to inconsistent
output when rotating by multiples of 90 degrees. For example:

    >>> transform.rotate(np.zeros((470, 230)), angle=90, resize=True).shape
    (231, 470)

The new behavior correctly produces an output shape of (230, 470).

### DIFF
--- a/skimage/transform/_warps.py
+++ b/skimage/transform/_warps.py
@@ -371,7 +371,7 @@ def rotate(image, angle, resize=False, center=None, order=1, mode='constant',
         maxr = corners[:, 1].max()
         out_rows = maxr - minr + 1
         out_cols = maxc - minc + 1
-        output_shape = np.ceil((out_rows, out_cols))
+        output_shape = np.round((out_rows, out_cols))
 
         # fit output image in new shape
         translation = (minc, minr)
--- a/skimage/transform/_warps.py
+++ b/skimage/transform/_warps.py
@@ -371,7 +371,7 @@ def rotate(image, angle, resize=False, center=None, order=1, mode='constant',
         maxr = corners[:, 1].max()
         out_rows = maxr - minr + 1
         out_cols = maxc - minc + 1
-        output_shape = np.round((out_rows, out_cols))
+        output_shape = np.around((out_rows, out_cols))
 
         # fit output image in new shape
         translation = (minc, minr)
--- a/skimage/transform/tests/test_warps.py
+++ b/skimage/transform/tests/test_warps.py
@@ -160,6 +160,11 @@ def test_rotate_resize_center():
     assert_equal(x45, ref_x45)
 
 
+def test_rotate_resize_90():
+    x90 = rotate(np.zeros((470, 230), dtype=np.double), 90, resize=True)
+    assert x90.shape == (230, 470)
+
+
 def test_rescale():
     # same scale factor
     x = np.zeros((5, 5), dtype=np.double)


skimage/transform/_warps.py
```diff
@@ -371,7 +371,7 @@ def rotate(image, angle, resize=False, center=None, order=1, mode='constant',
         maxr = corners[:, 1].max()
         out_rows = maxr - minr + 1
         out_cols = maxc - minc + 1
-        output_shape = np.ceil((out_rows, out_cols))
+        output_shape = np.round((out_rows, out_cols))
```
soupault: Not sure about this. In Python 3 this rounds to the nearest even value, which means that the function might still produce a differently shaped output if some dimensions of the input image are odd. Should this line use `floor` instead?

P.S. To the best of my knowledge, `np.round` is deprecated, and `np.around` is recommended instead.

ahojnnes: The even/odd behavior is not an issue, since it only takes effect at the 0.5 boundary, otherwise the behavior is expected. np.round/np.around is the correct choice here.

emmanuelle:
Thanks @ahojnnes This indeed solves #3071, at the cost of losing a bit of subpixel information on edges. Are we comfortable wtth this? The idea of ``resize=True`` is to keep all the pixel information. 
I'm inclined to say that it's a very small loss, but I wonder what others think... (@jni?)

jni:
I'm :+1: on this change.

stefanv:
Round feels too magical to me. The fact that there is a decision boundary 
on 0.5 makes the decision of whether to go up or down arbitrary (even 
though we round to even, consistently, what is the defense for that case?). 
Rounding up is safer, and users always have the option of specifying a size 
manually. The 90 degree rotation is annoying. Is it because of floating 
point imprecision that we go one pixel over?

JDWarner:
Rounding to evens is IEEE standards correct.  It eliminates bias which would otherwise be present during the rounding operation.

stefanv:
I understand that it is "correct" for reducing bias. But it is not 
necessarily correct for information preservation, or for managing expectations!

jni:
@stefanv my expectation would be to use round here. If you are only doing a single transformation, the subpixel accuracy is unlikely to matter. If you are chaining many transformations through their outputs (instead of through the transformation chaining API), then you will have much bigger problems than a missing edge pixel.

The 90° rotation is a clear use case where ceil is broken. I think it should be a priority to fix it. Got any better ideas than this?

ahojnnes:
My rationale was that even if we use ceil, the output image will be cut off “abpruptly” for any interpolation order but nearest neighbor interpolation. So, ceil is not any better than round in these cases, and then round is just better because it solves the rot90 problem.

stefanv:
So imagine two images, each scaled by 0.5x.  The shapes are ``(10, 10)`` and ``(7, 7)``.  What should the output sizes ideally be?  This PR says ``(5, 5)`` and ``(3, 3)``, if I understand correctly.

This means that at least one data-point now falls outside of the new image.

jni:
@stefanv nearest even argues the new size would be (4, 4)? But, I get your point, (9, 9) would go to (4, 4). I don't think this is a huge problem as the image should be pre-filtered for downsampling. And in the converse case of (5, 5), now you are getting data that is mostly synthesised outside of the input image. Is that worse?

ahojnnes:
We are talking about rotating an image and not rescaling? If at all, the output image will be bigger than the input image if `resize=True`.

jni:
@ahojnnes I think @stefanv wanted to use the example of rescaling as a case study of what *should* happen. But actually for rescaling it seems we are already rounding:

```python
>>> import numpy as np
>>> from skimage import transform
>>> image = np.random.random((9, 9))
>>> transform.rescale(image, 0.5, multichannel=False).shape
(4, 4)
```

@stefanv does this existing behaviour change your mind about what should happen with rotate?

ahojnnes:
@jni In the `rescale` case, the function just rounds the output size, but the actual rescaling factor used during interpolation is then determined based on the rounded coordinates, so no information is lost in this case for sure.

jni:
> the actual rescaling factor used during interpolation is then determined based on the rounded coordinates

Whoa! I'm not sure I like that, but ok, thanks for the info! =)

axu2:
Unsure of what exactly you want, but maybe if a user wants to rotate an image by a multiple of 90 degrees, they shouldn't be using transform.rotate and instead use np.rot90?

https://docs.scipy.org/doc/numpy/reference/generated/numpy.rot90.html

Maybe transform.rot can call np.rotate when its a multiple of 90 deg?

stefanv:
Eh, yes, not sure I like the magic of rescale either.

If we want to commit to magical fixing of coordinates, why not shrink the output coordinates by 5 * eps (or some chosen delta) and take the ceil? Still a consistent recipe, fixes the rot90 case, but does not cause sizes like 4.5 to go down to 4.

ahojnnes:
@stefanv I don't think you last proposal a good fix for the problem. This "magical threshold" that you mention will depend on the image size and might not work in many cases. How about we do a check for ``angle mod 90 == 0`` and enforce the dimensions to be correct for this specific case?

alexandrejaguar:
Hey everyone,
any news on this one? :)

jni:
@stefanv is the blocker here, you should go bug him in person @alexandrejaguar =P

Reading the whole discussion again, I still think rounding is the correct and least-surprising choice here.

stefanv:
It's consistent, it fixes the issue; let's put it in.

Issue #3001, opened 2018-04-01
Crash on resize() with 1 pixel image and mode="reflect"


waleedka:
This code reproduces the issue:

```
image = np.zeros([1, 2])
image = skimage.transform.resize(image, (2, 4), order=1, mode="reflect")
```

This causes the Python interpreter to crash with a "floating point exception". I'm running this on Ubuntu 16.04, Python 3.5.2 and scikit-image 0.13.1.

Changing mode to `constant` or changing order to `0` avoids the problem.

jni:
Well, this crash makes perfect sense but I don't know how to fix it. =P

@waleedka thanks for the report! Would you expect it to work or just to throw a meaningful error? I don't think reflect makes sense when the image is one-pixel-thick in one dimension.

Coincidentally, this is probably an argument in favour of making `mode='edge'` the default as proposed in #2990!

waleedka:
> Would you expect it to work or just to throw a meaningful error? I don't think reflect makes sense when the image is one-pixel-thick in one dimension.

I'm no expert in the intricate details and edge cases of image resizing, but I would expect it to work rather than throw an error. And I don't see a problem with applying reflect on a 1-pixel image, it should simply repeat that pixel, similar to how Numpy handles it.
```
np.pad([[7]], 1, 'reflect')

# Out:
# array([[7, 7, 7],
#        [7, 7, 7],
#        [7, 7, 7]])
```

jni:
Wow! That's very interesting, because that really doesn't jive with the definition of 'reflect'. From the `pad` docstring:

```
    >>> a = [1, 2, 3, 4, 5]
    >>> np.lib.pad(a, (2, 3), 'reflect')
    array([3, 2, 1, 2, 3, 4, 5, 4, 3, 2])
```

Note how the edge value (1) is not repeated: you are taking a reflection about that value to get the padding. In the length-1 case, you are trying to reflect from beyond the edge of the image to the other side, so you have no values to pull from...

waleedka:
I see your point. You're right, it's an edge case. But I think repeating the 1 pixel is okay because the spirit of the `reflect` mode is to reduce discontinuities at the edges, and this achieves it.

stefanv:
We should never segfault.

stefanv:
```
[...]

PR #3174, opened 2018-06-11
Repeat pixels in reflect mode when image has dimension 1

ahojnnes:
This replicates the `np.pad` behavior and addresses issue https://github.com/scikit-image/scikit-image/issues/3001

### DIFF
--- a/skimage/_shared/interpolation.pxd
+++ b/skimage/_shared/interpolation.pxd
@@ -330,8 +330,7 @@ cdef inline Py_ssize_t coord_map(Py_ssize_t dim, long coord, char mode) nogil:
         Whether to wrap, symmetric reflect, reflect or use the nearest
         coordinate if `coord` falls outside [0, dim).
     """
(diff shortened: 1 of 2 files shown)


jni:
@ahojnnes for the reasons stated in #3001, I would prefer if we raised an error in this case. It is more consistent with the definition of `reflect`. What do you think?

JDWarner:
@jni I think it's a hard sell to conflict with the operation of NumPy.  This is the fix that makes the most sense, as far as what the user would be intending with the operation.

Reflect vs. edge are both trying to use the image data.  When there's only one index to choose from, their behavior converging doesn't seem like an error to me - and I can't think of any cases treating it this way would break.

jni:
@JDWarner you make some very compelling points. NumPy compatibility in particular is a very good aim. Merging!

jni:
@MeeseeksDev backport to v0.14.x

jni:
Wow, magic. =D

PR #3187, opened 2018-06-13
Speedup rgb2gray using matrix multiply

hmaarrfk:
1000 x 1000 x 3 array:
Original: 15ms
As proposed: 8-9ms

Run cell by cell
```python
import numpy as np

a = np.random.rand(1000, 1000, 3)

# %%
%%timeit

def rgb2gray(rgb):
    coeffs = np.array([0.2125, 0.7154, 0.0721], dtype=rgb.dtype)
    return rgb @ coeffs

d = rgb2gray(a)

# %%
%%timeit
def rgb2gray(rgb):
    gray = 0.2125 * rgb[..., 0]
    gray += 0.7154 * rgb[..., 1]
    gray += 0.0721 * rgb[..., 2]
    return gray

d = rgb2gray(a)
```

## References

## For reviewers

  later.

### DIFF
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -801,12 +801,8 @@ def rgb2gray(rgb):
         return np.ascontiguousarray(rgb)
 
     rgb = _prepare_colorarray(rgb[..., :3])
-
-    gray = 0.2125 * rgb[..., 0]
-    gray[:] += 0.7154 * rgb[..., 1]
-    gray[:] += 0.0721 * rgb[..., 2]
-
-    return gray
+    coeffs = np.array([0.2125, 0.7154, 0.0721], dtype=rgb.dtype)
+    return rgb @ coeffs
 
 
 rgb2grey = rgb2gray


skimage/color/colorconv.py
```diff
@@ -801,12 +801,8 @@ def rgb2gray(rgb):
         return np.ascontiguousarray(rgb)
 
     rgb = _prepare_colorarray(rgb[..., :3])
-
-    gray = 0.2125 * rgb[..., 0]
-    gray[:] += 0.7154 * rgb[..., 1]
-    gray[:] += 0.0721 * rgb[..., 2]
-
-    return gray
+    coeffs = np.array([0.2125, 0.7154, 0.0721], dtype=rgb.dtype)
```
jni: Why `rgb.dtype`? Is `rgb` guaranteed to be a float type?

hmaarrfk: yeah, the prepare function makes it a float of the correct range.

I feel like removing that would be considered "premature optimization"

PR #3199, opened 2018-06-14
WIP: Feature nd colorconv

### LINKED ISSUE #3185: ND multichannel arrays in color conversion
https://github.com/scikit-image/scikit-image/blob/ac4a8f1f31a28d34b81871ad3a6864b917474263/skimage/color/colorconv.py#L152

The line above only allows you to have 4D multichannel arrays.

Would you accept patches that ensures that we can work with ND multichannel arrays?

This is useful for example when you are trying to stitch tiles of acquired images with a microscope.

You might have large arrays, each RGB (3 dimensions) tiled in the x, y, and time dimension. 
6 dimensions! I'm sure somebody could think of more.

I know there is talk about using an other backend for color conversion, but the skimage color backend seems to do the trick for now.

hmaarrfk:
Allows one to perform color conversion with ND arrays. This makes it easier to apply color conversions to sets of arrays.

WIP. Fixes #3185

TODO: 
- [x] change 3D assumed slicing to use `...` allowing for arbitrary dimensions to be used. `[:, :, i] -> [..., i]`
- [ ] allow ND dimension arrays to pass through where only 3D/4D dimension arrays were used.
- [ ] decide if we want `gray2rgb(rgbimage)` to act as a no-op if a gray image is already passed in.
- [ ] write a test for some kind of ND array.
- [ ] backport ? 

## Checklist
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/adapt_rgb.py
+++ b/skimage/color/adapt_rgb.py
@@ -55,10 +55,10 @@ def hsv_value(image_filter, image, *args, **kwargs):
         Input image. Note that RGBA images are treated as RGB.
     """
     # Slice the first three channels so that we remove any alpha channels.
-    hsv = color.rgb2hsv(image[:, :, :3])
-    value = hsv[:, :, 2].copy()
+    hsv = color.rgb2hsv(image[..., :3])
+    value = hsv[..., 2].copy()
     value = image_filter(value, *args, **kwargs)
-    hsv[:, :, 2] = convert(value, hsv.dtype)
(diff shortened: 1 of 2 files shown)


hmaarrfk:
@jni, sorry for leaving this one dormant for so long. Basically I couldn't find a good way to gracefully deprecate the automagic features of colorconv. I fear that maybe we would have to resort to creating a new module `ndcolor`. Shouldn't be too hard, but there would definitely be some duplication.

PR #3202, opened 2018-06-14
Big image labeling fix

blochl:
## Description
Previously, labeling a large dataset would silently fail (falsely mis-label all the data after a certain point). With this fix, that is avoided by using long integers for the calculations.

In the image below (a scaled-down slice of a 3D dataset) this issue is illustrated: after a certain point, all the regions were being given the label "1".

Also, some comments are fixed.

label_edge

## For reviewers
  later.

### DIFF
--- a/skimage/measure/_ccomp.pyx
+++ b/skimage/measure/_ccomp.pyx
@@ -68,7 +68,7 @@ cdef enum:
     D_COUNT
 
 
-# Structure for centralised access to shape data
+# Structure for centralized access to shape data
 # Contains information related to the shape of the input array
 cdef struct s_shpinfo:
     INTS_t x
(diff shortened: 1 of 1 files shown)


soupault:
What is the memory use increase on this image? :)

blochl:
> What is the memory use increase on this image? :)

Not noticeable. The arrays themselves were using int64 before. This just makes some auxiliary data int64 as well.

blochl:
> Approved, pending a test case.

A test would be problematic, as the issue occurs only with very large datasets.

stefanv:
Would even a checkerboard pattern be prohibitive?

blochl:
> Would even a checkerboard pattern be prohibitive?

I guess that a 3D checkerboard of ~1300x1300x1300 would have this issue. But that's not feasible for testing.

soupault:
I think, that is about as much RAM as Travis provides :).

stefanv:
OK, let's move on :)

blochl:
Thanks! :)

PR #3204, opened 2018-06-14
On Travis-CI, test with Python 3.6 against the latest NumPy nightly

stefanv:
From a tip by @matthew-brett on the NumPy list

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
       env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 BUILD_DOCS=1 DEPLOY_DOCS=1
     - os: linux
       python: 3.6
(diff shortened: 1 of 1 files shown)


jni:
> despite that this will occasionally cause lots of failing builds

:joy: 

https://api.travis-ci.org/v3/job/392361657/log.txt

Question: are boolean arrays deprecated??? Or is the nightly build broken??? Also, what's with all the `[1m ... [0m`?

soupault:
@jni it is a syntax for colors in Bash`ian. :)

stefanv:
Ah, good, we already caught something :)

jni:
@stefanv do we still want this in?

hmaarrfk:
xref: https://github.com/scikit-image/scikit-image/pull/3729

I think this should just be closed. this infrastructure stuff goes stale quickly.

emmanuelle:
Let's close this one given the above discussion and what was done in #3729. Stéfan if you have second thoughts you can of course re-open it.

Issue #3012, opened 2018-04-11
Feature request: histogram matching


stefanv:
See, e.g. https://stackoverflow.com/a/33047048/214686

solutusimmensus:
I'd like to work on this feature, but am not sure as to what the unit tests should look like. 
Can I just import some images from skimage.data (both rgb and gray-scale) and check whether the histograms of source and output are sufficiently close?

jni:
@solutusimmensus yes that's what I would try I think!

PR #3208, opened 2018-06-16
New feature: histogram match (Closes: Feature request: histogram matching #3012)

solutusimmensus:
## Description
Implement histogram matching for images with arbitrary number of channels (each matching is done channel-wise). 
Add unit tests for histogram_matching.

## Checklist
  existing benchmark
- [ ] Unit tests

## References
http://paulbourke.net/miscellaneous/equalisation/
^ Not really a paper, but I had a hard time finding a proper research document
## For reviewers

  later.


solutusimmensus:
Guess I have to add an example in `doc/examples/transform`, hence PR status changed to WIP.
Should I add some benchmarks? `./benchmarks` seems to be quite outdated.

solutusimmensus:
Example added in `./doc`. 
I also had to raise unit tests tolerance due to Travis build failing for one version of Python.

solutusimmensus:
One Travis build fails, but it seems to be caused by the internet connectivity problem.

skimage/transform/tests/test_histogram_matching.py
```diff
@@ -0,0 +1,104 @@
...
+
+image_rgb = data.chelsea()
+template_rgb = data.astronaut()
+
+
+@pytest.mark.parametrize('image, reference', [
+    (image_rgb[:, :, 0], template_rgb[:, :, 0]),
+    (image_rgb, template_rgb)
+])
+def test_match_histograms(image, reference):
```
lagru: A suggestion: You might want to combine unit tests for the same function in a class. In this case you could save the global variables `image_rgb` and `template_rgb` as class attributes which may be a more contained solution.

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,80 @@
...
+    """Return an array consisting of all channels of an image"""
+    if len(image.shape) < 3:
+        return np.asarray([image])
+    else:
+        return image.transpose(2, 0, 1)
+
+
+def _match_array_values(a, b):
+    """Return modified a array with interpolated values that correspond to the
+    values of the b array"""
```
lagru: Using `source` instead of `a` and `template` instead of `b` (like in the original example from stackoverflow) might make the code clearer and easier to read.

E.g. `a_quantiles` -> `src_quantiles`, `b_quantities` -> `tmp_quantities`

skimage/transform/tests/test_histogram_matching.py
solutusimmensus: Good point, especially with the helper function. 
Moved functional tests to separate function `TestMetchHistogram`.

skimage/transform/histogram_matching.py
solutusimmensus: Right, renamed the variables accordingly.

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Raises
+    ------
+    ValueError
+        Thrown when the number of channels in the input image and the reference
+        differ
+
+    References
+    ----------
+    .. [1] http://paulbourke.net/miscellaneous/equalisation/
```
lagru: You may want to add an Example section if you can think of a simple example to demonstrate the usage.

solutusimmensus: There is an example in `doc/examples/transform/plot_histogram_matching.py` with plots. 
I don't think it would be very informative as a console based example.

lagru: Sounds reasonable.

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
+
+
+def _get_separate_channels(image):
+    """Return an array consisting of all channels of an image"""
+    if len(image.shape) < 3:
```
stefanv: image.ndim

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,80 @@
...
+        return np.asarray([image])
```
stefanv: Perhaps `np.array(image, copy=False, ndmin=3)` ?

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,80 @@
...
+    else:
+        return image.transpose(2, 0, 1)
+
+
+def _match_array_values(source, template):
+    """Return modified source array with interpolated values that correspond to the
```
stefanv: Can this documentation be clarified?

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,80 @@
...
+
+    Parameters
+    ----------
+    image : ndarray
+        Input image. Can be gray-scale or in color
+    reference : ndarray
+        Image to match histogram of. Must have the same number of channels as
+        image, otherwise an exception is thrown
+
+     Returns
```
stefanv: typo

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,80 @@
...
+
+    tmpl_quantiles = np.cumsum(tmpl_counts).astype(np.float64)
+    tmpl_quantiles /= tmpl_quantiles[-1]
+
+    interp_a_values = np.interp(src_quantiles, tmpl_quantiles, tmpl_values)
+    return interp_a_values[src_unique_indices]
+
+
+def match_histograms(image, reference):
+    """Adjust the pixel values of an image so that its histogram and a target's
```
stefanv: Single sentence description, followed by a paragraph with more detail.  It is not necessary to mention explicitly that the image will be preserved (this should be a given), especially since the function has a return value.  Perhaps:

```
Adjust an image so that its histogram matches that of another.
```

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
+
+
+def _get_separate_channels(image):
```
stefanv: Having this as a separate function may be overkill, if you can run something like `array(..., ndmin=3)` on any input, and then proceed from there.

skimage/transform/histogram_matching.py
solutusimmensus: Renamed the function to `_match_cumulative_cdf`, reformulated the docstring.

skimage/transform/histogram_matching.py
solutusimmensus: Missing space added.

skimage/transform/histogram_matching.py
solutusimmensus: Rephrased the docstring, mentioned that the adjustment is applied channel-wise.

skimage/transform/histogram_matching.py
solutusimmensus: Got rid of the `_get_separate_channels`. 
Had to change the order of transposition, since `ndim` prepends the shape with ones (so transposition is no longer needed for these arrays).

solutusimmensus:
Soo... Is there anything else I should change in the code?

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,78 @@
...
+
+    The adjustment is applied separately for each channel.
+
+    Parameters
+    ----------
+    image : ndarray
+        Input image. Can be gray-scale or in color
+    reference : ndarray
+        Image to match histogram of. Must have the same number of channels as
+        image, otherwise an exception is thrown
```
sciunto: please, put a full stop "." at the end of all your sentences in doctrings,

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,78 @@
...
+                         'image must match!')
+
+    matched_channels = np.empty(image.shape)
+
+    for channel in range(len(image)):
+        im_channel = image[channel].ravel()
+        ref_channel = reference[channel].ravel()
+
+        matched_channel = _match_cumulative_cdf(im_channel, ref_channel)
+        matched_channel = matched_channel.reshape(shape[0:2])
```
sciunto: With the above recommendation, you can reshape in _match_cumulative_cdf()

skimage/transform/histogram_matching.py
```diff
@@ -0,0 +1,78 @@
...
+        matched_channels[channel] = matched_channel
+
+    matched = matched_channels.transpose([1, 2, 0]).reshape(shape)
```
sciunto: Could you explain the purpose of this line please? (add a comment)

doc/examples/transform/plot_histogram_matching.py
```diff
@@ -0,0 +1,46 @@
...
+lighting).
+"""
+
+import matplotlib.pyplot as plt
+
+from skimage import data
+from skimage.transform import match_histograms
+
+reference = data.astronaut()
+image = data.chelsea()
```
sciunto: I think we can improve the example by adding more suitable images, but I would suggest to do that in a separate PR.

sciunto:
@scikit-image/core The user closed his account. How do you advise to continue this work? Copy paste by hand? Attribute to the user name in the commit message?

jni:
@sciunto we don't have an explicit process for this. Your suggested course of action sounds reasonable to me.

stefanv:
But the account @SolutusImmensus seems to exist still.

PR #3213, opened 2018-06-19
Removed dtype constructor call in exposure.rescale_intensity

### LINKED ISSUE #3044: Rescale intensity returns only a float given an array
## Description

I start with a numpy `array` that has `array.shape == (1, 1)` and `array.dtype == np.float64`. Let's define an example `array = np.array([[0]], dtype=np.float64)`.

If I then pass this float64 array as the first argument of `exposure.rescale_intensity`, the function gives `np.float64(0.0)` as an output instead of the expected `np.array([[0.0]], dtype=np.float64)`.

This is caused by [line 298 of exposure.py](https://github.com/scikit-image/scikit-image/blob/b02e2ce7014f245ff20d692519f2c36ed4bf8269/skimage/exposure/exposure.py#L298)

We should replace this line's `dtype(image * (omax - omin) + omin)` with this line:

```python
np.array(image * (omax - omin) + omin, dtype=dtype)
```

When called as a function, `np.float64` has the unexpected behavior that this is true: 

```python
np.float64(np.array([[0]])) == np.float64(0.0)
```

Therefore, contrary to expected behavior, this is false:

```python
type(np.float64(np.array([[0]]))) == np.ndarray
```

I've raised an [issue at numpy/numpy](https://github.com/numpy/numpy/issues/10995), but it appears this behavior will not be corrected given that the `np.float64(array)` syntax is undocumented. I believe all code that depends on numpy should move away from using `np.float64(*)` as an array constructor. It is always safer to use `np.array(*, dtype=np.float64)`.

## Way to reproduce

```python
import numpy as np
import skimage.exposure

array = np.array([[0]], dtype=np.float64)
output = skimage.exposure.rescale_intensity(array, (0, 1))
assert type(output) == np.ndarray #assertion fails
```
Iff `dtype == np.float64`, instead of `[[0.0]]`, rescale_intensity returns`0.0`

- ubuntu 16.04 LTS
- python 3.6.4
- scikit-image 0.13.1

fivemok:
## Description
Removed dtype constructor call in exposure.rescale_intensity. Calling np.array constructor with dtype keyword instead.

## Checklist
  existing benchmark
- [ ] Unit tests

## References
Closes #3044 

## For reviewers
  later.

### DIFF
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -294,7 +294,7 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     image = np.clip(image, imin, imax)
 
     image = (image - imin) / float(imax - imin)
-    return dtype(image * (omax - omin) + omin)
+    return np.array(image * (omax - omin) + omin, dtype=dtype)
 
 
 def _assert_non_negative(image):


jni:
Thanks @fivemok!

jni:
Come to think of it we should probably add a test to check this, so we don't get a regression in the future... ;)

PR #3215, opened 2018-06-20
Various updates to the ASV benchmarks

TomAugspurger:
## Description

A few fixups for the ASV benchmarks:

1. ASV expects a `__init__.py` in the benchmarks dir
2. For `mem_` tests, the object has to be returned
3. Added the generated `.asv` directory to the `.gitignore` 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ doc/release/_release_notes_for_docs.txt
 doc/gh-pages
 wheels
 skimage/morphology/_skeletonize_3d_cy.pyx
+.asv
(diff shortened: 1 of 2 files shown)


stefanv:
Thanks!

jni:
@TomAugspurger ah, but I don't think we care very much about the result. So there is no way to measure the *internal* memory used by a function?

stefanv:
Right, good point.

TomAugspurger:
I wonder, is
https://github.com/airspeed-velocity/asv/blob/master/docs/source/writing_benchmarks.rst#peak-memory
what we want?

jni:
@TomAugspurger yes that is exactly the right thing! =) Mesohappy.

PR #3216, opened 2018-06-21
draw_random_shapes: add background argument and avoid invisible objects

### LINKED ISSUE #3083: Add `background` keyword argument to `draw.random_shapes`
## Description
In #3029, @stefanv noted that if we allow an intensity range for the random shapes, we could/should also allow specifying the background, rather than hardcoding it as white. 

Additionally, the docstring needs fixing, as intensity range values can be of any int type, not just uint8. Their values need to be restricted to the uint8 range, but that's not the same thing.

cmohl2013:
## Description
The new optional background argument defines a custom background color. Before, background was hard coded to 255 for each channel.

The PR also includes a fix for randomly picking colors of foreground objects: Foreground objects are not allowed to have background color any more (i.e. making them invisible).

## Checklist
  existing benchmark
- [x] Unit tests

## References
Closes #3083

## For reviewers

  later.


skimage/draw/_random_shapes.py
```diff
@@ -200,13 +205,34 @@ def _generate_random_colors(num_colors, num_channels, intensity_range, random):
...
     if num_channels == 1:
-        intensity_range = (intensity_range, )
+        intensity_range_tpl = (intensity_range, )
     elif len(intensity_range) == 1:
-        intensity_range = intensity_range * num_channels
+        intensity_range_tpl = intensity_range * num_channels
+    else:
+        intensity_range_tpl = intensity_range
+
+    if np.diff(np.array(intensity_range_tpl)).sum() == 0:
```
stefanv: Is this check overkill?  Even if not, I think the error message should at least be improved.  But what do we lose if we remove it entirely?

skimage/draw/_random_shapes.py
```diff
@@ -200,13 +205,34 @@ def _generate_random_colors(num_colors, num_channels, intensity_range, random):
...
+            raise ValueError(msg)
+        return np.repeat(colors, num_colors, axis=0)
+
     colors = [random.randint(r[0], r[1]+1, size=num_colors)
-              for r in intensity_range]
-    return np.transpose(colors)
+              for r in intensity_range_tpl]
+    colors = np.transpose(colors)
+
+    to_replace = np.array([(exclude == color).all() for color in colors])
```
stefanv: First convert to a numpy array and then perform array operations?

skimage/draw/_random_shapes.py
```diff
@@ -312,16 +344,28 @@ def random_shapes(image_shape,
                     msg = 'Intensity range must lie within (0, 255) interval'
                     raise ValueError(msg)
 
+    if background is None:
+        background = [255] * num_channels
+    elif isinstance(background, int):
+        background = [background]
```
stefanv: [background] * num_channels?

stefanv: Also, what about the case where the user wants to specify RGB background.

stefanv (commented):
Thanks for your contribution, Christoph.  I've left some initial feedback.

skimage/draw/_random_shapes.py
cmohl2013: If intensity range spans only a single color, and this color equals the color to be excluded, the recursive call in line 230 will end up in an infinite loop. Therefore this check is important. The error message could be more specific, indeed.

skimage/draw/_random_shapes.py
cmohl2013: Yes, would be more efficient.

skimage/draw/_random_shapes.py
cmohl2013: ```[background] * num_channels```: good idea. My initial idea was to not allow single values for background in multichannel images. But this is more convenient.

I'm not shure if I understand you correctly regarding RGB background. The user can specify background for images of three channels with a tuple of length 3. Or do you think of matplotlib-like color strings as e.g.```background='r'``` for red background?

skimage/draw/_random_shapes.py
cmohl2013: something like this: ```to_replace = np.equal(color, exclude).all(axis=1)```

stefanv:
No, I was thinking of the 3-tuple case.  I missed seeing that that is
already supported.

skimage/draw/_random_shapes.py
```diff
@@ -312,16 +344,28 @@ def random_shapes(image_shape,
...
+    if len(background) != num_channels:
+        raise ValueError('Nr of background values must match nr of channels')
+    for intensity in background:
```
soupault: It seems that operations on `background` take too many lines. Let's just cast it to ndarray from the beginning, and use the vectorized ops.

skimage/draw/_random_shapes.py
```diff
@@ -257,10 +284,15 @@ def random_shapes(image_shape,
...
         The range of values to sample pixel values from. For grayscale images
         the format is (min, max). For multichannel - ((min, max),) if the
-        ranges are equal across the channels, and ((min_0, max_0), ... (min_N, max_N))
-        if they differ. As the function supports generation of uint8 arrays only,
-        the maximum range is (0, 255). If None, set to (0, 254) for each
-        channel reserving color of intensity = 255 for background.
+        ranges are equal across the channels, and
+        ((min_0, max_0), ... (min_N, max_N)) if they differ. As the function
+        supports generation of uint8 arrays only, the maximum range is
+        (0, 255). If None, set to (0, 255) for each channel reserving color of
```
soupault: `reserving color of intensity = 255 for background.` <- this should probably be moreved. The same info is provided 2 lines below.

skimage/draw/_random_shapes.py
```diff
@@ -200,13 +205,34 @@ def _generate_random_colors(num_colors, num_channels, intensity_range, random):
...
     """
+
     if num_channels == 1:
-        intensity_range = (intensity_range, )
+        intensity_range_tpl = (intensity_range, )
     elif len(intensity_range) == 1:
-        intensity_range = intensity_range * num_channels
+        intensity_range_tpl = intensity_range * num_channels
+    else:
+        intensity_range_tpl = intensity_range
```
soupault: Is there really a need in changing this block?

skimage/draw/_random_shapes.py
soupault: `.sum()` is not a 100% valid check here. Will fail on, for example, `((0, 1), (1, 0), (0, 0))`.

skimage/draw/_random_shapes.py
```diff
@@ -200,13 +205,34 @@ def _generate_random_colors(num_colors, num_channels, intensity_range, random):
...
+
+    if np.diff(np.array(intensity_range_tpl)).sum() == 0:
+        # if intensity range only spans a single color
+        colors = np.array([np.array(intensity_range_tpl)[:, 0]])
```
soupault: To many `np.array` constructors here, above, and below. Please, cast `intensity_range_tpl` to ndarray as early as possible, and use the vectorized ops.

skimage/draw/_random_shapes.py
```diff
@@ -312,16 +344,28 @@ def random_shapes(image_shape,
                     msg = 'Intensity range must lie within (0, 255) interval'
                     raise ValueError(msg)
 
+    if background is None:
+        background = [255] * num_channels
+    elif isinstance(background, int):
+        background = [background]
+    if len(background) != num_channels:
+        raise ValueError('Nr of background values must match nr of channels')
```
soupault: `Nr/nr` abbreviation should be expanded.

skimage/draw/tests/test_random_shapes.py
```diff
@@ -135,9 +136,137 @@ def test_random_shapes_is_reproducible_with_seed():
     assert all(other == labels[0] for other in labels[1:])
 
 
-def test_generates_white_image_when_intensity_range_255():
-    image, labels = random_shapes((128, 128), max_shapes=3,
-                                  intensity_range=((255, 255),),
+def test_throws_when_intensity_range_255():
```
soupault: `test_throws_when_intensity_range_255` -> `test_throws_when_intensity_range_equals_background` or something. Also, please specify the background value explicitly in the function call below.

skimage/draw/tests/test_random_shapes.py
```diff
@@ -135,9 +136,137 @@ def test_random_shapes_is_reproducible_with_seed():
...
+    with testing.raises(ValueError):
+        random_shapes((128, 128), max_shapes=3,
+                      intensity_range=((255, 255),),
+                      random_seed=42)
+
+
+def test_pick_random_colors_within_range():
```
soupault: This could be rewritten using `parametrize` decorator.

skimage/draw/tests/test_random_shapes.py
```diff
@@ -135,9 +136,137 @@ def test_random_shapes_is_reproducible_with_seed():
...
+    intensity_range = ((20, 21), (25, 26))
+    colors = _generate_random_colors(num_colors, num_channels,
+                                     intensity_range, random,
+                                     exclude=(21, 25))
+    assert len(colors) == 10
+    for color in colors:
+        assert tuple(color) in [(20, 25), (20, 26), (21, 26)]
+
+
+def test_throws_when_intensity_range_equals_excluded_intensities():
```
soupault: Erm, this essentially repeats `test_throws_when_intensity_range_255`...

skimage/draw/tests/test_random_shapes.py
```diff
@@ -135,9 +136,137 @@ def test_random_shapes_is_reproducible_with_seed():
...
+                                  num_channels=2,
+                                  intensity_range=intensity_range,
+                                  background=(21, 31),
+                                  random_seed=42)
+    print(image.shape)
+    assert set(image[:, :, 0].flatten()) == {20, 21}
+    assert set(image[:, :, 1].flatten()) == {30, 31}
+
+
+def test_throws_when_backgound_out_of_range():
```
soupault: This function and the one below could be also refactored using `parametrize`.

soupault (commented):
@cmohl2013 very good starting point, thanks! Let's polish the PR a bit. :)

skimage/draw/_random_shapes.py
cmohl2013: I wanted to keep the original intensity_range value for using in the recursive call in line 230. But I see that this approach is confusing, makes the code less readable. Indeed this can be done outside the block in a more simple way.

cmohl2013:
@stefanv @soupault Thank you for your descent review! I think all the improvements you suggested are clear to me. I'll try to update the PR in the next couple of days.

skimage/draw/_random_shapes.py
hmaarrfk: Sorry to interject, but `asarray` might be appropriate to avoid a copy early on.

Also, are you looking for `np.newaxis`. you would slice with `[:, 0, np.newaxis]`

cmohl2013: @hmaarrfk thanks for the hint!

cmohl2013:
I updated the PR according the comments. Should the example in the gallery also be updated with a background argument example?

soupault (changes requested):

soupault:
@cmohl2013 Thank you for the changes! The current implementation is much better, although, I still see some issues (see the comments above).

skimage/draw/_random_shapes.py
```diff
@@ -257,10 +284,15 @@ def random_shapes(image_shape,
...
-        channel reserving color of intensity = 255 for background.
+        ranges are equal across the channels, and
+        ((min_0, max_0), ... (min_N, max_N)) if they differ. As the function
+        supports generation of uint8 arrays only, the maximum range is
+        (0, 255). If None, set to (0, 255) for each channel. Color of
+        background is always excluded.
+    background : {tuple of ints, int}, optional
+        Pixel intensities for background. Values between 0 and 255 are allowed.
+        For multichannel, a tuple of length num_channels is required. If None,
+        set to 255 for each channel.
```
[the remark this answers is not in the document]
cmohl2013: You're right, better we do not support ```None```. Solves also some of your comments below.

cmohl2013:
@soupault Thank you again for the review. I'll try to change implementation according your comments.

cmohl2013:
@soupault 
I implemented changes according to your requests.

soupault (changes requested):
@cmohl2013 I think, I've found a bug in the current implementation. Once fixed, this PR should be good to go.

skimage/draw/_random_shapes.py
```diff
@@ -200,13 +205,30 @@ def _generate_random_colors(num_colors, num_channels, intensity_range, random):
...
+        colors = np.broadcast_to(intensity_range[:, 0],
+                                 (num_colors, num_channels))
+        if exclude is not None and (colors[0, :] == exclude).all():
+            msg = ('Shape colors can not be selected. '
+                   'Intensity range spans only background value.')
+            raise ValueError(msg)
+        else:
+            return colors
+    color_candidates = [np.arange(r[0], r[1] + 1) for r in intensity_range]
+    color_candidates = [np.setdiff1d(c, e) for c, e
```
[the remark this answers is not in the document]
cmohl2013: wow, a bug indeed, not so easy to find I guess. Yes I'll add a test and see how to fix this. Thanks!

matthew-brett:

<details>

## Rationale

</details>

Issue #3172, opened 2018-06-09
Modernise build process


hmaarrfk:
The build process used to have requirements scattered across many scripts in travis. 
This was probably due to the fact that PyPi didn't have wheels and thus some optimization was done around that.
There seems to have been an attempt at cleaning this up a while back, but the build process is still needlessly complicated.

Many of these are small changes to files, but combined, it can be confusing to understand what happens. I'll try to split them up as different PRs.

 * [x] 0. Ensure that Travis actually fails when the test script fails. 
    -  [x] Master  #3170. 
    -  [x] 0.14.x #3169 

**Everything below needs to be rebased onto 0. to ensure that a new build is actually triggered.**
 * [ ] 4. Move the cythonize script outside of the `skimage` directory. PR #3158.
 * [x] 1. Remove stray pip commands in travis install that don't use a requirements file. PR  #3227
 * [x] 2. Pin all requirements files and not just default.txt  PR #3227
 * [ ] 5. Use `pip install .` to install the package as opposed to calling the setup.py script directly.
    * Since we don't have a wheelhouse anymore, we can ask pip to install all the dependencies listed in the `setup.py` file itself.
    * Will have to use `pytest --pyargs skimage` as opposed to `pytest skimage` since scikit-image isn't installed in development mode anymore.
        * [ ] We install only build dependencies before the build.
 * [x] 6. Consistent Cython versioning: should all require `0.23.4`: PR #3171
 * [x] 7. Improve `build_versions.py` such that it is not hard coded. PR #3211
     * ~~Remove the tools/build_versions.sh script. This functionality is provided by `pip list` and breaks things as we add or remove dependencies.~~ @stefanv made good arguments to keep this scritpt PR #3181
 * [ ] 8. Remove matplotlib as a core dependency: WIP #3129
 * [x] 9. Prefer imageio over PIL (though PIL is still a dependency for tiffle): WIP #3126
 * [x] 10. Reduce the verbosity on the builds as travis is having a hard time keeping up. PR #3182
    * [x] Decrease Python messages
    * [x] run `pytest` without `-v`

Many of these fixes are in #3159, but that PR got too big to handle.

hmaarrfk:
I think this is mostly done. Remains to remove matplotlib as a core dependency. but that has its own PR.

PR #3221, opened 2018-06-27
WIP: Remove the wheelhouse in favour of pypi

hmaarrfk:
1. Remove stray pip commands in travis install that don't use a requirements file.
2. Pin all requirements files and not just default.txt
3. Remove the wheelhouse

Motivation to use the wheelhouse is that pypi is now quite good.,
We also had commands like 
```
pip install numpy
```
that didn't have pinning.
All of these changes fall into cleaning up the pip commands used during the build process.

### misc change
There was also the added addition of test duplication that was a cause of two large changes changes: a previous big fix to the build process #3072 and #3000 that removed python 2 compatibility

## References
Part of the larger effort to modernise the build process #3172

## For reviewers

  later.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -37,15 +37,16 @@ install:
   - "curl https://raw.githubusercontent.com/chemeris/msinttypes/master/stdint.h -o skimage/external/tifffile/stdint.h"
 
   # Install the build and runtime dependencies of the project.
-  # The --pre flag is necessary to grab a SciPy wheel, which is in
-  # pre-release at the time of writing (03-10-2017)
-  - pip install --retries 3 --pre -r requirements.txt
   - pip install --retries 3 -r requirements/build.txt
+  # scipy, from default is required to build until #3158 is fixed.
+  - pip install --retries 3 -r requirements/default.txt
   - python setup.py bdist_wheel bdist_wininst
   - ps: "ls dist"
 
   # Install the generated wheel package to test it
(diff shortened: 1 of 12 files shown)


matthew-brett:
I think you will need the wheels for installing in 32-bit Linux.  I suppose you may not want to support that, but one advantage is that it can be easier to debug 32-bit failures on Linux than Windows, and Windows 32-bit is still alive (about 30% of pip downloads).

matthew-brett:
Yes, Matplotlib does not provide 32-bit wheels, partly because it's test suite is terribly sensitive to small variations in the output of Freetype.  I believe the various Linux distributions do provide 32-bit packages.

hmaarrfk:
I don't envy their situation. Testing graphical interfaces is no fun.

They seem to provide 32bit wheels for windows, though not linux.
That said, it seems that the recommended way to develop (and install) applications is to use a virtual env so as to allow one to install the optimal versions of packages. 

Without wheels, users are stuck with the system version.

Maybe the more correct solution would be to enable circle-ci so we can actually test 32bit with a CI.

hmaarrfk:
@matthew-brett thanks for walking me through the packaging issues.

pypi packaging isn't as perfect as I thought it was in 2018. I guess that until we decide to remove 32bit support for linux, it is too early to remove code associated with the wheelhouse.

I have an other branch in the works that aims to cleanup the pip commands.

PR #3222, opened 2018-06-27
Build tools: Dedicate a --pre build in appveyor and ensure other builds don't download --pre

hmaarrfk:
Pre was only needed in 2017 before wheels became widely available.

pre was failing on appveyor.
now it is fixed because it doen't install everything in requirements from prereleases..... (and likely does what we want it to do)

@jni

## For reviewers

  later.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -39,15 +39,16 @@ install:
   - "curl https://raw.githubusercontent.com/chemeris/msinttypes/master/stdint.h -o skimage/external/tifffile/stdint.h"
 
   # Install the build and runtime dependencies of the project.
-  # The --pre flag is necessary to grab a SciPy wheel, which is in
-  # pre-release at the time of writing (03-10-2017)
-  - pip install --retries 3 --pre -r requirements.txt
   - pip install --retries 3 -r requirements/build.txt
+  # scipy, from default is required to build until #3158 is fixed.
+  - pip install --retries 3 -r requirements/default.txt
   - python setup.py bdist_wheel bdist_wininst
   - ps: "ls dist"
 
   # Install the generated wheel package to test it
-  - "pip install --pre --no-index --find-links dist/ scikit-image"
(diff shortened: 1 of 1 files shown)


jni:
> can appveyor or travis "warn" you that something is failing?

That's the point of the build matrix and having a specific build set aside for PRE. I don't know of a way to change the behaviour of the tick marks to be e.g. a red tick, probably that doesn't exist, but that is the burden of a maintainer: to investigate all failures and figure out whether it is actually blocking for the PR. =)

jni:
> The other question is: why did the pre travis builds not catch this?

It's entirely possible that pre builds exhibit some windows-specific failures.

hmaarrfk:
Right, so the windows builds don't have dedicated pre builds, which was weird.

they were always on pre.....

hmaarrfk:
@jni I just added a single pre build. It now fails. Less green, but at least testing is "more complete"

.appveyor.yml
```diff
@@ -39,15 +41,15 @@ install:
   - "curl https://raw.githubusercontent.com/chemeris/msinttypes/master/stdint.h -o skimage/external/tifffile/stdint.h"
 
   # Install the build and runtime dependencies of the project.
-  # The --pre flag is necessary to grab a SciPy wheel, which is in
-  # pre-release at the time of writing (03-10-2017)
-  - pip install --retries 3 --pre -r requirements.txt
```
hmaarrfk: @jni this is the problematic line.

hmaarrfk:
@jni, oh it still seems that non-pre builds are failure. what is with dask...

https://ci.appveyor.com/project/scikit-image/scikit-image/build/1.0.868/job/odd5r9ptvc6s8ugb#L3056

hmaarrfk:
@scikit-image/core what do we think about this?

soupault:
Looks good, thanks!

PR #3224, opened 2018-06-27
Enable flake8

hmaarrfk:
Flake8 was enabled, but we were ignoring the output so it might as well have been removed.
That said, it is useful to check unused variables, and undefined variables.

I disabled most of the style options that don't make sense when you write mathematical code,
added a few `# noqa` to keep the useful warnings for future code.

Things are mostly split by logical modifications in a few different commits 

I setup flake8 such that it would be much less strict than before and only run on a single build 

## For reviewers

  later.


hmaarrfk:
@scikit-image/core I don't know if you care about flake8, but I find it gives much better error message than running code through python when you have a syntax error. The commit message are now much improved compared to the first version I submitted a few months ago.

.travis.yml
```diff
@@ -38,7 +38,7 @@ matrix:
   include:
     - os: linux
       python: 3.5
-      env: OPTIONAL_DEPS=1 WITH_PYSIDE=1 BUILD_DOCS=1
+      env: OPTIONAL_DEPS=1 WITH_PYSIDE=1 BUILD_DOCS=1 TEST_FLAKE8=1
```
soupault: `RUN_FLAKE8` or smth?

.travis.yml
```diff
@@ -79,6 +79,15 @@ before_install:
     - which python; python --version
     - tools/build_versions.py
     - tools/check_sdist.py
+    # It is really annoying if flake8 causes your tests to fail first
+    # but in the event that you have a typo, flake8 is pretty good at
+    # catching it.
+    # Therefore, it is best to run flake8 first, on a single build
+    # with the oldest version of python we support
+    - if [[ $TEST_FLAKE8 == "1" ]]; then
```
soupault: Should all of this be in a dedicated `tools\flake8.py`?

.travis.yml
```diff
@@ -79,6 +79,15 @@ before_install:
...
+        pip install flake8;
+        flake8;
```
soupault: You probably need to keep the same flake8 arguments as before.

skimage/io/_plugins/util.py
```diff
@@ -6,11 +6,8 @@
 
 # utilities to make life easier for plugin writers.
 
-try:
-    import multiprocessing
-    CPU_COUNT = multiprocessing.cpu_count()
-except:
-    CPU_COUNT = 2
+import multiprocessing
```
soupault: Already in. https://github.com/scikit-image/scikit-image/pull/3434

.travis.yml
hmaarrfk: I think it is all in the `setup.cfg`. I added a dummy commit to trip up flake8 to test it.

.travis.yml
hmaarrfk: done. `flake8.sh`

.travis.yml
hmaarrfk: ok.

hmaarrfk:
@soupault see this build https://travis-ci.org/scikit-image/scikit-image/jobs/436291887#L905 showing we don't need to setup any flake8 command line arguments.

hmaarrfk:
Yeah. Let me think about this again. Flake8 is a little bit of an annoyance when it causes something to fail for no reason. This is especially true when they change their mind about the style guide and whatnot. 

The worst part about this script, is that if somebody else introduces a flake8 error, all future contributors would see red. 

My goal is to have contributors get to green as easily as possible. The goal of these tests should be to guide contributors through the mundane stuff (style, and test infrastructure), while helping them with the technical parts.

soupault:
@hmaarrfk well, a build failing with a redundant whitespace is not what I would like to see either. But having flake8 executed during the early phase of CI routine is pretty useful.
You're right, flake8 is PR-agnostic in this case, and may lead to a chain of red builds.

What is we adjust pep8speaks bot to perform only the most significant checks (as, currently, specified in the PR), and remove flake8 completely (or, at least, disable its influence on the build status)?

Other changes in the PR are very welcome.

setup.cfg
```diff
@@ -1,2 +1,47 @@
 [metadata]
 license_file = LICENSE.txt
+
+[flake8]
+ignore =
```
Borda: you can use full names instead of shortcuts

setup.cfg
```diff
@@ -1,2 +1,47 @@
...
+    F841,
+    # F401 'package' imported but unused. This shows up too much in __init__ files
+    F401,
+    # F403 'from .manage_plugins import *' used; unable to detect undefined names
+    F403,
+    # W391 blank line at end of file
+    W391
+
+exclude =
+    test_*,
```
Borda: why excluding tests?

skimage/io/__init__.py
```diff
@@ -4,7 +4,9 @@
 
 """
 
-from .manage_plugins import *
+from .manage_plugins import (use_plugin, call_plugin, plugin_info,
+                             plugin_order, reset_plugins,
+                             find_available_plugins, available_plugins)
 from .sift import *
```
Borda: why here is not the names list too?

skimage/measure/mc_meta/createluts.py
```diff
@@ -134,10 +134,11 @@ def search_line(lines, refline, start=0):
 
 
 
-def getLutNames(prefix):
+def getLutNames(prefix, luts):
```
Borda: from the review scope, it is not clear but `luts` does not seems to be used

skimage/segmentation/random_walker_segmentation.py
```diff
@@ -30,7 +30,7 @@ def new_del(self):
             pass
     umfpack.UmfpackContext.__del__ = new_del
     UmfpackContext = umfpack.UmfpackContext()
-except:
+except:  # noqa
```
Borda: an exception should be specified et leas by `Exception`

setup.cfg
hmaarrfk: ??? How?

Borda: my bad, it seems that it works only for pylint

setup.cfg
hmaarrfk: Tests often include many data arrays, that are well justified not to follow pep8.

skimage/io/__init__.py
hmaarrfk: you could. I think i gave up on changing style, I mostly like flake8 to find python typos, and not imposing style.

setup.cfg
jni: @hmaarrfk this is true, but test functions themselves should follow PEP8. I'd prefer finer-grained control here. Maybe it's not doable but I prefer to ignore these warnings rather than silence them completely.

skimage/segmentation/random_walker_segmentation.py
jni: I agree, and here we can probably do even better. @JDWarner might have some thoughts?

setup.cfg
hmaarrfk: I really believe that warnings should be set to 0, except when they would add more noise (like in examples).

The idea is that currently the flake8 stuff is just fully being ignore.

skimage/segmentation/random_walker_segmentation.py
hmaarrfk: These kind of maintenance changes really require 3 devs working according to our policies. Will not work with such large sweeping changes without focused effort ;)

stefanv:
Love this; thanks @hmaarrfk! Needs a quick rebase, hopefully not too honerous (err, may well be).

Reminds me about the conversations around black, @jni.  I'd be okay with the way it's used in Napari, e.g.

jni:
Yeah, pre-commit hooks are the right way to implement these things. Unfortunately, though, it makes it harder to get started developing for skimage...

hmaarrfk:
THe flake8 changes I had to actually enact are things like tabulation, and trailing whitespace, I see these as "minor" and better to deal with near the end of the PR, instead of upfront with a pre-commit hook

stefanv:
> THe flake8 changes I had to actually enact are things like tabulation, and trailing whitespace, I see these as "minor" and better to deal with near the end of the PR, instead of upfront with a pre-commit hook

Agreed, for now. My suggestion was for the long term, where I think we would benefit from enforced compliance that simultaneously relieves the burden on the contributor.

I don't think it's much of a problem to help new users with this.  For other projects, I have a single command they can run to install the commit hook, e.g. `make lint-install` (or, perhaps, `python .tools/install-lint.py`.

doc/release/contribs.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python
```
grlee77, line 1: Did this file get added to the PR by accident? It does not seem related to anything else here.

grlee77:
This now needs a rebase, but I would be fine with merging once the CI is green. We can always refine things again later as needed.

stefanv:
Implemented using ruff in #6729

PR #3225, opened 2018-06-27
Respect input array type in apply_parallel by default

jakirkham:
## Description

Sets `compute=None` by default and fixes this value to `True` or `False` based on the array type (if the user leaves `compute` unspecified) based on [this discussion]( https://github.com/scikit-image/scikit-image/pull/3121#pullrequestreview-132169980 ).

## Checklist
  existing benchmark
- [ ] Unit tests

## References

Follow-up for lazy `apply_parallel` added in PR ( https://github.com/scikit-image/scikit-image/pull/3121 ).

## For reviewers

  later.

### DIFF
--- a/skimage/util/apply_parallel.py
+++ b/skimage/util/apply_parallel.py
@@ -58,7 +58,7 @@ def _ensure_dask_array(array, chunks=None):
 
 
 def apply_parallel(function, array, chunks=None, depth=0, mode=None,
-                   extra_arguments=(), extra_keywords={}, *, compute=True):
+                   extra_arguments=(), extra_keywords={}, *, compute=None):
     """Map a function in parallel across an array.
 
     Split an array into possibly overlapping chunks of a given depth and
@@ -90,8 +90,10 @@ def apply_parallel(function, array, chunks=None, depth=0, mode=None,
     extra_keywords : dictionary, optional
         Dictionary of keyword arguments to be passed to the function.
     compute : bool, optional
-        Whether to compute right away (default) or
-        skip computing and return a dask Array.
+        If ``True``, compute eagerly returning a NumPy Array.
+        If ``False``, compute lazily returning a Dask Array.
+        If ``None`` (default), compute based on array type provided
+        (eagerly for NumPy Arrays and lazily for Dask Arrays).
 
     Returns
     -------
@@ -112,6 +114,9 @@ def apply_parallel(function, array, chunks=None, depth=0, mode=None,
     if not dask_available:
         raise RuntimeError("Could not import 'dask'.  Please install "
                            "using 'pip install dask'")
+
+    if compute is None:
+        compute = not isinstance(array, da.Array)
 
     if chunks is None:
         shape = array.shape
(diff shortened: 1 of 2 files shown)


skimage/util/apply_parallel.py
```diff
@@ -90,8 +90,9 @@ def apply_parallel(function, array, chunks=None, depth=0, mode=None,
     extra_keywords : dictionary, optional
         Dictionary of keyword arguments to be passed to the function.
     compute : bool, optional
-        Whether to compute right away (default) or
-        skip computing and return a dask Array.
+        Whether to compute eagerly or lazily (returning a Dask Array).
```
soupault: From this docstring it is not clear what `True` and `False` for `compute` actually do. Could you elaborate on the correspondence, please?

jni: what he said. ⬆️ 😛

skimage/util/apply_parallel.py
```diff
@@ -90,8 +90,9 @@ def apply_parallel(function, array, chunks=None, depth=0, mode=None,
...
+        By default eagerly or lazily computes based on array type.
+        So eagerly for NumPy Arrays and lazily for Dask Arrays.
```
jni: Indeed, this paragraph could be cleaned up a fair bit.

skimage/util/apply_parallel.py
```diff
@@ -113,6 +114,12 @@ def apply_parallel(function, array, chunks=None, depth=0, mode=None,
         raise RuntimeError("Could not import 'dask'.  Please install "
                            "using 'pip install dask'")
 
+    if compute is None:
+        if isinstance(array, da.Array):
+            compute = False
+        else:
+            compute = True
```
jni:
```
if compute is None:
    compute = not isinstance(array, da.Array)
```

skimage/util/tests/test_apply_parallel.py
```diff
@@ -47,8 +47,7 @@ def test_apply_parallel_lazy():
     # apply the filter on a Dask Array
     result2 = apply_parallel(threshold_local, d, depth=5,
                              extra_arguments=(3,),
-                             extra_keywords={'mode': 'reflect'},
-                             compute=False)
+                             extra_keywords={'mode': 'reflect'})
```
jni: I would very much like an additional test where the input is a Dask Array and `compute` is True.

skimage/util/apply_parallel.py
jakirkham: Feel free to suggest things here. ;) Had a hard time capturing this concisely.

skimage/util/tests/test_apply_parallel.py
jakirkham: Done in `test_apply_parallel`.

skimage/util/apply_parallel.py
jakirkham: Fixed.

skimage/util/apply_parallel.py
jakirkham: So I tried something. If it works or needs minor fixes, happy to clean up as needed. If it's way off, please suggest what you would like to see here instead. :)

jni:
Thanks @jakirkham and @soupault!

Issue #2956, opened 2018-02-04
Investigate Cython + Pythran as an acceleration option


stefanv:
Currently, we heavily employ Cython for optimization, but there is now an experimental feature to support NumPy constructs via Cythran.  We should investigate.

/cc @serge-sans-paille

serge-sans-paille:
@stefanv can you point me at some kernels that would be a good start to port?

stefanv:
I think any existing piece of Cython code would be a good target for benchmarking vs Pythran.

serge-sans-paille:
I've been wandering in the codebase and here is my educated guess: there is very few places where cython + pythran integration would be of any help, as most of the cython kernels don't make use of ufunc and cannot be transformed into using it with ease. yet most of the kernels are well withing the bounds of what pythran can compile, I provided a sample impelemntation here https://github.com/serge-sans-paille/scikit-image/commit/793612df141bb722a3d50fa83fbb3225c594f8f4 .

Let me know if you want me to dig that way!

jni:
@serge-sans-paille that's very clean! Nice!

However, I'm unable to compile, getting the error `NameError: name 'intp' is not defined`. Full traceback below:

```pytb
 ~/projects/scikit-image (feature/test-pythran-support)
 $ pip install --no-deps -U -e .
Obtaining file:///home/jni/projects/scikit-image
    Complete output from command python setup.py egg_info:
    /home/jni/miniconda3/envs/cf/lib/python3.6/site-packages/setuptools/dist.py:388: UserWarning: Normalizing '0.15dev' to '0.15.dev0'
      normalized_version,
    Partial import of skimage during the build process.
    WARNING: Disabling color, you really want to install colorlog.
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/home/jni/projects/scikit-image/setup.py", line 161, in <module>
        **extra
      File "/home/jni/miniconda3/envs/cf/lib/python3.6/site-packages/numpy/distutils/core.py", line 135, in setup
        config = configuration()
      File "/home/jni/projects/scikit-image/setup.py", line 81, in configuration
        config.add_subpackage('skimage')
      File "/home/jni/miniconda3/envs/cf/lib/python3.6/site-packages/numpy/distutils/misc_util.py", line 1024, in add_subpackage
        caller_level = 2)
[...]

PR #3226, opened 2018-06-27
Add Pythran support to build, convert two functions

serge-sans-paille:
## Description

Use Pythran instead of cython to compile several kernels, while keeping
somewhat high level and pythonic code.

## Checklist
  existing benchmark
- [x] Unit tests

## References

#2956 

## For reviewers

  later.


serge-sans-paille:
@jni / @stefanv : short status update. I updated kernel code to be pep8 compilant (wasn't the case for original cython code). Due to a mvsc update, some kernels no longer compile under windows but that's fixed in pythran master version.

Until then, I'd like some guidance on integration of pythran kernel into asv-powered benchmarks!

serge-sans-paille:
For the record: I just opened https://github.com/serge-sans-paille/pythran/pull/955 to make sure all scikit-image kernels are and remain correctly compiled by pythran. I'll feed the kernels with default values to make sure the behavior is unchanged, this is probably related to the bencmakrs task (I need guidance for valid / reasonable input)

serge-sans-paille:
For the record, https://github.com/serge-sans-paille/pythran/pull/955 now validates ok on windows.

serge-sans-paille:
@jni / @stefanv : all scikit-image kernels have been added to the pythran validation suite and compile OK on linux and windows! It's all in master branch, so next step on my side would be to make a Pythran release compatible with scikit-image, but I'd like to be sure there is no other blocker before.

serge-sans-paille:
> Or alternatively keep the existing Cython around and use some kind of setup.py flag to determine compilation with one lib or the other.

That looks better than Pure Python mode, performance wise. I can handle that task.

serge-sans-paille:
@jni just pushed a tentative cython fallback, but as stated in the commit message, I'm not sure that's the way we should take. But a Python fallback would be extreemly slow too...

jni:
@serge-sans-paille it's a very clean extension, thank you!

I absolutely sympathise with not wanting to maintain two separate code bases, believe me. =) But the reality is that we are quite risk averse at scikit-image, with good reason. So switching all-out to Pythran, which is a much younger project than Cython with a smaller community, is probably not feasible in the short term. But a hybrid approach might be feasible, and might be a good way for the core team to get practice working with Pythran, get confident in using it, and eventually decide to take the leap. The alternative is to leave this PR open for a long time, which is its own maintenance burden.

There's again another possibility, which is to wait until we go 3.6+, and then we can use this for Cython:

https://github.com/AlanCristhian/statically

I also presume something like this would be possible with Pythran, too? Actually a decorator is much more Pythonic than a comment.

@stefanv @soupault @emmanuelle do you guys have any input here? To summarise:

- Pythran offers similar performance to Cython while retaining pure Python syntax (using type inference for any internal code)
- It compiles to C++, so it doesn't have the distribution problems that Numba has
- @serge-sans-paille has made heroic efforts and converted I think *all* (?) of our Cython codebase to Pythran
- what now?

serge-sans-paille:
>  with good reason. So switching all-out to Pythran, which is a much younger project than Cython with a smaller community, is probably not feasible in the short term. But a hybrid approach might be feasible, and might be a good way for the core team to get practice working with Pythran,

OK, this totally makes sense. If I understand correctly, I won't need to play with the benchmarks, so my part of the job is almost done :-)

Concerning the hybrid path, It's probably better to give an option to the user to control which back-end to use, a flag to setup.py is feasible, or maybe an environment variable (easier to setup :-))?

I'll probably do a pythran release in august, that would include all the patches required for the scikit-image integration.

serge-sans-paille:
@jni up! Any step missing on my side?

jni:
@serge-sans-paille very cool. =) What do you think about the idea of implementing Pythran as a decorator instead of a comment + external command?

serge-sans-paille:
@jni. I'm not a big fan of this idea, because pythran is an ahead of time compiler: compile speed has never been a real focus, and it's not rare to have a compile step that takes 5 to 10 seconds. Even if cached, this is not acceptable for a decorator.

Moreover, Pythran only handles full modules. Being able to extract the relevant subpart of the module is an interesting subject and we may want to support that at some point, but it's definitly not possible now.

jni:
@serge-sans-paille very good points!

How about the middle ground: use Python 3.5+ type annotations, if available, instead of comments. But build still happens ahead of time.

serge-sans-paille:
@jni. Yeah, that's in the plan, once Python2.7 support is dropped.

serge-sans-paille:
Is that a blocking feature for that PR?

serge-sans-paille:
hi @jni. Just updated the integration step, with an optional env variable to disable pythran to lower the friction.

serge-sans-paille:
(rebased on master)

.appveyor.yml
```diff
@@ -1,5 +1,7 @@
 # AppVeyor.com is a Continuous Integration service to build and run tests under
 # Windows
+image:
+- Visual Studio 2017
```
hmaarrfk: I'm not sure conda is running Visual Studio 2017. Can you confirm that we would be able to ship this package with conda?

doc/Makefile
```diff
@@ -3,8 +3,8 @@
 
 # You can set these variables from the command line.
 PYTHON        ?= python
-SPHINXOPTS    ?= -j 1
-SPHINXBUILD   ?= python $(shell which sphinx-build)
+SPHINXOPTS    ?= -j $(shell nproc)
```
hmaarrfk: This doesn't really belong in this PR and seems to have been done in #3478 whats up with github?

doc/source/gitwash/development_workflow.rst
```diff
@@ -334,7 +334,7 @@ Rewriting commit history
 
    Do this only for your own feature branches.
 
-There's an embarassing typo in a commit you made? Or perhaps the you
+There's an embarrassing typo in a commit you made? Or perhaps the you
```
hmaarrfk: Haven't these been merged already? Whats up with github? #3480

skimage/_build.py
```diff
@@ -35,8 +38,7 @@ def cython(pyx_files, working_path=''):
     except ImportError:
         # If cython is not found, the build will make use of
         # the distributed .c files if present
-        c_files = [f.replace('.pyx.in', '.c').replace('.pyx', '.c') for f in pyx_files]
-        for cfile in [os.path.join(working_path, f) for f in c_files]:
+        for cfile in abs_c_files:
```
hmaarrfk: These changes are the subject of this PR #3311 . It has been in the works for a while, but keeps slipping as there are always more pressing issues. Is it necessary for your PR?

hmaarrfk:
So I am listening to your talk from 2013 (I guess that is what I found on Youtube) and it seems pretty awesome!!!

1. I don't think PyPi or conda (defaults or conda-forge) compiles with AVX instructions just yet, these are probably the two most popular ways to distribute scikit-image. Any thoughts on how that might affect the benchmarks?

The ability to use AVX instructions from native python is awesome!

2. I know some memory alignment is sometimes necessary to use AVX, does this impose a different minimum numpy version to take advantage of this.

skimage/_build.py
```diff
@@ -73,3 +75,33 @@ def process_tempita_pyx(fromfile):
...
+    The scenario first tries Pythran generation, then Cython generation,
+    then falls back to plain C code.
+
+    set SKIMAGE_PYTHRAN env to 0 to prevent using Pythran.
+    '''
+    working_path = kwargs.pop('working_path', None)
+    use_pythran = os.environ.get('SKIMAGE_NO_PYTHRAN', 1)
+    try:
+        from pythran.dist import PythranExtension
+        use_pythran &= 1
```
hmaarrfk:
```
  File "/home/mark/git/scikit-image/skimage/_build.py", line 94, in HybridExtension
    use_pythran &= 1
TypeError: unsupported operand type(s) for &=: 'str' and 'int'
```

skimage/_build.py
```diff
@@ -73,3 +75,33 @@ def process_tempita_pyx(fromfile):
...
+    Create a Pythran or Cython Extension, based on user flag and available
+    dependencies.
+
+    The scenario first tries Pythran generation, then Cython generation,
+    then falls back to plain C code.
+
+    set SKIMAGE_PYTHRAN env to 0 to prevent using Pythran.
+    '''
+    working_path = kwargs.pop('working_path', None)
+    use_pythran = os.environ.get('SKIMAGE_NO_PYTHRAN', 1)
```
hmaarrfk: Should be `SKIMAGE_PYTHRAN`

hmaarrfk: Well really `int(os.environ.get('SKIMAGE_PYTHRAN', '1'))`

hmaarrfk:
Trying to run hessian_matrix_det I ran into the following issues:

  * The pythran libraries seem to be output at the top level directories, not in their respective places. Anyway to fix that?
  * Needed to cast python floats to numpy floats.
  * It seems to run slower than cython (165 ms Pythran on my benchmark vs 100 ms Cython)

I don't think it is wise to have 2 code paths. It makes testing and debugging much harder. We should toward doing this piecemeal and have this be a benchmark driven exercise. 

<details> <summary>hessian_matrix_det benchmark</summary>

I ran this after rebasing and fixing the location of the so files manually.
```python
import numpy as np

from skimage import data, img_as_float
from skimage.feature import hessian_matrix_det


class HessianSuite:
    """Benchmark for exposure routines in scikit-image."""
    def setup(self):
        self.image = np.tile(img_as_float(data.coins()), (5, 5))

    def time_hessian_matrix_det(self):
        hessian_matrix_det(self.image)
```
</details>

hmaarrfk:
@serge-sans-paille you know pythran best, maybe pick the one where you think it would really shine. It doesn't have to be hessian. That was just the first one I noticed in this PR.

Honestly, it could just be generating an image pyramid where we need to convolve with simple kernels.

Your talks seemed to explain how numpy was particularly bad at that case due to data locality.

skimage/_build.py
hmaarrfk: nevermind, this is a different thing.

skimage/feature/_hessian_det_appx.py
```diff
@@ -0,0 +1,138 @@
...
+    c = _clip(c, 0, img.shape[1] - 1)
+
+    r2 = _clip(r + rl, 0, img.shape[0] - 1)
+    c2 = _clip(c + cl, 0, img.shape[1] - 1)
+
+    ans = img[r, c] + img[r2, c2] - img[r, c2] - img[r2, c]
+    return max(0, ans)
+
+
+# pythran export _hessian_matrix_det(float64[:,:], float)
```
hmaarrfk: if we have a single code path, I think casting to a numpy float is OK.

skimage/_build.py
```diff
@@ -73,3 +77,31 @@ def process_tempita_pyx(fromfile):
...
+
+def HybridExtension(name, py_sources, pyx_sources, **kwargs):
+    '''
+    Create a Pythran or Cython Extension, based on user flag and available
+    dependencies.
+
+    The scenario first tries Pythran generation, then Cython generation,
+    then falls back to plain C code.
+
+    set SKIMAGE_PYTHRAN env to 0 to prevent using Pythran.
```
hmaarrfk: `SKIMAGE_PYTHRAN` -> `SKIMAGE_NO_PYTHRAN`

skimage/feature/_hessian_det_appx.py
```diff
@@ -0,0 +1,138 @@
+import numpy as np
+
+
+def _clip(x, low, high):
```
hmaarrfk: Will Pythran be able to inline this?

serge-sans-paille: the underlying compiler, yes.

skimage/feature/_hessian_det_appx.py
serge-sans-paille: OK

serge-sans-paille:
@hmaarrfk I benchamrked hessian and found the same speed problem as you have. Th main issue is that Pythran fails to prove that all access are in-bounds. It should be deducible from clip, but it fails to do so. I opened a bug for that https://github.com/serge-sans-paille/pythran/issues/1059.

I'll try a simpler kernel :-)

I've also simplified the hybrid stuff to make sure the native moduels are generated in the correct place.

hmaarrfk:
@serge-sans-paille thanks for looking into that. I figured it was something fishy like that.

I hope you created a backup branch for all the other kernels you wrote.

serge-sans-paille:
@hmaarrfk changed to a simpler kernel, to test the whole thing. On my laptop, cython and Pythran performance are the same.

serge-sans-paille:
And added a simple benchmark \o/

tools/travis/script.sh
```diff
@@ -16,7 +16,7 @@ tools/build_versions.py
 section_end "List.installed.dependencies"
 
 section "Test"
-pytest $TEST_ARGS skimage
+PY_IGNORE_IMPORTMISMATCH=1 pytest $TEST_ARGS skimage
```
hmaarrfk: Is there a way to avoid needing this? This would warrant the creation of a "test" function again (I am the one who removed it). `skimage.test` or something like that.

serge-sans-paille: ``py.test`` gets confused by the ``.py`` vs ``.so`` files in the directory... do you have a better solution?

serge-sans-paille: cython does not have this issue because ``py.test`` ignores ``.pyx`` files...

hmaarrfk:
@serge-sans-paille I think you've done quite a bit already:
1. Extended the build infrastructure to enable hybrid pythran/cython. Cython is default.
2. Added a benchmark.
3. Attempted to speed up brief. 

Getting the build infrastructure right is going to be tricky. 

We still need to:
1. Run pythran compiled code on travis.
2. Run pythran compiled on appveyor

I'm not against this kind of `experimentation` but I don't have merge power either.
Getting the build infrastructure right to enable this kind of compilation is tricky and I don't want to see this work go to waste.

If `experimetnation` is allowed, I'm not sure we want experimental `py` files you created to be in the `non-experimental` subdirectory. This might confuse users looking deeper into the code. Maybe instead of hybrid compilation you could have a `pythran` subdirectory we place these experimental kernels and add benchmarks that compare the two approaches on the same graph. I tried to do this https://github.com/numpy/numpy/pull/11965 maybe a similar approach is applicable here?

tools/travis/script.sh
hmaarrfk: Not exactly. I get why you decided you kept the same extension but as is, `_brief_loop` still looks like fortran code ;) and would run horribly without compilation.

Maybe it is just time to bring back the `skimage.test` function, though I really like the simplicity of just running `pytest --pyargs skimage` since I don't have to remember each project's own special test functions.

serge-sans-paille:
@hmaarrfk green on windows, green on linux and one fail on OSX, due to installation issue (difficult for me to reproduce that, and no log does not help)

@jni we've been making significant improvements, what's the status on your side?

hmaarrfk:
1 build of travis will almost surely fail due to that kind of error. I with those issuing the PRs could restart their builds.

serge-sans-paille:
OK. I don't have the credentials to do that for skimage :-)
So let's say the build is fine! what is the next step?

hmaarrfk:
Take everything I say as a suggestion. 

I want to see the pythran code compiled and running in the default build. This is mostly because we need to test a "wheel"/conda package that has this integrated in it. Wheels have very specific restrictions on what can be shipped and conda likes to use their own compilers.

1. I'm not really a fan of `SKIMAGE_NO_PYTHRAN`, I would simply move the pythran accelerated stuff to `future` and compile them every run. If you need to disable them for python 3.5, do it as pythran is likely longer term and 3.5 support will likely be dropped (conda already dropped it).

2. Do you need a specific test to trigger that code path now that we moved it to a subdirectory?

Personally, performance equal to Cython isn't super exciting. But what is exciting is how this tool might be used in the future for things Cython is bad at (multiple types, fused types are a pain, reorderinng instructions, AVX).

I'm hoping that initially, we can separate `experimental code path` so that other devs can see what they can do with image processing and pythran without dealing with all these build issues themselves. I think this is the intent of https://github.com/scikit-image/scikit-image/tree/master/skimage/future

My thoughts are that once a benchmark shows clear advantage of the pythran code vs cython code, we transition it to the main library. We would need to test out wheels (and conda packages???) before doing so though.

Showing an advantage and building the infrastructure to do it are likely 2 separate steps in my mind. Both having their own challenges.

hmaarrfk:
I would wait for @jni's input before changing anything you have now. I'm just looking back on the experience I've had shipping out products that relied on compiled dependencies and it is nice to be able to test the shipping process.

hmaarrfk:
For me, equal performance is good enough for this "infrastructure step". 

Don't remove any of the work you have done just yet. The bounds checking issue was also pretty interesting. I mostly liked your reaction once you found the cause of the bug.

stefanv:
On Thu, 18 Oct 2018 04:28:45 -0700, Mark Harfouche wrote:
> @serge-sans-paille you know pythran best, maybe pick the one where you
> think it would really shine. It doesn't have to be hessian. That was
> just the first one I noticed in this PR.

FWIW, a code path that gets exercised a lot, and that also receives
numerous complaints for slowness (even though it is already optimized),
is image transformations & rescaling.

serge-sans-paille:
@stefanv can you point me to the exact kernel(s) involved?

stefanv:
Ah, these are not kernels, sorry.

hmaarrfk:
I would have to triple check, but I think a large fraction of the time is
calling `scipy.ndimage.gaussian_filter`

https://github.com/scikit-image/scikit-image/blob/master/skimage/filters/_gaussian.py#L133

I'm not sure if you are already working with scipy to accelerate their
stuff.

serge-sans-paille:
ok, after a first check, tehre is optimization room here, cool.

I'm still waiting for the build infrastructure to be approved before getting any further. better avoid big PR.

serge-sans-paille:
Just for the record, and even if that PR has been stalling for a while, I made some generic improvements to pythran that are relevant to at least one scikit-image kernel, so I think I'd share it here: http://serge-sans-paille.github.io/pythran-stories/basic-value-range-analysis.html

jni:
Thanks @serge-sans-paille ! That is a 😍 blog post!

serge-sans-paille:
Folks, now that scipy has been clearing the land and pythran has somehow passed the test of fire, would that make sense for you If I rebase & update this PR?

serge-sans-paille:
@jni this should pass validation now

grlee77:
Thanks @serge-sans-paille! I kicked off the GitHub Actions CI here to check it as well. It seems there is a related error coming up while running the the doctests. Is this one you have seen before?
https://github.com/scikit-image/scikit-image/pull/3226/checks?check_run_id=3524808791#step:5:218

Pythran also came up a little while back as a possibility in this open PR: https://github.com/scikit-image/scikit-image/pull/5362#issuecomment-828803090, if you happen to have any suggestions there (the PR itself is currently using Cython, but a Pythran implementation is in the comment I linked). The existing `rgb2hsv` and `hsv2rgb` conversions are quite slow and get >10x speedup using Cython or Pythran. 

Some other conversions like `xyz2lab`, `lab2xyz` and `luv2xy` should also benefit to some extent (there are also element-wise CUDA versions of these [in this file in cuCIM](https://github.com/rapidsai/cucim/blob/27da55a89321fead6847700868df4b40f8657204/python/cucim/src/cucim/skimage/color/colorconv.py#L309-L357)). I mention these as examples of simple functions where Pythran would provide a substantial acceleration over the current Python code.

serge-sans-paille:
@grlee77 this is an interesting issue: pytest is unhappy about a `.py` file being loaded as a `.so` file. It considers this could be a source of error. To workaround that, I've instructed pythran to generate a module named `a.so`  from a module named `a_pythran.py`. Can you trigger the gh-action tests?

Concerning  #5362, I'll take a deeper look soon :-)

serge-sans-paille:
\o/ All green \o/
Anything more I could do on this?

grlee77:
I agree that Pythran is more mature now and am not opposed to adding it to scikit-image.

For the specific function illustrated in this PR, though, I do not currently measure a benefit. It doesn't look like we currently have benchmarks for the `blob_doh` or `hessian_matrix_det` functions that make use of the Pythran code added here. When I tried the following simple script for this branch with and without Pythran enabled, performance was about 2.6x better with Cython. Do you see a similar result on your end? Naively, I would have expected fairly similar performance.


```Python
import numpy as np

from skimage import feature, data

# make a 2k x 2k image
img = np.tile(data.camera(), [4, 4])

%timeit feature.hessian_matrix_det(img)
```

I get with Pythran:
```
436 ms ± 11.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
vs. Cython
```
166 ms ± 2.97 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

serge-sans-paille:
Thanks for providing an easy way to benchmark. I'll investigate and report :-)

serge-sans-paille:
Problems spotted. I need a minor pythran release to fix them, but basically pythran needs some help through an assert to correctly compute array access bounds and perform unchecked array access. @grlee77 I'll ping you once that lands.

serge-sans-paille:
@grlee77 can you check performance with the new release?

stefanv:
I get very similar timings between the two now.

serge-sans-paille:
@stefanv that's already a first achievement : pure Python down to Cython-like performance. I'll try a kernel that yields speedup as a next step.

stefanv:
For me, the increased readability while matching Cython speed is a big win already!

serge-sans-paille:
Good, we have two kernels know, That's probably enough material for you to decide if we want to dive in that direction.

serge-sans-paille:
Gentle ping @grlee77 / @stefanv / @jni, what would be the next step if any?

grlee77:
> Gentle ping @grlee77 / @stefanv / @jni, what would be the next step if any?

I can also confirm the hessian matrix performance is now about the same as the Cython implementation. 

I am in favor of adding Pythran support to scikit-image. It is a new, but fairly lightweight build time dependency and should not add any new runtime dependencies. What do others think? In terms of time scale, should we target this for the 1.0 release?

stefanv:
I will ping the team about this.

rfezzani:
I have some concerns about adding new dependencies to the package as pointed out by @andyfaff https://github.com/scipy/scipy/issues/14335#issuecomment-877050156.
This is a risk we can't neglect :confused:

paugier:
> I have some concerns about adding new dependencies to the package as pointed out by @andyfaff [scipy/scipy#14335 (comment)](https://github.com/scipy/scipy/issues/14335#issuecomment-877050156). This is a risk we can't neglect confused

Note that this comment is outdated. Some times ago pythran depends on networkx and networkx changed and started to depend on too many packages (for example matplotlib!). So pythran included some networkx code and now does not depend on networkx anymore.

Now, one gets in a new empty virtualenv:

```
$ pip install pythran                            

Collecting pythran
  Using cached pythran-0.10.0-py3-none-any.whl (4.3 MB)
Collecting ply>=3.4
  Using cached ply-3.11-py2.py3-none-any.whl (49 kB)
Collecting beniget~=0.4.0
  Using cached beniget-0.4.1-py3-none-any.whl (9.4 kB)
Collecting numpy
  Downloading numpy-1.21.2-cp38-cp38-manylinux_2_12_x86_64.manylinux2010_x86_64.whl (15.8 MB)
     |████████████████████████████████| 15.8 MB 1.9 MB/s            
Collecting gast~=0.5.0
  Using cached gast-0.5.2-py3-none-any.whl (10 kB)
Installing collected packages: gast, ply, numpy, beniget, pythran
Successfully installed beniget-0.4.1 gast-0.5.2 numpy-1.21.2 ply-3.11 pythran-0.10.0
```

Gast and beniget are 2 small pure python packages maintained by @serge-sans-paille.

rfezzani:
OK, `pythran` will be a build dependency, so no impact at runtime from that side. I :-1: my previous concern ^^

I am not familiar with `pythran`, and I don't really know how wide is its adoption. Will it get as much support as `Cython` has?

Is the plan to drop `Cython` dependency or do we want to support both packages for accelerating Python code? I don't think that the last option is a good move :confused:.
Getting rid of `Cython` is a huge amount of work (54 `*.pyx` files in the repository), we can't only rely on @serge-sans-paille (thank you Serge by the way for your excellent job :wink:) to translate all the `Cython` code.

stefanv:
I think we should consider Pythran because it allows for cleaner code than Cython does at a comparable speed.

Cython activity has slowed down significantly, so here you can compare the two:

https://github.com/cython/cython/graphs/commit-activity
https://github.com/serge-sans-paille/pythran/graphs/commit-activity

Active contributors is about 2 vs 2 (maybe 1.5 :) ).  I can't comment on how easy it is to contribute to either project.

I think @serge-sans-paille is dedicated to keeping the project going, so I don't think the change would increase our risk  exposure by much.

serge-sans-paille:
Gentle ping ;-)

jni:
I think we need to delete `_hessian_det_appx.pyx` here?

jni:
@serge-sans-paille, @grlee77 and I are sprinting to make a push on 0.19, so I'm going to push some changes to this branch. Just letting you know in case you're still up and working on my comment from 10min ago. 😜 Thanks!

jni:
@serge-sans-paille looks like the Azure builds are failing with:

> msvcrt.lib(std_nothrow.obj) : fatal error LNK1112: module machine type 'x86' conflicts with target machine type 'x64'

See an example failure [here](https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=6990&view=logs&j=eeafdf3e-6518-596e-3f31-4f6be6dd535c&t=80b64133-034e-5caa-8a9b-a6fbec7619c5&l=1076).

Would you mind troubleshooting it?

jni:
Looks like the benchmark build fails also:

https://github.com/scikit-image/scikit-image/runs/4133309630?check_suite_focus=true#step:7:2385

I thought rebasing on main would fix it but that wasn't the issue. It looks like the GHA for benchmarks ([here](https://github.com/scikit-image/scikit-image/blob/a9e292d58585a1097d27e13cc3f95af10e1efeb4/.github/workflows/benchmarks.yml)) will need updating also...

serge-sans-paille:
@jni : trying to fix the issue you mentioned in #6014 here to help.

jni:
@serge-sans-paille Thanks!

I don't understand why but the Azure workflows aren't even launching now... 😂 

Re your comment on #6014:

> for some reason the xsimd package is not installed in the benchmark gh action.

Is it supposed to be there by default? Can it be installed? I notice that there is a [conda-forge feedstock](https://github.com/conda-forge/xsimd-feedstock), so if the benchmark env comes from conda then we should be able to install it...

jni:
I realise that I have no idea how the benchmark action creates its environments... Maybe @jaimergp can chime in? 😬 🙏

jaimergp:
@jni Dependencies are specified in [this matrix](https://github.com/scikit-image/scikit-image/blob/main/asv.conf.json#L19-L24). [More info](https://asv.readthedocs.io/en/stable/asv.conf.json.html#conda-channels),¡.

serge-sans-paille:
@jni fixed the azue-pipeline parts, investigating the benchmark part

serge-sans-paille:
@jni : green again!

jni:
Hmm, the benchmark build is still not working... Maybe the xsimd conda package doesn't ship the headers...? @jaimergp is there any way to build in the parent env and then just use the wheel in the asv env? Or any other ideas?

Also, we should change the benchmark action to run on the head and abort if the benchmarks don't run on the head. Currently we have to wait a long time just to know that only the base branch has been run. 🤦

jaimergp:
The `conda-forge` package does ship the headers:

```
xsimd-8.0.3-h4bd325d_1/include
└── xsimd
    ├── arch
    │   ├── generic
    │   │   ├── xsimd_generic_arithmetic.hpp
    │   │   ├── xsimd_generic_complex.hpp
    │   │   ├── xsimd_generic_details.hpp
    │   │   ├── xsimd_generic_logical.hpp
    │   │   ├── xsimd_generic_math.hpp
    │   │   ├── xsimd_generic_memory.hpp
    │   │   ├── xsimd_generic_rounding.hpp
    │   │   └── xsimd_generic_trigo.hpp
    │   ├── xsimd_avx.hpp
    │   ├── xsimd_avx2.hpp
    │   ├── xsimd_avx512bw.hpp
    │   ├── xsimd_avx512cd.hpp
    │   ├── xsimd_avx512dq.hpp
    │   ├── xsimd_avx512f.hpp
    │   ├── xsimd_constants.hpp
    │   ├── xsimd_fma3.hpp
    │   ├── xsimd_fma4.hpp
    │   ├── xsimd_fma5.hpp
    │   ├── xsimd_generic.hpp
    │   ├── xsimd_generic_fwd.hpp
    │   ├── xsimd_isa.hpp
    │   ├── xsimd_neon.hpp
    │   ├── xsimd_neon64.hpp
    │   ├── xsimd_scalar.hpp
    │   ├── xsimd_sse2.hpp
    │   ├── xsimd_sse3.hpp
    │   ├── xsimd_sse4_1.hpp
    │   ├── xsimd_sse4_2.hpp
    │   └── xsimd_ssse3.hpp
    ├── config
    │   ├── xsimd_arch.hpp
    │   ├── xsimd_config.hpp
    │   └── xsimd_cpuid.hpp
    ├── math
    │   └── xsimd_rem_pio2.hpp
    ├── memory
    │   ├── xsimd_aligned_allocator.hpp
    │   └── xsimd_alignment.hpp
    ├── stl
    │   └── algorithms.hpp
    ├── types
    │   ├── xsimd_all_registers.hpp
    │   ├── xsimd_api.hpp
    │   ├── xsimd_avx2_register.hpp
    │   ├── xsimd_avx512bw_register.hpp
    │   ├── xsimd_avx512cd_register.hpp
    │   ├── xsimd_avx512dq_register.hpp
    │   ├── xsimd_avx512f_register.hpp
    │   ├── xsimd_avx_register.hpp
    │   ├── xsimd_batch.hpp
    │   ├── xsimd_batch_constant.hpp
    │   ├── xsimd_fma3_register.hpp
    │   ├── xsimd_fma5_register.hpp
    │   ├── xsimd_generic_arch.hpp
    │   ├── xsimd_neon64_register.hpp
    │   ├── xsimd_neon_register.hpp
    │   ├── xsimd_register.hpp
    │   ├── xsimd_sse2_register.hpp
    │   ├── xsimd_sse3_register.hpp
    │   ├── xsimd_sse4_1_register.hpp
    │   ├── xsimd_sse4_2_register.hpp
    │   ├── xsimd_ssse3_register.hpp
    │   ├── xsimd_traits.hpp
    │   └── xsimd_utils.hpp
    └── xsimd.hpp
```

I think you can use `-E` to specify a manually created environment. See [here](https://asv.readthedocs.io/en/stable/commands.html#asv-continuous).

I'd try adding `--verbose` to the asv flags to see if we are missing something obvious in the env creation. 

> Also, we should change the benchmark action to run on the head and abort if the benchmarks don't run on the head. Currently we have to wait a long time just to know that only the base branch has been run.

This is worth an issue in the `asv` repo, I'd say.

serge-sans-paille:
@jni : where do you see the failing benchmark step?

jni:
@serge-sans-paille you need to remove then add the run-benchmarks label to see it. It then gets added to the latest run matrix but I guess it doesn't persist? I can't see it now. You can find the runs here:

https://github.com/scikit-image/scikit-image/actions/workflows/benchmarks.yml

Here's the latest run, which failed at the compilation step:

https://github.com/scikit-image/scikit-image/runs/4148854373?check_suite_focus=true

I've invited you to the repo with the triage role @serge-sans-paille so that you can re-run at will! :pray:

jaimergp:
The [benchmarks readme ](https://github.com/scikit-image/scikit-image/blob/main/benchmarks/README_CI.md#running-the-benchmarks-on-github-actions)has some more info if you need too!

serge-sans-paille:
@jni I've been able to reproduce the issue locally, and it seems to be codna specific. At least using virtualenv as asv backend works locally. To my great shame, I cannot find how to update the label, if you could retrigger it?

grlee77:
> To my great shame, I cannot find how to update the label, if you could retrigger it?

you just use the labels menu to the right and remove the "run-benchmark" label and then add it again to restart. I went ahead and did it just now, but something is odd because it completed almost instantly without doing anything: https://github.com/scikit-image/scikit-image/actions/runs/1454438133.

grlee77:
I pushed a couple of commits that installs Python 3.7 and `virtualenv` in the benchmarks workflow. It is now running at: https://github.com/scikit-image/scikit-image/runs/4194731156?check_suite_focus=true

serge-sans-paille:
@grlee77 ok, seems I don't have the rights to change the labels, thanks for doing it. Do you know if there's any advantage in using mamba/conda env vs. virtualenv as asv backend? Access to more binary packages like compilers etc?

grlee77:
> ok, seems I don't have the rights to change the labels

Did you get an invite to join the triage team? (I think @jni said he added you to it). I would have thought accepting that would have given you permission to add/remove labels.

> Do you know if there's any advantage in using mamba/conda env vs. virtualenv as asv backend? Access to more binary packages like compilers etc?

I don't think it should matter too much in our case. The dependencies we need should be available from pypi as well, so virtualenv is fine with me.

serge-sans-paille:
@grlee77 indeed, the invite went right to my spam folders :'(. I can now update labels.

All lights are green, do you want me to rebase and cleanup the history?

jni (approved):
@serge-sans-paille no need! The changes are actually compact enough that I think squashing is appropriate. Congrats! It's only taken three years to get this in!!! 😂 🎉 🚀 

Now that I am less worried about all of our infrastructure, I am looking more slowly at the actual implementation details and damn it's awesome. 😃 Thank you @serge-sans-paille!

serge-sans-paille:
\o/ I won't submit more kernel until next skimage release, so that we have some feedbacks about how wheels are built etc.

grlee77:
@serge-sans-paille
We have been updating our wheel building scripts in #6021 to prepare for a new release. After merging in this Pythran support I am seeing a new Windows failure related to unresolved symbols: https://github.com/grlee77/scikit-image/runs/4230656524?check_suite_focus=true#step:6:1399

If you have an idea of what might be the cause, please let us know!

<details>

```
       Creating library build\temp.win32-3.7\Release\skimage/feature\brief_cy.cp37-win32.lib and object build\temp.win32-3.7\Release\skimage/feature\brief_cy.cp37-win32.exp
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyImport_ImportModule
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyObject_GetAttrString
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyCapsule_Type
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyExc_RuntimeError
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyErr_SetString
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyExc_AttributeError
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyCapsule_GetPointer
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyErr_Format
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyErr_Print
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyExc_ImportError
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyModule_Create2
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_Py_BuildValue
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyModule_AddObject
    brief_pythran.obj : error LNK2001: unresolved external symbol __security_cookie
    brief_pythran.obj : error LNK2001: unresolved external symbol __security_check_cookie
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyArg_ParseTupleAndKeywords
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp__Py_NoneStruct
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyErr_Clear
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyType_IsSubtype
    brief_pythran.obj : error LNK2001: unresolved external symbol __CxxFrameHandler3
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyExc_ValueError
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyExc_MemoryError
    brief_pythran.obj : error LNK2001: unresolved external symbol memcpy
    brief_pythran.obj : error LNK2001: unresolved external symbol "void * __cdecl operator new(unsigned __int64)" (??2@YAPEAX_K@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "struct std::nothrow_t const std::nothrow" (?nothrow@std@@3Unothrow_t@1@B)
    brief_pythran.obj : error LNK2001: unresolved external symbol "void * __cdecl operator new(unsigned __int64,struct std::nothrow_t const &)" (??2@YAPEAX_KAEBUnothrow_t@std@@@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "void __cdecl operator delete(void *,unsigned __int64)" (??3@YAXPEAX_K@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::~basic_streambuf<char,struct std::char_traits<char> >(void)" (__imp_??1?$basic_streambuf@DU?$char_traits@D@std@@@std@@UEAA@XZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: virtual __cdecl std::ios_base::~ios_base(void)" (__imp_??1ios_base@std@@UEAA@XZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp__invalid_parameter_noinfo_noreturn
    brief_pythran.obj : error LNK2001: unresolved external symbol "void __cdecl operator delete(void *)" (??3@YAXPEAX@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol __std_terminate
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyDict_Size
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyDict_Next
    brief_pythran.obj : error LNK2001: unresolved external symbol memmove
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyExc_TypeError
    brief_pythran.obj : error LNK2001: unresolved external symbol memchr
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyEval_SaveThread
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyEval_RestoreThread
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_free
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) protected: __cdecl std::ios_base::ios_base(void)" (__imp_??0ios_base@std@@IEAA@XZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) protected: void __cdecl std::basic_ios<char,struct std::char_traits<char> >::init(class std::basic_streambuf<char,struct std::char_traits<char> > *,bool)" (__imp_?init@?$basic_ios@DU?$char_traits@D@std@@@std@@IEAAXPEAV?$basic_streambuf@DU?$char_traits@D@std@@@2@_N@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) protected: __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::basic_streambuf<char,struct std::char_traits<char> >(void)" (__imp_??0?$basic_streambuf@DU?$char_traits@D@std@@@std@@IEAA@XZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: int __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::sputc(char)" (__imp_?sputc@?$basic_streambuf@DU?$char_traits@D@std@@@std@@QEAAHD@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: void __cdecl std::basic_ios<char,struct std::char_traits<char> >::clear(int,bool)" (__imp_?clear@?$basic_ios@DU?$char_traits@D@std@@@std@@QEAAXH_N@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "bool __cdecl std::uncaught_exception(void)" (?uncaught_exception@std@@YA_NXZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: void __cdecl std::basic_ostream<char,struct std::char_traits<char> >::_Osfx(void)" (__imp_?_Osfx@?$basic_ostream@DU?$char_traits@D@std@@@std@@QEAAXXZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol strlen
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PySet_Type
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyObject_GetIter
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyIter_Next
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyObject_Not
    brief_pythran.obj : error LNK2001: unresolved external symbol __imp_PyCapsule_GetName
    brief_pythran.obj : error LNK2001: unresolved external symbol __std_exception_destroy
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: bool __cdecl std::ios_base::good(void)const " (__imp_?good@ios_base@std@@QEBA_NXZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: class std::basic_ostream<char,struct std::char_traits<char> > & __cdecl std::basic_ostream<char,struct std::char_traits<char> >::flush(void)" (__imp_?flush@?$basic_ostream@DU?$char_traits@D@std@@@std@@QEAAAEAV12@XZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "void __cdecl std::_Xlength_error(char const *)" (?_Xlength_error@std@@YAXPEBD@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "const type_info::`vftable'" (??_7type_info@@6B@)
    brief_pythran.obj : error LNK2001: unresolved external symbol "public: virtual void __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::_Lock(void)" (?_Lock@?$basic_streambuf@DU?$char_traits@D@std@@@std@@UEAAXXZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "public: virtual void __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::_Unlock(void)" (?_Unlock@?$basic_streambuf@DU?$char_traits@D@std@@@std@@UEAAXXZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "protected: virtual __int64 __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::showmanyc(void)" (?showmanyc@?$basic_streambuf@DU?$char_traits@D@std@@@std@@MEAA_JXZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "protected: virtual int __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::uflow(void)" (?uflow@?$basic_streambuf@DU?$char_traits@D@std@@@std@@MEAAHXZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "protected: virtual __int64 __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::xsgetn(char *,__int64)" (?xsgetn@?$basic_streambuf@DU?$char_traits@D@std@@@std@@MEAA_JPEAD_J@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "protected: virtual __int64 __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::xsputn(char const *,__int64)" (?xsputn@?$basic_streambuf@DU?$char_traits@D@std@@@std@@MEAA_JPEBD_J@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "protected: virtual class std::basic_streambuf<char,struct std::char_traits<char> > * __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::setbuf(char *,__int64)" (?setbuf@?$basic_streambuf@DU?$char_traits@D@std@@@std@@MEAAPEAV12@PEAD_J@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol "protected: virtual int __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::sync(void)" (?sync@?$basic_streambuf@DU?$char_traits@D@std@@@std@@MEAAHXZ)
    brief_pythran.obj : error LNK2001: unresolved external symbol "protected: virtual void __cdecl std::basic_streambuf<char,struct std::char_traits<char> >::imbue(class std::locale const &)" (?imbue@?$basic_streambuf@DU?$char_traits@D@std@@@std@@MEAAXAEBVlocale@2@@Z)
    brief_pythran.obj : error LNK2001: unresolved external symbol _fltused
    LINK : error LNK2001: unresolved external symbol _DllMainCRTStartup
    build\lib.win32-3.7\skimage\feature\brief_cy.cp37-win32.pyd : fatal error LNK1120: 69 unresolved externals
    error: Command "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\bin\HostX86\x86\link.exe /nologo /INCREMENTAL:NO /LTCG /DLL /MANIFEST:EMBED,ID=2 /MANIFESTUAC:NO /LIBPATH:C:\cibw\python\pythonx86.3.7.9\tools\libs /LIBPATH:C:\cibw\python\pythonx86.3.7.9\tools\PCbuild\win32 /LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\ATLMFC\lib\x86 /LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\lib\x86 /LIBPATH:C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\lib\um\x86 /LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.22000.0\ucrt\x86 /LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.22000.0\um\x86 /EXPORT:PyInit_brief_cy build\temp.win32-3.7\Release\skimage/feature/brief_pythran.obj /OUT:build\lib.win32-3.7\skimage\feature\brief_cy.cp37-win32.pyd /IMPLIB:build\temp.win32-3.7\Release\skimage/feature\brief_cy.cp37-win32.lib" failed with exit status 1120
    Building wheel for scikit-image (PEP 517): finished with status 'error'
    ERROR: Failed building wheel for scikit-image
  ERROR: Failed to build one or more wheels
  Failed to build scikit-image
  Exception information:
    File "C:\cibw\python\pythonx86.3.7.9\tools\lib\site-packages\pip\_internal\cli\base_command.py", line 173, in _main
      status = self.run(options, args)
    File "C:\cibw\python\pythonx86.3.7.9\tools\lib\site-packages\pip\_internal\cli\req_command.py", line 203, in wrapper
      return func(self, options, args)
    File "C:\cibw\python\pythonx86.3.7.9\tools\lib\site-packages\pip\_internal\commands\wheel.py", line 173, in run
      "Failed to build one or more wheels"
  pip._internal.exceptions.CommandError: Failed to build one or more wheels
  Removed build tracker: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\pip-req-tracker-qbd3p826'
Error: Command ['python', '-m', 'pip', 'wheel', 'D:\\a\\scikit-image\\scikit-image', '--wheel-dir=C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\cibuildwheelna3k2hgh\\built_wheel', '--no-deps', '-vv'] failed with code 1. None
```

<\details>

rgommers:
Just a guess: `clang-cl` is not installed in the wheel build actions, so the build uses MSVC and that doesn't work. See how `clang-cl` is installed in this PR.

grlee77:
> Just a guess: clang-cl is not installed in the wheel build actions, so the build uses MSVC and that doesn't work. See how clang-cl is installed in this PR.

Thanks, I will try this! In the meantime, I did confirm that AMD64 cases work okay and it is just x86 that is failing

grlee77:
Adding the following on GHA seems to fix things on 32-bit. I added `--force` to the choco install because otherwise it did nothing because it said LLVM was already present, but I wasn't sure if it was the same version.

```
      # install choco package manager
      - uses: crazy-max/ghaction-chocolatey@v1
        name: Install chocolatey

      # install clang-cl (needed for Pythran)
      - name: Install clang-cl
        run: |
          choco install --force --x86 llvm
          SET PATH="C:\\Program Files\\LLVM\\bin;%PATH%"
          SET PATH="C:\\Program Files (x86)\\LLVM\\bin;%PATH%"
```

PR #3231, opened 2018-06-28
Do not merge! Travis: explicitely run optional tests and QT test for OSX

### LINKED ISSUE #3084: Travis: OSX can't install `./requirements/docs.txt`
Version 0.14dev, on travis, we can't run the examples because we can't install `./requirements/docs.txt`

See
https://travis-ci.org/scikit-image/scikit-image/jobs/380507785#L9511

### LINKED ISSUE #3131: Simplify travis parameters
## Description
Currently we have a few travis parameters that pertain to handling how the tests/documentation behaves for different configurations.

These include:
  - `MINIMUM_REQUIREMENTS`
  - `OPTIONAL_DEPS`
  - `BUILD_DOCS`

I think the only necessary one is `MINIMUM_REQUIREMENTS`. The others were necessary for 2.7/3 compatibility. 

I think we should cleanup the logic such that 
  - `OPTIONAL_DEPS = not MINIMUM_REQUIREMENTS`
  - `BUILD_DOCS = not MINIMUM_REQUIREMENTS`

hmaarrfk:
This is being abbandonned in favour of 4 or 5 smaller PRs

  * https://github.com/scikit-image/scikit-image/pull/3421
  * https://github.com/scikit-image/scikit-image/pull/3422
  * https://github.com/scikit-image/scikit-image/pull/3422
  * https://github.com/scikit-image/scikit-image/pull/3423
  * https://github.com/scikit-image/scikit-image/pull/3431

Sorry for so many commits refractoring the tests. I know these PRs are the hardest to review since there can be silent failures. A previous PR of mine likely caused benefit 4. not to appear sooner.

These became necessary when trying to pull out the matplotlib dependency PR #3129. That PR is almost ready too, but I had to include these 50 line changes which I figured were easier to review as a separate PR.

The tests were refractored to remove much of the logic in how they are run. 
I think this became easier to do when we stopped supporting python 2.7.
I find that too many if and else statements make things harder to read:

### Noticeable benefits
  1. Added a OSX Python 3.6 build (python 3.7 is out now ;) )
  2. OSX also installs and tests Qt
  3. PyAMG is an optional dependency: it isn't treated as a special case anymore
  4. OSX wasn't really testing with optional dependencies. This hid a bug in the version of libpng that was installed in travis. I explicitly upgrade libpng in OSX now

### Test and build steps:
1. Compile
2. Install
3. Setup matplotlib (will be moved below in PR #3129)
3. Test with minimal
4. Install additional dependencies and Qt
5. Test again and report coverage
6. Build docs (This didn't work if I tried to enable it in every branch, it complained that files already existed, maybe because of caching??)

Note, we probably should refractor appveyor to run the same steps.

# Xref
Fixes #3131
Fixes #3084 
## For reviewers

  later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,16 +38,16 @@ matrix:
   include:
     - os: linux
       python: 3.5
-      env: OPTIONAL_DEPS=1 WITH_PYSIDE=1 BUILD_DOCS=1
+      env: BUILD_DOCS=1
     - os: linux
       python: 3.5
       env: QT=PyQt5 WITH_PYAMG=1 MINIMUM_REQUIREMENTS=1
     - os: linux
       python: 3.6
-      env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 BUILD_DOCS=1 DEPLOY_DOCS=1
+      env: QT=PyQt5 WITH_PYAMG=1 BUILD_DOCS=1 DEPLOY_DOCS=1
     - os: linux
       python: 3.6
-      env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 PIP_FLAGS="--pre"
+      env: QT=PyQt5 WITH_PYAMG=1 PIP_FLAGS="--pre"
     - os: osx
       osx_image: xcode9
       language: objective-c
@@ -88,18 +88,10 @@ install:
       else
         export MPL_DIR=${HOME}/.config/matplotlib
       fi
-    - mkdir -p ${MPL_DIR}
-    - touch ${MPL_DIR}/matplotlibrc
-    # Install most of the optional packages
-    - |
-      if [[ "${OPTIONAL_DEPS}" == "1" ]]; then
-        pip install --retries 3 -q -r ./requirements/optional.txt $WHEELHOUSE
-      fi
     - |
       if [[ "${WITH_PYAMG}" == "1" ]]; then
         pip install --retries 3 -q pyamg
       fi
-    - tools/travis/install_qt.sh
 
 script: tools/travis/script.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,13 +41,13 @@ matrix:
       env: BUILD_DOCS=1
     - os: linux
       python: 3.5
-      env: QT=PyQt5 MINIMUM_REQUIREMENTS=1
+      env: MINIMUM_REQUIREMENTS=1
     - os: linux
       python: 3.6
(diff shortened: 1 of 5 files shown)


grlee77:
I looked through the commits here and it looks good to me. Running the tests twice with and without optional dependencies is nice for completeness as long as everything is still well within the CI time limits.

unfortunately it looks like the PyQt5 problem is still the same as before, though

hmaarrfk:
At least they are all failing together. Ill have a look at them when I get
home tonight.

hmaarrfk:
Something is wrong with the matplotlibrc file. For some reason, matplotlib is going back to qt4 and not qt5....

hmaarrfk:
I stepped through the import process, apparently it was missing `sip`. Lets see what happens now.

hmaarrfk:
@grlee77 it now works! Its probably because pyqt5 removed it's `sip` dependency.

hmaarrfk:
FYI: @jni

jni:
You're running the full test suite twice per build now? I don't necessarily think this is a nice plan, as it will take longer for Travis to complete... (And it already takes a long time...)

jni:
I don't understand the sip discussion... So PyQt5 still depends on sip but they don't explicitly declare that dependency?

hmaarrfk:
Apparently there is a "new-first" option in `pytests`.
I'll have to explore it https://docs.pytest.org/en/latest/changelog.html#id64

hmaarrfk:
Ok the SIP issue was actually a Matplotlib, and not a qtpy issue that was fixed in this Commit:
https://github.com/matplotlib/matplotlib/commit/260e6d611afae9adaac528c28a4879097c357b74#diff-025508b6963efcfa97ba27389f2d2b86

I guess we have to wait until it gets pushed upstream.

hmaarrfk:
@jni, I would like to get the ball rolling on this one before travis breaks again. Actually the matplotlib PR  #3129 passed, but I would like to harden the build before we review that PR. It contains a subset of the modifications I made here.

You had an outstanding issue which was running the tests twice. You are right in that it takes more time to run the tests twice, but I think the main thing that takes time is the examples.

Those take a long time even on my machine.

I decided to actually create 2 linux builds just for the examples. (The OSX machines seems to finish in under 30 mins, so I think we have room to spare there).

Let me know what you think.

Note: Maybe cancel some of my previous builds on travis if you see this in time.

hmaarrfk:
@jni seems like my assumptions about the examples taking time was wrong. :smile_cat: 

Oh well, at least what is on a table seems to run each build in under 30 mins.

PR #3233, opened 2018-06-30
 MAINT: Release GIL where possible in watershed algorithm

lagru:
## Description
Releasing the GIL makes using the watershed algorithm in parallel easier / possible. As this is a computationally intensive function this should be worth it and have no performance penalties.

## Checklist
- [x] Unit tests (already exist)

## For reviewers

  later.


soupault:
@lagru could you, please, present the performance change on a sample data?

jni:
Yes, I thought I'd added a checklist item for adding benchmarks to `benchmarks/`... ;)

lagru:
@jni Yes, you did. However, two things made me hesitate.

- I'm not sure asv is able to handle and measure multiprocessing correctly but I guess it might still be useful to catch performance regressions
- The current state of the benchmarks directory compared to SciPy or NumPy made me think that you are still setting up asv for scikit-image and not ready to take PRs yet. You were missing the asv.conf.json (I now found it on the project's top level) and seem to be using one file for all benchmarks? Is the latter a temporary solution until there are enough benchmarks to start using a structure that more closely resembles scikit-images namespaces?

Anyway, I'll put benchmarks back on the menu. :smile:

jni:
I considered a full benchmark suite too ambitious for any one PR. Instead I suggest that any PR that touches some function should provide a benchmark for that function. This is *especially* true if the PR is about performance! ;)

I agree that asv might not be able to catch multicore appropriately. This is certainly true if the benchmark machine only has one core associated with the benchmarking task. @TomAugspurger any comments?

Regarding the structure: there should be one file per scikit-image top-level module, at least until we outgrow this structure.

TomAugspurger:
The benchmark machine has 8 cores.

I *think* the runner has all 8 available, but it's possible its being
limited somewhere.

lagru:
@soupault Here is a code snippet (using IPython!) that you can use to compare the effect multithreading has with and without releasing the GIL.

```python
In[1]: from multiprocessing.pool import ThreadPool
  ...: from skimage.morphology import watershed
  ...: from skimage.data import coins
  ...: image = coins()

In[2]: %%time
  ...: with ThreadPool(4) as pool:
  ...:     pool.starmap(watershed, ((image, 200) for _ in range(500)))
```

Opening a system monitor one can easily see how only the GIL-releasing version is able to use more than one CPU. On my machine (4 x Intel Core i5-4200U CPU @ 1.60GHz)  the results are:
- with GIL (f21c52bc248b59197129d60b97285e1579fee6bf): 
  ```
  CPU times: user 17.8 s, sys: 1.4 s, total: 19.2 s
  Wall time: 16.5 s
  ```
- without GIL (6526c87e5a8c60f0c3a05215b1e83c57926ddf49):
  ```
  CPU times: user 25.1 s, sys: 1.08 s, total: 26.2 s
  Wall time: 7.33 s
  ```

I'm not sure how to present this otherwise. At least this way you can quickly check the qualitative results yourself if you want.

lagru:
@jni With #3234 closed, are there any concerns left?

jni:
@lagru thanks for the ping!

1. Well, the idea of #3234 was to rebase this one after that one was merged. ;) A rebase is a good idea to get proper CI now that it's working. And,
2. It just occurred to me that #3234 doesn't cover the multi-threaded case. After rebase, could you add such a benchmark? It can be part of this PR, and it can more or less follow your informal benchmark above (threadpool.starmap), but with a smaller total runtime. =)

lagru:
Oh, you are right. I'll do that!

lagru:
@emmanuelle I already had the benchmark in my offline repository but didn't upload it for a while because I encountered several problems / hiccups (see review comments) which I wanted to fix before uploading. Sorry for the delay.

benchmarks/benchmark_morphology.py
```diff
@@ -36,3 +37,14 @@ def peakmem_reference(self, *args):
...
+
+
+class WatershedParallel(object):
+
+    def setup(self):
+        image = filters.sobel(data.coins())
+        self.images = ((image, 100) for _ in range(4))
+
+    def time_watershed_parallel(self):
+        with ThreadPool(4) as pool:
```
lagru: Initially I created the ThreadPool inside the setup method so that this initialization wouldn't taint the benchmark. However the benchmark would always fail when using `asv run` with `ValueError: Pool not running` despite passing with with `asv dev`. Moving the initialization into the `time_`-function seems to work now. I figure asv does some threading behind the scenes which interfered with the ThreadPool?

benchmarks/benchmark_morphology.py
```diff
@@ -36,3 +37,14 @@ def peakmem_reference(self, *args):
     def peakmem_watershed(self, seed_count, connectivity, compactness):
         morphology.watershed(self.image, seed_count, connectivity,
                              compactness=compactness)
+
+
+class WatershedParallel(object):
+
+    def setup(self):
+        image = filters.sobel(data.coins())
+        self.images = ((image, 100) for _ in range(4))
```
lagru: I'm not happy with the adaption of the informal benchmark to a smaller runtime. For the current configuration the GIL-releasing version is faster but only by a factor of 0.86. Increasing the number of "jobs" leads to the GIL-releasing version unexpectedly being slower! The same thing happens if I change the pool size meaning the current pool and job size of 4 seems to be a small sweet spot were the GIL-releasing version is slower.

Would be interesting if this only happens on my local hardware or if this behavior is universal. Right now I have a rather low confidence in the results of this particular benchmark...

emmanuelle:
thank you @lagru ! Just to be sure to understand, what would be a typical usecase where performance could be increased thanks to your modifications? When the watershed function is called on the same image with different parameters in different threads, for example?

lagru:
Also relevant, the contribution guide recommends this as well:
https://github.com/scikit-image/scikit-image/blob/942ed2996daffd88eec7e5d1e351c3171693b946/CONTRIBUTING.txt#L228-L229

sciunto:
kind ping to @emmanuelle and @jni on this? Is this PR ready for you?

PR #3236, opened 2018-07-01
Corrected docstring formatting so that code block is displayed as code

emmanuelle:
Very minor change :-). I built the doc to verify it does what it should. See http://scikit-image.org/docs/stable/api/skimage.measure.html#profile-line to see the problematic formatting

### DIFF
--- a/skimage/measure/profile.py
+++ b/skimage/measure/profile.py
@@ -51,6 +51,7 @@ def profile_line(image, src, dst, linewidth=1,
     The destination point is included in the profile, in contrast to
     standard numpy indexing.
     For example:
+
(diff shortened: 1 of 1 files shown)


hmaarrfk:

hmaarrfk:
@soupault here is the second test for meeseeksdev powers

hmaarrfk:
Got it!

JDWarner:

jni:
Hmm, for some reason this was not picked up by the old contrib script. (The new one works only by date so it was picking up too much non-0.14 noise...) Will add to notes.

Issue #2269, opened 2016-08-28
Build the doc with CircleCI?


emmanuelle:
At the moment, when a change is made to the documentation, the persons who are reviewing the PR have to (or should :-)) fetch the corresponding branch and build the doc on their machine. We could maybe use CircleCI and configure it to build the doc, so that it would be possible to see the new doc online, without having to build it locally.

What do you think?

sciunto:
I think there are pro and cons. Clearly, it will make our life easier for PR where new features with documentation are added because very often screenshots are not attached. On the other way, it will take time to setup and to maintain just because we don't have a hand on these servers and we will depend on them if them change anything in the configuration. For travis for example, we have to spend quite a lot of time to my taste to maintain configurations just because we do not always have the right versions or so.

soupault:
I agree, that would be a nice thing to have, although I've no idea how to approach it.
Basically speaking, we would like to have a fair web-deployed copy of documentation tree for each existing PR. That's a huge overhead.

Reasonable alternative is to enable `artifacts creation` for Travis (quick googling has indicated that there is such an option https://blog.travis-ci.com/2012-12-18-travis-artifacts/. In any case, at least AppVeyor has it), where on each commit a PR is not just being built, but also the outcomes of the building process are stored. This way, one could download automatically built copy of `skimage` and evaluate corresponding `.html` doc files. That is already much faster than manual building. To reduce the overhead for storing artifacts (I believe there is a limit of available storage for free accounts), we could save only `doc` folders as CI artifacts.
The most complex thing from my point of view is how do we select PRs of interest. At very rough estimate, only 20% of recent PRs introduce some changes to the documentation tree. Maybe we could use some of neat GitHub bots (for example, see https://github.com/rust-lang/cargo/pull/3039#issuecomment-242317255 and https://github.com/tensorflow/tensorflow/pull/3960#issuecomment-241488915) to detect if a PR changes anything in `doc` folder and trigger the corresponding option for Travis. This will, obviously, require some work to be implemented.

Finally, we have to developed a process to remove old/stored artifacts (applicable both for AppVeyor, Travis or even our website). Currently, I have no ideas how to make this work.

soupault:
Whoa! There is one more neat bot -> https://github.com/rust-lang/cargo/pull/3039#issuecomment-242478135 . This one can be triggered to automatically build and merge PRs. I we decide on this solution, I believe that it is not so hard to write a customized bots for our needs.

[...]

Issue #2097, opened 2016-05-21
Use stacklevel in warnings


stefanv:
When you do ``warn('blah')`` somewhere in scikit-image, the user gets *that* line-number whenever the warning is raised.  If the warning message is "you've used a deprecated keyword argument", e.g., the correct line to return is *where the user called skimage*.  To fix this, use the *correct* stacklevel argument in ``warn``.

For an example of correct usage, take a look at numpy.

blink1073:
Didn't we fix this in https://github.com/scikit-image/scikit-image/pull/1916?

meriki:
@stefanv @blink1073 Can I look into this?

blink1073:
Hi @meriki, yes, please!  At this point I believe the task is to make sure we always pass `stacklevel=2` in when we call `warnings.warn`.  It could be that #1916 already did that, in which case there is nothing more to do and we can close the issue.

arjunsuri:
@stefanv @blink1073 hey! I also want to look into the issue.

blink1073:
Sure, it is still an open question.

lagru:
The current implementation and usage introduced with #1916 doesn't show the behavior @stefanv intends. For that the default stacklevel would have to be 3 because `_shared.utils.warn` itself adds one stacklevel. This would cover most cases inside public functions. For callers that are more stacklevels deep this should be set manually. 

If you agree, I'll raise a PR.

Also, what was the reasoning for wrapping `warnings.warn`? I have a gut feeling that using `warnings.warn` would be more explicit and easier to read / understand.

hmaarrfk:
@lagru see https://github.com/scikit-image/scikit-image/pull/3438 help would be appreciated.

grlee77:
closed by #6183

PR #3245, opened 2018-07-05
FEAT: Flood fill in n-dimensions

### LINKED ISSUE #2876: Add flood fill algorithm

JDWarner:
# Flood fill has arrived!

Builds upon the excellent work in #3022 by @lagru.

I'm very excited to have this functionality in the package!  Took a bit of extra time, but I believe this is now fully general; adding the tolerance options makes it more generally useful.  Every option I can think of is included.  It's quite fast; should be suitable for interactive use in GUI widgets or similar.

Closes #2876.

Hopefully this can be backported to 0.14.

  later.


JDWarner:
Test failures appear entirely unrelated.

skimage/util/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,212 @@
...
+    while queue_pop(queue_ptr, &current_index):
+        # Look at all neighboring samples
+        for i in range(neighbor_offsets.shape[0]):
+            neighbor = current_index + neighbor_offsets[i]
+
+            # Indexing sanity check
+            if neighbor < 0:
+                continue
+            if neighbor >= image.shape[0]:
+                continue
```
lagru: Instead of just continuing for faulty indices it may be possible to raise an error like so:
```python
if neighbor < 0 or neighbor >= image.shape[0]:
    with gil:
        raise RuntimeError("encountered invalid index {} in _flood_fill_do_equal"
                           .format(neighbor))
```
I can't image no performance penalty as acquiring the GIL is only performed if the if-statement evaluates to true but would raise a nice explicit error message.

skimage/util/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,212 @@
...
+
+            # Shortcut if neighbor is already part of fill
+            if flags[neighbor] == UNKNOWN:
+                if image[neighbor] == seed_value:
+                # Neighbor is part of fill - but must check its neighbors too.
+                    flags[neighbor] = FILL
+                    queue_push(queue_ptr, &neighbor)
+                else:
+                    # Do not check this point again
+                    flags[neighbor] = NOT_FILL
```
lagru: I think this may possibly be simplified further.  If checking `flags[neighbor] == UNKNOWN` has nearly the same performance as `image[neighbor] == seed_value` omitting the outer if-else-clause (with the expensive `flags[neighbor] = NOT_FILL`) may have the same or better performance and nearly the same functionality:
```python
if image[neighbor] == seed_value:
    # Neighbor is part of fill - but must check its neighbors too.
    flags[neighbor] = FILL
    queue_push(queue_ptr, &neighbor)
```

May be worth trying out and doing a quick benchmark...

skimage/util/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,212 @@
...
+            # Indexing sanity check
+            if neighbor < 0:
+                continue
+            if neighbor >= image.shape[0]:
+                continue
+
+            # Only do comparisons on points not (yet) part of fill
+            if flags[neighbor] == UNKNOWN:
+                if image[neighbor] <= high_tol:
+                    if image[neighbor] >= low_tol:
```
lagru: Is there a reason you are using two nested if-else-statements which could easily be merged? Is this for readability? 

Actually I think that
```python
if low_tol <= image[neighbor] <= high_tol:
    # Neighbor is in fill; must check its neighbors too.
    flags[neighbor] = FILL
    queue_push(queue_ptr, &neighbor)
else:
    flags[neighbor] = NOT_FILL
```
is clearer.

skimage/util/_flood_fill_cy.pyx
JDWarner: Before adding these there were frequent faults due to referencing invalid neighbors (easy to reproduce, just take these out and start any fill at the origin; watch the kernel crash).  These are common at edges, and just need to be excluded.

Happy to combine the if-conditionals here and elsewhere.

lagru: Ah I see, I didn't think of the difference between how this algorithm handles the image edge vs the one in #3022.

skimage/util/_flood_fill_cy.pyx
lagru: You're right, this would endlessly queue values again and again. However I still think that the last else-block with `flags[neighbor] = NOT_FILL` can be removed which should have a slight performance advantage. That point will be checked again anyway via other queued neighbors either in `== UNKOWN` or `== seed_value`.

JDWarner:
@lagru You were right, removing the additional logic flagging checked points actually was beneficial - small but significant (10-15%) reduction in runtime by chucking the whole thing.  Out it goes!

Also combined `if`-statements and some minor cleanup.

JDWarner:
@lagru Also happens you were right about handling borders; there was a serious bug where the fill wrapped around.  Returned to a similar approach as in extrema, regression tests added.

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,182 @@
+"""flood_fill.py - inplace flood fill algorithm
+
+This module provides a function to fill all equal (or within tolerance) values
+connected to a given seed point with a different value.
+"""
+
+import numpy as np
+
+
+def flood_fill(image, seed_point, new_value, selem=None, connectivity=None,
```
soupault: Before diving into the code, I'd like to propose some changes to the API.
For me, one of the most common use cases for `flood fill` is to color the image regions given their boundaries. For this purpose, I'd expect the parameter `seed_point` to accept also a list of points with the corresponding labels, or a ndarray with binary (`True` marks the seed points of the same object class) or uint mask (non-zero elements mark the seed points and their values indicate the class index). I understand, that the current implementation is just one loop away from this API, but it is the change, I think, worth making.
Another thing: could you, please, explicitly specify the supported dimensionalities of `image`?

JDWarner: There are not any restrictions on the rank of the input image.

I have a few reservations to complicating the API further.  First, our hypothetical user who wants to wield the tool multiple times can do so as easily as building that input list with multiple calls and the inplace kwarg.  Second, the output logic would branch much further; off the top of my head it would be dangerous to allow input like this and also permit indices to be true for memory reasons.  Third, with tolerance available the result of successive fills is not commutative and this will result in user confusion.

If such functionality would be desired it would entail a simple wrapper for any specific use case, but making it general is not trivial - and the added logic would slow its use as a low level tool.  If we wanted to ship a multi_fill function in the future that would be fine, exporting all of these issues to a separate function and discussion.

Open to potentially be convinced otherwise, but for now I think that is out scope here.

skimage/util/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,190 @@
...
+                    ravelled_seed_idx, seed_value)
+            else:
+                _flood_fill_do_tol(image, flags, neighbor_offsets, &queue,
+                    ravelled_seed_idx, seed_value, high_tol, low_tol)
+        finally:
+            # Ensure that memory is released again
+            queue_exit(&queue)
+
+
+cdef inline void _flood_fill_do_equal(
```
lagru: Just a minor suggestion: How about naming this one `_fill_plateau` or `_fill_equal` and the other one below `_fill_with_tolerance`? But this is not a strong opinion.

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,182 @@
...
+    # Must annotate borders
+    working_image = _fast_pad(image, image.min())
+
+    # Correct start point in ravelled image
+    ravelled_seed_idx = np.ravel_multi_index([i+1 for i in seed_point],
+                                             working_image.shape)
+    neighbor_offsets = _offsets_to_raveled_neighbors(
+        working_image.shape, selem, center=((1,) * image.ndim))
+
+    # Use a set of flags; see _flood_fil.pyx for meanings
```
lagru: Typo in file name.

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,182 @@
...
+                    ravelled_seed_idx, seed_value, do_tol, high_tol, low_tol)
+    except TypeError:
+        if working_image.dtype == np.float16:
+            # Provide the user with clearer error message
+            raise TypeError("dtype of `image` is float16 which is not "
+                            "supported, try upcasting to float32")
+        else:
+            raise
+
+    original_slice = tuple(slice(1, -1) for _ in range(image.ndim))
```
lagru: Even if it may not matter in the big picture `(slice(1, -1),) * image.ndim` is faster and scales better.

skimage/util/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,190 @@
...
+        Lower limit for tolerance comparison (ignored if `do_tol` is 0).
+    """
+    cdef:
+        QueueWithHistory queue
+        unsigned char last_dim_in_neighbors
+
+    with nogil:
+        # Initialize a buffer used to queue positions while evaluating each
+        # potential maximum (flagged with 2)
+        queue_init(&queue, 64)
```
lagru: Actually, there is no need to initialize the queue outside the functions using it. It may even be worth it to remove this function entirely and just access the two wrapped functions directly from Python. Maybe worth it to make the algorithm's structure less deep.

skimage/util/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,190 @@
...
+    start_index : int
+        Start position for the flood-fill.
+    seed_value :
+        Value of `image[start_index]`.
+    """
+    cdef:
+        QueueItem current_index, neighbor
+
+    # Queue start position after clearing the buffer
+    queue_clear(queue_ptr)
```
lagru: You don't need to clear a freshly initialized queue.

JDWarner:
Making the Cython called directly from Python is the only suggestion I've yet to implement, @lagru; is there concern with changing the Cython functions from pure C to `cpdef` hybrids?  I thought there was some benefit to a Cython hybrid intermediary to get to pure C, but happy to do this if it would be beneficial.

The other point I'm unsure about when changing the architecture clearing the queue & freeing memory properly.

JDWarner:
@lagru thanks for the pointers.  Implemented, with `cpdef`.  Tested, no significant change in execution time but it is now simpler.

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,186 @@
...
+        returned with values equal to (or within tolerance) of the seed point
+        set to `new_value`.  If `indices` is true, a tuple of one-dimensional
+        arrays containing the coordinates (indices) of all found maxima.
+
+    See Also
+    --------
+    skimage.morphology.local_maxima
+    skimage.morphology.local_minima
+    skimage.morphology.h_maxima
+    skimage.morphology.h_minima
```
lagru: Why are referencing these methods? How are they related to flood-fill? In my opinion `morphology.watershed` has a stronger relation to this.

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,186 @@
...
+    >>> flood_fill(image, (0, 0), 5, tolerance=1)
+    array([[5, 5, 5, 5, 5, 5, 5],
+           [5, 5, 5, 5, 2, 2, 5],
+           [5, 5, 5, 5, 2, 2, 5],
+           [5, 5, 5, 5, 5, 5, 3]])
+    """
+    from ..morphology.extrema import (_resolve_neighborhood,
+                                      _set_edge_values_inplace, _fast_pad)
+    from ..morphology.watershed import _offsets_to_raveled_neighbors
+    from ._flood_fill_cy import _flood_fill_equal, _flood_fill_tolerance
```
lagru: This gives me the idea that this function would fit better into the `morphology` module. It's related to `watershed` and deals with topology inside a digital image... which sounds like an morphological operation to me. 
Furthermore we'd gain the advantage that we could move these lazy imports outside the function.

I hope I haven't missed a discussion on this. Are their reasons to include the algorithm here? Are you using lazy imports for performance reasons?

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,186 @@
...
+            raise
+
+    # Output what the user requested
+    original_slice = (slice(1, -1),) * image.ndim
+    if indices:
+        return np.nonzero(flags[original_slice] == 1)
+    else:
+        if not inplace:
+            output = working_image[original_slice]
+            output[flags[original_slice] == 1] = new_value
```
lagru: Just an idea: Have you experimented with setting `new_value` inside Cython? I not sure whether it would be faster but would be certainly interested in the result. :)

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,186 @@
...
+        same shape as `image`.
+    tolerance : ``selem`` type, optional
+        If None (default), adjacent values must be strictly equal to the
+        initial value of `image` at `seed_point`.  This is fastest.  If a value
+        is given, a comparison will be done at every point and this tolerance
+        on each side of the initial value will also be filled (inclusive).
+    inplace : bool, optional
+        If True, flood filling is applied to `image` inplace and nothing is
+        returned.  If False, the flood filled result is returned without
+        modifying the input `image`.  Ignored if `indices` is True.
```
lagru: > If True, [...] nothing is returned.

Right now this statement is wrong. If we pass in an empty image the function returns an empty array as well and not None (see line 118).
Why not always return the image? That way we could simplify the control flow in and the behavior of the function and remove the special cases as seen here.

However this is not a strong opinion. I really like the API already, especially that `*` in the function signature! :D

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,186 @@
...
+
+    Returns
+    -------
+    filled : ndarray or tuple[ndarray]
+        If `indices` is false, an array with the same shape as `image` is
+        returned with values equal to (or within tolerance) of the seed point
+        set to `new_value`.  If `indices` is true, a tuple of one-dimensional
+        arrays containing the coordinates (indices) of all found maxima.
```
lagru: If we keep the behavior and `inplace=True` returns `None` I really think we should mention this here.

skimage/util/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,143 @@
...
+        Value of `image[start_index]`.
+    """
+    cdef:
+        QueueWithHistory queue
+        QueueItem current_index, neighbor
+
+    # Initialize the queue and push start position
+    queue_init(&queue, 64)
+    queue_push(&queue, &start_index)
+    flags[start_index] = FILL
```
lagru: I would really recommend moving the above 2 lines into the try-finally-block as well. That way we can be sure the memory is released even if something goes wrong in these lines. Applies to the other function below as well.

skimage/util/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,143 @@
...
+    unsigned char FILL = 1
+    # Not checked yet
+    unsigned char UNKNOWN = 0
+
+
+cpdef inline void _flood_fill_equal(dtype_t[::1] image, 
+                                    unsigned char[::1] flags,
+                                    Py_ssize_t[::1] neighbor_offsets, 
+                                    Py_ssize_t start_index, 
+                                    dtype_t seed_value) nogil:
```
lagru: Right now the `nogil` statement has no effect on this function when calling from Python. I'd recommend enclosing everything after line 64 with a `with nogil:` block. 
And I'm not sure how this would interact with the `cpdef`, `inline` and `nogil` in the functions signature so I'd change that one to `def` for now.

skimage/util/_flood_fill.py
JDWarner: This decision was mine.  I view `flood_fill` as a very general operation, for example it can be used as a pseudo-threshold (with a wide tolerance anyway, other similar algorithms are in `filters` & `segmentation`), or for interactive drawing (`draw`, or `viewer`), color substitution (`color`), and yes also `morphology` - though I would posit that most users would think of it less as a morphological tool than the other uses.  With so many use cases, it seemed most appropriate to expose as a general utility function in `util` for ease of accessibility.  As the structure was based on extrema finding, I borrowed several helper functions.

The lazy imports are unfortunately required because the util module is brought in as a partial import during the build process, prior to Cython extensions being built.  Attempting to import any Cython at top level makes it impossible to build scikit-image.

skimage/util/_flood_fill.py
JDWarner: That's a good point, and I haven't.  If it works and has some benefit, might be worth considering refactoring to `_shared` because this is a really common operation.

skimage/util/_flood_fill.py
JDWarner: Some good points here.  I think the majority of NumPy/SciPy `inplace` functions do not return anything, which is what I was emulating, but I am not against returning the inplace modified image.  The empty array case logic definitely needs to be revised, good catch there.

skimage/util/_flood_fill.py
JDWarner: Gave it a shot, there was not a significant difference.

JDWarner:
Thanks for the review, @lagru!  Everything now implemented except the API suggestion (returning input image when inplace is True).  I am not opposed to returning the inplace-modified image, if another couple sets of eyes from @skimage/core agree.

If anyone has ideas about alternatives for the lazy imports I'm all ears.  Tried a couple things but no working alternatives yet.

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,185 @@
...
+        is given, a comparison will be done at every point and this tolerance
+        on each side of the initial value will also be filled (inclusive).
+    inplace : bool, optional
+        If True, flood filling is applied to `image` inplace and nothing is
+        returned.  If False, the flood filled result is returned without
+        modifying the input `image`.  Ignored if `indices` is True.
+
+    Returns
+    -------
+    filled : ndarray or tuple[ndarray]
```
lagru: I don't want to be nitpicky but you could add an `or None` here as well?

skimage/util/_flood_fill.py
lagru: Reading this I have to agree, using utils seems justified. But maybe we could do something to enhance its discoverability by adding this function to the see-also section for similar functions (I only know of watershed)?
And thank you for providing your thoughts on this decision.

> The lazy imports are unfortunately required because the util module is brought in as a partial import during the build process, prior to Cython extensions being built. Attempting to import any Cython at top level makes it impossible to build scikit-image.

Interesting, I didn't know that!

skimage/util/_flood_fill.py
lagru: Which functions do you have in mind? I was under the impression that NumPy uses the `out` parameter to provide this kind of functionality and still returns the reference even if `out` is provided, e.g. [mean](https://docs.scipy.org/doc/numpy/reference/generated/numpy.mean.html). The only place I could find this API in SciPy ([winsorize](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.winsorize.html#scipy.stats.mstats.winsorize)) the output is still returned. 

Anyway, your argument is quite good, consistency with NumPy / SciPy is desirable. Right now I'm just not sure which way is the consistent one..

lagru:
I don't want to push the scope of this PR but an idea to keep around for the future might be to try modifying this into a [Scanline Fill](https://en.wikipedia.org/wiki/Flood_fill#Scanline_fill). Shouldn't be to hard to do in a  successive PR and might yield an even faster function.

skimage/util/_flood_fill.py
JDWarner: @lagru I went looking for the API examples I thought I recalled and can't seem to find any.  There are some instances of this in our hidden helper functions, and Pandas does a decent bit of this sort of thing, but I was on the wrong side of this one.  Very diplomatic response, but you're just right.  Also returning in this case lets users call the function and continue in that line with further NumPy methods.

So thanks for the persistence and I'll be making this change.

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,180 @@
...
+           [5, 5, 5, 5, 2, 2, 5],
+           [5, 5, 5, 5, 2, 2, 5],
+           [5, 5, 5, 5, 5, 5, 3]])
+    """
+    from ..morphology.extrema import (_resolve_neighborhood,
+                                      _set_edge_values_inplace, _fast_pad)
+    from ..morphology.watershed import _offsets_to_raveled_neighbors
+    from ._flood_fill_cy import _flood_fill_equal, _flood_fill_tolerance
+
+    image = np.asarray(image, order='C')
```
lagru: Will this clash with `inplace=True`? If `image` is not C-contiguous this will perform a copy and the original input image is lost.

skimage/util/_flood_fill.py
```diff
@@ -0,0 +1,180 @@
...
+
+This module provides a function to fill all equal (or within tolerance) values
+connected to a given seed point with a different value.
+"""
+
+import numpy as np
+
+
+def flood_fill(image, seed_point, new_value, *, selem=None, connectivity=None,
+               indices=False, tolerance=None, inplace=False):
```
jni: Don't mean to railroad you but both @stefanv and myself feel like util is not the right place for this... I suggest having it in morphology and cloning it to segmentation. 

Long term, we could consider a "Special" group for things that are more user-facing than util but still generic.

I guess the question I would ask is, if a user is crawling through the segmentation docs, do we want them to stumble upon this function?

lagru: An alternative would be to add entries to the right `See Also` sections in segementation.
(In case you missed it, there was already an earlier [discussion](https://github.com/scikit-image/scikit-image/pull/3245#discussion_r202028213) about this.)

JDWarner:
Latest patch is small but pretty neat; Fortran and C-contiguous arrays are both now supported.

JDWarner:
Summary of changes discussed at SciPy sprints with @jni and @stefanv 

* Home in `morphology`, exposed also in `segmentation`
* Split function into two to avoid conditionally branching output, second is `flood_fill_mask`
* Revise output to return the mask rather than indices; encourage users to convert mask to indices if they have memory concerns on repeated runs.  This is the fastest and most Pythonic in the vein of 'one best way'.
* Both function signatures pruned, now cleaner
* Revised helper functions as needed to allow either C- or Fortran contiguous memory.

jni:
@JDWarner 

Mind rebasing to see whether we can get some greenery happening in the CI? =)

I also would like to reconsider the API to be more along the lines of `flood_fill` returning a mask. (I know we already discussed this in person but want to see how others feel about this also.) In most uses I envision, I will perform the computation not on the image I want filled, but on e.g. a Sobel-filtered image. I also feel like this would be a much more general approach, and it wouldn't suffer from e.g. in-place modification of the input (which we are not doing now but which is the natural interpretation of the fill operation). Example usage:

```
image[flood_fill(image_sobel, (256, 256), tolerance=0.05)] = (255, 0, 0)
```

I think returning a mask enables the use of flood-fill in much more general settings. I know we have `flood_fill_mask` for this, but I find that name quite off-putting!

How about just `flood`?

```
image[morphology.flood(image_sobel, (256, 256), tolerance=0.05)] = (255, 0, 0)
```

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,294 @@
...
+    """Find mask corresponding to a flood fill applied to an image.
+
+    Starting at a specific `seed_point`, connected points equal or within
+    `tolerance` of the seed value are found, returned as a mask.
+
+    Parameters
+    ----------
+    image : ndarray
+        An n-dimensional array.
+    seed_point : tuple or int
```
emmanuelle: why "or int"? Is it for the case of 1-D array? I find this "or int" a bit confusing, but maybe I'm missing something.

emmanuelle:
This is impressive work! I hope we can merge it soon. Since it is a new feature, there should also be a gallery example. I'm curious to know for which image processing tasks/ pipelines you would typically use this algorithm?

jni:
@JDWarner :wave: do you have any comments on my suggestion above? Also, if you don't have time to push this through now, do you mind if I take over?

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,294 @@
...
+def flood_fill(image, seed_point, new_value, *, selem=None, connectivity=None,
+               tolerance=None, inplace=False):
+    """Perform flood filling on an image.
+
+    Starting at a specific `seed_point`, connected points equal or within
+    `tolerance` of the seed value are found, then set to `new_value`.
+
+    Parameters
+    ----------
+    image : ndarray
```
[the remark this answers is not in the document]
JDWarner: But they are technically supported.  I'm not entirely sure why someone would want to fill across channels, but that's absolutely possible.  This is a low-level tool, and I'm strongly inclined to follow a `scipy.ndimage` style where, if the tool _can_ operate on/with it, we place no further restrictions (express or implied).

skimage/morphology/_flood_fill.py
JDWarner: Yes, it's a convenience for the 1-D case.

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,294 @@
...
+        A structuring element used to determine the neighborhood of each
+        evaluated pixel. It must contain only 1's and 0's, have the same number
+        of dimensions as `image`. If not given, all adjacent pixels are
+        considered as part of the neighborhood (fully connected).
+    connectivity : int, optional
+        A number used to determine the neighborhood of each evaluated pixel.
+        Adjacent pixels whose squared distance from the center is larger or
+        equal to `connectivity` are considered neighbors. Ignored if
+        `selem` is not None.
+    tolerance : ``selem`` type, optional
```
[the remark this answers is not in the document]
jni: Although I very much like the idea of NumPy arg compatibility in general, I dislike `atol` and `rtol`. `abs_tolerance` or `absolute_tolerance` or `tolerance_absolute` are all preferable imho, if you feel that specifying the type of tolerance is important.

jni:
@JDWarner @scikit-image/core any thoughts on my API [suggestion](https://github.com/scikit-image/scikit-image/pull/3245#issuecomment-413479618) for just `flood`?

Also, thinking a bit more about this, is flood fill a morphology operation or a segmentation operation?

skimage/morphology/_flood_fill.py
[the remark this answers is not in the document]
JDWarner: Will clarify the dtype.  I'm in agreement with @jni on tolerance.

skimage/morphology/__init__.py
```diff
@@ -12,7 +12,7 @@
 from .greyreconstruct import reconstruction
 from .misc import remove_small_objects, remove_small_holes
 from .extrema import h_minima, h_maxima, local_maxima, local_minima
-from ._flood_fill import flood_fill, flood_fill_mask
+from ._flood_fill import flood, flood_fill
```
jni: Whoo! =D

JDWarner:
@jni after consideration I like `flood` (now replaces `flood_fill_mask`).

I have not added a References section as this algorithm doesn't strictly match a canonical flood fill, especially as the implementation is n-D, abstracted, and supports a tolerance.  There is no other implementation in Python I am aware of that offers these features.

Per @emmanuelle - this is incredibly widely used in GUI workflows.  From the very early GUI raster editors (even those as limited as MS Paint) there was a bucket tool; that's what this gives us.  However, the bucket is often pretty limited when applied to real images where color gradients and sensor noise even in flat fields are the norm.  That's why, when I saw how, I implemented it with a tolerance also.  If we get around to looking at a GUI again, in Jupyter[lab] or otherwise, this is basically expected.

It's such a basic operation it almost defies being placed in a specific category.  Algorithmically it's most similar to morphology, but I expect more than half of users would intuitively think about it like segmentation; limiting it to only one of these might be a mistake.

skimage/morphology/__init__.py
JDWarner: :+1:

JDWarner:
@emmanuelle now with gallery example.  The single branch build Travis failure is unrelated.

JDWarner:
@scikit-image/core this is ready for final review.

doc/examples/segmentation/plot_floodfill.py
```diff
@@ -0,0 +1,120 @@
...
+
+"""
+.. image:: PLOT2RST.current_figure
+
+Flood as mask
+-------------
+
+A sister function, `flood`, is available which returns a mask identifying the
+flood rather than modifying the image itself.  This is useful for
+segmentation purposes and more advanced analysis pipelines.
```
jni: Any chance you could add a section to use my example in the earlier comment (painting a color image)?

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,290 @@
...
+            _flood_fill_equal(working_image.ravel(),
+                              flags.ravel(),
+                              neighbor_offsets,
+                              ravelled_seed_idx,
+                              seed_value)
+    except TypeError:
+        if working_image.dtype == np.float16:
+            # Provide the user with clearer error message
+            raise TypeError("dtype of `image` is float16 which is not "
+                            "supported, try upcasting to float32")
```
jni: Why is this the case?

doc/examples/segmentation/plot_floodfill.py
```diff
@@ -0,0 +1,120 @@
...
+    m = (1 + i)//3
+    ax[m, n].imshow(output[i], cmap=plt.cm.gray)
+    ax[m, n].set_title('Tolerance {0}'.format(str(5 + 20*i)))
+    ax[m, n].axis('off')
+
+fig.tight_layout()
+plt.show()
+
+"""
+.. image:: PLOT2RST.current_figure
```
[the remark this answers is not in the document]
JDWarner: I'm not either in honesty, for getting multiple images to show this is the motif I've used previously.  Didn't see another option in our gallery but if there is, I'd be happy to tweak this.

doc/examples/segmentation/plot_floodfill.py
```diff
@@ -0,0 +1,120 @@
...
+# Initialize plot and place original image
+fig, ax = plt.subplots(nrows=3, ncols=3, figsize=(12, 12))
+ax[0, 0].imshow(cameraman, cmap=plt.cm.gray)
+ax[0, 0].set_title('Original')
+ax[0, 0].axis('off')
+
+# Plot all eight different tolerances for comparison.
+for i in range(8):
+    n = (1 + i)%3
+    m = (1 + i)//3
```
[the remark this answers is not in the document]
hmaarrfk: Or you can use my favorite function:

`for i, (n, m) in enumerate(np.ndindex((3, 3)):`

skimage/morphology/_flood_fill.py
JDWarner: Float16 is not among the fused types in Cython, so catch and pass an informative error.  Personally I feel float16 is dangerously low precision warranting its exclusion; precision error propagates fast.  We excluded float16 in the initial code supported by @lagru also.

doc/examples/segmentation/plot_floodfill.py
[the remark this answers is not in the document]
JDWarner: I've restyled the example, hoping the 'magic' properly places intermediate figures.

jni:
@JDWarner oops! Need to import numpy now. I've pushed a fix adding the import to the example.

jni:
@scikit-image/core this needs another review! ;)

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,290 @@
...
+    """
+    # Correct start point in ravelled image - only copy if non-contiguous
+    image = np.asarray(image)
+    if image.flags.f_contiguous is True:
+        order = 'F'
+    elif image.flags.c_contiguous is True:
+        order = 'C'
+    else:
+        warnings.warn('Non-contiguous array passed as `image`; this will be '
+                      'converted to a contiguous array as a copy.')
```
lagru: I really like how explicit this!  But you may want to add `stacklevel=3` here to raise the warning on the stack level of the user code instead. Same for the warning in `flood_fill`.

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,290 @@
...
+            output = image.copy()
+        else:
+            output = image
+        output[flood(image, seed_point, selem=selem, connectivity=connectivity,
+                     tolerance=tolerance)] = new_value
+        return output
+    else:  # inplace
+        image[flood(image, seed_point, selem=selem, connectivity=connectivity,
+                    tolerance=tolerance)] = new_value
+        return image
```
lagru: I find the nested workflow above confusing to read at a first glance. Maybe there is a simpler solution? Wouldn't

```python
if inplace and not image.flags.contiguous:
    warnings.warn('Non-contiguous array passed as `image`; this will '
                  'be converted to a contiguous array as a copy.')
    # Flood fill requires some sort of contiguity - this makes a copy
    image = np.ascontiguousarray(image)
elif not inplace:
    image = image.copy()

mask = flood(image, seed_point, selem=selem, connectivity=connectivity,
             tolerance=tolerance)
image[mask] = new_value
return image
```

do the same?

lagru: Wait, I think we don't even need to check for contiguity here because `flood` will do that for us. `image[mask] = new_value` should work for non-contiguous arrays just fine. The only case where this would bite us is if we pass in a non-contiguous image with `inplace=False` in which case the image would be copied twice (once in `flood_fill` and once in `flood`). So maybe having a check for contiguity in both functions is for the best.

lagru:
Do you plan to add benchmarks for these functions at a later stage?

jni:
@lagru I suggest we leave benchmarks for a future PR since we are still coming to grips with our benchmark infrastructure...

lagru:
@jni Fine by me. Has there been an issue or thread related to this I missed?

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def flood(image, seed_point, *, selem=None, connectivity=None, tolerance=None):
```
lagru: Has there been a discussion to add a parameter `indices` that will change the returned output similar to the behavior in `local_maxima`? 

https://github.com/scikit-image/scikit-image/blob/39176efc2ad4e1131b57bb14f42bca335d43e399/skimage/morphology/extrema.py#L356-L360

Might be useful for the user and make the API more consistent. Using `np.nonzero` it's really trivial to add.

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,290 @@
...
+        warnings.warn('Non-contiguous array passed as `image`; this will be '
+                      'converted to a contiguous array as a copy.')
+        image = np.ascontiguousarray(image)
+        order = 'C'
+
+    seed_value = image[seed_point]
+
+    # Shortcut for rank zero
+    if image.size == 0:
+        return np.array([], dtype=np.uint8)
```
lagru: The docstring promises that this function returns "A Boolean array with the same shape as `image`".

```suggestion
        return np.zeros(image.shape, dtype=np.uint8)
```

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,290 @@
...
+    except TypeError:
+        if working_image.dtype == np.float16:
+            # Provide the user with clearer error message
+            raise TypeError("dtype of `image` is float16 which is not "
+                            "supported, try upcasting to float32")
+        else:
+            raise
+
+    # Output what the user requested; view does not create a new copy.
+    return flags[(slice(1, -1),) * image.ndim].view(np.bool)
```
lagru: Seeing this: is there a convention in scikit-image on how functions return indices? E.g. as a boolean array of the same shape, as an array of 0s and 1s or an `np.nonzero(...)` of the former (see earlier comment)?

skimage/morphology/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,143 @@
...
+cdef:
+    # Border value - do not cross!
+    unsigned char BORDER = 2
+    # Part of the flood fill
+    unsigned char FILL = 1
+    # Not checked yet
+    unsigned char UNKNOWN = 0
+
+
+cpdef inline void _flood_fill_equal(dtype_t[::1] image,
```
lagru: Is there a reason for using `cpdef` here? As long as this function isn't called somewhere in C- or Cython-code `cpdef` isn't needed and `inline void`, `nogil` are actually ineffective. So for readabilities sake simplifying the signature might give a clearer intention to the reader.

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -0,0 +1,128 @@
...
+from pytest import raises
+
+from skimage.morphology import flood_fill
+
+eps = 1e-12
+
+
+def test_empty_input():
+    output = flood_fill(np.empty(0), (), 2)
+    assert output.size == 0
```
lagru: This should check for the dtype (and shape if implemented above) as well.

skimage/morphology/_flood_fill.py
jni: @lagru we had a discussion about this that GitHub has probably helpfully masked. =P Also some of it may have happened offline while at SciPyConf — sorry if we did an incomplete job transcribing to here.

The basic idea behind not going this route is:

- we want to return the `bool` array because it is produced as a side-effect of the running of the function, so calling `np.nonzero` on it is extra overhead that users may not need.
- we generally want to avoid changing return types based on the value of some inputs. Even though it is Pythonic to do this sort of thing, it makes it harder (sometimes impossible) to statically check the type correctness of code.

Even though we've got some examples where we do this, the question is how we want to proceed going forward.

Finally, since this is a trivial change to add later without breaking the existing API, I suggest that this should not hold up this PR.

skimage/morphology/_flood_fill.py
jni: I think @lagru has uncovered a valid concern here. As far as I can tell, `inplace=True` won't work for non-contiguous arrays, because `image` will have been copied. However this is not necessary because `image[mask] = new_value` will work fine for noncontiguous, so even though the Cython code requires contiguity, we can still do the fill in-place.

Am I wrong here? And, if not, @JDWarner can you change this behaviour and add a test for it?

skimage/morphology/_flood_fill.py
jni: 3? Not 2? But anyway I think this is probably the right place for the warning...

skimage/morphology/_flood_fill.py
jni: @lagru shouldn't the dtype be `bool`? But your suggestion about the shape is spot-on — even if the user passes an array of shape (0, 0), we want to preserve that dimensionality. It might be useful for stacking things later, for example.

skimage/morphology/_flood_fill.py
jni: Not currently. ;) I think we should return indices differently depending on the amount of work required. Since `flood` produces a boolean array as a side effect, I think we should return that, and not do extra work to get `np.nonzero` when users might not require it. Similarly, if something produced indices (`np.nonzero`-style) as a side effect, I don't think we should instantiate an enormous bool array.

Having said that, a counter-point: we can imagine a future in which NumPy arrays are expandable in amortised-constant time (similar to Python lists), in which case, the indices approach becomes much more natural, and we might want to switch to that. Do we plan for said future by normalising our API to the most "efficient" output perspective? I would argue not (who knows if expandable arrays will ever be a thing), but it's worth considering.

At any rate, I'll go back to what I said above: this should not hold up this PR.

skimage/morphology/_flood_fill_cy.pyx
jni: @lagru 

Interesting! Do you have a reference as to nogil not working in cpdef'd functions?

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -0,0 +1,128 @@
...
+
+    np.testing.assert_array_equal(image, expected)
+
+
+def test_inplace_float():
+    image = np.array([[0, 0, 0, 0, 0, 0, 0],
+                      [0, 1, 1, 0, 2, 2, 0],
+                      [0, 1, 1, 0, 2, 2, 0],
+                      [1, 0, 0, 0, 0, 0, 3],
+                      [0, 1, 1, 1, 3, 3, 4]])
```
jni: Erm, this isn't float? =D

I guess we need `test_inplace_non_contiguous`, too.

skimage/morphology/_flood_fill_cy.pyx
lagru: @jni As I understand it [`cpdef` will create two functions, equivalent to `def` and `cdef`](https://notes-on-cython.readthedocs.io/en/latest/function_declarations.html#cpdef-it-s-both) (note: I'm not sure whether that is an official document). So the second pure C-function is never used which won't do any harm apart from increasing the binary ever so slightly.

Concerning `nogil`: what I mean is that I'm not sure whether the `nogil` statement actually works together with the `def` part of the declaration (I have never seen it used that way). To add to this it clashes with the `with nogil` statement in the function body. So one of those isn't needed. For some reason I can't find the documentation to that keyword right now but I would recommend to just keep it as simple as possible. And as long as we don't need this functionality and keywords why risk it?

skimage/morphology/_flood_fill.py
lagru: > I think we should return indices differently depending on the amount of work required. Since `flood` produces a boolean array as a side effect, I think we should return that, [...]

Well it doesn't satisfy the perfectionist inside me but it's probably the more pragmatic approach. I think it would be worth it though to decide between returning a boolean array or an array of 0s and 1s for functions that don't use indices. Creating a view is really cheap and always returning the same type would make the user experience more consistent. And by that I don't mean that we go around changing stuff but just that we have a guideline for future changes.

> At any rate, I'll go back to what I said above: this should not hold up this PR.

Fully agreed.

skimage/morphology/_flood_fill.py
lagru: @jni Yeah, good catch! I totally overlooked that this function returns a boolean array.

skimage/morphology/_flood_fill.py
jni: > decide between returning a boolean array or an array of 0s and 1s for functions that don't use indices.

Definitely bool. 0s and 1s can't be used to index a NumPy array. (At least not to get the expected result!)

skimage/morphology/_flood_fill.py
lagru: @jni Okay, good to know. Sounds sensible. Perhaps we should transcribe that somewhere? Maybe the developer guide?

skimage/morphology/_flood_fill.py
jni: @lagru You literally mentioned it in your comment! :joy:

lagru: Yeah, I got to that part of the code after suggesting this change. :sweat_smile:

skimage/morphology/_flood_fill.py
lagru: Should we change that in `local_maxima` as well then?

jni: Holy crap! Yes! LOL

jni: The API breakage will be fun... =\

skimage/morphology/_flood_fill.py
lagru: @jni Yeah, that's what I meant. (I modified my original snippet.)

skimage/morphology/_flood_fill.py
lagru: No, you are right! I got confused because I remembered https://github.com/scikit-image/scikit-image/issues/2097#issuecomment-428118113. But that one uses the wrapped `warn` function unlike this case. So it should be 2.

skimage/morphology/_flood_fill_cy.pyx
JDWarner: Yes, you can use `with nogil` in a Cython function defined with `def`.  Here, the double `nogil` is because `cpdef` generates both the Python wrapper and the pure C-function - the C-function uses the declaration `nogil`, while the Python wrapper makes use of the `with nogil`. 

I generally use `cpdef` when reasonable, even if `def` would suffice at the time, to expose future possibilities.  We don't use it here right now, in this PR, but for the price of a function declaration we ship the ability for someone to be much faster if they want to call this from pure C.  And it's fast enough that it might be useful for real-time work, so I'm trying to be as sensitive to that as possible.

skimage/morphology/_flood_fill.py
JDWarner: As @jni said, this is basically due to minimizing computation time required to arrive at a result which meets NumPy requirements for indexing.  I think NumPy tightened their requirements in the last year or so to a Boolean array from looser prior tolerances.  I think I'd prefer a nested Tuple as well, but that's not realistic given how this algorithm works.

skimage/morphology/_flood_fill.py
JDWarner: Given that it's possible to fill inplace even with a noncontiguous array, I'm going to remove all of these warnings.  The original won't be made contiguous and will be filled, so I don't see why we need them.

skimage/morphology/_flood_fill.py
JDWarner: Done.

skimage/morphology/_flood_fill_cy.pyx
lagru: > We don't use it here right now, in this PR, but for the price of a function declaration we ship the ability for someone to be much faster if they want to call this from pure C.

I would like this argument more if the function in question was part of the user-facing API. Right now I would raise an eyebrow if a user started to rely on this private function. 

I guess it comes down to different philosophies in this case.  While I'd still prefer a simpler approach until the we actually need to call this from Cython code, I can see no real disadvantages to yours. So to be clear, I'm fine to leave it as is. :)

jni: I have to side with @lagru here — although I very much like the idea of providing C-accessible functionality, if we go down that road we should do it with a public facing API, rather than semi-encourage users to use a private function that we might want to mess with in the future. In the meantime, we should aim to minimise our API surface.

I don't think this should hold up this PR, but wanted to use this chance to (maybe) establish a policy on this sort of thing.

skimage/morphology/_flood_fill.py
```diff
@@ -98,30 +99,20 @@ def flood_fill(image, seed_point, new_value, *, selem=None, connectivity=None,
...
            [5, 5, 5, 5, 5, 5, 3]])
     """
-    # Correct start point in ravelled image - only copy if non-contiguous
-    image = np.asarray(image)
-    copied = False
+    if inplace and not image.flags.contiguous:
+        # Flood fill requires some sort of contiguity - this makes a copy
+        _image = np.ascontiguousarray(image)
+    else:
+        _image = image
```
jni: Is this actually needed at all? `flood` makes its own contiguity check and copy if necessary.

lagru: No, I think you're right. Nice!

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -0,0 +1,158 @@
...
+
+def test_inplace_noncontiguous():
+    image = np.array([[0, 0, 0, 0, 0, 0, 0],
+                      [0, 1, 1, 0, 2, 2, 0],
+                      [0, 1, 1, 0, 2, 2, 0],
+                      [1, 0, 0, 0, 0, 0, 3],
+                      [0, 1, 1, 1, 3, 3, 4]])
+
+    # Transpose is noncontiguous
+    image = image.T
```
jni: Transpose is F-contiguous, which appears to be handled by the code?

skimage/morphology/_flood_fill_cy.pyx
JDWarner: To be clear, I intend to use this function in this way in planned future work - relevant to interactive GUI as well as my research.  I'd very much like the pure C function to exist.

jni: Alright I'll allow it. But only because I like you. =P

Speaking of which, I'm not sure you ever gave me your guacamole recipe. =P And in looking for it in my email I just found a neglected to-do item — deprecating our deprecation policy! =P

jni:
@lagru want to give this one more pass?

doc/examples/segmentation/plot_floodfill.py
```diff
@@ -0,0 +1,141 @@
...
+# value, allowing use on real-world images.
+#
+# Here we will experiment a bit on the cameraman.  First, turning his coat from
+# dark to light.
+
+cameraman = data.camera()
+
+# Change the cameraman's coat from dark to light (255).  The seed point is
+# chosen as (200, 100),
+light_coat = flood_fill(cameraman, (200, 100), 255, tolerance=10)
```
stefanv: Suggestion: plot the seed point here and elsewhere.

doc/examples/segmentation/plot_floodfill.py
```diff
@@ -0,0 +1,141 @@
...
+##############################################################################
+# Flood as mask
+# -------------
+#
+# A sister function, `flood`, is available which returns a mask identifying
+# the flood rather than modifying the image itself.  This is useful for
+# segmentation purposes and more advanced analysis pipelines.
+#
+# Here we segment the nose of a cat.  However, multi-channel images are not
+# supported by flood[_fill].  Instead we Sobel filter the red channel to
```
stefanv: Why the red channel?  Green is probably a more likely choice.  Although anything is better than blue :)

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,274 @@
...
+        New value to set the entire fill.  This must be chosen in agreement
+        with the dtype of `image`.
+    selem : ndarray, optional
+        A structuring element used to determine the neighborhood of each
+        evaluated pixel. It must contain only 1's and 0's, have the same number
+        of dimensions as `image`. If not given, all adjacent pixels are
+        considered as part of the neighborhood (fully connected).
+    connectivity : int, optional
+        A number used to determine the neighborhood of each evaluated pixel.
+        Adjacent pixels whose squared distance from the center is larger or
```
stefanv: This description doesn't seem right.

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,274 @@
...
+        equal to `connectivity` are considered neighbors. Ignored if
+        `selem` is not None.
+    tolerance : float or int, optional
+        If None (default), adjacent values must be strictly equal to the
```
stefanv: This is not quite accurate: By default, in order to be filled, adjacent values must be strictly equal to the value of `image` at `seed_point`.

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,274 @@
...
+        initial value of `image` at `seed_point`.  This is fastest.  If a value
+        is given, a comparison will be done at every point and this tolerance
```
stefanv: This sentence doesn't read correctly.

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,274 @@
...
+        on each side of the initial value will also be filled (inclusive).
+    inplace : bool, optional
+        If True, flood filling is applied to `image` inplace.  If False, the
+        flood filled result is returned without modifying the input `image`
+        (default).
+
+    Returns
+    -------
+    filled : ndarray
+        An array with the same shape as `image` is returned with values equal
```
stefanv: This is also not correct.

doc/examples/segmentation/plot_floodfill.py
JDWarner: I experimented - red worked best for this example.  Green just wasn't quite as good.

skimage/morphology/_flood_fill.py
JDWarner: This is literally identical to how we describe connectivity elsewhere, most notably https://github.com/scikit-image/scikit-image/blob/29a22e17c8b4ccf26383e3007ab014eb18b81f12/skimage/morphology/extrema.py#L351

stefanv: It may be correct/incorrect in other places as well; but can we check carefully that this formulation is accurate?  It may well be, but I wanted more eyes on it.

stefanv: The word `larger` here is particularly suspect.

```
In [8]: connectivity = 1; ndi.generate_binary_structure(2, connectivity)                                                                                                             
Out[8]: 
array([[False,  True, False],
       [ True,  True,  True],
       [False,  True, False]])
```

Note the description from `ndimage`:

```
connectivity : int
     `connectivity` determines which elements of the output array belong
     to the structure, i.e. are considered as neighbors of the central
     element. Elements up to a squared distance of `connectivity` from
     the center are considered neighbors. `connectivity` may range from 1
     (no diagonal elements are neighbors) to `rank` (all elements are
     neighbors).
```

lagru: Oh, I think I messed that up when writing the docstring for `local_maxima`. I even remember looking up that docstring as a source. :see_no_evil:  If we change "larger" into "smaller" it should be correct though.

JDWarner:
Also fixed the docstring in `local_maxima` pertaining to `connectivity`.

jni:
@stefanv this is ready for another pass from you!

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -0,0 +1,164 @@
...
+    expected = np.array([[5, 0, 5, 0, 5, 0, 5],
+                         [0, 1, 1, 0, 2, 2, 0],
+                         [5, 1, 1, 0, 2, 2, 5],
+                         [1, 0, 0, 0, 0, 0, 3],
+                         [5, 1, 1, 1, 3, 3, 4]])
+
+    np.testing.assert_allclose(image, expected)
+
+
+def test_1d():
```
lagru: Could you add at least one test for the >= 3D case?

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -0,0 +1,164 @@
...
+                         [-1., -1., -1., 100., 0., 0., 0.],
+                         [-1., -1., -1., 100., 0., 0., 0.],
+                         [-1., -1., -1., 100., 0., 0., 0.],
+                         [-1., -1., -1., 100., 0., 0., 0.]])
+
+    np.testing.assert_equal(flood_fill(test, (0, 0), -1), expected)
+
+
+def test_neighbors():
+    # This test will only pass if the neighbors are exactly correct
```
lagru: Is this statement true? My impression is that this only checks if neighbors are connected in the last dimension? 

On connectivity and the structuring element: I guess a large part of that functionality is covered by the tests for `watershed` and `local_maxima` but ideally this module would have at least some basic tests for the parameters `connectivity` and `selem`. You could just copy from the tests of `local_maxima`.

JDWarner: No, it fills a column and then fills a row.  Both directions are checked.  Given the way the ravelling works, if it is going to break it will break in 2D.

JDWarner: I'll admit I ascribe to a KISS style when it comes to testing in general... connectivity is checked in the docstring example and not replicated here, for example.

I'll add a basic test for `selem` but not directly testing `selem` exhaustively here, it's entirely exported to the mechanisms in `.extrema.py` which are tested separately was deliberate.

This is partially motivated by the length of our test suite, which has been an issue with CI builds in the past.

skimage/morphology/tests/test_flood_fill.py
JDWarner: When there isn't loss of generality I stay 2D, because it's easier for future maintenance (embedding 3D expected results in the test suite is rarely intuitive or easy to parse).

JDWarner:
Coverage bot is just broken, probably related to some of the rebasing along the way.  This PR doesn't touch the viewer but it claims huge losses in coverage there.

I've made relevant PEP8 tweaks.  I invoke PEP8 itself against the remaining *supposed* PEP8 failures (E226), because per PEP8 one is supposed to omit whitespace in more complex expressions this way.  The PEP8 checking algorithms are not smart enough to do it right.  I blacklist E226 in my personal editors because literally nothing does it right.

I do not know of a way to capture the example outputs in the CI.  Matplotlib has certain mechanisms to test that sort of thing, with PNG comparisons and the like, but we do not use them for the examples.

skimage/morphology/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,143 @@
...
+            flags[start_index] = FILL
+            # Break loop if all queued positions were evaluated
+            while queue_pop(&queue, &current_index):
+                # Look at all neighboring samples
+                for i in range(neighbor_offsets.shape[0]):
+                    neighbor = current_index + neighbor_offsets[i]
+
+                    # Shortcut if neighbor is already part of fill
+                    if flags[neighbor] == UNKNOWN:
+                        if image[neighbor] == seed_value:
```
hmaarrfk: can we combine these two conditions into 1 if. My head hurts because of the deep nesting.

skimage/morphology/_flood_fill_cy.pyx
```diff
@@ -0,0 +1,143 @@
...
+            flags[start_index] = FILL
+            # Break loop if all queued positions were evaluated
+            while queue_pop(&queue, &current_index):
+                # Look at all neighboring samples
+                for i in range(neighbor_offsets.shape[0]):
+                    neighbor = current_index + neighbor_offsets[i]
+
+                    # Only do comparisons on points not (yet) part of fill
+                    if flags[neighbor] == UNKNOWN:
+                        if low_tol <= image[neighbor] <= high_tol:
```
hmaarrfk: Is this syntax correct in cython? or is it doing ((low_tol <= image) <= high_tol)

jni: @hmaarrfk it's correct. Cython tries very hard to be a superset of Python.

hmaarrfk: ok. 

I'll just leave an other comment that I really don't like this syntax. 
I know that working with lowest common denominators is "bad" but seeing as this causes many bugs in other languages, I think it is a good syntax "trick" to avoid.

jni: @hmaarrfk I disagree, I think all other languages are broken in this regard and major advantages of Python should be celebrated. =)

Having said that, I don't *strongly* disagree because, tragically, NumPy arrays don't have this syntax (I guess because Python as far as I know doesn't provide a way to override this behaviour for classes), and *that* has bitten me a couple of times. Not bitten me hard enough that I introduced a bug, but hard enough that I had to rewrite something because I forgot about this not working.

hmaarrfk: Truthfully, my issue is with CPython. The fact that
```
for i in range(10):
    pass
```
can't be optimized is a travesty

jni: I don't think this is the right forum for a chat about the Python language's weaknesses? =)

Anyway, "can't" is such a strong word... CPython is FOSS, so perhaps this can be your contribution to the language? ;)

skimage/morphology/tests/test_flood_fill.py
lagru: I can't follow that line of reasoning yet. In this case the ability to generalize between 2D and ND is an implementation detail of this particular algorithm. I feel like tests shouldn't loose or gain value if the internal implementation changes without a change to the API. If we replaced the algorithm in the future with one were this generalization doesn't hold true we'd have a gap in our test suite were the 3D case is not properly covered.

However I agree that tests for >2D are generally harder to read. Therefore how about a compromise that adds one basic test that can act as a canary for if that generalization is lost?

E.g.:

```python
cube = np.zeros((5, 5, 5))
cube[1:-1, 1:-1, 1:-1] = 1  # sum is 27
filled = flood_fill(cube, (2, 2, 2), 2)
assert filled.sum() == 54
assert_equal(filled[1:-1, 1:-1, 1:-1], np.ones((3, 3, 3)) * 2)
```

I guess I'd prioritize functionality over readability in tests but I recognize that as an opinion which you don't have to share. Please, feel free to mark this as resolved if you don't see the merit in my suggestion. :wink:

skimage/morphology/tests/test_flood_fill.py
lagru: I guess I just find that comment overly broad with potential for confusion. E.g. the test doesn't check for diagonal connections. I'd actually prefer something explicit like the examples from the docstring. Speaking of which, I testing docstrings is a nice feature to ensure that they are correct but I don't agree with the sentiment that they may replace tests. If someone changes them in the future they won't consider test coverage while doing so. I'd prefer the test coverage of a function to be independent of its docstring.

But same as above, feel free to resolve this. This is not worth blocking the PR for.

lagru:
> I do not know of a way to capture the example outputs in the CI. Matplotlib has certain mechanisms to test that sort of thing, with PNG comparisons and the like, but we do not use them for the examples.

@JDWarner Not sure if you misunderstood me. My question was more along the line of if it's possible to view the examples as they would appear in on the website without having to clone the PR and build the documentation locally.

jni:
I'd like to formally move to merge this and leave any future improvements to future PRs. The tests cover 100% of the new code — they may not be perfect but full coverage is good enough for me, especially when the PR has been in the works for more than six months. Things like minor syntax fixes, comments cleanup, and better tests can all occur in future PRs.

@stefanv the GitHub UI still records you as a dissenting voice, though that may have changed with @JDWarner's latest commits. =)

soupault:
>My question was more along the line of if it's possible to view the examples as they would appear in on the website without having to clone the PR and build the documentation locally.

@lagru Not yet, unfortunately. The relevant issue is here - https://github.com/scikit-image/scikit-image/issues/2269 .

skimage/morphology/_flood_fill.py
```diff
@@ -0,0 +1,275 @@
...
+        If None (default), adjacent values must be strictly equal to the
+        initial value of `image` at `seed_point`.  This is fastest.  If a value
+        is given, a comparison will be done at every point and if within
+        tolerance of the initial value will also be filled (inclusive).
+
+    Returns
+    -------
+    mask : ndarray
+        A Boolean array with the same shape as `image` is returned, with values
+        equal to 1 for areas connected to and equal (or within tolerance of)
```
soupault: `1` -> True, `zero` -> False.

skimage/morphology/extrema.py
```diff
@@ -350,7 +350,7 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
         considered as part of the neighborhood.
     connectivity : int, optional
         A number used to determine the neighborhood of each evaluated pixel.
-        Adjacent pixels whose squared distance from the center is larger or
+        Adjacent pixels whose squared distance from the center is less than or
```
soupault: Should be fixed also in `local_minima`.

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -0,0 +1,197 @@
...
+        flood_fill(image, 0, 1)
+
+
+def test_overrange_tolerance_int():
+    image = np.arange(256, dtype=np.uint8).reshape((8, 8, 4))
+    expected = np.zeros_like(image)
+
+    output = flood_fill(image, (7, 7, 3), 0, tolerance=379)
+
+    np.testing.assert_equal(output, expected)
```
soupault: `np.testing` should not be used directly. We have all the necessary wrappers in https://github.com/scikit-image/scikit-image/blob/master/skimage/_shared/testing.py .

jni: I thought that we'd reversed this policy? I remember the comment that having our own testing functions is one more thing to remember "just for skimage", and it resonated very strongly with me. We should use standard things except where it is an undue burden. I realise that the nose -> pytest transition was traumatic, but imho that is not sufficient justification to do our own thing, as we don't expect those transitions to be frequent.

jni: (Apologies if the above comment comes like 18 months too late or so. =P)

soupault: Why will anyone want to revert that? 😲 Frankly speaking, I'm hearing about this for the first time.

To me, the thin wrappers in `skimage._shared.testing` aren't just for potential convenient switch of the testing library. Besides that:

1. It is a single place where all the utilised testing functions and classes are presented. Even after our transition to `pytest` we kept using a lot of functionality from `numpy.testing`. This is due to the fact that neither of those libraries include all the necessary functions and one is not superset of the other. On top of this, `numpy.testing` was still using both `nose` and `pytest` under the hood, and for a skimage contributor it was a pain to understand which functions he or she can import from numpy. Having this thin wrappers, a contributor can just import `skimage._shared.testing` and enjoy the coding; 
2. Single point of deprecation. When `pytest` or `numpy.testing` will have some functionality deprecated, for us it is much easier to rewrite couple of lines in `skimage._shared.testing` than to change few hundreds of them all over the library;
3. Almost-0 maintenance cost.

If you have a strong argument why we should drop that, I'm happy to discuss, but, otherwise, I'd strongly prefer to keep things as is.

jni: @soupault I think I just got confused by the discussion [here](https://github.com/scikit-image/scikit-image/pull/3164), specifically [Mark's comment](
https://github.com/scikit-image/scikit-image/pull/3164#issuecomment-395649448). But, my overall feeling remains, so I'll address your points:

1. I get this, I just don't think that scikit-image should be that one place. `np.testing` is a much more logical place for that one place, especially since `_shared` is private, which means we can change things when we feel like it, which means that no outside project can use our wrappers.
2. In contrast, `np.testing` is a public API that is much less likely to be deprecated.
3. I would argue that postponing this PR (and countless others because everyone in the SciPy ecosystem is trained to use `np.testing` and `with pytest.raises`) is a maintenance cost.

Re (3), note that we don't have any reference to `skimage._shared.testing` in the contributing guide.

JDWarner: Agree with @jni.

I think I was deeply mired in surgical intern year when the conversion to pytest happened.  These shared wrappers were unknown to me.  Looking around the package, our usage remains heterogeneous.  I would have been :-1: at converting anything to using them, unless they bring new important functionality, and regret not being part of that discussion.

I strongly dislike forcing use of a hidden test API when NumPy's is perfectly suitable.  Where certain features may not exist at base in NumPy, we should publicly expose these (even if they are deeply nested, like `skimage.util.testing` or similar.  The potential portability advantages are not significant enough to justify this.  This is the same reason we generally don't reimplement `scipy.ndimage`.

Imagine a new user making a first contribution.  They generate a perfectly compliant test suite using guides they find online, via NumPy's suite.  We tell them - despite technical correctness - that we won't accept it?  Yikes.  All this really does is provide additional barriers to contribution.

Barriers to contribution are high enough already, IMHO, and we should be looking to reduce them.

stefanv: Background here: we switched before NumPy, so we had to take care of situations where they used nose specific functionality.  The idea behind the wrappers were that there would be a single place to modify things, instead of using pytest-specific decorators, etc.

But, yes, I don't see any reason why we need to maintain our own testing utilities, if the functionality overlaps almost identically with numpy.

stefanv:
On Sun, 24 Feb 2019 17:58:12 -0800, Josh Warner wrote:
> Imagine a new user making a first contribution.  They generate a
> perfectly compliant test suite using guides they find online, via
> NumPy's suite.  We tell them - despite technical correctness - that we
> won't accept it?  Yikes.  All this really does is provide additional
> barriers to contribution.

I don't think we ever do this.  We just go and make those stylistic
fixes on the PR, as described in the core developer guide.

The question to be discussed is whether we want to have something
pytest-specific in our test suite; and I'd argue you don't really want
to see the underlying technology show through; we've already made one
testing framework change.

So, having looked at our testing module a bit more carefully, I'll
soften my stance from the previous comment, and say that I think it
makes sense what we have done there.  At the same time, I think *not*
allowing functions from `np.testing` to be used is overkill: the
contract of these functions is simple enough: check a condition and
raise if not.  No matter what testing framework we choose to use, they
will have to support that.

jni:
> I don't think we ever do this. We just go and make those stylistic fixes on the PR, as described in the core developer guide.

Well, compliance with the above is haphazard, at best. We definitely need to do better here.

> The question to be discussed is whether we want to have something pytest-specific in our test suite; and I'd argue you don't really want to see the underlying technology show through; we've already made one testing framework change.

In this case, again, we have tons of places where we use pytest fixtures, pytest xfails, and pytest skipifs (try `git grep "import pytest"`). This is the correct approach imho. Test framework migrations are painful *but rare*, and I disagree with @soupault that the overhead is minimal. For one, the wrappers are not documented in-place, so we still have to go check the pytest or numpy documentation. It's a leaky abstraction.

> I think *not* allowing functions from `np.testing` to be used is overkill

Additionally, swapping `np.testing.assert_*` with an alternative throughout our codebase is quite easy these days with editors like PyCharm.

jni:
@JDWarner I was just about to prod you to fix the merge conflicts. =D Thanks!

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -0,0 +1,213 @@
...
+    expected = np.array([[  0,   0,   0,   0,   0,   0],
+                         [255, 255, 255, 255, 255,   0],
+                         [255, 255, 255, 255, 255,   0],
+                         [255, 255, 255, 255, 255,   0],
+                         [255, 255, 255, 255, 255,   0]], dtype=np.uint8)
+
+    np.testing.assert_equal(output, expected)
+
+
+def test_basic_nd():
```
JDWarner: Programmatic tests for 3d through 5d, just for @lagru ;)

stefanv:
On Sun, 24 Feb 2019 19:12:08 -0800, Juan Nunez-Iglesias wrote:
> > I don't think we ever do this. We just go and make those stylistic fixes on the PR, as described in the core developer guide.
> 
> Well, compliance with the above is haphazard, at best. We definitely
> need to do better here.

Yes, we should keep reminding one another of what we agreed to as a
team, and we should revise those guidelines when we realize a pattern
described there is not a good fit.

> In this case, again, we have tons of places where we use pytest
> fixtures, pytest xfails, and pytest skipifs (try `git grep "import
> pytest"`). This is the correct approach imho. Test framework
> migrations are painful *but rare*, and I disagree with @soupault that
> the overhead is minimal. For one, the wrappers are not documented
> in-place, so we still have to go check the pytest or numpy
> documentation. It's a leaky abstraction.

I think many contributors look at the existing test suite to guide them.
So, as long as we make the entire codebase consistent, it should be
fine.  I don't feel particularly strongly about the issue.  I think you
can see why the decision was made to abstract these into a module
(during the testing migration period, when everything was in flux all
over the place), and you can see why people would prefer not to have it
any longer.

Let's see how the change looks in a PR, then we can discuss it further
with something tangible in front of us.

jni:
This now has three approvals and three comments, and the single Travis failure is related to #3767, so IMMA GONNA MERGE! :grinning: Thanks @JDWarner, a headline feature for 0.15!

stefanv:
Thank you, @JDWarner, great piece of work; this is a long-awaited addition!

soupault:
Thanks @JDWarner for the very important addition! As usually, well executed.

PR #3247, opened 2018-07-05
WIP: tidy up register_translation a tiny bit

hmaarrfk:
I've been trying to accelerate register_translation for a personal project using `cupy`. Cupy complained a little bit about a few things that were being done.

These changes are mostly benign and make the code look good for sure and while addressing some of the issues I ran into.

There were a few other problems in certain spots, but those changes would actually make the code look worse.

## For reviewers

  later.

### DIFF
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -184,7 +184,7 @@ def register_translation(src_image, target_image, upsample_factor=1,
     # Locate maximum
     maxima = np.unravel_index(np.argmax(np.abs(cross_correlation)),
                               cross_correlation.shape)
-    midpoints = np.array([np.fix(axis_size / 2) for axis_size in shape])
+    midpoints = np.asarray(shape, dtype=np.floating) // 2
 
     shifts = np.array(maxima, dtype=np.float64)
     shifts[shifts > midpoints] -= np.array(shape)[shifts > midpoints]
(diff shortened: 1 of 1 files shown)


skimage/feature/register_translation.py
```diff
@@ -196,7 +196,6 @@ def register_translation(src_image, target_image, upsample_factor=1,
     # If upsampling > 1, then refine estimate with matrix multiply DFT
     else:
         # Initial shift estimate in upsampled grid
-        shifts = np.round(shifts * upsample_factor) / upsample_factor
```
hmaarrfk: Shifts is computed from above from the following code:
```python
shifts = np.array(maxima, dtype=np.float64)	     shifts = np.array(maxima, dtype=np.float64)
shifts[shifts > midpoints] -= np.array(shape)[shifts > midpoints]
```
where maxima are guaranteed to be integers converted to floats. This line doesn't do anything in my mind.

hmaarrfk:
There are also a few other things, but I didn't think they were were slam dunks

for example:
https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/register_translation.py#L169-L173
are in my mind implementation specific optimizations.

[cupy can compute FFTs in 32 bit precision](https://docs-cupy.chainer.org/en/stable/reference/fft.html)
and 

so will pyfftw
```
import pyfftw
pyfftw.interfaces.numpy_fft.fft2(np.zeros((4, 4), dtype=np.float32)).dtype
OUT: dtype('complex64')
```

Personally, I would just take them off.

skimage/feature/register_translation.py
hmaarrfk: hmm, interesting. The function definitely returns something. I'll have to look into it further.

hmaarrfk:
Thanks @emmanuelle for finding the edge case I didn't think of. I think the other fixups were incorporated  at some point.

Issue #2677, opened 2017-06-10
exposure.equalize_hist doesn't output the same type as the input


TheShadow29:
I am not completely sure if it should be a feature request or a bug.
Currently exposure.equalize_hist given an image of input of type uint8 outputs an image of type float64. The proposal is to give the output in uint8 format itself. 
Current implementation:
```
img1 = io.imread(path_to_file)
img11 = skimage.exposure.equalize_hist(img1)
print('Original Image is of dtype ', img1.dtype)
print('Histogram Equalized Image is of dtype', img11.dtype)
```
Output:
```
('Original Image is of dtype ', dtype('uint8'))
('Histogram Equalized Image is of dtype', dtype('float64'))
```
#1192 mentions this, but the issue is not really addressed.
To be honest I am not even sure why the output image is of type 'float64'. At no point of time should  image of type uint8 should be changed to float64, i.e. there should be no case where we should be working with floats, we should be in integer domain itself. Here is a nice text describing histogram equalization https://www.math.uci.edu/icamp/courses/math77c/demos/hist_eq.pdf

Histogram equalization would simply be a transformation curve, mapping one intensity to another corresponding intensity (the formulae given in the mentioned link) and it should be all in the integer domain when the input is of type uint8. 

Also as far as I know, MATLAB also does the same thing, that is given an input of uint8 returns output in uint8 itself. Link for the same https://www.mathworks.com/help/images/ref/histeq.html

In short, the current implementation of the equalize_hist is doing something similar but will give a different result from what a normal histogram equalization would return at least for an uint8 image. Also I think it is more sensible to return image of the same dtype.

TheShadow29:
@soupault can you have a look at this too?

jni:
Hi @TheShadow29 

I think the difference is much smaller than you think. One of the perils of working with uint8 instead of float! `abs(img12 - img2)` will result in negative values, anywhere `img12` is smaller than `img2`, looping back to 255. Try `diff = abs(img12.astype(float) - img2.astype(float)).astype('uint8')`, then compute the mean and max of that image.

I don't know that it's generally true that "it should be all in the integer domain when the input is of type uint8". The choice of `floor` in the paper you linked is a consequence of the desire to work in the integer domain. If you work in float, you don't need to use the floor at all, but rather you can use the exact value from the mapping, and that will be a more faithful histeq result.

Could you post links to your rice image and the Matlab-equalised image? I can investigate a bit more if I have some sample test data.

[...]

PR #3253, opened 2018-07-07
Enable warping with more than just doubles.

### LINKED ISSUE #1287: Use fused types for warping
Use fused types for image warping, so cast to double is not necessary.

hmaarrfk:
Fixes #1287 

~~With complex types too!!! That was more difficult than it sounds....~~ too difficult to do without help from the compiler.

~~This breaks things because people might have expected an implicit cast to float. Needs somebody to make a decision on how to deprecate this.~~ 

I think that calling anything like: `img_as_*` ~~, or any kind of range checking should be considered a mistake~~ without converting the data back to the original type, should be considered a mistake.
Personally, I think that the parameter should be ripped out. Instead, people should have to think about what type they want their data to be represented as.
Edit: I just think that types and units are different. The `img_as*` functions assume that the units you are I think that the `img_as_*` functions within algorithms is a mistake as it assumes a very specific unit system namely that uint8 is in ADC counts, and that floats are normalized ADC counts. That might be true for most cameras, but at the end of the day, it is just one unit system, and isn't universaly true, especially for other data that might have ND correlations.

~~I probably need some help writing tests for this, but for now it passes, so that is a good thing right??? maybe there is an implicit cast as float somewhere.~~

Edit: clarification on what I think should be considered a mistake.
## Checklist
  existing benchmark
- [ ] Unit tests (started, extra tests may depend reviewer decision)

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/fused_numerics.pxd
+++ b/skimage/_shared/fused_numerics.pxd
@@ -1,0 +1,34 @@
+cimport numpy as cnp
+import numpy as np
+
+ctypedef fused np_ints:
+    cnp.int8_t
+    cnp.int16_t
+    cnp.int32_t
+    cnp.int64_t
+
+ctypedef fused np_uints:
+    cnp.uint8_t
+    cnp.uint16_t
+    cnp.uint32_t
+    cnp.uint64_t
+
+ctypedef fused np_anyint:
+    np_uints
+    np_ints
+
+ctypedef fused np_floats:
+    cnp.float32_t
+    cnp.float64_t
+
+ctypedef fused np_complexes:
+    cnp.complex64_t
+    cnp.complex128_t
+
+ctypedef fused np_real_numeric:
+    np_anyint
+    np_floats
+
+ctypedef fused np_numeric:
+    np_real_numeric
+    np_complexes
--- a/skimage/_shared/interpolation.pxd
+++ b/skimage/_shared/interpolation.pxd
@@ -18,8 +18,12 @@ by 4 values on each side:
 """
 from libc.math cimport ceil, floor
 
+import numpy as np
+cimport numpy as np
+from .fused_numerics cimport np_real_numeric, np_floats
 
-cdef inline Py_ssize_t round(double r) nogil:
+
+cdef inline Py_ssize_t round(np_floats r) nogil:
     return <Py_ssize_t>((r + 0.5) if (r > 0.0) else (r - 0.5))
 
 cdef inline Py_ssize_t fmax(Py_ssize_t one, Py_ssize_t two) nogil:
@@ -28,62 +32,68 @@ cdef inline Py_ssize_t fmax(Py_ssize_t one, Py_ssize_t two) nogil:
 cdef inline Py_ssize_t fmin(Py_ssize_t one, Py_ssize_t two) nogil:
     return one if one < two else two
 
+# Redefine np_real_numeric to force cross type compilation
+# this allows the output type to be different than the input dtype
+# https://cython.readthedocs.io/en/latest/src/userguide/fusedtypes.html#fused-types-and-arrays
+ctypedef fused np_real_numeric_out:
+    np_real_numeric
 
-cdef inline double nearest_neighbour_interpolation(double* image,
-                                                   Py_ssize_t rows,
-                                                   Py_ssize_t cols, double r,
-                                                   double c, char mode,
-                                                   double cval) nogil:
+cdef inline void nearest_neighbour_interpolation(
+        np_real_numeric* image, Py_ssize_t rows, Py_ssize_t cols,
+        np_floats r, np_floats c, char mode, np_real_numeric cval,
+        np_real_numeric_out* out) nogil:
     """Nearest neighbour interpolation at a given position in the image.
 
     Parameters
     ----------
-    image : double array
+    image : numeric array
         Input image.
     rows, cols : int
         Shape of image.
-    r, c : double
+    r, c : np_float
         Position at which to interpolate.
     mode : {'C', 'W', 'S', 'E', 'R'}
         Wrapping mode. Constant, Wrap, Symmetric, Edge or Reflect.
-    cval : double
+    cval : numeric
         Constant value to use for constant mode.
 
     Returns
     -------
-    value : double
+    value : np_float
         Interpolated value.
 
     """
 
-    return get_pixel2d(image, rows, cols, round(r), round(c), mode, cval)
+    out[0] = <np_real_numeric_out>get_pixel2d(
+        image, rows, cols, round(r), round(c), mode, cval)
 
 
-cdef inline double bilinear_interpolation(double* image, Py_ssize_t rows,
-                                          Py_ssize_t cols, double r, double c,
-                                          char mode, double cval) nogil:
+cdef inline void bilinear_interpolation(
+        np_real_numeric* image, Py_ssize_t rows, Py_ssize_t cols,
+        np_floats r, np_floats c, char mode, np_real_numeric cval,
+        np_real_numeric_out out[0]) nogil:
     """Bilinear interpolation at a given position in the image.
 
     Parameters
     ----------
-    image : double array
+    image : numeric array
         Input image.
     rows, cols : int
         Shape of image.
-    r, c : double
+    r, c : np_float
         Position at which to interpolate.
     mode : {'C', 'W', 'S', 'E', 'R'}
         Wrapping mode. Constant, Wrap, Symmetric, Edge or Reflect.
-    cval : double
+    cval : numeric
         Constant value to use for constant mode.
 
     Returns
     -------
-    value : double
+    value : numeric
         Interpolated value.
 
     """
-    cdef double dr, dc
+    cdef np_floats dr, dc
     cdef long minr, minc, maxr, maxc
 
     minr = <long>floor(r)
@@ -92,30 +102,36 @@ cdef inline double bilinear_interpolation(double* image, Py_ssize_t rows,
     maxc = <long>ceil(c)
     dr = r - minr
     dc = c - minc
-    top = (1 - dc) * get_pixel2d(image, rows, cols, minr, minc, mode, cval) \
-          + dc * get_pixel2d(image, rows, cols, minr, maxc, mode, cval)
-    bottom = (1 - dc) * get_pixel2d(image, rows, cols, maxr, minc, mode,
-                                    cval) \
-             + dc * get_pixel2d(image, rows, cols, maxr, maxc, mode, cval)
-    return (1 - dr) * top + dr * bottom
 
+    cdef np.float64_t top
+    cdef np.float64_t bottom
 
-cdef inline double quadratic_interpolation(double x, double[3] f) nogil:
+    cdef np_real_numeric top_left = get_pixel2d(image, rows, cols, minr, minc, mode, cval)
+    cdef np_real_numeric top_right = get_pixel2d(image, rows, cols, minr, maxc, mode, cval)
+    cdef np_real_numeric bottom_left = get_pixel2d(image, rows, cols, maxr, minc, mode, cval)
+    cdef np_real_numeric bottom_right = get_pixel2d(image, rows, cols, maxr, maxc, mode, cval)
+
+    top = (1 - dc) * top_left + dc * top_right
+    bottom = (1 - dc) * bottom_left + dc * bottom_right
+    out[0] = <np_real_numeric_out> ((1 - dr) * top + dr * bottom)
+
+cdef inline np_floats quadratic_interpolation(np_floats x,
+                                              np_real_numeric[3] f) nogil:
     """WARNING: Do not use, not implemented correctly.
 
     Quadratic interpolation.
 
     Parameters
     ----------
-    x : double
+    x : np_float
         Position in the interval [0, 2].
-    f : double[3]
+    f : real numeric[3]
         Function values at positions [0, 2].
 
     Returns
     -------
-    value : double
-        Interpolated value.
+    value : np_float
+        Interpolated value to be used in biquadratic_interpolation.
 
     """
     return (x * f[2] * (x - 1)) / 2 - \
@@ -123,30 +139,30 @@ cdef inline double quadratic_interpolation(double x, double[3] f) nogil:
                     (f[0] * (x - 1) * (x - 2)) / 2
 
 
-cdef inline double biquadratic_interpolation(double* image, Py_ssize_t rows,
-                                             Py_ssize_t cols, double r,
-                                             double c, char mode,
-                                             double cval) nogil:
+cdef inline void biquadratic_interpolation(
+        np_real_numeric* image, Py_ssize_t rows, Py_ssize_t cols,
+        np_floats r, np_floats c, char mode, np_real_numeric cval,
+        np_real_numeric_out* out) nogil:
     """WARNING: Do not use, not implemented correctly.
 
     Biquadratic interpolation at a given position in the image.
 
     Parameters
     ----------
-    image : double array
+    image : numeric array
         Input image.
     rows, cols : int
         Shape of image.
-    r, c : double
+    r, c : np_float
         Position at which to interpolate.
     mode : {'C', 'W', 'S', 'E', 'R'}
         Wrapping mode. Constant, Wrap, Symmetric, Edge or Reflect.
-    cval : double
+    cval : numeric
         Constant value to use for constant mode.
 
     Returns
     -------
-    value : double
+    out : np_real_numeric
         Interpolated value.
 
     """
@@ -154,11 +170,11 @@ cdef inline double biquadratic_interpolation(double* image, Py_ssize_t rows,
     cdef long r0 = <long>round(r) - 1
     cdef long c0 = <long>round(c) - 1
 
-    cdef double xr = r - r0
-    cdef double xc = c - c0
+    cdef np_floats xr = r - r0
+    cdef np_floats xc = c - c0
 
-    cdef double fc[3]
-    cdef double fr[3]
+    cdef np_real_numeric fc[3]
+    cdef np_floats fr[3]
 
     cdef long pr, pc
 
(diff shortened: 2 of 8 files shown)


jni:
@hmaarrfk 

Cool stuff! 

> I think that calling anything like: img_as_* , ~~or any kind of range checking should be considered a mistake~~ without converting the data back to the original type, should be considered a mistake.

No, repeated uint8 -> float -> uint8 -> float... conversions result in artifacts. See one demonstration [here](https://github.com/scikit-image/scikit-image/issues/2677#issuecomment-309717979). As you know I am not a fan of our type and range conversions, but the conversion to float is there for a good reason and converting float -> uint8 should only happen at the very end of a pipeline — and we have no way of knowing when that is. So the user has to be responsible here, one way or another.

We also have to think about how to combine this with #3148...

Will have a chat with @stefanv about this, as well as the deprecation path.

> for now it passes, so that is a good thing right??? 

Yes, very good! =)

hmaarrfk:
I agree that working with low precision units is a bad idea as it introduces artifacts. That is something you learn eventually when doing data processing and data acquisition. To address this problem, matlab and numpy both default to foat64 which is likely overkill in many situations, but definitely safe for non-experts.

That said, having the ability to develop with high precision units, and speed up computation with low precision ones is invaluable once a processing pipeline is ready to be optimized. The example you gave is great if people are trying to differentiate between data that is close to the center of the distribution, but if the data is well separated, it might not matter and the "error" might be acceptable. The user should have that choice.

I think git will help figure out what happens between #3148.

My 2 cents on data conversion:
  * If an algorithm only supports one input type, for example, `rank` algorithms don't support floats, or 32bit datatypes, then it should be up to the user to explicitly convert his data to the appropriate type and in the correct range. 16 bit datatypes are also ambiguous as to whether the upper 12 or lower 12 bits should be used.
  * Forcing an image to be between `0.` and `1.` is a really narrow view of what image data is. What if we are analyzing airflows? or temperature data? moisture level in the soil? topographic data? Do we force the users to work with normalized units? The calling `img_as_*` as opposed to throwing an error is just so domain specific in my mind. This is why I spent some time trying to make warping work with complex numbers.

I'm glad to have

hmaarrfk:
while thinking about it. I think it is best for the user to be able to
provide an output dtype if we want to allow flexibility.

That would allow us to chain algorithms together if we need to, while
leaving the user’s units intact.

I also like the idea of providing an output array, but that is a different
story. Maybe there is place for both.

hmaarrfk:
@jni: I just improved the the PR taking into consideration your feedback. Honestly, its up to you and the BDFL team with regards to how you want to deal with types, I just wanted to add flexibility to the algorithms which I think this PR achieves without imposing my point of view on the codebase.

1. warping now allows you to specify the dtype of the output. This enables higher level algorithms to choose the precision of this "intermediate step".
2. The default is still to upvert everything to float64 and not to preserve the range.
3. The intermediate computation type is determined by the homography matrix. This is a weird hack. I wish I could just let the compiler decide what the appropriate intermediate type is. For example: uint8 computation can probably benefit from potential speedups from single point arithmetic. Though maybe that is "premature optimization". Specifically, this is commit: https://github.com/scikit-image/scikit-image/pull/3253/commits/8262d7379abf6e99e43fbc3c82b1c6d6ad5b9abd I'm kinda impartial to this so if you want, I can take it off.
4. I backed away from complex interpolation. To avoid code duplication, it really needs to be done with C/C++ macros so that the compiler can choose the appropriate types. I don't know how to do in Cython. Apparently you can use Jinja or something, but maybe just writing straight C code is easier. I also don't know how to combine this with point #1 as you cannot have complex->float casting.

hmaarrfk:
plot twist, this actuallly speeds up float64->float64 somehow, but slows down uint8->uint8 significantly. fun times!

skimage/feature/_texture.pyx
```diff
@@ -92,7 +85,7 @@ cdef inline int _bit_rotate_right(int value, int length) nogil:
     return (value >> 1) | ((value & 1) << (length - 1))
 
 
-def _local_binary_pattern(double[:, ::1] image,
+def _local_binary_pattern(cnp.float64_t[:, ::1] image,
```
jni: Why change this?

skimage/feature/_texture.pyx
```diff
@@ -148,14 +141,15 @@ def _local_binary_pattern(double[:, ::1] image,
...
+    cdef cnp.float64_t zero = 0
 
     with nogil:
         for r in range(image.shape[0]):
             for c in range(image.shape[1]):
                 for i in range(P):
-                    texture[i] = bilinear_interpolation(&image[0, 0], rows, cols,
-                                                        r + rp[i], c + cp[i],
-                                                        'C', 0)
+                    bilinear_interpolation[cnp.float64_t, double, double](
```
jni: Whoa, what's going on here? Do you have a reference for this syntax?

skimage/transform/_warps.py
```diff
@@ -657,8 +657,38 @@ def _clip_warp_output(input_image, output_image, order, mode, cval, clip):
...
+
+    This is a Python wrapper for the Cython function ``_warp_fast``.
+    It prepares the data and allocates the necessary memory.
+
+    Parameters
+    ----------
+    output_shape : tuple (rows, cols), optional
+        Shape of the output image generated (default None).
+    dtype :
+        The numric type of the output image.
```
jni: This docstring is pretty woefully inaccurate...

skimage/_shared/fused_numerics.pxd
```diff
@@ -0,0 +1,34 @@
...
+    cnp.complex64_t
+    cnp.complex128_t
+
+ctypedef fused np_real_numeric:
+    np_anyint
+    np_floats
+
+ctypedef fused np_numeric:
+    np_real_numeric
+    np_complexes
```
jni: I love this file, and have wanted it for a long time. Thank you!!!

skimage/_shared/interpolation.pxd
```diff
@@ -28,62 +32,65 @@ cdef inline Py_ssize_t fmax(Py_ssize_t one, Py_ssize_t two) nogil:
...
 
-cdef inline double nearest_neighbour_interpolation(double* image,
-                                                   Py_ssize_t rows,
-                                                   Py_ssize_t cols, double r,
-                                                   double c, char mode,
-                                                   double cval) nogil:
+cdef inline void nearest_neighbour_interpolation(
+    np_real_numeric* image, Py_ssize_t rows, Py_ssize_t cols,
+    np_floats r, np_floats c, char mode, np_real_numeric cval,
+    np_real_numeric_out* out) nogil:
```
jni: Can you indent this 8 spaces to differentiate from a code block?

skimage/_shared/interpolation.pxd
```diff
@@ -92,73 +99,79 @@ cdef inline double bilinear_interpolation(double* image, Py_ssize_t rows,
...
+
+    cdef np_real_numeric top_left = get_pixel2d(image, rows, cols, minr, minc, mode, cval)
+    cdef np_real_numeric top_right = get_pixel2d(image, rows, cols, minr, maxc, mode, cval)
+    cdef np_real_numeric bottom_left = get_pixel2d(image, rows, cols, maxr, minc, mode, cval)
+    cdef np_real_numeric bottom_right = get_pixel2d(image, rows, cols, maxr, maxc, mode, cval)
 
-cdef inline double quadratic_interpolation(double x, double[3] f) nogil:
+    top = (1 - dc) * top_left + dc * top_right
+    bottom = (1 - dc) * bottom_left + dc * bottom_right
+    out[0] = <np_real_numeric_out> ((1 - dr) * top + dr * bottom)
```
jni: Looking at this code now, it might not be as hard as I thought to generalise to nD... =) (Not related to this PR, I just meant to say that your code has made me wonder about this long-standing goal.

skimage/feature/_texture.pyx
hmaarrfk: i don't know. gardening. Raymond Hettinger taught me not to do it. I'll revert.

skimage/feature/_texture.pyx
hmaarrfk: https://cython.readthedocs.io/en/latest/src/userguide/fusedtypes.html#indexing

fused type indexing. I didn't want to make changes to this function.

skimage/feature/_texture.pyx
```diff
@@ -148,14 +141,15 @@ def _local_binary_pattern(double[:, ::1] image,
 
     # To compute the variance features
     cdef double sum_, var_, texture_i
+    cdef cnp.float64_t zero = 0
```
hmaarrfk: Why do i have this? I'm going to try and take it off.

skimage/transform/_warps.py
```diff
@@ -657,8 +657,38 @@ def _clip_warp_output(input_image, output_image, order, mode, cval, clip):
...
+    others:
+        See ``_warp_fast``
+    """
+    image = np.ascontiguousarray(image)
+    H = np.ascontiguousarray(H, dtype=np.float64)
+
+    if output_shape is None:
+        out = np.empty_like(image, dtype=dtype)
+    else:
+        out = np.empty(shape=output_shape[:2], dtype=dtype)
```
hmaarrfk: I'm not sure what happens if you pass a 3D image and specify output shape.

skimage/_shared/fused_numerics.pxd
hmaarrfk: This should really be part of cython....

skimage/transform/_warps.py
hmaarrfk: What needs to be added?

skimage/_shared/fused_numerics.pxd
hmaarrfk: Also, if you want this, you should take it out of this PR since I'm not sure where the performance drop is coming from here. Cython casting?

hmaarrfk:
@jni, I'm glad you like this, but the performance is about 30% on unint8, and 20% better on float64.

Now that I got asv running, I think I need to be a little more systematic about the development of this.

skimage/_shared/interpolation.pxd
```diff
@@ -28,62 +32,65 @@ cdef inline Py_ssize_t fmax(Py_ssize_t one, Py_ssize_t two) nogil:
 cdef inline Py_ssize_t fmin(Py_ssize_t one, Py_ssize_t two) nogil:
     return one if one < two else two
 
+ctypedef fused np_real_numeric_out:
```
stefanv: What's happening here?

hmaarrfk: I'm forcing cross type compilation too. so uint8->uint16, but more importantly uint8->float64

hmaarrfk:
@stefanv, the fact that `np_real_numeric` is the same type for `cval` and `image` means that they are necessarily the same time.

`np_real_numeric_out` allows the output to be of a different `dtype` following standard c type conversion.

However, I think you are correct in thinking that it is related to an extra cast somewhere.

skimage/_shared/fused_numerics.pxd
jni: I agree about Cython building this in. Perhaps that should be your next PR? ;) In the meantime I'm happy to wait for you to hunt down the errant cast... And I'm also happy to merge, since ~90% of our warping is probably float-based.

soupault: >Perhaps that should be your next PR? ;
I'd also suggest writing a book on Practical Cython 😅. Seems like @hmaarrfk is a rare expert in this matter.

skimage/_shared/interpolation.pxd
hmaarrfk: I added a note about this.

hmaarrfk:
Final closing remarks on my experience addign fused types in this:

It seems that might actually have sped up `uint8` by a little bit. A real benchmark on a dedicated computer would be necessary for this with interpolation order=0.

interpolation order=3 seems to be compute bound and not memory bound.
To speed that one up, we need to make use of the AVX instructions.
Unfortunately, I'm not a gcc guru so I don't know if setting `export CFLAGS=-march=native` was enough to enable it in cython.

It did help by about 20% in the CPU bound case of a 4096x4096 matrix.

This might hit the limitations of what cython can do. Not too sure. 

This might be a good demo for Pythran or Numba to attack.

hmaarrfk:
I don't know why no speedup is seen with the float32 operations. Seems like you should be able to get a 2x speed up from going to float32 because SSE(2) can make 2x as many vectorized operations.

Currently, this is an
O(NxM) algorith, with memory order 1. 

Maybe for vectorization, we can make this a
memory order (NxM) and computational order (NxM) algorithm.

Anyway, just some thoughts.

skimage/transform/_warps_cy.pyx
```diff
@@ -9,65 +9,70 @@ from .._shared.interpolation cimport (nearest_neighbour_interpolation,
...
-cdef inline void _transform_affine(double x, double y, double* H,
-                                   double *x_, double *y_) nogil:
+cdef inline void _transform_affine(np_floats x, np_floats y, np_floats* H,
+                                   np_floats *x_, np_floats *y_) nogil:
     """Apply an affine transformation to a coordinate.
 
     Parameters
     ----------
-    x, y : double
+    x, y : np_float
```
stefanv: Considering our other discussion, do you think `np_float` is unambiguous here?

hmaarrfk: haha totally, i'll revert the notation!

hmaarrfk:
@stefanv I'm really not sure how to interpret them :/

```
     [6f45aaa0]       [e4864e44]
     <master>         <feature_fused_warping>
+         149±9μs        302±100μs     2.03  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.uint16'>, 128, 0)
+         125±9μs          193±5μs     1.55  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.uint8'>, 128, 0)
+        375±20μs        581±200μs     1.55  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.uint8'>, 128, 1)
+         122±2μs          188±4μs     1.54  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.float64'>, 128, 0)
+     1.13±0.05ms       1.73±0.2ms     1.53  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.uint16'>, 128, 3)
+        433±10μs        647±100μs     1.49  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.uint16'>, 128, 1)
+         375±9μs         535±20μs     1.43  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.uint16'>, 128, 1)
+         372±6μs         500±30μs     1.34  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.uint8'>, 128, 1)
+         376±9μs         505±10μs     1.34  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.float64'>, 128, 1)
+         153±3μs         201±20μs     1.31  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.uint8'>, 128, 0)
+        145±10μs          188±4μs     1.29  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.float32'>, 128, 0)
+     1.02±0.01ms      1.21±0.03ms     1.19  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.uint16'>, 128, 3)
-        561±20ms         483±10ms     0.86  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.float32'>, 4096, 1)
-         266±6ms          219±7ms     0.82  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.float64'>, 4096, 0)
-        267±30ms          220±3ms     0.82  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.float64'>, 4096, 0)
-        258±20ms          196±3ms     0.76  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.float32'>, 4096, 0)
-      10.2±0.3ms       7.63±0.2ms     0.75  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.uint16'>, 1024, 0)
-        15.4±1ms       10.5±0.7ms     0.68  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.float64'>, 1024, 0)
-      10.9±0.8ms       7.36±0.2ms     0.68  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.uint8'>, 1024, 0)
-        264±40ms          173±4ms     0.66  benchmarks_transform_warp.WarpSuite.time_to_float64(<class 'numpy.uint8'>, 4096, 0)
-         295±8ms          176±4ms     0.60  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.float32'>, 4096, 0)
-      13.2±0.7ms       7.86±0.2ms     0.60  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.float32'>, 1024, 0)
-         286±6ms         150±10ms     0.52  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.uint16'>, 4096, 0)
-         279±9ms          121±8ms     0.43  benchmarks_transform_warp.WarpSuite.time_same_type(<class 'numpy.uint8'>, 4096, 0)
```

hmaarrfk:
I think @jni is more ahead of me on this one. I just want to see the results of a command like


```
asv continuous
```

jni:
@stefanv see #3486, which aims to supersede this one.

jni:
@hmaarrfk those benchmarks look like small sizes might be dominated by the dispatch time, whereas big sizes benefit from type-specific computing?

What asv command did you use to run that?

hmaarrfk:
I wish I remembered.....

hmaarrfk:
Here is the docs for it

https://asv.readthedocs.io/en/stable/commands.html#asv-continuous

PR #3255, opened 2018-07-07
Add benchmarks for morphology.local_maxima

lagru:
Basically two cases are benchmarked: A natural 2d image with many small extrema and a synthetic 3D "image" with few large extrema. The memory behavior of the latter is especially interesting because it requires the internal queue to grow quite a bit.

## Checklist
  later.
- [x] <strike>Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.</strike>
- [X] <strike>Consider backporting the PR with `@meeseeksdev backport to v0.14.x`</strike>

### DIFF
--- a/benchmarks/benchmark_morphology.py
+++ b/benchmarks/benchmark_morphology.py
@@ -239,7 +239,7 @@ class LocalMaxima(object):
     def peakmem_reference(self, *args):
         """Provide reference for memory measurement with empty benchmark.
 
-        .. [1]: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory
+        .. [1] https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory
         """
         pass
 
--- a/benchmarks/benchmark_morphology.py
+++ b/benchmarks/benchmark_morphology.py
@@ -220,3 +220,31 @@ class GrayReconstruction(object):
     def peakmem_reconstruction(self, shape, dtype):
         morphology.reconstruction(self.seed, self.mask)
 
+
+class LocalMaxima(object):
+
+    param_names = ["connectivity", "allow_borders"]
+    params = [(1, 2), (False, True)]
+
+    def setup(self, *args):
+        # Natural image with small extrema
+        self.image = data.moon()
+
+    def time_2d(self, connectivity, allow_borders):
+        morphology.local_maxima(
+            self.image, connectivity=connectivity,
+            allow_borders=allow_borders
+        )
+
+    def peakmem_reference(self, *args):
+        """Provide reference for memory measurement with empty benchmark.
+
+        .. [1]: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory
+        """
+        pass
+
+    def peakmem_2d(self,connectivity, allow_borders):
+        morphology.local_maxima(
+            self.image, connectivity=connectivity,
+            allow_borders=allow_borders
+        )


jni:
@lagru do you have a reference for how long these benchmarks take? I feel like the parameter matrix might be a bit too big. And I would argue that running `local_minima` on Hubble is unrealistic and not very relevant, while running it on the checkerboard is redundant.

You can run these locally quite easily. See the "asv run" help:

http://asv.readthedocs.io/en/latest/commands.html

You should probably limit yourself to a single environment from the benchmark matrix, as the envs take quite a while to make.

lagru:
@jni Ha :smile:, and here I expected that you'd have asked for benchmarks for `local_minima` if I had omitted them. My thought process was that `local_minima` using `local_maxima` under the hood is an implementation detail and not part of the API. Although I doubt that will change.

I haven't used `local_maxima` myself yet but using it without any kind of filtering on a natural image may not be "realistic" anyway. However I think these two extreme benchmarks cover a large part of `local_maxima`s use cases quite well.

I'm happy to do a quick check on my machine although I doubt that my notebook is a good reference.

lagru:
@jni On my machine (4 x Intel Core i5-4200U CPU @ 1.60GHz) running the 4 remaining benchmarks takes roughly 1 Minute and 20 seconds.

lagru:
@jni With 33b8572 all 4 benchmarks together take ~20 s on my machine. Is that enough? Otherwise our only options are to remove the parameter "connectivity" or skip the 3D benchmark.

jni:
@lagru why not just reduce the size of the 3D image by reducing the sidelength of the cube, e.g. from (10, 10, 10) to (5, 5, 5)? That should give an 8x reduction in runtime, while testing the exact same principles?

lagru:
@jni That has not much of an effect; the benchmarks take essentially the same time. The size of the array doesn't seem to be the factor here. Even if I use only really small arrays the 4 benchmarks take roughly 20 s. Of note is that these results are only true for the development version of asv (0.3.dev1274+a88114ce); for the most recent release 0.2.1 all 4 benchmarks take 45 seconds!

At this stage I really think that testing on my weak CPU may not be a good reference point. And I really want to avoid reducing the parameter matrix further because I think it's quite valuable to see the algorithms performance for different connectivity and border behavior.

Oh and I found a small bug: `local_maxima([1, 1], allow_borders=False)` should return `array([0, 0])` but raises a ValueError in `_offset_to_raveled_neighbors`. Should I open an issue for this or just raise a PR when I find the time?

sciunto:
@lagru If you think that you can open a PR in a short term, then fine. Otherwise, you are invited to open an issue to do not forget. :)

sciunto:
@lagru Can you rebase this PR please? Otherwise, it looks good to me. @jni, do you agree after the last @lagru's comment?

lagru:
Benchmark suite won't pass currently. See #4176 for a fix.

mkcor:
Hey @lagru!! :wink: 

Is this PR still relevant? The fourth task (out of four) is not relevant anymore; it may be either checked off or removed.

lagru:
@mkcor, relevant in as much that more benchmarks don't hurt? I updated it and would say it's ready to go.

benchmarks/benchmark_morphology.py
```diff
@@ -220,3 +220,31 @@ def peakmem_reference(self, *args):
...
+    def time_2d(self, connectivity, allow_borders):
+        morphology.local_maxima(
+            self.image, connectivity=connectivity,
+            allow_borders=allow_borders
+        )
+
+    def peakmem_reference(self, *args):
+        """Provide reference for memory measurement with empty benchmark.
+
+        .. [1]: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory
```
mkcor, line 242:
```suggestion
        .. [1] https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory
```
I can see this colon elsewhere in the benchmark docstrings, but I don't think it's standard in the rest of our docs.

mkcor (approved):

benchmarks/benchmark_morphology.py
lagru: Thanks. :)

mkcor:
> more benchmarks don't hurt

@lagru definitely! Since there hadn't been any activity in a while (:zzz: Dormant :wink:), I was wondering if maybe it was outdated. I guess this PR had simply fallen in the cracks! Should I wait for @jni's approval before merging? Or would you like to remove this review request?

lagru:
I guess it had just fallen out of sight. I was wondering how you found this. :smile: As far as I'm aware I addressed all of @jni's feedback so I'm happy to get this merged.

PR #3260, opened 2018-07-10
benchmark: add some warping benchmarks to create a baseline

hmaarrfk:
PR #3253 is about speeding up computation when you want to.

I figured I would add some benchmarks to create a baseline.

It considers 2 cases:
1. The case where want to go to "float64"
2. The case where we don't care and just want the warped result as the final image.

I really wish I could test this on my own machine but I tried to install asv:

```
pip install git+https://github.com/airspeed-velocity/asv.git@v0.2.2rc1#egg=asv
```
and
```
pip install git+https://github.com/airspeed-velocity/asv.git#egg=asv
```
but all failed with

```python
               File "/home/mark/miniconda3/envs/sk/lib/python3.6/site-packages/asv/benchmark.py", line 795, in <module>
                 commands[mode](args)
               File "/home/mark/miniconda3/envs/sk/lib/python3.6/site-packages/asv/benchmark.py", line 740, in main_discover
                 list_benchmarks(benchmark_dir, fp)
               File "/home/mark/miniconda3/envs/sk/lib/python3.6/site-packages/asv/benchmark.py", line 725, in list_benchmarks
                 for benchmark in disc_benchmarks(root):
               File "/home/mark/miniconda3/envs/sk/lib/python3.6/site-packages/asv/benchmark.py", line 710, in disc_benchmarks
                 for module in disc_files(root, os.path.basename(root) + '.'):
               File "/home/mark/miniconda3/envs/sk/lib/python3.6/site-packages/asv/benchmark.py", line 699, in disc_files
                 module = import_module(package + filename)
               File "/home/mark/git/scikit-image/.asv/env/eef864a9fed06fd59ca8023f9f5b02ad/lib/python3.6/importlib/__init__.py", line 126, in import_module
                 return _bootstrap._gcd_import(name[level:], package, level)
               File "<frozen importlib._bootstrap>", line 994, in _gcd_import
               File "<frozen importlib._bootstrap>", line 971, in _find_and_load
               File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
               File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
               File "<frozen importlib._bootstrap_external>", line 678, in exec_module
               File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
               File "/home/mark/git/scikit-image/benchmarks/transform_warp.py", line 2, in <module>
                 from skimage.transform import SimilarityTransform, warp

```

## Checklist
  existing benchmark
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/benchmarks/transform_warp.py
+++ b/benchmarks/transform_warp.py
@@ -1,0 +1,34 @@
+import numpy as np
+from skimage.transform import SimilarityTransform, warp
+from skimage.util.dtype import convert
+
+
+class WarpSuite:
+    params = ([np.uint8, np.uint16, np.float32, np.float64],
+              [128, 1024, 4096],
+              [0, 1, 3]
+              )
+    param_names = ['dtype_in', 'N', 'order']
+
+    def setup(self, dtype_in, N, order):
+        self.image = convert(np.random.random((N, N)), dtype=dtype_in)
+        self.tform = SimilarityTransform(scale=1, rotation=np.pi / 10,
+                                         translation=(0, 4))
+        self.order = order
+
+    def time_same_type(self):
+        """Test the case where the users wants to preserve their same low
+        precision data type."""
+        result = warp(self.image, self.tform, order=self.order,
+                      preserve_range=True)  # , dtype=self.image.dtype)
+
+        # With PR #3253, this line will be unecessary, and we can pass
+        # the parameter dtype_out above to specify casting back to the
+        # same time
+        result = result.astype(self.image.dtype)
+
+    def time_to_float64(self):
+        """Test the case where want to upvert to float64 for continued
+        transformations."""
+        result = warp(self.image, self.tform, order=self.order,
+                      preserve_range=True)  # , dtype=np.float64)
--- a/benchmarks/transform_warp.py
+++ b/benchmarks/transform_warp.py
@@ -1,6 +1,7 @@
 import numpy as np
 from skimage.transform import SimilarityTransform, warp
 from skimage.util.dtype import convert
+import warnings
 
 
 class WarpSuite:
@@ -11,12 +12,14 @@ class WarpSuite:
     param_names = ['dtype_in', 'N', 'order']
 
     def setup(self, dtype_in, N, order):
-        self.image = convert(np.random.random((N, N)), dtype=dtype_in)
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", "Possible precision loss")
+            self.image = convert(np.random.random((N, N)), dtype=dtype_in)
         self.tform = SimilarityTransform(scale=1, rotation=np.pi / 10,
                                          translation=(0, 4))
         self.order = order
 
-    def time_same_type(self):
+    def time_same_type(self, dtype_in, N, order):
         """Test the case where the users wants to preserve their same low
         precision data type."""
         result = warp(self.image, self.tform, order=self.order,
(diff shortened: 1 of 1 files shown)


benchmarks/transform_warp.py
```diff
@@ -0,0 +1,31 @@
...
+              )
+    param_names = ['dtype_in', 'N', 'order']
+
+    def setup(self, dtype_in, N, order):
+        self.image = convert(np.random.rand(N, N), dtype=dtype_in)
+        self.tform = SimilarityTransform(scale=1, rotation=np.pi / 10,
+                                         translation=(0, 4))
+        self.order = order
+
+    def same_type(self):
```
lagru: I think you'll need to  prefix these functions with [`time_`](https://asv.readthedocs.io/en/stable/writing_benchmarks.html#timing) otherwise asv won't run / find them. And the signature must be able to accept all parameters as positional arguments because asv passes them implicitly to all `setup`, `time_x` and `peakmem_` methods of the class.

lagru:
> I really wish I could test this on my own machine but I tried to install asv [...] but all failed with...

Is the traceback shown during installation of asv or when you try to run it?

I managed to install asv using either conda (conda-forge channel) or providing pip with the path to the cloned repository (which is similar to your attempts).

benchmarks/transform_warp.py
```diff
@@ -0,0 +1,31 @@
...
+        """Test the case where the users wants to preserve their same low
+        precision data type."""
+        result = warp(self.image, self.tform, order=self.order,
+                      preserve_range=True)
+        convert(result, dtype=self.image.dtype, copy=False)
```
lagru: Right now you are measuring the time it takes to execute `warp` *and* `convert` together. Is this intended? Wouldn't it be better to keep these measurements separate?

hmaarrfk:
@lagru. I’ll try to install again tonight.

I do want to benchmark conversion and warping together. Through
conversations with @jni, he mentionnés that it was crucial to try and keep
everything in the floating point representation during intermediate
processing steps for multi step algorithms.

That said, the original data that you might ty and warp might be in uint
data types. I’m wondering if the CPU implicit casts are faster than
converting everything to a given data type upfront before calling warp.

jni:
@hmaarrfk yes, floating point values are critical for precision, but it is still useful to measure the time for each step separately. Therefore I would recommend benchmarks with and without conversion.

hmaarrfk:
@lagru @jni thank you for your comments. 

I was treating the convert step as a "nop" in the `time_to_float64`. I feel like a common use case (at least for myself) is  to cast things back to the original type after using many scikit-image functions.

My hypothesis:
1. `time_to_float64` test the current performance where the output is a float 64. This should improve for cases where the `dtype` of the image is not `float64` with PR #3253 as the computation will implicitly upcast to float64 (hopefully) but stay the same when the dtype is `float64`.
2. `time_same_type`: should show improvements for the case when users need lower precision than 64bit. That said, the warping coefficients are still double. I should experiment to see if there are any gains in setting the warping coefficients to single precision when low precision types are used.

I had erroneously used `convert` as opposed to `astype` within the benchmarks.

hmaarrfk:
@jni done!

jni:
@hmaarrfk 

> i'm just blind.

I suspected that might be the case as I saw your comments rolling in... =D

hmaarrfk:
@jni, I realize i needed to install `asv` with system python3 and run it with bash or some combination thereof.

doesn't seem to be running correctly on my computer. mrg-1 for now.

hmaarrfk:
Closing in favour of just includeing this straight in #3253

PR #3288, opened 2018-07-24
bugfix: an index needs to be an integer not an float in CollectionViewer

Legodev:
## Description
there is an misplaced type conversation within the index calculation in `core.py -> CollectionViewer` that needs to be applied to the whole calculation not just a part of it

## Checklist
none

## References
none

## For reviewers

  later.

### DIFF
--- a/skimage/viewer/viewers/core.py
+++ b/skimage/viewer/viewers/core.py
@@ -386,7 +386,7 @@ class CollectionViewer(ImageViewer):
             key = event.key()
             # Number keys (code: 0 = key 48, 9 = key 57) move to deciles
             if 48 <= key < 58:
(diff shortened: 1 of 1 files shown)


sciunto:
@Legodev Then, you can use the function `round()` to get the closest round value, and then, cast to `int`

sciunto:
Be careful that the division doesn't have the same behavior with py2 and py3.

Legodev:
thanks for this hint, for my self i will stick to an simple cast

jni:
This looks good to me as is. I don't think an off-by-one error in this scenario is a big deal. Thanks @Legodev for the contribution and @sciunto for the initial review!

soupault:

PR #3292, opened 2018-07-26
Replace scipy.sparse.*.todense() with toarray()

jni:
NumPy 1.15 started warning about not using `np.matrix`, which breaks our
CI, because `todense()` returns a `np.matrix` rather than a `np.array`.
Replacing this with `toarray()` removes the warning and has no effect on
the expressions where it is being replaced.

## Checklist
  existing benchmark
- [x] Unit tests

## References
For an example of the errors, see [this build](https://travis-ci.org/scikit-image/scikit-image/jobs/407599944)

## For reviewers

  later.


sciunto:
I was also investigating this, but you were faster than me.

soupault:

PR #3295, opened 2018-07-27
BUILD: Add pyproject.toml to ensure cython is present

thewtex:
PEP 517, implemented by pip 10 and newer, will ensure that cython is
present for the build by specifying the build dependencies in
pyproject.toml.

## Description

This intends to address an issue reported on the mailing; the build will fail if Cython is not pre-installed.

## References

https://www.python.org/dev/peps/pep-0517/

## For reviewers

  later.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -6,6 +6,7 @@ include *.rst
 recursive-include requirements *.txt
 include requirements/README.md
 include Makefile
+include pyproject.toml
 include skimage/scripts/skivi
 recursive-include skimage *.pyx *.pxd *.pxi *.py *.c *.h *.ini *.md5 *.npy *.txt *.in *.cpp *.md
 recursive-include skimage/data *
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "wheel", "cython"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel", "cython"]
+requires = ["setuptools", "wheel", "numpy", "cython"]


soupault:
@thewtex Hi! See the discussion in https://github.com/scikit-image/scikit-image/pull/3020.

thewtex:
@soupault thanks for the note! -- I was not aware of that patch. However, I think this can still be merged as-is. Pip has many improvements and releases since then. NumPy or a version specification is not added, so there should be no conflicts.

This addresses the original bug report that the build fails if Cython was not installed.

jni:
I agree with @thewtex that this PR is a lot less ambitious, and simpler, than #3020 and solves a pressing need, so this should be merged.

@thewtex do you have a nice way to test this in CI to make sure it's working as advertised?

thewtex:
> @thewtex do you have a nice way to test this in CI to make sure it's working as advertised?

I verified locally with:

1. Ensuring that the `cython` executable is not available
2. Create a virtualenv
3. `pip install -e .`

and the error was reproduced and addressed. Looking at the CI scripts, it does not seem that they install cython directly, but with `requirements/build.txt` through some path. However, we should not remove cython there because it still needs to be installed explicitly if the legacy `python setup.py develop` is used instead of `pip install -e .`.

kne42:
Could we also get this backported to v0.14.x so that I no longer have to explicitly install Cython prior to scikit-image in my CI builds?

jni:
@thewtex (et al) thanks for the local test, but I kind of think we should be testing this in CI. Happy to leave this to a later PR though. If @scikit-image/core agrees, I'll raise an issue.

soupault:

hmaarrfk:
Sigh, I think I am chiming in too late, sorry about that.

The previous build system was actually designed so as not to have cython as an install time dependency from the `sdist`.
This is made clear by explicitely including the precompiled `.c` files and `.cpp` file in the `sdist`. The original issue arose because the `cythonize` function in `_build.py` file wasn't checking for  `_harr.cpp` that was generated from `_harr.pyx` it. It was checking for `_harr.c` whch doesn't exist.

I think that one of two things should be done:
1. `cython` should be removed from the install dependencies.
2. The logic for checking for `cythonized` modules should be removed.

Cython actually [recommends 1.](https://cython.readthedocs.io/en/latest/src/reference/compilation.html#distributing-cython-modules) though I guess that is for us to decide.

PR #3304, opened 2018-08-01
Backport #3242: handle NumPy 1.15 warnings in dependencies

jni:
See #3242 for details.

PR #3310, opened 2018-08-02
Backport of #3238

jni:
## Description
See #3238 for details.

@soupault it was only four commits so I just cherry-picked them individually. I agree that merge commits are messy. =)

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -28,7 +28,8 @@ Improvements
 API Changes
 -----------
 
-
+- ``rectangular_grid`` now returns a tuple instead of a list to improve
+  compatibility with NumPy 1.15.
 
 Deprecations
 ------------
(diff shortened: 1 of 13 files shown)


jni:
@soupault oops! Fixed!

hmaarrfk:
@jni I think this PR should be backported first. then rebase #3304 and the tests **should** pass.

hmaarrfk:
Does github allow you to set rules for what you can merge? Seems like @stefanv has BDFL powers. Did he not extend those to you @jni?

jni:
@hmaarrfk 

> Does github allow you to set rules for what you can merge? Seems like @stefanv has BDFL powers. Did he not extend those to you @jni?

We all have the same powers, but we endeavour to use them responsibly. =) By an unwritten agreement that needs to become written (I alluded to this in my roadmap proposal), we only merge after two reviews. But the backporting/LTS thing is new and we don't have a policy for these.

PR #3311, opened 2018-08-02
cython: check for presence of cpp files during install from sdist

hmaarrfk:
Only used if cython isn't present. Allows for the installation from sdist without 
cython.

This was brought up on the mailing list. The issue is that `_harr.pyx` compiles to `_harr.cpp` and previous to this PR, the logic only checked for `_harr.c`. This also checks for `.cpp` for all the `pyx` files provided.

This allows us to remove `cython` from `pyproject.toml`

~~This PR should be backported too.~~

## References
https://mail.python.org/pipermail/scikit-image/2018-July/005597.html

## For reviewers

  later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
   include:
     - os: linux
       python: 3.5
-      env: OPTIONAL_DEPS=1 WITH_PYSIDE=1 BUILD_DOCS=1
+      env: OPTIONAL_DEPS=1 WITH_PYSIDE=1 BUILD_DOCS=1 INSTALL_FROM_SDIST=1
     - os: linux
       python: 3.5
       env: QT=PyQt5 MINIMUM_REQUIREMENTS=1
@@ -87,8 +87,14 @@ before_install:
     - tools/check_sdist.py $SDIST_NAME
 
 install:
-    - python setup.py develop
     - ccache --show-stats
+    # Test installing without cython using the sdist
+    - if [[ "$INSTALL_FROM_SDIST" == "1" ]]; then
+        pip uninstall cython -y ;
+        pip install dist/scikit-image-*.tar.gz ;
+      else
+        python setup.py develop ;
+      fi
     # Install testing requirements
     - pip install --retries 3 -q $PIP_FLAGS -r requirements/test.txt
     # Matplotlib settings - do not show figures during doc examples
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,10 +90,10 @@ install:
     - ccache --show-stats
     # Test installing without cython using the sdist
     - if [[ "$INSTALL_FROM_SDIST" == "1" ]]; then
-        pip uninstall cython -y ;
-        pip install dist/scikit-image-*.tar.gz ;
+        pip uninstall cython -y;
+        pip install dist/scikit-image-*.tar.gz;
       else
-        python setup.py develop ;
+        pip install --no-build-isolation .;
       fi
     # Install testing requirements
     - pip install --retries 3 -q $PIP_FLAGS -r requirements/test.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,7 +89,7 @@ before_install:
 install:
     - ccache --show-stats
     # Test installing without cython using the sdist
-    - if [[ "$INSTALL_FROM_SDIST" == "1" ]]; then
+    - if [[ $INSTALL_FROM_SDIST ]]; then
         pip uninstall cython -y;
         pip install dist/scikit-image-*.tar.gz;
       else
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,6 +89,10 @@ before_install:
 install:
     - ccache --show-stats
     # Test installing without cython using the sdist
+    # --no-build-isolation ensures even in the presense of a pyproject.toml
+    # pip will not create a virtual environment, just for building the package
+    # This is problematic because the version of numpy of that virtual environment
+    # may be higher than the version we want to test with.
     - if [[ $INSTALL_FROM_SDIST ]]; then
         pip uninstall cython -y;
         pip install dist/scikit-image-*.tar.gz;
--- a/skimage/_build.py
+++ b/skimage/_build.py
@@ -13,6 +13,19 @@ except NameError:
         pass
 
 
+def _compiled_filename(f):
+    """Check for the presence of a .pyx[.in] file as a .c or .cpp."""
+    basename = f.replace('.in', '').replace('.pyx', '')
+    for ext in ('.c', '.cpp'):
+        filename = basename + ext
+        if os.path.exists(filename):
+            return filename
+    else:
+        raise RuntimeError('Cython >= %s is required to build '
+                           'scikit-image from git checkout' %
+                           CYTHON_VERSION)
+
+
 def cython(pyx_files, working_path=''):
     """Use Cython to convert the given files to C.
 
@@ -34,15 +47,12 @@ def cython(pyx_files, working_path=''):
         from Cython.Build import cythonize
     except ImportError:
         # If cython is not found, the build will make use of
-        # the distributed .c files if present
-        c_files = [f.replace('.pyx.in', '.c').replace('.pyx', '.c') for f in pyx_files]
-        for cfile in [os.path.join(working_path, f) for f in c_files]:
-            if not os.path.isfile(cfile):
-                raise RuntimeError('Cython >= %s is required to build scikit-image from git checkout' \
-                                   % CYTHON_VERSION)
+        # the distributed .c or .cpp files if present
+        c_files_used = [_compiled_filename(os.path.join(working_path, f))
+                        for f in pyx_files]
 
         print("Cython >= %s not found; falling back to pre-built %s" \
-              % (CYTHON_VERSION, " ".join(c_files)))
+              % (CYTHON_VERSION, " ".join(c_files_used)))
     else:
         pyx_files = [os.path.join(working_path, f) for f in pyx_files]
         for i, pyxfile in enumerate(pyx_files):
--- a/skimage/conftest.py
+++ b/skimage/conftest.py
@@ -8,15 +8,11 @@ if Version(np.__version__) >= Version('1.14'):
     np.set_printoptions(legacy='1.13')
 
 # List of files that pytest should ignore
-collect_ignore = ["setup.py",
-                  "skimage/io/_plugins",
-                  "doc/",
-                  "tools/",
-                  "viewer_examples"]
+collect_ignore = ["io/_plugins",]
 try:
     import visvis
 except ImportError:
-    collect_ignore.append("skimage/measure/mc_meta/visual_test.py")
+    collect_ignore.append("measure/mc_meta/visual_test.py")
 
 # importing skimage.novice issues some warnings. Without these lines,
 # pytest issues numerous warnings when crawling the package.
--- a/tools/travis/before_install.sh
+++ b/tools/travis/before_install.sh
@@ -29,8 +29,19 @@ export WHEELHOUSE
 export DISPLAY=:99.0
 # This causes way too many internal warnings within python.
 # export PYTHONWARNINGS="d,all:::skimage"
-# codecov will combine different paths
-export TEST_ARGS="--doctest-modules --cov=skimage"
+
+# Doctest's don't run well on other's systems because of the
+# need for conftest.py
+# Maybe it is time to bring back the test function
+# https://github.com/scikit-image/scikit-image/pull/3164
+# Though, the configuration of conftest would have to be added to it
+if [[ "$INSTALL_FROM_SDIST" != "1" ]]; then
+  # codecov will combine different paths
+  export TEST_ARGS="--doctest-modules --cov=skimage"
+fi
+if [[ "$OPTIONAL_DEPS" == "1" ]]; then
+  export TEST_ARGS="${TEST_ARGS} --cov=skimage"
+fi
 
 retry () {
     # https://gist.github.com/fungusakafungus/1026804
--- a/tools/travis/before_install.sh
+++ b/tools/travis/before_install.sh
@@ -30,18 +30,7 @@ export DISPLAY=:99.0
 # This causes way too many internal warnings within python.
 # export PYTHONWARNINGS="d,all:::skimage"
 
-# Doctest's don't run well on other's systems because of the
-# need for conftest.py
-# Maybe it is time to bring back the test function
-# https://github.com/scikit-image/scikit-image/pull/3164
-# Though, the configuration of conftest would have to be added to it
-if [[ "$INSTALL_FROM_SDIST" != "1" ]]; then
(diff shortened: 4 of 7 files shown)


hmaarrfk:
Adding the discussion from my previous post here: https://github.com/scikit-image/scikit-image/pull/3295#issuecomment-410016924

I chimed in to #3295 too late. Sorry.

The previous build system was actually designed so as not to have cython as an install time dependency from the `sdist`.
This is made clear by explicitely including the precompiled `.c` files and `.cpp` file in the `sdist`. The original issue arose because the `cythonize` function in `_build.py` file wasn't checking for  `_harr.cpp` that was generated from `_harr.pyx` it. It was checking for `_harr.c` whch doesn't exist. PR #3311 addresses this issue.

I think that one of two things should be done:
1. `cython` should be removed from the install dependencies.
2. The logic for checking for `cythonized` modules should be removed.

Cython actually [recommends 1.](https://cython.readthedocs.io/en/latest/src/reference/compilation.html#distributing-cython-modules) though I guess that is for us to decide.

jni:
@hmaarrfk it would be nice for this to be tested somehow...

hmaarrfk:
I'll gladly set us up on circle ci just for that :stars:

hmaarrfk:
@jni I think this should test installing from sdist without cython and avoid future problems.

skimage/_build.py
```diff
@@ -34,15 +34,25 @@ def cython(pyx_files, working_path=''):
...
-        # the distributed .c files if present
-        c_files = [f.replace('.pyx.in', '.c').replace('.pyx', '.c') for f in pyx_files]
-        for cfile in [os.path.join(working_path, f) for f in c_files]:
-            if not os.path.isfile(cfile):
-                raise RuntimeError('Cython >= %s is required to build scikit-image from git checkout' \
-                                   % CYTHON_VERSION)
+        # the distributed .c or .cpp files if present
+        c_files_used = []
+        for pyx_file in [os.path.join(working_path, f) for f in pyx_files]:
+            pyx_file = pyx_file.replace('.in', '')
```
stefanv: I like this PR in general, but I feel this part of the code path can be improved with something like:

```
def compiled_fn(f):
    base = f.replace('.pyx.in', '')
    for ext in ('.cpp', '.c'):
        fn = base + ext
        if os.path.exists(fn):
            return fn

    raise RuntimeError('...')

cxx_files = [compiled_fn(f) for f in pyx_files]
```

hmaarrfk: Thanks @stefanv. I think the code is much improved thanks to your idea. Hopefully the CIs pass and this PR can get merged in!!

jni:
@hmaarrfk nope:

```pytb
Processing ./dist/scikit-image-0.15.dev0.tar.gz
  Installing build dependencies ... -\|/done
    Complete output from command python setup.py egg_info:
      File "/tmp/pip-req-build-2ggc502q/skimage/_build.py", line 30, in cython
        from Cython import __version__
    ImportError: No module named 'Cython'
    
    During handling of the above exception, another exception occurred:
    
      File "/tmp/pip-req-build-2ggc502q/setup.py", line 222, in <module>
        **extra
      File "/tmp/pip-build-env-l7n4724k/lib/python3.5/site-packages/numpy/distutils/core.py", line 135, in setup
        config = configuration()
      File "/tmp/pip-req-build-2ggc502q/setup.py", line 137, in configuration
        config.add_subpackage('skimage')
      File "/tmp/pip-build-env-l7n4724k/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 1037, in add_subpackage
        caller_level = 2)
      File "/tmp/pip-build-env-l7n4724k/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 1006, in get_subpackage
        caller_level = caller_level + 1)
      File "/tmp/pip-build-env-l7n4724k/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 943, in _get_configuration_from_setup_py
        config = setup_module.configuration(*args)
      File "skimage/setup.py", line 9, in configuration
        config.add_subpackage('_shared')
      File "/tmp/pip-build-env-l7n4724k/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 1037, in add_subpackage
        caller_level = 2)
      File "/tmp/pip-build-env-l7n4724k/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 1006, in get_subpackage
        caller_level = caller_level + 1)
      File "/tmp/pip-build-env-l7n4724k/lib/python3.5/site-packages/numpy/distutils/misc_util.py", line 943, in _get_configuration_from_setup_py
        config = setup_module.configuration(*args)
      File "skimage/_shared/setup.py", line 18, in configuration
        'interpolation.pyx'], working_path=base_path)
      File "/tmp/pip-req-build-2ggc502q/skimage/_build.py", line 51, in cython
        for f in pyx_files]
      File "/tmp/pip-req-build-2ggc502q/skimage/_build.py", line 51, in <listcomp>
        for f in pyx_files]
      File "/tmp/pip-req-build-2ggc502q/skimage/_build.py", line 43, in compiled_filename
        if os.path.exists(fn):
    NameError: name 'fn' is not defined
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-req-build-2ggc502q/
The command "if [[ "$INSTALL_FROM_SDIST" == "1" ]]; then pip uninstall cython -y ; pip install dist/scikit-image-*.tar.gz ; else python setup.py develop ; fi" failed and exited with 1 during .
```

hmaarrfk:
The rule should be that one build must always hang

.travis.yml
```diff
@@ -85,8 +85,14 @@ before_install:
     - tools/check_sdist.py
 
 install:
-    - python setup.py develop
     - ccache --show-stats
+    # Test installing without cython using the sdist
+    - if [[ "$INSTALL_FROM_SDIST" == "1" ]]; then
+        pip uninstall cython -y ;
```
stefanv: minor nitpick: semi-colons are unnecessary here

tools/travis/before_install.sh
```diff
@@ -29,8 +29,19 @@ export WHEELHOUSE
 export DISPLAY=:99.0
 # This causes way too many internal warnings within python.
 # export PYTHONWARNINGS="d,all:::skimage"
-# codecov will combine different paths
-export TEST_ARGS="--doctest-modules --cov=skimage"
+
+# Doctest's don't run well on other's systems because of the
```
stefanv: What does this mean?

tools/travis/script.sh
```diff
@@ -16,7 +16,13 @@ tools/build_versions.py
 section_end "List.installed.dependencies"
 
 section "Test"
-pytest $TEST_ARGS skimage
+if [[ "$INSTALL_FROM_SDIST" == "1" ]]; then
+    pushd ..
```
stefanv: I don't understand what's happening here.  Can this just be `cmd ; cd ..`?

tools/travis/script.sh
```diff
@@ -16,7 +16,13 @@ tools/build_versions.py
...
+    pytest $TEST_ARGS --pyargs skimage
+    popd
+else
+    pytest $TEST_ARGS skimage
```
stefanv: Why do these two require different forms of `pytest`?

hmaarrfk:
Right... we can’t run pytest from the source dir anymore because the source
doesn’t contain the compiled modules. Just like the tests on appveyor.

That means the config isn’t loaded and thus the doctests all go crazy.

hmaarrfk:
You mean after installing it from sdist?

stefanv:
I'm just trying to figure out how many "if ... else" splits we really need.

hmaarrfk:
I don't like them either. It might be time to add the `skimage.test`
function again!!! I wonder who removed it.....

soupault:
Me too, Mark, me too. :rofl:  (https://github.com/scikit-image/scikit-image/pull/3164) #dontblamethecontributors.

.travis.yml
hmaarrfk: ok.

tools/travis/before_install.sh
hmaarrfk: I think I was able to fix this by moving the conftest file in the `skimage` directory.

tools/travis/script.sh
hmaarrfk: `--pyargs` asks pytest to look in the installed module. This is necessary since the development directory is pretty bare and doesn't include any cythonized modules.

tools/travis/script.sh
hmaarrfk: isn't `pushd popd` the same as `cd`??? I can't run pytest from the git directory when it has been installed from source. At least I don't think so.

.travis.yml
hmaarrfk: I think they are necessary. Travis unwraps all of these into a "single line command". Really annoying, but c'est la vie.

skimage/_build.py
```diff
@@ -34,15 +34,24 @@ def cython(pyx_files, working_path=''):
...
     except ImportError:
         # If cython is not found, the build will make use of
-        # the distributed .c files if present
-        c_files = [f.replace('.pyx.in', '.c').replace('.pyx', '.c') for f in pyx_files]
-        for cfile in [os.path.join(working_path, f) for f in c_files]:
-            if not os.path.isfile(cfile):
-                raise RuntimeError('Cython >= %s is required to build scikit-image from git checkout' \
-                                   % CYTHON_VERSION)
+        # the distributed .c or .cpp files if present
+        def compiled_filename(f):
```
jni: This function should not be nested.

skimage/conftest.py
```diff
@@ -8,15 +8,11 @@
     np.set_printoptions(legacy='1.13')
 
 # List of files that pytest should ignore
-collect_ignore = ["setup.py",
-                  "skimage/io/_plugins",
-                  "doc/",
-                  "tools/",
-                  "viewer_examples"]
+collect_ignore = ["io/_plugins",]
```
jni: Moving this file is pretty genius. =D

.travis.yml
```diff
@@ -85,8 +85,14 @@ before_install:
     - tools/check_sdist.py
 
 install:
-    - python setup.py develop
     - ccache --show-stats
+    # Test installing without cython using the sdist
+    - if [[ "$INSTALL_FROM_SDIST" == "1" ]]; then
```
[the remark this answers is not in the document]
hmaarrfk: Bash scripting is super annoying. I don't understand strings and whatnot.

We can try this syntax, but as I remember it:

- We don't define `INSTALL_FROM_SDIST` in all buildis.
- Even if we did, is "string 0" truthy or falsy?

I'm just bad at shell scripting....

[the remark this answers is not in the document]
jni: > We don't define `INSTALL_FROM_SDIST`

This is true of all the other flags, or at least a good number of them. `$WHATEVER_SOME_VARIABLE_THAT_DOESNT_EXIST` evaluates to nothing/false:

```bash
 $ echo $FOOBARUAU


```

Conversely 0 does *not* evaluate to false:

```bash
 $ FOOBARAU=0
 $ if [[ $FOOBARAU ]]; then echo yes; else echo no; fi
yes
 $ if [[ $NOVARIABLEDOESNTEXIST ]]; then echo yes; else echo no; fi
no
```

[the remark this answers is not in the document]
hmaarrfk: Thanks @jni. 

I probably authored many of the other shell script tests like `== "1"`. Do you want to keep the convention in this PR and issue an a seperate "cleanup" PR for your preferred syntax?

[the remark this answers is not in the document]
jni: Let's do it right in this PR, then a separate cleanup PR can come later. I don't care too much about consistency with bad conventions. ;) e.g. if some new PR came using `skimage.img_as_float` instead of `util.img_as_float`, I'd want it fixed before merging even though we still have a few of the former in the codebase (hopefully eliminated in my latest PR)

.travis.yml
```diff
@@ -85,8 +85,14 @@ before_install:
     - tools/check_sdist.py
 
 install:
-    - python setup.py develop
     - ccache --show-stats
+    # Test installing without cython using the sdist
+    - if [[ $INSTALL_FROM_SDIST ]]; then
+        pip uninstall cython -y;
+        pip install dist/scikit-image-*.tar.gz;
```
stefanv: Since you removed the setup.py develop above, where does the sdist get generated?

hmaarrfk: `check_sdist` builds one. https://github.com/scikit-image/scikit-image/blob/master/tools/check_sdist.py#L11

jni: Ah. Perhaps it's better to break things up: `python setup.py sdist; tools/check_sdist.py`. It would make things much more readable...

hmaarrfk:
@jni @stefanv is there anything else holding up this PR?

tools/travis/script.sh
stefanv: I meant this:

```
(py36) carbo:doc (core-dev-guide*) $ pwd        
/home/stefan/src/scikit-image/doc
(py36) carbo:doc (core-dev-guide*) $ (cd ..; pwd) 
/home/stefan/src/scikit-image
(py36) carbo:doc (core-dev-guide*) $ pwd
/home/stefan/src/scikit-image/doc
```

So that can just be `(cd .. ; pytest ...)`.

hmaarrfk:
Thanks for the help Stefan!

stefanv:
No, thank you; these Travis-CI issues can be very time consuming to debug!

soupault:
@hmaarrfk could you rebase?

hmaarrfk:
@jni @soupault is there anything left to be done here?

tools/travis/script.sh
```diff
@@ -16,7 +16,11 @@ tools/build_versions.py
 section_end "List.installed.dependencies"
 
 section "Test"
-pytest $TEST_ARGS skimage
+# We can't run it in the git directory since there is a folder called `skimage`
+# in there. pytest will crawl that instead of the module.
+pushd ..
+pytest $TEST_ARGS --pyargs skimage
+popd
```
jni: @hmaarrfk can you explain the broader part of your changes? Specifically, the bit above seems like a big change that affects not only the SDIST builds, and `--no-build-isolation` vs `setup.py develop` are different. Could you explain the rationale behind these changes?

.travis.yml
```diff
@@ -87,8 +87,14 @@ before_install:
...
+      else
+        pip install --no-build-isolation .;
```
stefanv: This flag requires a comment to explain why it is necessary, for future ourselves ;)

tools/travis/script.sh
stefanv: Did we change the syntax to:

```
(cd ..; pytest ...; cd ..)
```
?

jni:
```
cd ..; pytest [...]; cd -
```
surely? ;)

skimage/conftest.py
hmaarrfk: Windows still can't use it. doctests on windows are still broken, even after that change.

PR #3317, opened 2018-08-03
Cosine Similarity for Felzenszwalb Superpixels

cmarshak:
## Description

This pull request is a minor addition to the Felzenszwalb superpixel code. We provide a keyword argument to the user if they desire to use another similarity metric for comparing adjacent superixels as done in Eq. 6 of [1], namely instead of computing the Euclidean distance as is currently done, we allow for the use of the cosine of the angle as a similarity metric . The motivation for this pull request is to make the work from [1] accessible using the toolset within skimage as this library is already extremely valuable in remote sensing. We illustrated the feature in the gist [here](https://gist.github.com/cmarshak/618a1acf5f5db8b5c4d49e54c2f3cf32).

## Checklist
- [x] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/) - I used flake8 and there are only warnings in files that I have not modified.
  existing benchmark
- [x] Unit tests - I modified the test `skimage/segmention/test_felzenszwalb` and the function `test_colors` to include a simple test for the new metric. I updated the test image to be:

```
img[:10, :10, 0] = 1
img[10:, :10, 1] = 1
img[:10, 10:, :] = [-1, -1, -1]  # vector with negative dot product with standard basis
img[10:, 10:, 2] = 1
```
So that the segments would agree with the Euclidean (the previous test image does not).

## References
[1] [Thompson, David R., et al. "Superpixel endmember detection." IEEE Transactions on Geoscience and Remote Sensing 48.11 (2010): 4023-4033.](https://ml.jpl.nasa.gov/public/mls/papers/thompson/thompson-tgrs-10.pdf)

## For reviewers

  later.


jni:
Hey @cmarshak, and thanks for the contribution!

Two comments:

1) I suspect that the argument name "distance" is a better name than "similarity", since both options are actual distance metrics.
2) I don't really see a significant difference in your gist, with my untrained eye, between the two results. Would you be able to submit a gist showing a performance improvement with real data? Ultimately we would want to include this in the gallery example for Felzenszwalb segmentation. (So, ideally, the real data would be from `skimage.data`. If none of that is suitable, I wouldn't mind adding a remote sensing-relevant image for this.)

cmarshak:
Thank you, @jni, for your encouragement.

1. I agree - that is a much better name. I will change this later this week.
2. To be honest, I have not seen a dramatic improvement in my comparisons. I have a colleague that prefers the cosine metric in his work and thought this feature should be available in sklearn so that there was an open tool to compare the two methods (the paper I cited too uses the cosine distance). I will have to spend some time later this week putting something together. I will likely use a very small portion of a landsat tile to compare carefully.

jni:
Hi @cmarshak! Generally, in scikit-image, we are aiming to provide reference algorithms, that have been proven to be useful in many diverse scenarios. I'm not sure that this fits that description, since even you have struggled to see an improvement. For the sake of comparison, I think keeping this code in a separate package should be sufficient. Something that I've learned over many years is that code is costly to maintain (let alone improve), and adding extra code should be very carefully considered. Therefore, currently, if you say that you haven't seen a dramatic improvement in your own experience, I am more inclined to say that we should not merge this PR.

If your colleague can make a compelling case in his work, then I'm certainly open to changing my mind.

Let me know what you think...

cmarshak:
This makes a lot of sense. I will do a few more experiments later this
week. If there is no *dramatic* improvement I will let you know.

Thank you again.

skimage/segmentation/tests/test_felzenszwalb.py
```diff
@@ -61,14 +61,16 @@ def test_color():
...
+    img[:10, 10:, :] = [-1, -1, -1]  # vector with negative dot product with standard basis
     img[10:, 10:, 2] = 1
-    seg = felzenszwalb(img, sigma=0)
-    # we expect 4 segments:
-    assert_equal(len(np.unique(seg)), 4)
-    assert_array_equal(seg[:10, :10], 0)
-    assert_array_equal(seg[10:, :10], 2)
-    assert_array_equal(seg[:10, 10:], 1)
-    assert_array_equal(seg[10:, 10:], 3)
+    for similarity in ['euclidean', 'cosine']:
```
hmaarrfk: search for `parametrize` in the test file to see how it can be used. It makes it very clear which parameter is failing when your tests fail. pytest will report exactly what (combination of) parameter caused the test to fail, and which one passed.

cmarshak: This is very helpful to know. Thank you!

jni:
Thank you so much for your thorough experiments and for understanding our position!

Issue #3323, opened 2018-08-05
Discussion: solution for large data files


hmaarrfk:
The size of the wheel is quite big. This has been identified as an issue in #2978 #3263 #1991
Ideally, we would just package conda and pypi subpackages  but I don't think those exist yet.... @jakirkham maybe you can confirm my suspicion?

We could decrease the size of the test images, but I feel like 5x5 images are quite boring :smile:  and may not even be possible with some of the algorithms.

I've identified 3 main sources for the large package size:
1. `data` directory: 15 MB
2. `restoration/tests`: 8MB (about to get moved to data with if PR #3324 is accepted)
3. Cythonized libraries. On my machine, there are about 48 files, totalling 50 MB uncompressed (15 MB compressed) . The cython issues seems to have been reported here: https://github.com/cython/cython/issues/2102 and more specifically in this comment https://github.com/cython/cython/issues/2102#issuecomment-397755796 See PR #3325

This proposal would split `scikit-image` into two packages:  `scikit-image` and `scikit-image-data`. It might look something like this:

![image](https://user-images.githubusercontent.com/90008/43690444-21055286-98d8-11e8-8673-a3e37664262d.png)

`scikit-image-data` would install in the data in the same directory as `scikit-image` so if a user wants to, they can have most of functionality by just installing `scikit-image` but to use the datasets we provide, they would have to install scikit-image-data

I'm trying to make sure that `scikit-image-data` depends on the same version `scikit-image`. This used to be possible with rpms and was how they dealt with development package dependencies. I was having trouble with this, and am willing to bet that this is a "non-issue" for many python packages. 
I think the `.dev0` was giving it trouble.
I don't know if it is really possible to do it with python to be honest. We might need help from upstream.

A things to ponder on:
  * We would need to tag the tests so that we know whether or not they depend on data. 
  * there are some complications in having both of these in the same top level folder. Maybe it would be easier to just have the data in a different repository. Since the data is likely going to be less volatile than the repo, this may be a pretty good idea.

If such a proposal would be accepted, there are a few things to do:
  * [ ] Update the instructions
  * [ ] Ensure the build process passes 
  * [ ] Update the conda feedstock
  * [ ] Update the wheel generation
  * [ ] Update release notes
  * [ ] (Try to) add a human readable message when a user tries to import `skimage.data` without having installed `scikit-image-data`.
  * [ ] Make sure that both packages do not have overlapping files.

[...]

PR #3324, opened 2018-08-05
Move restoration and color data to the combined data directory

hmaarrfk:
It would make it easier to think about what we want to do with all this data:

https://github.com/scikit-image/scikit-image/issues/3323

~~(I'm not sure if this will pass on python 3.5 since I used `pathlib`)~~


hmaarrfk:
@jni, so it didn't work right, not with the min requirements. I'll just add a "'string" to the front of the path to cast it as one.

jni:
Oooooh, older NumPy has the same problem I mentioned above. Of course. =) 

goddammit

hmaarrfk:
I feel like this whole string -> path is a going to be a repeat of bytearray -> string
But hopefully it won't create a new version of python :D

skimage/restoration/tests/test_restoration.py
```diff
@@ -1,15 +1,22 @@
 from os.path import abspath, dirname, join as pjoin
+from pathlib import Path
```
soupault: I've probably missed a trend... Is `os.path.join` too mainstream now?

hmaarrfk: I decided I want to backport this, and it might be better to just use os.path.join for an other year.

soupault:
There is also some data in `skimage/color/tests/data/`. Let me know if you would like to handle those files in this PR or not.

hmaarrfk:
@soupault I moved that data, but it was a really small file. The data in restoration was quite large.

soupault:
@hmaarrfk I was mentioning the ones from `color/tests/data`, not `color/tests` - https://github.com/scikit-image/scikit-image/tree/master/skimage/color/tests/data :slightly_smiling_face: .

hmaarrfk:
:carousel_horse:

soupault:
@hmaarrfk Hmm, sorry, I've just noticed that those latter files are quite small, and we don't introduce too much sense moving them... Although, the consistency is a good thing, in any case. I think, I still prefer the PR in its current state. Thanks!
@jni are you OK with the changes?

hmaarrfk:
Thats ok. Depending on @jni's comments, I can revert them quite easily. Undoing the revert would be harder ;)

jni:
I agree with @soupault, I like the consistency of having all data files under one tree. However I wouldn't mind a third @scikit-image/core chiming in.

sciunto:
Another thought. In case we go to the online solution, we have to ensure the integrity of these images (imagine dealing with bug reports due to corrupted images...)

hmaarrfk:
Francois. The data used for tests is moved to the the data/tests directory.
A regular user should not be tempted to go there. Does that tick the box
for your requirements?

hmaarrfk:
got it!

emmanuelle:
@hmaarrfk are you considering moving the files to another directory inside ``data`` (like suggested by @sciunto ?)? [I'm going through the merge + 1 PRs and had the impression that this one was ready, but the suggestion of @sciunto seems like an improvement]

hmaarrfk:
lets see what happens with the cis.

sciunto:
it works :)

sciunto:
@hmaarrfk Should we close this PR? Turn it to an issue?

hmaarrfk:
I think it can just be closed. Pooch dealt with this

PR #3325, opened 2018-08-05
Cython: strip debug info to make smaller binaries

hmaarrfk:
## Description
Use what @cpcloud suggests to strip the debug info from the cythonized files.
https://github.com/cython/cython/issues/2102#issuecomment-401171477

I'm not sure if this is windows compatible. Lets see what appveyor says.

This brings the `.so` files on linux down from 50MB to 10MB. Compressed, they only take up 5MB of space. Most of this is taken by the rank filters.

If we also ship the data files seperately, then the core binary, that gives the scikit-image functionality is only 5MB.

## For reviewers

  later.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,12 +18,15 @@ URL = 'http://scikit-image.org'
 LICENSE = 'Modified BSD'
 DOWNLOAD_URL = 'http://github.com/scikit-image/scikit-image'
 
+import builtins
 import os
 import sys
-
-import setuptools
+import warnings
 from distutils.command.build_py import build_py
 from distutils.command.sdist import sdist
+
+import setuptools
+from numpy.distutils.command.build_ext import build_ext
 
 if sys.version_info < (3, 5):
 
@@ -39,7 +42,6 @@ For Python 2.7, please install the 0.14.x Long Term Support using:
     sys.stderr.write(error + "\n")
     sys.exit(1)
 
-import builtins
 
 # This is a bit (!) hackish: we are setting a global variable so that the main
 # skimage __init__ can detect if it is being loaded by the setup routine, to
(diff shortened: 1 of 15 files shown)


cpcloud:
Looks like the appveyor build is passing, but the job is trying to upload more than 6.25GB of artifacts.

hmaarrfk:
Awesome. Some other PR is taking care of this upload issue. 

I don’t have enough permissions to clear the artifacts myself. 

How should I attribute you the credit for this?

cpcloud:
@hmaarrfk No credit necessary, the @ mention is enough :)

hmaarrfk:
Oh your actual fix is even better. I was adding the link wrath manually to each package

soupault:
>I'm not sure if this is windows compatible.

What about MacOS? :)

```
/usr/bin/clang -bundle -undefined dynamic_lookup -arch i386 -arch x86_64 -g build/temp.macosx-10.6-intel-3.5/skimage/_shared/geometry.o -o skimage/_shared/geometry.cpython-35m-darwin.so -Wl,--strip-all
ld: unknown option: --strip-all
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ld: unknown option: --strip-all
clang: error: linker command failed with exit code 1 (use -v to see invocation)
error: Command "/usr/bin/clang -bundle -undefined dynamic_lookup -arch i386 -arch x86_64 -g build/temp.macosx-10.6-intel-3.5/skimage/_shared/geometry.o -o skimage/_shared/geometry.cpython-35m-darwin.so -Wl,--strip-all" failed with exit status 1
The command "python setup.py develop" failed and exited with 1 during .
```

hmaarrfk:
So I hacked a solution for OSX, but it isn't really robust. Namely, it checks for the OS and not the compiler.

The issue is that gcc (and MSVC?) use `--strip-all` while llvm uses `-strip-all`. Did you see that missing `-`? It isn't a typo 🐌 . 

Does anybody know how to detect which compiler will be chosen by numpy's distutils?

hmaarrfk:
We should follow what this post did: 
https://stackoverflow.com/questions/30985862/how-to-identify-compiler-before-defining-cython-extensionsv

hmaarrfk:
Interesting that the osx files are already 25MB compared to the linux ones which are 35MB.
https://anaconda.org/conda-forge/scikit-image/files

Maybe this is just a peculiarity about linux+Cython?

hmaarrfk:
It seems that on Pypi, the package is already at the suspicious 25MB instead of 35 on conda

image

hmaarrfk:
Of course @matthew-brett is already ahead of the game
https://github.com/matthew-brett/multibuild/commit/453308deaf6e045ef437d413f8791ec297fcc60d
:smile:

matthew-brett:
I should say that it wasn't me but @xhochy who was ahead of the game ... https://github.com/matthew-brett/multibuild/pull/140 - the joys of open source.

hmaarrfk:
My apologies @xhochy

hmaarrfk:
I should say that @xhochy's solution is actually supperior because there is no accurate way to test which compiler will actually be used.

PR #3326, opened 2018-08-06
codecov: ignore build scripts in coverage and don't comment on PRs

hmaarrfk:
Ignore known build scripts from codecov. This also supresses codecov from commenting on the PR, but the check at the bottom should remain.

https://docs.codecov.io/docs/ignoring-paths

Curious to see how much we are actually testing the code.

## For reviewers
  later.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,5 +5,14 @@ ignore:
   - "setup.py"
   - "**/setup.py"
 
+coverage:
+  status:
+    project:
+      default:
+        # Drops on the order 0.01% are typical even when no change occurs
+        # Having this threshold set a little higher (0.1%) than that makes it 
+        # a little more tolerant to fluctuations
+        threshold: 0.1%
+
 # Disable PR comment
 comment: off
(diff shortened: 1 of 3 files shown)


hmaarrfk:
@scikit-image/core how do you feel about this PR. As much as I would love to have github mark codecov as spam by default, it seems that this is the more direct fix.

.codecov.yml
```diff
@@ -0,0 +1,17 @@
...
+  - "scripts/skivi.py"
+  - "setup.py"
+  - "**/setup.py"
+
+coverage:
+  status:
+    project:
+      default:
+        # Allow the coverage to drop by 0.1% even when posting success
+        # Drops on the order 0.01% are typical even when no change occurs
```
soupault: Doesn't match still :smile: .

jni: @soupault yes it does. He's saying codecov can report 0.01% drop even when nothing has changed, so we set the tolerance to 0.1%.

hmaarrfk: @soupault not sure if the rewording helped. 0.1 and 0.01 look like typos ;)

Issue #3330, opened 2018-08-08
Masked image translation registration


LaurentRDC:
I'd like to gauge interest in masked image translation registration.

I often have to register translation between diffraction patterns. However, there are static image elements that throw off standard approaches based on normalized cross-correlation, like `skimage.feature.register_translation`.

I have implemented a normalized cross-correlation function that can take image masks as well, based on the following publication: [D. Padfield. Masked object registration in the Fourier domain. IEEE Transactions on Image Processing (2012)](http://www.dirkpadfield.com/Home/MaskedObjectRegistrationInTheFourierDomain.pdf?attredirects=0). The working principle is essentially the same as `skimage.feature.register_translation`, with the cross-correlation not taking into account the masked parts of images.

You can find a successful registration example [here](https://scikit-ued.readthedocs.io/en/master/tutorials/image.html#diffraction-pattern-alignment).

I have already written a Python implementation that mimics `skimage.feature.register_translation` as much as possible:
* The source code for the masked, normalized cross-correlation is [visible here as `mnxc2`](https://github.com/LaurentRDC/scikit-ued/blob/master/skued/image/correlation.py)
* The source code for the overall masked image translation registration is [visible here as `masked_register_translation`](https://github.com/LaurentRDC/scikit-ued/blob/master/skued/image/alignment.py)

The function `masked_register_translation` above passes the exact same tests as the MATLAB implementation from the original author. It also gives the same results as `skimage.feature.register_translation` in the case of trivial masks. 

I'd like to get some feedback regarding including this functionality directly in Scikit-image. The source code linked above provides a working starting point.

jni:
@LaurentRDC cool package! You should present it at a SciPy conference!

I would support inclusion of this if:

1. You could make it n-dimensional (seems doable from a cursory look at the code)
2. You could include it with minimal intrusion to both the API (adding `src_mask` and `target_mask` keyword arguments at the end) and the code (not too many additional branching if/elses inside the code). (Again, I think this seems doable.)

What do you think?

jni:
Oh, another point: at scikit-image we use `True` values in the mask to point to valid locations. I think that's the opposite of what your API does? Again, trivial fix.

LaurentRDC:
@jni  Thank you for the feedback!

I understand the need for n-dimensional operations and consistent mask definition. However, it seems you are advocating for including the masked image registration functionality inside of `register_translation`.

[...]

PR #3315, opened 2018-08-02
appveyor: don't upload any artifacts

hmaarrfk:
Issue #3307

I don't really think these wheels are even used as this repo https://github.com/scikit-image/scikit-image-wheels just rebuilds them from scratch.

Should be backported.

## For reviewers

(Don't remove the checklist below.)

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

PR #3334, opened 2018-08-09
Implementation of masked image translation registration

LaurentRDC:
## Description
This pull request is related to issue #3330. The main points are described below:

I often have to register translation between diffraction patterns. However, there are static image elements that throw off standard approaches based on normalized cross-correlation, like `skimage.feature.register_translation`.

I have implemented a normalized cross-correlation function that can take image masks as well, based on the following publication: [D. Padfield. Masked object registration in the Fourier domain. IEEE Transactions on Image Processing (2012)](http://www.dirkpadfield.com/Home/MaskedObjectRegistrationInTheFourierDomain.pdf?attredirects=0). The working principle is essentially the same as `skimage.feature.register_translation`, with the cross-correlation not taking into account the masked parts of images.

The function `skimage.feature.masked_register_translation` from this pull request passes the exact same tests as the MATLAB implementation from the original author. It also gives the same results as `skimage.feature.register_translation` in the case of trivial masks. 

You can find a successful registration example [here](https://scikit-ued.readthedocs.io/en/master/tutorials/image.html#diffraction-pattern-alignment).

## Checklist
  existing benchmark
- [x] Unit tests

## For reviewers

  later.


hmaarrfk:
@LaurentRDC what did you do the .appveyor file? why are all the lines changed?

LaurentRDC:
@hmaarrfk Same changes as your pull request #3315

hmaarrfk:
Could it be your line endings? `skimage/feature/__init__.py` seems to be heavily modified as well...

hmaarrfk:
Make sure you can revert your `appveyor` config. 
1. They will probably ask you not to include that.
2. They might still want for whatever reason to upload artifacts.

hmaarrfk:
`git checkout master skimage\feature\__init__.py` might do it. I sometimes just download the file from online and copy paste it in.

And for what its worth, you can configure git to typically `commit line endings as LF` and `keep locally as is`. Any decent editor should be able to handle LF endings. That is just my personal preference though.

hmaarrfk:
you can get really fancy with
`git rebase -i master` and `git push --force` but watch out since that really messes with git/github all the CIs.

That said, I do it all the time.

emmanuelle:
Hi @LaurentRDC thanks for this contribution. Since this is a new feature, we would need to illustrate it with a gallery example. Could you write such an example?

LaurentRDC:
@emmanuelle Good suggestion, it is now included.

BTW, tests won't pass until #3400 is merged.

doc/examples/transform/plot_masked_register_translation.py
```diff
@@ -0,0 +1,68 @@
...
+
+image = data.camera()
+shift = (-22, 13)
+
+# Define areas of the image which are invalid
+# Probability of an invalid pixel is 25%
+# This could be due to a faulty detector, or edges that
+# are not affected by translation (e.g. moving object in a window)
+# See reference paper for more examples
+corruped_pixels = np.random.choice([False, True], size = image.shape, p = [0.25, 0.75])
```
emmanuelle: corrupted_pixels

doc/examples/transform/plot_masked_register_translation.py
```diff
@@ -0,0 +1,68 @@
...
+import matplotlib.pyplot as plt
+
+from skimage import data
+from skimage.feature import masked_register_translation
+from scipy.ndimage import fourier_shift
+
+image = data.camera()
+shift = (-22, 13)
+
+# Define areas of the image which are invalid
```
emmanuelle: If you want this block to appear as a text block in the rendered html, you need to add a line of "######" above (see for example ``plot_local_binary_pattern.py``).

doc/examples/transform/plot_masked_register_translation.py
```diff
@@ -0,0 +1,68 @@
...
+# Probability of an invalid pixel is 25%
+# This could be due to a faulty detector, or edges that
+# are not affected by translation (e.g. moving object in a window)
+# See reference paper for more examples
+corruped_pixels = np.random.choice([False, True], size = image.shape, p = [0.25, 0.75])
+
+# The shift corresponds to the pixel offset relative to the reference image
+offset_image = fourier_shift(np.fft.fftn(image), shift)
+offset_image = np.fft.ifftn(offset_image)
```
emmanuelle: These two lines are hard to understand if you're not familiar with Fourier transform. We could instead use juste ``ndimage.shift``, which is slightly less accurate for the registration (it uses splines) but should still give good results. I know the original example of ``plot_register_translation.py`` does the same, so it's the same question for that example.

doc/examples/transform/plot_masked_register_translation.py
```diff
@@ -0,0 +1,68 @@
...
+offset_image *= corruped_pixels
+print("Known offset (y, x): {}".format(shift))
+
+# Determine what the mask is based on which pixels are invalid
+mask = np.logical_not(np.isclose(offset_image, 0))
```
emmanuelle: I understand you're trying here to show that the mask can be determined from the image, without prior knowledge of corrupted pixels, but it seems a bit artificial to me. Could we just have ``mask = corrupted_pixels``?

skimage/feature/masked_register_translation.py
```diff
@@ -0,0 +1,274 @@
...
+        (or 1) on valid pixels. ``src_mask`` should have the same shape
+        as ``src_mask``.
+    target_mask : ndarray or None, optional
+        Boolean mask for ``target_image``. The mask should evaluate to ``True``
+        (or 1) on valid pixels. ``target_mask`` should have the same shape
+        as ``target_image``. If ``None``, ``src_mask`` will be used.
+    overlap_ratio : float, optional
+        Maximum allowed overlap ratio between masks. The correlation at pixels
+        with overlap ratio higher than this threshold will be zeroed.
+        TODO: specify effect of smaller vs larger overlap_ratio
```
emmanuelle: Don't forget the TODO :-)

emmanuelle: This parameter is hard to understand. Please also comment about the default value.

emmanuelle:
Thanks for adding the gallery example. I left a few additional comments, but we're getting there!

doc/examples/transform/plot_masked_register_translation.py
LaurentRDC: Good trick, thanks

doc/examples/transform/plot_masked_register_translation.py
LaurentRDC: Replaced with `scipy.ndimage.shift` as suggested.

doc/examples/transform/plot_masked_register_translation.py
LaurentRDC: You are right that it would be clearer. Changed

emmanuelle:
I finally took the time to read the paper, and I had one more question regarding the explanation of ``overlap_ratio``. Otherwise, looks good to me!

emmanuelle:
@scikit-image/core we'll need a second review. The paper is well written and I found it easy to follow. (Google scholar does not reference the preprint, follow the link provided by @LaurentRDC).

LaurentRDC:
@emmanuelle Thank you so much for taking the time to read the original paper!

> I had one more question regarding the explanation of `overlap_ratio`
Has this question been answered by your reading?

sciunto:
I'll look at it, I'm very interested in this feature.

emmanuelle:
@LaurentRDC I left my question as a comment in the code, I had the impression that ``overlap_ratio`` is a minimum overlap ratio (corresponding to a maximum admissible translation), and that the wording in the docstring was not so clear.

emmanuelle:
Oh I just saw that my comment doesn't appear, something went wrong when I left it, sorry about that. So, in the docstring, I would express overlap_ratio as a minimum overlap, or as a maximum translation, but not as a maximum overlap. What do you think?

skimage/feature/masked_register_translation.py
```diff
@@ -0,0 +1,275 @@
...
+    Returns
+    -------
+    shifts : ndarray
+        Shift vector (in pixels) required to register ``target_image`` with
+        ``src_image``. Axis ordering is consistent with numpy (e.g. Z, Y, X)
+
+    References
+    ----------
+    .. [1] Dirk Padfield. Masked Object Registration in the Fourier Domain.
+        IEEE Transactions on Image Processing, vol.21(5), pp. 2706-2718 (2012).
```
sciunto: Please, make the citation consistent with the example. In particular, add the DOI.

skimage/feature/masked_register_translation.py
```diff
@@ -0,0 +1,275 @@
...
+
+    Raises
+    ------
+    ValueError : if correlation `mode` is not valid, or array dimensions along
+        non-transformation axes are not equal.
+
+    References
+    ----------
+    .. [1] Dirk Padfield. Masked Object Registration in the Fourier Domain.
+        IEEE Transactions on Image Processing, vol.21(5), pp. 2706-2718 (2012).
```
sciunto: same here

skimage/feature/masked_register_translation.py
```diff
@@ -0,0 +1,275 @@
...
+
+and the author's original MATLAB implementation, available on this website:
+http://www.dirkpadfield.com/
+"""
+
+import numpy as np
+from functools import partial
+
+# next_fast_len was implemented in scipy 0.18
+# In case it cannot be imported, we use the id function
```
sciunto: Please, add a note in the TODO.txt file under the section "Others" asking to remove this piece of code when the min dep. version of scipy is set to 0.18.

skimage/feature/masked_register_translation.py
```diff
@@ -0,0 +1,275 @@
...
+# It could be much slower, but I don't think we should back-port
+# next_fast_length in its entirety
+# Take a look at the source:
+# https://github.com/scipy/scipy/blob/v1.1.0/scipy/fftpack/helper.py#L53-L151
```
sciunto: I don't think the 4 lines above are necessary.

skimage/feature/masked_register_translation.py
```diff
@@ -0,0 +1,275 @@
...
+    moving_mask = np.array(m2, dtype=np.bool)
+    eps = np.finfo(np.float).eps
+
+    # Array dimensions along non-transformation axes should be equal.
+    all_axes = set(range(fixed_image.ndim))
+    for axis in (all_axes - set(axes)):
+        if fixed_image.shape[axis] != moving_image.shape[axis]:
+            raise ValueError(
+                'Array shapes along non-transformation axes should be \
+                    equal, but dimensions along axis {a} not'.format(a=axis))
```
sciunto: What are the two last characters on the two lines above?

LaurentRDC: These were '\' for multiline strings. I thought they were necessary, but turns out that's not true! Removed.

LaurentRDC:
@emmanuelle : about the `overlap_ratio` meaning. I was thinking about it backwards because of the following. Right now, the docstrings say:

```
overlap_ratio : float, optional
    Maximum allowed overlap ratio between masks. The correlation at pixels
    with an overlap ratio higher than this threshold will be zeroed. A 
    larger percentage leads to greater robustness to spurious matches, for 
    example due to noise. A smaller percentage leads to greater 
    registration range.
```
 ` … overlap ratio higher than this threshold will be zeroed` is __wrong__. I should have written `overlap ratio lower than this threshold will be zeroed`. 

Your observation is right. `overlap_ratio` is a minimum allowable overlap. While the calculation of `overlap_ratio` involves masks, it is really the overlap between images. Note that while I got the reasoning backwards, the implementation is correct.

I would therefore rephrase the docstrings as:

```
overlap_ratio : float, optional
    Minimum allowed overlap ratio between images. The correlation at pixels
    with an overlap ratio lower than this threshold will be ignored. A lower `overlap_ratio`
    leads to smaller maximum translation, while a higher `overlap_ratio` leads to greater 
    robustness against spurious matches due to small overlap between masked images.
```

Any thoughts?

sciunto:
@scikit-image/core I would like your comment here.
This PR introduces the function "mnxc" that calculates cross-correlations of images. The function `match_template` calculate effectively a cross-correlation with the ZNCC function (zero-normalized cross-correlation).

1. Bad historical choice for `match_template`. This function must have returned a shift (as for register_translation), not a correlation function. I would say that we should think about a renaming of match_template to zncc.
2. We are getting with this PR 2 functions which calculate cross correlations. I would vote for a dedicated file for those functions.

Any thought?

emmanuelle:
@LaurentRDC I find the wording better, but I would not speak about pixels as in ``correlation at pixels`` since the correlation is measured for translation vectors, not pixels (if I understand well). Replace with "the correlation for translations corresponding to an overlap ratio..." and I will be happy :-).

emmanuelle:
@sciunto the ``mnxc`` function is not exposed in the API (maybe it could be in the future? We don't want to populate the API too much, however), so for users it's not a problem. I'm fine with moving the two cross-correlation functions to a common file, but maybe in another PR? I have the impression we're close to merging for this one.

sciunto:
I'm fine doing that after merging this one (I'm always in favor of maintaining a reasonable workload per PR). 

Regarding the API exposure of cross-correlation functions, i believe it is important as it can be useful to some users (some might want to exploit the correlation image itself, some might want to check the cross correlation quality)...  I speak with the experience of using cross-correlation functions from scikit-image and I had to modify the code of `match_template` to return everything I needed.

stefanv:
Since Dirk Padfield's code does not have an explicit license, but was used in the construction of this PR, I will email him to ask explicitly to approve this PR.

emmanuelle:
@stefanv did you write to Dirk Padfield?

skimage/feature/masked_register_translation.py
```diff
@@ -0,0 +1,278 @@
...
+    -------
+    shifts : ndarray
+        Shift vector (in pixels) required to register ``target_image`` with
+        ``src_image``. Axis ordering is consistent with numpy (e.g. Z, Y, X)
+
+    References
+    ----------
+    .. [1] Dirk Padfield. Masked Object Registration in the Fourier Domain.
+           IEEE Transactions on Image Processing, vol. 21(5), 
+           pp. 2706-2718 (2012). :DOI:`10.1109/TIP.2011.2181402`
```
emmanuelle:
```suggestion
           pp. 2706-2718 (2012). :DOI:`10.1109/TIP.2011.2181402`
.. [2] D. Padfield. "Masked FFT registration". In Proc. Computer Vision and Pattern Recognition, pp. 2918-2925 (2010).  :DOI:`10.1109/CVPR.2010.5540032`         
```

emmanuelle:
I've added a suggestion (the new feature of Github, hope I did it right) for the second paper. After this change will it be OK to merge this PR? @sciunto

LaurentRDC:
Thank you very much everyone, especially @emmanuelle, for all the help with this PR.

stefanv:
@LaurentRDC Thank you; this is a very nice PR and a great contribution!

I have one outstanding issue I would like to address:

I think `mnxc` is cryptic, and not in line with our long-standing tradition of suggestive naming.

We have the possibility here to create `cross_correlate`, which supports normalization and masking.  Or, if we don't want to implement both normalized and non-normalized behavior, then at least `cross_correlate_normalized`, which supports being run with / without masking.

LaurentRDC:
@stefanv `mnxc` isn't a great name, you are right. It's Dirk Padfield's original name. I did not anticipate there would be a desire to expose it to users.

While `mnxc` can effectively calculate the non-masked normalized cross-correlation by using trivial masks, I don't you should use `mnxc` as a basis for a general `cross_correlate` function. The masked normalized cross-correlation algorithm requires 6 (!) FFTs vs 2 for the non-masked case. 

So I see the following two things happening:
1. Rename `mnxc` to something like `masked_cross_correlate`
2. Create a `cross_correlate` functions that wraps most (all?) use cases, with a fallback to `masked_cross_correlate` in the case of masking.

stefanv:
I think we have all the pieces in place for (2), so that'd be my first choice, but (1) is a good option for Right Now.  Perhaps `cross_correlate_masked`, since that is a naming pattern we have elsewhere: `threshold_otsu`, `threshold_yen`, etc.

LaurentRDC:
@stefanv Since this PR is merged, I've made another small PR #3497 .

doc/examples/transform/plot_masked_register_translation.py
```diff
@@ -0,0 +1,72 @@
...
+print("Known offset (y, x): {}".format(shift))
+
+# Determine what the mask is based on which pixels are invalid
+# In this case, we know what the mask should be since we corrupted 
+# the pixels ourselves
+mask = corrupted_pixels
+
+shift = masked_register_translation(image, offset_image, mask)
+
+print("Detected pixel offset (y, x): {}".format(shift))
```
soupault: From the generated webpage http://scikit-image.org/docs/dev/auto_examples/transform/plot_masked_register_translation.html#sphx-glr-auto-examples-transform-plot-masked-register-translation-py:
```
Known offset (y, x): (-22, 13)
Detected pixel offset (y, x): [ 22. -13.]
```
Shouldn't they match?

LaurentRDC: Following the logic of the `plot_register_translation.py` example ([link](http://scikit-image.org/docs/dev/auto_examples/transform/plot_register_translation.html#sphx-glr-auto-examples-transform-plot-register-translation-py)), the 'detected pixel offset' value is the shift required to _shift back_.

This makes sense when trying to correct for a shift, but for a gallery example, it is confusing. I've corrected this incongruence in #3528, so that they match.

Issue #3335, opened 2018-08-10
ARM build: No warning raised matching: divide by zero, invalid value


olebole:
During the Debian build, we observe the following [test error](https://buildd.debian.org/status/fetch.php?pkg=skimage&arch=armel&ver=0.14.0-1&stamp=1530063617&raw=0) on arm (32 bit, little endian, software floating point). The sister architecture with hardware floating point does not have this problem (but fails with #3092).
```
_______________________________ test_shape_index _______________________________

    def test_shape_index():
        square = np.zeros((5, 5))
        square[2, 2] = 4
        with expected_warnings(['divide by zero', 'invalid value']):
>           s = shape_index(square, sigma=0.1)

../debian/tmp/usr/lib/python2.7/dist-packages/skimage/feature/tests/test_corner.py:174: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python2.7/contextlib.py:24: in __exit__
    self.gen.next()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

matching = ['divide by zero', 'invalid value']

    @contextmanager
    def expected_warnings(matching):
        """[...]"""
        with all_warnings() as w:
            # enter context
            yield w
            # exited user context, check the recorded warnings
            remaining = [m for m in matching if '\A\Z' not in m.split('|')]
            for warn in w:
                found = False
                for match in matching:
                    if re.search(match, str(warn.message)) is not None:
                        found = True
                        if match in remaining:
                            remaining.remove(match)
                if not found:
                    raise ValueError('Unexpected warning: %s' % str(warn.message))
            if len(remaining) > 0:
                msg = 'No warning raised matching:\n%s' % '\n'.join(remaining)
>               raise ValueError(msg)
E               ValueError: No warning raised matching:
E               divide by zero
E               invalid value

../debian/tmp/usr/lib/python2.7/dist-packages/skimage/_shared/_warnings.py:124: ValueError
```
Are the warnings really required, or may the not appear as well?

Putting @yarikoptic and @tillea in Cc

hmaarrfk:
@olebole I was definitely agreesive with catching all warnings :

[...]

PR #3337, opened 2018-08-10
Build tool: Pass tests that don't raise floating point exceptions on arm with soft-fp

hmaarrfk:
Issue #3335

Ready if this passes tests.

## For reviewers

  later.

### DIFF
--- a/skimage/feature/tests/test_corner.py
+++ b/skimage/feature/tests/test_corner.py
@@ -168,9 +168,11 @@ def test_hessian_matrix_det_3d(im3d):
 
 
 def test_shape_index():
+    # software floating point arm doesn't raise a warning on divide by zero
+    # https://github.com/scikit-image/scikit-image/issues/3335
     square = np.zeros((5, 5))
     square[2, 2] = 4
-    with expected_warnings(['divide by zero', 'invalid value']):
+    with expected_warnings([r'divide by zero|\A\Z', r'invalid value|\A\Z']):
         s = shape_index(square, sigma=0.1)
     assert_almost_equal(
         s, np.array([[ np.nan, np.nan,   -0.5, np.nan, np.nan],


hmaarrfk:
@olebole can you try to build with this: [diff](https://patch-diff.githubusercontent.com/raw/scikit-image/scikit-image/pull/3337.diff) 
[patch](https://patch-diff.githubusercontent.com/raw/scikit-image/scikit-image/pull/3337.patch)

I don't know which you need.
I only disabled the check for 1 warning. If you still fail on the other warning, I'll disable it too.

olebole:
I will test it on the weekend (will take some time, since the CPU is quite slow).
However,  I am not sure about this; at least the error message in #3335 means that _both_ warnings are not there.
What speaks against making both optional by adding `r'|\A\Z'`? Are you sure that you want to take it an error if they (on other archs) do not appear? You are later checking the result for correctness anyway.

hmaarrfk:
@odebeir Can you just tell me what the warnings are when you run this code: should be much faster.
```python
import numpy as np
def shape_index():
    l2 = np.array([[ 0.,  0.,  0.,  0.,  0.],
       [ 0., -1.,  0., -1.,  0.],
       [ 0.,  0., -2.,  0.,  0.],
       [ 0., -1.,  0., -1.,  0.],
       [ 0.,  0.,  0.,  0.,  0.]])
    l1 = np.array([[ 0.,  0.,  2.,  0.,  0.],
       [ 0.,  1.,  0.,  1.,  0.],
       [ 2.,  0., -2.,  0.,  2.],
       [ 0.,  1.,  0.,  1.,  0.],
       [ 0.,  0.,  2.,  0.,  0.]])
    return (2.0 / np.pi) * np.arctan((l2 + l1) / (l2 - l1))
print(shape_index())
```
I would also like the output of the print statement.

@jni it fails on windows!!

hmaarrfk:
@odebeir the patchc now simply makes the warning optional.
I don't like optional warnings without explanations, but with the comment pointing to the issue, I think it is fine.

olebole:
It works after some trivial fixes -- the line should be
```
with expected_warnings([r'divide by zero|\A\Z', r'invalid value|\A\Z']):
```
I now however get another error:
```
________________________________ test_spacing_1 ________________________________

    def test_spacing_1():
        [...]
        # Test with `spacing` kwarg
        # First, anisotropic along Y
        with expected_warnings(['"cg" mode' + '|' + SCIPY_EXPECTED]):
            labels_aniso = random_walker(data_aniso, labels_aniso, mode='cg',
>                                        spacing=(1., 2., 1.))

../debian/tmp/usr/lib/python2.7/dist-packages/skimage/segmentation/tests/test_random_walker.py:
263: 
/usr/lib/python2.7/contextlib.py:24: in __exit__
    self.gen.next()

matching = ['"cg" mode|numpy.linalg.matrix_rank|\\A\\Z']

    @contextmanager
    def expected_warnings(matching):
        """[...]"""
        with all_warnings() as w:
            # enter context
            yield w
            # exited user context, check the recorded warnings
            remaining = [m for m in matching if '\A\Z' not in m.split('|')]
            for warn in w:
                found = False
                for match in matching:
                    if re.search(match, str(warn.message)) is not None:
                        found = True
                        if match in remaining:
                            remaining.remove(match)
                if not found:
>                   raise ValueError('Unexpected warning: %s' % str(warn.message))
E                   ValueError: Unexpected warning: overflow encountered in long_scalars
```
I can open a new issue for this -- shall I?

hmaarrfk:
The others were pretty safe to ignore. I wonder if this is a real bug.

hmaarrfk:
Gah thanks!

jni:

jni:
@hmaarrfk thanks for this fix! =)

Issue #3045, opened 2018-05-01
Proposal: `Default` class


jni:
This is inspired by some of the answers in [this SO question](https://stackoverflow.com/questions/14749328/python-how-to-check-whether-optional-function-parameter-is-set).

I recently needed to specify a default value to a parameter where `None` was a possible value that people might want to pass, and further it was the default value that needed deprecating, so our usual pathway of "set the default to None and then change it to a real value in the future" didn't work. I needed to distinguish between "default None" and "user-given None".

The top answers in SO suggested the following:

```python
DEFAULT = object()

def myfunc(param=DEFAULT):
    if param is DEFAULT:
        warn(...)   # etc
    # etc
```

This works, but it'll be messy in the documentation:

```python
>>> DEFAULT=object()
>>> str(DEFAULT)
'<object object at 0x11dfc3370>'
```

This whole thing has been turning in my head since then and I think that we can actually make a Default class that (a) prints the right thing to the docs, so that function introspection works, and (b) greatly simplifies our deprecation procedures for default value changes.

Here's the proposal in brief:

```python
class Default:
    def __init__(self, value, next_value=None,
                 change_version=None, param_name=None):
        self.value = value
        self.next_value = next_value
        self.change_version = change_version
        self.param_name = param_name
        if next_value is not None:
            self.message = (f'The default value of {param_name} will '
                            f'change to {next_value} in {change_version}.')
        else:
            self.message = ''

    def __str__(self):
        if self.next_value is None:
            return str(self.value)
        else:
            return (f'{self.value} ({self.next_value} after '
                    f'{self.change_version}')

    def __repr__(self):
        return (f'Default({repr(self.value)}, {repr(self.param_name}, '
                f'{repr(self.next_value)}, {repr(self.change_version)})')

    def warn(self):
        if self.next_value is not None:
            warn(self.message)
```

Use it like this. If we want to deprecate an argument default value from False to True, change:

```python
def my_func(my_deprecated_arg=False):
```

to:

```python
def my_func(my_deprecated_arg=Default(False, True, 'v0.16',
                                      'my_deprecated_arg')):
    if type(my_deprecated_arg) == Default:
        my_deprecated_arg.warn()
```

And that's it! No docstrings to update, with default information that may become stale, no inconsistent deprecation messages to write, and the docs write themselves. In fact, with the right Regex in the release docs, we don't even need to add an item to the to-do: the release manager can just search for `Default...v0.whatever`.

[...]

PR #3346, opened 2018-08-13
Demo: deprecators

hmaarrfk:
This is a small demo about how a decorator I wrote could be used to transition to keyword only arguments.

@jni 

If i can remember how to compile the docs, I'll try to show how they look in screenshots

But you can read more about the package here
https://deprecation-factory.readthedocs.io/en/latest/index.html
Some advantages of using this would be:

  * The library write the warning message in the docs.
  * The library warns your users telling them how to get rid of the warning too.
  * The library will point the user to their line of code, so that they can make the appropriate modifications.
  * It is even safe to leave the deprecators in place after the threshold version has been reached. The decorator will behave as a no-op and your library will use the updated version of your code. Deprecations should not have to be blockers for your development.

## References
https://github.com/scikit-image/scikit-image/issues/3045 (not exactly related to this particular demo, but the library can deprecate default values)

## For reviewers

  later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ doc/build
 source/api
 build
 dist
-skimage/version.py
 *.swp
 .coverage
 doc/source/auto_examples/**/*
--- a/doc/tools/build_modref_templates.py
+++ b/doc/tools/build_modref_templates.py
@@ -37,7 +37,7 @@ if __name__ == '__main__':
 
     installed_version = V(module.__version__)
 
-    source_lines = open('../skimage/__init__.py').readlines()
+    source_lines = open('../skimage/version.py').readlines()
     version = 'vUndefined'
     for l in source_lines:
         if l.startswith('__version__'):
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -7,4 +7,4 @@ imageio>=2.0.1
 PyWavelets>=0.4.0
 dask[array]>=0.9.0
 cloudpickle>=0.2.1
-deprecation_factory>=0.2.3
+deprecation_factory>=0.2.4
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -7,4 +7,4 @@ imageio>=2.0.1
 PyWavelets>=0.4.0
 dask[array]>=0.9.0
 cloudpickle>=0.2.1
-deprecation-factory>=0.2.2
+deprecation_factory>=0.2.2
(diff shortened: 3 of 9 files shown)


hmaarrfk:
image

jni:
I gotta admit, that's pretty gorgeous. =)

soupault:
Give me some time, please, I've to check this PR with Federal Antimonopoly Service. :trollface: 
The library look interesting, but, from my perspective, requires more work before we can make it a dependency (if we decide so). Thanks for the demo!

hmaarrfk:
Regarding making it a dependency, I would simply version what you want from. I don't think scikit-image can afford to depend on an alpha library....

I can give scikit-image an other license for it so they don't have to drag around my license text.

I just wanted to show that it is possible, to a certain extent, not have to create long todo lists with confusion default parameters when deprecating things.

hmaarrfk:
Finally, one last thing to note is weather or not you want this kind of "magic".

Setting something to `None` and then calling the deprecated behaviour is pretty understandable, though, cumbersome to maintain. 

Those decorators are not understandable. I can maybe improve that with some refactoring, but ultimately, they rely on python introspection which is essentially magic... Me of 6 months ago would have not understood them.

jni:
@hmaarrfk thanks again for this. I want to simplify the upstream code and change some names ("kwonly" sounds pretty wonky to me), but I'm very much leaning in favour of adopting something like this. It would simplify our deprecation procedures.

As to the library name, I humbly suggest `wabisabi`. From [Wikipedia](https://en.wikipedia.org/wiki/Wabi-sabi):

> In traditional Japanese aesthetics, wabi-sabi (侘寂) is a world view centered on the acceptance of transience and imperfection.

hmaarrfk:
I'm glad to take some simplifications upstream so long as breaking changes to the tests can be justified.

Also, I'm hoping that performance doesn't get affected too much by such heavy handed use of decorators.

I'm also happy to
1. Take the simplifications upstream
2. To give an unrestricted license to skimage for versioning the code (this would allow you to remove the license from the comments)

Preferably in that order in order to retain the ability to relicense wabisabi later on. If versionning is the way you want to move forward (which I recommend), feel free to issue the PR and CC me so that I can formally agree to giving license over to skimage on that PR.

Please open an issue on deprecation-factory for the name change, I'm hoping it won't happen more than once. I won't maintain the old names, hence why I recommend versionning.

hmaarrfk:
@jni, @alexdesiqueira this is the demo that I was talking about regarding how the decorator can be used to basically make the change "once". 

In version 1.0, you are left with "zombie" decorators that would basically do nothing.

https://github.com/hmaarrfk/wabisabi/blob/master/wabisabi/kwonly_change.py#L89

Issue #3228, opened 2018-06-27
Failures in --pre builds


hmaarrfk:
This first showed up in the windows builds failing.
this was due to to the fact that the --pre flag was enabled by default.

PR #3222 attempts to make this more obvious by dedicating a --pre build to the build matrix.

I have a hunch it is due to numpy 1.15.rc1

On travis this wasn't showing up for some reason. I triggered a build on my fork of master (updated) and the dedicated --pre build fails to pull in the numpy 1.15.rc1 dependency from upstream https://travis-ci.org/hmaarrfk/scikit-image/jobs/397451046#L861

When this PR builds, #3227, travis will pull in prereleases from pypi too the problem should show up for travis too.
I've found that while the --pre build compiles with numpy 1.15.rc1, it fails some tests on linux.

stefanv:
Travis-CI does fail against the latest pre-release!  See https://github.com/scikit-image/scikit-image/pull/3204

hmaarrfk:
Good catch. Haven't had a chance to keep up with the PRs lately.
I think the common issue that #3204 and #3227 address is that the following line is executed without the `$PIP_FLAGS`
https://github.com/scikit-image/scikit-image/blob/master/tools/travis/before_install.sh#L63

hmaarrfk:
@stefanv, how would you feel about allowing failures in the `--pre` builds?

I have a few branches that would show how it looks like on my fork:
  - Allow failures in travis: https://github.com/hmaarrfk/scikit-image/pull/4
  - Allow failures in appveyor: https://github.com/hmaarrfk/scikit-image/pull/3

The idea would be that contributors can see that their builds are actually passing on supported machines, while the more advanced devs and maintainers can check to see if scikit-image is working on builds such as `--pre` without having everything look red all the time.

I don't like the fact that the color is "green", I would have preferred a colour such as "blue", but then again, maybe that isn't ideal either.

jni:
@hmaarrfk whoa, very cool!

I'm *not* sure how I feel about it, but it's a tentative 👍 from me. It makes a lot of sense to not allow other projects to change our status to red. It just means one more tick box for reviewers in the PR checklist.

stefanv:
If we allow these failures, I am pretty sure no one will ever look at them.

Perhaps we should only do the --pre installs on our nightly builds, and have those emails go out to a wider audience?

hmaarrfk:
The nightly builds seem like an appropriate place for them.

[...]

PR #3378, opened 2018-08-28
Speed up inner loop of GLCM

### LINKED ISSUE #3377: Performance improvement for `scikit-image/skimage/feature/_texture.pyx`
## Description
When calculating GLCM using the _glcm_loop, the offset pixel calculation can be optimized slightly.
In lines 68-69 the pixel offset value is calculated per each element, those offset values are constant and require both the calculation of a trigonometric function and casting to a different variable type.
By calculating the rowOffset and columnOffset ouside of the loop those executions need only happen once.
                        row = r + <int>round(sin(angle) * distance)
                        col = c + <int>round(cos(angle) * distance)

Some further optimization can be done by using the offset size(s) to reduce the number of iterations by changing the range of the for loops in lines 63 and 64.
In my opinion it would be wiser to change the convention from distance and angle to  rowOffset and columnOffset, but that does not conform with the standard notations.

jni:
## Description
As suggested by @abTest-123 in #3377, pulling some computations and comparisons outside of the inner loop of the gray level co-occurrence matrix results in a significant speedup. In my machine, the included benchmark runs in 9.6ms with the new code, vs 14.7ms on master!

## Checklist
  existing benchmark
- [x] ~~Unit tests~~

## References
Closes #3377 

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_feature.py
+++ b/benchmarks/benchmark_feature.py
@@ -2,7 +2,7 @@
 # http://asv.readthedocs.io/en/latest/writing_benchmarks.html
 import numpy as np
 from scipy import ndimage as ndi
-from skimage import feature
+from skimage import feature, util
 
 
 class FeatureSuite:
@@ -15,5 +15,11 @@ class FeatureSuite:
         self.image = ndi.gaussian_filter(self.image, 4)
         self.image += 0.2 * np.random.random(self.image.shape)
 
+        self.image_ubyte = util.img_as_ubyte(np.clip(self.image, 0, 1))
+
     def time_canny(self):
         result = feature.canny(self.image)
+
+    def time_glcm(self):
+        pi = np.pi
+        result = feature.greycomatrix(self.image_ubyte, [1, 2], [0, pi/4, pi/2, 3*pi/4])
--- a/benchmarks/benchmark_feature.py
+++ b/benchmarks/benchmark_feature.py
@@ -22,4 +22,5 @@ class FeatureSuite:
 
     def time_glcm(self):
         pi = np.pi
-        result = feature.greycomatrix(self.image_ubyte, [1, 2], [0, pi/4, pi/2, 3*pi/4])
+        result = feature.greycomatrix(self.image_ubyte, distances=[1, 2],
+                                      angles=[0, pi/4, pi/2, 3*pi/4])
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,4 +1,6 @@
-pytest
+# pytest 3.7.3 is broken on macOS, see:
+# https://github.com/pytest-dev/pytest/issues/3888
+pytest!=3.7.3
 pytest-cov
 flake8
 codecov
--- a/skimage/feature/_texture.pyx
+++ b/skimage/feature/_texture.pyx
@@ -48,7 +48,8 @@ def _glcm_loop(any_int[:, ::1] image, double[:] distances,
     """
 
     cdef:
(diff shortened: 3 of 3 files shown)


jni:
By the way, while developing this I learned about the command `asv dev`, which is super useful: it runs the specified benchmarks in the current environment. (Rather than doing the full build matrix.)

jni:
Just fyi, I think the remaining PEP8 issues should be ignored, as the angles list is more readable this way.

hmaarrfk:
Does it make sense to have a separate PR for the benchmark to have this speed up visible in the history?

jni:
I'm working on a solution for this problem. I don't want to have to break every PR into two, though, so we'll punt on this until that's in place. It's possible to build a history retroactively, so we can do that periodically.

hmaarrfk:
Nice. I didn’t know we could build a history retroactively. That makes things much easier.

jni:
What the... Anyone know what is going on here?

https://travis-ci.org/scikit-image/scikit-image/jobs/421352873#L5050

Pytest 3.7.3 was released yesterday, which is very suspicious.

jni:
... Yup: https://github.com/pytest-dev/pytest/issues/3888

Options: downgrade/pin pytest, or wait for the fix.

hmaarrfk:
Can you exclude versions in the requirements file?

jni:
Looks like it!

```
 $ cat excl-requirements.txt 
numpy!=1.15.1

 $ pip install -r excl-requirements.txt 
Collecting numpy!=1.15.1 (from -r excl-requirements.txt (line 1))
  Using cached https://files.pythonhosted.org/packages/88/29/f4c845648ed23264e986cdc5fbab5f8eace1be5e62144ef69ccc7189461d/numpy-1.15.0-cp36-cp36m-manylinux1_x86_64.whl
mkl-random 1.0.1 requires cython, which is not installed.
mkl-fft 1.0.2 requires cython, which is not installed.
distributed 1.22.0 requires msgpack, which is not installed.
Installing collected packages: numpy
  Found existing installation: numpy 1.15.1
    Uninstalling numpy-1.15.1:
      Successfully uninstalled numpy-1.15.1
Successfully installed numpy-1.15.0
You are using pip version 10.0.1, however version 18.0 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

PR incoming!

jni:
Actually I just appended it to this one to avoid a rebase. Hopefully it gets merged quickly... ;)

jni:
*sigh* now what the hell is wrong with appveyor? This "morning project" is getting out of hand. LOL

hmaarrfk:
It found matplotlib 3!!!

hmaarrfk:
xref: PR #3222 <- only pull in --pre on one single Windows Build.
xref: https://github.com/scikit-image/scikit-image/issues/3228#issuecomment-400823864 <- allowing --pre failures

soupault:
>It found matplotlib 3!!!

Awesome news! Need to buy some matplotlib shares :smiley:.

abTest-123:
Not sure why - but it seems that the [pyx file](https://github.com/scikit-image/scikit-image/blob/da11c23ed9c2b5db146ffb96b5d05e4cf5206307/skimage/feature/_texture.pyx) hasn't updated. Not sure why/how to see the updated version?

soupault:
@abTest-123 you were looking at the old version (see `blob/...` in your URL). The latest is here - https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/_texture.pyx .

PR #2311, opened 2016-09-19
Corrections proposed + improvements.

alexandrejaguar:
## Description
Adding the function `threshold_multiotsu()`. This generates multiple thresholds for an input image, based on the Multi-Otsu implementation by Liao and Chung.

## Checklist
[It's fine to submit PRs which are a work in progress! But before they are merged, all PRs should provide:]
- [X] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
- [X] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [X] Gallery example in `./doc/examples` (new features only)
- [ ] Unit tests

## References
[1] Liao, P-S., Chen, T-S. and Chung, P-C., "A fast algorithm for multilevel thresholding", Journal of Information Science and Engineering 17 (5): 713-727, 2001.
[2] Tosa, Y., "Multi-Otsu Threshold", a java plugin for ImageJ. Available at: http://imagej.net/plugins/download/Multi_OtsuThreshold.java

PR #3385, opened 2018-09-01
Adding `threshold_multiotsu` (again).

alexandrejaguar:
A followup to #2311.


doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,77 @@
...
+image = data.camera()
+
+# Applying multi-Otsu threshold for the default value, generating
+# three classes.
+thresh = threshold_multiotsu(image)
+
+# Using the values on thresh, we generate the three regions.
+region1 = image <= thresh[0]
+region2 = (image > thresh[0]) & (image <= thresh[1])
+region3 = image > thresh[1]
```
jni: This might be a good place to use [`numpy.digitize`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.digitize.html). Then display the result with `color.label2rgb`. Maybe. =)

doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,77 @@
...
+Multi-Otsu threshold is used to separate the pixels of an input image
+into three different classes, each one obtained according to the
+intensity of the gray levels within the image.
+
+Multi-Otsu calculates several thresholds, determined by the number of desired
+classes. The default number of classes is 3: for obtaining three classes, the
+algorithm returns two threshold values. They are represented by a red line in
+the histogram below.
+
+.. [1] Liao, P-S. and Chung, P-C., "A fast algorithm for multilevel
```
sciunto: It looks like than an author is missing: http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.85.3669

doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,77 @@
...
+thresholding", Journal of Information Science and Engineering 17 (5):
+713-727, 2001.
+"""
+
+from skimage import data
```
sciunto: Usually, we import skimage stuffs after external libraries.

doc/examples/segmentation/plot_multiotsu.py
sciunto: Please, add the link as in the function's docstring (as I couldn't find a DOI...)

skimage/filters/thresholding.py
```diff
@@ -725,6 +726,146 @@ def threshold_triangle(image, nbins=256):
...
+    ----------
+    .. [1] Liao, P-S., Chen, T-S. and Chung, P-C., "A fast algorithm for
+    multilevel thresholding", Journal of Information Science and
+    Engineering 17 (5): 713-727, 2001. Available at:
+    http://www.iis.sinica.edu.tw/page/jise/2001/200109_01.html
+    .. [2] Tosa, Y., "Multi-Otsu Threshold", a java plugin for ImageJ.
+    Available at:
+    http://imagej.net/plugins/download/Multi_OtsuThreshold.java
+
+    >>> image = data.camera()
```
sciunto: from skimage import data :)

alexandrejaguar: Thanks! I missed that :)

doc/examples/segmentation/plot_multiotsu.py
alexandrejaguar: OK.

doc/examples/segmentation/plot_multiotsu.py
alexandrejaguar: It doesn't have DOI (again) :) I prioritized the journal link itself.

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,154 @@ def threshold_triangle(image, nbins=256):
...
+    -------
+    idx_thresh : (nclass) array
+        Array containing the threshold values for the desired classes.
+    max_sigma : float
+        Maximum sigma value achieved on the classes.
+
+    Notes
+    -----
+        The current implementation takes some time to calculate the
+        thresholds for five classes.
```
sciunto: I'm not sure to get this note. Why five classes specifically? Also, "some time" sounds quite vague. Perhaps, a better comment would be to indicate the computation time cost of this function?

alexandrejaguar: I was discussing this with @emmanuelle today. This implementation is not that efficient yet; I thought it would be nice to explain that, but I'm not sure the right way to... need some time to figure the computational cost. What about removing the note? I'll come back to this implementation soon.

alexandrejaguar:
OK, @sciunto @jni now it's good to go. Only need your comments in that. Thanks @emmanuelle for all the support!

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,149 @@ def threshold_triangle(image, nbins=256):
     return bin_centers[arg_level]
 
 
+def threshold_multiotsu(image, nclass=3, nbins=255):
+    """Generates multiple thresholds for an input image. Based on the
```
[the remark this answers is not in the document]
alexandrejaguar: OK

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,149 @@ def threshold_triangle(image, nbins=256):
...
+    >>> from skimage.color import label2rgb
+    >>> from skimage import data
+    >>> image = data.camera()
+    >>> thresh, _ = threshold_multiotsu(image)
+    >>> regions = np.digitize(image, bins=thresh)
+    >>> sep_regions = label2rgb(regions)
+    """
+    if image.shape[-1] in (3, 4):
+        raise TypeError("The input image seems to be RGB (shape: {0}. Please"
+                        "use a grayscale image.".format(image.shape))
```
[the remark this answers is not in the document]
alexandrejaguar: OK. Deleted that comparison.

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,149 @@ def threshold_triangle(image, nbins=256):
...
+
+    if image.min() == image.max():
+        raise TypeError("The input image seems to have only one color: {0}."
+                        "Please use a grayscale image.".format(image.min()))
```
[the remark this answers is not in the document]
alexandrejaguar: OK. Deleted as well.

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,149 @@ def threshold_triangle(image, nbins=256):
...
+           <http://imagej.net/plugins/download/Multi_OtsuThreshold.java>
+
+    Examples
+    --------
+    >>> from skimage.color import label2rgb
+    >>> from skimage import data
+    >>> image = data.camera()
+    >>> thresh, _ = threshold_multiotsu(image)
+    >>> regions = np.digitize(image, bins=thresh)
+    >>> sep_regions = label2rgb(regions)
```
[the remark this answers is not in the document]
alexandrejaguar: OK

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,149 @@ def threshold_triangle(image, nbins=256):
...
+    for u in range(1, nbins):
+        for v in range(u+1, nbins):
+            if (momP[u, v] != 0):
+                var_btwcls[u, v] = momS[u, v]**2 / momP[u, v]
+            else:
+                var_btwcls[u, v] = 0
+
+    # finding max threshold candidates, depending on nclass.
+    # number of thresholds is equal to number of classes - 1.
+    if nclass == 2:
```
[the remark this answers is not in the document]
alexandrejaguar: This would be pretty good, but I actually have no idea on how to walk within the `(idx[k]+1, bins-classes+k)` sets using `combinations()`. Could you clarify a little bit on how I could do that?

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,149 @@ def threshold_triangle(image, nbins=256):
...
+    if nclass not in np.array((2, 3, 4, 5)):
+        raise ValueError("Please choose a number of classes between "
+                         "2 and 5.")
+
+    # receiving minimum and maximum values for the image type.
+    type_min, type_max = dtype_limits(image)
+
+    # calculating the histogram and the probability of each gray level.
+    hist, _ = np.histogram(image.ravel(), bins=nbins,
+                           range=(type_min, type_max))
```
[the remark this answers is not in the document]
alexandrejaguar: That sounds good as well. So, I would use the `bin_edges` returned by `numpy.histogram()` instead of defining them myself, right?

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,149 @@ def threshold_triangle(image, nbins=256):
...
+
+    Parameters
+    ----------
+    image : (N, M) ndarray
+        Grayscale input image.
+    nclass : int, optional
+        Number of classes to be thresholded, i.e. the number of resulting
+        regions. Accepts an integer from 2 to 5. Default is 3.
+    nbins : int, optional
+        Number of bins used to calculate the histogram. Default is 255.
```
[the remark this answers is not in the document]
alexandrejaguar: OK

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,149 @@ def threshold_triangle(image, nbins=256):
...
+
+    Returns
+    -------
+    idx_thresh : (nclass) array
+        Array containing the threshold values for the desired classes.
+    max_sigma : float
+        Maximum sigma value achieved on the classes.
```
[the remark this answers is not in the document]
alexandrejaguar: Opted for (b).

skimage/filters/thresholding.py
```diff
@@ -763,111 +761,98 @@ def threshold_multiotsu(image, nclass=3, nbins=255):
...
     """
-    if image.shape[-1] in (3, 4):
-        raise TypeError("The input image seems to be RGB (shape: {0}. Please"
-                        "use a grayscale image.".format(image.shape))
-
-    if image.min() == image.max():
-        raise TypeError("The input image seems to have only one color: {0}."
-                        "Please use a grayscale image.".format(image.min()))
-
-    # check if nclass is between 2 and 5.
```
sciunto: the check of the correctness of nclass must be kept I think

jni: No, I don't want to merge this until nclass can be an arbitrary number until the number of bins.

sciunto: ok, I didn't get that. Sorry

jni:
@alexandrejaguar Something like this:

```python
from itertools import combinations
...
for idxs in combinations(bins, nclasses):
    limits = (0,) + idxs + (bins - 1,)
    part_sigma = sum(var_btwcls[i+1, j] for i, j in zip(limits, limits[1:]))
    if max_sigma < part_sigma:
        aux_thresh = idxs
        max_sigma = part_sigma
```

I haven't tested this but I think that should be close...

btw, should the iteration start at 0 instead of 1? ie is `var_btwcls[0, idx1]` not a sensical term?

I realised that I don't understand how `var_btwcls` works, so some comments explaining that part of the implementation would be useful.

Thanks, I hope that's readable/useful!

jni:
Sorry, that shouldn't be `combinations(bins, nclasses)` but `combinations(range(bins), nclasses)`, and you might need to adjust the start and end of the range.

skimage/filters/thresholding.py
jni: ... Having said that, once that is implemented, I certainly wouldn't be opposed to a check that nclasses is strictly below the number of bins, once that's done. =)

skimage/filters/thresholding.py
```diff
@@ -726,6 +727,134 @@ def threshold_triangle(image, nbins=256):
...
+    # receiving minimum and maximum values for the image type.
+    type_min, type_max = dtype_limits(image)
+
+    # calculating the histogram and the probability of each gray level.
+    hist, _ = np.histogram(image.ravel(), bins=bins,
+                           range=(type_min, type_max))
+    prob = hist / image.size
+
+    max_sigma = 0
+    momP, momS, var_btwcls = [np.zeros((bins, bins)) for n in range(3)]
```
emmanuelle: a comment explaining the meaning of these variables would be welcome

emmanuelle:
For 4 and 5 classes it might be worth cythonizing the loops. Please tell us if you feel comfortable with this or not (we can help!).

skimage/filters/tests/test_thresholding.py
```diff
@@ -426,3 +428,22 @@ def test_niblack_sauvola_pathological_image():
...
+
+
+def test_check_multiotsu_results():
+    image = 0.25 * np.array([[0, 1, 2, 3, 4],
+                             [0, 1, 2, 3, 4],
+                             [0, 1, 2, 3, 4],
+                             [0, 1, 2, 3, 4]])
+
+    for idx in range(3, 6):
+        thr_multi, _ = threshold_multiotsu(image,
```
emmanuelle: it's ``thr_multi = ....`` now that the function only returns the thresholds (this causes Travis to fail)

emmanuelle:
@jni  even with your example I don't understand how you would use ``combinations``. The nested for loops are cumbersome but at least they are quite readable.

jni:
@emmanuelle why is it confusing? `combinations(seq, n)` returns all possible (sorted) combinations of n elements in `seq`. This is *exactly* what you are doing in the for loops, expressed in a word. "For all combinations of 5 bin edges (thresholds), compute the total between-class variance, and keep the combination that maximises that." I think the itertools version makes that much clearer. (Though perhaps a comment to clarify the sum is warranted.)

alexandrejaguar:
@emmanuelle @jni I'm trying to implement using `combinations()` and it's slower than the original, even for 3 thresholds (talking about minutes more)... maybe cython could be a way out of it, but the results of the new function are also not the same. I'll take a while to understand what's happening and to adapt the code.

skimage/filters/tests/test_thresholding.py
alexandrejaguar: Thanks! Didn't see that 🤔😳

skimage/filters/thresholding.py
alexandrejaguar: `momP` and `momS` are moments of something :laughing: They are tables used for calculating the variances between classes (`var_btwcls` itself). I'll re-read the paper to make better comments on that.

jni:
@alexandrejaguar I think you might be iterating over fewer elements in your implementation, because `range` doesn't include the last index. Here is a simple test that shows `itertools.combinations` and the nested for-loops are equivalent, and that itertools is even a tiny bit faster:

`comb.py`:
```python
import itertools

def comb3(n):
    for i in range(1, n-3):
        for j in range(i+1, n-2):
            for k in range(j+1, n-1):
                yield i, j, k

def comb3_iter(n):
    yield from itertools.combinations(range(1, n-1), 3)
```

```python
In [1]: import comb

In [2]: %timeit list(comb.comb3(255))
247 ms ± 3.22 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit list(comb.comb3_iter(255))
224 ms ± 2.01 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: l1 = comb.comb3(255)

In [5]: l1 = list(comb.comb3(255))

In [6]: len(l1)
Out[6]: 2667126

In [7]: l2 = list(comb.comb3_iter(255))

In [8]: len(l2)
Out[8]: 2667126

In [9]: set(l1) == set(l2)
Out[9]: True
```

So whatever the source of the slowdown, it's not itertools...

emmanuelle:
Hi @alexandrejaguar I opened a pull request to your branch (I have the impression that I cannot push changes to your branch). https://github.com/alexandrejaguar/scikit-image/pull/2

I implemented a cython version, which is much faster. With the camera image, it takes less than one second even with 5 classes. Please merge if you're happy, so that we can discuss these changes here.

emmanuelle:
Hi @alexandrejaguar did you find the same to look at my PR?

sciunto:
In contrast to @jni's opinion, I would vote for a merge even if the current code can't detect more than 3 thresholds and leave the generalization to a future work. There are many situations where this level is already satisfying.

emmanuelle:
@sciunto I proposed in https://github.com/alexandrejaguar/scikit-image/pull/2 a cython implementation which is much faster, even with 5 classes. At the moment the main obstacle is the aesthetic validation from @jni :-p. @jni is in charge of improving the speed of a more elegant but slower cython version, we know in a few days if it's feasible or not. I'm also in favor of merging this feature ASAP!

jni:
LOL I am totally feeling the heat of the spotlight now. :joy: Note that my objections would be quite diminished for 3 classes only. And it is not just aesthetic: in that PR, @alexandrejaguar proposed that it might be possible to change the way the recurrence lists and variances are computed. The multi-nested version of the code might then require many changes in response, whereas a single loop would be much easier to maintain. Beauty and practicality are not necessarily opposed!

alexandrejaguar:
@jni now, I'm more in favor of "if isn't broken, don't fix it" :laughing: But I understand the functional part of beautiful code as well. For me, I'd like it to be as fast as possible :sweat_smile:

emmanuelle:
@stefanv I completely agree that this code smells bad, but at the moment I don't know how to start to improve it. Would love any hint for a solution which I could try...

stefanv:
I know it's easy to shout suggestions from the sideline.  But... I also know that once this goes in, we are unlikely to look at it again for a while.  So, I am hoping that we'll think as hard as we can, and if we cannot find a better solution, well, then so be it and we'll move ahead for sake of practicality.  I am not claiming I have a golden insight, so take my encouragement here as exactly that, and not criticism!

alexandrejaguar:
Hi everyone,
this is where we are. I'm having some issues trying to implement the `itertools` version.
I can't generate the intervals we're generating using `for`. These should be `[0, bins-classes]`, `[1, bins-classes+1]`, `[2, bins-classes+2]`, until `classes-1`. I sketched several different codes, and no one makes it properly...  one example follows. I think I should iterate the `-1` in `limits` somehow, but I worked on that so much I'm not seeing things properly. 

```
for idx in combinations(range(bins), classes):
    limits = (0,) + idx + (bins - 1,)
    part_sigma = sum(var_btwcls[i+1, j] for i, j in zip(limits, limits[1:]))
    if max_sigma < part_sigma:
        aux_thresh = idx
        max_sigma = part_sigma
```

Any ideas? :blush:

alexandrejaguar:
Hi everyone,
@jni I got some profiling results. After checked the entire code, I've narrowed it to this part:
```
    # finding max threshold candidates, depending on classes.
    # number of thresholds is equal to number of classes - 1.
    for idx in combinations(range(bins-1), classes-1):
        limits = (0,) + idx + (bins - classes + 1,)
        part_sigma = sum(var_btwcls[i+1, j] for i, j in zip(limits,
                                                            limits[1:]))
        if max_sigma < part_sigma:
            aux_thresh = idx
            max_sigma = part_sigma
```

The issue here is how we obtain `part_sigma`. For 5 classes, we have:

```
jaguar@spaghetti:~$ python threshold_multiotsu.py
* Thresholds:  [ 40.  93. 138. 168.]
         1185537508 function calls in 347.498 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
169362501  162.671    0.000  347.498    0.000 {built-in method builtins.sum}
1016175006  184.827    0.000  184.827    0.000 threshold_multiotsu.py:114(<genexpr>)
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

I'll think about it a little bit, but feel free to shout it you have any ideas.

jni:
@alexandrejaguar sorry that this took so long! I'm glad GitHub doesn't lose the code after you delete the branch because I'd like to pick this up at some point!

Issue #3384, opened 2018-08-31
Missing scientific (domain-specific) images and corresponding examples and tutorials


emmanuelle:
I think scientific users could be more interested in starting using scikit-image if our documentation showed more examples corresponding to applications in different scientific domains. This could include cell images, satellite / remote sensing images, hyperspectral images, materials science images, astronomy images, etc. We could ask users to contribute some open data and examples, but before doing that is there a consensus that it would be useful?

sciunto:
I think this is extremely valuable. However, we have the size issue... A discussion started here: #3323

emmanuelle:
@sciunto There is the size issue indeed but it seems that it is due primarily to cython files; moreover the largest data files are the astronaut HOG data which are used only for testing, maybe it's possible to decrease their size by using a subimage.

jni:
@emmanuelle just last week I gave a talk about scikit-image, and as I was browsing the gallery for useful examples, I noted this exact problem: we have a nice gallery of examples that feel too artificial to compel scientists. I ended up using the images from your blog posts. =) So, a big :+1: from me. We need to solve the scientific examples issue, and @sciunto if that means that we need to solve the size issue, then we need to solve that issue also. =)

Since we are considering moving data out of the core package and have it download on demand, it might be a good opportunity to revisit our example data also.

emmanuelle:
We might use images from Wikipedia. I was quickly browsing through wikipedia microscopy images, 
https://en.wikipedia.org/wiki/Agranulocyte#/media/File:Monocyte_no_vacuoles.JPG
https://commons.wikimedia.org/wiki/File:Hairy_cell_leukemia_smear_2009-08-20.JPG
https://commons.wikimedia.org/wiki/File:Zz_Plagiomnium_undulatum_fluorescence.jpg
https://commons.wikimedia.org/wiki/File:NdFeB-Domains.jpg
Is CC-by-SA (needs attribution) license OK, or do we need public domain images?

sciunto:
So far, we used only CC0 / public domain images, and If I'm correct, that was a requirement made by @stefanv I guess for the compatibility with the BSD license.

stefanv:
Two issues arise: 1) how to communicate such requirements with users 
effectively and 2) do we want to place potentially surprising restrictions 
in place. For number 2, we've argued strongly "no": you should be able to 
use the output of scikit-image without breaching a patent or attribution 
requirement.

hmaarrfk:
[...]

PR #3388, opened 2018-09-03
Show our data in the gallery

sciunto:
## Description

To pursue on the idea in #3384 I thought that it would be useful to show the data we are shipping with scikit-image. We could separate images for general purposes, scientific images and other specific images (like stereo).

## Checklist
  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


doc/examples/data/plot_general.py
```diff
@@ -0,0 +1,42 @@
...
+           'page',
+           'text',
+           'rocket',
+           )
+
+#'stereo_motorcycle'
+#'lfw_subset',
+
+fig, axes = plt.subplots(len(images), 1, figsize=(8, 4 * len(images)),
+                         sharex=False, sharey=False)
```
emmanuelle: isn't ``sharex/y`` False by default?

doc/examples/data/plot_general.py
```diff
@@ -0,0 +1,42 @@
...
+ax = axes.ravel()
+
+for i, image in enumerate(images):
+    caller = getattr(data, image)
+    ax[i].imshow(caller())
+    ax[i].set_title(image)
```
emmanuelle: ``ax[i].axis('off')```to gain some space in a crowded figure?

doc/examples/data/plot_general.py
```diff
@@ -0,0 +1,42 @@
+"""
+======================
+General purpose images
```
emmanuelle: General-purpose

sciunto:
You are absolutely right for the thumbnail. It's an issue. The idea I got in the mean time to solve this is to take one image as an example for the thumbnail, and then display them all. I'm afraid that, even if we succeed in generating a correct rendering, the future additions will destroy it anyway.

emmanuelle:
@sciunto you can do that by displaying a first image in the example, and then the whole subplots. I don't remember if sphinx-gallery takes the first or the last figure for the thumbnail, and whether it can be configured. The doc of sphinx-gallery is quite thorough, but its organization is sometimes a bit weird.

sciunto:
It's now ready for reviews.

sciunto:
@scikit-image/core Anyone for reviewing this please?

sciunto:
@soupault I increased the font size. Regarding the fact it doesn't show up in section, I guess it's because we use getattr, and not a direct call. A parser cannot get the call here... but i don't think it's significant miss.

doc/examples/data/plot_specific.py
```diff
@@ -0,0 +1,46 @@
...
+===============
+
+"""
+import numpy as np
+import matplotlib.pyplot as plt
+import matplotlib
+
+from skimage import data
+
+matplotlib.rcParams['font.size'] = 18
```
soupault: I've just noticed that you don't set titles in this example. The font config doesn't really make sense then.

soupault:
> I guess it's because we use getattr, and not a direct call
@sciunto Ah, you are right. That's, most likely, the case.

doc/examples/data/plot_specific.py
soupault: Other examples look great!

sciunto: It just to keep the tics homogeneous...

soupault:
Why don't we add the titles here also then? :)

sciunto:
Because we have a header for each image/set of images  that already introduces the content... I can add it, but I did not feel it's necessary.

emmanuelle:
@sciunto I run the example after merging your branch from master, and I got an error because of the ``lbp_frontal_face_cascade_filename`` which was added. A more robust way to exclude datasets could be 
```
exclude = ['data_dir', 'load', 'lbp_frontal_face_cascade_filename']

datasets = data.__all__ # remove load and data_dir
datasets = [dataset for dataset in datasets if not dataset in exclude]

```

sciunto:
@emmanuelle I'm confused. Where do I call `data.__all__`? I couldn't reproduce.

emmanuelle:
@sciunto sorry I got mixed up (with one of my local files). Sorry about that, ignore my comment!

emmanuelle:
LGTM. It could be good to order the sections of the gallery so that the section on data appears on top (as described here https://sphinx-gallery.readthedocs.io/en/stable/advanced_configuration.html?highlight=sorting#sub-gallery-order). But we can also merge this PR first and do the ordering later, depending on your time.

sciunto:
Good point. I open another issue for that. I think that ordering our examples would be useful beyond this PR (eg operation on numpy arrays could come among the first ones).

emmanuelle:
Merging, thanks!

PR #3394, opened 2018-09-08
Better behavior of Gaussian filter for arrays with a large number of dimensions

emmanuelle:
Follow-up of #2881


skimage/filters/_gaussian.py
```diff
@@ -110,7 +110,11 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
 
     """
 
-    spatial_dims = guess_spatial_dimensions(image)
+    spatial_dims = None
+    try:
+        spatial_dims = guess_spatial_dimensions(image)
```
soupault: I'd actually put the full code of `guess_spatial_dimensions` here. `gaussian` seems to be one of the very few places where this function is used (see https://github.com/scikit-image/scikit-image/search?q=guess_spatial_dimensions), and we decided to move towards explicit dimension handling (having the full code here will allow, occasionally, to easily refactor the behaviour).

jni: Wow, you're right, it's *the only* place where it's used! Do we want to deprecate it?

In general, though, I would leave that for another PR that also deprecates the magic behaviour here. One of my guiding principles when writing code is that no function should be longer than 25 lines, so that it can all be taken in in one look. (This happens to be in NASA's coding practices for bug-free code.) So I like the pattern of abstracting away input validation.

But, since this function is basically *only* input validation before calling out to `ndimage`, it could all live in one place.

emmanuelle:
I opened #3396 so that we do the deprecation as suggested by @soupault and @jni

PR #3398, opened 2018-09-10
Allow import/execution with -OO

### LINKED ISSUE #3392: TypeError when running in Python optimized mode
## Description
I'm investigating running a project, which use scikit-image, with python optimized mode.
The problem happens because the `_parse_docs()` function try to regex parse a docstring which is None.
I already fixed it on my side and can provide the patch ..

## Way to reproduce
scikit_image ==  0.14.0

`python2.7 -OO -c "import skimage.measure"`

gst:
## Description
Allow import/execution with `python -OO`

Closes #3392

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ matrix:
       python: 3.7
       dist: xenial # Required for Python 3.7
       sudo: true   # travis-ci/travis-ci#9069
-      env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 BUILD_DOCS=1 PYTHONOPTIMIZE=2
+      env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 BUILD_DOCS=1
     - os: linux
       python: 3.6
       env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 PIP_FLAGS="--pre"
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ matrix:
       python: 3.7
       dist: xenial # Required for Python 3.7
       sudo: true   # travis-ci/travis-ci#9069
-      env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 BUILD_DOCS=1
+      env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 BUILD_DOCS=1 PYTHONOPTIMIZE=2
     - os: linux
       python: 3.6
       env: QT=PyQt5 WITH_PYAMG=1 OPTIONAL_DEPS=1 PIP_FLAGS="--pre"
--- a/skimage/measure/_regionprops.py
+++ b/skimage/measure/_regionprops.py
@@ -631,7 +631,7 @@ def _parse_docs():
     import re
     import textwrap
 
-    doc = regionprops.__doc__
+    doc = regionprops.__doc__ or ''
     matches = re.finditer('\*\*(\w+)\*\* \:.*?\n(.*?)(?=\n    [\*\S]+)',
                           doc, flags=re.DOTALL)
     prop_doc = dict((m.group(1), textwrap.dedent(m.group(2))) for m in matches)
@@ -648,4 +648,6 @@ def _install_properties_docs():
         setattr(_RegionProperties, p, property(getattr(_RegionProperties, p)))
 
 
-_install_properties_docs()
+if __debug__:
+    # don't install docstrings when in optimized/non-debug mode
+    _install_properties_docs()
--- a/skimage/measure/tests/test_regionprops.py
+++ b/skimage/measure/tests/test_regionprops.py
@@ -476,6 +476,11 @@ def test_cache():
 
 
 def test_docstrings_and_props():
+    def foo():
+        """foo"""
+
+    has_docstrings = bool(foo.__doc__)
+
     region = regionprops(SAMPLE)[0]
 
     docs = _parse_docs()
@@ -483,11 +488,13 @@ def test_docstrings_and_props():
 
     nr_docs_parsed = len(docs)
     nr_props = len(props)
-    assert_equal(nr_docs_parsed, nr_props)
-
-    ds = docs['weighted_moments_normalized']
-    assert 'iteration' not in ds
-    assert len(ds.split('\n')) > 3
+    if has_docstrings:
+        assert_equal(nr_docs_parsed, nr_props)
+        ds = docs['weighted_moments_normalized']
+        assert 'iteration' not in ds
+        assert len(ds.split('\n')) > 3
+    else:
+        assert_equal(nr_docs_parsed, 0)
 
 
 def test_incorrect_coordinate_convention():
--- a/TODO.txt
+++ b/TODO.txt
@@ -47,4 +47,4 @@ Other
 * Monitor SimpleITK Pypi releases to find when a 3.7 release comes out.
   Remove version restriction in requirements file when appropriate.
   `https://github.com/SimpleITK/SimpleITK/issues/551`
-* Monitor next Pillow release to have a build with PYTHONOPTIMIZE=2.
+* Monitor next Pillow release to have a build with PYTHONOPTIMIZE=2 (See #3398).
--- a/TODO.txt
+++ b/TODO.txt
@@ -47,3 +47,4 @@ Other
 * Monitor SimpleITK Pypi releases to find when a 3.7 release comes out.
   Remove version restriction in requirements file when appropriate.
   `https://github.com/SimpleITK/SimpleITK/issues/551`
+* Monitor next Pillow release to have a build with PYTHONOPTIMIZE=2.


hmaarrfk:
Recreation of the issue:
```
conda create --name sk27 python=2.7 -y
conda activate sk27
pip install scikit-image
python2.7 -OO -c "import skimage.measure" 
```

```
  File "/home/mark/miniconda3/envs/sk27/lib/python2.7/site-packages/skimage/measure/__init__.py", line 3, in <module>
    from ._marching_cubes_classic import (marching_cubes_classic,
  File "/home/mark/miniconda3/envs/sk27/lib/python2.7/site-packages/skimage/measure/_marching_cubes_classic.py", line 2, in <module>
    import scipy.ndimage as ndi
  File "/home/mark/miniconda3/envs/sk27/lib/python2.7/site-packages/scipy/ndimage/__init__.py", line 161, in <module>
    from .filters import *
  File "/home/mark/miniconda3/envs/sk27/lib/python2.7/site-packages/scipy/ndimage/filters.py", line 38, in <module>
    from . import _ni_docstrings
  File "/home/mark/miniconda3/envs/sk27/lib/python2.7/site-packages/scipy/ndimage/_ni_docstrings.py", line 4, in <module>
    from scipy.misc import doccer
  File "/home/mark/miniconda3/envs/sk27/lib/python2.7/site-packages/scipy/misc/__init__.py", line 100, in <module>
    from .pilutil import *
  File "/home/mark/miniconda3/envs/sk27/lib/python2.7/site-packages/scipy/misc/pilutil.py", line 19, in <module>
    from PIL import Image, ImageFilter
  File "/home/mark/miniconda3/envs/sk27/lib/python2.7/site-packages/PIL/__init__.py", line 27, in <module>
    __doc__ = __doc__.format(__version__)  # include version in docstring
AttributeError: 'NoneType' object has no attribute 'format'
```

skimage/measure/_regionprops.py
```diff
@@ -631,7 +633,7 @@ def _parse_docs():
     import re
     import textwrap
 
-    doc = regionprops.__doc__
+    doc = _regionprops_doc
```
hmaarrfk: `doc = regionsprops.__doc__ if regionprops.__doc__ else ""`

skimage/measure/_regionprops.py
```diff
@@ -358,9 +358,7 @@ def __eq__(self, other):
         return True
 
 
-def regionprops(label_image, intensity_image=None, cache=True,
-                coordinates=None):
-    """Measure properties of labeled image regions.
+_regionprops_doc = """Measure properties of labeled image regions.
```
hmaarrfk: I really don't like seperating the docs from the function names.

skimage/measure/_regionprops.py
```diff
@@ -648,4 +650,5 @@ def _install_properties_docs():
         setattr(_RegionProperties, p, property(getattr(_RegionProperties, p)))
 
 
-_install_properties_docs()
+if __debug__:
```
hmaarrfk: I think with the fix above it should be ok. don't know if you need this.

hmaarrfk: I could be wrong.

hmaarrfk:
```
python3.6 -OO -c "import skimage.measure"                                                                                                                                                                 (sk36) 
  File "/home/mark/miniconda3/envs/sk36/lib/python3.6/site-packages/skimage/measure/__init__.py", line 3, in <module>
    from ._marching_cubes_classic import (marching_cubes_classic,
  File "/home/mark/miniconda3/envs/sk36/lib/python3.6/site-packages/skimage/measure/_marching_cubes_classic.py", line 2, in <module>
    import scipy.ndimage as ndi
  File "/home/mark/miniconda3/envs/sk36/lib/python3.6/site-packages/scipy/ndimage/__init__.py", line 161, in <module>
    from .filters import *
  File "/home/mark/miniconda3/envs/sk36/lib/python3.6/site-packages/scipy/ndimage/filters.py", line 38, in <module>
    from . import _ni_docstrings
  File "/home/mark/miniconda3/envs/sk36/lib/python3.6/site-packages/scipy/ndimage/_ni_docstrings.py", line 4, in <module>
    from scipy.misc import doccer
  File "/home/mark/miniconda3/envs/sk36/lib/python3.6/site-packages/scipy/misc/__init__.py", line 100, in <module>
    from .pilutil import *
  File "/home/mark/miniconda3/envs/sk36/lib/python3.6/site-packages/scipy/misc/pilutil.py", line 19, in <module>
    from PIL import Image, ImageFilter
  File "/home/mark/miniconda3/envs/sk36/lib/python3.6/site-packages/PIL/__init__.py", line 27, in <module>
    __doc__ = __doc__.format(__version__)  # include version in docstring
AttributeError: 'NoneType' object has no attribute 'format'
```
Seems like it also happens on 3.6

hmaarrfk:
How are you not hitting the PIL issue? do you have a patched version of PILLOW?

gst:
> How are you not hitting the PIL issue? do you have a patched version of PILLOW?
I've effectively workaround it too. I had not encountered it previously as I was on a really old version of it but after upgrade I faced it too.

skimage/measure/_regionprops.py
gst: this makes test_regionprops to fail when run with `python -OO -mpytest`

hmaarrfk: Really you should just `return` early.

Those tests are likely failing because they expect the docstrings to exits. If you want the tests to pass in `-OO` mode, you will need to change their behaviour in `-OO` mode. Instead of having mangled doc strings, the doc strings should be `None` it seems.

hmaarrfk:
i guess the goal of this patch would be:

1. Not to make the codebase look werid.
2. Not to undermine the point of `-OO` by forcing the doc strings to be added.
3. Not to fail tests

gst:
- Not to fail tests

with or without -OO ?

hmaarrfk:
You seem to think that they should pass with `-OO` that seems sensible. You are showing us that there is a usecase for having the code optimized with `-OO`. I'm honestly not even sure what happens with test in optimized mode :/ many of the tests use assert statements. (I think)

A useful strategy is to tell people:
1. please install `pytest`
2. run `pytest --pyargs skimage`

skimage/measure/_regionprops.py
```diff
@@ -627,25 +626,27 @@ def perimeter(image, neighbourhood=4):
...
-        getattr(_RegionProperties, p).__doc__ = prop_doc[p]
-        setattr(_RegionProperties, p, property(getattr(_RegionProperties, p)))
+        for p in [member for member in dir(_RegionProperties)
+                  if not member.startswith('_')]:
+            getattr(_RegionProperties, p).__doc__ = prop_doc[p]
+            setattr(_RegionProperties, p, property(getattr(_RegionProperties, p)))
 
 
-_install_properties_docs()
+    _install_properties_docs()
```
hmaarrfk: If this PR fixes your issue, 
It seems like all you need to do is have this one line in the __debug__.

hmaarrfk: Am I missing something?

gst: nothing missed.

I was hesitating to keep or not keep these 2 related functions in the path of the non-debug mode..

I finally choosed not to keep them but if you prefer to keep the 2 functions defined when they are unused then please acknowledge so.. if you prefer the simpler one liner fix so.

for the test : given by default tests are executed without -OO I've done nothing here.
but it's something pytest handles correctly as far as I see.

hmaarrfk: yeah, so I would remove the indentation. those functions may want to be called otherwise.

hmaarrfk:
@gst, so this helps you out by not executing that function in debug mode?

gst:
> @gst, so this helps you out by not executing that function in debug mode?

in _optimized_ mode, yes, that prevents the regex on the then None __doc__ function attribute.

gst:
By the way the problem is fixed too in master of Pillow/PIL : https://github.com/python-pillow/Pillow/commit/fc8717fb040a135b91be58af3c4faf86d9765516

hmaarrfk:
Great thanks for working on this! I know it is alot of back and forth, but I think we converged on a solution that is both elegant, and solved the issue at hand!

hmaarrfk:
@scikit-image/core this is ready for official review now.

gst:
Thanks for making the follow up.

skimage/measure/_regionprops.py
```diff
@@ -648,4 +648,5 @@ def _install_properties_docs():
         setattr(_RegionProperties, p, property(getattr(_RegionProperties, p)))
 
 
-_install_properties_docs()
+if __debug__:
```
soupault: Wth is that description https://docs.python.org/3.7/library/constants.html#__debug__:
```
This constant is true if Python was not started with an -O option.
```
Means that the default regime is a `debug` one? Does it make any sense to you? 🙂 

Regarding the change: it would be nice to have a short comment (with or without reference) on the necessity of the conditional.

hmaarrfk: yeah... I didn't know about this debug/no debug more either. I guess default python is started in debug mode.

Apparently many software actually crashes if you try running it with `-OO`

hmaarrfk:
@gst, I wonder if you can find a way to add a test to check if we can continue to run scikit-image in debug mode. I see that this might continue to happen if we don't have a test. 

I really like the idea of running all tests with command so ideally this would be integrated in the current `pytest` infrastructure (though I think scikit-image likes to abstract the idioms away). 

maybe something like "execute command python -OO import skimage.module name" in a test. And maybe try and identify all module names programatically?

It won't be perfect, but it would be a good start.

skimage/measure/_regionprops.py
gst: Effectively I think there are code still out there that crashes/misbehaves when executed with -O(O), which is a shame ;)

gst:
> It won't be perfect, but it would be a good start.

well, then why not add this in `tools/travis/script.sh`, after test with min requirements (or something/somewhere like that) : 

`python -OO -m pytest $TEST_ARGS skimage`

?

I don't really like to add test case specifically for this..  and with this you/we ensure that everything is really alright (an import succeed may not be sufficient).

hmaarrfk:
@gst great ideas!

1. The test suit is quite long and already takes for ever to run on my computer. Travis is kinda getting close to hitting the time limit.
2. Changes to the build matrix  are difficult to get passed by the review process. That said, you make a good point. **Maybe** you can add it to the 3.5 build https://github.com/scikit-image/scikit-image/blob/master/.travis.yml#L41
3. I'm mostly worried that somebody will come by in 2 years and refactor `__DEBUG__` away. but maybe they can use `__doc__` as strings assuming all our functions have docstrings in other locations (which would crash debug mode).
4. There is also a chance that we change the build process. Not too sure how environment variables will play into the new one. the testing suite is quite simple "just run pytest"
5. Finally, an other way to fix this is to handle `__doc__ is NULL` gracefully in the function calls. That is much easier to test, but an indirect test (not airtight as you say).

What will you do about PIL being broken in debug mode? Did they get a new release?

Whatever you do, revert your change first, run the test, then revert it back. Hopefully we will see the difference it makes to the test.

Final question: what is your usecase for using optimized python. Is it faster?

gst:
regarding your 5 : I can add something for this.. coming

regarding PIL : as I said it's already fixed in their master. On their next release that should be out.

regarding your final question: for now not really.. but I just found this odd that -OO wasn't handled gracefully. hence my ticket / PR.

gst:
the optimized build fails because of dask but I see it's at version
dask                0.9.0
while latest available version (0.19.1) doesn't suffer the (same kind of) problem.
Don't get/see from where comes this quite old version of dask.. ?

skimage/measure/tests/test_regionprops.py
```diff
@@ -476,18 +476,25 @@ def test_cache():
...
-
-    ds = docs['weighted_moments_normalized']
-    assert 'iteration' not in ds
-    assert len(ds.split('\n')) > 3
+    if has_docstrings:
+        assert_equal(nr_docs_parsed, nr_props)
+        ds = docs['weighted_moments_normalized']
+        assert 'iteration' not in ds
+        assert len(ds.split('\n')) > 3
+    else:
```
hmaarrfk: Why not create bar with no doc string?

skimage/measure/_regionprops.py
```diff
@@ -631,7 +631,7 @@ def _parse_docs():
     import re
     import textwrap
 
-    doc = regionprops.__doc__
+    doc = regionprops.__doc__ or ''
```
hmaarrfk: When calling _install_properties_docs on a null string, does it now behave as a no-op?

hmaarrfk:
dask 0.9.0 is only in 2016.

It is hard to set minimum version requirements. In an ideal world, people would be updated on all fronts, but quite often they can't. Setting dask to be in 2016 (only 2 years old) is kinda like saying:

If you want to upgrade scikit-image, you don't really need to check if your dask code will break.

I'm not sure we want to set an upgraded version of dask just yet.
Maybe when this [PR](https://github.com/dask/dask/pull/3964) gets merged we can set it to 0.19.2 :D
Even then, I would only set that for an example.

skimage/measure/_regionprops.py
gst: effectively. the regex then just returns nothing.

skimage/measure/tests/test_regionprops.py
gst: don't get you, how would you detect that it has been dropped when in -OO then ?

hmaarrfk: Basically -OO doesn’t assign that string to the property .doc. It is the same as simply omitting a docstring.

.travis.yml
```diff
@@ -42,6 +42,9 @@ matrix:
     - os: linux
       python: 3.5
       env: QT=PyQt5 WITH_PYAMG=1 MINIMUM_REQUIREMENTS=1
+    - os: linux
```
stefanv: Can't we rather add this flag to one of the existing matrix builds?  We have to be careful here, because we only have a fixed number of parallel Travis-CI builds available.

hmaarrfk: Oh yea, I was suggesting we add it to the one on line 44 just above it.
Good catch.

gst: just like this ?

skimage/measure/tests/test_regionprops.py
gst: > Basically -OO doesn’t assign that string to the property .doc. It is the same as simply omitting a docstring.

you got it :)

jni:
I'm confused about @stefanv's approval... The Travis failure appears to be exactly related to this PR?

hmaarrfk:
I think stefanv mostly put his seal of approval on the direction this change is going.

gst:
@jni the travis failure is because one of python3.5 build matrix entry is now executed with -OO mode and that trigger the same kind of problem within dask 0.9 that this PR resolves for skimage :)

hmaarrfk:
does `doctest_skip_parser` need to be updated? it splits `__doc__` without checking it

gst:
> does doctest_skip_parser need to be updated? it splits __doc__ without checking it

it probably will need too

jni:
Incidentally, I'd consider moving the PYTHONOPTIMIZE to a different build. MINIMUM_REQUIREMENTS often breaks, so it would be great to not confuse the two tags. Probably I'd put it in the first build, or one of the macOS builds, or 3.7 (now that that's working).

gst:
moved to python3.7 build

hmaarrfk:
@gst sorry for making you do more work. I'm just going through a few of these PRs and I think it might be useful for you if you broke this one into two.

1. With your fix and tests
2. With the addition to travis. We can mark the addition to travis as pending Pilliow 5.3.

Think about what you want.

hmaarrfk:
@soupault it fails tests because Pilow probably won't make a release for 4 months. I am really in favour of having passing tests to have a smooth process for other PRs.

gst:
should I just remove/comment the PYTHONOPTIMIZE for now ?

hmaarrfk:
@gst i think that whould be best, made a note in the TODO.txt to monitor Pillow and add it back when the new version is released

hmaarrfk:
Minimum requirements build is failing because of: https://github.com/scikit-image/scikit-image/pull/3432

sciunto:
I added a reference to this PR in the TODO file. As it is already validated by two of us and i'm fine with this, i merge this PR.

Many thanks to all of you.

PR #3402, opened 2018-09-13
Improve Li thresholding

### LINKED ISSUE #2337: threshold_li fails for "zeros" input images due to undefined `threshold`
## Description

When running `skimage.filters.threshold_li`, if the input image is all 0, the 
`tolerance` is set to zero causing the while loop never to execute and a somewhat cryptic error is
encountered because `threshold` never gets set. 

A typical traceback is 

```
  File "test.py", line 3, in <module>
    threshold_li(np.zeros((100,100)))
  File "./skimage/filters/thresholding.py", line 509, in threshold_li
    return threshold + immin
UnboundLocalError: local variable 'threshold' referenced before assignment
```

I recognise that this is a special case, but I just wanted to raise this issue as a way of
asking 

> do we want this somewhat cryptic error, or should we catch this 
> using a check on the input image"?  
## Way to reproduce

All that's needed to reproduce this (Python 3.4.1, skimage 0.13dev) is: 

```
import numpy as np
threshold_li(np.zeros((100,100)))
```

jni:
- images with np.nan and np.inf values are now processed correctly
- `temp` value becoming negative was impossible, and this line [was not
  covered](https://codecov.io/gh/scikit-image/scikit-image/src/86bbed0937aef504cc1a9837c1a21271cce5bef4/skimage/filters/thresholding.py), so removed it
- code contained numerous fixes for integer offset errors. Since we
  deal with floating point, these are unnecessary. The code is now
  much simpler
- it is also more correct: the tolerance offset that was applied
  previously in fact was giving an incorrect threshold. I verify this
  with a new private function, _cross_entropy, which estimates the
  cross-entropy between foreground and background as defined in the
  original paper.

The only fly in this sweet ointment is that for *one* of our test
images, coins, the threshold found by the iterative method is one off
(or rather 0.something off, but this happens to fall on either side of an
integer) from the optimal threshold. I believe this is an unavoidable
side effect of images being discretized into integer bins.

## Checklist
  existing benchmark
- [x] Unit tests

## References
Fixes #2337 

## For reviewers
  later.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -459,7 +459,10 @@ def threshold_isodata(image, nbins=256, return_all=False):
         return thresholds[0]
 
 
-def _cross_entropy(image, threshold, bins=np.arange(-0.5, 255.51, 1)):
+_DEFAULT_ENTROPY_BINS = tuple(np.arange(-0.5, 255.51, 1))
+
+
+def _cross_entropy(image, threshold, bins=_DEFAULT_ENTROPY_BINS):
     """Compute cross-entropy between distributions above and below a threshold.
 
     Parameters
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -459,6 +459,9 @@ def threshold_isodata(image, nbins=256, return_all=False):
         return thresholds[0]
 
 
+# Computing a histogram using np.histogram on a uint8 image with bins=256
+# doesn't work and results in aliasing problems. We use a fully specified set
+# of bins to ensure that each uint8 value false into its own bin.
 _DEFAULT_ENTROPY_BINS = tuple(np.arange(-0.5, 255.51, 1))
 
 
@@ -475,7 +478,8 @@ def _cross_entropy(image, threshold, bins=_DEFAULT_ENTROPY_BINS):
         The number of bins or the bin edges. (Any valid value to the ``bins``
         argument of ``np.histogram`` will work here.) For an exact calculation,
         each unique value should have its own bin. The default value for bins
-        ensures exact handling of uint8 images.
+        ensures exact handling of uint8 images: ``bins=256`` results in
+        aliasing problems due to bin width not being equal to 1.
 
     Returns
     -------
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -459,13 +459,66 @@ def threshold_isodata(image, nbins=256, return_all=False):
         return thresholds[0]
 
 
-def threshold_li(image):
-    """Return threshold value based on adaptation of Li's Minimum Cross Entropy method.
+def _cross_entropy(image, threshold, bins=np.arange(-0.5, 255.51, 1)):
+    """Compute cross-entropy between distributions above and below a threshold.
 
     Parameters
     ----------
-    image : (N, M) ndarray
+    image : array
+        The input array of values.
+    threshold : float
+        The value dividing the foreground and background in ``image``.
+    bins : int or array of float
+        The number of bins or the bin edges. (Any valid value to the ``bins``
+        argument of ``np.histogram`` will work here.) For an exact calculation,
+        each unique value should have its own bin. The default value for bins
+        ensures exact handling of uint8 images.
+
+    Returns
+    -------
+    nu : float
+        The cross-entropy target value as defined in [1]_.
+
+    Notes
+    -----
+    See Li and Lee, 1993 [1]_; this is the objective function `threshold_li`
+    minimizes. This function can be improved but this implementation most
+    closely matches equation 8 in [1]_ and equations 1-3 in [2]_.
+
+    References
+    ----------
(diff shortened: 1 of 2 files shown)


skimage/filters/thresholding.py
```diff
@@ -459,14 +459,67 @@ def threshold_isodata(image, nbins=256, return_all=False):
         return thresholds[0]
 
 
-def threshold_li(image):
-    """Return threshold value based on adaptation of Li's Minimum Cross Entropy method.
+def _cross_entropy(image, threshold, bins=np.arange(-0.5, 255.51, 1)):
```
sciunto: Is the default value intended?

skimage/filters/thresholding.py
```diff
@@ -459,14 +459,67 @@ def threshold_isodata(image, nbins=256, return_all=False):
...
+
+def threshold_li(image, *, tolerance=None):
+    """Compute threshold value by Li's iterative Minimum Cross Entropy method.
+
+    Parameters
+    ----------
+    image : ndarray
         Input image.
 
+    tolerance : float
```
sciunto: , optional.

skimage/filters/thresholding.py
```diff
@@ -459,14 +459,67 @@ def threshold_isodata(image, nbins=256, return_all=False):
...
+    m0b = np.sum(histogram[t:])
+    m1a = np.sum(histogram[:t] * bin_centers[:t])  # 1st moment, background
+    m1b = np.sum(histogram[t:] * bin_centers[t:])
+    mua = m1a / m0a  # mean value, background
+    mub = m1b / m0b
+    nu = -m1a * np.log(mua) - m1b * np.log(mub)
+    return nu
+
+
+def threshold_li(image, *, tolerance=None):
```
sciunto: This signature looks weird

skimage/filters/thresholding.py
jni: Yes. Unfortunately you get some nasty aliasing artifacts if you just write 255 or 256...

skimage/filters/thresholding.py
jni: it is a Python 3 keyword-only argument. https://www.python.org/dev/peps/pep-3102/

jni:
@sciunto thanks for the quick review! I've addressed your docstring comments. For the default value, it's a bit weird, but since it's correct and it's a private function, I figure it can stay like that. I could also set it to None and compute it inside the function. I'm ok with either option, depending on reviewer preferences. =)

skimage/filters/thresholding.py
sciunto: oh, yes, I forgot that one. However, don't we want to backport this? you said that the current version can return incorrect thresholds, which is embarrassing.

jni: Yes but my philosophy with backports is to code as we want to code (ie in Python 3 if it fits), and then manually fix backports if necessary. The whole idea of dropping support for 2.7 on master is to stop being hamstrung by the need to support old syntax.

In this specific case, we never supported tolerance before so we can hardcode it anyway.

sciunto: ok :)

skimage/filters/thresholding.py
sciunto: I would prefer to apply your alternative solution, ie, specifying this range before the function call instead as a default value.

skimage/filters/thresholding.py
```diff
@@ -459,14 +459,67 @@ def threshold_isodata(image, nbins=256, return_all=False):
...
+    """Compute threshold value by Li's iterative Minimum Cross Entropy method.
+
+    Parameters
+    ----------
+    image : ndarray
         Input image.
```
sciunto: blank line?

skimage/filters/thresholding.py
```diff
@@ -494,44 +547,47 @@ def threshold_li(image):
...
 
     # Stop the iterations when the difference between the
     # new and old threshold values is less than the tolerance
-    while abs(new_thresh - old_thresh) > tolerance:
-        old_thresh = new_thresh
-        threshold = old_thresh + tolerance   # range
-        # Calculate the means of background and object pixels
-        mean_back = image[image <= threshold].mean()
-        mean_obj = image[image > threshold].mean()
+    while abs(t_next - t_curr) > tolerance:
```
sciunto: Naive question. Do we have the warranty that it converges in any case?

sciunto:
I left few comments. Then, it'll be ready to go.

skimage/filters/thresholding.py
jni: @sciunto yes, the paper proves this, although I'll admit that I didn't follow the proof in detail. =)

jni:
@sciunto what do you think of the current solution?

sciunto:
Yeap, that's better. Sorry to bother you again. Could you had the argument you said here to give the reason of that choice? Then, I merge. Promised :)

sciunto:
Merged :)

jni:

jni:
I should have known. LOL

sciunto:
:D

PR #3403, opened 2018-09-13
Hardcode the inputs to test_ssim_grad

hmaarrfk:
This would randomly fail mostly when I run `pytest -n 4` to speed up the tests on 4 threads.

## Checklist
  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/tests/test_structural_similarity.py
+++ b/skimage/measure/tests/test_structural_similarity.py
@@ -60,7 +60,7 @@ def test_ssim_grad(seed):
     #       seeds that are known to work.
     #       The likely cause of this failure is that we are setting a hard
     #       threshold on the value of the gradient. Often the computed gradient
-    #       is only slightly larger than what whas measured.
+    #       is only slightly larger than what was measured.
     # X = np.random.rand(N, N) * 255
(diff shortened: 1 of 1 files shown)


emmanuelle:
I just installed ``pytest-xdist``, speeding up tests is pretty cool :-). I also get a failure here with ``n=4``. Instead of hardcoding, why don't you use a random number generator with a fixed seed (like in test_slic.py for example, test_color_2d)? Fixing the seed solves the problem on my machine.

hmaarrfk:
@emmanuelle even smarter

jni:
@hmaarrfk thanks again! =)

hmaarrfk:
Thanks @jni

hmaarrfk:
Thanks for reviewing @emmanuelle

hmaarrfk:
FYI, we need to find a solution for the tiffile doctest before you can run doctests in parallel too.

Those tests require a temporary file to be present which is created in the "first doc test".

I'm kinda in favour of helping tiffile package a wheel, then we can (optionally) depend on tiffile and not version it.

PR #3415, opened 2018-09-25
Add explicit Trove classifier for Python 3

hugovk:
## Description

scikit-image only supports Python 3.5 and up.

However, this is not clear from the [PyPI metadata](https://pypi.org/project/scikit-image/):

The `'Programming Language :: Python',` Trove classifier is used by many projects as shorthand for "Python 2".

This PR makes the supported versions clear by:

* Adding the Python-3 only classifier
* Adding explicit Python 3.5-3.7 classifiers

~This is important as the `python_requires` metadata (added in PR https://github.com/scikit-image/scikit-image/pull/3119 for https://github.com/scikit-image/scikit-image/issues/3117) is missing from PyPI. (I'll open another issue for this.)~ (Edit: oh, I see that the last 0.14.1 release was a point release, so not from master.)

## Checklist
- [n/a] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [n/a] Gallery example in `./doc/examples` (new features only)
- [n/a] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
- [n/a] Unit tests

## For reviewers

  later.


hmaarrfk:
scikit-image 0.14 still supports 2.7. 
If you want the metadata to be correct on PyPi, this PR will have to be adapted and be made explicitly against the 0.14.x branch.

As a note, 0.14.1 might be released soon, so this PR request is quite timely.

I don't know of the timeline for releasing 0.15 (which will be Python3 only).

jni:
@hmaarrfk I don't understand what you mean, this PR seems correct to me? It's against master and it specifies that master doesn't work for 3.4 and lower. There is a question of whether we want to backport it to 0.14.x with proper modifications, but it is orthogonal to this PR?

At any rate, LGTM, thanks @hugovk, I'll approve it let another dev comment on it and merge in case I'm missing something. =)

jni:
@hmaarrfk well, this change affects master. These classifiers are not getting pushed to PyPI until master gets pushed to PyPI. Which is 0.15. And different versions of a package can have different classifiers on PyPI.

hugovk:
Yes, I was initially mistaken in thinking the PyPI version had already dropped Python 2, and my edit wasn't very clear.

So this isn't as urgent as I thought, but I think this change is good for master (0.15).

I've made a similar PR for 0.14 as there's still 8 months until 0.15 is due out: https://github.com/scikit-image/scikit-image/pull/3417.

Sorry for the confusion!

sciunto:
All looks great! thanks.

PR #3432, opened 2018-09-30
Build tools: Fix PyQt installed in minimum requirements build

hmaarrfk:
Somewhere along the last few maintenance PRs the MINIMUM_REQUIREMENTS broke.

See this build that failed 
https://travis-ci.org/scikit-image/scikit-image/jobs/435259570

Hopefully the build process will be a little more normalized after this. `MINIMUM_REQUIREMENTS` is just really fragile....

## For reviewers

  later.

### DIFF
--- a/tools/travis/script.sh
+++ b/tools/travis/script.sh
@@ -10,6 +10,11 @@ if [[ $TRAVIS_OS_NAME == "osx" ]]; then
     echo 'backend : Template' > $MPL_DIR/matplotlibrc
 fi
 
+section "list.installed.dependencies"
+pip list
+tools/build_versions.py
+section_end "list.installed.dependencies"
+
 section "Test"
 pytest $TEST_ARGS skimage
 section_end "Test"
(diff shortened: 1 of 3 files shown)


tools/travis/script.sh
```diff
@@ -22,6 +27,8 @@ section "Tests.examples"
 # Run example applications
 echo Build or run examples
 pip install --retries 3 -q -r ./requirements/docs.txt
+pip list
+tools/build_versions.py
```
soupault: This looks redundant, doesn't it?

tools/travis/script.sh
```diff
@@ -10,6 +10,11 @@ if [[ $TRAVIS_OS_NAME == "osx" ]]; then
     echo 'backend : Template' > $MPL_DIR/matplotlibrc
 fi
 
+section "list.installed.dependencies"
+pip list
+tools/build_versions.py
```
soupault: Same here.

hmaarrfk: QT isn't included in the requirements file, so to see what was installed, I had to use the `pip list` command.

tools/travis/script.sh
hmaarrfk: I think both are useful. `build_versions` gives us a quick at a glance look of what we installed. `pip list` is very verbose. Maybe putting it in 

```
 section "examples.install.dependencies"
pip install --retries 3 -q -r ./requirements/docs.txt
pip list
tools/build_versions.py
 section_end "examples.install.dependencies"
```

would help clarify things?

tools/travis/script.sh
hmaarrfk: Maybe it would be clearer if this was moved to the `install` section?

PR #3433, opened 2018-09-30
Build tools: Add a PySide2 build

hmaarrfk:
The previous PySide environment variable was defunct. We should either take it off or use the more current flags.

logic was added in previous maintenance PR to be compatible with PySide2 but PySide2 wasn't really availble on PyPi.

This is rebased onto: https://github.com/scikit-image/scikit-image/pull/3432

Confirmation that this was run with pyside2 https://travis-ci.org/scikit-image/scikit-image/jobs/435296973#L2568

## For reviewers

  later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
   include:
     - os: linux
       python: 3.5
-      env: OPTIONAL_DEPS=1 WITH_PYSIDE=1 BUILD_DOCS=1
+      env: QT=PySide2 OPTIONAL_DEPS=1 BUILD_DOCS=1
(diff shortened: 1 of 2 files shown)


soupault:
@hmaarrfk what was the reason behind closing this PR?

hmaarrfk:
historically it seemed pyside was being tested. As the PySide project got orphaned by Qt, that test probably got dropped. Then recently PySide2 got released, and I wanted to add the code back.

But this was so low priority, and might cause maintenance headaches, so I figured it waasn't so necessary. 

We should still remove the environment variable at the very least.

PR #3440, opened 2018-10-03
MAINT Use np.full instead of cst*np.ones

rth:
This replaces `cst*np.ones` by `np.full` in a few places which avoids a memory copy and should be ~2x faster.

### DIFF
--- a/skimage/feature/orb.py
+++ b/skimage/feature/orb.py
@@ -187,7 +187,7 @@ class ORB(FeatureDetector, DescriptorExtractor):
             keypoints_list.append(keypoints * self.downscale ** octave)
             orientations_list.append(orientations)
             scales_list.append(np.full(
-                keypoints.shape[0], self.downscale ** octave, dtype=np.intp))
+                keypoints.shape[0], self.downscale ** octave, dtype='float64'))
             responses_list.append(responses)
 
         keypoints = np.vstack(keypoints_list)
--- a/skimage/feature/orb.py
+++ b/skimage/feature/orb.py
@@ -186,8 +186,8 @@ class ORB(FeatureDetector, DescriptorExtractor):
 
             keypoints_list.append(keypoints * self.downscale ** octave)
             orientations_list.append(orientations)
-            scales_list.append(self.downscale ** octave
-                               * np.ones(keypoints.shape[0], dtype=np.intp))
+            scales_list.append(np.full(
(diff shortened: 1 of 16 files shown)


rth:
The added changes should be PEP8 compliant; PEP8 warnings in other parts of the changed files don't seem to be too relevant to this PR.

skimage/feature/orb.py
```diff
@@ -186,8 +186,8 @@ def detect(self, image):
 
             keypoints_list.append(keypoints * self.downscale ** octave)
             orientations_list.append(orientations)
-            scales_list.append(self.downscale ** octave
-                               * np.ones(keypoints.shape[0], dtype=np.intp))
+            scales_list.append(np.full(
+                keypoints.shape[0], self.downscale ** octave, dtype=np.intp))
```
jni: It looks like this is causing a test failure: the unwanted copy, in this case, was also casting the result to float! I think changing dtype to float should fix things!

Other than that, thanks for a great PR! I found out about `np.full` quite late in the game, so it'll be great that people looking at our code will see the proper NumPythonic way of doing things! =)

rth: Thanks for the review @jni ! Fixed it, also specified a few dtypes manually as for numpy <1.12, np.full by default used the float64 dtype (same as np.ones), while after it uses the dtype of the second argument instead. So specifying it manually when it's not a float is safer..

skimage/util/_regular_grid.py
```diff
@@ -65,7 +65,8 @@ def regular_grid(ar_shape, n_points):
     space_size = float(np.prod(ar_shape))
     if space_size <= n_points:
         return (slice(None), ) * ndim
-    stepsizes = (space_size / n_points) ** (1.0 / ndim) * np.ones(ndim)
+    stepsizes = np.full(ndim, (space_size / n_points) ** (1.0 / ndim),
+                        dtype='float64')
```
hmaarrfk: Is there a reason to specify float64 here and elsewhere?

rth: In `np.full(x, fill_value)` with the newest numpy versions the resulting dtype will be of dtype of `fill_value`. In this case I'm not sure it there is a combinations of parameters that could make it an int, but in any case it's safer to specify the dtype explicitly..

hmaarrfk: Got it!

sciunto:
thank you @rth

PR #3442, opened 2018-10-03
Update instructions et al for releases after 0.14.1

jni:
After working on 0.14.1 I've made a few more changes to the release instructions. I've also forward-ported the 0.14.1 release notes to master, and updated a bit of the tooling.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,5 +19,6 @@ coverage:
 	$(PYTESTS) skimage --cov=skimage
 
 html:
-	pip install -q sphinx pytest-runner sphinx-gallery
+	pip install -q -r requirements/docs.txt
+	rm -rf doc/build
 	export SPHINXOPTS=-W; make -C doc html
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -19,5 +19,6 @@ coverage:
 	$(PYTESTS) skimage --cov=skimage
 
 html:
-	pip install -q sphinx pytest-runner sphinx-gallery
+	pip install -q -r requirements/docs.txt
+	rm -rf doc/build
```
soupault: Not sure about this line. I often partially rebuild the docs, especially, when working on a specific example.

Makefile
```diff
@@ -19,5 +19,6 @@ coverage:
 	$(PYTESTS) skimage --cov=skimage
 
 html:
-	pip install -q sphinx pytest-runner sphinx-gallery
+	pip install -q -r requirements/docs.txt
```
soupault: This is very cool!

Makefile
jni: ah, right, this isn't used only in releases. =) I'll add doc/build to make clean and I'll update the release instructions. I think it's important to do a clean build (since this is going to be the official documentation essentially forever), but you're right that it shouldn't be done always.

PR #3444, opened 2018-10-03
Fix default value of `methods` in `_try_all` to avoid exception

sciunto:
## Description

I discovered a trivial bug if method is used with the default value. len(None) will throw a TypeError. The bug was introduced by myself. However, it's not critical. The function is private.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -47,6 +47,9 @@ def _try_all(image, methods=None, figsize=None, num_cols=2, verbose=True):
     """
     from matplotlib import pyplot as plt
 
+    # Handle default value
+    methods = methods or {}
+
     num_rows = math.ceil((len(methods) + 1.) / num_cols)
     num_rows = int(num_rows)  # Python 2.7 support
     fig, ax = plt.subplots(num_rows, num_cols, figsize=figsize,


jni:
Erm, this doesn't actually fix the issue? :joy: `AttributeError: 'tuple' object has no attribute 'items'`. Since the input type is dict, I suggest either `dict()` or `None` with a line: `methods = methods or {}` inside the function. The second line is slightly preferable because it's considered bad practice to use mutable objects as default values.

sciunto:
argh, I read tuple instead of dict :(

sciunto:
Fixed. Thanks @jni for the hint.

skimage/filters/thresholding.py
```diff
@@ -47,6 +47,9 @@ def _try_all(image, methods=None, figsize=None, num_cols=2, verbose=True):
     """
     from matplotlib import pyplot as plt
 
+    # Handle default value
+    methods = methods or {}
```
hmaarrfk: I'm just going to say that I really don't like this line of logic. I get that `None is Flasy`, but man, is a `dictionary Truthy`?

I would much prefer
```
methods = {} if methods is None else method
```
or 
```
if methods is None:
    methods = {}
```

jni: @hmaarrfk the dictionary doesn't have to be truthy. (In fact it isn't: empty collections are `False`.) I only learned of this expression quite recently but I very much like it as a concise way of representing your alternatives. To me it reads quite naturally. Having said that I get the objection and will not myself object if others prefer one of your more explicit forms.

hmaarrfk: my issue with that particular line is that it has no equivalence in many other languages. Typically the result of an or expression is a Boolean, but now it is either the LHS or the RHS.


False, empty lists, empty tuples, zero are not valid inputs to this function and that line of logic might imply so in 4 years.

jni: But that's fine: all of the above inputs would cause a nasty `AttributeError` later on in the function with your alternative syntax, so I don't find them to be compelling reasons not to use this...

PR #3447, opened 2018-10-06
 Fix morphology.local_maxima for input with any dimension < 3

### LINKED ISSUE #3261: BUG: ValueError for input [1, 1] in morphology.local_maxima when [0, 0] is expected
The error can be reproduced with all development versions of scikit-image that include 912d9d2 and the following snippet:
```python
>>> from skimage.morphology import local_maxima
>>> local_maxima([1, 1], allow_borders=False)
```
This raises 
```
----> 1 local_maxima([1, 1], allow_borders=False)

~/Projects/skimage/skimage/morphology/extrema.py in local_maxima(image, selem, connectivity, indices, allow_borders)
    441     selem = _resolve_neighborhood(selem, connectivity, image.ndim)
    442     neighbor_offsets = _offsets_to_raveled_neighbors(
--> 443         image.shape, selem, center=((1,) * image.ndim))
    444 
    445     # Array of flags used to store the state of each pixel during evaluation.

~/Projects/skimage/skimage/morphology/watershed.py in _offsets_to_raveled_neighbors(image_shape, structure, center)
    148     structure[tuple(center)] = 0  # Ignore the center; it's not a neighbor
    149     connection_indices = np.transpose(np.nonzero(structure))
--> 150     offsets = (np.ravel_multi_index(connection_indices.T, image_shape) -
    151                np.ravel_multi_index(center, image_shape))
    152     squared_distances = np.sum((connection_indices - center) ** 2, axis=1)

ValueError: invalid entry in coordinates array
```
but should return `array([0, 0])`.

Until me or someone else finds the time to implement a fix and regression test this issue acts as a reminder.

Discovered in #3255.

lagru:
## Description

- `local_maxima` (and `local_minima`) failed with a ValueError for images with at least 1 dimension with fewer than 3 elements. This was due to the implementation which applies a structuring element with exactly 3 elements in each dimension to the image in question. This behavior was only visible if `allow_borders` was false, otherwise the image was padded internally. 
This fix makes these functions work for the mentioned edge cases by testing the image shape at the appropriate time.
- The  output of `local_maxima([], indices=True)` didn't match the output for non-empty arrays and the description of the docstring. This is fixed as well.
- Replaced `assert_equal` with `assert_array_equal` which is more restrictive and (I think) more inline with what we actually want to test.

## Checklist
- [x] <strike>Gallery example in `./doc/examples` (new features only)</strike>
- [x] <strike>Benchmark in `./benchmarks`, if your changes aren't covered by an existing benchmark</strike>
- [x] Unit tests

## References
Closes #3261 

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/extrema.py
+++ b/skimage/morphology/extrema.py
@@ -16,6 +16,7 @@ import numpy as np
 from scipy import ndimage as ndi
 
 from ..util import dtype_limits, invert, crop
+from .._shared.utils import warn
 from . import greyreconstruct
 from .watershed import _offsets_to_raveled_neighbors
 from ._extrema_cy import _local_maxima
@@ -368,6 +369,12 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
         (0 otherwise). If `indices` is true, a tuple of one-dimensional arrays
         containing the coordinates (indices) of all found maxima.
 
+    Warns
+    -----
+    UserWarning
+        If `allow_borders` is false and any dimension of the given `image` is
+        shorter than 3 samples, maxima can't exist and a warning is shown.
+
     See Also
     --------
     skimage.morphology.local_minima
@@ -451,9 +458,12 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
     _set_edge_values_inplace(flags, value=3)
 
     if any(s < 3 for s in image.shape):
-        # Skip if any dimension is smaller than 3
+        # Warn and skip if any dimension is smaller than 3
         # -> no maxima can exist & structuring element can't be applied
-        pass
+        warn(
+            "no maxima can exist for an image with any dimension smaller 3",
+            stacklevel=3
+        )
     else:
         selem = _resolve_neighborhood(selem, connectivity, image.ndim)
         neighbor_offsets = _offsets_to_raveled_neighbors(
--- a/skimage/morphology/extrema.py
+++ b/skimage/morphology/extrema.py
@@ -461,7 +461,8 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
         # Warn and skip if any dimension is smaller than 3
         # -> no maxima can exist & structuring element can't be applied
         warn(
-            "no maxima can exist for an image with any dimension smaller 3",
+            "maxima can't exist for an image with any dimension smaller 3 "
+            "if borders aren't allowed",
             stacklevel=3
         )
     else:
--- a/skimage/morphology/extrema.py
+++ b/skimage/morphology/extrema.py
@@ -434,7 +434,11 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
     image = np.asarray(image, order="C")
     if image.size == 0:
         # Return early for empty input
-        return np.array([], dtype=(np.intp if indices else np.uint8))
+        if indices:
+            # Make sure that output is a tuple of 1 empty array per dimension
+            return np.nonzero(image)
+        else:
+            return np.zeros(image.shape, dtype=np.uint8)
 
     if allow_borders:
         # Ensure that local maxima are always at least one smaller sample away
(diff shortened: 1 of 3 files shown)


skimage/morphology/tests/test_extrema.py
```diff
@@ -217,14 +217,24 @@ def test_empty(self):
...
+        assert result[0].dtype == np.intp
+        assert result[1].size == 0
+        assert result[1].dtype == np.intp
 
     def test_dtypes(self):
         """Test results with default configuration for all supported dtypes."""
         for dtype in self.supported_dtypes:
             result = extrema.local_maxima(self.image.astype(dtype))
-            assert_equal(result, self.expected_default)
+            assert_array_equal(result, self.expected_default)
```
emmanuelle: Why did you prefer ``assert_array_equal`` here? ``assert_equal`` calls ``assert_array_equal`` when the arguments are arrays.

emmanuelle:
Thanks @lagru . Should a warning be raised when one of the dimensions is <3 (in order to help the user understand why no maxima are detected)?

lagru:
>  Should a warning be raised when one of the dimensions is <3 (in order to help the user understand why no maxima are detected)?

Why not. Good idea! I'll add it.

skimage/morphology/tests/test_extrema.py
lagru: Just a hunch. I felt that it is more explicit, stricter and more inline with what we are actually testing. I have the impression that's advantageous for unit tests. Should I revert this?

skimage/morphology/extrema.py
```diff
@@ -431,31 +441,44 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
...
-            raise TypeError("dtype of `image` is float16 which is not "
-                            "supported, try upcasting to float32")
-        else:
-            raise
+    if any(s < 3 for s in image.shape):
+        # Warn and skip if any dimension is smaller than 3
+        # -> no maxima can exist & structuring element can't be applied
+        warn(
+            "no maxima can exist for an image with any dimension smaller 3",
+            stacklevel=3
```
lagru: See https://github.com/scikit-image/scikit-image/issues/2097#issuecomment-428118113.

lagru:
@sciunto As far as I remember I this is ready to merge (test failure is unrelated). However because this PR involved adding a warning I decided to wait until a decision is made how the scikit-image wants to raise warnings (see #2097, #3438 and #3467).

By the way, because `local_minima` wraps `local_maxima` the warning can only have the correct stacklevel for one of these functions (currently the latter).

sciunto:
I restarted the test, it's passing now.

skimage/morphology/tests/test_extrema.py
jni: I don't mind either convention, but I would in general discourage you and other contributors from making changes like this one, which touch a lot of lines for just a matter of taste, as it creates noise in the git history for no substantial benefit. This seems small but, since I've spent a few times chasing down decisions by looking at git histories, I can attest that changes like this *are* disruptive to this kind of "archeological" work. =)

skimage/morphology/extrema.py
```diff
@@ -431,31 +441,44 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
...
-            # Provide the user with clearer error message
-            raise TypeError("dtype of `image` is float16 which is not "
-                            "supported, try upcasting to float32")
-        else:
-            raise
+    if any(s < 3 for s in image.shape):
+        # Warn and skip if any dimension is smaller than 3
+        # -> no maxima can exist & structuring element can't be applied
+        warn(
+            "no maxima can exist for an image with any dimension smaller 3",
```
jni: @lagru Perhaps point users to use `allow_borders=True` here?

jni:
Thanks @lagru!

skimage/morphology/tests/test_extrema.py
lagru: Your arguments against this change are way better than my hunch! I'll revert it. :D

stefanv:
Thanks for catching these; I love the guiding warnings and error messages.

stefanv:
@meeseeksdev backport to 0.14.x

stefanv:

stefanv:
Manual backport at https://github.com/scikit-image/scikit-image/pull/3545

lagru:
You're welcome! :slightly_smiling_face:

PR #3450, opened 2018-10-09
Maintenance: R strings where needed.

hmaarrfk:
I have no idea how I got those DeprecationWarnings to come up. But I did, and I cleaned up some r-strings.

## For reviewers

  later.

### DIFF
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -167,7 +167,7 @@ def _prune_blobs(blobs_array, overlap):
 
 def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
              overlap=.5,):
-    """Finds blobs in the given grayscale image.
+    r"""Finds blobs in the given grayscale image.
 
     Blobs are found using the Difference of Gaussian (DoG) method [1]_.
     For each blob found, the method returns its coordinates and the standard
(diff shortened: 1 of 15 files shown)


skimage/feature/blob.py
```diff
@@ -167,7 +167,7 @@ def _prune_blobs(blobs_array, overlap):
 
 def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
              overlap=.5,):
-    """Finds blobs in the given grayscale image.
+    r"""Finds blobs in the given grayscale image.
```
jni: Oh man, this is super ugly... Is it really necessary? The docstrings render correctly in the docs. I'd rather we found a way to turn off the warnings from sphinx (or wherever they're coming from)...

hmaarrfk:
Whatever it is an r. Numpy doesn’t even use the first line.

skimage/feature/blob.py
jni: That's certainly worth something! I just saw that the official sphinx docs recommends escaping the backslashes, though. Anyway I'm gonna change it to that and merge.

jni:
I thought your comment displayed some frustration with my nitpicking, but maybe I was overly sensitive. =)

hmaarrfk:
No, this whole escape character stuff frustrates me too. I often write lines on bash that are like `echo \\\\\a` and keep randomly adding `\` until it works.

hmaarrfk:

PR #3451, opened 2018-10-09
Update colorconv.py

zhanwenchen:
Use np.clip instead of two mask-based setters for the same clipping functionality.

## Description

## Checklist
  existing benchmark
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -637,9 +637,7 @@ def xyz2rgb(xyz):
     mask = arr > 0.0031308
     arr[mask] = 1.055 * np.power(arr[mask], 1 / 2.4) - 0.055
     arr[~mask] *= 12.92
-    arr[arr < 0] = 0
-    arr[arr > 1] = 1
-    return arr
+    return np.clip(arr, 0, 1)
 
 
 def rgb2xyz(rgb):


skimage/color/colorconv.py
```diff
@@ -637,9 +637,7 @@ def xyz2rgb(xyz):
     mask = arr > 0.0031308
     arr[mask] = 1.055 * np.power(arr[mask], 1 / 2.4) - 0.055
     arr[~mask] *= 12.92
-    arr[arr < 0] = 0
-    arr[arr > 1] = 1
-    return arr
+    return np.clip(arr, 0, 1)
```
hmaarrfk: you should at least clip in place. Please use a kwarg for that parameter.

`a_min` and `a_max` may be used at your discretion.

hmaarrfk:
Cool stuff. It really sucks that the `MINIMUM_REQUIREMENTS` build failed for reasons that are not your fault. 

I'm really interested in that one.

hmaarrfk:
Thanks for this PR. with in place clipping it could really help :D

zhanwenchen:
@hmaarrfk You are so helpful! I changed the code to clipping in place. I see that you have already changed the title of this PR. Is there anything else you'd like me to do?

hmaarrfk:
@sciunto actually changed your PR title. I'm not allowed to do that 😆.

@zhanwenchen no problem, glad to help.

As a style change, I typically wouldn't care for this kind of PR. But maybe others might.
I think somebody as smart as you can find better lines of code to change :D. This is why I suggested the performance angle.

Performance is tricky. You have to ask, performance for who? Not only that, performance depends alot on the OS, the compiler, and the hardware. As a performance change, we typically require a benchmark that illustrates the improvement. 

For example, it was not obvious to me that this actually sped anything up.
<details> <summary>Rough benchmark sketch</summary>

```python
import numpy as np
N = (1000, 10)
a = np.random.random(N)
# %%
%%timeit
a[a>0.8] = 0.8
a[a<0.2] = 0.2

# %%

a = np.random.random(N)

# %%
%%timeit
b = np.clip(a, 0.2, 0.8)


# %%
a = np.random.random(N)
# %%
%%timeit
np.clip(a, 0.2, 0.8, a)
```
</details>

Setting `N=(1000, 1000)` did not show any improvement with `np.clip`. I think this is because allocation of small arrays is easy and quick. For `N=(1000, 10000)`

<details><summary>Rough benchmark results `N=(1000, 10000)` </summary>

```python
import numpy as np
N = (1000, 10000)
a = np.random.random(N)

%%timeit
a[a>0.8] = 0.8
a[a<0.2] = 0.2
22.7 ms ± 698 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

a = np.random.random(N)

%%timeit
b = np.clip(a, 0.2, 0.8)
67.8 ms ± 2.14 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit
np.clip(a, 0.2, 0.8, a)
9.92 ms ± 173 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
</details>

For tiny arrays smaller than say `N = (1000, 10)`, it becomes really hard to tell the difference between the two approaches.

<details><summary> Clip vs boolean indexing, small array</summary>

```python
import numpy as np
N = (1000, 10)
a = np.random.random(N)

%%timeit
a[a>0.8] = 0.8
a[a<0.2] = 0.2
8.06 µs ± 145 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

a = np.random.random(N)

%%timeit
b = np.clip(a, 0.2, 0.8)
35.9 µs ± 663 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

a = np.random.random(N)

%%timeit
np.clip(a, 0.2, 0.8, a)
9.2 µs ± 155 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
<details>

Finally, these toy benchmarks don't exactly benchmark the full operation. Often a full benchmark is preferred as it better recreates the scenerio you want. I'm not saying you **should** do it for this particular case, but if you do, you might realize that this may or may not be the performance bottlekneck for this function. 

I bet you can find other interesting speedups in `xyz2rgb` or in `skimage` if you are so inclined.

PR #3452, opened 2018-10-09
MNT; handle deprecation warnings in tifffile

hmaarrfk:
Apparently numpy.fromtstring shouldn't be used on bytes.

This came to me in the form of a warning from numpy.

## For reviewers
  later.

### DIFF
--- a/skimage/external/tifffile/tifffile.py
+++ b/skimage/external/tifffile/tifffile.py
@@ -70,6 +70,8 @@ Requirements
 
 Revisions
 ---------
+2018.10.08
+    No longer use numpy.fromstring just use numpy.frombuffer.
 2017.01.12
     Read Zeiss SEM metadata.
     Read OME-TIFF with invalid references to external files.
@@ -2539,13 +2541,13 @@ class TiffPage(object):
                         # needs the raw byte order
                         typecode = dtype
                     try:
-                        return numpy.fromstring(x, typecode)
+                        return numpy.frombuffer(x, typecode)
                     except ValueError as e:
                         # strips may be missing EOI
                         warnings.warn("unpack: %s" % e)
                         xlen = ((len(x) // (bits_per_sample // 8)) *
                                 (bits_per_sample // 8))
-                        return numpy.fromstring(x[:xlen], typecode)
+                        return numpy.frombuffer(x[:xlen], typecode)
 
             elif isinstance(bits_per_sample, tuple):
                 def unpack(x):
@@ -3600,7 +3602,7 @@ class FileHandle(object):
                             offset=self._offset + offset,
                             shape=shape, order=order)
 
-    def read_array(self, dtype, count=-1, sep=""):
+    def read_array(self, dtype, count=-1):
         """Return numpy array from file.
 
         Work around numpy issue #2230, "numpy.fromfile does not accept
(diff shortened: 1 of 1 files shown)


jni:
Side notes:

- can you contribute this "upstream"? I believe the most "official" version of tifffile is here: https://github.com/blink1073/tifffile
- should we stop vendoring it? It seems @blink1073 is doing a good job with regular releases. (👋 @blink1073!)

hmaarrfk:
We need to make wheels for it then:
https://github.com/hmaarrfk/tifffle-wheels

help appreciated.

hmaarrfk:
Seems like this has already been done. I'll try to update our version instead.

hmaarrfk:
Sorry @jni, seems like his newest version requires numpy 1.14. Too new for scikit-image.

soupault:
@jni I'm pretty sure that `upstream` is on the personal page of @cgohlke - https://www.lfd.uci.edu/~gohlke/code/tifffile.py.html . https://github.com/blink1073/tifffile distributes the very file, but sometimes has a small lag.

Regarding the PR: I think that the issue has to be reported to the `upstream` first. Once the solution is approved, we should update the shipped version.

>We need to make wheels for it then:
>https://github.com/hmaarrfk/tifffle-wheels

@blink1073 would you be interested in improving the CI with the above?

blink1073:
> @blink1073 would you be interested in improving the CI with the above?

I don't follow what you mean.

hmaarrfk:
@soupault thanks, it seems that upstream has addressed this in 2018.02 but I was probably too aggressive in removing all instances of fromstring.

Finally, the reason why I didn't just take upstream's file is because they require Numpy 1.14 in their latest version.

hmaarrfk:
Sorry guys, I fixed my typo so the links above probably stopped working.

@soupault maybe the wheel discussion is better suited for: https://github.com/blink1073/tifffile/issues/46

jni:
@soupault 

> I'm pretty sure that upstream is on the personal page of @cgohlke - https://www.lfd.uci.edu/~gohlke/code/tifffile.py.html

Hmm, you are right. I guessed from the PyPI version number, which matched @blink1073's. But from the readme on that page:

> This is a mirror of the code at http://www.lfd.uci.edu/~gohlke/code/tifffile.py.html. For any development concerns, please email Christoph Gohlke at cgohlke at uci.edu.

This is such a weird way to make contributions these days... :man_shrugging:

hmaarrfk:
@jni, it seems that tifffile will now be supported upstream by the original author. Except a dependency doesn't contain wheels for linux or mac. I suggest we impement these fixes, and try to work with upstream to get wheels for all up and running.

hmaarrfk:
Anyway, failing because of the patch coverage. Nothing tests these changes, so we have to be extra careful, but the changes look ok.

hmaarrfk:
@scikit-image/core friendly ping to get this PR through.

skimage/external/tifffile/tifffile.py
```diff
@@ -70,6 +70,8 @@
 
 Revisions
 ---------
+2018.10.08
+    No longer use numpy.fromstring just use numpy.frombuffer.
```
sciunto: Please, could you indicate what you saif about upstream requires a more recent version of scipy and that this revision is made in our copy to avoid to bump numpy's min version.

Also, is it worth to add something in the TODO list, for when our min numpy version will satisfy the one of tifffile?

I'm OK with this PR, but I just want to make sure that we still understand what we did in 1 year.

hmaarrfk: I think the removal of the vendored code isn't obvious because tifffile is a moving target. It is not something that can just be done blindly because it is very likely that getting to 1.14 will take some time for scikit-image.

I think an issue on github tracking how upstream is dealing with this issue (i.e. to wheels exists for all our platforms) is probably best.

sciunto:
Great! thanks @hmaarrfk Merging this...

PR #3458, opened 2018-10-09
TST: don't use the pytest.fixtures decorator anymore in class fixtures 

hmaarrfk:
as explained by a pytest warning

> RemovedInPytest4Warning: Fixture "setUp" called directly. Fixtures are not meant to be called directly, are created automatically when test functions request them as parameters. See https://docs.pytest.org/en/latest/fixture.html for more information.

If you can make sense of why this works, you are a hero. I can't, but it works.

## For reviewers

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

PR #3467, opened 2018-10-10
MNT: Set stacklevel manually for warnings.

hmaarrfk:
Add a test to catch future unset stacklevels.
The test added here needs this PR to pass: https://github.com/scikit-image/scikit-image/pull/3459

@lagru your help on this would be appreciated.

Either this, or https://github.com/scikit-image/scikit-image/pull/3438 should be pulled in.

Depends on what style you like. See discussion on #3438 for more details.

## Checklist
  existing benchmark
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -519,6 +519,13 @@ In most cases, it is appropriate to set the ``stacklevel`` to ``2``, but cases
 when warnings originate from helper routines internal to the scikit-image
 library, it is may be more appropriate to set the ``stacklevel`` to ``3``.
 
+To test if your warning is being emitted correctly, try calling the function
+from an python console. You it should point you to the python console itself
+instead of being emitted by the files in the scikit-image library
+
+* **Good**: ``ipython:1: UserWarning: ...``
+* **Bad**: ``scikit-image/skimage/measure/_structural_similarity.py:155: UserWarning:``
+
 Bugs
 ----
 
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -478,8 +478,8 @@ that has to be changed to
 
     def a_function(image, rescale=None):
         if rescale is None:
-            warn('The default value of rescale will change to `False` in '
-                 'version N+3', stacklevel=2)
+            warn('The default value, currently `True`, of rescale will change '
+                  'to  `False` in version N+3', stacklevel=2)
             rescale = True
         out = do_something(image, rescale=rescale)
         return out
@@ -507,6 +507,17 @@ Note that the 2-release deprecation cycle is not a strict rule and in some
 cases, the developers can agree on a different procedure upon justification
 (like when we can't detect the change, or it involves moving or deleting an
 entire function for example).
+
+As warnings are rather intrusive, their message and placement is designed to
+guide users to take immediate action so that important changes
+will not break their code in a future release.
+The ``stacklevel`` argument in the ``warn``
+`function <https://docs.python.org/3/library/warnings.html#warnings.warn>`_
+should be used in order to ensure the warning point to the user's line of code,
+where they have the ability to take an action.
+In most cases, it is appropriate to set the ``stacklevel`` to ``2``, but cases
+when warnings originate from helper routines internal to the scikit-image
+library, it is may be more appropriate to set the ``stacklevel`` to ``3``.
 
 Bugs
 ----
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -476,6 +476,8 @@ that has to be changed to
 
 .. code-block:: python
 
+    from warnings import warn
+
     def a_function(image, rescale=None):
         if rescale is None:
             warn('The default value, currently `True`, of rescale will change '
@@ -509,15 +511,16 @@ cases, the developers can agree on a different procedure upon justification
 entire function for example).
 
 As warnings are rather intrusive, their message and placement is designed to
-guide users to take immediate action so that important changes
-will not break their code in a future release.
-The ``stacklevel`` argument in the ``warn``
-`function <https://docs.python.org/3/library/warnings.html#warnings.warn>`_
-should be used in order to ensure the warning point to the user's line of code,
-where they have the ability to take an action.
-In most cases, it is appropriate to set the ``stacklevel`` to ``2``, but cases
-when warnings originate from helper routines internal to the scikit-image
-library, it is may be more appropriate to set the ``stacklevel`` to ``3``.
+guide users to take immediate action so that important changes will not break
+their code in a future release.  scikit-image uses warnings to highlight
+changes in its API so that users may update their code accordingly. The
+``stacklevel`` argument sets the location in the callstack where the warnings
+will point. In most cases, it is appropriate to set the ``stacklevel`` to
+``2``.  When warnings originate from helper routines internal to the
+scikit-image library, it is may be more appropriate to set the ``stacklevel``
+to ``3``. For more information, see the documentation of the `warn
+<https://docs.python.org/3/library/warnings.html#warnings.warn>`__ function in
+the Python standard library.
 
 To test if your warning is being emitted correctly, try calling the function
 from an python console. You it should point you to the python console itself
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -510,17 +510,14 @@ cases, the developers can agree on a different procedure upon justification
 (like when we can't detect the change, or it involves moving or deleting an
 entire function for example).
 
-As warnings are rather intrusive, their message and placement is designed to
-guide users to take immediate action so that important changes will not break
-their code in a future release.  scikit-image uses warnings to highlight
-changes in its API so that users may update their code accordingly. The
-``stacklevel`` argument sets the location in the callstack where the warnings
-will point. In most cases, it is appropriate to set the ``stacklevel`` to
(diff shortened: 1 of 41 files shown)


lagru:
> @lagru your help on this would be appreciated.

Of course. I'll have a look once I have access to my notebook again.

hmaarrfk:
I’m not so sure if this te correct way to go about it. This is a pretty big change to many different files. I don’t think warnings are as rigorously tested as other aspects of the code. I think the previous fun tools.partial solution is leas intrusive and we can “discourage” the use of that convenience if we want to.

hmaarrfk:
@stefanv I know this change is much more intrusive, but it allows the use of that scipy script you mentioned.

I think it is a much better fix long term.

Definitely in the right direction for usability.

hmaarrfk:
@scikit-image/core your input on how to deal with the code vendored from scipy would be helpful.

I can rebase once https://github.com/scikit-image/scikit-image/pull/3458 gets in.

skimage/viewer/__init__.py
```diff
@@ -1,6 +1,6 @@
-from .._shared.utils import warn
 from .viewers import ImageViewer, CollectionViewer
 from .qt import has_qt
 
 if not has_qt:
-    warn('Viewer requires Qt')
+    from warnings import warn
+    warn('Viewer requires Qt', stacklevel=2)
```
soupault: Should be 1?

skimage/feature/register_translation.py
```diff
@@ -58,7 +58,49 @@ def _upsampled_dft(data, upsampled_region_size,
             raise ValueError("number of axis offsets must be equal to input "
                              "data's number of dimensions.")
 
-    im2pi = 1j * 2 * np.pi
+    # For the 2D case, it is more efficient to use the well-optimized
```
soupault: Is this intended change to be here?

skimage/viewer/__init__.py
hmaarrfk: Yea, could be 1. I tested this quite a bit, doesn't actually matter on what you select.

skimage/feature/register_translation.py
jni: ping @hmaarrfk something weird in your history...

hmaarrfk: Yeah. I'm really not sure. I didn't have time to figure out why the rebase went astray

hmaarrfk: Thahnks you two.  I think what happened was that that function was simplified. I was more careful with my rebase now and I think all is well.

jni:
@hmaarrfk thanks for the update! You've missed a couple of the old-style imports it seems:

https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=231

hmaarrfk:
I guess a few more warns where added. I hope I got the stack levels right...

hmaarrfk:
@scikit-image/core I just rebased this for 0.16. If anybody wanted to look at this, I think it will help users follow our changes in their code and help them transition as we add/improve the library. thanks

skimage/_shared/tests/test_warnings.py
```diff
@@ -0,0 +1,166 @@
...
+        msg += ("warnings should have an appropriate stacklevel:\n"
+                "    {}".format("\n    ".join(bad_stacklevels)))
+
+    if msg:
+        raise AssertionError(msg)
+
+
+
+"""
+Copyright (c) 2001, 2002 Enthought, Inc.
```
stefanv: We distribute all license text in the main LICENSE.txt now.

stefanv:
Would it make sense to have a default stacklevel of 2, rather than specifying it *almost* everywhere.

Also, can we provide some guidance to developers on how to determine the correct stacklevel?

hmaarrfk:
I think i had a PR where the default stack level was simply set to 2, unfortunately, when refactoring code, it becomes common to put warnings in a common function where the stack level should be 3. In that case, incrementing the stacklevel often gets forgotten.

Warnings, as shown by many of Issues recently, warnings are quite intrusive (though not as intrusive as your code breaking) and thus I think they warrant some thought at the time the PR is made. I think the few times where stack level is set to 3 makes it quite important to ask the contributor to think about who should receive the warning.

Generally, you should set them so that users get pointed to their line of code, where they can take an action to remedy the situation.

hmaarrfk:
I think i had a PR where the default stack level was simply set to 2, unfortunately, when refactoring code, it becomes common to put warnings in a common function where the stack level should be 3.

Warnings, as shown by many of Issues recently, are quite intrusive (though not as intrusive as your code breaking) and thus I think they warrant some thought at the time the PR is made.

Finally, that script you found @stefanv is pretty godly, it parses the python AST to find where `warn` is called without a stack level. Not sure we can do that with functools partial.


Generally, you should set them so that users get pointed to their line of code, where the user can take an action to remedy the situation.

jni:
I would very much appreciate a paragraph or two about how to correctly use stacklevels in the developer guide. I'm still fuzzy about how to correctly think about them.

Yes, I remember your old PR and people complaining that maybe it wasn't the right thing always. :joy: can't please everyone! fwiw I think I like this better because contributors are more likely to be familiar with warnings.warn than skimage.util.warn.

hmaarrfk:
Yeah, generally I like the enforceability of this approach. I'll add that heuristic to the dev docs..

hmaarrfk:
An example of an unhelpful warning:
> /home/mark2/miniconda3/envs/skdev/lib/python3.7/site-packages/sphinx/util/docutils.py:311: RemovedInSphinx30Warning: function based directive support is now deprecated. Use class based directive instead.
  RemovedInSphinx30Warning)

CONTRIBUTING.txt
```diff
@@ -507,6 +508,17 @@ cases, the developers can agree on a different procedure upon justification
...
+guide users to take immediate action so that important changes
+will not break their code in a future release.
+The ``stacklevel`` argument in the ``warn``
+`function <https://docs.python.org/3/library/warnings.html#warnings.warn>`_
+should be used in order to ensure the warning point to the user's line of code,
+where they have the ability to take an action.
+In most cases, it is appropriate to set the ``stacklevel`` to ``2``, but cases
+when warnings originate from helper routines internal to the scikit-image
+library, it is may be more appropriate to set the ``stacklevel`` to ``3``.
+
```
stefanv: Maybe one sentence on how a developer can test whether they did it correctly?

CONTRIBUTING.txt
```diff
@@ -507,6 +508,24 @@ cases, the developers can agree on a different procedure upon justification
...
+To test if your warning is being emitted correctly, try calling the function
+from an python console. You it should point you to the python console itself
+instead of being emitted by the files in the scikit-image library
+
+* **Good**: ``ipython:1: UserWarning: ...``
+* **Bad**: ``scikit-image/skimage/measure/_structural_similarity.py:155: UserWarning:``
```
hmaarrfk: @stefanv is this a good enough description

skimage/_shared/tests/test_warnings.py
hmaarrfk: Do you want to remove it from this file as well? I find it useful to keep the license in the file for these single "borrowed" files from libraries.

stefanv: Either way is fine, as long as it is listed in the main license file as well.

CONTRIBUTING.txt
stefanv: This is good, but nowhere is there a hint on what the stacklevel argument *means*.

The first sentence is also a bit long.  Perhaps, instead of:

> As warnings are rather intrusive, their message and placement is designed to
guide users to take immediate action so that important changes
will not break their code in a future release.

Just:

> scikit-image uses warnings to highlight changes in its API, so that users may update their code accordingly. The ``stacklevel`` ...

hmaarrfk: hopefully that addresses things.

skimage/_shared/tests/test_warnings.py
hmaarrfk: Done.

hmaarrfk:
@stefanv @jni @scikit-image/core, this is ready for review again. Hopefully we can get this out before 0.16!

hmaarrfk:
@scikit-image/core  what is the best way to get this through? should we just `touchup` the internal warn function so that it does what it is intended to do and call it a day? ortherwise, this PR, with all the suggestion (automated testing, documentation) is very hard to keep updated, especially as many other more important PRs need our attention.

That said, a recent PR that added many deprecation warnings basically points users to some obscure location in internal scikit-image functions with no recourse on where in their code they need to make the modifications. I really think this PR adds alot of usability to the deprecation policy.

jni:
@hmaarrfk I still think I like this approach, sorry that it's been hard to get through. When I need to warn I typically `git grep warn` to see how it's done elsewhere in the library. Combined with my distaste for "just for skimage" approaches, my preference would be to see `warnings.warn` being used with the correct stacklevel. (#4062 notwithstanding, that can also work for people.)

hmaarrfk:
The only thing left in this PR is the test that goes through the AST and checks all warnings calls.

Clearly There isn't a concensus on using the internal warn function or the the standard python one.

I opened PRs where the only thing I did was correct the places in here where the warning stack level wasn't set up correctly.

I'm going to close this PR and others are free to reopen if consensus is reached.

Issue #3475, opened 2018-10-15
lookfor is broken with cython == 0.29.0


hmaarrfk:
## Description
`lookfor` seems to be broken on travis. 
See failures on PR https://github.com/scikit-image/scikit-image/pull/3472

I can reproduce this on my machine with pip. Seems to be potentially related to Cython

@jni found
```
 $ pytest --pdb --doctest-modules skimage/util/lookfor.py 
==================================================================================================================== test session starts =====================================================================================================================
platform linux -- Python 3.6.3, pytest-3.5.1, py-1.5.3, pluggy-0.6.0
rootdir: /home/jni/projects/scikit-image/skimage, inifile:
plugins: cov-2.5.1, hypothesis-3.56.9
collected 1 item                                                                                                                                                                                                                                             

skimage/util/lookfor.py F
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
007 
008     Parameters
009     ----------
010     what : str
011         Words to look for.
012 
013     Examples
014     --------
015     >>> import skimage
016     >>> skimage.lookfor('regular_grid')
UNEXPECTED EXCEPTION: TypeError("'NoneType' object is not iterable",)
Traceback (most recent call last):

  File "/home/jni/miniconda3/envs/cf/lib/python3.6/doctest.py", line 1330, in __run
    compileflags, 1), test.globs)

  File "<doctest skimage.util.lookfor.lookfor[1]>", line 1, in <module>

  File "/home/jni/projects/scikit-image/skimage/util/lookfor.py", line 24, in lookfor
    return np.lookfor(what, sys.modules[__name__.split('.')[0]])

  File "/home/jni/miniconda3/envs/cf/lib/python3.6/site-packages/numpy/lib/utils.py", line 754, in lookfor
    cache = _lookfor_generate_cache(module, import_modules, regenerate)

[...]

PR #3479, opened 2018-10-17
MAINT: Blacklist specific Cython versions

hmaarrfk:
I think #3477 was prematurely merged.

This fix is much less restrictive as it is likely that Cython will get fixed in the next minor release.

Also see discussion in #3475

## For reviewers
  later.

### DIFF
--- a/requirements/README.md
+++ b/requirements/README.md
@@ -30,5 +30,5 @@ $ pip install -U -r requirements/test.txt
 
 ## Blacklist justification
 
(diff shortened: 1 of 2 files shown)


stefanv:
I don't think it matters much how we avoid specific versions of Cython.  We can also set our requirement to the latest version, once that comes out.  Is the version of Cython explicitly tested by the build script?  If not, we may want to add that, since not everyone installs via requirements.

PR #3480, opened 2018-10-17
Fix typos

yangfl:
## For reviewers

  later.


skimage/feature/_texture.pyx
```diff
@@ -103,7 +103,7 @@ def _local_binary_pattern(double[:, ::1] image,
     image : (N, M) double array
         Graylevel image.
     P : int
-        Number of circularly symmetric neighbour set points (quantization of
+        Number of circular symmetric neighbour set points (quantization of
```
jni: I think this should be reverted. "circularly" may not be in standard dictionaries but a quick Google search suggests plenty of references in the technical literature. Same for other corrections of "circularly symmetric".

skimage/feature/texture.py
```diff
@@ -281,7 +281,7 @@ def local_binary_pattern(image, P, R, method='default'):
     image : (N, M) array
         Graylevel image.
     P : int
-        Number of circularly symmetric neighbour set points (quantization of
+        Number of circular symmetric neighbour set points (quantization of
```
jni: As above.

skimage/restoration/_cycle_spin.py
```diff
@@ -76,7 +76,7 @@ def cycle_spin(x, func, max_shifts, shift_steps=1, num_workers=None,
     x : array-like
         Data for input to ``func``.
     func : function
-        A function to apply to circularly shifted versions of ``x``.  Should
+        A function to apply to circular shifted versions of ``x``.  Should
```
jni: As above.

jni:
```suggestion
        A function to apply to circularly shifted versions of ``x``.  Should
```

doc/examples/filters/plot_cycle_spinning.py
```diff
@@ -12,7 +12,7 @@
 averaging the results of the following 3-step procedure for multiple spatial
 shifts, n:
 
-1.) (circularly) shift the signal by an amount, n
+1.) (circular) shift the signal by an amount, n
```
jni:
```suggestion
1.) (circularly) shift the signal by an amount, n
```

skimage/feature/_texture.pyx
jni:
```suggestion
        Number of circularly symmetric neighbour set points (quantization of
```

skimage/feature/texture.py
jni:
```suggestion
        Number of circularly symmetric neighbour set points (quantization of
```

stefanv:
@OrkoHunter Take a look at the useful PEP8 report above ;)  Can we strip out the "there are no PEP8 errors" comments?

stefanv:
@yangfl Did you run this through an automated spell checker of sorts?  If so, it would be helpful to make that part of our CI toolchain.

yangfl:
Yes, I use codespell. But as always, spell checkers make false positive, which may be a little bit annoying if added to CI toolchain.

PR #1977, opened 2016-03-03
Added fused type warping

akuchibotla:
Starting a PR for discussion on https://github.com/scikit-image/scikit-image/issues/1287

Added a fused type and modified a few arguments but I have a few questions:
1.  What data types should we be including into the fused type?
2.  What arguments can safely be converted into the fused type?
3.  Where exactly is the casting to double happening to begin with?

PR #3486, opened 2018-10-19
Use fused types in denoise, warp

### LINKED ISSUE #3449: Denoise throws ValueError
## Description

`_denoise_cy.pyx` throws a `ValueError` saying a `double` was expected but a `float` was received.

```
  File "/home/aether/programs/anaconda3/envs/cellprofiler/lib/python2.7/site-packages/skimage/restoration/_denoise.py", line 117, in denoise_bilateral
    bins, mode, cval)
  File "skimage/restoration/_denoise_cy.pyx", line 114, in skimage.restoration._denoise_cy._denoise_bilateral
ValueError: Buffer dtype mismatch, expected 'double' but got 'float'
```

## Way to reproduce

- [X] Code example
```
import numpy as np
import skimage.restoration

sigma = 16.0
sigma_range = .2
np.random.seed(0)
image = np.random.uniform(size=(100, 100)).astype(np.float32)
mask = np.ones(image.shape, bool)
mask[40:60, 45:65] = False
expected = skimage.restoration.denoise_bilateral(
    image=image,
    multichannel=False,
    sigma_color=sigma_range,
    sigma_spatial=sigma,
)
```
- [ ] Relevant images (if any)
- [X] Operating system and version: Ubuntu 16.04
- [X] Python version
```
>>> print(sys.version)
2.7.14 |Anaconda, Inc.| (default, Dec  7 2017, 17:05:42) 
[GCC 7.2.0]
```
- [X] scikit-image version (run `skimage.__version__`)
```
>>> skimage.__version__
'0.14.1'
```

jni:
## Description
Fixes #3449, #1287 

Supersedes #3469, #3253 (I think), #1977. 

All credit to @AetherUnbound and @hmaarrfk here, I have just rebased their work on the latest master after @hmaarrfk fixed our negative indexing use in Cython.

## Checklist
  existing benchmark
- [x] Unit tests

## For reviewers

  later.


hmaarrfk:
Why is OSX failing? https://travis-ci.org/scikit-image/scikit-image/jobs/443575582#L5300

hmaarrfk:
@jni, I'm going to say that though I appreciate that my original PR is taken seriously, I don't know how worthwhile it is to complicate the warping code if it doesn't really give a speedup. In my experience, templated code is so much harder to read and to adapt. By including this, we are forcing future developers to also develop warping algorithms that function for all numeric types. This is a huge potential burden.

I would look into why it was that any speedup was obtained for float64 types and maybe just apply that fix. It might have been passing a parameter by reference in a tight loop which was necessary to template the functions.

I think the assumption that integer arithmetic is faster than floatingpoint is seriously flawed due to massive hardware optimizations for the floating point units in modern processors, though I'm not an expert in this field.

float32/float64 seems to be a valid assumption if you can guarantee not upcasting. float16 only recently started to exist because GPUs are running out of memory for machine learning.

I can't speak specifically for @AetherUnbound's PR, I was mostly helping with syntax.

jni:
@hmaarrfk but speed is not the only advantage, there is also RAM usage, right? I think that's very valuable. And personally I don't find the code much more complicated. And you know I'm quite nitpicky about complicated code! =D

It looks like the test failure is to do with random noise values. Same failure on that Mac build as on AppVeyor:
https://travis-ci.org/scikit-image/scikit-image/jobs/443575582#L5285-L5301

Dunno why it would happen in this PR and not elsewhere, since that test appears to use float64 anyway... Any ideas?

hmaarrfk:
Probably because we added some tests, the seed is different than before. Numpy (the repo) actually prints the values of the arrays that failed.

Would be useful if we had that....

jni:
@hmaarrfk it's not that the seed is different, it's that it's not set. You can run

```
pytest skimage/restoration/tests/test_denoise.py::test_denoise_bilateral_2d
```

repeatedly to get different results each time.

hmaarrfk:
@jni really? i thought the python seed was set to some consistent value everytime. I guess this isn't matlab again. It probably needs an environment variable.

hmaarrfk:
```
export PYTHONHASHSEED=1   
pytest skimage/restoration/tests/test_denoise.py::test_denoise_bilateral_2d 
```
will fail, but I can't run it with `--pdb`

hmaarrfk:
```
    def test_denoise_bilateral_2d():
        img = checkerboard_gray.copy()[:50, :50]
        # add some random noise
        noise = 0.5 * img.std() * np.random.rand(*img.shape)
        img += noise
        img = np.clip(img, 0, 1)
    
        out1 = restoration.denoise_bilateral(img, sigma_color=0.1,
                                             sigma_spatial=10, multichannel=False)
        out2 = restoration.denoise_bilateral(img, sigma_color=0.2,
                                             sigma_spatial=20, multichannel=False)
    
        # make sure noise is reduced in the checkerboard cells
>       assert img[30:45, 5:15].std() > out1[30:45, 5:15].std()
E       assert 0.07044972960234154 > nan
E        +  where 0.07044972960234154 = <built-in method std of numpy.ndarray object at 0x7fcbe9b3ce40>()
E        +    where <built-in method std of numpy.ndarray object at 0x7fcbe9b3ce40> = array([[ 0.12671524,  0.15149392,  0.21206663,  0.15423946,  0.00545805,\n         0.16724259,  0.21282153,  0.1807636 ...1418,  0.08659465,  0.017158  ,  0.18889751,\n         0.09346233,  0.08582789,  0.12415705,  0.2002917 ,  0.07256701]]).std
E        +  and   nan = <built-in method std of numpy.ndarray object at 0x7fcbe9b3cdf0>()
E        +    where <built-in method std of numpy.ndarray object at 0x7fcbe9b3cdf0> = array([[ nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan],\n       [ nan,  nan,  nan,  nan,  nan,  nan,  nan,...  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan],\n       [ nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan,  nan]]).std
```

skimage/restoration/_denoise_cy.pyx
```diff
@@ -176,22 +133,21 @@ def _denoise_tv_bregman(image, double weight, int max_iter, double eps,
         Py_ssize_t total = rows * cols * dims
 
     shape_ext = (rows2, cols2, dims)
-    u = np.zeros(shape_ext, dtype=np.double)
+    u = out.copy()
 
     cdef:
-        double[:, :, ::1] cimage = np.ascontiguousarray(image)
-        double[:, :, ::1] cu = u
+        np_floats[:, :, ::1] cu = u
```
hmaarrfk: Can `cu` be removed? `u` is already a horrible name

hmaarrfk:
Do we have a benchmark? is it possible to see what happens when we change the looping order of the tight loop in `_denoise_tv_bregman` to `C` ordering?

skimage/restoration/_denoise.py
```diff
@@ -109,8 +122,43 @@ def denoise_bilateral(image, win_size=None, sigma_color=None, sigma_spatial=1,
...
+    image = np.atleast_3d(img_as_float(image))
+    image = np.ascontiguousarray(image)
+
+    sigma_color = sigma_color or image.std()
+
+    color_lut = _compute_color_lut(bins, sigma_color, max_value, image.dtype)
+
+    range_lut = np.empty(win_size * win_size, dtype=image.dtype)
+
+    out = np.empty(image.shape, dtype=image.dtype)
```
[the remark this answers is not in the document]
hmaarrfk: nevermind. this inst the right denoising algorithm.

skimage/restoration/_denoise_cy.pyx
```diff
@@ -111,22 +73,18 @@ def _denoise_bilateral(image, Py_ssize_t win_size, sigma_color,
                          "`edge`, `wrap`, `symmetric` or `reflect`.")
     cdef char cmode = ord(mode[0].upper())
 
-    cimage = np.ascontiguousarray(image)
 
-    out = np.zeros((rows, cols, dims), dtype=np.double)
-    color_lut = _compute_color_lut(bins, csigma_color, max_value)
-    range_lut = _compute_range_lut(win_size, sigma_spatial)
```
hmaarrfk: We need to call `_compute_range_lut`, which is never done in the new implementation. Rather it is replaced by empty.

skimage/restoration/_denoise.py
```diff
@@ -157,7 +205,18 @@ def denoise_tv_bregman(image, weight, max_iter=100, eps=1e-3, isotropic=True):
...
+
+    rows = image.shape[0]
+    cols = image.shape[1]
+    dims = image.shape[2]
+
+    shape_ext = (rows + 2, cols + 2, dims)
+
+    out = np.zeros(shape_ext, image.dtype)
+
+    return _denoise_tv_bregman(image, weight, max_iter, eps, isotropic, out)
```
hmaarrfk: I think good design dictates that a function shouldn't mutate a variable AND return a variable.

I know this is a common thread in numpy function design, but I think we should see that as an exception rather than the norm.

skimage/restoration/_denoise_cy.pyx
```diff
@@ -176,22 +133,21 @@ def _denoise_tv_bregman(image, double weight, int max_iter, double eps,
         Py_ssize_t total = rows * cols * dims
 
     shape_ext = (rows2, cols2, dims)
-    u = np.zeros(shape_ext, dtype=np.double)
+    u = out.copy()
```
hmaarrfk: By copying out, out doesn't become the output paramter anymore. And you are allocating a huge array without using it.

`u` should be removed in favour of `out` and other variables should be renamed accordingly.

hmaarrfk: Thee reason this wasn't caught, is that the function "seemed to mutate out", but in fact, it was simply returning the correct value.

skimage/restoration/_denoise.py
hmaarrfk: You can do the `sqeezing` back in python. It also makes more sense to make ti there since `python` created the array. I think `squeezing` returns a view anyway.

hmaarrfk:
Finally, when all those comments are addressed, you can release the gil. You will need this line of code somewhere

```
                      color_lut_bin = min(
                          <Py_ssize_t>(dist * dist_scale), max_color_lut_bin)
```

skimage/restoration/_denoise.py
```diff
@@ -2,13 +2,26 @@
...
 
 
+def _compute_color_lut(bins, sigma, max_value, dtype):
+    color_lut = np.empty(bins, dtype=dtype)
+
+    sigma *= sigma
+    max_value /= bins
+
+    for b in range(bins):
+        color_lut[b] = _gaussian_weight(sigma, b * max_value)
```
hmaarrfk: Python code should use numpy and not cythonisms.

skimage/restoration/_denoise_cy.pyx
hmaarrfk: I personally dislike cython code. Feel free to move this to Python.

hmaarrfk:
@jni, thanks for helping find the bug.

Can you please close this PR? it seems there is more work to do in the denoising PR.

jni:
@hmaarrfk I don't see why the debugging needs to happens specifically in one or the other PR? I'm happy to take this on. Thanks for the detailed and specific review either way!

hmaarrfk:
whatever works for you and @AetherUnbound \o/

skimage/_shared/interpolation.pxd
```diff
@@ -169,57 +185,58 @@ cdef inline double biquadratic_interpolation(double* image, Py_ssize_t rows,
                                  r0 + pr, c0 + pc, mode, cval)
         fr[pr] = quadratic_interpolation(xc, fc)
 
-    # cubic interpolation for interpolated values of each row
-    return quadratic_interpolation(xr, fr)
+    out[0] = <np_real_numeric_out>quadratic_interpolation(xr, fr)
 
 
-cdef inline double cubic_interpolation(double x, double[4] f) nogil:
+cdef inline np_floats cubic_interpolation(np_floats x, np_real_numeric[4] f) nogil:
```
hmaarrfk: @jni, I think one reason we aren't seeing a big speedup is probably because this might converting uint8->float64. I was hoping it would resolve it as as a float32 but that seems like alot to expect from Cython.

Maybe the correct way is to provide an `out` parameter and explicitely specify which one we want. Though that is more `compiler work` that I would like to avoid. 

Apparently people have been using templates to avoid this kind of explicit casting due to cython limitations.

jni:
@AetherUnbound I've given this another go. Sorry to miss your ping earlier this month. I think @hmaarrfk is on holiday in Japan atm. =D

Let's see how my latest commits fare in CI! =)

jni:
One last thing I want to do is move all of the LUT code out of Cython. But for now it's bedtime.

jni:
It may not look like it, but CI is actually happy! :stuck_out_tongue_closed_eyes: :tada: :tada: :tada: 

@scikit-image/core can we have some reviews please? =D This fixes a regression in 0.14.1 relative to 0.14.0 so it's relatively urgent (within scikit-image levels of urgency, anyway =P).

The failures are in `--pre` because, as far as I can tell, NumPy 1.16 changed the deprecation message for `numpy.matrix`, so all the work in #3242 is now obsolete. :man_facepalming: I'll raise a separate issue but I don't think it should hold up this PR. [edit: @sciunto already found it in #3571]

https://travis-ci.org/scikit-image/scikit-image/jobs/461638460#L3340

```
ValueError: Unexpected warning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
```

skimage/_shared/interpolation.pxd
```diff
@@ -92,73 +102,79 @@ cdef inline double bilinear_interpolation(double* image, Py_ssize_t rows,
...
-          + dc * get_pixel2d(image, rows, cols, minr, maxc, mode, cval)
-    bottom = (1 - dc) * get_pixel2d(image, rows, cols, maxr, minc, mode,
-                                    cval) \
-             + dc * get_pixel2d(image, rows, cols, maxr, maxc, mode, cval)
-    return (1 - dr) * top + dr * bottom
 
+    cdef np.float64_t top
+    cdef np.float64_t bottom
+
+    cdef np_real_numeric top_left = get_pixel2d(image, rows, cols, minr, minc, mode, cval)
```
AetherUnbound: This is much more readable, nice!

skimage/_shared/interpolation.pxd
```diff
@@ -28,62 +32,68 @@ cdef inline Py_ssize_t fmax(Py_ssize_t one, Py_ssize_t two) nogil:
...
+        image, rows, cols, round(r), round(c), mode, cval)
 
 
-cdef inline double bilinear_interpolation(double* image, Py_ssize_t rows,
-                                          Py_ssize_t cols, double r, double c,
-                                          char mode, double cval) nogil:
+cdef inline void bilinear_interpolation(
+        np_real_numeric* image, Py_ssize_t rows, Py_ssize_t cols,
+        np_floats r, np_floats c, char mode, np_real_numeric cval,
+        np_real_numeric_out out[0]) nogil:
```
AetherUnbound: What does the `out[0]` here do? Why does it need to be defined and used here and below?

skimage/_shared/interpolation.pxd
```diff
@@ -92,73 +102,79 @@ cdef inline double bilinear_interpolation(double* image, Py_ssize_t rows,
...
+    cdef np_real_numeric bottom_right = get_pixel2d(image, rows, cols, maxr, maxc, mode, cval)
 
-cdef inline double quadratic_interpolation(double x, double[3] f) nogil:
+    top = (1 - dc) * top_left + dc * top_right
+    bottom = (1 - dc) * bottom_left + dc * bottom_right
+    out[0] = <np_real_numeric_out> ((1 - dr) * top + dr * bottom)
+
+cdef inline np_floats quadratic_interpolation(np_floats x,
+                                              np_real_numeric[3] f) nogil:
     """WARNING: Do not use, not implemented correctly.
```
AetherUnbound: :thinking: Should this still be here?

skimage/restoration/_denoise_cy.pyx
```diff
@@ -277,5 +205,3 @@ def _denoise_tv_bregman(image, double weight, int max_iter, double eps,
 
         rmse = sqrt(rmse / total)
         i += 1
-
-    return np.squeeze(np.asarray(u[1:u_height-1, 1:u_width-1]))
```
AetherUnbound: Is it important that this return statement no longer exists here? I assume that's because `out` is handed in above?

skimage/restoration/tests/test_denoise.py
```diff
@@ -182,6 +183,18 @@ def test_denoise_bilateral_2d():
...
 
 
+@pytest.mark.parametrize('dtype', [np.float16, np.float32, np.double])
+def test_denoise_bilateral_types(dtype):
+    img = checkerboard_gray.copy()[:50, :50]
+    # add some random noise
+    img += 0.5 * img.std() * np.random.rand(*img.shape)
+    img = np.clip(img, 0, 1)
+
+    # make sure that it can actually process the float32 types
```
AetherUnbound: This should probably be changed to say `float32` to `all valid float` - I can take care of that if that's fine with you

AetherUnbound:
Ah, I don't have permission to push to your branch :stuck_out_tongue_winking_eye:

jni:
@AetherUnbound you don't have permission, but for future reference, you can always make a PR to my fork/branch.

skimage/_shared/interpolation.pxd
jni: @AetherUnbound I honestly have no idea. I'm not touching it and will let @ahojnnes comment / deal with it as the author of that line. ;)

jni: This came up in the core dev meeting. It should still be there *and* we should raise a NotImplementedError when `order=2` 😂  I leave this for another PR!

skimage/_shared/interpolation.pxd
jni: A great question! @hmaarrfk? My initial guess is that this is another way of writing `np_real_numeric_out* out`?

jni:
@scikit-image/core anyone want to take a stab at reviewing this?

jni:
@scikit-image/core ping! As we are gearing up to release 0.14.2 any day now (ASAP given that we are broken with the newly release NumPy 1.16), this pull request would be very nice to have, since it fixes a bug introduced in 0.14.1.

jni:
@scikit-image/core ok, having rebased this after setting the language level to 3 (#3707), we're now getting this build error:

```
 ~/projects/scikit-image (bugfix/float-cast)
 $ pip install --no-deps -U -e .
Obtaining file:///home/jni/projects/scikit-image
    Complete output from command python setup.py egg_info:
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
    """
    from libc.math cimport ceil, floor
    
    import numpy as np
    cimport numpy as np
    from .fused_numerics cimport np_real_numeric, np_floats
    ^
    ------------------------------------------------------------
    
    skimage/_shared/interpolation.pxd:23:0: relative cimport beyond main package is not allowed
```

This is the only information *anywhere* about this error that I could find:

https://stackoverflow.com/questions/33555927/cython-relative-cimport-beyond-main-package-is-not-allowed/33572497

It I tried the following patch, but it doesn't help:

```
diff --git a/skimage/_shared/setup.py b/skimage/_shared/setup.py
index 998fc784d..4194a2876 100644
--- a/skimage/_shared/setup.py
+++ b/skimage/_shared/setup.py
@@ -19,7 +19,8 @@ def configuration(parent_package='', top_path=None):
 
     config.add_extension('geometry', sources=['geometry.c'])
     config.add_extension('transform', sources=['transform.c'],
-                         include_dirs=[get_numpy_include_dirs()])
+                         include_dirs=[get_numpy_include_dirs(),
+                                       os.path.abspath(os.path.join(base_path, '..'))])
     config.add_extension('interpolation', sources=['interpolation.c'])
     return config
```

(and neither does `'.'`). **Edit:** of course it doesn't, those extensions are only useful once Cython has generated the .c files.

I suspect that our multiple setup.py's are biting us here. @stefanv can you tell me the rationale for having module-specific setup.pys, and would you object to unifying them all at the base level if it solves this issue?

Or, if anyone has any suggestions about how to fix this, I'm all ears!

hmaarrfk:
I think cython was working on sharing object code between modules to make the binaries smaller, might be causing issues.

I suspect so many setup.py files has to do with growing pains of python. I've seen many such things in different packages.

In my code, I use the following as the body of a loop that iterates of `pyx` files:

```python
        file = file.relative_to(Path(__file__).parent)
        if platform.system() == 'Windows':
            module_name = str(file.parent / file.stem).replace('\\', '.')
        else:
            module_name = str(file.parent / file.stem).replace('/', '.')

        ext_modules = ext_modules + cythonize(Extension(
            module_name, sources=[str(file)]),
            compiler_directives={'language_level': 3})
```

We can make the top most setup.py just crawl through all the directories. That is what I do. not sure about c++. I don't use that in my builds, but i suspect `cythonize` deals with it.

hmaarrfk:
I think it was a 1 character fix.
scikit-image is still compiling, but now at least it gets to the compilation phase.

stefanv:
Travis-CI seems happy :)

Re: multiple setup files.  We never had a fully-automated solution (varying header files, sometimes using C++, headers, etc.), so the thought was to keep the setup code "close" to the files it described.  Maintaining one big setup file felt messy by comparison.

jni:
:man_facepalming: Thanks @hmaarrfk! ... Now do you want to approve the PR? =D

hmaarrfk:
This PR needs a benchmark. It is unclear, without the benchmark, what goal it achieves other than making compilation time really long.

It doesn't need to benchmark all dtypes, but maybe uint8, float32, and float64.

jni:
@hmaarrfk 

> It is unclear, without the benchmark, what goal it achieves

Err, no, right there in the description, it fixes #3449, a critical regression introduced by 0.14.1. It is important to note that the CellProfiler team has had to keep skimage pinned to 0.14.0 because of this, so in fact we should get this merged and released ASAP.

hmaarrfk:
One benchmark is included in my original PR. Could you please cherry pick that and simply test it with modern Cython?

jni:
> One benchmark is included in my original PR. Could you please cherry pick that

Ah! I had probably intended to do that but missed it. Hold please!

jni:
@hmaarrfk here are the results. master:

```
 $ asv dev -b WarpSuite
· Discovering benchmarks
[  0.00%] ·· Building for existing-py_home_jni_miniconda3_envs_cf_bin_python
[  0.00%] ·· Benchmarking existing-py_home_jni_miniconda3_envs_cf_bin_python
[ 50.00%] ··· Running benchmark_transform_warp.WarpSuite.time_same_type                                                     ok
[ 50.00%] ···· 
               =============== ====== ======== ======== ========
               --                               order           
               ---------------------- --------------------------
                   dtype_in      N       0        1        3    
               =============== ====== ======== ======== ========
                 numpy.uint8    128    132μs    357μs    1.22ms 
                 numpy.uint8    1024   12.5ms   27.4ms   63.9ms 
                 numpy.uint8    4096   314ms    566ms    1.20s  
                 numpy.uint16   128    139μs    372μs    933μs  
                 numpy.uint16   1024   12.5ms   29.4ms   66.6ms 
                 numpy.uint16   4096   323ms    577ms    1.25s  
                numpy.float32   128    134μs    381μs    1.20ms 
                numpy.float32   1024   15.3ms   35.7ms   74.8ms 
                numpy.float32   4096   355ms    686ms    1.20s  
                numpy.float64   128    128μs    479μs    1.05ms 
                numpy.float64   1024   17.9ms   35.7ms   72.5ms 
                numpy.float64   4096   317ms    565ms    1.14s  
               =============== ====== ======== ======== ========
```

PR HEAD:

```
 $ asv dev -b WarpSuite
· Discovering benchmarks
[  0.00%] ·· Building for existing-py_home_jni_miniconda3_envs_cf_bin_python
[  0.00%] ·· Benchmarking existing-py_home_jni_miniconda3_envs_cf_bin_python
[ 50.00%] ··· Running benchmark_transform_warp.WarpSuite.time_same_type                                                     ok
[ 50.00%] ···· 
               =============== ====== ======== ======== ========
               --                               order           
               ---------------------- --------------------------
                   dtype_in      N       0        1        3    
               =============== ====== ======== ======== ========
                 numpy.uint8    128    134μs    344μs    929μs  
                 numpy.uint8    1024   12.4ms   33.2ms   63.7ms 
                 numpy.uint8    4096   320ms    617ms    1.26s  
                 numpy.uint16   128    135μs    360μs    895μs  
                 numpy.uint16   1024   12.3ms   28.3ms   63.3ms 
                 numpy.uint16   4096   342ms    577ms    1.16s  
                numpy.float32   128    131μs    369μs    971μs  
                numpy.float32   1024   14.6ms   31.6ms   67.2ms 
                numpy.float32   4096   340ms    588ms    1.14s  
                numpy.float64   128    128μs    407μs    954μs  
                numpy.float64   1024   17.8ms   35.9ms   76.7ms 
                numpy.float64   4096   317ms    572ms    1.25s  
               =============== ====== ======== ======== ========
```
master:

```
[100.00%] ··· Running benchmark_transform_warp.WarpSuite.time_to_float64                                                    ok
[100.00%] ···· 
               =============== ====== ======== ======== ========
               --                               order           
               ---------------------- --------------------------
                   dtype_in      N       0        1        3    
               =============== ====== ======== ======== ========
                 numpy.uint8    128    129μs    363μs    931μs  
                 numpy.uint8    1024   11.6ms   26.2ms   61.4ms 
                 numpy.uint8    4096   311ms    537ms    1.10s  
                 numpy.uint16   128    122μs    387μs    1.00ms 
                 numpy.uint16   1024   11.4ms   25.7ms   66.6ms 
                 numpy.uint16   4096   306ms    535ms    1.12s  
                numpy.float32   128    139μs    385μs    997μs  
                numpy.float32   1024   14.3ms   31.4ms   64.6ms 
                numpy.float32   4096   302ms    570ms    1.12s  
                numpy.float64   128    125μs    377μs    967μs  
                numpy.float64   1024   17.7ms   33.6ms   70.2ms 
                numpy.float64   4096   310ms    567ms    1.14s  
               =============== ====== ======== ======== ========
```
PR HEAD:
```
[100.00%] ··· Running benchmark_transform_warp.WarpSuite.time_to_float64                                                    ok
[100.00%] ···· 
               =============== ====== ======== ======== ========
               --                               order           
               ---------------------- --------------------------
                   dtype_in      N       0        1        3    
               =============== ====== ======== ======== ========
                 numpy.uint8    128    143μs    380μs    990μs  
                 numpy.uint8    1024   11.7ms   27.1ms   63.3ms 
                 numpy.uint8    4096   305ms    534ms    1.13s  
                 numpy.uint16   128    128μs    368μs    1.02ms 
                 numpy.uint16   1024   11.5ms   27.1ms   63.8ms 
                 numpy.uint16   4096   308ms    526ms    1.12s  
                numpy.float32   128    142μs    379μs    1.34ms 
                numpy.float32   1024   14.0ms   29.8ms   69.3ms 
                numpy.float32   4096   302ms    551ms    1.18s  
                numpy.float64   128    118μs    355μs    926μs  
                numpy.float64   1024   18.0ms   35.0ms   72.2ms 
                numpy.float64   4096   308ms    561ms    1.17s  
               =============== ====== ======== ======== ========
```

jni:
Pretty much seems like a wash to me... Almost to the point where I wonder if I did something wrong. But I don't think so...

hmaarrfk:
Could be a wash. There was an asv continuous command that could elucidate this but this is probably good enough maybe

hmaarrfk:
I think the command is:

```
asv continuous -b WarpSuite -E conda:3.7 HEAD master
```

You probably want to run after reducicng the asv matrix to a bare minimum

i set numpy1.16 and scipy 1.2 and cythn 0.29, but I might go to sleep soon.

jni:
@hmaarrfk just how much worse is it? Generally I am not super concerned about compile times, and I don't think there is a better way to address this. I prefer supporting float32s as a "first class citizen" to always upcasting.

@stefanv would you like to express an opinion here? =)

jni:
Also, thank you very much for running the benchmarks! I have a visitor this week so I have been focusing on napari. But I agree with you that it would be excellent to push both 0.14.3 and 0.15 out the door. I think it's feasible. If we get this merged I'll devote some time today for release notes!

hmaarrfk:
lets just say that I used to "wait for scikit image to comile. Now I go get a coffee.

jni:
@hmaarrfk this sounds like a win-win! :joy:

stefanv:
This changes the return from `skimage.transform.warp` substantially.  I am concerned about returning ints from an interpolation operation: there is significant precision loss there.  I can imagine a `dtype` argument to determine the output, but it seems dangerous to presume the user always wants the same dtype out as in (yes, I know ndimage does this).

Either way, this behavior change is big enough that it would require deprecation if it were to go through.

jni:
@stefanv I think you've misread the PR. The output from `warp` is determined here:

https://github.com/scikit-image/scikit-image/blob/f9046b4f21d090417fb28a038b612d120c9fa5db/skimage/transform/_warps_cy.pyx#L138

and this line is not touched. As far as the warps module is concerned, and as far as I can tell, only internal functions are touched — which might explain the benchmark results 😂.

stefanv:
OK, so maybe I also missed the intent / benefit of this PR then.  Do you mind summarizing?

jni:
@stefanv the most immediate benefit is that it fixes #3449, a bug that was introduced in 0.14.1 that is preventing CellProfiler from upgrading past 0.14.0. The warping changes lay the groundwork for allowing float32 warps. Really there wouldn't be any changes to warp but we needed fused types to fix the denoise error and @hmaarrfk had already implemented them in #3253, so we built off that.

stefanv:
OK, I am +1; @jni @hmaarrfk do you mind confirming that all inline comments have been addressed?  E.g., the `pointer[0]` syntax should probably be `pointer*`, but otherwise everything looks good to me.

skimage/_shared/interpolation.pxd
hmaarrfk: Probably. lets try it. I forget. `* out` might not be valid cython syntax.... no wait it is.

hmaarrfk:
```suggestion
        np_real_numeric_out* out) nogil:
```

skimage/_shared/interpolation.pxd
hmaarrfk: :+1:

hmaarrfk:
I think everything is in order @stefanv

stefanv:
Don't ask :)

jni:
I think there's very little chance that this will work but...

jni:
I've submitted a manual backport.

hmaarrfk:

Issue #2008, opened 2016-03-16
Faster inpainting


samverstocken:
Hi,

I was wondering whether there are developments to make the biharmonic inpainting algorithm any faster (e.g. with a Cython implementation) or more scalable for larger inpainting masks? On some astronomical images, I've seen it perform well for smaller masks but I had to abort the code once I was trying to remove somewhat larger sources from my images since it was running for more than an hour (on one single source) and I had to revert to an inferior inpainting method for that purpose. On the bright side, the biharmonic inpainting produces really good results for me.

Thanks,
Sam

soupault:
Hi Sam!
Thank you very much for your interest and feedback!

Indeed, the implementation of biharmonic inpainting is quite far from optimal in the sense of performance. Current version is pretty straightforward due to the fact that it was written focusing nD-input support as a main target.

Possible improvements to the implementation include, but not limited to:
1. Pre-generating of bilaplacians (atm it is computed for each masked pixel separately);
2. Mask partitioning into independent connected regions (atm single huge matrix is built);
3. Cythonization (not sure if I'm able to write nD code in Cython atm);
4. Faster `linsolve`;
5. Parallel implementation.

As a intermediate solution, one can try to implement faster Cythonized version for 2D[+color] (taking into consideration also other points above), as it is expected to be the most common use case.

We're always happy to see PRs! If you are ready to face the challenge, don't hesitate to create one!

stefanv:
Feel free to discuss further, but I'm closing this issue as it is not a bug report.

soupault:
@samuelstjean hi! Could you, please, elaborate on your use case: do you have many tiny regions to inpaint / several large regions / combination of two?

samuelstjean:
I think you have summoned the wrong person by mistake, maybe @samverstocken is more knowledgeable about inpainting than me though.

soupault:
@samuelstjean of course, of course, my sincere apologies about that (I've got brain-DDoSed). Always happy to see you here, though :wink:.

CMCDragonkai:
I'm running the algorithm on (4165, 4744) image with 338477 pixels masked as true, and it's taking more than 2 hrs and still hasn't finished. Although I noticed that there's been some performance improvements since 0.12.3?

soupault:
@CMCDragonkai thanks for your input. Yes, I've made some trivial changes to speedup the algorithm on images with several separate inpainting domains (see https://github.com/scikit-image/scikit-image/pull/2234).
[...]

PR #3489, opened 2018-10-21
speedup for inpaint

GalAvineri:
## Description
This is a speedup for the inpaint method, specifically for the '_inpaint_biharmonic_single_channel' method, in the part where the masked points are being iterated over.

I've added a bit of speedup in 2 steps, each in a separate commit:
1. I've made a bit of speedup by vectorizing a few operations
2. I've added a speedup (perhaps a more significant one) by converting the iteration to a parallel manner.

## Checklist
I'm new to contributing to open source projects so before I start working on the checklist I would like to receive your thoughts about it :)

Btw, as i'm new to contributing i'd love to hear your advice on my attempt to contribute and i'd appreciate your help in completing this feature enhancement :)

## References
This is an enhancement regarding this issue #2008

## For reviewers

  later.

### DIFF
--- a/skimage/restoration/inpaint.py
+++ b/skimage/restoration/inpaint.py
@@ -37,12 +37,13 @@ def _inpaint_biharmonic_single_channel(mask, out, limits):
 
         # Iterate over masked point's neighborhood
         it_inner = np.nditer(neigh_coef, flags=['multi_index'])
-        for coef in it_inner:
+        coefs, multi_indices = [(coef, it_inner.multi_index) for coef in it_inner]
+        multi_indices = np.vstack(multi_indices)
+        tmp_pt_idxs = multi_indices + b_lo
+        tmp_pt_is = np.ravel_multi_index(tmp_pt_idxs.transpose(), mask.shape)
+        for coef, tmp_pt_idx, tmp_pt_i in zip(coefs, tmp_pt_idxs, tmp_pt_is):
             if coef == 0:
                 continue
-            tmp_pt_idx = np.add(b_lo, it_inner.multi_index)
-            tmp_pt_i = np.ravel_multi_index(tmp_pt_idx, mask.shape)
-
             if mask[tuple(tmp_pt_idx)]:
                 matrix_unknown[mask_pt_n, tmp_pt_i] = coef
             else:
--- a/skimage/restoration/inpaint.py
+++ b/skimage/restoration/inpaint.py
@@ -6,6 +6,8 @@ import scipy.ndimage as ndi
 from scipy.ndimage.filters import laplace
 import skimage
 from ..measure import label
+from multiprocessing import Pool, Array
+from itertools import repeat
 
 
 def _get_neighborhood(nd_idx, radius, nd_shape):
@@ -14,7 +16,7 @@ def _get_neighborhood(nd_idx, radius, nd_shape):
     return bounds_lo, bounds_hi
 
 
-def _inpaint_biharmonic_single_channel(mask, out, limits):
+def _inpaint_biharmonic_single_channel(mask, out, limits, processes=1):
     # Initialize sparse matrices
     matrix_unknown = sparse.lil_matrix((np.sum(mask), out.size))
     matrix_known = sparse.lil_matrix((np.sum(mask), out.size))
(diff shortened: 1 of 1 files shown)


emmanuelle:
@GalAvineri thanks for your PR! At the moment it seems that tests are not passing. Can you execute ``pytest skimage/restoration/tests/test_inpaint.py`` in your branch and try to understand where the problem comes from?

skimage/restoration/inpaint.py
```diff
@@ -73,6 +62,30 @@ def _inpaint_biharmonic_single_channel(mask, out, limits):
...
 
 
+def _modify_matrices_by_point(mask_pt_idx, mask_pt_n, out_shape, mask, matrix_unknown, matrix_known):
+    # Get bounded neighborhood of selected radius
+    b_lo, b_hi = _get_neighborhood(mask_pt_idx, 2, out_shape)
+
+    # Create biharmonic coefficients ndarray
+    neigh_coef = np.zeros(b_hi - b_lo)
+    neigh_coef[tuple(mask_pt_idx - b_lo)] = 1
+    neigh_coef = laplace(laplace(neigh_coef))
```
hmaarrfk: I think the call to laplace is the only one that is `parallelizable`. The others are all `python` and thus will get held up in the GIL.

hmaarrfk:
@GalAvineri, @emmanuelle how much overhead does `Pool(1)` add? is it worth having two code paths?

skimage/restoration/inpaint.py
GalAvineri: it seems that each masked point has some variables that are unique to it.
These contain b_lo, b_hi, neight_cief, and the variables it_inner, coefs, tmp_pt_idx and tmp_pt_is.
All of these varibales aren't shared between the processes and thus computing them for each point can be parralilzed in my opinion, as the GIL will not intervene here. Do you agree with me on this point?

I do agree that the GIL might intervene with the modifications of the matrix_unknown and matrix_known variables.
So regarding this, I can suggest the following:
Each process will produce a list of changes that should be made to these matrices, and the main thread will make all of them.
This way the computation of which modification should be made will be done in parallel, and the modification themselves will be done serially. What do you say?

GalAvineri:
@emmanuelle I will happily do so once i find some spare time! :)
@hmaarrfk I'm not sure about the overhead, how do you recommend checking this?
and what two code paths are you referring to?

hmaarrfk:
Hi Gal, please disregard my comment about the GIL, I thought I read threads, but this Is probably starting up different putthon interpreters each having their own GIL.

I would start by defining a benchmark for your particular use case. It could be a simple one using %timeit but we can work on integrating it with ASV once it is more fleshed out. You might find that you get very different speed ups on different workloads.

hmaarrfk:
The two codepaths would be: 1 starts the pool, the other just calls the function directly without staring a pool of 1

GalAvineri:
> The two codepaths would be: 1 starts the pool, the other just calls the function directly without staring a pool of 1

I'm still not entirely sure how you would have these 2 codes paths *together*, as mentioned here:

> @GalAvineri, @emmanuelle how much overhead does `Pool(1)` add? is it worth having two code paths?

In case you are asking if the overhead of creating a pool is beneficial, I would say it depends on the amount of time that could be parallelized later.
I had an example (quite an extreme one perhaps) of an data structure of shape (50, 50, 96) with 99.3% pixels masked to be inpainted.
The result of the parallelization was a speed of of about 160%, which in my case was 2 minutes faster.
In this case the overhead of creating the pool was certainly negligible.

Did I answer your question? If not, can you try to elaborate more? :)

GalAvineri:
> Hi Gal, please disregard my comment about the GIL, I thought I read threads, but this Is probably starting up different putthon interpreters each having their own GIL.

I agree, although you might be correct about the modifications of the known and unknown matrices.

> I would start by defining a benchmark for your particular use case. It could be a simple one using %timeit but we can work on integrating it with ASV once it is more fleshed out. You might find that you get very different speed ups on different workloads.

I'd love to do so once I find time for it :)

hmaarrfk:
@GalAvineri I'm always pretty excited about performance PRs. I'm curious to see what will come from this experiment of yours! Looking forward to see your modifications when you have time to make them.

You asked for thoughts on your implementation. As it stands, even when `processes=1`, you are staring a `Pool`. Therefore, this leads me to believe that this will cause a slowdown for most people, simply by virtue of the call to Pool.

`Pool(1)` is basically doing nothing. If you have a main process, you don't need a Pool of 1 process. 

When passing
Here is a simple benchmark showing how costly `Pool(1)` might be.
```python
In [1]: from multiprocessing import Pool                                      

In [2]: %timeit -n 10 -r 10 Pool(1)                                           
5.61 ms ± 1.67 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)

In [3]: %timeit -n 10 -r 10 pass                                              
78.5 ns ± 41.5 ns per loop (mean ± std. dev. of 10 runs, 10 loops each)
```

This benchmark indicates that starting a Pool of 1 worker is basically taking on the order of 5ms. That isn't really negligible compared to the noop of 100 ns. 

This is the kind of preliminery benchmark that shows us the usecase you are considering. 

Maybe more importantly than proving the performance speedup is proving the correctness as @emmanuelle pointed out.

GalAvineri:
Thank you for the great elaboration!
As you suggested I'll start with the correctness and we can resume the speed up conversation afterwards :)

soupault:
Hi @GalAvineri ,
Thanks for you interest in contributing and, particularly, in `inpaint_biharmonic`! 🙂

As already mentioned by the other dev team members, it would be very natural (and also very convenient to the reviewers) to have PRs such as yours supplemented with the benchmarking info. A set of numbers for comparison (running time on small, middle, large image / small, middle, large inpainting domain) would set a very solid ground for a constructive discussion. Optionally, you may consider adding an ASV (AirSpeed Velocity) file, which implements an interface for automatic benchmarking of the function in our infrastructure. For multiprocessing implementation, you may consider other PR, e.g. https://github.com/scikit-image/scikit-image/pull/3233.

Now for the PR itself. I think (as mentioned in https://github.com/scikit-image/scikit-image/issues/2008#issuecomment-197349560), that the suggested change is, indeed, one of the ways of accelerating the function. From a brief profiling, it seems that a great amount of time comes from the `laplace` calls and sparse matrix operations. My intuition is that Cythonizing the function would be the most efficient option, although I foresee a lot of troubles in writing 3D+ Cython code. So, I'd be really interested in seeing how fast we can go with pure multithreaded Python.
The algorithm, of course, has its own limits, but the current implementation is quite far from optimal. Just for reference, attached below is the profiling results of the gallery example.

>I had an example (quite an extreme one perhaps) of an data structure of shape (50, 50, 96) with 99.3% pixels masked to be inpainted.

🤣 What kind of science you are doing?

image

soupault:
Superseded by #5240.
Thank you for the contribution and, please, feel free to open a new PR is there is something else you would like to suggest implementation-wise.

PR #3491, opened 2018-10-21
Added axis option to block_reduce

pabloduque0:
## Description
Adding the option of specifying the axis in the func called can be very helpful for certain scenarios.

## Checklist
  existing benchmark
- [x] Unit tests

## References

## For reviewers

  later.


grlee77:
Hi @pabloduque0.

I think perhaps you want to allow reduction over only a subset of the block axes? I think that is a reasonable proposal, however, I don't think the implementation here is quite correct (but perhaps I am misreading something?). In any case, you will need to add test cases to verify the expected behavior to `skimage/measure/tests/test_block.py`

If I am reading things correctly, the internal call to `view_as_blocks` transforms to a view with shape (in 2D) of: `(nr, nc, br, bc)` where `(nr, nc)` is the original image shape and `(br, bc)` is the block shape. The following `for` loop then loops over the last axis, progressively reducing the dimension by 1 so that by the end of the loop, reduction has been performed over all block axes and the output has the same number of dimensions as the input.

If you are going to provide a specific axis to reduce over, then shouldn't the for loop need to be modified to only iterate over the provided axis/axes within the block (i.e. axis values should be offset by ``out.ndim//2``)? As is, it seems like reduction could occur over some of the original array axes as opposed to the block axes.

grlee77:
I think if you use a `block_size` of 1 on the axes you don't want to reduce over, it should already provide the desired functionality. I think efficiency could be improved a bit by skipping the reduction on any axes where the `block_size` is 1.

skimage/measure/block.py
```diff
@@ -18,6 +18,8 @@ def block_reduce(image, block_size, func=np.sum, cval=0):
     cval : float
         Constant padding value if image is not perfectly divisible by the
         block size.
+    axis : int
+        Specify the axis on which func will be called. Default is -1
```
hmaarrfk: I would avoid saying "Default is" as people can infer this from the signature. See for example: http://scikit-image.org/docs/stable/api/skimage.graph.html#skimage.graph.MCP

This is useful if `None` is in the signature, but later down, the code would translate `None` into a valid option.

pabloduque0:
Hello @grlee77 !
You are totally correct on the things you pointed out. I missed out on that last loop and it was indeed not working as expected. Sorry about that.
I have made a few changes so now it accepts an list of indices to reduce over. If all indices are passed, it produces the same result as the function does now.

I was not sure what default value set, I went with None as default and handled it that way. However there might be a cleaner option. List of indices could be passed as any array_like.

Also added the missing tests for this option, with argmax as an example. Let me know if you would like me to add more tests for other cases.

Let me know if you see anything that does not look right! I would be happy to correct it

skimage/measure/block.py
pabloduque0: Thanks for the feedback. I agree! I made some modifications on the code and now the default value is None, therefore I applied what you pointed out here. I could still be improvable, let me know

pabloduque0:
Hello @stefanv !
Yes, sorry if I was not clear enough on the initial description. For example, in some cases and functions that also implement an axis parameter, and therefore could work with this `reduce_block`, we do not want to reduce over all axes. Lets say an argmax, for a 2D block_size, we might want to reduce over rows or columns but not both.

pabloduque0:
> I think if you use a `block_size` of 1 on the axes you don't want to reduce over, it should already provide the desired functionality. I think efficiency could be improved a bit by skipping the reduction on any axes where the `block_size` is 1.

@grlee77  I havent implemented this just yet, but I could add it to the PR if you guys think it will be helpful!

matthew-brett:

<details>

## Rationale

</details>

PR #3493, opened 2018-10-22
Cython touchups

hmaarrfk:
img_as_float stopped being 64bit floats. A few things lingered in cython that assumed that behaviour.

Unfortunately, this still doesn't fix the bugs in 32bit orb on linux.

## For reviewers

  later.


sciunto:
Just a thought, if necessary, we might think about dropping 32 bits support for versions coming after the LTS release.

sciunto:
Just a thought, if necessary, we might think about dropping 32 bits support for versions coming after the LTS release.

sciunto:
Just a thought, if necessary, we might think about dropping 32 bits support for versions coming after the LTS release.

sciunto:
Just a thought, if necessary, we might think about dropping 32 bits support for versions coming after the LTS release.

sciunto:
Just a thought, if necessary, we might think about dropping 32 bits support for versions coming after the LTS release.

sciunto:
Just a thought, if necessary, we might think about dropping 32 bits support for versions coming after the LTS release.

sciunto:
Just a thought, if necessary, we might think about dropping 32 bits support for versions coming after the LTS release.

sciunto:
Just a thought, if necessary, we might think about dropping 32 bits support for versions coming after the LTS release.

hmaarrfk:
32 bit support is super valuable for tembedded guys playing with Raspebeery Pis. I would keep it.

It is almost surely indicative of a real bug

stefanv:
On Mon, 22 Oct 2018 11:27:34 +0000, Mark Harfouche wrote:
> It is almost surely indicative of a real bug

My suspicions always go in that direction first.  We should investigate
properly; the XFAILs we currently employ are very unsatisfactory.

hmaarrfk:
Yeah sorry stefan I still haven’t found why it doesn’t work on 32 bit. I am trying to isolate the little extra things in the code tht are no longer necessary in insulated PRs like this one

hmaarrfk:
@grlee77 sorry for changing this PR last minute. I've been playing with `libc.math` function, I thought they were as efficient/more efficient than a naive implementation in `inline Cython`, but they are sometimes 2x slower! I don't know whats up with that. So I revered the change that removed the `_min` and `_max` definition since things aren't exactly covered by benchmarks.

I'm worried about a regression.

grlee77:
Okay, thanks for looking into that. I had assumed performance would be similar. Perhaps it is just the function call overhead that makes the `libc.math` versions slower.

Is there anything else still pending here or do you think we are good to go?

hmaarrfk:
I think this is good to go. The libc math probably needs to be called inline.

PR #3501, opened 2018-10-23
Improvement: rank_filters: Change how the bitdepth and max_bin are computed to ensure exact warnings.

hmaarrfk:
Rank was giving a non-sensical warning due to floating point rounding errors.

See for example the test

https://github.com/scikit-image/scikit-image/blob/6f45aaa0d58e442667ae3b4cecc70f26903daff7/skimage/filters/rank/tests/test_rank.py#L529-L534

I renamed max_bin to n_bin because max_bin was only the maximum value of a bin in 1 line. In Cython, `+=1` was applied to make it the `number of bins`, but it carried its name. It was really confusing reading `range(max_bin)` and having to think: 

* Well range is not inclusive
* but max_bin is actually the number of bins
* So there isn't an off by 1 error.

## For reviewers

  later.

### DIFF
--- a/skimage/filters/rank/_percentile.py
+++ b/skimage/filters/rank/_percentile.py
@@ -38,11 +38,11 @@ def _apply(func, image, selem, out, mask, shift_x, shift_y, p0, p1,
            out_dtype=None):
 
     assert_nD(image, 2)
-    image, selem, out, mask, max_bin = _handle_input(image, selem, out, mask,
-                                                     out_dtype)
+    image, selem, out, mask, n_bins = _handle_input(image, selem, out, mask,
+                                                    out_dtype)
 
     func(image, selem, shift_x=shift_x, shift_y=shift_y, mask=mask,
-         out=out, max_bin=max_bin, p0=p0, p1=p1)
+         out=out, n_bins=n_bins, p0=p0, p1=p1)
 
     return out.reshape(out.shape[:2])
 
--- a/skimage/filters/rank/generic.py
+++ b/skimage/filters/rank/generic.py
@@ -92,28 +92,28 @@ def _handle_input(image, selem, out, mask, out_dtype=None, pixel_size=1):
     is_8bit = image.dtype in (np.uint8, np.int8)
 
     if is_8bit:
-        max_bin = 256
+        n_bins = 256
     else:
         # Convert to a Python int to avoid the potential overflow when we add
         # 1 to the maximum of the image.
-        max_bin = int(max(3, image.max())) + 1
+        n_bins = int(max(3, image.max())) + 1
 
-    if max_bin > 2**10:
+    if n_bins > 2**10:
         warn("Bad rank filter performance is expected due to a "
              "large number of bins ({}), equivalent to an approximate "
-             "bitdepth of {:.1f}.".format(max_bin, np.log2(max_bin)))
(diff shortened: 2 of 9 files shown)


skimage/filters/rank/generic.py
```diff
@@ -92,39 +92,40 @@ def _handle_input(image, selem, out, mask, out_dtype=None, pixel_size=1):
     is_8bit = image.dtype in (np.uint8, np.int8)
 
     if is_8bit:
-        max_bin = 255
+        n_bins = 256
```
stefanv: Is this correct?

skimage/filters/rank/generic.py
```diff
@@ -92,39 +92,40 @@ def _handle_input(image, selem, out, mask, out_dtype=None, pixel_size=1):
...
     else:
-        max_bin = max(4, image.max())
+        # do the addition after the conversion to python to ensure
```
stefanv: I don't understand this comment.

skimage/filters/rank/_percentile.py
```diff
@@ -38,11 +38,11 @@ def _apply(func, image, selem, out, mask, shift_x, shift_y, p0, p1,
            out_dtype=None):
 
     assert_nD(image, 2)
-    image, selem, out, mask, max_bin = _handle_input(image, selem, out, mask,
-                                                     out_dtype)
+    image, selem, out, mask, n_bins = _handle_input(image, selem, out, mask,
```
stefanv: Why this rename of `max_bin`?  It makes it hard to find the actual change in this patch.

skimage/filters/rank/generic.py
hmaarrfk: I totally should have split this PR in two steps, but they are so intertwined.

In most of the cython code, what is called `max_bin` is actually the old `255` number `+1`.

skimage/filters/rank/generic.py
hmaarrfk: It is to avoid overflow when we add `1`
```
In [4]: np.empty(3, dtype='uint8').max().dtype                                
Out[4]: dtype('uint8')
```

hmaarrfk: Clearly, i am wrong though!!!
In [6]: max(3, np.empty(3, dtype='uint8').max()).__class__                    
Out[6]: numpy.uint8

skimage/filters/rank/_percentile.py
hmaarrfk: @stefanv, I totally feel your pain on not wanting to review a 200 line PR. I'll revert the rename of `max_bin`, but I think the name is misleading. I split this PR into two commits. You may want to look at the first commit in isolation:
https://github.com/scikit-image/scikit-image/pull/3501/commits/d396fe6ff71637d2ee21d096482dcf5e094c8409

The second commit is the rename, and I'm happy to revert it. Looking at the first one may make you see why `max_bin` isn't actually the variable of interest. The variable that is being used to iterate over is `n_bins`.

hmaarrfk: I should rephrase myself, I'll revert the rename if you insist.

stefanv: Thanks, this is hugely helpful.

hmaarrfk: glad I was able to help!

PR #3518, opened 2018-10-28
Raise meaningful exception in warping when image is empty

ahojnnes:
Previously, we did not raise a descriptive error when the input image has a zero dimension.


hmaarrfk:
why is codecov still speaking? I thought I disabled it.

ahojnnes:
@soupault What is your cosmetic change supposed to improve? It causes a syntax error though.

hmaarrfk:
f strings are not allowed in 3.5

need to still use format.

@ahojnnes the previous version didn't output a string. It outputed a tuple, the first being a string, the second being the size.

PR #3525, opened 2018-10-30
Add detail to ProjectiveTransform repr and str

Erotemic:
## Description
This is a very simple improvement. I've simply added information on the matrix parameters to ProjectiveTransform. My motivation to make this change is that I had a list of AffineTransforms and I wanted to visually see the difference between them. Sure you can just get the params attribute with a list comprehension, but I thought it might be nicer if the repr and str functions were a bit more detailed.

Before when you had a projective transform (or a list of them) and go the repr you would see: 

```
<skimage.transform._geometric.ProjectiveTransform at 0x7fc6d65e4cf8>
```

and now with this change you will see something like:

```
<skimage.transform._geometric.ProjectiveTransform([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]) at 0x7fc6d65e4cf8>'
```

Analogous to the defaults in object, the str method is the same except it doesn't have the hex-id of the object.

## Implementation Choices

To start, I made the simplest possible change I could by just adding the parameter info. I cast `self.params` to a list to ensure that messy newlines wouldn't enter the repr. 

If I was to implement this in one of my own libraries I would probably make the following changes: 

* Remove the modname before the classname repr so the above would reduce the repr to simply 

```
<ProjectiveTransform([[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]]) at 0x7fc6d65e4cf8>
```

* Use `''.join([p.strip() for p in repr(self.params).split('\n')])` instead of `str(self.params.tolist())`, which makes the parameter repr a bit more concise when that matrix parameters have lots of decimals. This would change the above to 

```
<ProjectiveTransform(array([[1., 0., 0.], [0., 1., 0.], [0., 0., 1.]])) at 0x7fc6d65e4cf8>
```

To be conservative I started with the simplest method, but I can modify the PR to include any of the above modifications as well. 

(Actually if I was really doing this in one of my libraries, I would inherit from [`ubelt.NiceRepr`](https://github.com/Erotemic/ubelt/blob/master/ubelt/util_mixins.py) and define the `__nice__` method to return `''.join([p.strip() for p in repr(self.params).split('\n')])`.)

## Checklist
- [X] Unit tests

I feel the other things in the checklist are NA to this change

  existing benchmark

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -719,6 +719,36 @@ class ProjectiveTransform(GeometricTransform):
             raise TypeError("Cannot combine transformations of differing "
                             "types.")
 
+    def __nice__(self):
+        """ common 'paramstr' used by __str__ and __repr__ """
+        # paramstr = str(self.params.tolist())  # original proposal
+        npstring = np.array2string(self.params, separator=', ')
+        # add a newline and indentation after the first square braket to ensure
+        # the columns of the params matrix are aligned in the repr
+        lines = npstring[1:].split('\n ')
+        indented_lines = ['    ' + p for p in lines]
+        paramstr = 'matrix=[\n' + '\n'.join(indented_lines)
+        print(paramstr)
+        return paramstr
+
+    def __repr__(self):
+        """ Add standard repr formatting around a nice string """
+        paramstr = self.__nice__()
+        classname = self.__class__.__name__
+        classstr = classname
+        # modname = self.__class__.__module__  # uncomment for more verbosity
+        # classstr = modname + '.' + classname
+        return '<{}({}) at {}>'.format(classstr, paramstr, hex(id(self)))
+
+    def __str__(self):
+        """ Add standard str formatting around a nice string """
+        paramstr = self.__nice__()
+        classname = self.__class__.__name__
+        classstr = classname
+        # modname = self.__class__.__module__  # uncomment for more verbosity
+        # classstr = modname + '.' + classname
+        return '<{}({})>'.format(classstr, paramstr)
+
 
 class AffineTransform(ProjectiveTransform):
     """2D affine transformation.
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -721,7 +721,6 @@ class ProjectiveTransform(GeometricTransform):
 
     def __nice__(self):
         """ common 'paramstr' used by __str__ and __repr__ """
-        # paramstr = str(self.params.tolist())  # original proposal
         npstring = np.array2string(self.params, separator=', ')
         # add a newline and indentation after the first square braket to ensure
         # the columns of the params matrix are aligned in the repr
@@ -735,8 +734,6 @@ class ProjectiveTransform(GeometricTransform):
         paramstr = self.__nice__()
         classname = self.__class__.__name__
         classstr = classname
-        # modname = self.__class__.__module__  # uncomment for more verbosity
-        # classstr = modname + '.' + classname
         return '<{}({}) at {}>'.format(classstr, paramstr, hex(id(self)))
 
     def __str__(self):
@@ -744,8 +741,6 @@ class ProjectiveTransform(GeometricTransform):
         paramstr = self.__nice__()
         classname = self.__class__.__name__
         classstr = classname
-        # modname = self.__class__.__module__  # uncomment for more verbosity
-        # classstr = modname + '.' + classname
         return '<{}({})>'.format(classstr, paramstr)
 
 
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -728,7 +728,6 @@ class ProjectiveTransform(GeometricTransform):
         lines = npstring[1:].split('\n ')
         indented_lines = ['    ' + p for p in lines]
         paramstr = 'matrix=[\n' + '\n'.join(indented_lines)
-        print(paramstr)
         return paramstr
 
     def __repr__(self):
--- a/skimage/transform/tests/test_geometric.py
+++ b/skimage/transform/tests/test_geometric.py
@@ -483,16 +483,25 @@ def test_projective_repr():
             [0., 1., 0.],
             [0., 0., 1.]]) at
         ''').strip()) + ' 0x[a-f0-9]+' + re.escape('>')
+    # Hack the escaped regex to allow whitespace before each number for
+    # compatibility with different numpy versions.
+    want = want.replace('0\\.', ' *0\\.')
+    want = want.replace('1\\.', ' *1\\.')
     assert re.match(want, repr(tform))
 
 
 def test_projective_str():
     tform = ProjectiveTransform()
-    want = textwrap.dedent(
+    want = re.escape(textwrap.dedent(
         '''
         <ProjectiveTransform(matrix=[
             [1., 0., 0.],
             [0., 1., 0.],
             [0., 0., 1.]])>
-        ''').strip()
-    assert str(tform) == want
+        ''').strip())
+    # Hack the escaped regex to allow whitespace before each number for
+    # compatibility with different numpy versions.
+    want = want.replace('0\\.', ' *0\\.')
+    want = want.replace('1\\.', ' *1\\.')
+    print(want)
+    assert re.match(want, str(tform))
--- a/skimage/transform/tests/test_geometric.py
+++ b/skimage/transform/tests/test_geometric.py
@@ -1,4 +1,5 @@
 import numpy as np
+import re
 from skimage.transform._geometric import GeometricTransform
 from skimage.transform import (estimate_transform, matrix_transform,
                                EuclideanTransform, SimilarityTransform,
--- a/skimage/transform/tests/test_geometric.py
+++ b/skimage/transform/tests/test_geometric.py
@@ -9,6 +9,7 @@ from skimage.transform import (estimate_transform, matrix_transform,
 
 from skimage._shared import testing
 from skimage._shared.testing import assert_equal, assert_almost_equal
+import textwrap
 
 
 SRC = np.array([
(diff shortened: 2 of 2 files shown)


soupault:
Hi @Erotemic,
Thanks for a thorough PR! I like your idea, but, probably, we should think a bit more about the implementation...
Looking at how such methods work in `sklearn`, my suggestion would be to have multiline repr (1st line similar to the functionality of `sklearn` - classname + the arguments used for instantiation, 2nd-... lines - names and values of the useful public and private attributes). Some kind of fusion of `SklearnEstimator.__repr__` and `UsualClass.__dict__`:
```
In [14]: class Someclass(): 
    ...:     def __init__(self, coeff):
    ...:         self._coeff = coeff
    ...:         self.params = [[1,2,3], [4,5,6]]
    ...:
    ...:     def __repr__(self):
    ...:         #...
    ...:         pass
    ...:

In [15]: s = Someclass(coeff=1)

In [16]: repr(s)
Out[16]: Someclass(coeff=1) at 0x0123456789ab,
               'params': [[1,2,3], [4,5,6]]
```
How do you feel about this?

stefanv:
From https://docs.python.org/3.4/library/functions.html#repr:

```
**repr(object)**

    Return a string containing a printable representation of an object. For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to eval()
```

soupault:
Thanks @stefanv ! :)
These two descriptions are, probably, more straightforward:
https://docs.python.org/3.7/reference/datamodel.html#object.__repr__
https://docs.python.org/3.7/reference/datamodel.html#object.__str__

So, basically, `__repr__` must comprise of instantiation signature only. For `__str__` we are free to implement any of the suggestions above (I'm still in favour of mine).

Erotemic:
@soupault and @stefanv lets definitely discuss! 

Some thoughts: 
* The docs on __repr__ only suggest that it return a exec-able string. I'm not against it being exec-able, but I do want to note that its not a hard requirement. For instance, the default object __repr__ is non-exec-able.
* I'm tend to like one-line __repr__ and __str__ functions best. 

How about something like this?

```
<skimage.transform.ProjectiveTransform(([[1.,0.,0.],[0.,1.,0.],[0.,0.,1.]]))>
```

hmaarrfk:
You know, I just changed all of my code to **not** use the `Transforms` from scikit-image because I felt they were too opaque. I would welcome this change. That said, I would prefer it if the parameters were printed in a matrix form

```
<skimage.transform._geometric.ProjectiveTransform(
  [[1.0, 0.0, 0.0], 
   [0.0, 1.0, 0.0],
   [0.0, 0.0, 1.0]]) at 0x7fc6d65e4cf8>
```

This at least makes me **see** the homography matrix, and I can physically see it move.

hmaarrfk:
I'm just going to say that I don't think anybody is going to be executing code from reprs. It just isn't happening.

I think we can either follow scikit-learn or xarray in terms of how they display things:

```
In [5]: print(repr(xr.DataArray(np.zeros((5, 5)))))                                                  
<xarray.DataArray (dim_0: 5, dim_1: 5)>
array([[0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.]])
Dimensions without coordinates: dim_0, dim_1

In [6]: print(str(xr.DataArray(np.zeros((5, 5)))))                                                   
<xarray.DataArray (dim_0: 5, dim_1: 5)>
array([[0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.]])
Dimensions without coordinates: dim_0, dim_1
```

It might be interesting to "interpret" the matrix, say give users what the translation/rotation/skew/scaling are. We can leave that in a second PR.

Erotemic:
Somewhat related to @hmaarrfk's point: I'd wager that the most common use of `repr` is in IPython displays. 

@hmaarrfk I like the idea of following xarray, but I'm also unfamiliar with the module. Is it really useful in this case to display anything more than `(shape=(5, 5))`, where the current `dim` syntax is? Do we even need the shape? Its usually going to be small shape, so I don't see how that's useful. How about something like: 

```
<ProjectiveTransform(
  [[1., 0., 0.], 
   [0., 1., 0.],
   [0., 0., 1.]]) at 0x7fc6d65e4cf8>
```

Also, ICYDK: The technical term for "interpreting a matrix" is to "decompose a matrix".

I think it might be nice to override the `__repr__` method on subclasses so it displayed the parameters rather than the matrix: e.g. `<AffineTransform(translation=[0., 0.], scale=[1., 1.], skew=0., radians=0.) at 0x7fc6d65e4cf8>`. However, I agree with you that we should leave that for a second PR.

hmaarrfk:
I feel like your latest suggestion sacrifices readability by "trying to compatible". 
That said, it is MUCH better than having a matrix unravelled.

Xarray is a pretty cool way to have "meta data" follow your numpy array. 

```python
In [5]: DataArray(np.zeros((3, 3)), dims=['y', 'x'], coords={'x': range(3), 'y':range(1, 4)})            
Out[5]: 
<xarray.DataArray (y: 3, x: 3)>
array([[0., 0., 0.],
       [0., 0., 0.],
       [0., 0., 0.]])
Coordinates:
  * x        (x) int64 0 1 2
  * y        (y) int64 1 2 3
```

!! Thanks for the math refresher :D !!

JDWarner:
I agree with @hmaarrfk, making `repr`'s output as human readable as reasonably possible is ideal, and from my standpoint we can safely disregard any concerns about executing what is returned by `repr`.

Erotemic:
@hmaarrfk, was the `ProjectiveTransform` or `AffineTransform` transformation the one you thought was less human readable? I fail to see how that is the case for either of those suggestions. 

How would you modify them to be more readable? Remove the hex address? I tend to like having that because it quickly lets me know if two transforms are the same object or not. I can see how the hex address might not be as useful to others.

hmaarrfk:
I just think you should describe what the matrix means

in our case it is `params`

so 

```
params=   [[1., 0., 0.], 
         [0., 1., 0.],
        [0., 0., 1.]]
````
you know, with indentation.

**Comment by Erotemic** [comment]
So:

```
<ProjectiveTransform(params=[
   [1., 0., 0.], 
   [0., 1., 0.],
   [0., 0., 1.]]) at 0x7fc6d65e4cf8>
```
?

**Comment by hmaarrfk** [comment]
Sure, for me, as long as I can "see a square" I'm good :D

don't let this "flavour of the week"

I would go for one of these two
```
<ProjectiveTransform(params=[[1., 0., 0.], 
                             [0., 1., 0.],
                             [0., 0., 1.]]) at 0x7fc6d65e4cf8>

<skimage.transforms.ProjectiveTransform at 0x7fc6d65e4cf8>
params = [[1., 0., 0.], 
          [0., 1., 0.],
          [0., 0., 1.]]
```

but my requirement is more that a 3x3 matrix should be on 3 lines :D.

**Comment by Erotemic** [comment]
I also like having a 3x3 matrix that spans 3 lines. My latest push generates reprs that look like this:

```
<AffineTransform(matrix=[
    [0.38942439, 0.12769202, 0.28326154],
    [0.58025116, 0.96227529, 0.20424059],
    [0.96858298, 0.93697663, 0.25709658]]) at 0x7f88b3fe69f8>
```

and strings that look like this
```
<AffineTransform(matrix=[
    [0.38942439, 0.12769202, 0.28326154],
    [0.58025116, 0.96227529, 0.20424059],
    [0.96858298, 0.93697663, 0.25709658]])>
```

Some decisions I made: 
* I ended up using `matrix` instead of `params` because it more accurately reflects the `kwargs` accepted by the classes. 
* I opted to use 4 lines instead of 3 because it makes it simpler to keep the columns aligned. If I used 3 lines I would have to know the length of the prefix string, which will change depending on which class is creating the string. I also think its easier to read.  

Note that the latest push breaks the test (which I haven't updated yet), I will fix that once the general formatting gets ok-ed.

**Comment by jni** [comment]
Thanks everyone for the discussion here and @Erotemic for your contribution! I just want to chime with my 👍 for this output, with perhaps the only exception that I might prefer the opening bracket for the matrix to be on the same line as the start of the matrix. ie:

```
<AffineTransform(matrix=
   [[0.38942439, 0.12769202, 0.28326154],
    [0.58025116, 0.96227529, 0.20424059],
    [0.96858298, 0.93697663, 0.25709658]])>
```

Just because that's how arrays are usually represented. However I see the role of the bracket in implying the line continuation, so I don't mind either approach.

**Comment by hmaarrfk** [comment]
@jni, now that makes it really not python friendly syntax.

**Comment by jni** [comment]
@hmaarrfk 

> now that makes it really not python friendly syntax.

No it doesn't! Since we are inside the `(` of the class constructor, line continuation is implied. iirc PEP8 is quite ambiguous about the best way to do line continuations. In this case, I think it makes sense to have the matrix nicely encapsulated.

**Comment by hmaarrfk** [comment]
mind blown
```
In [4]: print('hello', end= 
   ...: 'alsdkfja')                                                                                     
helloalsdkfja
```

soupault:
:+1: to @jni 's alternative. Otherwise, I feel quite good about the latest format.

Erotemic:
Cool, I will fixup the tests and then this will be ready for merge.

Erotemic:
I think this PR is all set for merge. I've rebased on master and the tests seem to past, but it looks like `test_imread` is broken on Python 3.5 on travis only? Is this an issue on master as well?

jni:
@Erotemic yes, that issue has just been fixed in #3960 and is unrelated to your work, don't worry about it! =)

jni:
I still prefer the opening square bracket in a new line, but I'll leave that for a future PR. Thank you very much @Erotemic, this is a *great* usability improvement! =)

Erotemic:
I will put up another PR for adjusting that square bracket. 

The fact that moving the square bracket to a new line is indeed valid Python (\cite the mind-blowing demonstration by @hmaarrfk) invalidates my argument for my format. I'd rather not force a design choice based on a feel-fact.

PR #3528, opened 2018-11-01
Better gallery examples and tests for masked translation registration

LaurentRDC:
## Description
I would like to propose an extension of #3334.

The original publication [D. Padfield. Masked object registration in the Fourier domain. IEEE Transactions on Image Processing (2012)](http://www.dirkpadfield.com/Home/MaskedObjectRegistrationInTheFourierDomain.pdf?attredirects=0) shows the algorithm at work. I would like to include an example and tests similar to Figure 5.

The images used in Figure 5 are available on the original author's website. The data is also used in tests (see MATLAB code [[Link to zip](http://www.dirkpadfield.com/Home/MaskedFFTRegistrationCode.zip)]. The main tests have been implemented in this PR.

I'd like to go further and include a gallery example of figure 5 as well. We would be killing two birds with one stone by adding theses images in `skimage`: 
* Better usage examples in the gallery;
* Better tests.

Since this data is specific to masked translation registration, I'm hesitating adding it to the `skimage.data` module. However, for the data to be available both in tests and in gallery examples, I see no other alternatives. 

Let me know what you think.

## Checklist
  existing benchmark
- [x] Unit tests

## For reviewers

  later.


stefanv:
Thank you, @LaurentRDC, I think these would be excellent additions.  The only thing we'd need to check is usage rights to those images; as far as I know, Dirk did not provide explicit copyright statements, but I think the intent was to make the software and data open.  I will email him to clarify the dataset license.

LaurentRDC:
@stefanv Thank you for taking care of this.

What do you think about keeping the images in the `skimage/feature/tests/images`? These aren't good test images for other purposes, so I wouldn't want to put them in `skimage/data`.

soupault:
@LaurentRDC so far the policy was to locate such images under `module/tests/data/` (see e.g. https://github.com/scikit-image/scikit-image/tree/master/skimage/color/tests/data).

LaurentRDC:
@soupault Great, I will do the same.

I'd also like to show these images in gallery example. Do you think it is appropriate to store these images in `doc/examples/transform/data` as well?

stefanv:
If the images are used in a gallery example, you have to expose them to users.  Then, the `data` module is the correct place for them.

sciunto:
To complete @stefanv 's comment, images must also be displayed in http://scikit-image.org/docs/dev/auto_examples/index.html#data (more specifically in scientific images).

LaurentRDC:
I've decided against adding more to the `skimage.data` module. The added gallery example makes use of existing images.

stefanv:
Via email, Dirk confirmed that we are able to distribute these images publicly.

sciunto:
thanks @stefanv !

LaurentRDC:
@stefanv Thank you for taking care of this. 

Using these images, we can be confident that the scikit-image implementation is functionally equivalent to the original author's MATLAB code.

sciunto:
@LaurentRDC Would you like to stop on this PR and do not work on moving this in the gallery (as stated in the title) or do you plan to continue?

If you don't we will fill another issue for the gallery.

thanks!

LaurentRDC:
@sciunto From my comment above:

> I've decided against adding more to the skimage.data module. The added gallery example makes use of existing images.

I've reproduced the spirit of the original paper using existing images in `skimage.data` and making masks using `skimage.draw`. With this, we've fulfilled the purpose of this PR.

doc/examples/transform/plot_masked_register_translation.py
```diff
@@ -47,9 +47,9 @@
 # the pixels ourselves
 mask = corrupted_pixels
 
-shift = masked_register_translation(image, offset_image, mask)
+detected_shift = masked_register_translation(image, offset_image, mask)
 
-print("Detected pixel offset (y, x): {}".format(shift))
+print("Detected pixel offset (y, x): {}".format(-detected_shift))
```
emmanuelle: Isn't it confusing to print the opposite value of the value returned by the function?

LaurentRDC: This point was [discussed](https://github.com/scikit-image/scikit-image/pull/3334#discussion_r231680321) at the end of #3334.

Just like the `register_translation` function, the shift returned by `masked_register_translation` is the shift required to _shift back_. I personally find this confusing to see in the a gallery example (see gallery example for `register_translation` [here](http://scikit-image.org/docs/stable/auto_examples/transform/plot_register_translation.html#sphx-glr-auto-examples-transform-plot-register-translation-py)). 

Maybe, instead of calling it _detected offset_, we should call it _corrective offset_?

jni: I'm ambivalent about all this (I can see arguments for either convention), but I would like to ask that `y, x` be changed to `r, c`. See [coordinate conventions](http://scikit-image.org/docs/dev/user_guide/numpy_images.html#coordinate-conventions).

LaurentRDC: @jni good suggestion, I've changed it in latest commit

stefanv:
Good work, thank you!

PR #3530, opened 2018-11-02
Add copybutton

thoo:
Close ##3482. 
Add sphinx-copybutton so that the code block generated by sphinx can be copied by a single click. 
More details about sphinix-copybutton can be found [here](https://github.com/choldgraf/sphinx-copybutton).
 Here is how it looks with the click button at the upper right corner.

### DIFF
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -333,10 +333,6 @@ the documentation.
     pip install sphinx_gallery
     pip install sphinx-copybutton
 
-`copybutton file <https://github.com/choldgraf/sphinx-copybutton/blob/master/sphinx_copybutton/_static/copybutton.css>`__ is modified to generate the online documentation by
-adding ``background: transparent no-repeat none; border: none;`` and
-adjusting ``margin: 0.5px; height: 1.6em; padding: 0 2px 0px;`` at ``a.copybtn``.
-
 **LaTeX Ubuntu:**
 
 .. code:: sh
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -330,6 +330,12 @@ the documentation.
 .. code:: sh
 
     pip install sphinx
+    pip install sphinx_gallery
+    pip install sphinx-copybutton
+
+`copybutton file <https://github.com/choldgraf/sphinx-copybutton/blob/master/sphinx_copybutton/_static/copybutton.css>`__ is modified to generate the online documentation by
+adding ``background: transparent no-repeat none; border: none;`` and
+adjusting ``margin: 0.5px; height: 1.6em; padding: 0 2px 0px;`` at ``a.copybtn``.
 
 **LaTeX Ubuntu:**
 
--- a/CONTRIBUTING.txt
+++ b/CONTRIBUTING.txt
@@ -317,7 +317,19 @@ Building docs
 -------------
 
 To build docs, run ``make`` from the ``doc`` directory. ``make help`` lists
-all targets.
+all targets. For example, to build the HTML documentation, you can run:
+
+.. code:: sh
+
+    make SPHINXOPTS= html
+
+Then, all the HTML files will be generated in ``scikit-image/doc/build/html/``.
+To rebuild a full clean documentation, run:
+
+.. code:: sh
+
+    make clean
+    make SPHINXOPTS= html
 
 Requirements
 ~~~~~~~~~~~~
@@ -327,11 +339,12 @@ the documentation.
 
 **Sphinx:**
 
+Sphinx and other python packages needed to build the documentation
+can be installed using ``scikit-image/requirements/docs.txt`` file.
+
 .. code:: sh
 
-    pip install sphinx
-    pip install sphinx_gallery
-    pip install sphinx-copybutton
+    pip install -r requirements/docs.txt
 
 **LaTeX Ubuntu:**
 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.txt
```diff
@@ -330,6 +330,12 @@ the documentation.
 .. code:: sh
 
     pip install sphinx
+    pip install sphinx_gallery
+    pip install sphinx-copybutton
+
+`copybutton file <https://github.com/choldgraf/sphinx-copybutton/blob/master/sphinx_copybutton/_static/copybutton.css>`__ is modified to generate the online documentation by
+adding ``background: transparent no-repeat none; border: none;`` and
+adjusting ``margin: 0.5px; height: 1.6em; padding: 0 2px 0px;`` at ``a.copybtn``.
```
jni: I don't understand why this comment is here?

thoo: @jni  I am also really glad to see the click button. Thanks for the suggestion. I have to use the custom css  ( modified CSS in sphinx-copybutton) to generate a nicer and transparent click-button because the button blocks some text in a few cases. Also I am not sure how the online document is generated. Should I point the forked repository with the custom css file for sphinx-copybutton installation.

thoo:
@jni All checks passed. Thanks for the review.

CONTRIBUTING.txt
```diff
@@ -330,6 +330,8 @@ the documentation.
 .. code:: sh
 
     pip install sphinx
```
stefanv: Let's replace this with:

```
pip install -r requirements/docs.txt
```

thoo: @stefanv Thanks for the review. Now it should be consistent when anyone runs ` make SPHINXOPTS= html`. I should have looked for a custom css file in the first place. :)

thoo:
@jni Can you re-review this?

CONTRIBUTING.txt
```diff
@@ -317,7 +317,19 @@ Building docs
 -------------
 
 To build docs, run ``make`` from the ``doc`` directory. ``make help`` lists
-all targets.
+all targets. For example, to build the HTML documentation, you can run:
+
+.. code:: sh
+
+    make SPHINXOPTS= html
```
stefanv: Not just `make html` ?

CONTRIBUTING.txt
```diff
@@ -327,9 +339,12 @@ the documentation.
 
 **Sphinx:**
 
+Sphinx and other python packages needed to build the documentation
+can be installed using ``scikit-image/requirements/docs.txt`` file.
```
stefanv: can be installed using:

doc/source/themes/scikit-image/static/css/custom.css
```diff
@@ -298,3 +298,15 @@ div.sphx-glr-download a:hover {
 p.sphx-glr-signature a.reference.external {
   display: none !important;
 }
+
+a.copybtn {
+    background: transparent no-repeat none;
+    border:none;
```
stefanv: inconsistent spacing: `border: none;`

doc/source/themes/scikit-image/static/css/custom.css
```diff
@@ -298,3 +298,15 @@ div.sphx-glr-download a:hover {
...
+    position: absolute;
+    top: 0;
+    right: 0;
+    margin: 0.5px !important;
```
stefanv: Why is the `!important` necessary here?

jni:
@thoo other than addressing @stefanv's comments, this looks good to me!

doc/source/themes/scikit-image/static/css/custom.css
thoo: I am not familiar with how sphinx and its extensions work. But  my css style won't change if I don't use `!important`. `sphinx-gallery` extension also used `!important` to enforce its style in `custom.css`.

CONTRIBUTING.txt
thoo: If I just run `make html`, I got the following error on my mac:
```
python tools/build_modref_templates.py
/Users/thein/hacktoberfest/scikit-image/skimage/novice/__init__.py:103: UserWarning: The `skimage.novice` module was deprecated in version 0.14. It will be removed in 0.16.
  warnings.warn("The `skimage.novice` module was deprecated in version 0.14. "
WARNING: Empty - skimage.scripts
outdir:
27 files written
Build API docs...done.
Copying release notes
make: nproc: Command not found
python -m sphinx -b html -d build/doctrees  -j  source build/html
usage: __main__.py [OPTIONS] SOURCEDIR OUTPUTDIR [FILENAMES...]
__main__.py: error: argument -j: invalid jobs_argument value: 'source'
make: *** [html] Error 2
```

stefanv:
It looks like you are running RedHat or a secured system that protects `nproc` (discovery of number of CPUs).  This is probably a bug in your `make` package—can you try and upgrade that?

The easiest quickfix on your system is probably to type:

```bash
nproc () {
echo 4
}
```

before doing `make html`.

thoo:
@stefanv Thanks. I didn't know `nproc` wasn't available on Mac OS X. I have to use `sysctl -n hw.ncpu` to make it work with `make html`. Could you also re-review this? Thanks.

emmanuelle:
I also checked on my machine and it looks good. Merging!

PR #3534, opened 2018-11-03
Get rid of the requirements-parser dependency

hmaarrfk:
I got rid of it in my own projects because when debugging, you things to be able to run on any python install.

Even that lightweight dependency is too much.

```
docs.txt
              sphinx is not installed
            numpydoc is not installed
      sphinx-gallery is not installed
       pytest-runner is not installed
        scikit-learn is not installed
default.txt
               numpy 1.15.3
               scipy 1.1.0
          matplotlib 3.0.1
            networkx 2.2
              pillow 5.3.0
             imageio 2.3.0
          PyWavelets 1.0.1
         dask[array] 0.20.0
         cloudpickle 0.6.1
build.txt
              Cython 0.29
               wheel 0.32.2
               numpy 1.15.3
test.txt
              pytest 3.9.3
          pytest-cov is not installed
              flake8 is not installed
             codecov is not installed
optional.txt
              imread is not installed
           SimpleITK is not installed
             astropy is not installed
            tifffile is not installed
                qtpy is not installed
               pyamg is not installed

```

Output is unchanged
## Checklist
  existing benchmark
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/tools/build_versions.py
+++ b/tools/build_versions.py
@@ -12,11 +12,22 @@ def main():
         print(requirement_file.name)
         with open(str(requirement_file), 'r') as f:
             for req in f:
+                # Remove trailing and leading whitespace.
                 req = req.strip()
-                if req[0] == '#':
+
+                if not req:
+                    # skip empty or white space only lines
                     continue
+                elif req.startswith('#'):
+                    continue
+
+                # Get the name of the package
                 req = re.split('<|>|=|!|;', req)[0]
                 try:
+                    # use pkg_resources to reliably get the version at install time
+                    # by package name. pkg_resources needs the name of the package
+                    # from pip, and not "import". 
+                    # e.g. req is 'scikit-learn', not 'sklearn'
                     version = pkg_resources.get_distribution(req).version
                     print(req.rjust(20), version)
                 except pkg_resources.DistributionNotFound:
--- a/tools/build_versions.py
+++ b/tools/build_versions.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
 
 from pathlib import Path
-import requirements
 import pkg_resources
+import re
 
 
 def main():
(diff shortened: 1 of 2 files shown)


hmaarrfk:
Seems like the CIs just kernel panicked

jni:
I'm :-1: on this change. I 100% prefer to have the blacklist justification right next to the requirement, and on top of it it adds complexity to the setup.py.

If you prefer to *vendor* requirements-parser, I might be happier with that.

hmaarrfk:
@jni, I removed the blacklist justification change. That requirement seems overkill for a simple regexp.

tools/build_versions.py
```diff
@@ -1,22 +1,26 @@
...
     requirements_dir = Path(__file__).parent / '..' / 'requirements'
 
     for requirement_file in requirements_dir.glob('*.txt'):
         print(requirement_file.name)
-        with open(str(requirement_file), 'r') as fd:
-            for req in requirements.parse(fd):
+        with open(str(requirement_file), 'r') as f:
+            for req in f:
+                req = req.strip()
+                if req[0] == '#':
```
jni: This will give an IndexError if we encounter a blank line. On the one hand, I suggest replacing with `if req.startswith('#')`. On the other, I'll take this opportunity to point out that robustness to this kind of error is the kind of thing that a dependency buys you.

tools/build_versions.py
```diff
@@ -1,22 +1,26 @@
...
+                    continue
+                req = re.split('<|>|=|!|;', req)[0]
                 try:
-                    version = pkg_resources.get_distribution(req.name).version
-                    print(req.name.rjust(20), version)
+                    version = pkg_resources.get_distribution(req).version
+                    print(req.rjust(20), version)
```
jni: Does this code not care at all about the version limiters that are in the file? The new code doesn't, but it looks like the old code didn't either? Is that what we want?

jni: Oh I see, this is just printing what's installed. Sorry, I didn't read this PR carefully and thought all this logic was part of the setup.py. I'm coming around to your point of view. =)

stefanv:
:+1:

hmaarrfk:
@jni, no problem. I sprinkled in a few comments since it was pretty obscur otherwise.

hmaarrfk:
Juan. I think that was my fault introducing changes that were not material
to the core of the idea.

Thanks for the detailed review.

hmaarrfk:
Nice. At least now we can tell people to run that script to tell us what they have installed when they are having trouble installing from source

PR #3541, opened 2018-11-09
HOG returns incorrect features if pixels_per_cell is odd

t-ae:
## Description

`skimage.feature.hog` was ignoring last row/column of cell if `pixels_per_cell` is odd number.

Suppose `pixels_per_cell == (3, 3)`

In here `cell_rows == pixels_per_cell[0]`.
https://github.com/scikit-image/scikit-image/blob/9c4632f43eb6f6e85bf33f9adf8627d01b024496/skimage/feature/_hoghistogram.pyx#L120-L127

```
range_rows_stop = cell_rows / 2 = 1
range_rows_start = -range_rows_stop = -1
```

`cell_rows` is 3 but `[range_row_start, range_rows_stop)` contains only 2 integers, so last row will be ignored.
I fixed `range_rows_stop` (and `range_columns_stop`) calculation.

## Checklist
  existing benchmark
- [x] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/feature/_hoghistogram.pyx
+++ b/skimage/feature/_hoghistogram.pyx
@@ -121,10 +121,10 @@ def hog_histograms(double[:, ::1] gradient_columns,
     c_0 = cell_columns / 2
     cc = cell_rows * number_of_cells_rows
     cr = cell_columns * number_of_cells_columns
-    range_rows_stop = cell_rows / 2
-    range_rows_start = -range_rows_stop
-    range_columns_stop = cell_columns / 2
-    range_columns_start = -range_columns_stop
+    range_rows_stop = (cell_rows+1) / 2
+    range_rows_start = -(cell_rows / 2)
+    range_columns_stop = (cell_columns + 1) / 2
+    range_columns_start = -(cell_columns / 2)
     number_of_orientations_per_180 = 180. / number_of_orientations
 
     with nogil:
--- a/skimage/feature/_hoghistogram.pyx
+++ b/skimage/feature/_hoghistogram.pyx
@@ -121,7 +121,7 @@ def hog_histograms(double[:, ::1] gradient_columns,
     c_0 = cell_columns / 2
     cc = cell_rows * number_of_cells_rows
     cr = cell_columns * number_of_cells_columns
-    range_rows_stop = (cell_rows+1) / 2
+    range_rows_stop = (cell_rows + 1) / 2
     range_rows_start = -(cell_rows / 2)
     range_columns_stop = (cell_columns + 1) / 2
     range_columns_start = -(cell_columns / 2)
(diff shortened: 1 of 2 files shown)


t-ae:
By the way I wanna add test code.
But [existing test](https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/tests/test_hog.py) is using prepared npy file.
How should I get the correct features?

emmanuelle:
Thanks for noticing this bug. The change seems to be correct. For adding tests, maybe you could use a very small image (like ``a = np.ones((9, 9)); a[-1] = 1; feature.hog(a, pixels_per_cell=(3, 3)``) and check that the result is different with your change (ie, the last line is taken into account).

t-ae:
The bug is that last row/column are skipped while scanning pixels.
To verify the problem, I added a test case about the image below.

```
[[0, 0, 0],
 [0, 0, 0],
 [0, 0, 1]]
```

The output feature will be non-zero if last row/column are not skipped.
(non-zero elements should be 0.5 but they appear as 0.4999775, maybe because `eps` is added while [L1 normalization](https://github.com/scikit-image/scikit-image/blob/f451bc0f6e3586fd66c7ddec84cbd15e168205e0/skimage/feature/_hog.py#L8).)

Now it's ready to be merged.

emmanuelle:
Great, I love this new test. Can you just remove all keyword arguments which have default values (orientations, etc.) and are therefore not needed, in order to keep a small codebase even in tests? @scikit-image/core this one is almost ready!

skimage/feature/_hoghistogram.pyx
```diff
@@ -121,10 +121,10 @@ def hog_histograms(double[:, ::1] gradient_columns,
     c_0 = cell_columns / 2
     cc = cell_rows * number_of_cells_rows
     cr = cell_columns * number_of_cells_columns
-    range_rows_stop = cell_rows / 2
-    range_rows_start = -range_rows_stop
-    range_columns_stop = cell_columns / 2
-    range_columns_start = -range_columns_stop
+    range_rows_stop = (cell_rows+1) / 2
```
Borda: add space around `+`

skimage/feature/tests/test_hog.py
```diff
@@ -51,6 +51,20 @@ def test_hog_image_size_cell_size_mismatch():
     assert len(fd) == 9 * (150 // 8) * (200 // 8)
 
 
+def test_hog_odd_cell_size():
+    img = np.zeros((3, 3))
+    img[2, 2] = 1
+
+    correct_output = np.zeros((9, ))
+    correct_output[0] = 0.4999775
```
Borda: would rather set the value to `0.5` as it may be more intuitive and later in the test set `decimal` precision to `1` which would become also more robust for test on different OS
https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.testing.assert_almost_equal.html

t-ae: Updated to use `decimal=1`.
It got better, thank you!

PR #3545, opened 2018-11-13
Backport #3447: Fix morphology.local_maxima for input with any dimension < 3

stefanv:
* Fix local_maxima for input with dimensions < 3

local_maxima (and local_minima) failed with an ValueError for images
with at least 1 dimension with fewer than 3 elements. This was due to
the implementation which applies a structuring element with exactly 3
elements in each dimension to the image in question. This behavior was
only visible if `allow_borders` was false, otherwise the image was
padded internally.
This fix makes these functions work for the mentioned edge cases by
testing the image shape at the appropriate time.

* Fix output of local_maxima for empty arrays

The old output for empty arrays if `indices` was true didn't correspond
to the output for non-empty arrays and the description in the docstring.
This ensures that the output will always be a tuple of empty arrays.

If `indices` is false, the returned empty array now machtes the shape of
`image` as well.

* Show warning if no maxima are possible

which is the case for dimensions with fewer than 3 samples. Because
`local_minima` wraps `local_maxima` the warning can only set the correct
stacklevel for one of these functions. The latter was chosen because it
stands to reason that it is / will be used more frequently.

If may be possible to suppress and re-raise the warning with the context
manager from the warnings module but that would introduce non-thread-
safe code and introduce perhaps unnecessary complexity.

* Point user to allow_borders in warning message

## Description

## Checklist
  existing benchmark
- [ ] Unit tests

## References

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/extrema.py
+++ b/skimage/morphology/extrema.py
@@ -13,6 +13,8 @@ from __future__ import absolute_import
 
 import numpy as np
 
+from ..util import dtype_limits, invert, crop
+from .._shared.utils import warn
 from . import greyreconstruct
 from ..util import dtype_limits
 
@@ -279,14 +281,22 @@ def local_maxima(image, selem=None):
 
     The resulting image will contain all 6 local maxima.
     """
-    # find the minimal grey level difference
-    h = _find_min_diff(image)
-    if h == 0:
-        return np.zeros(image.shape, np.uint8)
-    if not np.issubdtype(image.dtype, np.floating):
-        h = 1
-    local_max = h_maxima(image, h, selem=selem)
-    return local_max
+    if selem is None:
+        if connectivity is None:
+            connectivity = ndim
+        selem = ndi.generate_binary_structure(ndim, connectivity)
+    else:
+        # Validate custom structured element
+        selem = np.asarray(selem, dtype=bool)
+        # Must specify neighbors for all dimensions
+        if selem.ndim != ndim:
+            raise ValueError(
+                "structuring element and image must have the same number of "
+                "dimensions"
+            )
+        # Must only specify direct neighbors
+        if any(s != 3 for s in selem.shape):
+            raise ValueError("dimension size in structuring element is not 3")
 
 
 def local_minima(image, selem=None):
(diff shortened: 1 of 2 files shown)


jni:
Did you miss some commits or did we forget to backport a prerequisite PR? These changes clearly don't make sense on their own. See:
https://github.com/scikit-image/scikit-image/pull/3545/files#diff-1b49cb33a2ea212f0677bfe8ba6f3fcdR302

definition only has image and selem, but this PR appears to add many more.

stefanv:
I guess there must be other PRs that were missed; I wasn't tracking.

lagru:
> Did you miss some commits or did we forget to backport a prerequisite PR? 

Yeah, you missed #3022 which introduces the rewrite of `local_maxima`. I'd be willing to help. They way I understand it I'd just need to do a PR with the commits in question against `scikit-image:v0.14.x`. Correct?

jni:
Ooops! I can see that we even tried the auto backport and then totally dropped the ball when it failed. Yes @lagru that's indeed what's required. Maybe you can pile these on as well onto the same PR? =)

lagru:
Happy to do so. However maybe you give me some tips so I do this the right way from the start:

My first impulse is to cherry-pick the commits in these two PRs (sans merge commits?) onto a new branch (e.g. `backport_v0.14.x/3022_3447)` which is based on `v0.14.x`. But it looks like I'd have to do so for every commit in #3022 of which there are 35. Is there a better less error-prone way?

jni:
`git cherry-pick start_commit..end_commit` =)

lagru:
Thanks! Why didn't I think of that? :see_no_evil:

jni:
@lagru because git man-pages are appallingly bad. =D I highly recommend `tldr`:

```
 $ tldr git-cherry-pick
git cherry-pick
Apply the changes introduced by existing commits to the current branch.
To apply changes to another branch, first use git-checkout to switch to the desired branch.

 - Apply a commit to the current branch:
   git cherry-pick {{commit}}

 - Apply a range of commits to the current branch (see also 
   git rebase --onto
):
   git cherry-pick {{start_commit}}~..{{end_commit}}

 - Apply multiple (non-sequential) commits to the current branch:
   git cherry-pick {{commit_1}} {{commit_2}}
```

Incidentally, it looks like my advice was slightly wrong... There should have been a `~` after the first commit. Just tested it out and indeed it needs to be there. Looks like it's a half-open interval on the start side? Weird. Can you double check that you got all the commits you meant to in #3546? Or did you figure this out on your own in the end?

lagru:
@jni Yes, The PR has all commits except for the merge-commits.
I was slightly surprised that I got a merge conflict for the second cherry-picked commit when there shouldn't have been one. So I figured its an half-open interval and just specified the parent commit. :smile: 

> I highly recommend tldr

Thank you! This is gold!

PR #3563, opened 2018-11-25
Use 'getconf _NPROCESSORS_ONLN' as fallback for nproc in Makefile of docs

### LINKED ISSUE #3562: Building docs fails on macOS - nproc not found
## Description
Running `make` in the `doc` directory fails on macOS as the `Makefile` uses `nproc` to find the number of processors, but it wasn't installed. Full error below:

```
python tools/build_modref_templates.py
*WARNING* API documentation not generated: Installed version does not match source version
Build API docs...done.
Copying release notes
make: nproc: Command not found
python -m sphinx -b html -d build/doctrees  -j  source build/html
Sphinx v1.6.6
Usage: __main__.py [options] sourcedir outdir [filenames...]

Filename arguments:
  without -a and without filenames, write new and changed files.
  with -a, write all files.
  with filenames, write these.

__main__.py: error: option -j: invalid integer value: 'source'
make: *** [html] Error 2
```

I think this can be fixed by checking if `nproc` is available before using it, and if not, use some other method to find number of processors.

Python version - 3.6.4
OS - macOS Mojave - 10.14.1

PinkFloyded:
## Description
Use `getconf _NPROCESSORS_ONLN` as fallback for `nproc` as few systems don't have `nproc`.

## References
Closes #3562

## For reviewers

  later.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 PYTHON        ?= python
-SPHINXOPTS    ?= -j $(shell nproc)
+SPHINXOPTS    ?= -j $(shell nproc || getconf _NPROCESSORS_ONLN)
 SPHINXBUILD   ?= $(PYTHON) -m sphinx
 SPHINXCACHE   ?= build/doctrees
 PAPER         ?=
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 PYTHON        ?= python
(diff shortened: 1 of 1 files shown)


PinkFloyded:
Thank you!

Good point. It is possible because even `_NPROCESSORS_ONLN` is not POSIX mandated, so it might not be present. Will add it in.

sciunto:
Thanks @PinkFloyded !

thenickdude:
The "|| 1" doesn't actually work, that attempts to run a command called "1". You want "echo 1"

Issue #3571, opened 2018-11-29
Some failures in travis py3.6 for random walker


sciunto:
## Description

Se for instance: https://travis-ci.org/scikit-image/scikit-image/jobs/461265535

Not the first time I notice it.

[If reporting a bug, attach the entire traceback from Python.]
[If proposing an enhancement/new feature, provide links to related articles, reference examples, etc.]

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [ ] Code example
- [ ] Relevant images (if any)
- [ ] Operating system and version
- [ ] Python version
- [ ] scikit-image version (run `skimage.__version__`)

PR #3572, opened 2018-12-01
Filtered out DeprecationPendingWarning for matrix subclass

emmanuelle:
This PR should fix Travis, current failure is due to uncaught PendingDeprecationWarning raised by scipy sparse operations. Scipy will filter out these warnings in the next release, so it will be possible to remove this cumbersome warnings logics in the future (this was already mentioned in TODO.txt, therefore I didn't have to add it).

### DIFF
--- a/skimage/segmentation/tests/test_random_walker.py
+++ b/skimage/segmentation/tests/test_random_walker.py
@@ -19,7 +19,7 @@ if (Version(np.__version__) >= '1.15.0' and
         Version(scipy.__version__) <= '1.1.0'):
     NUMPY_MATRIX_WARNING = 'matrix subclass'
 else:
-    NUMPY_MATRIX_WARNING = None
+    NUMPY_MATRIX_WARNING = 'matrix subclass'
 
 
 def make_2d_syntheticdata(lx, ly=None):
--- a/skimage/segmentation/tests/test_random_walker.py
+++ b/skimage/segmentation/tests/test_random_walker.py
@@ -15,8 +15,7 @@ SCIPY_RANK_WARNING = r'numpy.linalg.matrix_rank|\A\Z'
 PYAMG_MISSING_WARNING = r'pyamg|\A\Z'
 PYAMG_OR_SCIPY_WARNING = SCIPY_RANK_WARNING + '|' + PYAMG_MISSING_WARNING
 
-if (Version(np.__version__) >= '1.15.0' and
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Nice!!! These were getting annoying

sciunto:
Must solve #3571

emmanuelle:
Hurray, travis is finally happy :-).

skimage/segmentation/tests/test_random_walker.py
```diff
@@ -15,8 +15,7 @@
 PYAMG_MISSING_WARNING = r'pyamg|\A\Z'
 PYAMG_OR_SCIPY_WARNING = SCIPY_RANK_WARNING + '|' + PYAMG_MISSING_WARNING
 
-if (Version(np.__version__) >= '1.15.0' and
-        Version(scipy.__version__) <= '1.1.0'):
```
hmaarrfk: Why is this line removed? Did scipy 1.1.0 change?

hmaarrfk: It should probably be set to < 1.2
https://scipy.github.io/devdocs/release.1.2.0.html

jni:
@hmaarrfk well, Travis is happy...

@emmanuelle maybe you could add/edit item in to-do to point to the relevant issue and PR history here? It might be very useful when all these checks break again...

jni:
I take it back, looking at the SciPy 1.2 release notes as suggested by @hmaarrfk, it does look like they will filter things (do a Ctrl-F for "8887"). @emmanuelle can you add the check for SciPy <1.2?

stefanv:
+1, after integrating the `SciPy <1.2` fix.

emmanuelle:
@jni @hmaarrfk @stefanv It does not work as shown by Travis failure for the "pip --pre" build. Indeed the warning is still raised with scipy release candidate (0.12.0rc1), I don't understand exactly why since a PR has been merged which intends to suppress such warnings, but I also checked on my machine and the warning does fire. This is why I had removed the test on scipy version until we know exactly which version of scipy solves the issue.

emmanuelle:
@hmaarrfk yes this should be investigated but since it's annoying on the short term that Travis is red, I'll revert just the commit.

hmaarrfk:
agreed

jni:
Good points @emmanuelle, I didn't realise 1.2rc was out. Thanks for handling this!

hmaarrfk:

Issue #543, opened 2013-05-05
Image conversion routines should allow users to suppress warnings


tonysyu:
The image conversion routines (i.e. `img_as_*`) should allow users to suppress warnings due to loss-of-precision/-sign. These warnings are particularly annoying when a program prints other info to stdout.

goretkin:
I skimmed this thread a few times looking for the solution, but missed it even though it was alluded to a few times. For someone else's sake, concretely, if you don't want a call to emit any warnings, you can do:

```python
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    skimage.io.imsave("test.png", all_white_image
```

mpt5:
thanks @goretkin this works nicely -- do you know if this has any major impact on performance? I will be looping across thousands of images and this will be inside the loop so I can catch other warnings. I guess best would be to have it outside, but then it would suppress other warnings.

goretkin:
@mpt5 I'm glad it helped. I don't know about the performance of suppressing and unsuppressing warnings, but that is probably something that isn't specific to this project.

kootenpv:
It's absurd to expect such a code for end users, I'd almost say it is unacceptable. Even if I want to explicitly cast, I don't see any way to avoid these messages (other than using a context manager which is super ugly in this case). Very frustrating.

soupault:
Hello @kootenpv ! Thanks for the negative feedback, but we'd strongly encourage you to approach this issue in a constructive way as well. Remind yourself, that `skimage` is an open-source library (means that we are open for the proposals), and it serves many purposes (each with different level of required control on image processing process). It might perform not as expected in your particular case, but is optimal in general.

Context manager seems like a natural method of wrapping things in Python, and I personally do not see any issues with that. We have some plans to refactor that piece of code, but, in any case, it will stay the main pain point: a bridge between a static and a dynamic typing.

jni:
@soupault "is optimal in general" seems like a very strong statement. =)

@kootenpv I think you're right regarding warnings on conversions. As @soupault hinted, we are welcome to suggestions. Here are two options off the top of my head:

- we have our own context manager that is perhaps nicer to use. We could move it to a more public space:

```python
In [4]: from skimage._shared._warnings import expected_warnings

In [5]: image = np.ones((5, 5))

In [6]: with expected_warnings(['precision']):
   ...:     image_uint8 = img_as_ubyte(image)
   ...:

In [7]:
```

[...]

Issue #3080, opened 2018-05-15
uint8 -> int16 does not have precision loss


hmaarrfk:
https://github.com/scikit-image/scikit-image/blob/05ac822219304d00426fb4e3b423c90c7b916fbf/skimage/_shared/testing.py#L130

Am I missing something or is there a bug in the above?

```python
def color_check(plugin, fmt='png'):
    """Check roundtrip behavior for color images.
    All major input types should be handled as ubytes and read
    back correctly.
    """
    img = img_as_ubyte(data.chelsea())
    # [...]
    with expected_warnings(['precision loss']):
        img4 = img_as_int(img)
```

  - `img` is a `uint8`
  - `img_as_int` says that it converts to `int16`
  - `uint8 -> int16` has no precision loss in my books.

soupault:
Doesn't seem necessary, indeed.

PR #3575, opened 2018-12-03
Remove precision and sign loss warnings from util.img_as_

### LINKED ISSUE #2602: Add warn keyword to img_as_* functions
Image conversion from float to uint8 is very, very common, and usually people know what they're doing when using it, and don't appreciate the annoying warning. A brief discussion in https://github.com/scikit-image/scikit-image/issues/543#issuecomment-292221966 suggests that users might like a `img_as_uint8(..., warn=False)` keyword argument as a way to suppress the warnings.

The same would be true for `imsave` and low contrast images.

Another idea would be to only warn when the apparent dynamic range is dramatically compressed, even when `warn=True`.

Discuss. =P

jni:
## Description
See discussions:
#543 (comment)
#2602

Specifically, @stefanv said [1]_:

> Why don't we just remove this warning entirely? It seemed like the
> proper thing to do at the time, but it's never been useful.

Therefore, I set the default to False, but leave the option for warning
there since it's potentially useful to some. When and if we stop doing
automatic dtype conversions (see #3009), we might be able to remove
warnings altogether.

..[1]: #2602 (comment) 

The PR touches a lot of lines so I suggest looking at individual commits when reviewing. Thanks @scikit-image/core!

## Checklist
  existing benchmark
- [x] Unit tests

## References
Closes #2602 

## For reviewers

  later.

### DIFF
--- a/doc/source/user_guide/data_types.rst
+++ b/doc/source/user_guide/data_types.rst
@@ -33,7 +33,7 @@ functions that convert dtypes and properly rescale image intensities (see
 `Input types`_). You should **never use** ``astype`` on an image, because it
 violates these assumptions about the dtype range::
 
-   >>> from skimage import img_as_float
+   >>> from skimage.util import img_as_float
    >>> image = np.arange(0, 50, 10, dtype=np.uint8)
    >>> print(image.astype(np.float)) # These float values are out of range.
    [  0.  10.  20.  30.  40.]
@@ -65,7 +65,7 @@ img_as_int     Convert to 16-bit int.
 These functions convert images to the desired dtype and *properly rescale their
 values*::
 
-   >>> from skimage import img_as_ubyte
+   >>> from skimage.util import img_as_ubyte
    >>> image = np.array([0, 0.5, 1], dtype=float)
    >>> img_as_ubyte(image)
    array([  0, 128, 255], dtype=uint8)
@@ -124,7 +124,7 @@ unnecessary data copies take place.
 A user that requires a specific type of output (e.g., for display purposes),
 may write::
 
-   >>> from skimage import img_as_uint
+   >>> from skimage.util import img_as_uint
    >>> out = img_as_uint(sobel(image))
    >>> plt.imshow(out)
 
@@ -160,7 +160,7 @@ If cv_image is an array of unsigned bytes, ``skimage`` will understand it by
 default. If you prefer working with floating point images, :func:`img_as_float`
 can be used to convert the image::
 
-    >>> from skimage import img_as_float
+    >>> from skimage.util import img_as_float
     >>> image = img_as_float(any_opencv_image)
 
 Using an image from ``skimage`` with OpenCV
@@ -168,7 +168,7 @@ Using an image from ``skimage`` with OpenCV
 
 The reverse can be achieved with :func:`img_as_ubyte`::
 
-    >>> from skimage import img_as_ubyte
+    >>> from skimage.util import img_as_ubyte
     >>> cv_image = img_as_ubyte(any_skimage_image)
 
 
@@ -181,7 +181,7 @@ a custom function that requires a particular dtype, you should call one of the
 dtype conversion functions (here, ``func1`` and ``func2`` are ``skimage``
 functions)::
 
-   >>> from skimage import img_as_float
+   >>> from skimage.util import img_as_float
    >>> image = img_as_float(func1(func2(image)))
    >>> processed_image = custom_func(image)
 
--- a/doc/source/user_guide/data_types.rst
+++ b/doc/source/user_guide/data_types.rst
@@ -70,22 +70,12 @@ values*::
    >>> img_as_ubyte(image)
    array([  0, 128, 255], dtype=uint8)
 
-These conversions can result in a loss of precision, since 8 bits cannot hold
-the same amount of information as 64 bits::
+Be careful! These conversions can result in a loss of precision, since 8 bits
+cannot hold the same amount of information as 64 bits::
 
    >>> image = np.array([0, 0.5, 0.503, 1], dtype=float)
    >>> image_as_ubyte(image)
    array([  0, 128, 128, 255], dtype=uint8)
-
-If desired, the user can be warned when such loss of information happens, by
-using a keyword argument::
-
-   >>> image_as_ubyte(image, warn_on_precision_loss=True)
-   UserWarning: Possible precision loss when converting from float64 to uint8
-   array([  0, 128, 128, 255], dtype=uint8)
-
-There is also a keyword argument to warn when it is possible to lose sign
-information, ``warn_on_sign_loss=True``.
 
 Additionally, some functions take a ``preserve_range`` argument where a range
 conversion is convenient but not necessary. For example, interpolation in
--- a/doc/source/user_guide/data_types.rst
+++ b/doc/source/user_guide/data_types.rst
@@ -63,24 +63,29 @@ img_as_int     Convert to 16-bit int.
 =============  =================================
 
 These functions convert images to the desired dtype and *properly rescale their
-values*. If conversion reduces the precision of the image, then a warning is
-issued::
+values*::
 
    >>> from skimage import img_as_ubyte
    >>> image = np.array([0, 0.5, 1], dtype=float)
    >>> img_as_ubyte(image)
-   WARNING:dtype_converter:Possible precision loss when converting from
-   float64 to uint8
    array([  0, 128, 255], dtype=uint8)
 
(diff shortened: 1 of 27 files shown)


hmaarrfk:
There don't seem to be any tests that actually test that the warnings are correct. Maybe you can adapt these: https://github.com/scikit-image/scikit-image/pull/3055/files#diff-74fec0b61086721cf4ec492e880e1d6cR153

jni:
@hmaarrfk I just left a few tests open where the warnings are on. Search for `expected_warnings.*precision`, or `warn_on_precision_loss=True`.

skimage/filters/rank/generic.py
```diff
@@ -51,7 +51,7 @@
 import functools
 import numpy as np
 from scipy import ndimage as ndi
-from ... import img_as_ubyte
+from ...util import img_as_ubyte
```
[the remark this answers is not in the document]
jni: This is very much a fair point. I'll let @scikit-image/core comment before I make that change, but I am :+1: on this suggestion.

hmaarrfk:
@jni, they just don't seem systematically tested.

See this issue for example
https://github.com/scikit-image/scikit-image/issues/3080

https://github.com/scikit-image/scikit-image/pull/3055 had fixed it, but there was too much magic going on in one of the conversions, so I stopped pushing for it.

jni:
:scream: wow. Yeah, I don't particularly want to get deep into the logic of `convert` for this PR...

alexandrejaguar:
Hi all,
I think we should believe in the users a little bit here. Wouldn't it be easier for us to only remove the warning, as @stefanv suggested, and pointing the precautions users should have when dealing with data conversion in a tutorial, or something like that?
Personally, I don't see a situation where maintaining the hidden warning would be useful. OK, maybe one or two cases popped on my mind now, but I still think we should let that in the users hands; my 2 ¢.

doc/source/user_guide/data_types.rst
```diff
@@ -176,7 +181,7 @@ a custom function that requires a particular dtype, you should call one of the
 dtype conversion functions (here, ``func1`` and ``func2`` are ``skimage``
 functions)::
 
-   >>> from skimage import img_as_float
+   >>> from skimage.util import img_as_float
```
stefanv: Why these changes?  I think those function remain exposed on the default namespace, without plans to change that any time soon or ever.

jni: @stefanv I object to the second clause in that sentence. =P 

I want this to be deprecated in the same way that `np.matrix` was "deprecated" for a long time before any messages appeared. ie, discouraged. For me it makes much more sense for these functions to be in `util` than "polluting" the top-level domain, which is otherwise only modules and the `data_dir` variable.

jni:
Ok, to quote the great Samuel L Jackson, hold on to your butts!

Since we have two core devs in favour of removing altogether, and no one (as far as I can tell) strongly against it, I've indeed removed all the warnings wholesale. Let's see if anyone complains... ;)

sciunto:
Travis complains :D

doc/source/user_guide/data_types.rst
stefanv: I should have been clearer, I guess: without your plans having been made known to the rest of us? :)

Seriously, though, those functions are strewn throughout *all* our lecture materials, so it would be quite painful to get rid of it.  Now, w.r.t. discouragement, I could go for that, but do you have a specific rationale for why `skimage.util.img_as_float` is better than `skimage.img_as_float`, other than the slight aesthetic benefit of having only submodules in the namespace?  I guess then we'll also need to get rid of `skimage.test()`, which leaves only submodules + `__version__`.

hmaarrfk: I once listened to a talk that was advocate against this kind of nested API. They were saying that we should strive for a "flat" API design. not sure how applicable that is to skimage.

a flat API can be easier to learn, at least easier to remember what to import from.

jni: I was under the impression that this idea had been lightly discussed elsewhere, but I can't find any evidence of that, so, sorry about that.

I don't consider "this is the way we've always done it" to be a big deal, especially in the context of a soft deprecation. There is zero urgency in this removal, it is just something that I think we should do "right" going forward, and especially in our own code base. There's outdated tutorials for everything online, and this should not hold us back, any more than "there's lots of tutorials and books written in Python 2" should hold back the transition to 3. Now, I was auditing the use of these functions in our codebase so I took the opportunity to update these. (And note that we had both conventions floating about in our codebase, which is a jarring inconsistency.)

I actually think `test` and `lookfor` and maybe even `data_dir` should stay in the top level as being "package-wide" things. Is `img_as_float` something that refers to our package as a whole, or a utility function? For me it's a slam dunk for the latter. Put another way: does `img_as_float` belong in the same bucket as `lookfor` and `test`?

jni:
@sciunto wow it's funny how a refactor can uncover unexpected badness! There were a lot of `from skimage._shared.testing import expected_warnings`, but that function is in `_shared._warnings` — it was in `_shared.testing` only as a side effect of it being used in that file. When I removed that use, all those imports broke. :man_facepalming: 

Anyway, :crossed_fingers: !

doc/source/user_guide/data_types.rst
stefanv: Perhaps what bothers me is the nondescript `util` name, which will not help users discover these.  They could potentially live in `io`, or in their own submodule.

You are arguing that these functions are not-package wide, which I agree with.  But another perspective could be that they belong there because they are invoked on every single pipeline, and should therefore be highly visible.

Hopefully they will be needed less as we embrace float images without defined ranges, wherever possible.

jni: Hey I'm the one who argued for a flat namespace to help discoverability and was vetoed. =P

More seriously, as you say, I hope that long term they are not so indispensable.

Finally, perhaps a new module, `skimage.dtypes`? Or somesuch?

jni:
Oh man, Travis is **so** unhappy...

doc/source/user_guide/data_types.rst
hmaarrfk: It seems that maybe you should restrict the scope of the PR. I don't think there is much resistance to any of the proposed changes, but it seems that together, they are hard to justify at once as some are cosmetic, and some are behavioural.

soupault:
@jni Could you rebase, please?

jni:
I doubt it'll help but sure. =D Hold please...

jni:
Erm, where did Travis go?

jni:
@jakirkham thanks for the `--allow-empty` tip! ;) :point_up: 

Hmmm, your instincts were right, @soupault, only two failures!

https://travis-ci.org/scikit-image/scikit-image/jobs/490881001#L2027-L2148

I don't understand either of them though... :joy:

jni:
@scikit-image/core I think my latest two commits should fix the remaining two errors. Please give it another round of review!

jni:
@hmaarrfk eh, I'm in favour of leaving that to #3009. ;)

skimage/filters/rank/generic.py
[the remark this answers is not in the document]
stefanv: Do you mind expanding on the issue here? What is being proposed?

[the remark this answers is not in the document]
jni: @stefanv I think @hmaarrfk is making a distinction between precision loss warnings stemming from IO and user calls to `img_as_ubyte`, which should be silenced, and precision loss warnings because rank filters convert your images to uint8, and now do so completely silently.

Having said that I would like to punt on this issue, as I don't consider it that critical or urgent. (See main conversation thread on the PR.)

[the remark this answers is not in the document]
soupault: >convert your images to uint8, and now do so completely silently

I think that happens in multiple other places also. 👍 to issue the warning.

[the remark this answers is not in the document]
jni: Done.

jni:
@scikit-image/core the napari team just ran into the "excessive warnings" issue. This was approved by @hmaarrfk already. Could someone else review and merge? I've rebased on master, the tests all pass locally, let's see what Travis & Co say, and hopefully get merging!

soupault:
Thanks @jni ! Nice piece of work!

jni:
Thanks @soupault! Nice work with the title and final touches! =)

hmaarrfk:
It is almost worth release 0.16 just for this.

jni:
:joy: I agree. Actually I'd like to get it out soon after SciPy, but I probably won't be able to get to it until I'm back in Melbourne.

Issue #3373, opened 2018-08-23
Code audit: usage of `img_as_float`


jni:
There's been lots of discussion about our automatic range conversions, especially when it comes to `img_as_float`. (See for example #3009 and #3052.) In #3052 @stefanv suggested that we do an audit to check which functions will be affected if we don't ensure that floats are in the range [0, 1]. I went through every non-testing use of `img_as_float` in our code base to check. Below, I've divided the functions into three cases: (1) function will not work correctly if image is out of range; (2) function will return a correspondingly-scaled image (ie it will work correctly, but the output is not scale invariant), and (3) functions where the output will be unchanged (ie we can safely remove range conversion altogether).

## Range essential for correctness

- `exposure.intensity_range` will return [0, 1] for `range_values='dtype'` or `range_values='float'`. However this behaviour in no way depends on `img_as_float`, let alone on `img_as_float` raising an error.
- `feature.draw_haar_like_features` does need the image to be in the correct range because colours are normalised to be in [0, 1]
- Same for `feature.draw_multiblock_lbp`.
- `feature.plot_matches` needs the range to make sure that the random color line drawn over it makes sense.
- `restoration.denoise_wavelet` probably needs a specific range, as it converts images to ycbcr and color conversion would assume a certain range for RGB input.
- `segmentation.felzenszwalb` method divides scale by 255 to be consistent with the reference implementation, I presume by assuming that the image values are in [0, 1] instead of [0, 255]. Therefore, here we would need to be careful to make sure that the scale matches the input image scale.
- `segmentation.mark_boundaries` "paints" colour onto the image, so the input image needs to be in [0, 1]. (Note that it currently doesn't check whether a float image is in that range, it just uses img_as_float; so this needs fixing, as do other cases of drawing covered in this list.)
- `future.rag.show_rag` needs the [0, 1] magnitude to display correctly in mpl

## Corresponding magnitude change in output

- `feature.corner.hessian_matrix_det` and `hessian_matrix_elems` use it. There will be a difference in magnitude if the image is out of range but everything should still "just work".
- Same for `feature.corner_moravec`
- `feature.structure_tensor` as above.
- `feature.CENSURE` -- not sure about this one, but it has some thresholding which probably needs to change if we change the scale of the image. Uses `_prepare_grayscale_input_2D`, a convenience function that calls `img_as_float`.
- `feature.corner_fast` uses a threshold to compare to image sums, so threshold choice is dependent on image range. Uses `_prepare_...`
[...]

PR #3390, opened 2018-09-06
Backport PR #3052 on branch v0.14.x (Allow float->float conversion of any range)

meeseeksmachine:
Backport PR #3052: Allow float->float conversion of any range

PR #3584, opened 2018-12-06
Fix the data scaling and clipping behavior of denoise_wavelet

grlee77:
## Description

This PR fixes two separate issues with data scaling in `denoise_wavelet`.

- The changes in #3052 (see also #3373), mean that floating point inputs will no longer be automatically rescaled to the range [0, 1] or [-1, 1]. I think this is desirable, but it means that it is no longer appropriate to clip the output to this range when the input is a float.

- In cases where `img_as_float` does rescale the image, the same rescaling should also be applied to `sigma`, but this was not previously being done. This was okay for the default `sigma=None` as the scaling occurs prior to the internal call to `estimate_sigma`. 

I added some test cases that check both PSNR and the output range for various dtypes.  All of these test cases fail without the changes in this PR.

The previously existing tests were passing because the inputs had already been scaled via `img_as_float` prior to calling `denoise_wavelet`.

## Checklist
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/restoration/_denoise.py
+++ b/skimage/restoration/_denoise.py
@@ -671,8 +671,14 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
     image, but larger thresholds also decrease the detail present in the image.
 
     If the input is 3D, this function performs wavelet denoising on each color
-    plane separately. The output image is clipped between either [-1, 1] and
-    [0, 1] depending on the input image range.
+    plane separately.
+
+    For floating point inputs, the original input range is maintained and there
+    is no clipping applied to the output. Other input types will be converted
+    to a floating point value in the range [-1, 1] or [0, 1] depending on the
+    input image range. Any rescaling applied to the `image` will also be
+    applied to `sigma` to maintain the same relative amplitude for the supplied
+    noise standard deviation.
 
     When YCbCr conversion is done, every color channel is scaled between 0
     and 1, and `sigma` values are applied to these scaled color channels.
@@ -722,11 +728,22 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
             ('Invalid method: {}. The currently supported methods are '
              '"BayesShrink" and "VisuShrink"').format(method))
 
-    image = img_as_float(image)
-
     if multichannel:
         if isinstance(sigma, numbers.Number) or sigma is None:
             sigma = [sigma] * image.shape[-1]
+
+    clip_output = False
+    if image.dtype.kind != 'f':
+        clip_output = True
+        data_range_pre = image.max() - image.min()
+        image = img_as_float(image)
+        data_range_post = image.max() - image.min()
+        scale_factor = data_range_post / data_range_pre
+        if sigma is not None:
+            if multichannel:
+                sigma = [s * scale_factor for s in sigma]
+            else:
+                sigma *= scale_factor
 
     if multichannel:
         if convert2ycbcr:
@@ -735,13 +752,15 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
                 # renormalizing this color channel to live in [0, 1]
                 min, max = out[..., i].min(), out[..., i].max()
                 channel = out[..., i] - min
-                channel /= max - min
+                sf = max - min
+                channel /= sf
+                _sigma = sigma[i]/sf if sigma[i] is not None else None
                 out[..., i] = denoise_wavelet(channel, wavelet=wavelet,
-                                              method=method, sigma=sigma[i],
+                                              method=method, sigma=_sigma,
                                               mode=mode,
                                               wavelet_levels=wavelet_levels)
 
-                out[..., i] = out[..., i] * (max - min)
+                out[..., i] = out[..., i] * sf
                 out[..., i] += min
             out = color.ycbcr2rgb(out)
         else:
@@ -757,8 +776,11 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
                                  sigma=sigma, mode=mode,
                                  wavelet_levels=wavelet_levels)
 
-    clip_range = (-1, 1) if image.min() < 0 else (0, 1)
-    return np.clip(out, *clip_range)
+    if clip_output:
+        clip_range = (-1, 1) if image.min() < 0 else (0, 1)
+        return np.clip(out, *clip_range)
+    else:
+        return out
 
 
 def estimate_sigma(image, average_sigmas=False, multichannel=False):
--- a/skimage/restoration/_denoise.py
+++ b/skimage/restoration/_denoise.py
@@ -619,6 +619,32 @@ def _wavelet_threshold(image, wavelet, method=None, threshold=None,
     return pywt.waverecn(denoised_coeffs, wavelet)[original_extent]
 
 
+def _scale_sigma_and_image_consistently(image, sigma, multichannel):
+    """If the `image` is to be rescaled, also rescale `sigma` consistently.
+
+    Images that are not floating point will be rescaled via `img_as_float`.
+
+    Sigma gets rescaled by the same amount, under the assumption that any
+    user-specified sigma would be at the same magnitude scale as the image.
+    """
+    if multichannel:
+        if isinstance(sigma, numbers.Number) or sigma is None:
+            sigma = [sigma] * image.shape[-1]
+
+    if image.dtype.kind != 'f':
+        range_pre = image.max() - image.min()
+        image = img_as_float(image)
+        range_post = image.max() - image.min()
+        # apply the same magnitude scaling to sigma
+        scale_factor = range_post / range_pre
+        if multichannel:
+            sigma = [s * scale_factor if s is not None else s for s in sigma]
+        elif sigma is not None:
+            sigma *= scale_factor
+
+    return image, sigma
+
+
 def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
                     wavelet_levels=None, multichannel=False,
                     convert2ycbcr=False, method='BayesShrink'):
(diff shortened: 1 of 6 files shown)


grlee77:
we can probably also backport this to `0.14.x` because #3052 was already backported in #3390.

skimage/restoration/_denoise.py
```diff
@@ -594,26 +600,39 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
...
+
     if multichannel:
         if convert2ycbcr:
             out = color.rgb2ycbcr(image)
             for i in range(3):
                 # renormalizing this color channel to live in [0, 1]
                 min, max = out[..., i].min(), out[..., i].max()
                 channel = out[..., i] - min
-                channel /= max - min
+                sf = max - min
```
jni: I'd rename this to `scale_factor`, and also, I know this isn't this PR's issue, but we should aim not to override Python built-in names, so `min` and `max` should probably be renamed to `channel_min` and `channel_max`, for example.

jni: Also, we should probably check for 0 range and exit early, rather than dividing by 0.

skimage/restoration/_denoise.py
```diff
@@ -594,26 +600,39 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
             ('Invalid method: {}. The currently supported methods are '
              '"BayesShrink" and "VisuShrink"').format(method))
 
-    image = img_as_float(image)
-
     if multichannel:
         if isinstance(sigma, numbers.Number) or sigma is None:
             sigma = [sigma] * image.shape[-1]
 
+    clip_output = False
```
jni: You could just write `clip_output = image.dtype.kind != 'f'`

skimage/restoration/_denoise.py
```diff
@@ -629,8 +648,11 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
                                  sigma=sigma, mode=mode,
                                  wavelet_levels=wavelet_levels)
 
-    clip_range = (-1, 1) if image.min() < 0 else (0, 1)
-    return np.clip(out, *clip_range)
+    if clip_output:
```
jni: I suggest:

```
if clip_output:
    np.clip(out, *clip_range, out=out)
return out
```

skimage/restoration/tests/test_denoise.py
```diff
@@ -386,6 +386,49 @@ def test_wavelet_denoising():
         assert_(np.sum(res1**2) <= np.sum(res2**2))
 
 
+def test_wavelet_denoising_scaling():
+    """Test case for floating point input outside the [-1, 1] range."""
+    rstate = np.random.RandomState(1234)
+    for dtype in [np.float16, np.float32, np.float64, np.int16, np.uint8]:
```
jni: Possibly this should be done with `pytest.parametrize`, which will then tell you which types failed, if any.

skimage/restoration/_denoise.py
```diff
@@ -594,26 +600,39 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
...
+        clip_output = True
+        data_range_pre = image.max() - image.min()
+        image = img_as_float(image)
+        data_range_post = image.max() - image.min()
+        scale_factor = data_range_post / data_range_pre
+        if sigma is not None:
+            if multichannel:
+                sigma = [s * scale_factor for s in sigma]
+            else:
+                sigma *= scale_factor
```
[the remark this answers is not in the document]
grlee77: I agree that it may warrant that. It is clearly a change in behavior, but I think will be more intuitive going forward.

`estimate_sigma` does not do any rescaling, so an input in range [0, 1e6] might produce a huge sigma value of 1e4 or so which would clearly give a crazy oversmoothing when applied to an image that has been rescaled to range [0, 1]!

For the current release, `sigma` only really works if the user computed it for an `image` scaled to [0, 1] or [1, -1] (or if it is left to the default of `None`).

skimage/restoration/_denoise.py
grlee77: Good idea on the range check. I think we can just skip "empty" channels via.
```python
                if scale_factor == 0:
                    # nothing to denoise if channel has zero amplitude
                    continue
```

grlee77: Actually, I need to revisit the whole YCbCr case and add tests for it. I think to do that color conversion we will need to rescale the floating-point RGB values to [0, 1] range first.
I am actually not sure why this per-channel rescaling after YCbCr conversion was used in the first place, but the author explicitly mentions in the docstring:

`
    When YCbCr conversion is done, every color channel is scaled between 0
    and 1, and `sigma` values are applied to these scaled color channels.
`

grlee77: Looking into `rgb2ycbcr`, it calls `_prepare_color_array` internally which uses `img_as_float` and thus does not automatically rescale floating point inputs. The docstring states that: `Y is between 16 and 235.`, but this is not true for floating point inputs unless they are in [0, 1].

example:
```Python
# uint8 case
np.max(skimage.color.rgb2ycbcr(data.chelsea()))
Out[75]: 182.74402352941178

# floats in range [0, 255.]
np.max(skimage.color.rgb2ycbcr(data.chelsea().astype(float)))
Out[76]: 42535.726
```

For that reason, it is good that we are also rescaling the YCbCr channels again in the inner channel denoising loop. The question is how to best handle sigma in this case:

Perhaps we should apply a global sigma scaling across all channels as opposed to doing per-channel sigma scaling as is being done currently in this PR.  That would be more consistent with the current YCbCr description in the denoise_wavelet docstring.

grlee77:
I messed up the scaling in the YCbCr case and still have to revisit that (see discussion above).

grlee77:
This is still broken in the YCbCr case. I need to add a test case with a color image to the paramtrization

skimage/restoration/_denoise.py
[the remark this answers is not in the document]
jni: Oh yes, I am absolutely not arguing that we shouldn't have the change, just whether it should be a silent/hard change or a deprecation. I think we might have many people who had their sigma set (empirically) 255 times lower than intended, and now they will find that their code does no denoising at all, and be at a loss as to why.

skimage/restoration/_denoise.py
grlee77: I have updated the YCbCr case with a proper solution (see 7e25ef3) for scaling `sigma` using the color conversion matrix (assuming that the noise across the R, G, B channels is independently distributed).

grlee77: I made the changes suggested above by @jni here

skimage/restoration/_denoise.py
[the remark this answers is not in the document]
grlee77: I think the `clip_output` flag to avoid clipping floating point outputs to [0, 1] or [-1, 1] should be applied now as a bug fix.

I think the automatic rescaling of sigma when the image must be rescaled is convenient and the right thing to do, but that could potentially go through a deprecation cycle for the reason you mention.

grlee77:
There was one CI failure in `TestRank.test_rank_filter[windowed_histogram]` unrelated to this PR.

I think the primary outstanding issue is how to handle the transition to this new, more user-friendly behavior. After these changes, the user can compute `sigma` for whatever the current values are in `image`.  If the image is rescaled internally, `sigma` gets automatically rescaled in the same way. This is much less likely to result in misuse of the function or unexpected behavior due to internal rescaling that would not be obvious to the user.

grlee77:
> Oh yes, I am absolutely not arguing that we shouldn't have the change, just whether it should be a silent/hard change or a deprecation. I think we might have many people who had their sigma set (empirically) 255 times lower than intended, and now they will find that their code does no denoising at all, and be at a loss as to why.

I would really like to get this fix in for `0.16`.  I think Juan's comment quoted above is the only thing left to address here. As currently implemented, this is a "hard change", but I can update to go the deprecation route if that is preferred.

jni:
I'm still in favour of a deprecation pathway, but could be swayed by other voices! What's your current thinking @grlee77?

skimage/restoration/_denoise.py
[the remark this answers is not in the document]
stefanv: Frustrating—so close to a bug fix, but not quite!  We may argue for an accelerated deprecation in this case, perhaps; i.e., that the behavior changes immediately but that old users at least get a warning?  Either that, or the more conservative option of deprecating.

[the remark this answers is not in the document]
jni: @stefanv I'd be on board with "accelerated deprecation" in cases of bug fixes that can affect correct running of code. @grlee77 what do you think? My main concern with this kind of warning is that it should be easy to disable.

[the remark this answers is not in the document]
grlee77: I would lean toward "accelerated deprecation".  I think it is likely not uncommon that users with floating point data would be getting bad results currently.

> My main concern with this kind of warning is that it should be easy to disable.

Is there a recommended way to do this? I am only familiar with the user needing to use `with warnings.catch_warnings` context to disable the unwanted warnings?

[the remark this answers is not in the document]
jni: I'm thinking some kind of long-lived keyword argument, `rescale_sigma=None`, which means `True`, and if set to None (default), a warning is issued, but if set to `True` there is no warning. If set to `False` you get the old behaviour but a warning that this option will go away in the future. In both the docstring and the warning, explain the issue and maybe link to this discussion.

Thoughts?

The point is that `catch_warnings` is super-verbose and people are unlikely to want to pollute their codebase with it.

The main problem with the long-lived kwarg is that it results in yet another deprecation in the future. I don't see a better way around it though...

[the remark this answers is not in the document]
jni: @JDWarner tends to have opinions about these things... =)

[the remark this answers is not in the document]
grlee77: I think adding a `rescale_sigma=None`argument sounds good and gives an easier way to bypass any warning if desired.

skimage/restoration/_denoise.py
```diff
@@ -610,11 +669,14 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
...
+    plane separately.
 
-    When YCbCr conversion is done, every color channel is scaled between 0
-    and 1, and `sigma` values are applied to these scaled color channels.
+    For floating point inputs, the original input range is maintained and there
+    is no clipping applied to the output. Other input types will be converted
+    to a floating point value in the range [-1, 1] or [0, 1] depending on the
+    input image range. Any rescaling applied to the `image` will also be
+    applied to `sigma` to maintain the same relative amplitude for the supplied
+    noise standard deviation.
```
jni: @grlee77 Sorry, this just occurred to me. Could you add a heading to this section, `.. versionchanged:: 0.16`? This makes it easier for people scouring the docs for reasons why their output has changed. Do `git grep versionchanged` in our repo for example usages. It's helped me a few times with NumPy changes.

skimage/restoration/_denoise.py
```diff
@@ -594,6 +647,12 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
     method : {'BayesShrink', 'VisuShrink'}, optional
         Thresholding method to be used. The currently supported methods are
         "BayesShrink" [1]_ and "VisuShrink" [2]_. Defaults to "BayesShrink".
+    rescale_sigma : bool or None, optional
+        If False, no rescaling of the user-provided ``sigma`` will be
+        performed. The default of ``None`` rescales sigma appropriately if the
+        image is rescaled internally. A ``DeprecationWarning`` is raised to
+        warn the user about this new behaviour. This warning can be avoided
+        by setting ``rescale_sigma=True`.
```
jni: As below: could you add `.. versionadded:: 0.16`?

skimage/restoration/_denoise.py
grlee77: I agree it can be useful. I have added it now.

jni:
@grlee77 as usual with sphinx errors, it's hard to tell what's wrong, but comparing your formatting with the [existing ones](https://github.com/scikit-image/scikit-image/blob/7ec25d986ef42e1dd3f2c5e0ff962de12bbade64/skimage/filters/_median.py#L60) ([which work](https://scikit-image.org/docs/dev/api/skimage.filters.html#skimage.filters.median)), the only thing I can spot is three spaces to indent the subsequent paragraph, rather than four. Because why not.

grlee77:
Well, 3 spaces didn't seem to work either. I will try testing a bit more locally to not tie up the CI. I did check that it appeared to render okay prior to uploading the prior commit, but perhaps there was a warning I overlooked.

grlee77:
@jni: Adding a `backtick` fixed the documentation build issue

There is one unrelated failure on Travis in `test_rank_filter`.
Two others timed out downloading something via `homebrew` so I tried restarting those.

jni:
Woot! The rank filters failure is known, see #4052. Thanks @grlee77!

grlee77:
Updated the tests to account for the rename of `compare_psnr` -> `peak_signal_noise_ratio`. Should be good to go now.

grlee77:
Any objections to merging this now so it will be in 0.16? (did not want to self-merge my own PR)

jni:
@scikit-image/core anyone around to pull the trigger? =) Clearly no objections from me, @grlee77! =)

NelleV:
Tagging this as critical for the release.

alexdesiqueira:
Thanks for the fix, @grlee77!

NelleV:
Thanks everyone for contributing and merging! Will backport this to 0.16 (aka, I'll probably just rebase…)

PR #3592, opened 2018-12-17
added positivity constraint to Chambolle-Pock TV

alemirone:
## Description
Positivity constraint in Chambolle-Pock TV
(min max -> max min and  the positivity indicator function disappears during the implicit resolution of min).

### DIFF
--- a/skimage/restoration/_denoise.py
+++ b/skimage/restoration/_denoise.py
@@ -160,7 +160,7 @@ def denoise_tv_bregman(image, weight, max_iter=100, eps=1e-3, isotropic=True):
     return _denoise_tv_bregman(image, weight, max_iter, eps, isotropic)
 
 
-def _denoise_tv_chambolle_nd(image, weight=0.1, eps=2.e-4, n_iter_max=200):
+def _denoise_tv_chambolle_nd(image, weight=0.1, eps=2.e-4, n_iter_max=200, positivity=False):
     """Perform total-variation denoising on n-dimensional images.
 
     Parameters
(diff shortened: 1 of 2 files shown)


emmanuelle:
Hi @alemirone why would you want to force the positivity of the output?

alemirone:
Hello Emmanuelle,
  now that I think to it, not only positivity could be enforced but more in general the constraint to a convex domain :  I could add a minimum and a maximum.
So why?
Because at the same cost that you need to  solve
argmin_x  (  1/2 L2(x-z)^2 + w TV(x) )
you can solve
argmin_x  (  1/2 ||x-z||^2 + w TV(x) ) + Indicator(x, domain)
and keeping the same  structure of the current implementation.
In the current implementation, calling p the dual variables of the variations ,
the following equation is implicitly solved :
min_x 1/2 ||x -(z+div(p) ||^2   + Indicator(x, domain) ......
The extra Indicator term implies an extra projection that is what I have added in the routine.
( the same  would not be trivial  for a more general problem containing
|| M.x - (z+div(p) ||+...  with M a matrix, which should be solved by keeping the primal-dual method explicit  and thus with more dual variables ( see Convex optimization problem prototyping for image reconstruction in computed tomography with the Chambolle-Pock algorithm Emil Y. Sidky, Jakob H. Jørgensen, and Xiaochuan Pan ))
Do you want me to generalise to a convex domain ( minimum and maximum values given as either None, or a float or an array of floats) ( and always with M= identity)?
To tell you more about the usage it is used to reconstruct volumes on ID20, at given photon-out energy, using several crystal analyzers in 1:1 imaging configuration. The deconvolution may introduce artifact, especially using super resolution, the pixel being smaller in general than the point spread function of the optics. And sometimes the deconvoluted signal may have some negative artefacts ( may be some absorption effect?).
So that's why I adapted the CP_tv to clip optionally on the positive domain ,
Have a nice day
   Alessandro

matthew-brett:

<details>

## Rationale

</details>

PR #2536, opened 2017-02-23
Tuple for 2D window sizes in threshold_niblack & threshold_sauvola

sarwatfatimam:
## Description
The `filters.threshold_{niblack,sauvola}` accepts a tuple for 2D window size. We are working on 3D window size. 

## References
It implements portion of issue #2442 

Sarwat Fatima & Aqsa Ahmed

PR #3596, opened 2018-12-18
EHN: allow tuple to define kernel in threshold_niblack and threshold_sauvola

### LINKED ISSUE #2442: Allow rectangular window sizes in threshold_niblack and threshold_sauvola
## Description
The functions `filters.threshold_{niblack,sauvola}` accept a `window_size` parameter, which defines a square/cubic window along which to threshold. However, any cuboid shape would work fine, and would be desirable e.g. in anisotropic 3D data, or even in 2D data where the lighting changes more rapidly along one dimension than the other.

mrastgoo:
## Description

Add support for `tuple` to specify the size of a kernel in functions `threshold_niblack` and `threshold_sauvola`

## Checklist
  existing benchmark
- [x] Unit tests

## References

closes #2442 
superseded #2536

## For reviewers

  later.

### DIFF
--- a/doc/source/api_changes.rst
+++ b/doc/source/api_changes.rst
@@ -2,7 +2,10 @@ Version 0.15
 ------------
 - ``skimage.feature.canny`` now uses a more accurate Gaussian filter
   internally; output values will be different from 0.14.
-
+- ``skimage.filters.threshold_niblack`` and
+  ``skimage.filters.threshold_sauvola``
+  now accepts a tuple as ``window_size`` besides integers.
+  
 Version 0.14
 ------------
 - ``skimage.filters.gaussian_filter`` has been removed. Use
--- a/doc/source/api_changes.rst
+++ b/doc/source/api_changes.rst
@@ -4,7 +4,7 @@ Version 0.15
   internally; output values will be different from 0.14.
 - ``skimage.filters.threshold_niblack`` and
   ``skimage.filters.threshold_sauvola``
-  now accepts a tuple as ``window_size`` besides integers.
+  now accept a tuple as ``window_size`` besides integers.
   
 Version 0.14
 ------------
--- a/skimage/filters/tests/test_thresholding.py
+++ b/skimage/filters/tests/test_thresholding.py
@@ -187,6 +187,30 @@ class TestSimpleImage():
         out = self.image > thres
         assert_equal(ref, out)
 
+    def test_threshold_niblack_tuple_window_size(self):
+        ref = np.array(
+            [[False, False, False, True, True],
+             [False, True, True, True, True],
+             [False, True, True, True, False],
+             [False, True, True, True, True],
+             [True, True, False, False, False]]
+        )
+        thres = threshold_niblack(self.image, window_size=(3, 3), k=0.5)
+        out = self.image > thres
+        assert_equal(ref, out)
+
+    def test_threshold_sauvola_tuple_window_size(self):
+        ref = np.array(
+            [[False, False, False, True, True],
+             [False, False, True, True, True],
+             [False, False, True, True, False],
+             [False, True, True, True, False],
+             [True, True, False, False, False]]
+        )
+        thres = threshold_sauvola(self.image, window_size=(3, 3), k=0.2, r=128)
+        out = self.image > thres
+        assert_equal(ref, out)
+
 
 def test_otsu_camera_image():
     camera = skimage.img_as_ubyte(data.camera())
--- a/skimage/filters/tests/test_thresholding.py
+++ b/skimage/filters/tests/test_thresholding.py
@@ -1,7 +1,7 @@
 import pytest
 import numpy as np
 from scipy import ndimage as ndi
-
+from collections import Iterable
 import skimage
 from skimage import data
 from skimage._shared._warnings import expected_warnings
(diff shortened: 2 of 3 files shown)


skimage/filters/thresholding.py
```diff
@@ -818,27 +819,40 @@ def _mean_std(image, w):
            Retrieval XV, (San Jose, USA), Jan. 2008.
            :DOI:`10.1117/12.767755`
     """
-    if w == 1 or w % 2 == 0:
-        raise ValueError(
-            "Window size w = %s must be odd and greater than 1." % w)
+    def _sanity_check_window_size(axis_size):
+        if axis_size == 1 or axis_size % 2 == 0:
+            raise ValueError(
+                "Window size w = {} must be odd and greater than 1."
```
jni: Why does it need to be greater than 1?

skimage/filters/thresholding.py
```diff
@@ -818,27 +819,40 @@ def _mean_std(image, w):
...
+                .format(axis_size)
+            )
+
+    if isinstance(w, tuple):
```
jni: here it might be better to check for `isinstance(w, collections.Iterable)`, so that people can pass in a list. In many APIs they are used interchangeably.

mrastgoo:
I took over what seems a stalled PR #2442. Hope that nobody else took it.

skimage/filters/tests/test_thresholding.py
```diff
@@ -187,6 +188,30 @@ def test_threshold_sauvola(self):
...
+
+    def test_threshold_sauvola_tuple_window_size(self):
+        ref = np.array(
+            [[False, False, False, True, True],
+             [False, False, True, True, True],
+             [False, False, True, True, False],
+             [False, True, True, True, False],
+             [True, True, False, False, False]]
+        )
+        thres = threshold_sauvola(self.image, window_size=(3, 3), k=0.2, r=128)
```
jni: Can you test tuple inputs where the values are not identical?

doc/source/api_changes.rst
```diff
@@ -2,7 +2,10 @@ Version 0.15
 ------------
 - ``skimage.feature.canny`` now uses a more accurate Gaussian filter
   internally; output values will be different from 0.14.
-
+- ``skimage.filters.threshold_niblack`` and
+  ``skimage.filters.threshold_sauvola``
+  now accepts a tuple as ``window_size`` besides integers.
```
emmanuelle: now accept

skimage/filters/thresholding.py
mrastgoo: I am confuse, what would be effect of kernel of size 1 ? Doesn't it indicate that mean will be the pixel value it self and std 0 ?

emmanuelle: the case ``w=1`` is trivial (unchanged image for mean, zero standard deviation)

skimage/filters/thresholding.py
```diff
@@ -801,8 +801,10 @@ def _mean_std(image, w):
     ----------
     image : ndarray
         Input image.
-    w : int
-        Odd window size (e.g. 3, 5, 7, ..., 21, ...).
+    w : int, tuple
```
emmanuelle: at the beginning of the docstring you should also modify "rectangular window with size w times w".

skimage/filters/thresholding.py
```diff
@@ -912,7 +929,9 @@ def threshold_sauvola(image, window_size=15, k=0.2, r=None):
     image: (N, M) ndarray
```
emmanuelle: the image can be n-d I think, at least 3D, therefore the `(N, M)` shape should be removed.

skimage/filters/thresholding.py
```diff
@@ -863,7 +878,9 @@ def threshold_niblack(image, window_size=15, k=0.2):
...
```
emmanuelle: image can be n-d, remove `(N, M)`

emmanuelle:
I also left a couple of comments, I hope they should be easy to address. Thank you @mrastgoo :-)

skimage/filters/thresholding.py
```diff
@@ -931,7 +951,7 @@ def threshold_sauvola(image, window_size=15, k=0.2, r=None):
 
     References
     ----------
-    .. [1] J. Sauvola and M. Pietikainen, "Adaptive document image
+    .. [1] J. Sauvola and M". Pietikainen, Adaptive document image
```
emmanuelle: this is weird, I think this should be reverted

skimage/filters/thresholding.py
jni: Guys, the whole point of this PR is to allow rectangular window sizes. A window size of (1, 1, 1) indeed makes no sense, but (1, 50, 50) might make sense for an image stack.

jni: Or it could just be ignored? If someone gives w=1, a whole bunch of computation will be wasted, but nothing more dramatic will happen.

skimage/filters/thresholding.py
```diff
@@ -860,10 +876,12 @@ def threshold_niblack(image, window_size=15, k=0.2):
 
     Parameters
     ----------
-    image: (N, M) ndarray
-        Grayscale input image.
-    window_size : int, optional
-        Odd size of pixel neighborhood window (e.g. 3, 5, 7...).
+    image: ndarray
+        Input image.
+    window_size : int, tuple or list of integer
```
sciunto:
```suggestion
     window_size : int, tuple or list of integer, optional
```

skimage/filters/thresholding.py
```diff
@@ -909,10 +927,12 @@ def threshold_sauvola(image, window_size=15, k=0.2, r=None):
...
```
sciunto:
```suggestion
     window_size : int, tuple or list of integer, optional
```

skimage/filters/thresholding.py
sciunto: I agree with @jni to ignore it.

jni:
Thanks @mrastgoo!!! This PR makes me very happy. =)

PR #3598, opened 2018-12-18
Update chat location

stefanv:


jni:
I merged directly since it's bad to point to gitter at this point, and it's just a README fix.

PR #3600, opened 2018-12-18
DOC: "feature.peak_local_max" : explanation of multiple same-intensity peaks returned by the function; added details on `exclude_border` parameter 

### LINKED ISSUE #2407: Better documentation for `peak_local_max`s `exclude_border` boolean variant
## Description

The documentation of [`peak_local_max`](http://scikit-image.org/docs/dev/api/skimage.feature.html#skimage.feature.peak_local_max) says nothing about the boolean variant of the `exclude_border` parameter, although it defaults to `True`. So it's unclear from the documentation what that is supposed to mean.

By looking at the corresponding [code](https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/peak.py#L115), one can see that `min_distance` is used for `exclude_border` when it is passed `True`.

**First improvement:** Update the corresponding documentation to make this invariant clear.

Additionally, I think the default value of `True` might lead to subtle problems, that people might not be aware of. For instance, I wasn't aware of this fact until now and just found it by visual inspection. My mental model of NMS doesn't seem to fit the default implementation of `peak_local_max`, which I find problematic.

**Second improvement:** Don't use `True` as default for `exclude_border`, instead use `0`.

shcrela:
## Description
- Explained the fact of the function returning all off the same intensity peaks from the region defined with min_distance.
- Added the details on the `exclude_border` parameter behavior and the bool vs. int parameter values.

## Checklist
  existing benchmark
- [ ] Unit tests

## References
closes #2407

## For reviewers

  later.

### DIFF
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -55,9 +55,9 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
     exclude_border : int or bool, optional
         If nonzero int, `exclude_border` excludes peaks from
         within `exclude_border`-pixels of the border of the image.
-	If True, takes the `min_distance` parametar as value.
+	If True, takes the `min_distance` parameter as value.
 	If zero or False, peaks are identified regardless of their
-	distance from the border
+	distance from the border.
     indices : bool, optional
         If True, the output will be an array representing peak
         coordinates.  If False, the output will be a boolean array shaped as


skimage/feature/peak.py
```diff
@@ -51,9 +52,12 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
         the minimum intensity of the image.
     threshold_rel : float, optional
         Minimum intensity of peaks, calculated as `max(image) * threshold_rel`.
-    exclude_border : int, optional
-        If nonzero, `exclude_border` excludes peaks from
+    exclude_border : int or bool, optional
+        If nonzero int, `exclude_border` excludes peaks from
         within `exclude_border`-pixels of the border of the image.
+	If True, takes the `min_distance` parametar as value.
+	If zero or False, the peak is identified regardless of its'
```
emmanuelle: I would use plural instead "peaks are identified regardless of their distance from the border".

skimage/feature/peak.py
```diff
@@ -51,9 +52,12 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
         the minimum intensity of the image.
     threshold_rel : float, optional
         Minimum intensity of peaks, calculated as `max(image) * threshold_rel`.
-    exclude_border : int, optional
-        If nonzero, `exclude_border` excludes peaks from
+    exclude_border : int or bool, optional
+        If nonzero int, `exclude_border` excludes peaks from
         within `exclude_border`-pixels of the border of the image.
+	If True, takes the `min_distance` parametar as value.
```
sciunto:
```suggestion
 	If True, takes the `min_distance` parameter as value.
```

skimage/feature/peak.py
```diff
@@ -51,9 +52,12 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
+	If zero or False, peaks are identified regardless of their
+	distance from the border
```
sciunto:
```suggestion
 	distance from the border.
```

emmanuelle:
Merging, thanks!

skimage/feature/peak.py
```diff
@@ -51,9 +52,12 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
     threshold_rel : float, optional
         Minimum intensity of peaks, calculated as `max(image) * threshold_rel`.
-    exclude_border : int, optional
-        If nonzero, `exclude_border` excludes peaks from
+    exclude_border : int or bool, optional
+        If nonzero int, `exclude_border` excludes peaks from
         within `exclude_border`-pixels of the border of the image.
+	If True, takes the `min_distance` parameter as value.
+	If zero or False, peaks are identified regardless of their
+	distance from the border.
```
soupault: These tabs should be replaced by whitespaces.

PR #3607, opened 2018-12-18
Add rgb to hsv example in the gallery

### LINKED ISSUE #1231: Examples wanted for the gallery
While starting working on the user guide, I noticed that several functions or functionalities are not illustrated in any example of the gallery. Here are a few examples of functionalities that could make nice examples for the gallery (and hence sections of the user guide). Feel free to ask more item to the wishlist in the conversation to follow.
- [x] RGB to HSV conversion, for example for image enhancement (as increasing image saturation), or for image segmentation (in the `data.coffee()` image, a mask of the cup + saucer can be obtained from thresholding the hue channel). WIP: #3607 
-  [x] RGB to gray conversion. Comment on the different weights of the 3 channels, and on the float dtype of the output. 
- [X] gamma and log contrast adjustement. Plot the different kinds of transformations and show the result when applied to a grayscale image such as `moon` DONE: #1670
- [X] Richardson-Lucy deconvolution. DONE: #2376 
- [ ] LPIFilter2D / inverse filters (I have no idea how they are used. @stefanv ?)
- [ ] ``measure.block_reduce``
- [ ] ``measure.profile_line``

alexis-cvetkov:
## Description
Added rgb to hsv example in the gallery, with an illustration of hsv conversion for color-based segmentation.

gallery1

## Checklist
  existing benchmark
- [ ] Unit tests

## References
closes #1231 

## For reviewers

  later.


doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,44 @@
+"""
+================
```
soupault:
```suggestion
==========
```

doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,44 @@
...
+RGB to HSV
+================
```
soupault:
```suggestion
==========
```

doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,44 @@
...
+
+This example converts an image with RGB channels into an image with HSV (Hue,
+Saturation,Value) channels.
```
soupault:
```suggestion
Saturation, Value) channels.
```

doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,44 @@
...
+
+This coordinate system is useful for color-based thresholding since the color
```
soupault: `coordinate system` -> `representation` or smth similar.
I'd suggest to elaborate in this paragraph a bit more: say, for example, that in many cases objects in image can have distinct intensities or hues, therefore, HSV representation can be effectively used to perform simple segmentation tasks merely by thresholding.

doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,44 @@
...
+of a pixel is only described by its Hue value (and not the R,G,B values). [1]_
+
+References
```
soupault: Remove the section header. See how the references are formatted in other examples (e.g. https://github.com/scikit-image/scikit-image/blob/master/doc/examples/xx_applications/plot_thresholding.py#L24-L29).

doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,44 @@
...
+----------
+.. [1] https://en.wikipedia.org/wiki/HSL_and_HSV
+
+"""
+import matplotlib.pyplot as plt
+
+from skimage import data
+from skimage.color import rgb2hsv
+
+original = data.coffee()
```
soupault: As your example is on conversion _from_ _RGB_, I suggest to rename `original` to `rgb_img`, and update the corresponding figure title.

emmanuelle:
Very nice example @alexis-cvetkov thank you! In addition to the comments of @soupault , I wonder whether it makes sense to represent the HSV image with imshow (which assumes the channels are RBG). Would it make sense to represent only the hue channel, using a colormap? I'm not sure about this, comments are welcome.

jni:
@emmanuelle a very good point! I might suggest the hsv colormap. ;) It might be instructive to also show the saturation and value channels separately, which is doable with a 2x3 panel figure.

emmanuelle:
@jni +1 for using the `hsv` colormap for the hue channel, it kinda makes sense :-). However, I'm not in favor of displaying the other channels, for the sake of keeping the subplots large enough. (we have a lot of examples with very small subplots)

sciunto:
An alternative: You can put the most appealing images in the first plot (used for the thumbnail) and then, explain with a second figure (where the channels can be shown).

alexis-cvetkov:
Thank you all for your feedback ! I tried to improve the example following your suggestions:

gallery1
gallery2
gallery3

jni:
@alexis-cvetkov cool, thank you! Just one more suggestion: I was wondering why the hsv image looked so weird (it should correspond to the dominant colour in the image). It's because it should show the full range of hsv, 0-1. I just noticed that you set the vmax manually. Could you remove it or change it to 1? (The exact max is 0.99934641, but I don't think that'll make much difference, and it's not worth polluting the example with that weird value.) The result is the image below. Yes, it's a lot of red, but it makes it obvious what the hsv transformation is doing. Also, the whole point of using the hsv colourmap is that hue is cyclical. You break that assumption by setting the vmax to 0.12, and so at that point using hsv is meaningless or even harmful (because things at 0.12 and above end up close to things at 0, which is not correct).

Here's what it looks like with vmax=1:

hsv-coffee

I think it clearly shows the two regions of the image (red and orange-red) in a way that corresponds to the actual image colours.

alexis-cvetkov:
@jni I'll fix that ! I initially wanted to emphasize the contrast between the table and the cup (as there are very few pixels with large Hue values) but as you said it's very confusing.

doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,83 @@
...
+
+fig.tight_layout()
+
+##############################################################################
+# We finally perform an additional thresholding on the Value channel to partly
+# remove the shadow of the cup:
+
+fig, ax0 = plt.subplots(figsize=(6, 4))
+
+value_threshold = 0.09
```
emmanuelle: Here I found the result to be visually more appealing with 0.10.

doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,83 @@
...
+binary_img = (hue_img > hue_threshold) | (value_img < value_threshold)
+
+ax0.imshow(binary_img)
+ax0.set_title("Hue and value thresholded image")
+ax0.axis('off')
+
+fig.tight_layout()
```
emmanuelle: fig.tight_layout()
plt.show()

emmanuelle:
Great example @alexis-cvetkov ! I had two minor comments. @soupault are you happy with the changes?

doc/examples/color_exposure/plot_rgb_to_hsv.py
```diff
@@ -0,0 +1,83 @@
...
+##############################################################################
+# We first load the RGB image and extract the Hue and Value channels:
+
+import matplotlib
+import matplotlib.pyplot as plt
+
+from skimage import data
+from skimage.color import rgb2hsv
+
+matplotlib.rcParams['font.size'] = 12
```
soupault: I think this call can be omitted if the `figsize`s are properly configured. I'm fine with the current version, anyway. Thanks!

alexis-cvetkov:
I have added your suggestions, I think it is good to go now !

jni:
Great, thank you @alexis-cvetkov!

PR #3610, opened 2018-12-19
Additionnal mask option to clear_border

### LINKED ISSUE #3594: More general clear_border?
A colleague suggested that it might be interesting to have in ``segmentation`` a more general version of ``clear_border``, which would take a mask as optional argument and remove objects touching the edge of the mask (with respect to the border of the whole image as it is now). It should be straightforward to adapt the existing function. Would there be enough interest to include this feature or not?

PeterJackNaylor:
## Description
Enhancement of the clear_border function by adding an optional mask argument and to remove objects intersecting with the mask.

## Checklist
- [ ] Unit tests

## References
Enhancement and closes #3594 

## For reviewers

  later.


skimage/segmentation/_clear_border.py
```diff
@@ -16,7 +16,8 @@ def clear_border(labels, buffer_size=0, bgval=0, in_place=False):
         Cleared objects are set to this value.
     in_place : bool, optional
         Whether or not to manipulate the labels array in-place.
-
+    mask : ndarray of bool, same shape as `image`, optional.
+        Image data mask.
```
emmanuelle: Could you explain a bit more what ``mask`` is doing? Like "objects containing True pixels of ``mask`` will be removed".

skimage/segmentation/_clear_border.py
```diff
@@ -16,7 +16,9 @@ def clear_border(labels, buffer_size=0, bgval=0, in_place=False):
         Cleared objects are set to this value.
     in_place : bool, optional
         Whether or not to manipulate the labels array in-place.
-
+    mask : ndarray of bool, same shape as `image`, optional.
+        Image data mask. Objects in labels image overlapping with
+        True pixels of mask will be removed.
```
jni: Two points:

- I think we should reverse this logic. In the rest of the library, `mask==True` corresponds to the bits of the image that are kept/processed. So for consistency I think the same should be true here.
- Add a note that the borders are ignored if `mask` is given.

skimage/segmentation/_clear_border.py
```diff
@@ -39,26 +41,47 @@ def clear_border(labels, buffer_size=0, bgval=0, in_place=False):
...
+    ...                  [1, 0, 0, 0, 0, 0, 0, 0, 0],
+    ...                  [1, 0, 0, 0, 0, 0, 0, 0, 0],
+    ...                  [1, 1, 1, 1, 1, 1, 1, 1, 1]]).astype(np.bool)
+    >>> clear_border(labels, mask=mask)
+    array([[0, 0, 0, 0, 0, 0, 0, 1, 0],
+           [0, 0, 0, 0, 1, 0, 0, 0, 0],
+           [0, 0, 0, 1, 0, 1, 0, 0, 0],
+           [0, 0, 1, 1, 1, 1, 1, 0, 0],
+           [0, 1, 1, 1, 1, 1, 1, 1, 0],
+           [0, 0, 0, 0, 0, 0, 0, 0, 0]])
```
jni: I think the docstring example has a lot of room for improvement, actually. (Sorry, this isn't your fault, but it's a good opportunity to tweak it.) Specifically, I think the border objects should be extended by a pixel to show that they just need to *touch* the border/mask to be removed, rather than it just being a single-pixel-wide removal.

skimage/segmentation/_clear_border.py
PeterJackNaylor: ok, I think that it is logical to keep it consitent. I'll commit the corresponding changes.

skimage/segmentation/_clear_border.py
PeterJackNaylor: Ok, I think I understood what you meant. I did a small tweak where I increased the size of the border objects. Should the example be improved still?

skimage/segmentation/_clear_border.py
alexandrejaguar: Hi @PeterJackNaylor, thank you for working on that!
I would remove the info "same shape as `image`" here and put it in the description. I'd put also `labels` in its place, since `image` is not one of the parameters, and not everyone will check the code :wink: Something like that:

```suggestion
    mask : (M[, N[, ..., P]]) array of bool, optional.
        Image data mask with same shape as `labels`. Objects in labels
        image overlapping with True pixels of mask will be removed.
```

What do you both (you and @jni) think about that?
Thanks again!

jni: @alexandrejaguar good suggestion but:

Objects in labels image overlapping with False pixels of mask will be removed.

;)

jni:
@PeterJackNaylor yep, thanks for your tweaks, exactly what I had in mind! Thanks very much for this!

Issue #3599, opened 2018-12-18
local tophat from rank filters has an unclear/unsufficient documentation


sciunto:
## Description

We had a quick look with @ThomasWalter and @emmanuelle on the top hat filter from rank and the docstring/behavior are unclear.

For instance, I understand

```
img = color.rgb2gray(data.hubble_deep_field())[:500, :500]

selem =  morphology.disk(1)
res = morphology.white_tophat(img, selem)

fig, ax = plt.subplots(ncols=3, figsize=(20, 8))
ax[0].set_title('Original')
ax[0].imshow(img, cmap='gray')
ax[1].set_title('White tophat')
ax[1].imshow(res, cmap='gray')
ax[2].set_title('Complementary')
ax[2].imshow(img - res, cmap='gray')

plt.show()
```

but not 
```
res = rank.tophat(img, selem)


fig, ax = plt.subplots(ncols=3, figsize=(20, 8))
ax[0].set_title('Original')
ax[0].imshow(img, cmap='gray')
ax[1].set_title('White tophat')
ax[1].imshow(res, cmap='gray')
ax[2].set_title('Complementary')
ax[2].imshow(img - res, cmap='gray')

```

soupault:
@sciunto see https://github.com/scikit-image/scikit-image/issues/3580.

sciunto:
I know, I opened that one :)

ThomasWalter:
I looked at the image, and the function does not compute a top-hat. 

Just for recall: a white top-hat is defined as WTH = f - opening(f), a black top-hat as BTH = closing(f) - f. 

`rank.tophat` calculates an external gradient, i.e. dilation(f) - f: 

```img_orig = color.rgb2gray(data.hubble_deep_field())[:500, :500]
img = (255*img_orig).astype(np.uint8)
grad = morphology.dilation(img, selem) - img
th_rank = rank.tophat(img, selem)
delta = grad - th_rank
print(delta.max())
delta = th_rank - grad
print(delta.max())
```

The function needs to be renamed, I guess.

sciunto:
I vote for removing the function. Other optinions @scikit-image/core

soupault:
:+1: for removal. I think the next step should be to wrap scipy morphology functions - https://github.com/scikit-image/scikit-image/issues/1917.

rfezzani:
Closing since #4423 deprecates filters.rank.tophat ;-)

PR #3614, opened 2018-12-19
removed tophat and bottomhat from filters.rank

ThomasWalter:
## Description
PR in response to discussion #3599 : removed bottomhat and tophat from `filters.rank`. These filters were named incorrectly. They are effectively calculating internal and external morphological gradients. After discussion with @sciunto and @emmanuelle we came to the conclusion that probably, it does not make so much sense to keep these functions. 

## Checklist
  existing benchmark
- [ ] Unit tests

## References
Closes issue #3599

## For reviewers

  later.


emmanuelle:
If this is approved, ``doc/release/release_dev.rst`` will need to be updated as well. I'm not sure we need to use the deprecation process here since I consider these functions to be quite hidden in the library and my gut feeling is that nobody is using them. But of course, not deprecating/warning users is quite brutal, so please comment @scikit-image/core

jni:
Oh yeah, I don't think we can just *remove* remove them. Since they are incorrect I don't think we should use deprecation, but rather, try to use import hooks or at least a redefinition of them so that a `NotImplementedError` or similar can be raised, mentioning that they were removed in 0.14.2/0.15 because they were incorrect, pointing users to how to reimplement the "morphological gradient" (in case they were getting good results with that behaviour), and to use ndimage for tophat/bottomhat. And also maybe point to the relevant discussion in #3599. Yes it will be a long message but a long message is warranted when you pull the rug from under people!

sciunto:
Postponed to 0.16. @ThomasWalter would mind adding an exception and a task in the todo list please?

grlee77:
I rebased this to resolve merge conflicts from a recent 3d rank filters PR. Now that 0.18 was released, I think this should be ready to merge if all is green on CI (docstrings since the 0.17 release have indicated these would be removed in 0.19)

grlee77 (approved):

stefanv:
Thanks for your input, everyone!

stefanv:
And, of course, @ThomasWalter who stuck with this for two years ;)

PR #1422, opened 2015-03-13
Implement median filter for float types

spark1729:
Median filter was only available for integer-type images, as raised in Issue #1419. Impemented a new funtions `median_float` which internally uses `nd_image.filters._rank_filter`.

PR #3616, opened 2018-12-19
[MRG] EHN: median filters will accept floating image

### LINKED ISSUE #1419: median filter is only available for integer-type images
`filters.median` only operates on integer images. For float images, we could return the output of `ndimage.median_filter`.

### LINKED ISSUE #2943: Median filter does not preserve dtype
The median filter documentation states that the output image will be of the same dtype and the input. However that's not true for gray (float64) inputs.

**Step to reproduce**

```py
from skimage.data import chelsea
from skimage.filters.rank import median
from skimage.color import rgb2gray

X_gray = rgb2gray(chelsea())
print("Input dtype: ", X_gray.dtype)
out = median(X_gray)
print("Output dtype:", out.dtype)
```
which returns
```
Input dtype:  float64
/home/datageek/src/scikit-image/skimage/util/dtype.py:123: UserWarning: Possible precision loss when converting from float64 to uint8
  .format(dtypeobj_in, dtypeobj_out))
Output dtype: uint8
```

Tested on the current master branch (0.14.dev) and 0.13.1 with Python 3.6 on Linux.  
cc @rogersamso

glemaitre:
## Description
Support floating image in `skimage.filters.median`. This PR will controversial, but let's start with a potential solution.

Things to considered:

* Floating images cannot be handled by `skimage.filters.rank.median`
* `scipy.ndimage.median_filter` would be a good candidate for this use case

However,

* `scipy.ndimage.median_filter` does not accept parameter such as `mask` or `shift_*`
* there is no way to go back to the behavior of `skimage.filters.rank.median` using `scipy.ndimage.median_filter`: (i) skimage discard the pixel outside of the structural element when computing the median while scipy will pad the image beforehand and (ii) the tie breaking is done differently --- floor for skimage and ceiling for scipy.

Therefore my proposal is the following: introduce a parameter `behavior` allowing to switch to the scipy implementation using all the available parameters. By default, we will fall back on the previous behavior for 2 releases. In 2 releases, the scipy function will be the default and in 4 releases, the parameters of the old behavior will be removed at the same time than the `behavior` parameter.

I could not come with a cleaner design than those deprecation cycles.

## Checklist

  existing benchmark
- [x] Unit tests

## References
closes #1419 
closes #1422
closes #2943

## For reviewers

  later.

### DIFF
--- a/skimage/filters/_median.py
+++ b/skimage/filters/_median.py
@@ -9,7 +9,7 @@ from .rank import generic
 
 @generic._default_selem
 def median(image, selem=None, out=None, mask=None, shift_x=False,
-           shift_y=False, mode='nearest', cval=0.0, behavior='old'):
+           shift_y=False, mode='nearest', cval=0.0, behavior='rank'):
     """Return local median of an image.
 
     Parameters
@@ -17,8 +17,8 @@ def median(image, selem=None, out=None, mask=None, shift_x=False,
     image : array-like
         Input image.
     selem : ndarray
-        If ``behavior=='old'``, ``selem`` is a 2-D array of 1's and 0's.
-        If ``behavior=='new'``, ``selem`` is a N-D array of 1's and 0's with
+        If ``behavior=='rank'``, ``selem`` is a 2-D array of 1's and 0's.
+        If ``behavior=='ndimage'``, ``selem`` is a N-D array of 1's and 0's with
         the same number of dimension than ``image``.
         If None, ``selem`` will be a N-D array with 3 elements for each
         dimension (e.g., vector, square, cube, etc.)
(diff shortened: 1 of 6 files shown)


stefanv:
Perhaps the following deprecation pathway:

1. Add a behavior flag, set to 'rank' for now, but with the option to set it to 'ndimage'.  If the user does not specify this flag, warn that we will switch behavior to ndimage in the future.
2. After one release, switch the default to ndimage.  If the flag is not set, warn the user that this function has changed, and that they should adapt their code.
3. After one more release, remove the behavior flag and make `ndimage` the default.

In addition, update the docstring See Also section with a reference to the rank filters, so that users know that this fast option is available for integer images.

Add a docstring to either `filters` or `filters.rank`, describing what these are, so that they are more discoverable.

In general, we should consider adding paragraph descriptions to each of the submodule `__init__.py`.

stefanv:
Ah, @glemaitre I did not read your PR description fully before commenting, so, yes, I guess we are in agreement.  Except that I wanted us to move a little bit more quickly.

stefanv:
For reference, our [deprecation cycle documentation](http://scikit-image.org/docs/dev/contribute.html#deprecation-cycle).

glemaitre:
> Ah, @glemaitre I did not read your PR description fully before commenting, so, yes, I guess we are in agreement. Except that I wanted us to move a little bit more quickly.

I think that this is a pretty good idea to move faster for this case.

skimage/filters/_median.py
```diff
@@ -0,0 +1,98 @@
...
+@generic._default_selem
+def median(image, selem=None, out=None, mask=None, shift_x=False,
+           shift_y=False, mode='nearest', cval=0.0, behavior='rank'):
+    """Return local median of an image.
+
+    Parameters
+    ----------
+    image : array-like
+        Input image.
+    selem : ndarray
```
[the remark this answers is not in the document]
sciunto: damn, it looks like github added a space before the variable. I didn't intend to do that. Just add the optional information....

skimage/filters/_median.py
```diff
@@ -0,0 +1,98 @@
...
+    mask : ndarray
+        Mask array that defines (>0) area of the image included in the local
+        neighborhood. If None, the complete image is used (default). Only valid
+        when ``behavior='rank'``
+
+        .. deprecated:: 0.16
+           ``mask`` is deprecated in 0.16 and will be removed 0.17.
+    shift_x, shift_y : int
+        Offset added to the structuring element center point. Shift is bounded
+        to the structuring element sizes (center must be inside the given
```
emmanuelle: bounded by

skimage/filters/_median.py
```diff
@@ -0,0 +1,98 @@
...
+        structuring element). Only valid when ``behavior='ndimage'``.
```
emmanuelle: when ``behavior='rank'`` (if I understand well??)

glemaitre:
> Also, when the input is a floating point array and behavior='rank', a warning should be raised I think.

It will be already the case as the other rank functions.

emmanuelle:
@glemaitre I was thinking of a warning advising to use the ndimage mode, but maybe the existing warning is enough.

emmanuelle:
Looks good to me!

PR #3617, opened 2018-12-20
Update Travis-CI to xcode 10.1

stefanv:
Let's see if this fixes our test segfault.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,15 +57,15 @@ matrix:
       python: 3.6
       env: QT=PyQt5 OPTIONAL_DEPS=1 PIP_FLAGS="--pre"
     - os: osx
-      osx_image: xcode9
+      osx_image: xcode10.1
       language: objective-c
       env: TRAVIS_PYTHON_VERSION=3.5
     - os: osx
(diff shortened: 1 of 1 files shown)


stefanv:
That seems to do the trick!

soupault:
What do you think about moving MacOS builds to Azure?

hmaarrfk:
I guess I missed the segfault. But I'm not sure if you can change the wheel one because of how it is supposed to target older macs too? :/

stefanv:
On Thu, 20 Dec 2018 05:53:08 -0800, Mark Harfouche wrote:
> I guess I missed the segfault. But I'm not sure if you can change the
> wheel one because of how it is supposed to target older macs too? :/

Yes, that was my impression too.

hmaarrfk:
The other issue is that it may be an actual bug, but that we dont see anymore in light of how new compilers are organizing the code

hmaarrfk:
can we downgrade xcode?

jni:
@scikit-image/core can we merge this? We could raise another issue and maybe a PR to revert the change, but I would very much like to have our CI go green again.

PR #3621, opened 2018-12-22
Sort gallery examples semantically

### LINKED ISSUE #3462: order the examples in our gallery
## Description

As suggested by @emmanuelle in #3388, it would be interesting to use the [ordering feature of sphinx-gallery](https://sphinx-gallery.readthedocs.io/en/stable/advanced_configuration.html?highlight=sorting#sub-gallery-order) a set a more logical order among sections and examples (eg operation on numpy arrays could come among the first ones).

PinkFloyded:
## Description
Sorts gallery subsections and the examples inside each subsection semantically. I used complexity as the main criteria to sort. [New gallery order](https://user-images.githubusercontent.com/11929269/50377825-4070ed80-064a-11e9-9a10-6eae31c35628.png).

The metric I've used is very subjective. So, please let me know if I need to change the order.

## Checklist
Closes #3462 

## For reviewers

  later.


emmanuelle:
Thank you very much! Explicit ordering of gallery sections is a real progress, I'm less sure about the order of examples within a subgallery (it's one more thing to ask to people contributing an example to the gallery. Or is it possible to specify the order for a few items and the other ones come next?). Also, Travis is failing with minimal dependencies (do we need to increase the minimal version for sphinx-gallery?).

sciunto:
Glad to see a PR on this.

@PinkFloyded To avoid the test failure, you must  add your gallery_order dependency in requirements/docs.txt please.

PinkFloyded:
@sciunto It's a module I've newly created which was at `doc/source` and now has been moved to `doc/ext`. It was causing `ModuleNotFoundError` as `doc/source` wasn't in `sys.path`. Not sure how it was working on Python 3.6.

> I'm less sure about the order of examples within a subgallery (it's one more thing to ask to people contributing an example to the gallery. Or is it possible to specify the order for a few items and the other ones come next?)

@emmanuelle Makes sense. I've made changes to add examples at the end if it's order is not explicitly specified.

hmaarrfk:
Is this maintainable? Are there notes on where new examples should be added?

PinkFloyded:
@hmaarrfk It should be. If the examples are to be shuffled around, only the order in the gallery variable needs to be changed.

> Are there notes on where new examples should be added?

I'm not sure where this should be documented. One option could be in the [guidelines section of the contribution guide](http://scikit-image.org/docs/stable/contribute.html#guidelines). The last bullet can be edited to say "Examples in the gallery should have a maximum figure width of 8 inches and it's position within the subsection should be specified in `doc/ext/gallery_order.py`". Another approach is to throw an error if it's order is not specified. This forces the contributor to think about it's order, but also as @emmanuelle mentioned adds another thing he has to do in order to add an example. Let me know if there is another way this can be documented.

hmaarrfk:
I generally don't like editing multiple files for the "addition" or "removal" of something.

Would something like `__gallery_priority__ = 10` make sense? Then we sort them in decreasing order.
If `__gallery_priority__` isn't defined, then it takes the value `-1`???

This adds noise to the example files though....

emmanuelle:
And what would you think of ordering only sections, not examples?

PinkFloyded:
@emmanuelle This would be the simplest. But, the examples are sorted by default using [`NumberOfCodeLinesSortKey`](https://sphinx-gallery.readthedocs.io/en/stable/gen_modules/sphinx_gallery.sorting.html#sphinx_gallery.sorting.NumberOfCodeLinesSortKey), which may lead to arbitrary orders. Though, this should not be such a big concern and can be added later when the control is worth the complexity.

PR #3622, opened 2018-12-22
WIP: Reduce the default tolerance in threshold_li

### LINKED ISSUE #3605: threshold_li changed from version 0.14.0 to version 0.14.1
## Description

```python
import numpy as np

image = np.load("test.npy")
thresh = threshold_li(image)

ax1 = plt.subplot(121)
ax1.imshow(image)

ax2 = plt.subplot(122)

plt.savefig("result.png")
```

Old code (0.14.0):
result_14 0

New code (0.14.1):
result_14 1

jni:
Fixes #3605

In that issue, I found that a tolerance of `range / 2**15` was
sufficient.  However, I don't know whether this is generally true, so an
extra factor of 2 is probably a good margin, and gets us to a nice
round number.

Still to do:

- [ ] Find an appropriate test case. I might end up using pooch/external data for this, so that I can use the example image from that issue.
- [ ] Update the release notes both for dev and 0.14.1.
- [ ] Write a dev gallery example about `threshold_li`. Possibly after merging `attribute_operators`.

## Checklist
  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -527,8 +527,8 @@ def threshold_li(image, *, tolerance=None):
 
     tolerance : float, optional
         Finish the computation when the change in the threshold in an iteration
-        is less than this value. By default, this is the smallest difference
-        between intensity values in ``image``.
+        is less than this value. By default, this is half the smallest
+        difference between intensity values in ``image``.
 
     Returns
     -------
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -527,8 +527,8 @@ def threshold_li(image, *, tolerance=None):
 
     tolerance : float, optional
         Finish the computation when the change in the threshold in an iteration
-        is less than this value. By default, this is half of the range of the
-        input image, divided by 256.
+        is less than this value. By default, this is the smallest difference
+        between intensity values in ``image``.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


jni:
A small update: while I was playing around with the test image, I indeed found that a factor of 2 over the empirically found limit above was insufficient to ensure robustness. After I sampled pixels from the test image in #3605 to get a 1024x1024 image, `range / 2**16` still got stuck in the low-gradient area. Part of the problem was that the image has a range greater than 8, but 98% of the pixels are in a much smaller range of less than 1. So the image range is inflated by outliers. But I think this is a general problem, regardless of the number of outliers.

In the end, I found a threshold that I think works in all cases: the tolerance should be strictly less than the smallest difference between two consecutive intensity values. I've made that (half the smallest gap) the default. I think in some cases you might end up making more updates than if you just brute-forced the threshold, but at least it will always be correct (and on average it should be much faster).

There's a chance that we could dramatically speed this up by sorting the intensities and only updating the foreground and background means incrementally, but I leave that for a later PR.

The rest of my to-do list remains to be done, just wanted to update with the current status.

skimage/filters/thresholding.py
```diff
@@ -578,8 +578,7 @@ def threshold_li(image, *, tolerance=None):
     # Li's algorithm requires positive image (because of log(mean))
     image_min = np.min(image)
     image -= image_min
-    image_range = np.max(image)
-    tolerance = tolerance or 0.5 * image_range / 256
+    tolerance = tolerance or np.min(np.diff(np.unique(image))) / 2
```
soupault: From above:
>By default, this is the smallest difference between intensity values in ``image``.

Here, it is a half of that difference.

jni: @soupault thanks, fixed!

PR #3626, opened 2018-12-25
Added glossary to the doc

emmanuelle:
I added a glossary to the doc, in order to define terms which we use a lot in the documentation. The idea is to add more redundancy and cross-linking within the doc, so that users are more likely to find what they're looking for. I only added a few terms for a start, but it can be completed later on (I can open an issue with a wishlist of terms). If we have a lot of terms in the glossary, we could organize it with sections, as in the one of scikit-learn https://scikit-learn.org/stable/glossary.html

I have used the built-in glossary structure of sphinx.

screenshot from 2018-12-26 13-56-05


emmanuelle:
Hi @alexandrejaguar it's not section and subsections but rather synonyms or close words which correspond to the same entry. I agree it looks a bit weird. If there were more space between entries it would look better, I'll see if it's possible.

jni:
Could synonyms be comma-separated instead of newline separated?

doc/source/glossary.rst
```diff
@@ -0,0 +1,78 @@
...
+    contrast
+        Differences of intensity or color in an image, which make objects
+        distinguishable. Several functions to manipulate the contrast of an
+        image are available in :mod:`skimage.exposure`. See :ref:`exposure`.
+
+    float
+    float values
+        Representation of real numbers, for example as :obj:`np.float32` or
+        :obj:`np.float64`. See :ref:`data_types`. Some operations on images
+        need a float datatype (such as multiplicating image values with
```
jni:
```suggestion
        need a float datatype (such as multiplying image values with
```

doc/source/glossary.rst
```diff
@@ -0,0 +1,78 @@
...
+        exponential prefactors in :func:`filters.gaussian`), so that 
+        image of integer type are converted to float type internally. Also
```
jni:
```suggestion
        images of integer type are often converted to float type internally. Also
```

doc/source/glossary.rst
```diff
@@ -0,0 +1,78 @@
...
+    int
+    int values
+        Representation of integer numbers, which can be signed or not, and
+        encoded on one, two, four or eight bytes according to the maximum value
+        which needs to be represented. In ``scikit-image``, the most common
+        integer types are :obj:`np.int64` (for large integer values) and
+        :obj:`np.uint8` (for small integer values, typically images of labels
+        with less than 255 labels). See :ref:`data_types`. 
+
+    labels
```
jni:
```suggestion
    labels
    label image
```

emmanuelle:
@jni hum, probably with some css...

emmanuelle:
So, I did not find an easy way to add colons between synonyms only (and not after each entry), therefore I added some space between entries using css. I updated the picture above.

sciunto:
Great addition. I agree that comas would be better, but this can be implemented later as an enhancement. @emmanuelle If you are done with it, it's ready to go in master and be a place for future other contributions.

jni:
I love it! The spacing makes a huge difference! =) Thanks @emmanuelle! Further terms can be added in future PRs! =)

soupault:
This is a great addition! Thanks @emmanuelle !

PR #3627, opened 2018-12-26
Add explicit ordering of gallery sections

### LINKED ISSUE #3462: order the examples in our gallery

PinkFloyded:
## Description
Add explicit ordering of gallery sections.

## References
Closes #3462 
Related discussion at: #3621 

## For reviewers

  later.


jni:
Thanks @PinkFloyded!

jni:
btw, what happens if a section is not mentioned in the `ExplicitOrder`? Does it not get served? If so, then we should have a reminder in the guide that new sections need to be added here.

PinkFloyded:
If it is not mentioned in the `ExplicitOrder` then it doesn't build and throws an error:
```
ValueError: If you use an explicit folder ordering, you must specify all folders. Explicit order not found for ../examples/new_section
```
I'm not sure where this can be documented. Please let me know and I will add it in.

jni:
Personally, I think that error message is explicit enough that we don't need to add a note, but for the record it would go in the [contributing guide](http://scikit-image.org/docs/dev/contribute.html), the source of which is [here](https://github.com/scikit-image/scikit-image/blob/master/CONTRIBUTING.txt). Looking it over, there isn't a natural place there (it does not explicitly state how to add examples to the gallery), so it would require extensive modifications.

PR #3634, opened 2018-12-30
cloudpickle is really a doc dependency, not a core one

hmaarrfk:
I added a comment to the dependency since the dependency was missed when this was merged

https://github.com/scikit-image/scikit-image/pull/3582

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -7,3 +7,5 @@ sphinx-copybutton
 pytest-runner
 scikit-learn
 dask[array]>=0.9.0
+# cloudpickle is necessary to provide the 'processes' scheduler for dask
+cloudpickle>=0.2.1
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -5,4 +5,3 @@ networkx>=2.0
 pillow>=4.3.0
 imageio>=2.0.1
 PyWavelets>=0.4.0
-cloudpickle>=0.2.1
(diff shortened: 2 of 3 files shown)


requirements/docs.txt
```diff
@@ -7,3 +7,5 @@ sphinx-copybutton
 pytest-runner
 scikit-learn
 dask[array]>=0.9.0
+# cloudpickle is necessary to provide the 'processes' scheduler for dask
+cloudpickle>=0.2.1
```
soupault: According to the https://github.com/scikit-image/scikit-image/pull/3582/files, this should be added also to `requirements/optional.txt`.

hmaarrfk: I'm not too sure about that. It is only useful if you want the `'processes'` backend for computation.

See related PR and the issue it closes.
https://github.com/dask/dask/pull/4342

jni: Hmmm, I'm surprised that cloudpickle isn't required for the distributed backend?

soupault: >I'm not too sure about that. It is only useful if you want the 'processes' backend for computation.

I guess it wouldn't harm to add it to avoid potential further issues (i.e. when those 2 backends are chosen by user)?

hmaarrfk: but as dask isn't part of the depenencies anymore, should it be part of the "optional" dependencies?

stefanv:
Yes, since that change optional dependency makes more sense to me.

hmaarrfk:
Truthfully, we should probabaly implement scikit-image[all] for 14.2

PR #3639, opened 2019-01-02
Add more details about the home page build process

jni:
## Description
A recent conversation with @stefanv revealed that I had missed a step after building the site home page, and that my oversight had been due to a missing step in the instructions. (I think others have fallen into this trap, which might explain why the "news" section on the home page was so often out of date.) 

This PR clarifies the procedure for updating the home page after a new release.

Should come in handy in the next few days... ;)

### DIFF
--- a/RELEASE.txt
+++ b/RELEASE.txt
@@ -124,7 +124,9 @@ the corresponding release branch.
    twine upload dist/scikit-image-<major>.<minor>.0.tar.gz
 
 - Update the web frontpage:
-  The webpage is kept in a separate repo: scikit-image-web
+  The webpage source is kept in a separate repo: scikit-image-web. The
+  live webpage lives in a second repo, scikit-image.github.com, that you only
+  interact with via the ``make`` command within scikit-image-web.
 
   - Sync your branch with the remote repo: ``git pull``.
     If you try to ``make gh-pages`` when your branch is out of sync, it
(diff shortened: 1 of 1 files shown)


soupault:
I'd really love seeing us using 2 repos instead of 4 :/...

stefanv:
@soupault Which four repos are you referring to? We have the docs in the scikit-image repo, and then a website source and a published website.

soupault:
@stefanv exactly those: scikit-image, docs, website source, website published. I might not remember all the peculiarities of the docs building and website publishing processes, but, presumably, 2 repos should be enough for everything: one with the source code, another with the built and composed website.

jni:
@soupault docs are part of the main repo. =)

soupault:
@jni, but the website source is not! Let me unwrap my ASCII tools :slightly_smiling_face: :
```
*********************************** Currently ***********************************

scikit-image     -> docs (generated)                    -> [docs] (served)
scikit-image-web -> scikit-image.github.com (generated) -> [main page] (served)

*********************************** Imagined ************************************

scikit-image  |  -> [main page] (generated) --> scikit-image-web -> [website] (served)
              |  -> [docs] (generated)      -/
```

stefanv:
Yes, the website is its own thing, and its sources lives in [its own repo](https://github.com/scikit-image/scikit-image-web/).  I think there's general agreement that we don't want our website mixed up with our source code.

The [docs live with the source](https://github.com/scikit-image/scikit-image/tree/master/doc), because they are closely related.

There is then [one place](https://github.com/scikit-image/scikit-image.github.com) from which GitHub serves the website.  This is where all published products go.

soupault:
>This is where all published products go.

I'm not exactly sure what you meant there, but https://github.com/scikit-image/docs/ is served independently as http://scikit-image.org/docs/ (see https://github.com/scikit-image/docs/settings).

Anyway, if there are any concerns about my proposal, I'm fine with not changing anything. :slightly_smiling_face:

Issue #3082, opened 2018-05-16
Forward Energy Filter for Improved Seam Carving


axu2:
## Description
So I implemented the forward energy algorithm from this paper:
http://www.faculty.idc.ac.il/arik/SCWeb/vidret/index.html

Basically, it is a new way of calculating energies of an image for seam carving that isn't afraid to shorten continuous edges.

This is the original image from the website.

![image](https://user-images.githubusercontent.com/20757319/40094370-972f0e7c-5894-11e8-9619-588e6efebd9c.png)

This is the energy from the built in sobel filter (backwards energy)

![image](https://user-images.githubusercontent.com/20757319/40094386-b49952ce-5894-11e8-8543-be45c7b31a48.png)

And this is the energy from the function I implemented:

![image](https://user-images.githubusercontent.com/20757319/40094402-cbf35438-5894-11e8-9fb3-395394ae03d9.png)

It's darker around the middle since a seam can pass through the horizontal boards and not destroy the continuous edge.

It's quite fast, and the difference in results is quite dramatic when I seam carve 200 pixels:

![image](https://user-images.githubusercontent.com/20757319/40094420-ed835a58-5894-11e8-95d4-532a94c59f4e.png)

Speed difference: 3 sec vs 8 sec.

Do you think it's worth for me to build a pull request for this feature?

Here is my current implementation:

```
def forward_energy(img, flag=False):
    height = img.shape[0]
    width = img.shape[1]
    I = color.rgb2gray(img)
    
    energy = np.zeros((height, width))
    m = np.zeros((height, width))
    
    U = np.roll(I, 1, axis=0)
    L = np.roll(I, 1, axis=1)
    R = np.roll(I, -1, axis=1)
    
    cU = np.abs(R - L)
    cL = np.abs(U - L) + cU
    cR = np.abs(U - R) + cU
    
    for i in range(1, height):
        mU = m[i-1]
        mL = np.roll(mU, 1)
        mR = np.roll(mU, -1)
        
        mULR = np.array([mU, mL, mR])
        cULR = np.array([cU[i], cL[i], cR[i]])
        mULR += cULR

        argmins = np.argmin(mULR, axis=0)
        m[i] = np.choose(argmins, mULR)
        energy[i] = np.choose(argmins, cULR)
        
    return energy
```

Here is the non-vectorized implementation:

```
def slow_forward_energy(img):
    height = img.shape[0]
    width = img.shape[1]
    
    I = color.rgb2gray(img)
[...]

PR #3643, opened 2019-01-06
add forward energy edge filter for seam carving

axu2:
## Description
Looking for code review as my first real open source contribution!

For more details see the original issue #3082

I added an example in the pr.

## Checklist
  existing benchmark
- [x] Unit tests

## References
Implements this paper:
http://www.faculty.idc.ac.il/arik/SCWeb/vidret/index.html

## For reviewers

  later.


axu2:
Also, unsure why CI failed?

hmaarrfk:
you broke all of python!!!!

hmaarrfk:
probably some bug in a pytest 4.1, you can try to exclude it in the test depencendies

see the `requirements` directory

axu2:
Looks ok now.

jni:
@axu2 nice work! The code is very clear.

Could I ask you to improve the example to discuss *why* you would use forward energy over Sobel? My idea for what the example should look like is:

- old example (with the rocket)
- start of a second section: what about this other image? (bench)
- this doesn't look great, and the reason is [why sobel fails in this use case]
- to help with this, Rubinstein et al proposed the *forward energy filter*, which measures [short description of the forward energy filter]
- this results in this better output [side by side bench results with Sobel and forward energy].

The short description should also go in the docstring of the forward energy filter, by the way. It would help me in my review to know *what* the filter is trying to measure. Currently I can read the steps but I'm missing the big picture.

Finally, we use the [NumPy docstring format](https://docs.scipy.org/doc/numpy/docs/howto_document.html) rather than the ReStructuredText format.

Thanks again, I'll give this another round once you've added those descriptions!

axu2:
@jni How's that?

doc/examples/transform/plot_seam_carving.py
```diff
@@ -83,3 +139,63 @@
...
+######################################################################
+
+out = seam_carve(img, energy, 'vertical', 400)
+plt.figure()
+plt.title('Resized using Seam Carving')
+plt.imshow(out)
+
+######################################################################
+# Seam carving using the Sobel edge filter distorts the person
+# and unncessarily preserves pixels along the bench and the fence
```
jni:
```suggestion
# and unnecessarily preserves pixels along the bench and the fence
```

skimage/filters/edges.py
```diff
@@ -534,3 +534,73 @@ def laplace(image, ksize=3, mask=None):
...
+        mL = np.roll(mU, 1)
+        mR = np.roll(mU, -1)
+
+        mULR = np.array([mU, mL, mR])
+        cULR = np.array([cU[i], cL[i], cR[i]])
+        mULR += cULR
+
+        argmins = np.argmin(mULR, axis=0)
+        m[i] = np.choose(argmins, mULR)
+        energy[i] = np.choose(argmins, cULR)
```
jni: I'm still unclear on what the implementation is doing. Probably some comments to map these arrays to the intuition behind the method (the forward energy) would be useful.

skimage/filters/edges.py
```diff
@@ -534,3 +534,73 @@ def laplace(image, ksize=3, mask=None):
...
     result = convolve(image, laplace_op)
     return _mask_filter_result(result, mask)
+
+
+def forward_energy(image, mode, mask=None):
+    """Find the edge magnitude using forward energy for seam carving.
+
+    Depending on the direction in `mode`, determines the magnitude of
+    each pixel based on the new edges created after removing a seam
+    containing that pixel.
```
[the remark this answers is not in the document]
axu2: https://www.youtube.com/watch?v=AJtE8afwJEg

This video would probably be helpful for understanding.

stefanv:
I'm sorry, folks, but sadly @jni is right.  This algorithm [is patented](https://patents.google.com/patent/US8213745B2/en) and will need to go.

I'm OK with someone taking the code and making it available somewhere else.  It would be good to make explicit mention of the issue (although our license does not require that, specifically).

jni:
@axu2 I'm sorry, but for the reasons stated above, we cannot accept this code into scikit-image (and the original code is in the process of being removed). It's a shame because it is good work. If you host it somewhere public, feel free to post a link to this thread for future reference.

grlee77:
As detailed in https://github.com/scikit-image/scikit-image/issues/3082#issuecomment-1083631678, I think this patent is now expired for non-payment of maintenance fees, so we can possibly resurrect this!

Issue #1605, opened 2015-07-13
Remove IO package in favor of imageio


blink1073:
This is intended for `scikit-image` v1.0:
- Replace `PIL` dependency with `imageio`
- Import all of `imageio` API into `skimage.io` namespace.
- Move `ImageCollection` to `imageio`
- Keep image stack
- Remove `external.tifffile`
- Use `matplotlib` only for `show`.

I have added plugins to `imageio` for `tifffile`, `gdal`, `fits`, and `itk`.

`imageio` only depends on `libfreeimage`, which is included in Debian repos, and will be included within the wheel installs for OSX and Windows.  There is a cross-platform `conda` recipe for `libfreeimage`.

Benefits:
- No `PIL`!
- Plugins are selected based on file extension, but can be overridden.
- A sane API for multi-image read/write.
- We benefit from any additional plugins that are added.

blink1073:
cc @almarklein

matthew-brett:
I notice that imageio does not currently have an OSX wheel on pypi - @almarklein - would you like me to set up a repo to build this, on the same lines as:

* https://github.com/MacPython/numpy-wheels
* https://github.com/MacPython/scipy-wheels

etc?

almarklein:
Support for wheels is in progress. I think we've got the OSX wheels covered, right @blink1073 ?

blink1073:
@matthew-brett, I am new to OSX wheels, your expertise would be greatly appreciated.  I just naively created `imageio-1.3-cp27-none-macosx_10_5_x86_64.whl` based on the library that @almarklein had pre-compiled: https://github.com/imageio/imageio-binaries.


https://github.com/imageio/imageio/pull/108

vighneshbirodkar:
If I understand correctly,  we plan to remove the io module completely?

blink1073:
It would contain `stack`, `imshow` and `show`, and then the `imread` and friends from `imageio`'s API: https://imageio.readthedocs.org/en/latest/userapi.html

matthew-brett:
Ah - OK - I will create a MacPython repo for this.

The tricks are:

* to make the wheels compatible with Python.org and System Pythons, which requires a dual-arch build of the libraries and the Python extensions, and:
* embedding the dynamic library in the wheel, maybe with `delocate` or doing static linking, so the personal installing the wheel does not need to also copy the library.

vighneshbirodkar:
@blink1073 What do you mean by friends ? 
[...]

PR #3645, opened 2019-01-08
optional dependencies on pip

hmaarrfk:
Just a warning, there is no wheel file for [`imread`](https://pypi.org/project/imread/#files), an optional dependency, so there is virtually 0 way that this will work on anybody's machine....

Test on your machine
```
conda create --name sk-test-pip-optional --offline --yes python=3
conda activate sk-test-pip-optional
pip install numpy cython
pip install -e ".[test]"  --disable-pip-version-check --no-build-isolation
```

Output:
```
[...] # long stuff
Successfully installed codecov-2.0.15 coverage-4.5.2 flake8-3.6.0 mccabe-0.6.1 pycodestyle-2.4.0 pyflakes-2.0.0 pytest-cov-2.6.1 scikit-image
```

## Checklist
  existing benchmark
- [ ] Unit tests

## References

## For reviewers

  later.


jni:
@baerbock please refrain from making such comments on any scikit-image forum. Please refer to our [community guidelines](https://scikit-image.org/community_guidelines.html) and only make constructive criticism.

Additionally, regardless of any *valid* criticism you might have of conda and pip (the above clearly isn't), they are the de facto standard distribution channels in Python and we *must* support them.

setup.py
```diff
@@ -112,8 +112,20 @@ def build_extensions(self):
...
+        requires = [l.strip() for l in fid.readlines() if l]
+
+    return requires
+
+
+INSTALL_REQUIRES = parse_requirements_file('requirements/default.txt')
+extras_require = {
+    'docs': parse_requirements_file('requirements/docs.txt'),
+    'optional': parse_requirements_file('requirements/optional.txt'),
+    'test': parse_requirements_file('requirements/test.txt'),
```
hmaarrfk: i'm not too sure about this test dependency.

soupault: Same can be said for `docs` :). I'm fine with the current version.

hmaarrfk:
@scikit-image/core can we drop the dependency on `imread`??? It is considered beta software without a wheel on pypi.

jni:
I don't have any insight about the imread dependency. @stefanv? I'd be in favour of dropping it since Pillow seems to cover most of our bases.

soupault:
@jni why Pillow? I though we're going to switch eventually to imageio - https://github.com/scikit-image/scikit-image/issues/1605.

hmaarrfk:
"not depending" on pillow is a little moot since "imageio" depends on it

the default imageio backend is pillow. They just figured out how to avoid a memory copy for pillow.

I went down the rabbit hole a while back and figured out that imread was using pillow quite effectively:
https://github.com/hmaarrfk/redpil
(see benchmarks at bottom).

jni:
@soupault what @hmaarrfk said. And anyway the point I was making is that we *do* depend on pillow right now (see `requirements/default.txt`), so probably we don't need to depend on `imread` *right now*. @stefanv what do you think about removing this dependency?

@hmaarrfk very cool benchmarks, and I appreciate the multi-level pun there! =D

setup.py
```diff
@@ -112,8 +112,19 @@ def build_extensions(self):
...
+    with open(filename) as fid:
+        requires = [l.strip() for l in fid.readlines() if l]
+
+    return requires
+
+
+INSTALL_REQUIRES = parse_requirements_file('requirements/default.txt')
+extras_require = {
+    dep : parse_requirements_file('requirements/' + dep + '.txt')
+    for dep in ['docs', 'optional', 'test']
```
sciunto: Perhaps, a comment to say that extra is omitted on purpose.

hmaarrfk: I hope that one helps! thanks for the input.

PR #3648, opened 2019-01-14
Delay matplotlib import in skimage/future/manual_segmentation.py

GenevieveBuckley:
## Description
Bugfix for missing matplotlib import error in `skimage.future.manual_segmentation.py`.

Currently if you try to run the docstring examples from the `manual_polygon_segmentaion` or `manual_lasso_segmentation` functions, this error is produced (also verified independently by @jni):
```
File '...\scikit-image\skimage\future\manual_segmentation.py", line 92, in manual_polygon_segmentation
    undo_button = matplotlib.widgets.Button(undo_pos, u'\u27F2')
NameError: name 'matplotlib' is not defined
```

It seems that as part of the larger effort to remove matplotlib as a dependency (https://github.com/scikit-image/scikit-image/pull/3129), the line `import matplotlib` inadvertently got dropped in PR https://github.com/scikit-image/scikit-image/pull/3533, commit https://github.com/scikit-image/scikit-image/commit/fb81456bf527f93ec0ad7a968b02de6eba0ad8cb. It's necessary for the undo buttons, but flew under the radar because pytest skips these docstring examples.

@hmaarrfk 
Mark, is this the right place to add that line back in, or should it go next to the `import matplotlib.pyplot as plt` lines inside the manual segmentation functions? You've spent a lot more time thinking about how to handle this.

## Checklist
  existing benchmark
- [ ] Unit tests

## References
PR https://github.com/scikit-image/scikit-image/pull/3533, commit https://github.com/scikit-image/scikit-image/commit/fb81456bf527f93ec0ad7a968b02de6eba0ad8cb

## For reviewers

  later.

### DIFF
--- a/skimage/future/manual_segmentation.py
+++ b/skimage/future/manual_segmentation.py
@@ -1,8 +1,5 @@
 from functools import reduce
 import numpy as np
-import matplotlib
-from matplotlib.patches import Polygon
-from matplotlib.collections import PatchCollection
 from ..draw import polygon
 
 
@@ -64,6 +61,9 @@ def manual_polygon_segmentation(image, alpha=0.4, return_all=False):
     >>> io.imshow(mask)  # doctest: +SKIP
     >>> io.show()  # doctest: +SKIP
     """
+    import matplotlib
+    from matplotlib.patches import Polygon
+    from matplotlib.collections import PatchCollection
     import matplotlib.pyplot as plt
 
     list_of_vertex_lists = []
@@ -176,6 +176,9 @@ def manual_lasso_segmentation(image, alpha=0.4, return_all=False):
     >>> io.imshow(mask)  # doctest: +SKIP
     >>> io.show()  # doctest: +SKIP
     """
+    import matplotlib
+    from matplotlib.patches import Polygon
+    from matplotlib.collections import PatchCollection
     import matplotlib.pyplot as plt
 
     list_of_vertex_lists = []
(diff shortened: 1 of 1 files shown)


jni:
btw I think this is the right place for this import. `import matplotlib` does not set the backend, which I believe was the whole problem that #3533 was trying to solve.

hmaarrfk:
sorry for the delay, many PRs were comming through and I didn't have a chance to see it.


I think jni is right that this doesn't set the backend.

https://github.com/scikit-image/scikit-image/pull/3129 needs a good push forward.

hmaarrfk:
honestly, can we place this inside the function that needs it.

I'm not sure there is enough bandwidth to completely remove matplotlib as a dependency, but at least delaying importing can allow downstream developer to "not install matplotlib" and scikit-image would still work (but the tests would fail).

GenevieveBuckley:
@hmaarrfk - do I understand correctly, you would like to delay importing all the matplotlib imports currently at the top of the manual_segmentation.py file?

```
import matplotlib
from matplotlib.patches import Polygon
from matplotlib.collections import PatchCollection
```

hmaarrfk:
yeah, that is effectively the last change needed to move away from a hard dependency on matplotlib.

at that point matplotlib would only be a "testing" dependency (and for polyclip).

but please just take this as a suggestion. my main gripe with matplotlib is more that I don't want it in my ci tests.

jni:
Woot! @scikit-image/core still waiting for another review...

GenevieveBuckley:
Thanks @soupault - I've updated with the change you suggest.

PR #3650, opened 2019-01-15
Use user-installed tifffile if available

### LINKED ISSUE #3644: tifffile: try to use the one in the user's install first
Should we try importing tifffile before using the one we versionned it?

jni:
Fixes #3644

Note that we already have tifffile in our optional dependencies, so
these lines will be tested by one configuration or another of our test
matrix. (If I remember correctly, there's no easy way for coverage to be
measured across the whole matrix.)

- [x] ~~Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark~~
- [x] Unit tests

## For reviewers

  later.


skimage/io/_plugins/tifffile_plugin.py
```diff
@@ -1,4 +1,7 @@
-from ...external.tifffile import TiffFile, imsave
+try:
+    from tifffile import TiffFile, imsave
+except ImportError:
```
hmaarrfk: Does this work in 2.7 and 3.5?

hmaarrfk:
Coverage has an option to combine reports from different cis

sciunto:
then, shouldn't we test against the last version of tifffile in our CI?

hmaarrfk:
tifffile 2019 is installed 
https://travis-ci.org/scikit-image/scikit-image/jobs/479729277#L2617

hmaarrfk:
This may cause a conflict with https://github.com/scikit-image/scikit-image/pull/3675
so i'm going to wait to hear back on 3675 before merging this.

jni:
:+1:

hmaarrfk:
I think the question now is how to accomodate the different parameters in the various versions of tifffile.

cc: @cgohlke

cgohlke:
The next version of tifffile will ignore the `fastij` and `multifile_close` parameters, and map the `pages` parameter to `key` in `tifffile.imread`. That should be (mostly) backwards compatible for skimage if the skimage tifffile plugin calls `tifffile.imread` instead of `with TiffFile... tif.save...`. Using both `pages` and `key` in `imread` will raise a `TypeError` as does passing `pages` to `TiffFile.__init__`.

hmaarrfk:
Is the issue:

```
    if 'img_num' in kwargs:
        kwargs['key'] = kwargs.pop('img_num')
    # parse_kwargs will extract keyword arguments intended for the TiffFile 
    # class and remove them from the kwargs dictionary in-place
    tiff_keys = ['multifile', 'multifile_close', 'pages', 'fastij', 'is_ome']
    kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
```

I feel like the solution is to:
1. Add a deprecation warning to discourage the use of `img_num`?

scikit-image should not encourage the use of old APIs without a deprecation policy

hmaarrfk:
I restarted the builds on travis. If they pass, I'll merge.

We can address compatibility with future tifffile in the future.

Issue #2531, opened 2017-02-21
"imsave" is too slow to check for "is_low_contrast", which is just for a warning.


cesc-park:
## Description
[Please provide a general introduction to the issue/proposal.]

For rgb (jpg) image.

```
from PIL import Image
image = Image.open(f)
image.save(save_name)

```
The code above is 10 times faster than the code below.
```
from skimage import io
image = io.imread(f)
io.imsave(save_name, image)

```
I think it is because of checking "is_low_contrast".
If this happens not only to me but also to the general, I think it is too big a waste to give a warning.

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [V] Code example

christian-rauch:
I agree with this. I am regularly exporting pixel-wise label images, where the grey value encodes the ID of an object. Obviously, these images are not natural images and have low contrast. Checking and printing this warning seems to slow down the image export by quite some factor.
I am in favour of completely removing this check. A low-contrast check is only useful for proper photo processing applications like GIMP.

hmaarrfk:
@christian-rauch I am sympathetic with your pains.

I don't think imageio does any of these checks. skimage mostly moved to imageio in (0.15???) anyway.

soupault:
Closed via #3653

PR #3653, opened 2019-01-16
make the low contrast check optional when saving images

christian-rauch:
## Description
This PR makes the low constrast check optional when saving images.
As described in https://github.com/scikit-image/scikit-image/issues/2531 the low contrast check delays the image export and is generally only useful for natural images.

## References
Fixe #2531 

## For reviewers

  later.

### DIFF
--- a/skimage/io/_io.py
+++ b/skimage/io/_io.py
@@ -104,7 +104,7 @@ def imread_collection(load_pattern, conserve_memory=True,
                        plugin=plugin, **plugin_args)
 
 
-def imsave(fname, arr, plugin=None, **plugin_args):
+def imsave(fname, arr, plugin=None, check_contrast=True, **plugin_args):
     """Save an image to file.
 
     Parameters
@@ -113,11 +113,13 @@ def imsave(fname, arr, plugin=None, **plugin_args):
         Target filename.
     arr : ndarray of shape (M,N) or (M,N,3) or (M,N,4)
         Image data.
-    plugin : str
+    plugin : str, optional
         Name of plugin to use.  By default, the different plugins are
         tried (starting with imageio) until a suitable
         candidate is found.  If not given and fname is a tiff file, the
         tifffile plugin will be used.
+    check_contrast : bool, optional
+        Check for low contrast and print warning (default: True).
 
     Other parameters
     ----------------
@@ -135,7 +137,7 @@ def imsave(fname, arr, plugin=None, **plugin_args):
     if plugin is None and hasattr(fname, 'lower'):
         if fname.lower().endswith(('.tiff', '.tif')):
             plugin = 'tifffile'
-    if is_low_contrast(arr):
+    if check_contrast and is_low_contrast(arr):
         warn('%s is a low contrast image' % fname)
     if arr.dtype == bool:
         warn('%s is a boolean image: setting True to 1 and False to 0' % fname)
--- a/skimage/io/tests/test_pil.py
+++ b/skimage/io/tests/test_pil.py
@@ -203,6 +203,10 @@ def test_imsave_incorrect_dimension():
         with testing.raises(ValueError):
             with expected_warnings([fname + ' is a low contrast image']):
                 imsave(fname, np.zeros((2, 3, 2)))
+        # test that low contrast check is ignored
+        with testing.raises(ValueError):
+            with expected_warnings([]):
+                imsave(fname, np.zeros((2, 3, 2)), check_contrast=False)
 
 
 def test_imsave_filelike():


skimage/io/_io.py
```diff
@@ -104,7 +104,7 @@ def imread_collection(load_pattern, conserve_memory=True,
                        plugin=plugin, **plugin_args)
 
 
-def imsave(fname, arr, plugin=None, **plugin_args):
+def imsave(fname, arr, check_contrast=False, plugin=None, **plugin_args):
```
hmaarrfk: Until keyword only arguments are standard, this has to come after "plugin"

hmaarrfk: Unless you want to suggest a change in default behavior, I suggest you set this to `True`.

skimage/io/_io.py
```diff
@@ -113,6 +113,8 @@ def imsave(fname, arr, plugin=None, **plugin_args):
         Target filename.
     arr : ndarray of shape (M,N) or (M,N,3) or (M,N,4)
         Image data.
+    check_contrast : bool
```
hmaarrfk: The default value should be specified.

skimage/io/tests/test_pil.py
```diff
@@ -199,10 +199,10 @@ def test_imsave_incorrect_dimension():
     with temporary_file(suffix='.png') as fname:
         with testing.raises(ValueError):
             with expected_warnings([fname + ' is a low contrast image']):
-                imsave(fname, np.zeros((2, 3, 3, 1)))
+                imsave(fname, np.zeros((2, 3, 3, 1)), check_contrast=True)
```
hmaarrfk: I think when you revert the default behaviour, you can avoid specifying `check_contrast` here

skimage/io/tests/test_pil.py
```diff
@@ -213,7 +213,7 @@ def test_imsave_filelike():
     # save to file-like object
     with expected_warnings(['precision loss',
                             'is a low contrast image']):
-        imsave(s, image)
+        imsave(s, image, check_contrast=True)
```
hmaarrfk: You should add a test where no warning is raised.

skimage/io/_io.py
christian-rauch: Yes, I propose a change in the default behaviour because of the before mentioned issue.

skimage/io/tests/test_pil.py
christian-rauch: It looks like the original test only checked that a warning is raised. I am not intending to change the test.

hmaarrfk:
@christian-rauch I'm having the discussion here because github often closes the mini discussions that are around comments:

Unfortunately, scikit-image has a deprecation cycle policy.
http://scikit-image.org/docs/stable/contribute.html#deprecation-cycle

Sometimes, often for warnings, it is broken, but often requires discussion.

In this particular case, mostly because the release of 0.14.2 should be imminent (due to the fact that numpy 1.16 broke scikit-image), I suggest you keep the default behaviour change as a separate PR. It is therefore in your best interest to minimize the discussion as much as possible.

jni:
Thanks for the comments and PR, @christian-rauch! And thanks for chiming in, @hmaarrfk! Good advice. =) I think either way this won't be in 0.14.2 though, @stefanv and I are working on getting this out the door today. (Though no promises because crap always goes wrong. LOL)

Anyway, regarding low contrast, the problem is that before that message, we would always get people saying "I used scikit-image to save an image and it's all black when I open it in [Preview/Photoshop/Whatever]". *All the time*. That's what the warning is supposed to prevent, and actually my totally unscientific impression is that it's worked.

So my intuition is that we should:
- definitely indeed add an option to skip the check
- don't skip by default
- add a documentation page that explains the issue (especially with regards to data types), and make the warning both explain how to turn it off and link to the page with more info about the whole thing.

I am amenable to be convinced that I should just delete the damn thing altogether, but that'll take a bit more work. =)

I do also think long term we want to offload *all* io to imageio, but that is also a hard sell.

christian-rauch:
To increase the probability of being merged (and to make exporting less annoying) I reverted to the default behaviour. To prevent the check and warnings, `check_contrast` has to be set to False by the user.
I will leave the discussion on whether or not to remove the low contrast check completely for a later PR :-)

skimage/io/_io.py
hmaarrfk: should be: `check_contrast : bool, optional`

hmaarrfk: i guess the other params dont specify optional, but maybe you can add optional to them?

skimage/io/_io.py
```diff
@@ -113,7 +113,9 @@ def imsave(fname, arr, plugin=None, **plugin_args):
         Target filename.
     arr : ndarray of shape (M,N) or (M,N,3) or (M,N,4)
         Image data.
-    plugin : str
+    check_contrast : bool, optional
```
hmaarrfk: should be moved "below" plugin

skimage/io/tests/test_pil.py
```diff
@@ -203,6 +203,10 @@ def test_imsave_incorrect_dimension():
         with testing.raises(ValueError):
             with expected_warnings([fname + ' is a low contrast image']):
                 imsave(fname, np.zeros((2, 3, 2)))
+        # test that low contrast check is ignored
+        with testing.raises(ValueError):
```
hmaarrfk: why does this raise valueerror?

christian-rauch: I don't know. I just amended the part that is related to the low contrast warning exception.

hmaarrfk: It just doesn't like the fact that there are 2 channels in the last dimension of the array.
Not an ideal test, but I'm ok with how this looks now.

Thanks!

hmaarrfk:
@jni, this is looking good. Minimally invasive 9 line PR, maybe good enough to get backported too ;)

skimage/io/tests/test_pil.py
hmaarrfk: I looked into it a little further, there isn't currently a place to put these "tests for all plugins". So this is as good of a place as any. I don't think PIL is going away any time soon.

hmaarrfk:
@scikit-image/core I think this PR is good to go if anybody has time to give it a second pass.

ahojnnes:
Looks great to me, thanks for the contribution!

jni:
:tada:

stefanv:
Should this PR be accompanied by plans for deprecation?

jni:
@stefanv imho it's a valuable check to have around. See my comment above regarding its original motivation. However I won't stand in the way if others prefer making "off" the default.

PR #3657, opened 2019-01-17
Remove MANIFEST file when making the 'clean' target

stefanv:


Makefile
```diff
@@ -8,6 +8,7 @@ all:
 clean:
 	find . -name "*.so" -o -name "*.pyc" -o -name "*.md5" -o -name "*.pyd" | xargs rm -f
 	find . -name "*.pyx" -exec ./tools/rm_pyx_c_file.sh {} \;
+	rm -f MANIFEST
```
soupault: Where do you get this file from?

stefanv: When you do source builds, a `MANIFEST` file is generated from `MANIFEST.in`.

soupault: It is actually created during `python setup.py sdist` run.

tools/rm_pyx_c_file.sh
```diff
@@ -4,3 +4,8 @@ filename="${1%.*}".c
 if [ -e "$filename" ]; then
 	rm -f "$filename";
 fi
+
+filename="${1%.*}".cpp
```
soupault: I'd rename the script then.

Makefile
stefanv: How is that not a source build? :)

tools/rm_pyx_c_file.sh
stefanv: Sure, let's do that in another PR, if you want.

Makefile
soupault: For me, `python setup.py build_ext` or `pip install -e .` are source builds. `sdist` is `sdist`. :) Anyways...

PR #3658, opened 2019-01-17
Refine release process notes

stefanv:


RELEASE.txt
```diff
@@ -32,29 +32,55 @@ the corresponding release branch.
...
+      ``make clean; make html; make gh-pages``.
+    - In the ``gh-pages/`` directory:
+      - Update the symlink to ``stable`` and commit.
+      - Upload the docs: ``git push origin gh-pages``.
+
+At this point, you are ready to tag the new version.  This "seals the
+deal", so we want to make sure everything is working.  Go to
+https://travis-ci.org/scikit-image/scikit-image/branches, click on the
+release branch, and ensure that all checks have passed for the latest
+build.  Do the same for AppVeyor at
```
jni: @stefanv this is necessary, but not sufficient to check that the build will work on `scikit-image-wheels`. Is there a way to check *those* builds on the latest release tip *without* making a release?

RELEASE.txt
```diff
@@ -98,14 +117,12 @@ the corresponding release branch.
   - Open that URL and check for the wheels as
     ``scikit_image-<versiontag>-<platforminfo>.whl``
 
-- Update the docs:
+- Make sure you have all the required write permissions (if not, you will need
+  to ask Stéfan or Juan to grant you the access), specifically, to:
```
jni: Possibly this should actually be moved to the very top? It is frustrating to get to this point and then be stuck, especially if on a different timezone.

stefanv:
@jni Updated with your feedback, and mentioned that binary wheels should be built once on dev branch before doing the final build.

RELEASE.txt
```diff
@@ -101,14 +113,24 @@ the corresponding release branch.
...
+   ``github.com:scikit-image/scikit-image`` repository.)
 
-  - Edit ``doc/source/_static/docversions.js`` and commit
-  - Build a clean version of the docs. In the root directory, run
-    ``python setup.py install``, then ``make cleandoc; make html``.
-  - In the ``doc/`` directory, build using ``make gh-pages``.
-  - Update the symlink to ``stable``.
-  - Push upstream: ``git push origin gh-pages`` in ``doc/gh-pages``.
+- Build the package wheels again, but change BUILD_COMMIT in both
+  Travis-CI and AppVeyor to the release version, e.g., ``v0.16.0``.
```
hmaarrfk: some notes should be added below ensuring that you are using a clean up to date environment for sdist generation. sdists ship generated c code and should be done with the latest cython version to support the most python versions. (old versions of cython will not support say 3.8)

stefanv: We should probably stop shipping generated C code, it's easy enough to install Cython nowadays.

stefanv: (We had a great mechanism before, which was to generate .md5 hash files upon Cythonization.  We would *not* ship these files, so whenever a user who had Cython installed tried to compile skimage, Cython would be called.  Otherwise, it would fall back on the pre-generated files.  When we switched to using Cython's internal mechanism for determining when to recompile, we lost that advantage.)

hmaarrfk: Right. I can make that PR.

stefanv: Thanks!

PR #3661, opened 2019-01-17
Contribution script: allow specification of GitHub development branch

stefanv:
Also some cleanups, such as using f-strings, and more clearly describing
how to generate the GH_TOKEN.

### DIFF
--- a/doc/release/contribs.py
+++ b/doc/release/contribs.py
@@ -69,11 +69,10 @@ def get_user(login):
     return user
 
 
-def get_merged_pulls(user, repo, date, page=1, results=0):
+def get_merged_pulls(user, repo, date, page=1, results=0, branch='master'):
     # See https://developer.github.com/v3/search/#search-issues
     # See https://help.github.com/articles/understanding-the-search-syntax/#query-for-dates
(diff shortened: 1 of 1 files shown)


doc/release/contribs.py
```diff
@@ -69,11 +69,10 @@ def get_user(login):
     return user
 
 
-def get_merged_pulls(user, repo, date, page=1, results=0):
+def get_merged_pulls(user, repo, date, page=1, results=0, branch='master'):
     # See https://developer.github.com/v3/search/#search-issues
     # See https://help.github.com/articles/understanding-the-search-syntax/#query-for-dates
-    query = 'repo:{user}/{repo} merged:>{date} sort:created-asc'
-    query = query.format(user=user, repo=repo, date=date)
+    query = f'repo:{user}/{repo} merged:>{date} sort:created-asc base:{branch}'
```
hmaarrfk: This kind of stuff doesn't work for python 3.5. it means you need 3.6 to run. not really LTS friendly.....

stefanv: Only developers run these scripts, so I don't think it matters.

PR #3663, opened 2019-01-18
WIP Skip doctest to see if it's responsible for segfault

jni:
Just experimenting with Travis, don't merge!


jni:
TIL that `#doctest +SKIP` just stops the output from being checked, it doesn't prevent the function call. :joy:

hmaarrfk:
why don't you test with verbose?
Then test only that one file?

hmaarrfk:
Eventually, you can try using the same xcode version as conda-forge, but I'm pretty sure conda-forge is broken today;) https://github.com/conda-forge/scikit-image-feedstock/pull/40

jni:
@hmaarrfk 

> why don't you test with verbose?
> Then test only that one file?

Because I am in a frustration-and-tiredness-induced brain coma. =P Very good suggestions, gonna go with that now. =)

jni:
@hmaarrfk Are. You. &$^#ing. Kidding me.

https://travis-ci.org/scikit-image/scikit-image/builds/481212183#L4745-L4765

stefanv:
I presume this is an old experiment, so closing for now.

PR #3675, opened 2019-01-23
Pass parameters through tifffile plugin

### LINKED ISSUE #3674: "pages" parameter is not passed to tifffile's imread
## Description
When trying to load a tif file with the additional "pages" parameter I get an error message:

```

  File "<ipython-input-18-4a8d1197932c>", line 1, in <module>
    io.imread('dataRaw_0002_ang090.tif',plugin='tifffile',pages=[3])

  File "C:\ProgramData\Anaconda3\lib\site-packages\skimage\io\_io.py", line 61, in imread
    img = call_plugin('imread', fname, plugin=plugin, **plugin_args)

  File "C:\ProgramData\Anaconda3\lib\site-packages\skimage\io\manage_plugins.py", line 211, in call_plugin
    return func(*args, **kwargs)

  File "C:\ProgramData\Anaconda3\lib\site-packages\skimage\io\_plugins\tifffile_plugin.py", line 31, in imread
    return tif.asarray(**kwargs)

TypeError: asarray() got an unexpected keyword argument 'pages'
```
According to 
http://scikit-image.org/docs/0.13.x/api/skimage.io.html#skimage.io.imread
it should be possible to pass plugin args, however when you look into 

> skimage\io\_plugins\tifffile_plugin.py line 30
(this also persists in the newest version though a bit changed code: https://github.com/scikit-image/scikit-image/blob/master/skimage/io/_plugins/tifffile_plugin.py)

you see that no arguments are passed to the Tifffile class, and instead all arguments are passed to the asarray function one line after.

## Way to reproduce
- Code example: `io.imread('sample.tif',plugin='tifffile',pages=[3])`
- Operating system and version: Windows 10 64bit, 
- Python 3.7
- skimage 0.13.1

MarcelBeining:
## Description
This fixes the bug that, contrary to the documentation, arguments such as 'pages' were not passed to the TiffFile reader.

## Checklist
  existing benchmark
- [x] Unit tests

## References
closes #3674 

## For reviewers

  later.

### DIFF
--- a/skimage/io/_plugins/tifffile_plugin.py
+++ b/skimage/io/_plugins/tifffile_plugin.py
@@ -1,6 +1,6 @@
 from ...external.tifffile import TiffFile, imsave, parse_kwargs
 
-                 
+
 def imread(fname, **kwargs):
     """Load a tiff image from file.
 
@@ -24,17 +24,17 @@ def imread(fname, **kwargs):
     .. [1] http://www.lfd.uci.edu/~gohlke/code/tifffile.py
 
     """
-    
+
     if 'img_num' in kwargs:
         kwargs['key'] = kwargs.pop('img_num')
-        
+
     # separate TiffFile kwargs from other kwargs
     tiff_keys = ['multifile', 'multifile_close', 'pages', 'fastij', 'is_ome',
                  'pattern']
     kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
     for key in tiff_keys:
         kwargs.pop(key, None)
-    
+
     # read and return tiff as numpy array
     with TiffFile(fname, **kwargs_tiff) as tif:
         return tif.asarray(**kwargs)
--- a/skimage/io/_plugins/tifffile_plugin.py
+++ b/skimage/io/_plugins/tifffile_plugin.py
@@ -28,11 +28,10 @@ def imread(fname, dtype=None, **kwargs):
     if 'img_num' in kwargs:
         kwargs['key'] = kwargs.pop('img_num')
 
-    # separate TiffFile kwargs from other kwargs
+    # parse_kwargs will extract keyword arguments intended for the TiffFile 
+    # class and remove them from the kwargs dictionary in-place
     tiff_keys = ['multifile', 'multifile_close', 'pages', 'fastij', 'is_ome']
     kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
-    for key in tiff_keys:
-        kwargs.pop(key, None)
 
     # read and return tiff as numpy array
     with TiffFile(fname, **kwargs_tiff) as tif:
--- a/skimage/io/_plugins/tifffile_plugin.py
+++ b/skimage/io/_plugins/tifffile_plugin.py
@@ -24,6 +24,7 @@ def imread(fname, **kwargs):
     .. [1] http://www.lfd.uci.edu/~gohlke/code/tifffile.py
 
     """
+    
     if 'img_num' in kwargs:
         kwargs['key'] = kwargs.pop('img_num')
         
@@ -32,7 +33,8 @@ def imread(fname, **kwargs):
                  'pattern']
     kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
     for key in tiff_keys:
-        kwargs.pop(key,None)
-
+        kwargs.pop(key, None)
+    
+    # read and return tiff as numpy array
     with TiffFile(fname, **kwargs_tiff) as tif:
         return tif.asarray(**kwargs)
--- a/skimage/io/_plugins/tifffile_plugin.py
+++ b/skimage/io/_plugins/tifffile_plugin.py
@@ -1,7 +1,7 @@
-from ...external.tifffile import TiffFile, imsave
+from ...external.tifffile import TiffFile, imsave, parse_kwargs
 
-
-def imread(fname, dtype=None, **kwargs):
+                 
+def imread(fname, **kwargs):
     """Load a tiff image from file.
 
     Parameters
@@ -26,5 +26,13 @@ def imread(fname, dtype=None, **kwargs):
     """
     if 'img_num' in kwargs:
         kwargs['key'] = kwargs.pop('img_num')
-    with TiffFile(fname) as tif:
+        
+    # separate TiffFile kwargs from other kwargs
+    tiff_keys = ['multifile', 'multifile_close', 'pages', 'fastij', 'is_ome',
+                 'pattern']
+    kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
+    for key in tiff_keys:
+        kwargs.pop(key,None)
+
+    with TiffFile(fname, **kwargs_tiff) as tif:
         return tif.asarray(**kwargs)
--- a/skimage/io/_plugins/tifffile_plugin.py
+++ b/skimage/io/_plugins/tifffile_plugin.py
@@ -29,8 +29,7 @@ def imread(fname, dtype=None, **kwargs):
         kwargs['key'] = kwargs.pop('img_num')
 
     # separate TiffFile kwargs from other kwargs
-    tiff_keys = ['multifile', 'multifile_close', 'pages', 'fastij', 'is_ome',
-                 'pattern']
+    tiff_keys = ['multifile', 'multifile_close', 'pages', 'fastij', 'is_ome']
     kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
     for key in tiff_keys:
         kwargs.pop(key, None)
--- a/skimage/io/_plugins/tifffile_plugin.py
+++ b/skimage/io/_plugins/tifffile_plugin.py
@@ -1,7 +1,7 @@
 from ...external.tifffile import TiffFile, imsave, parse_kwargs
 
 
-def imread(fname, **kwargs):
+def imread(fname, dtype=None, **kwargs):
     """Load a tiff image from file.
 
     Parameters
--- a/skimage/io/tests/test_tifffile.py
+++ b/skimage/io/tests/test_tifffile.py
@@ -37,6 +37,11 @@ def test_imread_multipage_rgb_tif():
     img = imread(os.path.join(data_dir, 'multipage_rgb.tif'))
     assert img.shape == (2, 10, 10, 3), img.shape
 
+def test_tifffile_kwarg_passthrough ():
+    img = imread(os.path.join(data_dir, 'multipage.tif'), pages=[1], 
+                 multifile=False, multifile_close=True, fastij=True, 
+                 is_ome=True)
+    assert img.shape == (15, 10), img.shape
 
 def test_imread_handle():
     expected = np.load(os.path.join(data_dir, 'chessboard_GRAY_U8.npy'))
--- a/skimage/external/tifffile/__init__.py
+++ b/skimage/external/tifffile/__init__.py
@@ -1,5 +1,5 @@
-from .tifffile import imsave, imread, imshow, TiffFile, TiffWriter, TiffSequence
+from .tifffile import imsave, imread, imshow, TiffFile, TiffWriter, TiffSequence, parse_kwargs
 
 __version__ = '0.6.2'
 __all__ = ('imsave', 'imread', 'imshow', 'TiffFile', 'TiffWriter',
-           'TiffSequence')
+           'TiffSequence','parse_kwargs')
--- a/skimage/external/tifffile/__init__.py
+++ b/skimage/external/tifffile/__init__.py
@@ -1,5 +1,6 @@
-from .tifffile import imsave, imread, imshow, TiffFile, TiffWriter, TiffSequence, parse_kwargs
+from .tifffile import imsave, imread, imshow, TiffFile, TiffWriter, \
+    TiffSequence, parse_kwargs
 
 __version__ = '0.6.2'
 __all__ = ('imsave', 'imread', 'imshow', 'TiffFile', 'TiffWriter',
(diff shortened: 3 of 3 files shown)


skimage/io/_plugins/tifffile_plugin.py
```diff
@@ -1,7 +1,7 @@
-from ...external.tifffile import TiffFile, imsave
+from ...external.tifffile import TiffFile, imsave, parse_kwargs
```
jni: According to our tests, this `parse_kwargs` function doesn't exist...?

hmaarrfk: It isn't exported from `tiffille.__init__` https://github.com/scikit-image/scikit-image/blob/master/skimage/external/tifffile/__init__.py

You need to import it from `external.tifffile.tifffile`

stefanv:
Also, we can double check, but `skimage.io.imread` should absorb all imsave keyword arguments other than the tifffile specific ones, so once you reach this point I think you can write the function without the special keyword handling?

MarcelBeining:
> Also, we can double check, but `skimage.io.imread` should absorb all imsave keyword arguments other than the tifffile specific ones, so once you reach this point I think you can write the function without the special keyword handling?

Hi stefanv. I am not completely sure what you mean. Guess you meant the imread keyword arguments? the reason I did not forward all TiffFile arguments was to be conform with the skimage docs about what is fowarded. Also the kwargs have to be split because other keywords are passed to the .asarray method (`tifffile_plugin.py line 40`). I could use something like `inspect.signature(TiffFile)` to get all keywords an

MarcelBeining:
any reason left why not allow merging? :-)

hmaarrfk:
@MarcelBeining it seems that your mods are causing the tests to fail. I think you should keep the dtype parameter there, even though it isn't used.

Inertia is too powerful to stop :D.

MarcelBeining:
@hmaarrfk yes that did it, thanks! how did you know, or do you have any idea how to find out why the check has failed? i am new to testing...

jni:
@MarcelBeining thanks for contributing! Here's how you can check the status of the automatic tests. First, you may need to click on "show all checks":

screen shot 2019-02-02 at 12 33 15 pm

Then you can click on a specific check, in this case, Travis, for example:

screen shot 2019-02-02 at 12 33 27 pm

Finally, you can click on specific builds within that system:

screen shot 2019-02-02 at 12 37 16 pm

This will take you to a build log that shows all of the tests that were run, and any warnings and errors that were raised.

stefanv:
So, which other arguments get passed in through `kwargs` that introduce the need to have it filtered?  Just clarifying here for my own understanding.

MarcelBeining:
@hmaarrfk : I understand the concept of testing, but unfortunately, I never wrote testing functions in python so far and thus have not the slightest idea how that should look like, even after checking `skimage/io/tests` folder :-( If you could point me to some commit, where an understandable change + test was made, maybe I'll know what to do ^^

@stefanv : I refer to these docs:
http://scikit-image.org/docs/dev/api/skimage.external.tifffile.html#skimage.external.tifffile.imread

There it says 
> kwargs : dict
> Parameters ‘multifile’, ‘multifile_close’, ‘pages’, ‘fastij’, and ‘is_ome’ are passed to the TiffFile class. The ‘pattern’ parameter is passed to the TiffSequence class. Other parameters are passed to the asarray functions. The first image series is returned if no arguments are provided.

which is why I think that kwargs splitting is necessary to stick with the docs.

hmaarrfk:
Have a look at this function
https://github.com/scikit-image/scikit-image/blob/master/skimage/io/tests/test_tifffile.py#L41

Write an other one, maybe `test_tifffile_kwarg_passthrough` where an assert statement depends on the results of a kwarg you pass to tifffile. 

You didn't provide a minimal code example in https://github.com/scikit-image/scikit-image/issues/3674. If you would have, the tests would be as easy as adding an assert statement at the end of the ~10 lines of code you used to create the issue.

Maybe you can go back in your notes and pull that code out?

Finally, if you need to add any (small) test images, feel free to a dd them in the same test directory.

MarcelBeining:
OK, I am done with everything, ready for merging if you are :-)

skimage/io/_plugins/tifffile_plugin.py
```diff
@@ -24,7 +24,16 @@ def imread(fname, dtype=None, **kwargs):
...
     """
+
     if 'img_num' in kwargs:
         kwargs['key'] = kwargs.pop('img_num')
-    with TiffFile(fname) as tif:
+
+    # separate TiffFile kwargs from other kwargs
+    tiff_keys = ['multifile', 'multifile_close', 'pages', 'fastij', 'is_ome']
+    kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
+    for key in tiff_keys:
```
hmaarrfk: Why is this necessary?

I'm inclined to omit the need to call `parse_kwargs` unless we want to pass other kwargs to a different function. 

My reasoning is that 
1. It adds unecessary overhead.
2. We don't actually test this functionality in the test suite. (i.e. the test suite as is will likely pass if we remove these 3 lines)
3. By parsing the kwargs, we don't issue any error messages if a user provides a faulty kwargs.

Thoughts?

cgohlke: > Why is this necessary?
> I'm inclined to omit the need to call `parse_kwargs` unless we want to pass other kwargs to a different function.

Exactly, it's to separate the arguments passed to `TiffFile(fname, **kwargs_tiff)` and `tif.asarray(**kwargs)`.

`parse_kwargs` should remove the keys from dict, so I'm not sure why `for key in tiff_keys: kwargs.pop(key, None)` is necessary.

jni: @hmaarrfk @cgohlke I feel like you are saying different things? @hmaarrfk are you saying that we can just pass all kwargs unmodified to `TiffFile` and then to `tif.asarray`? That seems like poor practice to me.

However, if `parse_kwargs` modifies the dictionary in place as @cgohlke suggests, then, sure, let's remove the for-loop, but let's add a comment above the `parse_kwargs` line that records this, as in-place modifications are not obvious on reading the code.

skimage/io/_plugins/tifffile_plugin.py
```diff
@@ -24,7 +24,16 @@ def imread(fname, dtype=None, **kwargs):
...
 
     """
+
     if 'img_num' in kwargs:
         kwargs['key'] = kwargs.pop('img_num')
-    with TiffFile(fname) as tif:
+
+    # separate TiffFile kwargs from other kwargs
+    tiff_keys = ['multifile', 'multifile_close', 'pages', 'fastij', 'is_ome']
+    kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
```
jni:
```suggestion
    # parse_kwargs will extract keyword arguments intended for the TiffFile class
    # and remove them from the tiff_keys dictionary in-place
    kwargs_tiff = parse_kwargs(kwargs, *tiff_keys)
```

skimage/io/_plugins/tifffile_plugin.py
jni:
```suggestion
```

skimage/io/_plugins/tifffile_plugin.py
```diff
@@ -24,7 +24,16 @@ def imread(fname, dtype=None, **kwargs):
...
+    for key in tiff_keys:
+        kwargs.pop(key, None)
```
jni:
```suggestion
```

skimage/io/_plugins/tifffile_plugin.py
hmaarrfk: Oh i get it. From the docstring

```
Return dict with keys from keys|keyvals and values from kwargs|keyvals.

    Existing keys are deleted from kwargs.
```

@MarcelBeining please remove the `kwargs.pop` construct

hmaarrfk:
Yeah @jni's suggested changes would address my concerns.

skimage/io/_plugins/tifffile_plugin.py
cgohlke: > just pass all kwargs unmodified to `TiffFile` and then to `tif.asarray`

I think that will fail for non-empty `kwargs`.

skimage/io/_plugins/tifffile_plugin.py
cgohlke: > and remove them from the tiff_keys dictionary in-place

and remove them from the **kwargs** dictionary in-place

jni: Oops, thanks @cgohlke! Fixed!

skimage/io/_plugins/tifffile_plugin.py
cgohlke: > if `parse_kwargs` modifies the dictionary in place

I know, a little un-intuitive, but it's a useful helper function in my packages...

```
>>> help(parse_kwargs)
Help on function parse_kwargs in module skimage.external.tifffile.tifffile:

parse_kwargs(kwargs, *keys, **keyvalues)
    Return dict with keys from keys|keyvals and values from kwargs|keyvals.

    Existing keys are deleted from kwargs.

    >>> kwargs = {'one': 1, 'two': 2, 'four': 4}
    >>> kwargs2 = parse_kwargs(kwargs, 'two', 'three', four=None, five=5)
    >>> kwargs == {'one': 1}
    True
    >>> kwargs2 == {'two': 2, 'four': 4, 'five': 5}
    True
```

hmaarrfk: Yeah,  I've seen a few people create these helper functions that would split dictionaries into two. I think it is fine to use it here as it part of the tifffile plugin.

skimage/io/tests/test_tifffile.py
```diff
@@ -37,6 +37,11 @@ def test_imread_multipage_rgb_tif():
     img = imread(os.path.join(data_dir, 'multipage_rgb.tif'))
     assert img.shape == (2, 10, 10, 3), img.shape
 
+def test_tifffile_kwarg_passthrough ():
+    img = imread(os.path.join(data_dir, 'multipage.tif'), pages=[1], 
+                 multifile=False, multifile_close=True, fastij=True,
```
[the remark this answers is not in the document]
jni: Crud. LOL Is there any replacement for them or recommended way of handling them?

At any rate, the plan is to merge this first, so once that happens, I'll update #3650 to deal with this correctly (I hope with your help, @cgohlke!).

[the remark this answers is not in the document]
jni: (so, please continue discussion of this issue there...!)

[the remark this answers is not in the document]
cgohlke: > Is there any replacement for them

Those were optimizations for very specific use cases (TIFF files with tens/hundreds of thousands of IFDs; multi-file OME-TIFF which exceeded the number of allowed open file handles; 32 bit TIFF files > 4 GB) and were made obsolete by a new implementation.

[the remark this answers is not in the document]
hmaarrfk: but would it be wrong to parse them? does `asarray` need them?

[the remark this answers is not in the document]
jni: I think the point is that you'll get a TypeError if you pass in a non-existing kwarg...?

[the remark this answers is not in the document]
hmaarrfk: so I feel like this would be "safe" in both versions right? maybe the function should be reversed, and we parse away the kwargs to `asarray`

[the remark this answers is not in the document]
hmaarrfk: But I second @jni's comments, we should work with @MarcelBeining to get this PR merged in only considering the bundled plugin compatibility. We can worry about bleeding edge compatibility in the other PR.

cgohlke:
I don't quite understand why `tifffile_plugin.imread` partially re-implements `tifffile.imread`? This should work and leaves the details of dispatching `kwargs` to the `tifffile` library:

```Python
def imread(fname, dtype=None, **kwargs):
    """ ... """
    if 'img_num' in kwargs:
        kwargs['key'] = kwargs.pop('img_num')
    tifffile.imread(fname, **kwargs)
```

jni:
I don't know either, but I again suggest leaving that for a later PR. =)

MarcelBeining:
Lol guys, I totally lost overview of which of your suggested changes I should commit now and which not (as I never used this GitHub feature before). All of them? And if yes, is the order important?

hmaarrfk:
Yeah. All of them ;)

Or you could simply remove the pop constructs as they are done for you by parse_kwargs

MarcelBeining:
> Yeah. All of them ;)
> 
> Or you could simply remove the pop constructs as they are done for you by parse_kwargs

Done :-)

hmaarrfk:
Ok, this is great! Thanks for your help.

@MarcelBeining do you use python 2.7?

@jni do we want to backport this anyway?

MarcelBeining:
@hmaarrfk No, Python 3.6, why?
Nice we finally did it :-)

jni:
@hmaarrfk my position on backports is *critical* bug fixes, and anything else "by popular demand". By critical I mean "your code will be incorrect without this fix", as with the error I made computing inertia tensor in 3D+. So, I suggest we don't bother with this one, unless someone explicitly asks for it.

@MarcelBeining thank you very much for contributing to scikit-image! And for your exemplary patience. =)

hmaarrfk:
I would have backportrd if you did. But I feel like we want to focus on finally releasing 0.15. So since you don’t care about python 2.7 I’ll just let somebody else open the issue asking for it.

Issue #3680, opened 2019-01-25
non-isotropic blob finding


ambrosejcarr:
## Non-isotropic blob finding (in 3d)

I work with RNA-FISH data that is captured by images that are isotropic in x, y, but _samples_ the z plane. The workflow involves blob finding to identify spots in the image, but the non-isotropic nature of the data require a smaller sigma in z. Concretely, spots may be 5-9 pixels wide in (x, y) but only 2-3 pixels in z. As a result, implemented skimage spot finders that support 3d (`blob_log, blob_dog`) have difficulty finding spots in my data. 

Would it be feasible to provide support for non-isotropic spot finding in `skimage`? At the moment we are using the `locate` function in `trackpy`, which implements support for this use case: 

http://soft-matter.github.io/trackpy/dev/generated/trackpy.locate.html

jni:
Hi @ambrosejcarr! Thanks for the suggestion! I agree that the non-isotropic use case is important. It looks like this is tricky with the API, which effectively mirrors the arguments to `np.linspace` for `sigma`. I see two options for supporting this:

- allow vector-like `min_sigma` etc, broadcasting them together. So you would type `min_sigma=(2, 5, 5), max_sigma=(3, 9, 9), num_sigma=3`.
- allow a new argument `sigmas=` that provides directly a 2D array of size `(num_sigmas, ndim)`.

Both of these are ok, and maybe both are worth supporting. Do you have any thoughts about this?

After we settle the API question, from what I can tell the implementation will be easy.

ambrosejcarr:
@jni Thanks for the quick response. I want to check my understanding -- the advantage of the latter case of directly providing the array is that a user could provide non-linear sigmas (e.g. log-space sigmas)? 

Our workflow won't require this, and the former case of vector-like sigmas + `num_sigma` seems simpler for a user, so I'd vote (weakly) in favor of the vector-sigmas case unless you can think of a use case requiring the latter. Both implementations would solve my issue, however, so I'd be fine with either choice.

ambrosejcarr:
I investigated a bit more and I think allowing users to pass an array of sigmas could be potentially dangerous. I'm not an image-processing expert, but it looks like `blob_dog` expects sigmas to be linearly increasing over `(min_sigma, max_sigma)`. Allowing users to pass `sigmas=` could enable users to break the expectations of the algorithm. 

Some other questions arise for `blob_dog`:

[...]

PR #3690, opened 2019-01-30
Add support for anisotropic blob detection in blob_log and blob_dog

ambrosejcarr:
## Add support for anisotropic blob detection in blob_log and blob_dog

This PR enables a user to submit an anisotropic gaussian kernel to blob_log or blob_dog. 

During testing on real data, I noticed that the exclude_border functionality of peak_local_max is hard coded to "off", however in my data I see significant border effects within ~ 1 sigma of the image border. I've added that parameter to the `blob_log` and `blob_dog` methods with the default set to the existing functionality (no exclusion). 

I don't believe there are existing benchmarks for spot finding, but changing the kernel width should not affect the performance in any way, as gaussian convolution is implemented as a series of linear convolutions, and there is no performance gain from subsequent kernels having the same width.

## Checklist
  existing benchmark
- [x] Unit tests

## References
closes issue #3680 

## For reviewers

  later.

### DIFF
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -259,7 +259,7 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
     if isinstance(max_sigma, (int, float)):
         max_sigma = np.full(len(image.shape), max_sigma, dtype=np.float)
     if isinstance(min_sigma, (int, float)):
-        min_sigma = np.full(len(image.shape), min_sigma, dtype=np.float)
+        min_sigma = np.full(image.ndim, min_sigma, dtype=float)
 
     # Convert sequence types to array
     min_sigma = np.asarray(min_sigma, dtype=np.float)
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -257,7 +257,7 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
     # Gaussian filter requires that sequence-type sigmas have same
     # dimensionality as image. This broadcasts scalar kernels
     if isinstance(max_sigma, (int, float)):
-        max_sigma = np.full(len(image.shape), max_sigma, dtype=np.float)
+        max_sigma = np.full(image.ndim, max_sigma, dtype=float)
     if isinstance(min_sigma, (int, float)):
         min_sigma = np.full(image.ndim, min_sigma, dtype=float)
 
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -309,7 +309,7 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
 
 
 def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
-             overlap=.5, log_scale=False, exclude_border=False):
+             overlap=.5, log_scale=False, *, exclude_border=False):
     r"""Finds blobs in the given grayscale image.
 
     Blobs are found using the Laplacian of Gaussian (LoG) method [1]_.
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -166,7 +166,7 @@ def _prune_blobs(blobs_array, overlap):
 
 
 def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
-             overlap=.5,):
+             overlap=.5, exclude_border=False):
     r"""Finds blobs in the given grayscale image.
 
     Blobs are found using the Difference of Gaussian (DoG) method [1]_.
@@ -178,12 +178,16 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
     image : 2D or 3D ndarray
         Input grayscale image, blobs are assumed to be light on dark
         background (white on black).
-    min_sigma : float, optional
-        The minimum standard deviation for Gaussian Kernel. Keep this low to
-        detect smaller blobs.
-    max_sigma : float, optional
-        The maximum standard deviation for Gaussian Kernel. Keep this high to
-        detect larger blobs.
+    min_sigma : scalar or sequence of scalars, optional
+        the minimum standard deviation for Gaussian kernel. Keep this low to
+        detect smaller blobs. The standard deviations of the Gaussian filter
+        are given for each axis as a sequence, or as a single number, in
+        which case it is equal for all axes.
+    max_sigma : scalar or sequence of scalars, optional
+        The maximum standard deviation for Gaussian kernel. Keep this high to
+        detect larger blobs. The standard deviations of the Gaussian filter
+        are given for each axis as a sequence, or as a single number, in
+        which case it is equal for all axes.
     sigma_ratio : float, optional
         The ratio between the standard deviation of Gaussian Kernels used for
         computing the Difference of Gaussians
@@ -194,15 +198,21 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
     overlap : float, optional
         A value between 0 and 1. If the area of two blobs overlaps by a
         fraction greater than `threshold`, the smaller blob is eliminated.
+    exclude_border : int or bool, optional
+        If nonzero int, `exclude_border` excludes blobs from
+        within `exclude_border`-pixels of the border of the image.
 
     Returns
     -------
-    A : (n, image.ndim + 1) ndarray
-        A 2d array with each row representing 3 values for a 2D image,
-        and 4 values for a 3D image: ``(r, c, sigma)`` or ``(p, r, c, sigma)``
-        where ``(r, c)`` or ``(p, r, c)`` are coordinates of the blob and
-        ``sigma`` is the standard deviation of the Gaussian kernel which
-        detected the blob.
+    A : (n, image.ndim + sigma) ndarray
+        A 2d array with each row representing 2 coordinate values for a 2D
+        image, and 3 coordinate values for a 3D image, plus the sigma(s) used.
+        When a single sigma is passed, outputs are:
+        ``(r, c, sigma)`` or ``(p, r, c, sigma)`` where ``(r, c)`` or
+        ``(p, r, c)`` are coordinates of the blob and ``sigma`` is the standard
+        deviation of the Gaussian kernel which detected the blob. When an
+        anisotropic gaussian is used (sigmas per dimension), the detected sigma
+        is returned for each dimension.
 
     References
     ----------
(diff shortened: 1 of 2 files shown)


ambrosejcarr:
I'm happy to fix the above pep8 issues, however it leads to (in my opinion) less readable code. I will wait for instructions, as I see other places in the codebase where the standard is not enforced.

stefanv:
It's not a big deal, don't worry about it.  We've debated making the limit higher, but the 80 number comes a long way.

skimage/feature/blob.py
```diff
@@ -178,12 +178,16 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
...
         Input grayscale image, blobs are assumed to be light on dark
         background (white on black).
-    min_sigma : float, optional
-        The minimum standard deviation for Gaussian Kernel. Keep this low to
-        detect smaller blobs.
-    max_sigma : float, optional
-        The maximum standard deviation for Gaussian Kernel. Keep this high to
-        detect larger blobs.
+    min_sigma : scalar or sequence of scalars, optional
+        the minimum standard deviation for Gaussian kernel. Keep this low to
```
jni:
```suggestion
        The minimum standard deviation for Gaussian kernel. Keep this low to
```

skimage/feature/blob.py
```diff
@@ -194,15 +198,21 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
...
-        detected the blob.
+    A : (n, image.ndim + sigma) ndarray
+        A 2d array with each row representing 2 coordinate values for a 2D
+        image, and 3 coordinate values for a 3D image, plus the sigma(s) used.
+        When a single sigma is passed, outputs are:
+        ``(r, c, sigma)`` or ``(p, r, c, sigma)`` where ``(r, c)`` or
+        ``(p, r, c)`` are coordinates of the blob and ``sigma`` is the standard
+        deviation of the Gaussian kernel which detected the blob. When an
+        anisotropic gaussian is used (sigmas per dimension), the detected sigma
+        is returned for each dimension.
```
jni: 👍 !

skimage/feature/blob.py
```diff
@@ -244,8 +254,19 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
     """
     image = img_as_float(image)
 
+    # Gaussian filter requires that sequence-type sigmas have same
+    # dimensionality as image. This broadcasts scalar kernels
+    if isinstance(max_sigma, (int, float)):
+        max_sigma = np.full(len(image.shape), max_sigma, dtype=np.float)
+    if isinstance(min_sigma, (int, float)):
+        min_sigma = np.full(len(image.shape), min_sigma, dtype=np.float)
```
jni: Here, I suggest using `np.isscalar` instead of `isinstance`, because, for example, a NumPy int64 is not an instance of int (and one can get this type when, eg, grabbing a value from a NumPy array):

```python
In [7]: np.int_(4)
Out[7]: 4

In [8]: type(np.int_(4))
Out[8]: numpy.int64

In [9]: isinstance(np.int_(4), int)
Out[9]: False
```

(Weirdly, NumPy floats are instances of float.)

Additionally:

```suggestion
        min_sigma = np.full(image.ndim, min_sigma, dtype=float)
```

Little known factoid, `np.float` is just an alias for Python's `float`:

```
In [10]: np.float is float
Out[10]: True

In [11]: np.float_
Out[11]: numpy.float64
```

skimage/feature/blob.py
```diff
@@ -244,8 +254,19 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
     """
     image = img_as_float(image)
 
+    # Gaussian filter requires that sequence-type sigmas have same
+    # dimensionality as image. This broadcasts scalar kernels
+    if isinstance(max_sigma, (int, float)):
+        max_sigma = np.full(len(image.shape), max_sigma, dtype=np.float)
```
jni:
```suggestion
        max_sigma = np.full(image.ndim, max_sigma, dtype=float)
```

skimage/feature/blob.py
```diff
@@ -244,8 +254,19 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
...
+    if isinstance(min_sigma, (int, float)):
+        min_sigma = np.full(len(image.shape), min_sigma, dtype=np.float)
+
+    # Convert sequence types to array
+    min_sigma = np.asarray(min_sigma, dtype=np.float)
```
jni:
```suggestion
    min_sigma = np.asarray(min_sigma, dtype=float)
```

skimage/feature/blob.py
```diff
@@ -244,8 +254,19 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
...
+    max_sigma = np.asarray(max_sigma, dtype=np.float)
```
jni:
```suggestion
    max_sigma = np.asarray(max_sigma, dtype=float)
```

skimage/feature/blob.py
```diff
@@ -254,29 +275,41 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
...
-    lm[:, -1] = sigma_list[local_maxima[:, -1]]
+
+    # translate final column of lm, which contains the index of the
+    # sigma that produced the maximum intensity value, into the sigma
+    sigmas_of_peaks = sigma_list[local_maxima[:, -1]]
+
+    # if the gaussian is isotropic, the stdev across dimensions are
+    # identical, so return only the stdev deviation of the first dimension
+    if np.unique(min_sigma).shape == (1,) and np.unique(max_sigma).shape == (1,):
+        sigmas_of_peaks = sigmas_of_peaks[:, 0][:, None]
```
jni: Can you explain what is happening here? I don't understand the logic here and in the lines above. Also, I think, rather than use `np.unique`, maybe set a flag at the start (when we detect the inputs with `np.isscalar`) and use that here. I can, for example, imagine that someone might want to pass in an array of identical values for sigmas as a way to force the format to be consistent between isotropic and nonisotropic datasets.

skimage/feature/blob.py
```diff
@@ -353,30 +396,57 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
     """
     image = img_as_float(image)
 
+    # Gaussian filter requires that sequence-type sigmas have same
+    # dimensionality as image. This broadcasts scalar kernels
+    if isinstance(max_sigma, (int, float)):
+        max_sigma = np.full(len(image.shape), max_sigma, dtype=np.float)
+    if isinstance(min_sigma, (int, float)):
+        min_sigma = np.full(len(image.shape), min_sigma, dtype=np.float)
```
jni: As above.

skimage/feature/blob.py
```diff
@@ -308,15 +345,21 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
         If set intermediate values of standard deviations are interpolated
         using a logarithmic scale to the base `10`. If not, linear
         interpolation is used.
+    exclude_border : int or bool, optional
+        If nonzero int, `exclude_border` excludes blobs from
+        within `exclude_border`-pixels of the border of the image.
```
jni: If you want to add this in this PR, could you add a test for it?

skimage/feature/blob.py
```diff
@@ -254,29 +275,41 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
...
+
+    # Remove sigma index and replace with sigmas
+    lm = np.hstack([lm[:, :-1], sigmas_of_peaks])
+
     return _prune_blobs(lm, overlap)
 
 
 def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
-             overlap=.5, log_scale=False):
+             overlap=.5, log_scale=False, exclude_border=False):
```
jni:
```suggestion
             overlap=.5, log_scale=False, *, exclude_border=False):
```

We are using [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/) going forward as a way to get more flexibility when modifying the API in the future.

skimage/feature/blob.py
```diff
@@ -166,7 +166,7 @@ def _prune_blobs(blobs_array, overlap):
 
 
 def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
-             overlap=.5,):
+             overlap=.5, exclude_border=False):
```
jni:
```suggestion
             overlap=.5, *, exclude_border=False):
```

We are using [keyword-only arguments](https://www.python.org/dev/peps/pep-3102/) going forward as a way to get more flexibility when modifying the API in the future.

skimage/feature/blob.py
ambrosejcarr: Test added in bcee760

skimage/feature/blob.py
ambrosejcarr: Missed this one, pushed it up myself.

ambrosejcarr:
@jni Thanks for the thoughtful comments. There are some places in my codebases that will receive some `np.isscalar` substitutions. I think my confusion comes from the flexibility of the mypy type system, where `Union[int, float]` _does_ cover the set of numpy numeric types. 

I think I've made the requested changes and this is ready for re-review.

skimage/feature/blob.py
```diff
@@ -244,8 +254,24 @@ def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
     """
     image = img_as_float(image)
 
+    # if both min and max sigma are scalar, function returns only one sigma
+    scalar_sigma = (
+        True if np.isscalar(max_sigma) and np.isscalar(min_sigma) else False
+    )
```
jni: This could just be `scalar_sigma = np.isscalar(max_sigma) and np.isscalar(min_sigma)`. =)

skimage/feature/blob.py
jni: Oh, right! I momentarily forgot that `np.newaxis` is `None`. Which hints at a slightly preferred solution. ;)

I suggest here, though:

```suggestion
        # select one sigma column, keeping dimension
        sigmas_of_peaks = sigmas_of_peaks[:, 0:1]
```

skimage/feature/blob.py
```diff
@@ -353,30 +401,62 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
...
-    lm[:, -1] = sigma_list[local_maxima[:, -1]]
+
+    # translate final column of lm, which contains the index of the
+    # sigma that produced the maximum intensity value, into the sigma
+    sigmas_of_peaks = sigma_list[local_maxima[:, -1]]
+
+    # if the gaussian is isotropic, the stdev across dimensions are
+    # identical, so return only the stdev deviation of the first dimension
+    if scalar_sigma:
+        sigmas_of_peaks = sigmas_of_peaks[:, 0][:, None]
```
jni:
```suggestion
        # select one sigma column, keeping dimension
        sigmas_of_peaks = sigmas_of_peaks[:, 0:1]
```

ambrosejcarr: 🤦‍♂️ Good idea.

ambrosejcarr:
Comments addressed. Thanks!

skimage/feature/blob.py
ambrosejcarr: Fixed below!

jni:
Fantastic! @scikit-image/core this is ready to merge imho!

JDWarner:
This is great work @ambrosejcarr - always glad to see more anisotropic generalization in the library!

PR #3692, opened 2019-01-31
collections.abc nit

hmaarrfk:
calling `collections.abc` `coll` is kinda confusing.

## Checklist
  existing benchmark
- [ ] Unit tests

## References

## For reviewers

  later.


stefanv:
Note from docs:

> Checking isinstance(obj, Iterable) detects classes that are registered as Iterable or that have an __iter__() method, but it does not detect classes that iterate with the __getitem__() method. The only reliable way to determine whether an object is iterable is to call iter(obj).

But I think this fix is fine for now.

PR #3695, opened 2019-01-31
Numerous backports to 0.14.x to avoid warnings in tests and python 3.7

hmaarrfk:
## Description
This backport removes many nuisance warnings in the tests, and potentially for users when they use scikit-image.

This also includes preemptive backports of 
-  Enable the faulthandler module during testing #3708 
-  Fix 3.7 builds on travis #3709 

## References
xref:  #3684
xref:  #3687

## For reviewers
- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

PR #3696, opened 2019-02-01
Fix potential use of NULL pointer

cgohlke:
## Description

## Checklist
  existing benchmark
- [ ] Unit tests

## References

## For reviewers

  later.
- [x] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

### DIFF
--- a/skimage/filters/_ctmf.pyx
+++ b/skimage/filters/_ctmf.pyx
@@ -188,6 +188,8 @@ cdef Histograms *allocate_histograms(Py_ssize_t rows,
                    (sizeof(Histogram) + sizeof(PixelCount)) +
                    sizeof(Histograms) + 64)
     ptr = malloc(memory_size)
+    if ptr == NULL:
+        return NULL
     memset(ptr, 0, memory_size)
     # align ph.accumulator to 32-byte boundary
     roundoff = (<Py_ssize_t> ptr + 31) % 32


hmaarrfk:
honest question, does one have to test against malloc failling after every call?

jni:
@hmaarrfk as I understand it, you should check it before you touch that memory. The easiest place to do so is right after the malloc, because it's unlikely that you would malloc it and then maybe not use it.

@cgohlke does this supposedly fix the segfaults we've been seeing in our windows builds? Or did you discover this independently?

Anyway all our builds are failing on 0.14 except for 3.7, which is crazy. I don't know what's happened. I'm currently hosting @jakirkham and we are doing some dask-image development till next Thu, so no bandwidth to investigate until then.

cgohlke:
> honest question, does one have to test against malloc failling after every call?

I think one should. Otherwise one enters into undefined territory (depending on compiler, C runtime, OS) which usually ends with an unhandled access violation when using the NULL pointer. It is not uncommon for an imaging application to run low or out of memory such that malloc of even small amounts of memory can fail.

> does this supposedly fix the segfaults we've been seeing in our windows builds? Or did you discover this independently?

The latter. I got a fee license for the "PVS-Studio Analyzer", a static C code checker for Windows, and are slowly trying to fix the low hanging fruits in the scipy stack. The reports are quite noisy, especially for Cython generated C code and older C libraries.

I can try to reproduce the segfaults locally with a debug build and memory bounds checker...

jni:
@cgohlke That would be great, thanks! Here's a failing build as a starting point:

https://travis-ci.org/scikit-image/scikit-image/jobs/484620536#L7444

cgohlke:
> Here's a failing build as a starting point:

I could not reproduce any crash in feature/corner.py. 

There is however a crash during `graph/tests/test_connect.py::test_connections` at https://github.com/scikit-image/scikit-image/blob/master/skimage/graph/_mcp.pyx#L541 when trying to call a `cpdef`'ed method in the Cython extension type. Using `cdef` instead or overriding the function in Python fixes the crash. Could be a Cython 0.29.3 bug...

jni:
@cgohlke in my experiments, the segfault only occurs when the full test suite is run rather than when running `corners.py` docstrings alone.

cgohlke:
> the segfault only occurs when the full test suite is run

How do you run the tests? I am using
```
> py -m pytest skimage -v --pyargs
========================================================================= test session starts ==========================================================================
platform win32 -- Python 3.7.2, pytest-4.2.0, py-1.7.0, pluggy-0.8.1 -- X:\Python37\python.exe
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase(<snip>)
rootdir: <snip>\test, inifile:
plugins: xdist-1.26.1, pep8-1.0.6, forked-1.0.1, faulthandler-1.5.0, cov-2.6.1, palladium-1.2.1.1, hypothesis-3.86.7, celery-4.2.1
collected 5368 items
<snip>
```

jni:
@cgohlke I think you need to add `--doctest-modules`. Our tests on Travis run as:

```
export TEST_ARGS="--doctest-modules --cov=skimage"
pytest $TEST_ARGS skimage
```

Don't ask me why these two lines are in separate files... 🤦‍♂️

cgohlke:
> add `--doctest-modules`

Still no segfault here.

jni:
@cgohlke probably something about the amount of RAM available on Travis machines? Fun times.

cgohlke:
I think the segfault at _mcp.pyx#L54 is due to a Cython bug, where "it" tries to access the non-existing `__dict__` of the MCP Extension type. The CPython function to get the pointer of the dict is not quite correct either...

jni:
@cgohlke would adding a PR to v0.14.x adding faulthandler help us to narrow it down? master isn't creating the segfault because we are using the newer xcode there.

hmaarrfk:
I've pre-emptively added faulthandler to the v0.14.x PR I made https://github.com/scikit-image/scikit-image/pull/3695

hmaarrfk:

cgohlke:
> I think the segfault at _mcp.pyx#L54 is due to a Cython bug

This is fixed in https://github.com/cython/cython/commit/253c25a902aef2c54de31baa22af73d0dbad2a06 which will be released with Cython 0.29.5.

hmaarrfk:
Time to go bleeding edge!

jni:
> Time to go bleeding edge!

I agree. I think Cython is a dependency that we should move aggressively on, since it is only for build-time, not runtime, so most users don't even need it.

PR #3706, opened 2019-02-02
Add offset to regionprops

jni:
## Description

To calculate region properties in a bigger volume than can be held in one image, we need to be able to specify the coordinates of the input image as an input to the regionprops function. Amazingly, as far as I can tell, this only affects the "coordinates" and "centroid" properties.

## Checklist

- [ ] ~~Gallery example in `./doc/examples` (new features only)~~ (**Note:** this will help dask-image, less helpful here...)
  existing benchmark~~
- [x] Unit tests

## For reviewers

  later.


jni:
Note: this doesn't quite address [this SO question](https://stackoverflow.com/questions/53502161/manipulating-scikit-image-regionprops-results), but it gets us partway there.

stefanv:
Why not just slice the image before computing regionprops?

jni:
You slice it, then you can manually add in the slice start indices. Or you can slice it and add the appropriate `offset=` and get correctly-computed props in the global coordinate system.

jakirkham:
Does this seem reasonable @stefanv? @jni and I were interested in this because we were thinking how to handle a Dask-based `regionprops` implementation leveraging scikit-image under-the-hood.

skimage/measure/_regionprops.py
```diff
@@ -383,6 +383,10 @@ def regionprops(label_image, intensity_image=None, cache=True,
     coordinates : 'rc' or 'xy', optional
         Coordinate conventions for 2D images. (Only 'rc' coordinates are
         supported for 3D images.)
+    offset : array-like of int, shape `(label_image.ndim,)`, optional
+        Coordinates of the origin ("top-left" corner) of the label image.
+        Normally this is ([0, ]0, 0), but it might be different if one wants
+        to obtain regionprops of subvolumes within a much larger volume.
```
mkcor, line 389:
```suggestion
        to obtain regionprops of subvolumes within a larger volume.
```
?

mkcor (approved):
I find this very elegant. :heart_eyes: Thanks for this addition, @jni!

stefanv (approved):
Approved, pending integration of @mkcor's suggestion and fixing of the conflict.

jarrodmillman (approved):

jarrodmillman:
@jni This looks good to me (pending resolving the merge conflict and maybe rebasing on main). I am thinking about releasing 0.20.0rc0 tomorrow and wanted to check whether you thought we should merge this before or after 0.20. If before, feel free to make the updates and merge once the tests pass.

skimage/measure/_regionprops.py
```diff
@@ -303,6 +304,7 @@ def __init__(self, slice, label, label_image, intensity_image,
             multichannel = False
 
         self.label = label
+        self.offset = offset or np.zeros((label_image.ndim,), dtype=int)
```
jakirkham, line 307: Think this might be causing the test failures

```suggestion
        self.offset = (offset is None) or np.zeros((label_image.ndim,), dtype=int)
```

stefanv: Thanks, @jakirkham!

stefanv:
@jni I'm afraid we need some test updates too.

jni:
Sorry, I got distracted and wasn't able to investigate at the time. I've fixed the issue (with a force, had done the same thing and was easier than trying to do a bunch of pulling/merging/cherry-picking, sorry and thank you @jakirkham and @stefanv!), but there's a bit of an API design issue too, which is, self.slice in some use cases would itself have to include the offset. Just like we have `coords` and `coords_scaled`, we also might want `slice` and `slice_offset`. For example, if one did `regionprops` in a dask array, we would want to have the slice objects be valid not on chunks, but on the whole image.

However, there's several steps between this PR and dask_image.measure.regionprops. For example, we need to be able to combine features between chunks. So I don't think that this decision should hold up merge — this is already very useful to start experimenting with, I think.

stefanv:
Thanks @jni! Feel free to open an issue for anything you still want to address.

jakirkham:
Thanks all for shepherding this one in! 😄

jni:
👋 @jakirkham long time no see. 😊 let me know if you want to pair up on dask_image regionprops any time soon. 😃

PR #3710, opened 2019-02-03
Replace np.einsum with np.tensordot in _upsampled_dft

taylor-scott:
## Description

In feature/register_translation.py, _upsampled_dft uses np.einsum to compute the upsampled DFT of an image. This is not easily generalizable and in numpy 1.11 is (extremely) slow because no optimize flag is available.

np.einsum can be replaced with sequential calls to np.tensordot which (should) generalize to higher dimensions.

E.g.,

    import numpy as np
    image = np.random.uniform(10, size=(10, 15, 20))
    kernels = [np.random.uniform(10, size=(2*i, i)) for i in image.shape]
    einsum_result = np.einsum('ijk,li,mj,nk->lmn', image, *kernels, optimize=True)
    tensordot_result = np.copy(image)
    for kernel in kernels[::-1]:
        tensordot_result = np.tensordot(kernel, tensordot_result, axes=(1, -1))
    
    print(np.abs(tensordot_result - einsum_result).max()) # 0.0

It should also be faster in numpy 1.11 and, from testing on my latptop, not slower than np.einsum in numpy 1.16. It seems to be slightly faster for some inputs. 

Is there a standard way to benchmark functions? I see there's a benchmarks folder but I'm not sure how to use it.

## Todo
- [x] Benchmark in numpy 1.16 (current) ~and numpy 1.11~
- [x] Extend to nD images
- [x] Write tests for nD images
- [x] Clean up kernel statement

## Checklist

- [ ] ~Gallery example in `./doc/examples` (new features only)~
  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -63,13 +63,12 @@ def _upsampled_dft(data, upsampled_region_size,
     dim_properties = list(zip(data.shape, upsampled_region_size, axis_offsets))
 
     for (n_items, ups_size, ax_offset) in dim_properties[::-1]:
-        kernel = np.exp(np.dot(
-                (-im2pi / (n_items * upsample_factor)) *
-                (np.arange(upsampled_region_size[0])[:, None] - ax_offset),
-                (np.fft.ifftshift(np.arange(n_items))[None, :] - n_items // 2)))
+        kernel = np.fft.fftfreq(n_items, upsample_factor)[:, None] *\
+                 (np.arange(upsampled_region_size[0]) - ax_offset)
+        kernel = np.exp(-im2pi * kernel)
         # Equivalent to:
-        #   data[i, j, k] = (kernel[i, :] * data[j, k, :]).sum()
-        data = np.tensordot(kernel, data, axes=(1, -1))
+        #   data[i, j, k] = kernel[:, i] @ data[j, k].T
+        data = np.tensordot(kernel, data, axes=(0, -1))
     return data
 
 
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -27,7 +27,7 @@ def _upsampled_dft(data, upsampled_region_size,
 
     Parameters
     ----------
-    data : nD ndarray
+    data : array
         The input data array (DFT of original data) to upsample.
     upsampled_region_size : integer or tuple of integers, optional
         The size of the region to be sampled.  If one integer is provided, it
@@ -40,7 +40,7 @@ def _upsampled_dft(data, upsampled_region_size,
 
     Returns
     -------
-    output : nD ndarray
+    output : ndarray
             The upsampled DFT of the specified region.
     """
     # if people pass in an integer, expand it to a list of equal-sized sections
@@ -117,9 +117,9 @@ def register_translation(src_image, target_image, upsample_factor=1,
 
     Parameters
     ----------
-    src_image : ndarray
+    src_image : array
         Reference image.
-    target_image : ndarray
+    target_image : array
         Image to register.  Must be same dimensionality as ``src_image``.
     upsample_factor : int, optional
         Upsampling factor. Images will be registered to within
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -58,44 +58,19 @@ def _upsampled_dft(data, upsampled_region_size,
             raise ValueError("number of axis offsets must be equal to input "
                              "data's number of dimensions.")
 
-    # For the 2D case, it is more efficient to use the well-optimized
-    # dot product instead of einsum
-    if data.ndim == 2:
-        col_kernel = np.exp(
-            (-1j * 2 * np.pi / (data.shape[1] * upsample_factor)) *
-            (np.fft.ifftshift(np.arange(data.shape[1]))[:, None] -
-             np.floor(data.shape[1] / 2)).dot(
-                np.arange(upsampled_region_size[1])[None, :] - axis_offsets[1])
-        )
-        row_kernel = np.exp(
-            (-1j * 2 * np.pi / (data.shape[0] * upsample_factor)) *
-            (np.arange(upsampled_region_size[0])[:, None] -
-             axis_offsets[0]).dot(
-                np.fft.ifftshift(np.arange(data.shape[0]))[None, :] -
-                np.floor(data.shape[0] / 2))
-        )
+    im2pi = 1j * 2 * np.pi
 
-        return row_kernel.dot(data).dot(col_kernel)
+    dim_properties = list(zip(data.shape, upsampled_region_size, axis_offsets))
 
-    # For the general nD case, use sequential calls to tensordot to calculate
-    # the DFT
-    elif data.ndim > 2:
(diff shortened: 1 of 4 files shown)


jni:
@taylor-scott very nice! I'm not super concerned with supporting NumPy 1.11, which is getting a bit long in the tooth, but the code is significantly cleaner now so I think this is a worthwhile change either way!

Regarding benchmarks, I think it's a good idea to add a couple (one for 2D and one for 3D), since performance is part of the motivation for this PR. I recommend you look at https://github.com/scikit-image/scikit-image/blob/8135c2f3505d77134162846abf1b06c0b18d5a56/benchmarks/benchmark_morphology.py for how to parameterise tests, and then add a `class RegisterTranslation` to `benchmark_feature.py`. As long as the class has a `setup` method and a `time_register_translation` method, it'll get picked up by airspeed velocity for benchmarks.

jni:
By the way, if this work sounds too onerous, we can do it the old-fashioned way: run a benchmark script on your machine and share with us the script and the results. =)

hmaarrfk:
Honestly, I wouldn't worry about numpy 1.11 much. It is almost 3 years old now, so not so important.

I should write these out somewhere:

```
pip install asv
```
or 
```
conda install asv
```

Then to test your benchmark
```
asv run -E existing -b RankSuite
```

To compare your results to master
```
asv continuous -b RankSuite -E conda:3.7 master simplify_benchmarks 
```

you might have to swap the order the branch names.

skimage/feature/register_translation.py
```diff
@@ -27,7 +27,7 @@ def _upsampled_dft(data, upsampled_region_size,
 
     Parameters
     ----------
-    data : 2D or 3D ndarray
+    data : nD ndarray
```
stefanv: Just `ndarray`?

stefanv:
On Mon, 04 Feb 2019 21:24:22 -0800, Taylor D. Scott wrote:
> Thoughts on using tensordot in the 2D case? It's slower, but the code
> cleaner with no special cases.

I think 50 µsec is nothing to be concerned about at all, especially
given that the code is clear and expressive.

taylor-scott:
I think I'm nearly done with this... 

A couple more things I noticed:

1. `_upsampled_dft` calculates the DFT but we actually want the inverse DFT. That's way the function call is `_upsampled_df(image_product.conj(), ...).conj()`. 

I think a good way to handle this is to add an optional flag `inverse=False` to `_uspampled_dft`. If true, use positive `im2pi` when calculating the kernel (to give the inverse DFT). If false, use `-im2pi` as it is now. 

The `conj` calls can get expensive for large images (~10 msec for a 1040x1392) so I think it's good to get rid of them if possible. 

2. [L256](https://github.com/scikit-image/scikit-image/blob/a3d2557209fb5913609749f06e70b367cef00617/skimage/feature/register_translation.py#L256): This might be returning the right thing by coincidence but max() sorts lexicographically for complex numbers. Looking at the matlab code, it's supposed to be the maximum cross_correlation sorted by absolute value. I'll fix this one tonight.

stefanv:
On Tue, 05 Feb 2019 18:04:00 +0000, Taylor D. Scott wrote:
> 1. `_upsampled_dft` calculates the DFT but we actually want the inverse DFT. That's way the function call is `_upsampled_df(image_product.conj(), ...).conj()`. 
> 
> I think a good way to handle this is to add an optional flag
> `inverse=False` to `_uspampled_dft`. If true, use positive `im2pi`
> when calculating the kernel (to give the inverse DFT). If false, use
> `-im2pi` as it is now.

Could you expand a bit more on this item?

> 2. [L256](https://github.com/scikit-image/scikit-image/blob/a3d2557209fb5913609749f06e70b367cef00617/skimage/feature/register_translation.py#L256):
> This might be returning the right thing by coincidence but max() sorts
> lexicographically for complex numbers. Looking at the matlab code,
> it's supposed to be the maximum cross_correlation sorted by absolute
> value. I'll fix this one tonight.

Thanks!

taylor-scott:
The [MATLAB code ](https://www.mathworks.com/matlabcentral/fileexchange/18401-efficient-subpixel-image-registration-by-cross-correlation) is here:

    CCabs = abs(CC);
    [rloc, cloc] = find(CCabs == max(CCabs(:)),1,'first');
    CCmax = CC(rloc,cloc);

It takes the absolute value of the cross-correlation, finds the argmax, then return the cross-correlation at that point.

The code in register_translation, `CCmax = cross_correlation.max()` returns the maximum value based on how numpy sorts complex numbers, which is not by absolute value. Instead, it sorts by real part first, then by complex part.

So:

    np.array([1+1j, 0+10j]).max()

gives `1 + 1j` even though `abs(0+10j) > abs(1 + 1j)`

See the discussion here: numpy/numpy#8151

stefanv:
@taylor-scott Sorry if my response was confusing; I was asking about item #1.

taylor-scott:
@stefanv I'm sorry, I completely misread your question.

To find the cross-correlation, we take the inverse DFT of the image product. In pixel-level registration, this is just `cross_correlation = ifftn(image_product)` but for subpixel registration the function calls `_upsampled_dft` which embeds the image product into a large image and calculates the DFT of the larger image.

This almost gives the cross-correlation, but we actually need the inverse DFT of the larger image. So to calculate the cross-correlation, `register_translation` uses the [identity ](https://en.wikipedia.org/wiki/Discrete_Fourier_transform#Expressing_the_inverse_DFT_in_terms_of_the_DFT) `fft(image.conj).conj/N = ifft(image)` where `N` is the image size.

For large images, the time to call `conj` is not trivial. For example, calling `image.conj()` on one of my images (1040 x 1392) takes ~11 msec and registering two images takes ~230 msec total. 

To avoid this call to `conj`, I proposed adding an optional flag `inverse=False` to `_upsampled_dft`. When false, return the DFT as the function does now. When true, return the inverse DFT which is what we want in this case.

This is easy to accomplish while generating the kernel:
```python
    kernel = ((np.arange(upsampled_region_size[0]) - ax_offset)[:, None]
              * np.fft.fftfreq(n_items, upsample_factor))
    if inverse:
        # We can also add the normalizing factor here if desired
        kernel = np.exp(im2pi * kernel)
    else:
        kernel = np.exp(-im2pi * kernel)
```

taylor-scott:
Also, after thinking about it a bit more, this may be an issue better handled in a separate pull request after some more discussion. I think `_upsampled_dft` could be generally useful, but IMO this aspect and how the image is embedded in the upsampled image are non-intuitive.

stefanv:
@taylor-scott Thank you for that very clear explanation.  Yikes, such subtle tricks are very easy to miss, and I agree with you that making it explicit would be an improvement.  If you are willing to make the change in another PR, then great—let's get this one merged so long.  Feel free to add your name to CONTRIBUTORS.txt if you want.

@scikit-image/core Ready for final review.

taylor-scott:
Ok, great. I've removed the inverse flag from this PR and I'll open a separate PR to address that. Thanks @stefanv !

stefanv:
Thanks, @taylor-scott!

PR #3711, opened 2019-02-04
Simplify benchmarks pinnings

hmaarrfk:
## Description
Things shouldn't be pinned so hard in the benchmarks. Lets following new software!

cc: @TomAugspurger could this cause any troubles?

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -2,137 +2,27 @@
     // The version of the config file format.  Do not change, unless
     // you know what you are doing.
     "version": 1,
+    // For other options, see the documentation
+    // https://asv.readthedocs.io/en/stable/reference.html
 
-    // The name of the project being benchmarked
     "project": "scikit-image",
-
-    // The project's homepage
     "project_url": "https://scikit-image.org/",
-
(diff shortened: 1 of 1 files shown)


TomAugspurger:
I don't think it's cause any problems.

jni:
My concern is that it will make it harder to detect when a performance regression or improvement is caused by an upstream package change. Could we have one configuration pinned (to current package versions), and one cutting edge? That would make me happy. (Conversely, it'll allow us to *detect* when a regression is due to updates to packages.)

Having said all this, the two major "regressions" we've spotted have been hard/impossible to reproduce on local machines, so I don't know what to think anymore... =\

hmaarrfk:
Yeah I agree

hmaarrfk:
I would just unpin everything and let this "maintain itself". The burden of maintaining version numbers is a pain.

hmaarrfk:
@jni did you want to review this again? I mostly removed the giant comments that we never configured and pointed to the docs, (with the exception to the one thing we aren't supposed to touch).

PR #3717, opened 2019-02-04
Fix potential use of NULL pointers

cgohlke:
## Description

Always verify that malloc succeeded.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_marching_cubes_lewiner_cy.pyx
+++ b/skimage/measure/_marching_cubes_lewiner_cy.pyx
@@ -25,10 +25,7 @@ cimport numpy as np
 import cython
 
 # Enable low level memory management
-# from libc.stdlib cimport malloc, free
-cdef extern from "stdlib.h": # The cimport does not work on my Linux Laptop
-   void free(void* ptr)
-   void* malloc(size_t size)
+from libc.stdlib cimport malloc, free
 
 # Define tiny winy number
 cdef double FLT_EPSILON = np.spacing(1.0) #0.0000001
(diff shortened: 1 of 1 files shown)


hmaarrfk:
C++ got alot better when they standardized smart pointers.

stefanv:
What can we do to safeguard ourselves from all the various types of memory mistakes that we've been seeing?

skimage/measure/_marching_cubes_lewiner_cy.pyx
```diff
@@ -190,6 +187,13 @@ cdef class Cell:
         # Allocate face layers
         self.faceLayer1 = <int *>malloc(self.nx*self.ny*4 * sizeof(int))
         self.faceLayer2 = <int *>malloc(self.nx*self.ny*4 * sizeof(int))
+
+        if (self.faceLayer1 is NULL or self.faceLayer2 is NULL or
+            self.vv is NULL or self.vg is NULL or self._vertices is NULL or
+            self._normals is NULL or self._values is NULL or
+            self._faces is NULL):
+            raise MemoryError()
+
```
stefanv: What happens here if we do not de-allocate the values that already succeeded?

hmaarrfk:
you can just use numpy to allocate arrays. there is no reason to use malloc today. Maybe in the old days of cython that wasn't so true.

hmaarrfk:
You can still call numpy from within cython, it is just rather annoying. It is done in a few locations in skimage already.

hmaarrfk:
Yeah, but it is harder to "review" and makes the cython functions really verbose.

cgohlke:
> you can just use numpy to allocate arrays.

can't do that without GIL. Not sure if that'll work for OpenMP thread local buffers.

skimage/measure/_marching_cubes_lewiner_cy.pyx
cgohlke: The should be released in `__dealloc__`, no?

hmaarrfk: is __dealloc__ safe to any failed memory allocations? seems like they all need to be guarded right?

jni: Yes, `free` is a no-op with NULL pointers, and will work correctly with allocated pointers. It is *not* safe to `free` a pointer that has already been freed. However that should not happen in this case, I believe.

PR #3721, opened 2019-02-04
Enable OpenMP for msvc

cgohlke:
msvc uses the '/openmp' compiler flag to enable OpenMP.

Two potential issues:

1) Extension modules using OpenMP will depend on VCOMP*.DLL, which either needs to be shipped with the wheel or installed by the user using a matching Visual C++ Redistributable.

2) This patch probably does not work if skimage is compiled with mingw.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,9 +56,12 @@ import builtins
 builtins.__SKIMAGE_SETUP__ = True
 
 # Support for openmp
-
-compile_flags = ['-fopenmp']
-link_flags = ['-fopenmp']
+if sys.platform == 'win32':
+    compile_flags = ['/openmp']
+    link_flags = []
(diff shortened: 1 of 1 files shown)


hmaarrfk:
A step in the right direction. 

But I'm not sure about the issue listed in 1. Makes me worried

hmaarrfk:
Not sure if you meant to close this.

In either case, I wouldn't be opposed to "testing" with openmp, but distributing software isn't something that we really consider here on this repo.

The wheels are built in the scikit-image-wheels repo.

Therefore, if we could "turn it off" on windows, then maybe we can test it here at least?

PR #3723, opened 2019-02-05
Downgrade Xcode to 9.4 on master

hmaarrfk:
## Description

This is the default version from travis which seems to be used by the wheel builder.

Should be merged in conjunction with https://github.com/scikit-image/scikit-image/pull/3724

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,16 +58,20 @@ matrix:
     - os: linux
       python: 3.6
       env: QT=PyQt5 OPTIONAL_DEPS=1 PIP_FLAGS="--pre"
+    # For smooth deployment, the osx_image here should match
+    # what we set in the wheel generation travis images.
+    # If not set, it will use the default version from Travis
+    # https://docs.travis-ci.com/user/reference/osx/#xcode-version
     - os: osx
-      osx_image: xcode10.1
+      osx_image: xcode9.4
       language: objective-c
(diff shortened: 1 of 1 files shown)


stefanv:
What is the reason for downgrading?  Why not upgrade all builds to 10.4?

hmaarrfk:
The wheel builder uses 9.4.1 https://github.com/scikit-image/scikit-image-wheels/blob/master/.travis.yml#L77 as it is left empty.

hmaarrfk:
I think.

stefanv:
Right, but why not just upgrade both?

hmaarrfk:
That works too. I'm just mostly worried about the multibuild blackmagic voodo that I don't know how to use.

This is a more familar change to me.

hmaarrfk:
The bonus is that Xcode 9.4 is used by conda-forge
https://github.com/conda-forge/staged-recipes/blob/master/.travis.yml#L8

And keeping up with that is probably a good enough reason for me

stefanv:
That's fine with me, but then it should be documented in the developer tech 
manual
What manual you ask. Yes, the one we should create to document updating the 
website, eg.

stefanv:
No, certainly not!

Issue #3713, opened 2019-02-04
scikit-image 0.14.2 failes to install into virtual environment with python 3.4.9.


odinsbane:
## Description

When attempting to pip install scikit-image the process fails with ImportError: No module named 'skimage'.

Here is a paste of the full output. https://paste.pound-python.org/show/hZVjxLQPbFLZGFmY8GLE/

## Way to reproduce

virtualenv3 test-env
./test-env/bin/pip install scikit-image

# Paste the output of the following python commands
```
>>> from __future__ import print_function
>>> import sys; print(sys.version)
3.4.9 (default, Aug 14 2018, 21:28:57) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-28)]
>>> import platform; print(platform.platform())
Linux-3.10.0-957.1.3.el7.x86_64-x86_64-with-centos-7.6.1810-Core
>>> 
```

soupault:
@odinsbane could you, please, try to install Cython before executing `pip install scikit-image` and report if that helps?

odinsbane:
When I pip install cython from a fresh virtual environment nothing changes.

Installing collected packages: cython
Successfully installed cython-0.29.4

```
./scikit-test/bin/pip install cython
DEPRECATION: Python 3.4 support has been deprecated. pip 19.1 will be the last one supporting it. Please upgrade your Python as Python 3.4 won't be maintained after March 2019 (cf PEP 429).
Collecting cython
  Using cached https://files.pythonhosted.org/packages/63/a4/80dd4bacb66736a7762b2a5b8dfc66ce18d52793fd884771db104f567390/Cython-0.29.4-cp34-cp34m-manylinux1_x86_64.whl
Installing collected packages: cython
Successfully installed cython-0.29.4
```
The output of of the pip install scikit-image doesn't appear to change. I can repaste it if the details matter.

odinsbane:
Yes we are currently using centos. Thank you for the suggestions. We can get it to work, but I thought this should be reported. I can see needing to improve our install process in the near future.

It seemed odd to break an installation process at a minor revision, 0.14.2 vs 0.14.1. I also thought that pip/pypi has some sort of version checks so that if a version doesn't work with a python version it doesn't get selected. For example if version >=0.14.2 doesn't work with python 3.4.9, then that version wouldn't be available through pip.

hmaarrfk:
[...]

PR #3726, opened 2019-02-05
pypi: explicitly exclude Python 3.1, 3.2, and 3.3

stefanv:
See also: https://github.com/scikit-image/scikit-image/issues/3713

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,7 +142,7 @@ if __name__ == "__main__":
         ],
         install_requires=INSTALL_REQUIRES,
         requires=REQUIRES,
-        python_requires='>=2.7',
+        python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
         packages=setuptools.find_packages(exclude=['doc']),
         include_package_data=True,
         zip_safe=False,  # the package can run out of an .egg file


hmaarrfk:
Are you going to rebuild 3.4?

stefanv:
Cannot rebuild 3.4 without a working Matplotlib, since *that* is not yet an optional dependency.  @matthew-brett uses [this script](https://github.com/matthew-brett/manylinux-builds/blob/master/build_matplotlibs.sh) to build Matplotlib.  We only need one version that works on every platform.

hmaarrfk:
you could skip the tests no?

Matplotlib isn't necessary to build.

In either case, if you wanted matplotlib to be an optional dependency, I can put the final 30 mins of work to clean it up, but since it touches the build system, it is hard to keep the PR in a mergable state.

stefanv:
Thinking about it more: these wheels would not be installable unless we made Matplotlib fully optional.

hmaarrfk:
The work is already done for making matplotlib optional. We can add a switch that would fall back on the poolygon clip algorithm from AGG if matplotlib is installed

stefanv:
I think this is probably exactly what Matplotlib uses: https://github.com/stefanv/supreme/blob/master/supreme/ext/polygon.c#L121

hmaarrfk:
Ok, I suggested it, on the PR, but I don't think the implementation detail
is what is holding up that PR.
That implementation detail could also be fixed in a separate issue.

hmaarrfk:
Isn't matplotlib available on pypi for 3.4???
https://pypi.org/project/matplotlib/2.2.3/#files

stefanv:
On Mon, 04 Feb 2019 17:53:10 -0800, Mark Harfouche wrote:
> Isn't matplotlib available on pypi for 3.4???
> https://pypi.org/project/matplotlib/2.2.3/#files

Looks like they are; I was thinking of 32-bit.

Should check why those 0.14.x builders never built our 3.4 wheels,
then.

Something like this:
https://travis-ci.org/scikit-image/scikit-image-wheels/builds/481253099

hmaarrfk:
thats the dask array dependency, that I think I just fixed.

I think the easiest solution there is just to release 0.14.3

https://github.com/scikit-image/scikit-image/blob/v0.14.x/requirements/optional.txt#L13

hmaarrfk:
It also seems that some where missing matplotlib. that can be for 0.14.4 ;)

stefanv:
Agreed, 0.14.3 should be easy enough.

PR #3729, opened 2019-02-06
[WIP] MAINT activating test against nightly cpython build

NelleV:
## Description
This is work in progress.

This pull request activates the tests against travis' nightly build of CPython (3.8 is in pre-release right now). As per the suggestion of @stefanv , I switched things around to keep the same number of matrix entries.

This PR will probably not work out of the box.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,9 +43,6 @@ matrix:
       python: 3.5
       env: QT=PyQt5 MINIMUM_REQUIREMENTS=1
     - os: linux
-      python: 3.5
-      env: QT=PyQt5 OPTIONAL_DEPS=1 MINIMUM_REQUIREMENTS=1
-    - os: linux
       python: 3.6
       env: QT=PyQt5 OPTIONAL_DEPS=1 BUILD_DOCS=1 DEPLOY_DOCS=1
     - os: linux
@@ -53,6 +50,15 @@ matrix:
       dist: xenial # Required for Python 3.7
       sudo: true   # travis-ci/travis-ci#9069
       env: QT=PyQt5 OPTIONAL_DEPS=1 BUILD_DOCS=1
+    - os: linux
+      python: "nightly"
+      env: QT=PyQt5 OPTIONAL_DEPS=1 MINIMUM_REQUIREMENTS=1
+      addons:
+        apt:
+          packages:
+            - liblapack-dev
+            - libblas-dev
+            - gfortran
     - os: linux
       python: 3.6
       env: QT=PyQt5 OPTIONAL_DEPS=1 PIP_FLAGS="--pre"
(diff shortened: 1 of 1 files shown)


NelleV:
Also, I'm going to cancel all travis build that are not related to the one I'm adding until it runs.

NelleV:
No, that's just me screwing up. It's actually still using 3.7 on this build. I almost finished to set it up for sphinx-gallery (the codecov part is failing, dunno why), so I'm pretty sure it's possible for skimage. scipy's going to be the biggest problem, as it requires compiling from scratch and takes 10 minutes: that might be the reason to reconsider testing against the nightly build.

There might be the option on testing against a pre-release build, but I'm not sure dependencies are built against it yet so scipy might also be the main issue.

(yes, once again, I thought it take me less than 1hour…)

jni:
> (yes, once again, I thought it take me less than 1hour…)

:joy:

hmaarrfk:
Is there something upstream in the nightly builds that would help us?
Cython is just a build time dependency. People could temporary downgrade if newer versions of cython break things for us.

NelleV:
There might be a way to pull a nightly wheel built from scipy. I need to chat more with Stéfan about that option (as it requires more work, but would benefit a large portion of the community)

hmaarrfk:
Right, but is this any different than testing against pre-releases? In practice I mean.

NelleV:
Yes: the nightly build tests againts the latest master commit (at a specific time everyday) of CPython.

hmaarrfk:
ok, I can see the value in that, but in general, I think that cython is a pretty mature project, that has been pretty good at issuing warnings about depredations.

We've hit a few snags with it, but because we can always fallback for code generation, it isn't that big of a deal, I think????

stefanv:
CPython vs Cython :)

stefanv:
Or just: Python

hmaarrfk:
omg.... ok, that makes more sense...

hmaarrfk:
I'm really -1 on this.

This is ALOT of maintenance effort to get these "PreRelease" tests going with virtually 0 payout for the users.

Scikit-image doesn't do crazy things with python. It is mostly procedural code.

Dependency tracking is a mess.

conda (-forge) took a while to release 3.7. It is unlikely that scikit-image will be a blocker for anybody releasing their versions.

soupault:
I agree with Mark. I cannot recall if we ever had any issues with the latest versions of Python.

Anyway, when the new version is released, the community usually spends couple of months to adapt and upgrade. As we almost immediately add the latest versions into our build matrix, there is always enough time to fix any (I'd say, highly unlikely) issues.

Please, feel free to reopen in case of a strong argument.

PR #3732, opened 2019-02-07
Improve visual_test.py

cgohlke:
## Description

Do not run the script when running pytest on an inplace build.
Clean-up and speed-up (10x) calculations of the donuts volume.
Many style fixes, passing flake8.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.
- [x] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

### DIFF
--- a/skimage/measure/mc_meta/visual_test.py
+++ b/skimage/measure/mc_meta/visual_test.py
@@ -77,27 +77,21 @@ def donuts():
     b = -1.25 * 8.0
     c = 16.0 - 1.85 * 1.85
     d = 64.0
-    vol = np.empty((n, n, n), 'float32')
-    for iz in range(n):
-        z = iz * a + b
-        z *= z
-        zc = z + c
-        for iy in range(n):
-            y = iy * a + b
-            y1 = y - 2
-            y1 *= y1
-            y2 = y + 2
-            y2 *= y2
-            for ix in range(n):
-                x = ix * a + b
-                x *= x
-                x1 = x + y1 + zc
-                x1 *= x1
-                x2 = x + y2 + zc
-                x2 *= x2
-                vol[iz, iy, ix] = ((x1 - d * (x + y1)) * (x2 - d * (z + y2)))
-    vol += 1025.0
-    return vol
+    
+    i = np.arange(n, dtype=int)
+    ia_plus_b = i * a + b
+    ia_plus_b_square = ia_plus_b ** 2
+    z = ia_plus_b_square[:, np.newaxis, np.newaxis]
+    zc = z + c
+    
+    y1 = ((ia_plus_b - 2) ** 2)[np.newaxis, :, np.newaxis]
+    y2 = ((ia_plus_b + 2) ** 2)[np.newaxis, :, np.newaxis]
+    
+    x = ia_plus_b_square[np.newaxis, np.newaxis, :]
+    x1 = (x + y1 + zc) ** 2
+    x2 = (x + y2 + zc) ** 2
+    
+    return ((x1 - d * (x + y1)) * (x2 - d * (z + y2))) + 1025
 
 
 @contextmanager
--- a/skimage/measure/mc_meta/visual_test.py
+++ b/skimage/measure/mc_meta/visual_test.py
@@ -6,6 +6,7 @@ data.
 from __future__ import print_function
 
 import time
+from contextlib import contextmanager
 
 import numpy as np
 
@@ -33,7 +34,7 @@ def main(select=3, **kwargs):
         vol = vv.aVolume(20, 128)
         isovalue = kwargs.pop('level', 0.2)
     elif select == 3:
-        with Timer('computing donuts'):
+        with timer('computing donuts'):
             vol = donuts()
         isovalue = kwargs.pop('level', 0.0)
         # Uncommenting the line below will yield different results for
@@ -46,10 +47,10 @@ def main(select=3, **kwargs):
         raise ValueError('invalid selection')
 
     # Get surface meshes
-    with Timer('finding surface lewiner'):
+    with timer('finding surface lewiner'):
         vertices1, faces1 = marching_cubes_lewiner(vol, isovalue, **kwargs)[:2]
 
-    with Timer('finding surface classic'):
+    with timer('finding surface classic'):
         vertices2, faces2 = marching_cubes_classic(vol, isovalue, **kwargs)
 
     # Show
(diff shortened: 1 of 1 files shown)


skimage/measure/mc_meta/visual_test.py
```diff
@@ -3,70 +3,118 @@
 data.
 """
 
+from __future__ import print_function
```
jni: We don't support Py2.7 on master so this is unnecessary.

skimage/measure/mc_meta/visual_test.py
```diff
@@ -3,70 +3,118 @@
...
+            y2 = y + 2
+            y2 *= y2
+            for ix in range(n):
+                x = ix * a + b
+                x *= x
+                x1 = x + y1 + zc
+                x1 *= x1
+                x2 = x + y2 + zc
+                x2 *= x2
+                vol[iz, iy, ix] = ((x1 - d * (x + y1)) * (x2 - d * (z + y2)))
```
jni: This looks eminently vectorisable...?

skimage/measure/mc_meta/visual_test.py
```diff
@@ -3,70 +3,118 @@
...
+        self.start = 0
+
+    def __enter__(self):
+        """Enter timer context."""
+        self.start = time.time()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """Exit timer context."""
+        print('took %1.0f ms' % (1000 * (time.time() - self.start)))
+
```
jni: This is fine, but imho it is more readable using the contextlib contextmanager decorator:

```python
from contextlib import contextmanager

@contextmanager
def timer(message):
    print(message, end=' ')
    start = time.time()
    yield
    print('took {:1.0f} ms'.format(1000 * (time.time() - start)))
```

skimage/measure/mc_meta/visual_test.py
cgohlke: I added it for backport to 0.14.x

skimage/measure/mc_meta/visual_test.py
hmaarrfk: @jni, is there a way to make context managers for devices? 

In my own projects, i have "classes" to describe hardware devices, and I would like to add context managers that would simply call the "open" and "close" functions for me. Ideas?

cgohlke: Why not add `__enter__` and `__exit__` methods to the "classes"?

hmaarrfk: yeah, that is what I do now. Wondering if that is the correct approach or not.

skimage/measure/mc_meta/visual_test.py
cgohlke: I don't think this is relevant since this function is not part of the skimage "library". It can easily be speed up as is by orders of magnitude with numba.jit or cythonizing with prange, Vectorizing would probably make it less readable and use more memory.

skimage/measure/mc_meta/visual_test.py
jni: My policy generally has been to leave any backport-required code for the backport, but to each their own. ;)

skimage/measure/mc_meta/visual_test.py
jni: I agree that it'll use more memory, but I disagree that it'll be less readable. It would presumably read like an actual formula, which right now I can't for the life of me figure out based on this code. But on the other hand this was even more true in the previous version of the code! So, as mentioned before, this is fine. =)

hmaarrfk: Can we at least agree that constructs such as:
```python
z = iz * a + b
z *= z
``` 
should be replaced by
```python
z = (iz * a + b) ** 2
```

There are many of them in here, and github's suggestion feature doesn't see appropriate.

hmaarrfk: I guess not, it makes things slower. 40 ms as written, and 47 ms with the `(  ) ** 2` suggestion. 

Using fully vectorized code, it goes down to 2ms

```
def donuts():
    """Return volume of two donuts based on a formula by Thomas Lewiner."""
    n = 48
    a = 2.5 / n * 8.0
    b = -1.25 * 8.0
    c = 16.0 - 1.85 * 1.85
    d = 64.0
    
    i = np.arange(n, dtype=int)
    ia_plus_b = i * a + b
    ia_plus_b_square = ia_plus_b ** 2
    z = ia_plus_b_square[:, np.newaxis, np.newaxis]
    zc = z + c
    
    y1 = ((ia_plus_b - 2) ** 2)[np.newaxis, :, np.newaxis]
    y2 = ((ia_plus_b + 2) ** 2)[np.newaxis, :, np.newaxis]
    
    x = ia_plus_b_square[np.newaxis, np.newaxis, :]
    x1 = (ia_plus_b_square + y1 + zc) ** 2
    x2 = (ia_plus_b_square + y2 + zc) ** 2
    
    return ((x1 - d * (x + y1)) * (x2 - d * (z + y2))) + 1025
```

hmaarrfk:
I don't mind pep8 so much for whiltespace lines, most of my editors remove them automatically. spyder wasn't configured correctly...

hmaarrfk:
@cgohlke I think you are right that writing a for loop isn't a terrible idea, but this would be a perfect place where a project like pythran might be useful.

`range` and iterables in python are horrendously slow. it is one of my biggest gripes with the language.

Whats the review process say now @jni. I can't merge this myself can I? Do we have to wait for an other core-dev?

skimage/measure/mc_meta/visual_test.py
jni: > I guess not, it makes things slower. 40 ms as written, and 47 ms with the `( ) ** 2` suggestion.

Well, in this case, readability trumps speed, or we wouldn't be doing such a loop in Python.

But I agree that the vectorised version is better.

jni:
@hmaarrfk 

> Whats the review process say now @jni. I can't merge this myself can I? Do we have to wait for an other core-dev?

Great questions all! In fact we don't have a guideline for this. In the past my internal guide has been "merge ok if changes are minor (typos, comments, variable naming), wait if changes are significant (major code restructuring, logic changes, and *definitely* for API changes)". We should probably add something to the core dev guide about this.

In this specific case, the changes are significant, but it is a minor file that is rarely run, so I'd be ok with merging *if* @cgohlke is ok with the changes you made. (Personally I like them. And I've verified locally that it works.) @cgohlke?

hmaarrfk:
Yeah. Sorry about just pushing to your branch. I figured you knew how to revise to remove the changes ;)

cgohlke:
I think it's time to move on and not get lost in ms level premature optimizations on code that has not been run successfully for many years... The main purpose of this PR is **not** to run the code during tests of inplace builds.

PR #3733, opened 2019-02-09
Updated painttool to work with color images and properly scale labels.

brianhendee:
## Description

PaintTool and LabelPlugin are handy for labelling data, but when I tried to use them, they weren't actually painting the labels onto the image. It looks like there are a couple of factors.

For color images, the overlay shape is getting set as MxNx3 instead of MxN. This causes an exception when the points are assigned to scalar values. I updated the code to just use the first two dimensions and  updated the unit tests.

The matplotlib.colors.NoNorm used for scaling the label layer has some subtle behavior. It is supposed to be a "dummy replacement for Normalize." However, the documentation (https://matplotlib.org/api/_as_gen/matplotlib.colors.NoNorm.html) states that it "returns 0 if vmin==vmax." The parameters vmin and vmax were being set from the initial label layer (all zeros), causing the label layer to always get mapped to zero. I updated the NoNorm call to explicitly set vmin and vmax to span the range of the colormap, and updated the unit tests.

## For reviewers

  later.

### DIFF
--- a/skimage/viewer/tests/test_tools.py
+++ b/skimage/viewer/tests/test_tools.py
@@ -8,7 +8,7 @@ from skimage.viewer.canvastools import (
 from skimage.viewer.canvastools.base import CanvasToolBase
 
 from skimage._shared import testing
-from skimage._shared.testing import assert_equal
+from skimage._shared.testing import assert_equal, parametrize
 
 try:
     from matplotlib.testing.decorators import cleanup
@@ -160,8 +160,8 @@ def test_rect_tool():
 
 @cleanup
 @testing.skipif(not has_qt, reason="Qt not installed")
-def test_paint_tool():
-    img = data.astronaut()
+@parametrize('img', [data.moon(), data.astronaut()])
+def test_paint_tool(img):
     viewer = ImageViewer(img)
 
     tool = PaintTool(viewer, img.shape)
--- a/skimage/viewer/tests/test_tools.py
+++ b/skimage/viewer/tests/test_tools.py
@@ -190,7 +190,6 @@ def test_paint_tool(img):
 
     tool.overlay = tool.overlay * 0
     assert_equal(tool.overlay.sum(), 0)
-
     assert_equal(tool.cmap.N, tool._overlay_plot.norm.vmax)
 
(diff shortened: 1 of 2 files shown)


skimage/viewer/tests/test_tools.py
```diff
@@ -161,7 +161,7 @@ def test_rect_tool():
 @cleanup
 @testing.skipif(not has_qt, reason="Qt not installed")
 def test_paint_tool():
-    img = data.moon()
```
hmaarrfk: could you parameterize this tests taht way we know it still works for gray images.

```python
from skimage._shared.testing import parametrize

@parametrize('img', [data.moon(), data.astronaut()])
def test_pain_tool(img):
```

brianhendee: Sure no problem.

hmaarrfk: cool thanks, and I think just for you qt pushed an update today that should break the tests ;)

brianhendee: Is there a way to make Travis-CI rerun the tests?

hmaarrfk: you gotta ask a core dev nicely ;)
I did it for you.
fixes for qt 5.12 just came in, so your tests will likely pass.

Optionally, I would amend a commit and force push. Force pushing is a bad habbit though.... but with the fork, branch, PR model, I think it is OK.

```
git commit --amend
git push --force
```

hmaarrfk: I had to retrigger a build. I think we solved the particular issue on master, but travis didn't pick up the change. I simply removed a blank line, I hope that is ok.

hmaarrfk: I think it will work now,
https://travis-ci.org/scikit-image/scikit-image/jobs/494273469#L159

the qt5dev stuff was missing at the end of that line before.

brianhendee: That works...thanks.

hmaarrfk:
Thanks! Merging.

PR #3738, opened 2019-02-10
Add image.sc forum badge to README

imagejan:
This adds a badge that links to the `scikit-image` tag on the forum. As `scikit-image` is a community partner project, it should link to the forum as its primary discussion channel, that's why I put this *before* the stackoverflow badge.


stefanv:
Looks good to me.

jni:
Thanks @imagejan!

PR #3740, opened 2019-02-10
Fix several issues with relabel_sequential

uschmidt83:
## Description

- Fixing several issues with `skimage.segmentation.relabel_sequential` (see added tests).
- Slightly changing the semantic of the function by providing data type stability (if posssible).

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -119,7 +119,7 @@ def relabel_sequential(label_field, offset=1):
     offset = int(offset)
     if offset <= 0:
         raise ValueError("Offset must be strictly positive.")
-    if int(label_field.min()) < 0:
+    if np.min(label_field) < 0:
         raise ValueError("Cannot relabel array that contains negative values.")
     m = label_field.max()
     if not np.issubdtype(label_field.dtype, np.integer):
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -62,7 +62,7 @@ def relabel_sequential(label_field, offset=1):
     Parameters
     ----------
     label_field : numpy array of int, arbitrary shape
-        An array of labels.
+        An array of labels, which must be non-negative integers.
     offset : int, optional
         The return labels will start at `offset`, which should be
         strictly positive.
@@ -72,14 +72,16 @@ def relabel_sequential(label_field, offset=1):
     relabeled : numpy array of int, same shape as `label_field`
         The input label field with labels mapped to
         {offset, ..., number_of_labels + offset - 1}.
+        The data type will be the same as `label_field`, except when
+        offset + number_of_labels causes overflow of the current data type.
     forward_map : numpy array of int, shape ``(label_field.max() + 1,)``
         The map from the original label space to the returned label
         space. Can be used to re-apply the same mapping. See examples
-        for usage.
+        for usage. The data type will be the same as `relabeled`.
     inverse_map : 1D numpy array of int, of length offset + number of labels
         The map from the new label space to the original space. This
         can be used to reconstruct the original label field from the
-        relabeled one.
+        relabeled one. The data type will be the same as `relabeled`.
 
     Notes
     -----
@@ -114,20 +116,29 @@ def relabel_sequential(label_field, offset=1):
     >>> relab
     array([5, 5, 6, 6, 7, 9, 8])
     """
+    offset = int(offset)
+    if offset <= 0:
+        raise ValueError("Offset must be strictly positive.")
+    if int(label_field.min()) < 0:
+        raise ValueError("Cannot relabel array that contains negative values.")
     m = label_field.max()
-    if not np.issubdtype(label_field.dtype, np.signedinteger):
+    if not np.issubdtype(label_field.dtype, np.integer):
         new_type = np.min_scalar_type(int(m))
         label_field = label_field.astype(new_type)
         m = m.astype(new_type)  # Ensures m is an integer
     labels = np.unique(label_field)
     labels0 = labels[labels != 0]
-    if m == len(labels0):  # nothing to do, already 1...n labels
+    required_type = np.min_scalar_type(offset + len(labels0))
+    if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
+        label_field = label_field.astype(required_type)
+    new_labels0 = np.arange(offset, offset + len(labels0))
+    if np.all(labels0 == new_labels0):
         return label_field, labels, labels
-    forward_map = np.zeros(m + 1, int)
-    forward_map[labels0] = np.arange(offset, offset + len(labels0))
+    forward_map = np.zeros(int(m + 1), dtype=label_field.dtype)
+    forward_map[labels0] = new_labels0
     if not (labels == 0).any():
         labels = np.concatenate(([0], labels))
-    inverse_map = np.zeros(offset - 1 + len(labels), dtype=np.intp)
+    inverse_map = np.zeros(offset - 1 + len(labels), dtype=label_field.dtype)
     inverse_map[(offset - 1):] = labels
     relabeled = forward_map[label_field]
     return relabeled, forward_map, inverse_map
--- a/skimage/segmentation/tests/test_join.py
+++ b/skimage/segmentation/tests/test_join.py
@@ -109,7 +109,7 @@ def test_relabel_sequential_int_dtype_stability(data_already_sequential,
 
 def test_relabel_sequential_int_dtype_overflow():
     ar = np.array([1, 3, 0, 2, 5, 4], dtype=np.uint8)
-    offset = 255
+    offset = 254
     ar_relab, fw, inv = relabel_sequential(ar, offset=offset)
     assert all(a.dtype == np.uint16 for a in (ar_relab, fw, inv))
     ar_relab_ref = ar.astype(np.int, copy=True)
--- a/skimage/segmentation/tests/test_join.py
+++ b/skimage/segmentation/tests/test_join.py
@@ -3,6 +3,7 @@ from skimage.segmentation import join_segmentations, relabel_sequential
 
 from skimage._shared import testing
 from skimage._shared.testing import assert_array_equal
+import pytest
 
 
 def test_join_segmentations():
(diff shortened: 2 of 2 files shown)


skimage/segmentation/_join.py
```diff
@@ -114,20 +116,29 @@ def relabel_sequential(label_field, offset=1):
     >>> relab
     array([5, 5, 6, 6, 7, 9, 8])
     """
+    offset = int(offset)
+    if offset <= 0:
+        raise ValueError("Offset must be strictly positive.")
+    if int(label_field.min()) < 0:
```
hmaarrfk: The challenge with these kinds of checks is that they are expensive. What was the error message without that check?

hmaarrfk: I guess max is used below. so this isn't really that big of an increase in the cost.. Generally I'm against these kinds of checks.

skimage/segmentation/tests/test_join.py
```diff
@@ -91,3 +92,57 @@ def test_relabel_sequential_dtype():
...
+    if data_already_sequential:
+        ar = np.array([1, 3, 0, 2, 5, 4], dtype=dtype)
+    else:
+        ar = np.array([1, 1, 5, 5, 8, 99, 42, 0], dtype=dtype)
+    assert all(a.dtype == dtype for a in relabel_sequential(ar))
+
+
+def test_relabel_sequential_int_dtype_overflow():
+    ar = np.array([1, 3, 0, 2, 5, 4], dtype=np.uint8)
+    offset = 255
```
hmaarrfk: 255 is quite special. Can you make it 254?
```suggestion
    offset = 254
```

skimage/segmentation/_join.py
uschmidt83: I generally agree, but `relabel_sequential` is not a function you'd typically call in a hot loop.

skimage/segmentation/_join.py
```diff
@@ -114,20 +116,29 @@ def relabel_sequential(label_field, offset=1):
...
     m = label_field.max()
-    if not np.issubdtype(label_field.dtype, np.signedinteger):
+    if not np.issubdtype(label_field.dtype, np.integer):
         new_type = np.min_scalar_type(int(m))
         label_field = label_field.astype(new_type)
         m = m.astype(new_type)  # Ensures m is an integer
     labels = np.unique(label_field)
     labels0 = labels[labels != 0]
-    if m == len(labels0):  # nothing to do, already 1...n labels
+    required_type = np.min_scalar_type(offset + len(labels0))
```
hmaarrfk: min_scalar_type is a little annoying. It can return unsigned types. I think I've personally come around to not liking the use of "unsigned" unless you need specific "unsigned" behavior, that is, `255+2 == 1` is something you want to be true in your math.

Thoughts?

uschmidt83: The whole `required_type` thing is really covering an edge case that I don't think will happen much in practice, unless someone purposefully uses restricted data types (especially `uint8`).

I frequently save label masks to disk as `uint16` TIFF files. Hence, they have this type when loaded again from disk. I like working with 16-bit integers for space reasons, which can be important for large 3D arrays.

skimage/segmentation/_join.py
```diff
@@ -114,20 +116,29 @@ def relabel_sequential(label_field, offset=1):
     >>> relab
     array([5, 5, 6, 6, 7, 9, 8])
     """
+    offset = int(offset)
+    if offset <= 0:
+        raise ValueError("Offset must be strictly positive.")
```
hmaarrfk: Nice one, yeah the error message was non-sensical before
```python
In [14]: relabel_sequential(np.asarray([2, 3, 4]), -1)                                                  
ValueError                                Traceback (most recent call last)
<ipython-input-14-e5025e3e4264> in <module>
----> 1 relabel_sequential(np.asarray([2, 3, 4]), -1)

~/miniconda3/envs/owl/lib/python3.7/site-packages/skimage/segmentation/_join.py in relabel_sequential(label_field, offset)
    129         labels = np.concatenate(([0], labels))
    130     inverse_map = np.zeros(offset - 1 + len(labels), dtype=np.intp)
--> 131     inverse_map[(offset - 1):] = labels
    132     relabeled = forward_map[label_field]
    133     return relabeled, forward_map, inverse_map

ValueError: could not broadcast input array from shape (4) into shape (2)
```

skimage/segmentation/tests/test_join.py
```diff
@@ -91,3 +92,57 @@ def test_relabel_sequential_dtype():
...
+    assert all(a.dtype == np.uint16 for a in (ar_relab, fw, inv))
+    ar_relab_ref = ar.astype(np.int, copy=True)
+    ar_relab_ref[ar_relab_ref > 0] += offset - 1
+    assert_array_equal(ar_relab, ar_relab_ref)
+
+
+def test_relabel_sequential_negative_values():
+    ar = np.array([1, 1, 5, -5, 8, 99, 42, 0])
+    with pytest.raises(ValueError):
+        relabel_sequential(ar)
```
jni: Yikes, I did not realise just how bad the results are right now when you have negative labels:

```python
In [3]: segmentation.relabel_sequential(np.array([-1, 0, 2, 2, 3, 8, -1]))
Out[3]: 
(array([4, 0, 2, 2, 3, 4, 4]),
 array([0, 0, 2, 3, 0, 0, 0, 0, 4]),
 array([-1,  0,  2,  3,  8]))
```

Note that 4 ends up representing both 8 and -1!

jni: Personally, though, I might offset the labels with the minimum value + 1 if there are negative labels, and then proceed as normal. Thoughts?

skimage/segmentation/tests/test_join.py
```diff
@@ -91,3 +92,57 @@ def test_relabel_sequential_dtype():
...
+@pytest.mark.parametrize('offset', (1, 5))
+@pytest.mark.parametrize('with0', (False, True))
+def test_relabel_sequential_already_sequential(offset, with0):
+    if with0:
+        ar = np.array([1, 3, 0, 2, 5, 4])
+    else:
+        ar = np.array([1, 3, 2, 5, 4])
+    ar_relab, fw, inv = relabel_sequential(ar, offset=offset)
+    ar_relab_ref = ar.copy()
+    ar_relab_ref[ar_relab_ref > 0] += offset - 1
```
jni: alternatively, `ar_relab_ref = np.where(ar > 0, ar + offset - 1, 0)`

skimage/segmentation/_join.py
jni: :+1:

skimage/segmentation/_join.py
jni: The `int` seems unnecessary? 

```suggestion
    if np.min(label_field) < 0:
```

skimage/segmentation/_join.py
```diff
@@ -114,20 +116,29 @@ def relabel_sequential(label_field, offset=1):
...
+    if not np.issubdtype(label_field.dtype, np.integer):
         new_type = np.min_scalar_type(int(m))
         label_field = label_field.astype(new_type)
         m = m.astype(new_type)  # Ensures m is an integer
     labels = np.unique(label_field)
     labels0 = labels[labels != 0]
-    if m == len(labels0):  # nothing to do, already 1...n labels
+    required_type = np.min_scalar_type(offset + len(labels0))
+    if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
+        label_field = label_field.astype(required_type)
```
jni: Very nice.

skimage/segmentation/_join.py
hmaarrfk: Want to allow a dtype parameter to ensure uint16???

jni: I think the new logic (use the larger of min_dtype and input dtype) is sufficient here.

hmaarrfk: Sure.

skimage/segmentation/tests/test_join.py
uschmidt83: I don't think that works with the current semantics of the function that the input can contain any negative values, since `label_field` can be used to index into `fw`. From the example in the docstring:
```python
>>> relab, fw, inv = relabel_sequential(label_field)
>>> (fw[label_field] == relab).all()
```

jni: A very good point! Ignore me! =P

skimage/segmentation/tests/test_join.py
uschmidt83: Sure, that's better.

skimage/segmentation/_join.py
uschmidt83: It is unnecessary if `label_field` is an integer array. But the code also covers the case where the input is of a different data type (e.g. float). That's why I thought it doesn't hurt to wrap it in `int()` just to be sure.

jni: Hmm. I'm surprised to find that the function indeed works with float types, but I guess if it does, let's not break it. However, `int(-0.01)` is 0, and I wonder whether I'd rather we raised in that case...

skimage/segmentation/_join.py
```diff
@@ -114,20 +116,29 @@ def relabel_sequential(label_field, offset=1):
...
+    new_labels0 = np.arange(offset, offset + len(labels0))
+    if np.all(labels0 == new_labels0):
         return label_field, labels, labels
-    forward_map = np.zeros(m + 1, int)
-    forward_map[labels0] = np.arange(offset, offset + len(labels0))
+    forward_map = np.zeros(int(m + 1), dtype=label_field.dtype)
```
uschmidt83: Btw, I replaced `m + 1` with `int(m + 1)` to fix an issue when `m` is of type `np.uint64`.

Is this intended behavior or a numpy bug?
```python
>>> (np.uint64(5) + 1).dtype
dtype('float64')
```

skimage/segmentation/_join.py
uschmidt83: I don't really care, to be honest. Your call.

skimage/segmentation/tests/test_join.py
uschmidt83: Should I change and commit this?

skimage/segmentation/_join.py
jni: yikes! I think actually I came across this and the answer was that it is indeed intended, because (a) NumPy's type promotion rules do not depend on the input values, only on the types, and (b) they are supposed to be safe in the sense of being able to represent the result, and `float64` is the only thing that can represent anything with a Python int, since they are unbounded.

At least, that's my memory of it. Perhaps @stefanv has more details. But, either way, thanks for the fix!

skimage/segmentation/tests/test_join.py
jni: Yeah go for it!

skimage/segmentation/_join.py
jni: My slight preference is that you accept the suggestion. =) It is more readable and I think if someone is sending negative floats to this function, they should get an error back.

skimage/segmentation/tests/test_join.py
uschmidt83: Done.

skimage/segmentation/_join.py
hmaarrfk: What is this sorcery!!!!

skimage/segmentation/_join.py
```diff
@@ -114,20 +116,29 @@ def relabel_sequential(label_field, offset=1):
...
+        raise ValueError("Cannot relabel array that contains negative values.")
     m = label_field.max()
-    if not np.issubdtype(label_field.dtype, np.signedinteger):
+    if not np.issubdtype(label_field.dtype, np.integer):
         new_type = np.min_scalar_type(int(m))
         label_field = label_field.astype(new_type)
         m = m.astype(new_type)  # Ensures m is an integer
     labels = np.unique(label_field)
     labels0 = labels[labels != 0]
-    if m == len(labels0):  # nothing to do, already 1...n labels
```
uschmidt83: Btw, the previous line
```python
if m == len(labels0):  # nothing to do, already 1...n labels
```
was only valid for `offset = 1`. This was only mentioned in the comment but not checked, i.e. it should've been
```python
if offset == 1 and m == len(labels0):  # nothing to do, already 1...n labels
```
Anyway, I guess hardly anyone uses `offset != 1`. Why was the previous function `relabel_from_one` replaced with `relabel_sequential`? (Since the `offset` introduces quite a bit of (subtle) complexity.)

Also, `relabel_sequential` assumes that the background has value 0, but I noticed `color.label2rgb` assumes (by default) that the background label has value -1.

jni: > Why was the previous function `relabel_from_one` replaced with `relabel_sequential`?

Someone requested the functionality and it seemed like the right thing to do and an easy fix. Oops. =)

> Also, `relabel_sequential` assumes that the background has value 0, but I noticed `color.label2rgb` assumes (by default) that the background label has value -1.

Yes, historically, skimage used -1 as the background label, but we have slowly started homogenising to 0. But this will take time.

uschmidt83: Good to know, thanks!

jni:
@scikit-image/core anyone else want to review this? The Travis failures are due to the Qt 5.12 bug that was fixed on master.

uschmidt83:
Anything else I can do to get this merged?

skimage/segmentation/_join.py
stefanv: Looks like the argument Juan outlines is being followed here.  Counter-intuitive, but probably correct?  Same as `(np.uint8(3)/2).dtype`.

jni:
Thanks for the ping @uschmidt83!

Issue #3739, opened 2019-02-10
Builds broken with QT 5.12


hmaarrfk:
## Description
Qt 5.12 was released today, and likely broke matplotlib. 
https://travis-ci.org/scikit-image/scikit-image/jobs/491337465#L1931

https://travis-ci.org/scikit-image/scikit-image/jobs/491337465#L2058

Windows seems ok???

## Way to reproduce
See travis, linux 5.12

jni:
Fun times! Shall we pin PyQt!=5.12 in our testing dependencies?

hmaarrfk:
5.12.0, or 5.12.1???

jni:
5.12.anything, no? I mean, I don't know, you tell me LOL.

hmaarrfk:
Did upstream identify anything?

hmaarrfk:
OMG I pushed to master......... So sorry

hmaarrfk:
Well it seemed to have worked..... But maybe we should keep the issue for now

jni:
@hmarrfk I’d prefer if you reset the changes to master (being extra careful it should go without saying) and issued a pull request as usual.

hmaarrfk:
Ok. I didn’t know if two wrongs made a right. I didn’t notice that the default behavior of the editor changed....

jni:
@hmaarrfk well I used the word "reset" on purpose but it seems we don't have that power, which is a relief, tbh. ;) So it wasn't so much two wrongs (although they were two wrongs) as sweeping the first wrong under the rug.

At any rate, it seems my historical distrust of editor-based git remains warranted.

tacaswell:
If you reset master there is a race condition where someone could have pulled after the bad push but before the reset. You will have to find everyone who did and hard-reset their local repositories.  I think once it is pushed to master on github it is out there with no (or only very very expensive) take-backs.

stefanv:
Branch protection is enabled for `master`, I see.  @jni @hmaarrfk Did you toggle this after the incident?  If not, I am unsure why the commit went in.

hmaarrfk:
no no, i didn't reset it, i just pushed an other commit. I've dealt with enough race conditions on my own machines to know better.

I Ididn't change any settings on scikit-image. Trying to use my rights sparingly.....

jni:
@tacaswell 

> If you reset master there is a race condition where someone could have pulled after the bad push but before the reset.

[...]

PR #3744, opened 2019-02-12
Fix matplotlib + qt 5.12 the same way upstream does it

hmaarrfk:
## Description
closes: https://github.com/scikit-image/scikit-image/issues/3739
xref: https://github.com/matplotlib/matplotlib/pull/13412

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,9 @@ addons:
     - texlive
     - texlive-latex-extra
     - dvipng
+    # this became necessary to use matplotlib and qt5.12
+    # https://github.com/matplotlib/matplotlib/pull/13412
+    - qtbase5-dev
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ addons:
     - dvipng
     # this became necessary to use matplotlib and qt5.12
     # https://github.com/matplotlib/matplotlib/pull/13412
-    # - qtbase5-dev
+    - qtbase5-dev
 
 env:
   global:
--- a/tools/travis/install_qt.sh
+++ b/tools/travis/install_qt.sh
@@ -11,6 +11,10 @@ if [[ "${QT}" == "PyQt5" ]]; then
         # Which conflicts with how matplotlib detects the
         # presence of PyQt before MPL 2.2.3
         pip install --retries 3 -q $PIP_FLAGS "pyqt5<5.11"
+    elif [[ `lsb_release  -r -s` == "14.04" ]]; then
+        # Apparently Qt 5.12 is only supported by ubuntu 16.04
+        # https://github.com/scikit-image/scikit-image/pull/3744#issuecomment-463450663
+        pip install --retries 3 -q $PIP_FLAGS "pyqt5<5.12"
     else
         pip install --retries 3 -q $PIP_FLAGS pyqt5
     fi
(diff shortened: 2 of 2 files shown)


jni:
There's always something... =P

This doesn't seem to be working (the two failures are identical)...

tacaswell:
@smheidrich figured it out.  You may need to also clear the caches in travis...

jni:
@tacaswell thanks! Ok I've tried deleting recent caches (couldn't do all for some reason), and restarting the failed builds, let's see how we go...

hmaarrfk:
@jni, I still see some caches from PRs. Did you want to "delete all caches". what are the caches actually helping?

jni:
@hmaarrfk I tried "delete all caches" but Travis is unhappy about doing that it seems. I clicked on as many as I could, including all the ones more recent than 24h. Clearly not sufficient. =\

smheidrich:
According to the [docs](https://docs.travis-ci.com/user/caching/), "[i]f a branch does not have its own cache, Travis CI fetches the cache of the repository’s default branch", so it should be enough to just delete the cache for master. If you did that and it still doesn't work, then probably the cache isn't the issue.

hmaarrfk:
I tried to restart one build in isolation. Do you think it is getting the cache from the previous builds in the same PR?

hmaarrfk:
seems like that isn't the issue :/

smheidrich:
I ran a build with more verbose ``pytest`` output, and [what it says](https://travis-ci.org/smheidrich/scikit-image/jobs/492916293#L2765) is:
```
/home/travis/venv/bin/python: relocation error: /home/travis/venv/lib/python3.6/site-packages/PyQt5/Qt/plugins/platforms/../../lib/libQt5DBus.so.5: symbol dbus_message_get_allow_interactive_authorization, version LIBDBUS_1_3 not defined in file libdbus-1.so.3 with link time reference
```
I think this is different from the [error in matplotlib's builds](https://travis-ci.org/matplotlib/matplotlib/jobs/491475297#L2239-L2256), but I'm not sure what to do about it nor why it doesn't appear for matplotlib.

tacaswell:
we bumped to xenial to get py37 iirc

smheidrich:
Hmm you're right, and for the earlier Python versions, switching to xenial didn't help in my tests... in that case, I'm out of ideas.

hmaarrfk:
so we should unpin for 5.12, but keep at 5.10 for anything less than xenial.

That sounds like a plan

hmaarrfk:
@smheidrich it seems xenial helped your tests. You probably got stuck on some xkb bug that we fixed recently (i htink). At least your xenial tests went super far.

smheidrich:
@hmaarrfk One weird thing is: the error message mentioning xcb (if you meant that instead of xkb) is the one that appeared in matplotlib's builds prior to adding ``qtbase5-dev``, except in scikit-image's case it also mentions [``could not connect to display``](https://travis-ci.org/smheidrich/scikit-image/jobs/493015222#L7897). And I guess it comes from a different place in the Python code.

hmaarrfk:
So i think I finally got it, we need both `qtbase5-dev` and the check for ubuntu 16.04.

Ubuntu 14.04 is going to be EOL in April, so I guess we should continue testing on that?
https://wiki.ubuntu.com/Releases

jni:
@scikit-image/core Can someone merge? Let's fix our build matrix!

PR #3749, opened 2019-02-14
Make local_maxima return a boolean array

### LINKED ISSUE #3746: local_maxima should return boolean array
Basically the output of `local_maxima` (and `local_minima`) should return a boolean array. Right now the returned dtype uses uint8 (1s and 0s) which can't be used to index numpy arrays as a mask. See this discussion thread in https://github.com/scikit-image/scikit-image/pull/3245#discussion_r255344610.

What is the correct depreciation strategy here?

cc @jni

lagru:
## Description

* Make `skimage.morphology.local_maxima` (and `~.local_minima`) return a boolean array if `indices=False`. This way the output can be used as a mask to index an array of the same shape (see also #3746)
* Make the parameter `selem` expect a boolean array which is more in line with [`scipy.ndimage.generate_binary_structure`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.generate_binary_structure.html) and the output of this function itself. Passing in an array of 0s and 1s (the old behavior) still works and is covered by tests as well. 

Closes #3746.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/extrema.py
+++ b/skimage/morphology/extrema.py
@@ -317,7 +317,7 @@ def _resolve_neighborhood(selem, connectivity, ndim):
         selem = ndi.generate_binary_structure(ndim, connectivity)
     else:
         # Validate custom structured element
-        selem = np.asarray(selem, dtype=bool)
+        selem = np.asarray(selem, dtype=np.bool)
         # Must specify neighbors for all dimensions
         if selem.ndim != ndim:
             raise ValueError(
@@ -345,9 +345,9 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
         An n-dimensional array.
     selem : ndarray, optional
         A structuring element used to determine the neighborhood of each
-        evaluated pixel. It must contain only 1's and 0's, have the same number
-        of dimensions as `image`. If not given, all adjacent pixels are
-        considered as part of the neighborhood.
+        evaluated pixel (``True`` denotes a connected pixel). It must be a
+        boolean array and have the same number of dimensions as `image`. If not
+        given, all adjacent pixels are considered as part of the neighborhood.
     connectivity : int, optional
         A number used to determine the neighborhood of each evaluated pixel.
         Adjacent pixels whose squared distance from the center is larger or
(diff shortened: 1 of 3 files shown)


skimage/morphology/tests/test_extrema.py
```diff
@@ -266,50 +267,64 @@ def test_dtypes_old(self):
...
         result_conn3 = extrema.local_maxima(self.image, connectivity=3)
+        assert result_conn3.dtype == np.bool
         assert_equal(result_conn3, self.expected_default)
 
     def test_selem(self):
-        """Test results if selem is an array."""
-        selem_cross = np.array([[0, 1, 0], [1, 1, 1], [0, 1, 0]])
+        """Test results if selem is given."""
+        selem_cross = np.array(
+            [[0, 1, 0], [1, 1, 1], [0, 1, 0]], dtype=np.bool)
```
hmaarrfk: I feel like you should keep the input dtypes the same.

To test for new dtypes, parametrize the test.

hmaarrfk: I guess that is done below.

skimage/morphology/extrema.py
```diff
@@ -345,9 +345,9 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
         An n-dimensional array.
     selem : ndarray, optional
         A structuring element used to determine the neighborhood of each
-        evaluated pixel. It must contain only 1's and 0's, have the same number
-        of dimensions as `image`. If not given, all adjacent pixels are
-        considered as part of the neighborhood.
+        evaluated pixel (``True`` denotes a connected pixel). It must be a
+        boolean array and have the same number of dimensions as `image`. If not
+        given, all adjacent pixels are considered as part of the neighborhood.
```
jni: Strictly speaking, *if neither selem nor connectivity are given*...

skimage/morphology/extrema.py
```diff
@@ -415,28 +414,28 @@ def local_maxima(image, selem=None, connectivity=None, indices=False,
...
 
     >>> local_maxima(image)
-    array([[0, 0, 0, 0, 0, 0, 0],
-           [0, 1, 1, 0, 0, 0, 0],
-           [0, 1, 1, 0, 0, 0, 0],
-           [1, 0, 0, 0, 0, 0, 1]], dtype=uint8)
+    array([[False, False, False, False, False, False, False],
+           [False,  True,  True, False, False, False, False],
+           [False,  True,  True, False, False, False, False],
+           [ True, False, False, False, False, False,  True]], dtype=bool)
```
jni: Typically in these situations I would use `local_maxima(image).astype(int)` so that the printout is readable. But this is (a) not critical, and (b) debatable anyway. Just thought I would put it out there.

lagru: Yeah, I'm on the fence as well. `astype(int)` would make the printed output more readable but hide the fact that the returned array is of boolean type. I'm +0 on leaving it like this.

JDWarner: I'd be fine with it as-is.  Another option would be changing the print options at the top of the docstring example, like

``` python
>>> np.set_printoptions(formatter={'bool':lambda x: str(int(x))})
```

so no need to do `.astype()` or `.view()` on every line.

JDWarner:
Merging, thanks @lagru!

This would be a good candidate for a backport.

JDWarner:

PR #3750, opened 2019-02-14
[WIP] Add axes argument to register_translation

taylor-scott:
## Description

Add optional argument `axes` to `register_translation` that allows users to register over a subset of the axes in their image. 

The use case I have in mind is stabilizing a time lapse by registering each frame to the previous:

```python
# image.shape = (t, r, c)
shifts, _, _ = register_translation(image[:-1], image[1:], axes=2)
# shifts.shape = (t, 2)
```

While I'm working on this, `axes=n` will register the last `n` axes of the image. Once I get the logic figured out, I'll make it so `n > 0` registers the first `n`  axes and `n < 0` registers the last `n` frames.

I'll also add an example (especially if #3745 gets merged before this one!).

## To do
- [ ] Subpixel registration
- [ ] Write tests for error and phase calculation
- [ ] Positive and negative axes arguments
- [ ] Example in documentation

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -76,7 +76,7 @@ def _upsampled_dft(data, upsampled_region_size,
 
         shifts = -ax_offset * np.fft.fftfreq(n_items, upsample_factor)[:, None]
         shifts = np.squeeze(np.exp(-im2pi * shifts.T))
-        data *= shifts.reshape(*shifts.shape, *(1, )*(nreg - 1))
+        data *= shifts.reshape(*shifts.shape, *(1, ) * (nreg - 1))
 
         # Equivalent to:
         #   data[i, j, k] = kernel[i, :] @ data[j, k].T
@@ -240,7 +240,7 @@ def register_translation(src_image, target_image, upsample_factor=1,
         flat_maxima = np.argmax(flat_CC, axis=-1)
         CCmax = flat_CC[..., flat_maxima]
         maxima = np.stack(np.unravel_index(flat_maxima,
-                                           (upsampled_region_size,)*axes),
+                                           (upsampled_region_size,) * axes),
                           axis=-1)
 
         shifts = shifts + (maxima - dftshift) / upsample_factor
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -45,8 +45,12 @@ def _upsampled_dft(data, upsampled_region_size,
     """
     if axes is None:
         axes = np.arange(data.ndim)
-
+    else:
+        axes = np.array(axes)
+    data_shape = np.array(data.shape)
+    register_shape = data_shape[axes]
     nreg = len(axes)
+    broadcast_shape = data_shape[:data.ndim - nreg]
     # if people pass in an integer, expand it to a list of equal-sized sections
     if not hasattr(upsampled_region_size, "__iter__"):
         upsampled_region_size = np.asarray([upsampled_region_size, ] * nreg)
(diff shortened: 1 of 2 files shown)


stefanv:
Will the result be a tuple of shifts, or will the shifts be combined?

taylor-scott:
The plan is to return an array where the shifts are sort of broadcast along the unregistered axes.

So, if `image.shape = (d_1, d_2, ..., d_N)` then `register_translation(..., axes=-m)` will return an array of shape `(d1, d2, ..., d_(N-m), m)`.

Returning a tuple is just as easy though if that makes the most sense.

skimage/feature/register_translation.py
```diff
@@ -191,20 +191,20 @@ def register_translation(src_image, target_image, upsample_factor=1,
...
     if upsample_factor == 1:
         if return_error:
             src_amp = np.sum(np.abs(src_freq) ** 2, axis=register_axes) \
-                             / src_freq.size
+                / np.product(register_shape)
             target_amp = np.sum(np.abs(target_freq) ** 2, axis=register_axes) \
-                                / target_freq.size
-            CCmax = cross_correlation[maxima]
+                / np.product(register_shape)
+            CCmax = flat_CC[..., flat_maxima]
```
hmaarrfk: can you please explain the bugfix here. The changes look more stylistic than anything else.

Is there a reason why you prefer `np.product(shape)` vs `size`

hmaarrfk:
@taylor-scott, will you be adding a loop inside? or is this bulk operation parallelizable?

skimage/feature/register_translation.py
taylor-scott: I'm trying to get just the size of the registered axes instead of the size of the entire image. I use `np.product(register_shape)` several times so it's probably worth defining `register_size = np.product(register_shape)` to make this more clear.

taylor-scott:
@hmaarrfk I don't *think* I will need to loop, everything should vectorize. Granted, I'm still working on this and (in particular) `_upsampled_dft` is a little fragile at the moment...

skimage/feature/register_translation.py
hmaarrfk: Just use the numpy attribute `size` like it was in the original code:
```python
In [1]: import numpy as np                                                                                                                            

In [2]: a = np.zeros((4, 5))                                                                                                                          

In [3]: a.shape                                                                                                                                       
Out[3]: (4, 5)

In [4]: a.size                                                                                                                                        
Out[4]: 20
```

There is one commit where you say "bugfix" and it is unclear if you are refactoring, or actually fixing a bug. If you fixing a bug, It would be cool to actually port it back while you are working on this.

hmaarrfk: My bad on my previous comment, you are right. ok. I get what is happening. yeah, maybe defining it as a seperate variable is good. The numpy guys just told me not to repeatidly use `np.product` on tuples, because the numpy coercion is quite slow.

taylor-scott: Sorry for the unclear commit message -- there is no bug in the original code. 

I will define it as a separate variable to avoid the repeated `np.product` calls.

hmaarrfk: thats fine, I was just trying to better understand what you were doing. bit size review you know ;)

hmaarrfk:
lets discuss `_upsample_dft` in the other issue you raised.

hmaarrfk:
I think I can see the use of this. It MIGHT perform better.
Make sure you include a benchmark. Before with a loop, after with your vectorized code.

The following:
```
register_translation(image[:-1], image[1:], axes=2)
```
I think you will have to use `space='fourier'` or whatever the argument should be to see improvements. otherwise, you are computing the large fourier transforms twice. 

it seems that FFTW does have an API for doing multiple identical transforms:
http://www.fftw.org/fftw3_doc/Advanced-Complex-DFTs.html
so maybe you can compute the transforms quickly once.

I'm not sure if people actually use it in python.

taylor-scott:
I agree about needing `space="fourier"`. Otherwise the execution time will be dominated by the fft calls.

I will be sure to include benchmarks once I'm done. Maybe I'll get asv working this time...

matthew-brett:

<details>

## Rationale

</details>

PR #3751, opened 2019-02-15
Remove seam carving

### LINKED ISSUE #3646: Seam carving is patented; remove from library
https://patents.google.com/patent/US8213745B2/en

Ref: #2808 #3082 #3643 

This code may be made available outside of scikit-image under the same license, but our policy is not to include algorithms covered under active patents.

jni:
Seam carving is a patented algorithm and so cannot be distributed with
scikit-image, which must offer code that can be used "without charge and
without restriction".

Currently, attempting to run `seam_carve` will cause a
`NotImplementedError`, and then it can be removed altogether in 0.16.

Fixes #3646

## For reviewers

  later.


sciunto:
Is there a plan to keep this code somewhere in another repository, to do not lost the implementation for those who are not impacted by software patents?

stefanv:
@sciunto If someone wants to save the code for their own use, they are most welcome to do so. I have a strong preference to not host it under the scikit-image org, however.

jni:
I'm with @stefanv. Even hosting a reference implementation is costly, and anyway we'd need to figure out the legalities of it.Anyway we are not going so far as to remove it from our git history, so the code is there for posterity, on any of the release tags that contained it. Which reminds me, I should make a note in the release notes. Hold please.

soupault:
I assume we also need to backport this, at least, to 0.14.x.

JDWarner:
I really hate the idea of backporting this, and I can't state that strongly enough.  When I consider a legacy major version, backports should fix bugs and make sure there's forward compatibility.  That's all.  I really, *really* would dislike a backport that removed functionality (there would be four-letter words aimed at the devs).  We're setting people up for a situation where they would have to pin 0.14.2 specifically and then manually cherry-pick our backports EXCEPT this one.  That's a great way to get people to abandon using a library.

It sets a precedent where anyone using a given feature on any given major point version may have that pulled out from under them - even if they pin that major point version for future compatibility!  Scary.

jni:
@JDWarner I can sympathise, and my mind is certainly not made up about this. However if we agree with the general sentiment that we shouldn't distribute patented algorithms, then it makes sense to backport. Consider the more extreme case where we discovered that a contribution violated copyright: we would be legally obligated to remove and update, and possibly even to pull/replace offending versions from PyPI and conda-forge.

I don't think it's the case here, and if @stefanv agrees, I'm always happy to do less work. ;) 

Incidentally, though, affected users wouldn't need to pin skimage to a patch version, but rather grab the seam carving code and do with it what they will.

JDWarner:
Conceptually, despite the fact that we 'distribute' the entire library with minor point releases a case can be made that we're really providing incremental changes/fixes.  Further 0.14.x releases don't distribute anything new with regard to this algorithm.  And unless we have somewhere - as in, at bare minimum a Pip installable package - to point everyone who was using this before, "just grab the code" is user hostile.

Unless we are going scorched earth, and it doesn't look like it, I am firmly opposed to backporting this.  Doing so would be a very scary precedent.

The closest precedent I know of was the Lenna image.  And there we didn't go scorched earth, and to my knowledge that image was never removed from prior versions in a backport (granted, backports were harder at the time).  "Fix things moving forward, but don't change history in a user hostile way" is how I would word a guideline on this.

jni:
Yep, I'm on board. =)

hmaarrfk:
```
generating gallery for auto_examples/transform... [ 50%] plot_seam_carving.py   
```

soupault:
@JDWarner I understand your concerns, especially regarding the precedents, but the user experience is, unfortunately, not the only thing we have to consider here. Please, notice that the LTS release was also a precedent for us (to the best of my knowledge), so there is, definitely, a need to discuss and agree on the considered issue.

I'm personally not an expert in the legal field, but it is quite obvious that there is always a balance between the spirit and the letter of the law. According to the letter, most likely, we are obliged to remove the implementation from all the packaged versions over all the distribution sources (PyPi, Conda, GitHub history). In practice, presumably, in order to reasonably lower the infringement risks we have to do this only for the versions we're actively maintaining and/or promoting and/or distributing, which means (as of Feb 2019) 0.14.x and 0.15dev/master. By applying the change only to 0.15, we're, essentially, balancing between increased user dissatisfaction and increased risks for the project owners. I'm not certain if such decision would be optimal, but it is, definitely, a suitable one.
You're right regarding the removal of Lena image: we performed that in a forward-only diff mode and didn't have any issues (at least, I'm not aware of any).

Of course, it is ultimately up to the project owners to decide how much risk they're willing to take, but, from my perspective, removing the functionality from the both active branches is the safest choice.
If Juan is fine with your proposal, I certainly don't have any objections to that.

JDWarner:
@soupault First let's be clear: the 'law' in question is our own policy.  There is no pending legal action, no expectation of pending legal action (Patent is owned by Google, months before the original PR the author was supported by Google doing a Scikit-Image GSoC, Google uses Python extensively and is very aware of this project), and as above we are not seriously considering going scorched earth (which would be attempting to remove all Git history and prior shipped versions).  This is out of an abundance of caution and spiritual adherence to FOSS principles, not a legal requirement.

The risk is already negligible, or we would be having a different discussion.

There is a reason we aren't going scorched earth.  Ripping that much history out is simply now how Git is built, it'll be difficult to do this wading through 4 years of commits.  Furthermore, as just one example, PyPi refuses to allow any changes to existing versions.  You can hide them, but never change them.  So anyone that wants or needs 0.14.[012] would simply be out of luck, as to the public these versions would no longer be obtainable.

We *cannot* be cavalier or underestimate the terrible precedent this would create, nor underestimate the user reaction.  FOSS and Git are designed the way they are to limit the ability to change history.  We would deservedly be crucified by everyone affected in the community - and scare a much wider subset from using the library - if we backport this.

jni:
@JDWarner the patent is owned by Mitsubishi if I understand correctly?

@soupault I mostly agree with @JDWarner that the risk is minor, and anyway that we would be fine ripping it out *if asked* by the patent holder(s). Ultimately though I would like to leave this decision to @stefanv as the creator of the library, originator of the policy, and the only one (as far as I can tell) to have run afoul of software patents in the past (if I understand correctly he was told by the SIFT patent holders, if you want the algo in skimage, pay up). 

In the meantime, the decision **has** been made that it'll be removed in 0.15 and above, and Travis is happy now! ;)

hmaarrfk:
>  SIFT patent holders, if you want the algo in skimage, pay up

I didn't know about that!

Does opencv pay? I guess they already got axed from opencv
https://www.pyimagesearch.com/2015/07/16/where-did-sift-and-surf-go-in-opencv-3/

jni:
> I guess they already got axed from opencv

Presumably for the same reasons. I disagree with the author of that blog post about the harm differential of the two options. (And we don't know what negotiations, if any, happened behind the scenes.)

btw @hmaarrfk you have the power to push the green button here. ;)

hmaarrfk:
I mean, I'm still kinda observing this discussion.

It is unclear to me if anybody will actually go after scikit-image. the library is a little slow....
I think opencv has some real marketshare, as large companies are including it in their productlines (xilinx, intel vision). 

Finally, just about everything is patented. You might as well go after all professors who write papers, claiming that their work, tenure, grant awards, couldn't have been completed without your patents. And as such, you should be compensated for your work.

Many people just collect patents. Is an implementation of this algorithm for sale elsewhere? If scikit-image is to be used by people learning about image processing, should we be stopping them from learning?

Should we setup an other "non-free" repo and move that there?

Nobody has sent us a letter, and it is unclear to me that anybody is being harmed by this inclusion. Therefore, I think it is OK to keep discussing it

jni:
> Finally, just about everything is patented.

This is patently 😛 false. Specifically, any method for which a paper was published in a journal *before* a patent was applied for *cannot* be patented as it would have existed as "prior art".

> Should we setup an other "non-free" repo and move that there?

See the discussion above, @stefanv strongly prefers for this to happen not under skimage. Feel free to copy it over!

> Is an implementation of this algorithm for sale elsewhere?

Apparently Adobe paid to put it in Photoshop (though the reference link in Wikipedia is dead).

> it is unclear to me that anybody is being harmed by this inclusion.

Our message is that scikit-image hosts algorithms that are free of charge and free of restriction. This is the first thing on the home page, and it means that users should be able to use scikit-image without fear of being sued by a big company for part of your profits. This doesn't happen in the research circles we move in but it happens *all the time* in corporate circles. No, if someone uses seam carving from skimage in their paper, probably noting would happen. If someone makes $200K selling a small app that uses skimage to do seam carving, there is a real risk that MERL would come in for a piece of that pie. We *could* start offering patented algorithms with a big warning in the docstring or something, but that is not a space we want to enter, as it means (among other things) that we would have to put a big fat asterisk on our front page message.

JDWarner:
I'm going to go ahead and merge this, though we should continue the discussion.  No backport, for reasons I've articulated above.

I do agree that keeping it in has minimal risk, @hmaarrfk.  I would also not be surprised if an exhaustive global patent search would find additional supposedly patented algorithms (you might well find some overly broad patents that snuck through with vague claims boiling down to "image analysis on computer", which would invalidate the whole library!).  At some level you have to stop the insanity.  Hopefully we get some improved legislation on this in the USA soon.

The model with a nonfree separate repo that could be combined by the user would be another good option, used by multiple FOSS projects including opencv.  It's certainly possible, but since @stefanv didn't want this to be hosted under the organization doing so in an official manner is difficult.

hmaarrfk:
sold.

hmaarrfk:
I think maybe we can come to a compromise, which is to release 0.14.3 as 0.15, and to release what we wanted to be 0.15 as 0.16. 

I think that solves the "breaking things accross a release" and protects skimage's brand as not (knowingly) releasing patented software, and keeps us honest with version numbers.

That said, maybe we should as a lawyer for future issues like this?

JDWarner:
I'm :-1: on this, as it's just a backport in different clothing - also contradicts the long-term support guarantee and statements made about 0.14.x being the final 2.x support release.  Also, the LTS of 0.14.x is broken unless we want to continually do TWO backports from the dev branch moving forward.

soupault:
@JDWarner thanks for your elaborate message. I tend to agree with you on the level of risks and potential implications of different decision options. For my personal projects I'd, most likely, take the same path. I'm happy with both the discussion happening and the reached agreement.

>as large companies are including it in their productlines (xilinx, intel vision)

@hmaarrfk FYI, Itseez (who is the major developer of OpenCV) is a part of Intel since recently.

>No, if someone uses seam carving from skimage in their paper, probably noting would happen. If someone makes $200K selling a small app that uses skimage to do seam carving, there is a real risk that MERL would come in for a piece of that pie. We _could_ start offering patented algorithms with a big warning in the docstring or something, but that is not a space we want to enter, as it means (among other things) that we would have to put a big fat asterisk on our front page message.

@jni Perhaps, we should extend the text of our license then (http://scikit-image.org/docs/dev/license.html) with clarifications on permitted commercial use of the shipped implementations?

jni:
> Perhaps, we should extend the text of our license then (http://scikit-image.org/docs/dev/license.html)

No, we are [not creating a custom license](https://www.reddit.com/r/opensource/comments/31r7ml/please_dont_roll_your_own_license/). =) BSD actually covers us (as I understand it) from being held liable for anything to do with use of scikit-image, other than copyright. The patent issue is to help *users* not get sued. We could of course say we are under no obligation to do this, but that's not nice.

We *should* add a few lines to the contributor guide about patented algorithms, though.

stefanv:
All the various points have been covered above, but I want to emphasize a few of them:

- *Distributing* patented algorithms is not illegal.
- *Using* patented algorithms can be illegal, if you don't pay the patent holder.
- `scikit-image` is a collection of free (libre) algorithms.  The expectation from our users is that these algorithms may be *applied freely*, and distributing patented algorithms may therefore lead them to break the law.
- This situation can be avoided by changing the library branding, updating the license, and adding big warnings in the docstrings.  I don't have an appetite for this approach, and I don't believe the majority of the core team does either.

I am -1 on culling seam carving from 0.14.3, however; what's done is done, and we have a stronger commitment to not breaking our APIs (we are already making an exception to remove this function without following the standard deprecation cycle).  We should check that the docstring of the function currently distributed mentions the patent, and update it if not.

For preservation of the code, that may happen outside the scikit-image org.

We will soon have a *mission/vision/values* document, and I hope to renew our commitment to fully free work therein.

EngrStudent:
> Seam carving is a patented algorithm and so cannot be distributed with scikit-image, which must offer code that can be used "without charge and without restriction".
> 
> Currently, attempting to run `seam_carve` will cause a `NotImplementedError`, and then it can be removed altogether in 0.16.
> 
> Fixes #3646
> 
> ## For reviewers
> * [ ]  Check that the PR title is short, concise, and will make sense 1 year
>   later.
> * [ ]  Check that new functions are imported in corresponding `__init__.py`.
> * [ ]  Check that new features, API changes, and deprecations are mentioned in
>   `doc/release/release_dev.rst`.
> * [ ]  Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

It was published in 2007. so any patent after that is invalid because the published article constitutes prior art.
Also, it is textbook content from computational photography in the OMSCS program at GA Tech.
So the patent of the idea is invalid.

The code existed in this repo first, under the appropriate license.  That means copyright content existed before the patent.

And it is already gone, so the trolls already got what they wanted... darnit.

stefanv:
Thanks for taking another look at the issue. 

I should add that it is not a very useful algorithm in practice, and I may have added it more because I found it entertaining and a neat application of shortest paths (a Dijkstra algorithm I love!). So, all in all, I don't think it impacts our scientific mission much, even if we lost out on that bit of fun.

EngrStudent:
> Thanks for taking another look at the issue.
> 
> I should add that it is not a very useful algorithm in practice, and I may have added it more because I found it entertaining and a neat application of shortest paths (a Dijkstra algorithm I love!). So, all in all, I don't think it impacts our scientific mission much, even if we lost out on that bit of fun.

The challenge with your statement is that when it comes to innovation it is hard to understand the future utility of a potential component of something until it is used.  I agree that the method is a little awkward, a little cellular-automata-ish, but that says nothing about its potential utility.

I see it less of Dijkstra, and a little more of geospatial bootstrapping, in a sense more akin to a random forest.  ;)

Issue #3758, opened 2019-02-19
convex_hull_image only works for C-contiguous arrays


Dimchord:
## Description
The convex_hull_image function only works for C-contiguous arrays, not for Fortran-contiguous or non-contiguous arrays.

## Way to reproduce
```python
from skimage.morphology import convex_hull_image
convex_hull_image(np.ones((2,2), order='F'))
```
Output:
```python
Traceback (most recent call last):

  File "<ipython-input-47-960dea2150d3>", line 1, in <module>
    convex_hull_image(mask == 1)

  File "C:\Users\robk\Anaconda3\lib\site-packages\skimage\morphology\convex_hull.py", line 59, in convex_hull_image
    coords = possible_hull(image.astype(np.uint8))

  File "skimage\morphology\_convex_hull.pyx", line 10, in skimage.morphology._convex_hull.possible_hull

  File "stringsource", line 653, in View.MemoryView.memoryview_cwrapper

  File "stringsource", line 348, in View.MemoryView.memoryview.__cinit__

ValueError: ndarray is not C-contiguous
```

Should be easy to fix by changing the offending line in convex_hull.py to

```python
coords = possible_hull(np.ascontiguousarray(image.astype(np.uint8)))
```

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.6.5 |Anaconda custom (64-bit)| (default, Mar 29 2018, 13:32:41) [MSC v.1900 64 bit (AMD64)]
Windows-10-10.0.17763-SP0
scikit-image version: 0.14.1
numpy version: 1.14.3
```

hmaarrfk:
Do you want to submit the patch with the little test?

Dimchord:
Something like this?

[patch_convex_hull.txt](https://github.com/scikit-image/scikit-image/files/2883840/patch_convex_hull.txt)

hmaarrfk:
looks good :)

jni:
@Dimchord very nice! (And thanks for noticing!) Would you like to submit a pull request (see our [contributing guide](http://scikit-image.org/docs/dev/contribute.html)) or should we just pilfer your diff? =)

Dimchord:
I'll have a go at a pull request later today.

lagru:
[...]

PR #3762, opened 2019-02-21
Allow for f-contiguous 2D arrays in convex_hull_image

Dimchord:
## Description
This is the bug fix for issue #3758.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/convex_hull.py
+++ b/skimage/morphology/convex_hull.py
@@ -56,7 +56,7 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
     # the starting or ending pixel of a row or column.  This vastly
     # limits the number of coordinates to examine for the virtual hull.
     if ndim == 2:
-        coords = possible_hull(image.astype(np.uint8))
+        coords = possible_hull(np.ascontiguousarray(image.astype(np.uint8)))
     else:
         coords = np.transpose(np.nonzero(image))
         if offset_coordinates:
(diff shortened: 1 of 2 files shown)


skimage/morphology/convex_hull.py
```diff
@@ -56,7 +56,7 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
     # the starting or ending pixel of a row or column.  This vastly
     # limits the number of coordinates to examine for the virtual hull.
     if ndim == 2:
-        coords = possible_hull(image.astype(np.uint8))
+        coords = possible_hull(np.ascontiguousarray(image.astype(np.uint8)))
```
hmaarrfk: I think maybe you should use the dtype parameter in the ascontiguousarray
https://docs.scipy.org/doc/numpy/reference/generated/numpy.ascontiguousarray.html#numpy-ascontiguousarray

```suggestion
        coords = possible_hull(np.ascontiguousarray(image, dtype=np.uint8))
```

hmaarrfk:
Do you want to add a test for 3D images too? It seems that 2D images and 3D images go through different code paths.

jni:
@hmaarrfk the 3D code path calls out to SciPy, so it's *probably* fine?

hmaarrfk:
Yeah, I mostly don't know. I like to think that scikit-image is easy enough to mod that the next person who runs into the issue will be able to submit a small fix like this.

skimage/morphology/convex_hull.py
Dimchord: I added this suggestion, as well as some tests for 3D arrays.

hmaarrfk:
Thanks @Dimchord

PR #3768, opened 2019-02-24
Avoid NumPy warning while stacking arrays.

stefanv:
Newer versions of NumPy require that a list/tuple is passed in, and here we are using a generator.


hmaarrfk:
Travis failure is related to https://github.com/scikit-image/scikit-image/pull/3767

I'm merging.

hmaarrfk:

PR #2788, opened 2017-09-01
[WIP] Optional multichannel histograms 

maaleske:
## Description
Doing multichannel histograms by applying histogram separately for each channel is fine if you only care about the separate histograms, but comparing and visualizing histograms requires in general for them to be calculated using the same binning for each channel. Generating single-binning multichannel histograms is easiest to implement in histogram by adding a multichannel parameter which does exactly this. This would default to False (or None) which would replicate current behaviour, and when True would return a 2d array of histograms (one per channel).

## Checklist
Calculation of histograms for all the channels using
- [x] same number of bins
- [ ] the same bins (needs a way to specify either bins or range + nbins)

API changes
- [x] A new option to handle multichannel histogram generation in histogram

Documentation
- [ ] Example of multichannel=True usage in docstring
- [ ] Example in Gallery

## References
- Since single-channel images and multichannel images use the same single-channel histogram calculation, #2786 makes code branching cleaner to implement (and this PR builds on top of it).
- Implementing the common binning may requires either separate range parameter (discussed at #2479) or an explicit way to pass the bins.

## For reviewers

(Don't remove the checklist below.)

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.

PR #3772, opened 2019-02-26
Compute histograms for multichannel images

sciunto:
## Description

Superseeds #2788 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/exposure/exposure.py
```diff
@@ -120,10 +125,27 @@ def histogram(image, nbins=256, source_range='image', normalize=False):
...
+             "apply this function to each color channel, or set "
+             "multichannel=True")
+
+    if multichannel:
+        channels = sh[-1]
+        hist = []
+        bin_centers = []
+        for chan in range(channels):
+            h, bc = _histogram(image[..., chan], nbins, source_range, normalize)
+            hist.append(h)
```
emmanuelle: since the bins are the same, hist could be transformed into an array before being returned.

emmanuelle:
You were mentioning doubts about the API, would you like to discuss it here?

sciunto:
My remark is intended to make me think. I was in an automatic mode when I refactored the code. :)

Thanks for the ping btw!

skimage/exposure/exposure.py
sciunto: You're right. I'll fix that.

emmanuelle:
OK, the API looked reasonable to me. No worries, pinging is easy to do :-).

sciunto:
The API sounds effectively reasonable. I made your suggestion Emmanuelle.

skimage/exposure/exposure.py
```diff
@@ -74,16 +74,18 @@ def _bincount_histogram(image, source_range):
     return hist, bin_centers
 
 
-def histogram(image, nbins=256, source_range='image', normalize=False):
+def histogram(image, nbins=256, source_range='image', normalize=False, multichannel=False):
```
jni:
```suggestion
def histogram(image, nbins=256, source_range='image', normalize=False,
              multichannel=False):
```
(PEP8)

skimage/exposure/exposure.py
```diff
@@ -74,16 +74,18 @@ def _bincount_histogram(image, source_range):
...
     its own bin, which improves speed and intensity-resolution.
 
-    The histogram is computed on the flattened image: for color images, the
-    function should be used separately on each channel to obtain a histogram
-    for each color channel.
+    If multichannel is not set, the histogram is computed on the flattened
+    image. For color or multichannel images, the function should either be
+    the function should be either be used separately on each
+    channel to obtain a histogram for each color channel with separate binning,
+    or by setting multichannel=True to use a common binning for all channels.
```
jni: Typos above (repeated "the function should either be"). Anyway, I think in this case we should promote the use of `multichannel` first. Suggested new wording:

```
For color or multichannel images, use the `multichannel=True` keyword argument.
This will use the same histogram binning for all channels. To obtain a different binning
for each individual channel, call the function separately for each channel, keeping
`multichannel=False`.
```

skimage/exposure/exposure.py
```diff
@@ -98,6 +100,9 @@ def histogram(image, nbins=256, source_range='image', normalize=False):
         of that data type.
     normalize : bool, optional
         If True, normalize the histogram by the sum of its values.
+    multichannel : bool, optional
+        If True, calculates the histogram for all the channels using the same
```
jni:
```suggestion
        If True, calculates an individiual histogram for each channel, using a common
```

skimage/exposure/exposure.py
jni: This is not actually correct: if the different channels have different data ranges, and the source range is image, then each histogram will have different bins.

skimage/exposure/exposure.py
```diff
@@ -120,10 +125,29 @@ def histogram(image, nbins=256, source_range='image', normalize=False):
...
+    if multichannel:
+        channels = sh[-1]
+        hist = []
+        bin_centers = []
+        for chan in range(channels):
+            h, bc = _histogram(image[..., chan], nbins, source_range, normalize)
+            hist.append(h)
+        # Convert to numpy arrays
+        bin_centers = np.array(bc)
+        hist = np.array(hist)
```
jni: `bc` is already a numpy array, and hist should only be converted *after* the loop, not inside the loop.

jni (dismissed):
@sciunto this will be a great usability enhancement, but I think right now the implementation is incorrect (though do correct me if I'm wrong about that). I think currently nothing enforces that the bins should be the same between channels, other than luck.

skimage/exposure/exposure.py
sciunto: looks outside...

sciunto:
It's actually half correct -- only if source_range=='dtype'

skimage/exposure/exposure.py
jni: Oops! The comment messed with my indentation detector! :sweat_smile:

skimage/exposure/tests/test_exposure.py
```diff
@@ -115,6 +115,23 @@ def test_normalize():
...
 
+# Test multichannel histograms
+# ============================
+
+def test_multichannel_hist_common_bins_uint8():
+    """Check that all channels use the same binning."""
+    # First channel contains zeros
+    # Second channel contains 255
+    im = np.array((np.zeros((5, 5)), np.ones((5, 5))*255), dtype=np.uint8)
+    im = np.moveaxis(im, source=0, destination=2)
```
jni: WOW, `moveaxis` is so much more intuitive when you include the keyword arguments! :clap:

jni:
@sciunto I think the correct thing here is to compute the bins on the raveled image, then use those bins as input to the individual channel histograms. I'm hoping passing manual bins works for our histogram function!

skimage/exposure/tests/test_exposure.py
hmaarrfk: I'm quoting numpy:
> save those O(100) nanoseconds!
> umr_maximum = um.maximum.reduce

https://github.com/numpy/numpy/blob/master/numpy/core/_methods.py#L16

emmanuelle:
I agree with @jni's comment (compute the bins on the raveled image, which will ensure to have the same bins for all channels). @sciunto do you have the time to work on this or would you prefer me to take over?

sciunto:
Please, go ahead @emmanuelle

grlee77:
I volunteer to resume this PR (assuming @emmanuelle has little free time to get to this in the immediate future)

skimage/exposure/exposure.py
grlee77: I updated the text here to recommend `channel_axis` first (I replaced `multichannel` with `channel_axis` in this PR since we are deprecating multichannel elsewhere)

grlee77:
@jni: this is ready for another round of review. I implemented your suggestion to compute the bins upfront on the raveled array in the multichannel case.

Otherwise, I adapted this to use `channel_axis` instead of `multichannel` and expanded the test cases a bit.

grlee77 (approved):
I think this looks good now, but it could use a second review of the new code I added.

sciunto:
Thanks @grlee77 to make it happen! :)

skimage/exposure/exposure.py
grlee77: This was resolved by passing `bins` instead of `nbins` in ec7edf36ef467d3a398673b4000060d724566d33

PR #3776, opened 2019-03-01
Add image enhancement measure (EME)

iliailmer:
## Description

This PR contains implementation of an image quality measure called EME, based on [this paper](https://www.semanticscholar.org/paper/A-New-Measure-of-Image-Enhancement-Agaian-Lentz/5adac5932775f089eaace7ebc45a6cba89809134).

It is based no the human visual system, as mentioned in the paper, and it can be used as a metric for enhancement algorithms to see if they work or not.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


grlee77:
Hi @iliailmer,

Thank you for the contribution! I have not yet done a detailed review of the publication, but it has a substantial number of citations and the implementation looks simple and should be easy to maintain going forward. I think this is a reasonable candidate for inclusion in scikit-image.

I can do a more thorough review over the coming week, but a few initial comments:

1.) The function should not be stored in the top-level `skimage` directory.  A more appropriate location would be in the `metrics` submodule, perhaps as an addition to the current [simple_metrics.py](https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/simple_metrics.py)

2.) Use a more descriptive name for the function than `quality`. Perhaps `enhancement`, `enhancement_measure` or something similarly descriptive. For example the existing functions `compare_psnr` and `compare_ssim` are measures of image similarity often referred to as image quality metrics. Having PSNR or SSIM in the function name gives some immediate indication of what "quality" is being measured.

3.) It looks like this can be trivially extended to the n-dimensional case rather than just 2d and 3d. We can help with this if you have questions on how to approach it. A convention in the library is to include a `multichannel` argument so that the user can indicate whether the final axis of an array is a "spatial" axis or whether it represents channels (e.g R, G, B channels for a color image, for example).

iliailmer:
Hi @grlee77 , thanks for your comments!

Regarding the multidimensional extension, I realize now that the max/min filters used here are, by default, usable with any number of dimensions, so one does not need to check the length of the array. In this case the `multidimensional` parameter may not be required, and the filter shape (`footprint`) will be inferred automatically.

skimage/image_quality_measure.py
```diff
@@ -0,0 +1,53 @@
...
+        --------
+        >>> from skimage.data import camera
+        >>> from skimage.exposure import equalize_hist
+        >>> img = camera()
+        >>> print("Image quality:\n")
+            Image quality:
+        >>> print(f"\tbefore: {quality(img)}")
+                before: 0.9096745071475523
+        >>> print(f"\tafter: {quality(equalize_hist(img))}")
+                after: 1.299327371881219
```
lagru: The docstring formatting is a little unconventional which may be the reason for the doctest failure on travis. What is you reasoning for using print functions? I think the indentation after the print statements confuses the parser. You could use comments instead to annotate "before" and "after". E.g.:

```python
>>> quality(img)  # before
0.9096745071475523
>>> quality(equalize_hist(img))  # after
1.299327371881219
```

And just a nitpick: could you remove one indentation level from the docstring? It's not really necessary.

lagru:
Looks promising! I'll have a more thorough look later.

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,50 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+            Image quality:
+        >>> print(f"\tbefore: {quality(img)}")
+                before: 0.9096745071475523
+        >>> print(f"\tafter: {quality(equalize_hist(img))}")
+                after: 1.299327371881219
+
+    """
+    image = img_as_float(image)
+    eme = np.zeros_like(image)
+    footprint = tuple([size] * len(image.shape))
```
lagru: You don't need to construct the `footprint` here. You can just pass `size` directly to the filter functions as an argument of the same name.

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,50 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+    eme = np.divide(maximum_filter(image, footprint=footprint) + 1,
+                    minimum_filter(image, footprint=footprint) + 1)
```
lagru: Is adding `+1` here supported by the linked paper? Why is this necessary? Do you want to avoid division by 0 this way?

skimage/image_quality_measure.py
lagru: Oh, I commented in the wrong file. But it applies to the other one as well.

lagru (commented):
Thanks for addressing @grlee77's feedback. Good work so far! With the recent changes this should work for the n-dimensional case, perhaps we could mention that in the docstring. I think you forgot to remove the now obsolete file `image_quality_measure`.  Could you fix that?

Otherwise I don't yet fully understand your implementation when looking at the linked paper (see comment). Do you or someone else know a source for the clock image used in the paper? It would be useful to compare results of this function to the paper.

skimage/image_quality_measure.py
iliailmer: Thanks for the comments @lagru, I did not really think about why I put the print statements there. Makes sense to remove them in this contexts, true.

skimage/measure/simple_metrics.py
iliailmer: So, in the paper the authors don't add anything, but division by zero is a possibility here, so to avoid it I add +1 to both numerator and denominator. At least in my experiments I used the function in this form a lot. It might make sense to use something like an `eps = 1` as a parameter passed to the function. I will check which values make sense.
Should the implementation follow the same comparison pattern as with PSNE and other functions? `compare_enhancement` or something like that?

iliailmer: Sorry, forgot to address the question about the image. I don't personally have it, but google search gave me [this link](https://link.springer.com/chapter/10.1007%2F978-3-642-23154-4_11) to the source paper and [the image](https://www.google.com/search?sa=G&hl=en&q=monochrome&tbm=isch&source=iu&ictx=1&tbs=simg:CAESvgIJt1bAc_1cJHhMasgILEKjU2AQaBAgUCAoMCxCwjKcIGmEKXwgDEieVAYESlwH6AYMS7AV_1qgbMAvsBiz7wJ6c27if2NaU29DbyJ6I2kz4aMBrLMEAUP2pMseBtDd5bfccLZkF_1HpcsD1s4kW6GfqcgdI3SV3g-nobzAkj7i1DbYSAEDAsQjq7-CBoKCggIARIE6a_1fCgwLEJ3twQkangEKHQoKbW9ub2Nocm9tZdqliPYDCwoJL20vMDFtd2tmCh4KC2FsYXJtIGNsb2Nr2qWI9gMLCgkvbS8wNDZkbHIKGgoHYW50aXF1ZdqliPYDCwoJL20vMDF5czJzCiIKD2JsYWNrLWFuZC13aGl0ZdqliPYDCwoJL20vMDFnNmdzCh0KCnN0aWxsIGxpZmXapYj2AwsKCS9tLzAxc3B6cww&fir=VQ-TGHm2qMMWYM%253A%252CvLAo2pnVeI90MM%252C_&vet=1&usg=AI4_-kQx9DbbxFxkhS3ywjL0w-seNQZmRQ&ved=2ahUKEwjR3435g_TgAhUJVN8KHRurCmoQ9QEwAnoECAAQBA#imgrc=VQ-TGHm2qMMWYM:). Should the image be added to tests here?

lagru:
@iliailmer 

> I think you forgot to remove the now obsolete file image_quality_measure. Could you fix that?

I'm getting confused now. The diff on GitHub shows that this PR still includes 2 redundant files with nearly the same code.  Could you ensure that the redundant files are removed? We only want to keep the changes in `skimage/measure/simple_metrics.py` as suggested by @grlee77.

skimage/measure/simple_metrics.py
lagru: I'm not sure that simply adding +1 is the right way to go. According to [this guide](http://scikit-image.org/docs/stable/user_guide/data_types.html) images of dtype float may also range from -1 to 1. If we pass in an image with a -1 somewhere adding +1 doesn't work.

No, adding the image to the tests shouldn't be necessary. If you want to use a natural image there you can use the [ones already shipped with scikit-image](http://scikit-image.org/docs/stable/api/skimage.data.html).

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +146,47 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+        >>> quality(img) # before
+            0.9096745071475523
+        >>> quality(equalize_hist(img))
+            1.299327371881219
+
+    """
+    image = img_as_float(image)
+    eme = np.zeros_like(image)
+    eme = np.divide(maximum_filter(image, tuple([size]*len(image.shape))) + 1,
+                    minimum_filter(image, tuple([size]*len(image.shape))) + 1)
```
lagru: Sorry, what I meant earlier is that [`maximum_filter`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.maximum_filter.html) already supports the parameter `size` directly. E.g.

```python
eme = np.divide(maximum_filter(image, size=size)) + 1,
                minimum_filter(image, size=size)) + 1)
```
Or am I reading that wrong?

iliailmer: I think you're right. I was using the `footprint` parameter, because I got the impression from the docs that it is necessary for proper shape of the filter. I tested with just using `size' and it seems to work as well.

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +147,51 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+        .. [1] Agaian, Sos S., Karen Panetta, and Artyom M. Grigoryan.
+               "A new measure of image enhancement."
+               IASTED International Conference on Signal Processing
+               & Communication. Citeseer, 2000.
+        Examples
+        --------
+        >>> from skimage.data import camera
+        >>> from skimage.exposure import equalize_hist
+        >>> img = camera()
+        >>> quality(img) # before
```
lagru: `quality` is no longer the correct function name which is the reason for the failures on travis.

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +147,51 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+            0.9096745071475523
+        >>> quality(equalize_hist(img))
+            1.299327371881219
+
+    """
+
+    image = img_as_float(image)
+    image = rescale_intensity(image, out_range=(0., 1.))
+    eme = np.zeros_like(image)
```
lagru: I think this line is not needed.

lagru (commented):
@iliailmer thanks for addressing my feedback. The solution with `eps` as an optional parameter is much better now but I'm not sure whether this is the right way to solve this. As I understand the paper the formula is meant to quantify a contrast improvement somehow. While rescaling the image won't affect the [Weber contrast](https://en.wikipedia.org/wiki/Contrast_(vision)#Weber_contrast) I'm unsure about this measurement. 

Is there a reason to rescale between [0, 1] in particular? If we were to chose a value above zero as the lower bound we could omit the `eps` parameter altogether.

skimage/measure/simple_metrics.py
iliailmer: > Is there a reason to rescale between [0, 1] in particular? If we were to chose a value above zero as the lower bound we could omit the `eps` parameter altogether.

Thanks for the feedback, I will get to all the stuff mentioned above as soon as I can.
Meanwhile, regarding rescaling, when I was testing images with range `[-1,1]` I ran into issues with logarithms, since some pixel neighborhoods can have multiple negative values, the logarithm would try to evaluate negatives and return `nan`. Hence the rescaling to `[0,1]`.

emmanuelle:
Hi @iliailmer thanks for your PR. It looks good, but we're missing unit tests (write unit tests on small images, 2D and 3D, with int and float types for example) as well as a gallery example.

iliailmer:
@emmanuelle Thank you, I will get to it as soon as I can. Should the test images come from the `data` module of scikit-image?

emmanuelle:
@iliailmer yes, either images from ``skimage.data`` (you can downscale them for faster execution, or choose a ROI, since we want tests to execute fast enough) or synthetic data which you create yourself. You can take a look at the different ``tests`` folders in each submodule of scikit-image, to get an idea of how tests are written.

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,48 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+    """ The image enhancement measure called EME based on [1]_.
+
+        Parameters
+        ----------
+        image : array
+            Input image of which the quality should be assessed.
+            Can be either 3-channel RGB or 1-channel grayscale.
+        size : int
+            Size of the window.
+        eps : float
```
emmanuelle: Do you need ``eps`` as a parameter or could you choose a value inside the function once and for all?

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,48 @@ def compare_psnr(im_true, im_test, data_range=None):
 
     err = compare_mse(im_true, im_test)
     return 10 * np.log10((data_range ** 2) / err)
+
+
+def enhancement_measure(image: np.ndarray,
+                        eps: float = 1e-6,
+                        size: int = 3) -> float:
+    """ The image enhancement measure called EME based on [1]_.
```
emmanuelle: Add one sentence to explain what EME is measuring, like "EME is a measure of average local contrast, measured on a logarithmic scale inside windows of user-provided size".

emmanuelle:
Thanks for the update! I left two more minor comments, and we would also need a gallery example :-).

skimage/measure/simple_metrics.py
iliailmer: I think it is better to let the user control the `eps` parameter because if the potential image/matrix is sparse. A larger `eps` could help to avoid unnecessary negative values.

skimage/measure/simple_metrics.py
iliailmer: I added more explanations to the doc files, thanks for pointing that out.

doc/examples/filters/plot_enhancemet_measure.py
```diff
@@ -0,0 +1,65 @@
+"""
```
emmanuelle: There is a typo in the file name enhancemet --> enhancement

doc/examples/filters/plot_enhancemet_measure.py
```diff
@@ -0,0 +1,65 @@
...
+
+It is defined as a log of the ratio of local maximum to local minimum
+within a sliding window. The code allows the user to choose the window size.
+This function was defined to quantify improvement of the image after processing.
+
+"""
+
+import numpy as np
+import matplotlib
+matplotlib.use("TkAgg")
```
emmanuelle: These two lines above are not needed and should be removed.

emmanuelle (commented):
Thanks, we're really getting there. Two more very minor things to correct in the example; plus there is some weird doc build failure related to the example, I'll have to investigate more to understand what's going on. @scikit-image/core we'll soon need a second review.

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,52 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+
+        The function uses a sliding window of user-provided size to measure
+        the mean of log of maximal and minimal intensity ratio within the window.
+
+        Parameters
+        ----------
+        image : array
+            Input image of which the quality should be assessed.
+            Can be either 3-channel RGB or 1-channel grayscale.
+        size : int
```
sciunto:
```suggestion
        size : int, optional
```

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,52 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+            Size of the window.
+        eps : float
```
sciunto:
```suggestion
        eps : float, optional
```

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,52 @@ def compare_psnr(im_true, im_test, data_range=None):
 
     err = compare_mse(im_true, im_test)
     return 10 * np.log10((data_range ** 2) / err)
+
+
+def enhancement_measure(image: np.ndarray,
+                        eps: float = 1e-6,
+                        size: int = 3) -> float:
```
sciunto: Please, keep a consistent arg order in the signature and the docstring.

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,52 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+        -------
+        eme : float
+            The number describing image quality.
+
+        References
+        ----------
+        .. [1] Agaian, Sos S., Karen Panetta, and Artyom M. Grigoryan.
+               "A new measure of image enhancement."
+               IASTED International Conference on Signal Processing
+               & Communication. Citeseer, 2000.
```
sciunto: Please, add a DOI with the format (auto formatting):

:DOI:`xxx`

doc/examples/filters/plot_enhancemet_measure.py
```diff
@@ -0,0 +1,65 @@
...
+# histogram equalization with `256*3` bins. Histogram equalization
+# brought out details and the EME value is higher than before the transform.
+
+############################################################################
+# References
+# ----------
+# .. [1] Agaian, Sos S., Karen Panetta, and Artyom M. Grigoryan.
+# "A new measure of image enhancement."
+# IASTED International Conference on Signal Processing
+# & Communication. Citeseer, 2000.
```
sciunto: Please, add the DOI here as well.

doc/examples/filters/plot_enhancemet_measure.py
iliailmer: Thank you for your comments! I fixed the issues.

skimage/measure/simple_metrics.py
iliailmer: @sciunto Thank you for pointing that out, I fixed it.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
+"""
+=========================================
+Quantitative Measure of Image Enhancement
+=========================================
+
+In this example, we will measure the result of applying contrast enhancement
```
stefanv:
```suggestion
In this example, we will quantify the effect of applying contrast enhancement
```

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
...
+to the image. We will use a measure called EME [1]_.
```
stefanv:
```suggestion
to an image using a measure called EME [1]_.
```

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
...
+
+It is defined as a log of the ratio of local maximum to local minimum
```
stefanv:
```suggestion
It is defined as the log of the ratio of local maximum to minimum
```

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
...
+within a sliding window. The code allows the user to choose the window size.
```
stefanv:
```suggestion
within a sliding window of given size.
```

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
...
+This function was defined to quantify improvement of the image after processing.
```
stefanv:
```suggestion
This measure aims to provide a perceptual estimate of image quality, useful when comparing image enhancement algorithms.
```

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
...
+
+"""
+
+import numpy as np
+import matplotlib.pyplot as plt
+from skimage import data
+from skimage.filters import gaussian
+from skimage.exposure import equalize_adapthist
+from skimage.measure.simple_metrics import enhancement_measure
```
stefanv: We don't typically do nested level imports in scikit-image.  Please import directly from `skimage.measure`.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+img_gray = data.camera()
+img_rgb = data.astronaut()
+
+
+def plotter(image_1: np.ndarray, image_2: np.ndarray):
```
stefanv: We need a better name than `plotter` here: `compare_images` or `side_by_side_display` or something.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
...
+# We will create a function that takes an original image as `before`
+# and applies a `transform` to it. We will use a gaussian filter to
+# strongly blur the image and an adaptive histogram equalization as
+# transforms.
+
+def compare_eme(before: np.ndarray, transform, kwargs=None):
+    after = transform(before, **kwargs)
+    plotter(before, after)
+
+compare_eme(before=img_gray, transform=gaussian, kwargs={'sigma':10})
```
stefanv: This is only run twice, so probably overkill.  How about simply:

```
compare(image, gaussian(image, sigma=10))
compare(image, equalize_adapthist(image, nbins=256 * 3))
```

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,64 @@
...
+# differences and, as a result, to lower value of EME.
+# On the other hand, the `astronaut` image was enhanced by adaptive
+# histogram equalization with `256*3` bins. Histogram equalization
+# brought out details and the EME value is higher than before the transform.
+
+############################################################################
+# References
+# ----------
+# .. [1] Agaian, Sos S., Karen Panetta, and Artyom M. Grigoryan.
+# "A new measure of image enhancement."
```
stefanv: This reference does not seem to be correctly formatted; needs some indentation?

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,53 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+               IASTED International Conference on Signal Processing
+               & Communication. Citeseer, 2000,
+               :DOI:10.1.1.35.4021
+
+        Examples
+        --------
+        >>> from skimage.data import camera
+        >>> from skimage.exposure import equalize_hist
+        >>> img = camera()
+        >>> before = enhancement_measure(img) # before
```
stefanv:
```suggestion
        >>> before = enhancement_measure(img)  # before
```

skimage/measure/tests/test_simple_metrics.py
```diff
@@ -75,3 +74,23 @@ def test_NRMSE_errors():
     # invalid normalization name
     with testing.raises(ValueError):
         compare_nrmse(x, x, 'foo')
+
+
+def test_EME_greyscale():
```
stefanv: Does the original paper provide any absolute values for given images that we can try and reproduce?  Only checking whether the value goes down after enhancement is not quite strict enough.

stefanv (changes requested):
Thanks for your contribution!  I've added a few small suggestions.

skimage/measure/tests/test_simple_metrics.py
lagru: This came up earlier (see [here](https://github.com/scikit-image/scikit-image/pull/3776#discussion_r263983133)). The image in question could be a good candidate for comparison. Unfortunately its source isn't mentioned inside the paper and the one found through searching may not be *exactly* the same.

Another issue is that the absolute value may change depending on `eps`. That parameter can help to avoid division by zero but is an addition of this PR and not part of the paper.

lagru: Furthermore the rescaling between 0 and 1 isn't part of the paper either, so I doubt we'd get comparable absolute values even if we used the same image as in the paper.

stefanv: That's fine; I was hoping for some rough agreement between their work and ours, otherwise how do we know we're even in the right ballpark?  Overall, any test that we can define that is a bit more strict will be helpful.

iliailmer: @stefanv @lagru I looked at the experiment section of the paper, I managed to find the moon surface image [here](https://homepages.inf.ed.ac.uk/rbf/HIPR2/images/moo1.gif), running the code with that image without converting to float and (0,1)-rescaling, I can get the value to be ~6.67 if the `size` is 6 and `eps` is `1e-8`. The same parameters yield ~6.28 when converted to float and rescaled.

stefanv: Does this correspond to the values in the paper?

iliailmer: Yes, the paper lists 6.78 on page 4, Figure 1b.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,69 @@
+"""
+=========================================
+Quantitative Measure of Image Enhancement
+=========================================
+
+In this example, we will quantify the effect of applying contrast enhancement
+to an image using a measure called EME [1]_.
+
+It is defined as the log of the ratio of local maximum to minimum
+<<<<<<< HEAD
```
stefanv: woops.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,69 @@
...
+within a sliding window. The code allows the user to choose the window size.
```
stefanv:
```suggestion
within a sliding window of given size.
```

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,69 @@
...
+This function was defined to quantify improvement of the image
+after processing.
+=======
```
stefanv: Another artifact.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,69 @@
...
+Quantitative Measure of Image Enhancement
+=========================================
+
+In this example, we will quantify the effect of applying contrast enhancement
+to an image using a measure called EME [1]_.
+
+It is defined as the log of the ratio of local maximum to minimum
+<<<<<<< HEAD
+within a sliding window. The code allows the user to choose the window size.
+This function was defined to quantify improvement of the image
```
stefanv: How about: This method was developed to quantify improvement of the image after processing.

skimage/measure/simple_metrics.py
```diff
@@ -142,3 +145,53 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+               IASTED International Conference on Signal Processing
+               & Communication, Citeseer, 2000,
+               :DOI:10.1.1.35.4021
+
+        Examples
+        --------
+        >>> from skimage.data import camera
+        >>> from skimage.exposure import equalize_hist
+        >>> img = camera()
+        >>> before = enhancement_measure(img)  # before
```
stefanv:
```suggestion
        >>> before = enhancement_measure(img)
```

skimage/measure/tests/test_simple_metrics.py
stefanv: OK, so we can tighten up one of the tests to match against a specific value.

Why are the values different depending on the dtype and scaling?

iliailmer: I need to think some more about it, right now I am not sure why this is the case.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,61 @@
...
+# differences and, as a result, to lower value of EME.
+# On the other hand, the `astronaut` image was enhanced by adaptive
+# histogram equalization with `256*3` bins. Histogram equalization
+# brought out details and the EME value is higher than before the transform.
+
+############################################################################
+# References
+# ----------
+# .. [1] Sos S. Agaian, Karen Panetta, and Artyom M. Grigoryan.
+# "A new measure of image enhancement.",
```
lagru: I'm guessing that the failures on Travis are due to the missing indent here and the three lines below. Adding 3 spaces should fix this.

sciunto:
I mark this PR to 0.17 as a targeted milestone to get time for finishing this PR. Some conflicts are due to a recent code refactoring.

doc/examples/filters/plot_enhancement_measure.py
iliailmer: I added the corrections, thanks for pointing that out

iliailmer:
I keep getting 
`Warning, treated as error:
C:\hostedtoolcache\windows\Python\3.6.8\x64\lib\site-packages\skimage\measure\simple_metrics.py:docstring of skimage.measure.compare_mse:23:Explicit markup ends without a blank line; unexpected unindent.
`
`
mingw32-make[1]: *** [Makefile:53: html] Error 2
`
`
mingw32-make[1]: Leaving directory 'd:/a/1/s/doc'
`
`
mingw32-make: *** [Makefile:27: html] Error 2
`

Tried to fix docstring couple of times already, not sure where the unindent is happening

rfezzani:
>Tried to fix docstring couple of times already, not sure where the unindent is happening

Have you tried the solution proposed [here](https://stackoverflow.com/questions/38761546/markup-ending-without-empty-line-but-markup-is-only-thing-in-comment)?

iliailmer:
@rfezzani Thank you, it worked.

skimage/measure/simple_metrics.py
```diff
@@ -71,12 +78,63 @@ def compare_psnr(im_true, im_test, data_range=None):
...
+    eme : float
+        The number describing image quality.
+
+    References
+    ----------
+    .. [1] Sos S. Agaian,  Karen Panetta, and Artyom M. Grigoryan.
+            "A new measure of image enhancement.",
+            IASTED International Conference on Signal Processing
+            & Communication, Citeseer, 2000,
+            :DOI:10.1.1.35.4021
```
sciunto:
```suggestion
            :DOI:`10.1.1.35.4021`
```

sciunto: The DOI is incorrect. Also, backquotes must be set as suggested.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,61 @@
...
+# brought out details and the EME value is higher than before the transform.
+#
+# References
+# ----------
+#
+# .. [1] Sos S. Agaian, Karen Panetta, and Artyom M. Grigoryan.
+#        "A new measure of image enhancement.",
+#        IASTED International Conference on Signal Processing
+#        & Communication, Citeseer, 2000,
+#        :DOI:10.1.1.35.4021
```
sciunto: Same here.

skimage/measure/simple_metrics.py
iliailmer: Hi @sciunto, thanks for pointing that our!
I was having troubles with the DOI though.

The only source that lists the DOI is [this](https://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.35.4021) link, and I doubt that there is a better version or at least i couldn't find one. The [doi.org](doi.org) says this is only a prefix, but google search for doi=10.1.1.35.4021 returns the link above no problem. The original journal is not available (see [here](https://www.iasted.org/CONFERENCES/pastinfo-320.html)) so I am not sure what is the best way to prcoceed here.

sciunto: I think doi.org must be the reference. If no working doi can be found, it is better to do not display a doi. Eventually, a link to the pdf file can be added instead.

iliailmer: I removed it for now.

iliailmer:
> Thanks for your contribution! I've added a few small suggestions.

Hi @stefanv, could you please take a look and check if anything is amiss? All the checks are passed so far. Thanks!

lagru:
Hey @iliailmer, sorry for the slow review process. I still want to get this into scikit-image! There are some merge conflicts now, do you want to resolve these or shall I? Afterwards I can do another review and hopefully approve to get this going again. :pray:

lagru:
Hey, I took the liberty to merge this which was a little complicated due to us moving `simple_metrics` from `skimage/measure` to `skimage/metrics` on `main`. I hope I got every relevant change pertaining to adding the new function `enhancement_measure`.

skimage/metrics/simple_metrics.py
```diff
@@ -259,3 +263,53 @@ def normalized_mutual_information(image0, image1, *, bins=100):
...
+
+    Examples
+    --------
+    >>> from skimage.data import camera
+    >>> from skimage.exposure import equalize_hist
+    >>> img = camera()
+    >>> before = enhancement_measure(img)
+    >>> after = enhancement_measure(equalize_hist(img))
+    >>> before < after
+    True
```
lagru, line 306: Other measure functions in this module are applied on and compare two images. As far as I understand, the EME of an image is not useful on its own but rather it's difference to the EME of an improved image. It might be worth thinking about making the function take two images and return a bool or difference between the EME of both images. That way we also prevent users from assigning to much meaning to the absolute value of the measure.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,61 @@
...
+=========================================
+In this example, we will quantify the effect of applying contrast enhancement
+to an image using a measure called EME [1]_.
+
+It is defined as the log of the ratio of local maximum to minimum within a
+sliding window of given size. This measure aims to provide a perceptual
+estimate of image quality, useful when comparing image enhancement algorithms.
+
+"""
+import numpy as np
```
lagru, line 13:
```suggestion
```
Not used.

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,61 @@
...
+        ax[idx].imshow(image)
+        ax[idx].axis('off')
+        ax[idx].set_title("EME = {:.2f} ".format(
+            simple_metrics.enhancement_measure(image, size=3)))
+    fig.show()
+
+
+#####################################################################
+# We will create a function that takes an original image as `before`
+# and applies a `transform` to it. We will use a gaussian filter to
```
lagru, line 36: Nitpick: the function doesn't actually apply a transform to it. It just calls `enhancement_measure` on both and plots the results. Rewording this might be a lot clearer to new users.

skimage/metrics/simple_metrics.py
```diff
@@ -259,3 +263,51 @@ def normalized_mutual_information(image0, image1, *, bins=100):
...
+    The enhancement measure (EME) of an image, aims to provide a perceptual
+    estimate of image quality, based on the luminance ratio within a sliding
+    window [1]_. The resulting measure is useful when comparing image
+    enhancement algorithms.
+
+    Parameters
+    ----------
+    image : array
+        Input image of which the quality should be assessed.
+        Can be either 3-channel RGB or 1-channel grayscale.
```
lagru, line 280: We should probably support only grayscale images as input to make sure that the user determines [how the total luminance of a colored image is calculated](https://scikit-image.org/docs/stable/api/skimage.color.html#skimage.color.rgb2gray). Currently, the way that `maximum_filter` and `minimum_filter` is combining color channels is not something that is covered by the paper.

skimage/metrics/simple_metrics.py
```diff
@@ -259,3 +263,51 @@ def normalized_mutual_information(image0, image1, *, bins=100):
     H01 = entropy(np.reshape(hist, -1))
 
     return (H0 + H1) / H01
+
+
+def enhancement_measure(image, size=3, eps=1e-6):
```
lagru, line 268: Is there some reasoning between choosing `eps=1e-6` as the default? Is something else more reasonable?

doc/examples/filters/plot_enhancement_measure.py
```diff
@@ -0,0 +1,60 @@
...
+from skimage.filters import gaussian
+from skimage.exposure import equalize_adapthist
+from skimage.metrics import simple_metrics
+
+img_gray = data.camera()
+img_rgb = data.astronaut()
+
+
+def compare_side_by_side(before, after):
+    fig, ax = plt.subplots(nrows=1, ncols=2, figsize=(16, 14))
```
lagru, line 25:
```suggestion
    fig, ax = plt.subplots(nrows=1, ncols=2)
```
I think removing the forced figure size leads to a nicer layout. Otherwise the images seem very small.

skimage/metrics/tests/test_simple_metrics.py
```diff
@@ -137,3 +140,23 @@ def test_nmi_random_3d():
...
+
+def test_EME_greyscale():
+    orig = cam
+    enhanced = equalize_hist(orig)
+    assert_less(enhancement_measure(orig),
+                enhancement_measure(enhanced))
+    assert_less(enhancement_measure(orig, size=5),
+                enhancement_measure(enhanced, size=5))
+    assert_less(enhancement_measure(orig, size=11),
+                enhancement_measure(enhanced, size=11))
```
lagru, line 153: We should probably check the actual values of the EMEs and that they are within an expected range. That way our tests would inform us if some change leads to significantly different EMEs.

lagru (commented):
I took the liberty and tried to make the docstring use a more precise description of the EME. However, I'm still a bit vague on how to EME should be interpreted precisely and what kind of "enhancement" it measures. Does it purely measure contrast enhancement? To me that seems to me where we can gain the most value yet.

Otherwise I suggested some smaller but easier changes that should probably be addressed.

lagru:
Hey @iliailmer, would you be able to look into the feedback? This seems like something that would be cool to have in scikit-image. It's also totally fine if you don't have the bandwidth in which case maybe someone else could take over. :)

matthew-brett:

<details>

## Rationale

</details>

PR #3781, opened 2019-03-04
Backport: Reduce default tolerance in threshold_li (#3622)

### LINKED ISSUE #3605: threshold_li changed from version 0.14.0 to version 0.14.1

jni:
* Reduce the default tolerance in threshold_li

Fixes #3605

In that issue, I found that a tolerance of `range / 2**15` was
sufficient. However, the only thing that guarantees convergence is the
smallest intensity difference in the image (it's impossible for the
update to change if the threshold doesn't change). I've set this as the
new default, as the only downside is that it might take a long time to
converge, but that seems preferable to finding an incorrect threshold.

* Fix documentation of tolerance in threshold_li

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1,3 +1,4 @@
+from __future__ import division, print_function
 import itertools
 import math
 import numpy as np
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -529,11 +529,6 @@ def threshold_li(image):
     image : ndarray
         Input image.
 
-    tolerance : float, optional
-        Finish the computation when the change in the threshold in an iteration
-        is less than this value. By default, this is half of the range of the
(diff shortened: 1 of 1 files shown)


soupault:
@jni tests are not OK.

jni:
@soupault Why is life always hard? =P

Since the failures were in 2.7, I presume it's a float issue, but I haven't hunted it down. Let's see if this new import fixes things...!

jni:
Booyah!

PR #3791, opened 2019-03-07
Unify LICENSE files for easier interpretation

stefanv:


hmaarrfk:
I feel like the seperate files hsould be kept in addition to having them listed in the main license document. no?

stefanv:
No, there isn't a "should" in this case.  Debian, for their packaging, already centralizes the licenses into one file.  The problem with the current way of doing things is that users have to do a search through the repo to get a full picture of the license; now they can look and see, in one place, that it is all BSD.

We should add the LICENSE.txt file to our sdist package; I don't think it is currently included, which I think is in breach of the BSD license.

hmaarrfk:
oh you mean we should include all the different license files in the .txts. yeah that is probably necessary.

stefanv:
No, I mean we should include our single LICENSE.txt in the sdist source archive.

LICENSE.txt
```diff
@@ -29,3 +26,120 @@ HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
...
+    ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+    LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+    CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+    SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+    INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+    CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+    ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+    POSSIBILITY OF SUCH DAMAGE.
+
+skimage/filters/thresholding.py:threshold_li
```
jni: I'm not sure how long licenses should stick around, but I essentially reimplemented Li from the original paper in #3402.

stefanv: Fantastic, then that can be removed.

jni:

PR #3798, opened 2019-03-09
Use float32 when warping when provided by the user.

hmaarrfk:
## Description

I don't want to enable integer warping right away, but this is at least a good start.

I probably need to write a failer for iradon
xref: https://github.com/scikit-image/scikit-image/pull/3486

```python
pytest skimage/transform/tests/test_warps.py                                  
============================= test session starts ==============================
platform linux -- Python 3.7.1, pytest-4.3.0, py-1.8.0, pluggy-0.9.0
rootdir: /home/mark/git/scikit-image/skimage/transform, inifile:
plugins: xdist-1.26.1, localserver-0.5.0, forked-1.0.2, faulthandler-1.5.0
collected 41 items                                                             

skimage/transform/tests/test_warps.py .................................. [ 82%]
.....F.                                                                  [100%]

__________________________ test_warp_floats[float32] ___________________________

dtype = <class 'numpy.float32'>

    @parametrize('dtype', [np.float32, np.float64])
    def test_warp_floats(dtype):
        x = np.zeros((5, 5), dtype=dtype)
        tform = SimilarityTransform(scale=1)
    
        x_tform = warp(x, tform, order=1)
>       assert x_tform.dtype == x.dtype
E       AssertionError: assert dtype('float64') == dtype('float32')
E        +  where dtype('float64') = array([[ 0.,  0.,  0.,  0.,  0.],\n       [ 0.,  0.,  0.,  0.,  0.],\n       [ 0.,  0.,  0.,  0.,  0.],\n       [ 0.,  0.,  0.,  0.,  0.],\n       [ 0.,  0.,  0.,  0.,  0.]]).dtype
E        +  and   dtype('float32') = array([[ 0.,  0.,  0.,  0.,  0.],\n       [ 0.,  0.,  0.,  0.,  0.],\n       [ 0.,  0.,  0.,  0.,  0.],\n       [ 0.,  0.,  0.,  0.,  0.],\n       [ 0.,  0.,  0.,  0.,  0.]], dtype=float32).dtype

skimage/transform/tests/test_warps.py:533: AssertionError
===================== 1 failed, 40 passed in 0.29 seconds ======
```

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_transform_warp.py
+++ b/benchmarks/benchmark_transform_warp.py
@@ -7,7 +7,7 @@ import inspect
 
 
 class WarpSuite:
-    params = ([np.uint8, np.uint16, np.float32, np.float64],
+    params = ([np.float32, np.float64],
               [128, 1024, 4096],
               [0, 1, 3],
               # [np.float32, np.float64]
@@ -20,24 +20,20 @@ class WarpSuite:
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", "Possible precision loss")
             self.image = convert(np.random.random((N, N)), dtype=dtype_in)
-        self.tform = SimilarityTransform(scale=1, rotation=np.pi / 10,
-                                         translation=(0, 4))
-        self.tform.params = self.tform.params.astype('float32')
-        self.order = order
+        tform = SimilarityTransform(scale=1, rotation=np.pi / 10,
+                                    translation=(0, 4))
+        tform.params = tform.params.astype('float32')
+        order = order
 
-        if 'dtype' in inspect.signature(warp).parameters:
-            self.warp = functools.partial(warp, dtype=self.image.dtype)
-        else:
-            # Keep a call to functools to have the same number of python
-            # function calls
-            self.warp = functools.partial(warp)
+        self.warp = functools.partial(
+            warp, inverse_map=tform,
+            order=order, preserve_range=True)
 
     # def time_same_type(self, dtype_in, N, order, dtype_tform):
     def time_same_type(self, dtype_in, N, order):
         """Test the case where the users wants to preserve their same low
         precision data type."""
-        result = self.warp(self.image, self.tform, order=self.order,
-                           preserve_range=True)
+        result = self.warp(self.image)
 
         # convert back to input type, no-op if same type
         result = result.astype(dtype_in, copy=False)
@@ -46,5 +42,4 @@ class WarpSuite:
     def time_to_float64(self, dtype_in, N, order):
         """Test the case where want to upvert to float64 for continued
         transformations."""
-        result = warp(self.image, self.tform, order=self.order,
-                      preserve_range=True)
+        result = self.warp(self.image)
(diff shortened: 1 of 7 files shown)


hmaarrfk:
Anyway, radon doesn't even call img_as_float anywhere, meaning that uint8 images behave very differently. oh well, not part of this PR for now.

hmaarrfk:
Without my radon hackery, you would get things like:

```
>   def _warp_fast(np_floats[:, :] image, np_floats_H[:, :] H,
E   TypeError: Function call with ambiguous argument types
```
when calling with uint8 for example.

hmaarrfk:
I'm really not sure what is going on. Benchmarks for float32 and float64 yield the same results.

```
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[  0.00%] ·· Benchmarking existing-py_home_mark2_miniconda3_envs_skdev_bin_python
[ 50.00%] ··· Running (benchmark_transform_warp.WarpSuite.time_warp--).
[100.00%] ··· benchmark_transform_warp.WarpSuite.time_warp                                          ok
[100.00%] ··· =============== ====== ============ ============ =============
              --                                      order                 
              ---------------------- ---------------------------------------
                  dtype_in      N         0            1             3      
              =============== ====== ============ ============ =============
               numpy.float32   128     138±6μs      410±20μs    1.17±0.06ms 
               numpy.float32   1024   12.3±0.6ms    29.4±2ms      75.7±2ms  
               numpy.float32   4096    299±5ms      583±5ms      1.36±0.03s 
               numpy.float64   128     124±3μs      400±10μs    1.12±0.04ms 
               numpy.float64   1024    16.0±2ms    33.0±0.8ms     79.3±1ms  
               numpy.float64   4096    273±5ms      557±10ms     1.32±0.03s 
              =============== ====== ============ ============ =============
```

ahojnnes:
@hmaarrfk Have you looked at the generated assembly instructions? Unless the compiler automatically vectorizes some of the operations, you are unlikely to see any performance benefit of float vs. double. This is because non-vectorized float/double operations are equally fast (on most of today's CPUs). Unless you are memory-bandwidth bound (which does not seem to be the case here), there won't be a gain in runtime.

hmaarrfk:
I have not looked at the gnerated assembly instructions.

Not too sure how to do that.

But yeah, i was hoping to see some difference when we benchmark larger image, which would be memory bound as it doesn't fit in CPU cache. Maybe that premise is wrong?

ahojnnes:
There is no need to fit the whole image into the cache. Since we ensure the image is stored contiguously in row major order and we are iterating through the image row by row, it's only necessary to fit a few rows of the image into the cache. It seems like we are keeping the CPU busy enough such that it can pull the next image rows fast enough.

There are a few opportunities for speeding up the warping:
- For multi-channel images, we currently iterate through the image three times. This is not ideal and we could do much faster in a single pass by using a stride equal to the number of channels.
- Hand-code vectorized SIMD instructions for the warping / interpolation code.

hmaarrfk:
Hmm, seems like you are right about the operation speed.

http://nicolas.limare.net/pro/notes/2014/12/12_arit_speed/

32 bit floating point multiplication only seems marginally faster than 64 bit floating point.

I would still rather my images don't double in terms of space in RAM :/

The other things you mentionned are probably more important though.

hmaarrfk:
I think somebody else implemented this.

PR #3809, opened 2019-03-19
Ensure watershed auto-markers respect mask

### LINKED ISSUE #3808: skimage.segmentation.watershed not giving expected output.
## Description
watershed method from skimage.segmentation is not working as it supposed to with mask parameter (or at least not as I assumed)
I think it is even segmenting area outside of mask. Some kind of dots (at regular interval) are in segmented image.
I have attached original (image which I want to segment), mask and segmented image.

## Way to reproduce
```python
import SimpleITK
from skimage.segmentation import clear_border, watershed, mark_boundaries

org = SimpleITK.ReadImage("original.png")
mask = SimpleITK.ReadImage("mask.png")
org_array = SimpleITK.GetArrayFromImage(org)
mask_array = SimpleITK.GetArrayFromImage(mask)

segments_watershed = watershed(org_array, 1000, mask=mask_array)
plt.imshow(segments_watershed, cmap="gray")
```
original
mask
seg

## Version information
```python
```

```python
# your output here
3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 18:50:55) [MSC v.1915 64 bit (AMD64)]
Windows-10-10.0.17134-SP0
scikit-image version: 0.14.1
numpy version: 1.15.4
```

jni:
## Description

Fixes #3808 

Additionally, I took the opportunity to allow watershed without passing in markers. Then the local minima in the image are used.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/morphology/watershed.py
```diff
@@ -154,17 +172,19 @@ def _offsets_to_raveled_neighbors(image_shape, structure, center, order='C'):
...
     Parameters
     ----------
     image: ndarray (2-D, 3-D, ...) of integers
         Data array where the lowest value points are labeled first.
-    markers: int, or ndarray of int, same shape as `image`
+    markers: int, or ndarray of int, same shape as `image`, optional
         The desired number of markers, or an array marking the basins with the
-        values to be assigned in the label matrix. Zero means not a marker.
+        values to be assigned in the label matrix. Zero means not a marker. If
+        no markers are given, the local minima of the image are used as
```
lagru: Nitpick: I think "no markers" may be confused with giving 0 or passing in an array that is False everywhere.

```suggestion
        no ``None``, the local minima of the image are used as
```

Maybe remove "Zero means not a marker." and just display a warning if the input leads to no seeds being defined. Is there a use case where this doesn't hint at a usage error?

skimage/morphology/tests/test_watershed.py
```diff
@@ -489,5 +471,28 @@ def test_incorrect_markers_shape():
...
+        image = np.ones((5, 6))
+        mask = np.ones((5, 7))
+        output = watershed(image, markers=4, mask=mask)
+
+
+def test_markers_in_mask():
+    data = blob
+    mask = (data != 255)
+    out = watershed(data, 25, connectivity=2, mask=mask)
+    # There should be no markers where the mask is false
```
emmanuelle: Why do you call this markers? For me markers and seeds are synonyms, corresponding to marked pixels from which the watershed flooding is done. How about "labeled pixels"? Sorry to be nitpicking about a comment in a test ;-).

emmanuelle: OK, I think I got it, we're talking about the same thing for markers, you just want to check that no marker was chosen outside the mask, so that all pixels outside the mask in the segmentation are set to zero. OK!

skimage/morphology/watershed.py
emmanuelle: If ``None`` (no markers provided), the local minimal of the image...

emmanuelle:
The code looks fine to me. Just a question, why did you use ``local_minimal`` and not ``peak_local_max``? With ``peak_local_max`` it would be possible to add a minimal distance between local minima, but then it adds another parameter... Well, I guess the watershed only provides good results with well-chosen markers anyway.

skimage/morphology/watershed.py
```diff
@@ -55,17 +57,33 @@ def _validate_inputs(image, markers, mask):
...
         mask = np.ones(image.shape, bool)
+    else:
+        mask = np.asanyarray(mask, dtype=bool)
+        n_pixels = np.sum(mask)
+        if mask.shape != image.shape:
+            message = ("`mask` (shape {}) must have same shape as "
+                       "`image` (shape {})".format(mask.shape, image.shape))
+            raise ValueError(message)
+    if markers is None:
+        from .extrema import local_minima
```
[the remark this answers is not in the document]
jni: extrema imports from watershed (the offsets to raveled neighbors), so not using this resulted in a circular import. I guess the correct way is to move offsets to its own utility file.

skimage/morphology/watershed.py
```diff
@@ -55,17 +57,33 @@ def _validate_inputs(image, markers, mask):
...
+        markers_bool = local_minima(image, connectivity=connectivity) * mask
+        markers = ndi.label(markers_bool)[0]
+    elif not isinstance(markers, (np.ndarray, list, tuple)):
```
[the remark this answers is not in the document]
jni: Possibly this is true, I'm not sure about the original rationale, this is new code. I'll change it.

jni:
@emmanuelle we've had a *lot* of issues with `peak_local_max` being buggy or having strange behaviour, many of which are unresolved, see:

https://github.com/scikit-image/scikit-image/issues?utf8=✓&q=is%3Aissue+is%3Aopen+peak_local_max

So I much prefer using `local_minima`, which actually matches the original watershed paper anyway. I'm less interested in making the watershed function do *all possible* seeding strategies than in providing seedless watershed as originally proposed.

jni:
@lagru @emmanuelle I believe I have addressed your concerns. =)

jni:
Hmm, [failure in Py3.7](https://travis-ci.org/scikit-image/scikit-image/jobs/511462972#L4299) that seems unrelated to this PR:

> Warning, treated as error:
> /home/travis/build/scikit-image/scikit-image/doc/source/auto_examples/color_exposure/plot_rgb_to_hsv.rst:26:Explicit markup ends without a blank line; unexpected unindent.

I'll see if I can fix it.

emmanuelle:
Merging, thanks!

PR #3811, opened 2019-03-22
win_size checks dims except last

Rochan-A:
## Description

Previously, win_size was checked against all the dimensions (including the last dimension, ie. pixel channel) of the input image. This PR fixes it by checking the size against all dimensions but the last one.

PS: I have made a few assumptions about how the function works. Please do let me know if I have misunderstood something.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_structural_similarity.py
+++ b/skimage/measure/_structural_similarity.py
@@ -142,7 +142,7 @@ def compare_ssim(X, Y, win_size=None, gradient=False,
         else:
             win_size = 7   # backwards compatibility
 
-    if np.any((np.asarray(X.shape) - win_size) < 0):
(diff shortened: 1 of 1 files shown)


jni:
Thanks for contributing, @Rochan-A! We always appreciate new contributors. But, unfortunately, this PR is indeed incorrect: if the image is multichannel, you need to pass the `multichannel=True` flag. Then the code follows a different path, which this check doesn't apply to. Have a look at the block that starts with `if multichannel:` earlier in the function.

I'm going to close this, feel free to reopen if I have misunderstood something.

Rochan-A:
Hi @jni, Thanks for clarifying. Cheers

PR #3822, opened 2019-03-29
ENHANCEMENT: Compatibility with dask for image thresholding methods

GenevieveBuckley:
## Description

We can make most of the scikit-image thresholding functions compatible with dask with a very small change. These functions are now compatible with dask arrays:
* threshold_isodata
* threshold_mean
* threshold_minimum
* threshold_otsu
* threshold_triangle
* threshold_yen

### Details
I wrap two  integer values with `int()` in the `__bincount_histogram()` from `skimage/exposure/exposure.py`. 

1. If the input is a numpy array, this is redundant and has no effect. 
2. If the input is a dask array it will force computation & let us avoid the error that happens when you pass a dask object in as a keyword argument to `numpy.bincount()` a few lines later.

I also allow users to control the histogram keyword argument  `source_range` from the thresholding functions (something we already do for the keyword argument controlling the number of histogram bins). This is of particular importance to dask users, who are likely to want to use `source_range='dtype'` as a way to keep computations as lazy as possible.

## Checklist

- [ ] ~~[Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py) (no new functions here)~~
  existing benchmark (I've given this a go, any feedback here? It's the first time I've written benchmarks for scikit-image)
- [ ] Unit tests (Should I write some that explicitly include dask arrays? What's the stance on compatibility between libraries and the scikit-image test suite?)

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -66,7 +66,8 @@ def _bincount_histogram(image, source_range):
     elif source_range == 'dtype':
         image_min, image_max = dtype_limits(image, clip_negative=False)
     image, offset = _offset_array(image, image_min, image_max)
-    hist = np.bincount(image.ravel(), minlength=image_max - image_min + 1)
+    hist = np.bincount(image.ravel(),
+                       minlength=int(image_max) - int(image_min) + 1)
     bin_centers = np.arange(image_min, image_max + 1)
     if source_range == 'image':
         idx = max(image_min, 0)
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -61,13 +61,12 @@ def _bincount_histogram(image, source_range):
     if source_range not in ['image', 'dtype']:
         raise ValueError('Incorrect value for `source_range` argument: {}'.format(source_range))
     if source_range == 'image':
-        image_min = np.min(image).astype(np.int64)
-        image_max = np.max(image).astype(np.int64)
+        image_min = int(np.min(image).astype(np.int64))
+        image_max = int(np.max(image).astype(np.int64))
     elif source_range == 'dtype':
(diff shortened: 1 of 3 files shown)


skimage/filters/thresholding.py
```diff
@@ -599,7 +616,7 @@ def threshold_li(image, *, tolerance=None):
     return threshold
 
 
-def threshold_minimum(image, nbins=256, max_iter=10000):
+def threshold_minimum(image, nbins=256, source_range='image', max_iter=10000):
```
jni: This is an API change and has to go at the end of the function, preferable with a keyword-only flag to prevent future similar problems

skimage/filters/thresholding.py
```diff
@@ -241,6 +241,11 @@ def threshold_otsu(image, nbins=256):
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
+    source_range : 'image' or 'dtype' (default value is 'image').
+    source_range : string, optional
```
jni: Why two lines here?

skimage/exposure/exposure.py
```diff
@@ -61,8 +61,8 @@ def _bincount_histogram(image, source_range):
     if source_range not in ['image', 'dtype']:
         raise ValueError('Incorrect value for `source_range` argument: {}'.format(source_range))
     if source_range == 'image':
-        image_min = np.min(image).astype(np.int64)
-        image_max = np.max(image).astype(np.int64)
+        image_min = int(np.min(image).astype(np.int64))
+        image_max = int(np.max(image).astype(np.int64))
```
jni: Wow, very cool! =)

GenevieveBuckley:
Thank you for your comments - I will reopen this as two separate pull requests, as you suggest.

PR #3823, opened 2019-03-29
ENHANCEMENT: Compatibility with dask for image thresholding methods

GenevieveBuckley:
## Description

We can make most of the scikit-image thresholding functions compatible with dask with a very small change. These functions are now compatible with dask arrays:
* threshold_isodata
* threshold_mean
* threshold_minimum
* threshold_otsu
* threshold_triangle
* threshold_yen

### Details
I wrap two  integer values with `int()` in the `_bincount_histogram()` from `skimage/exposure/exposure.py`. 

1. If the input is a numpy array, this is redundant and has no effect. 
2. If the input is a dask array it will force computation & let us avoid the error that happens when you pass a dask object in as a keyword argument to `numpy.bincount()` a few lines later.

Note: this does mean any threshold methods with dask won't strictly have lazy computation. I'm putting together another PR that will give users an option to avoid forced this computation.

## Checklist

- [ ] ~~[Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py) (no new functions here)~~
  existing benchmark
- [x] Unit tests

## For reviewers

  later.


GenevieveBuckley:
> This is great, and I'd love to see more of it in other areas of the package.

Us too!

(And if you have the time or inclination to help out let us know, I'm happy to share anything I know that's useful 😄)

skimage/exposure/exposure.py
```diff
@@ -61,8 +61,8 @@ def _bincount_histogram(image, source_range):
     if source_range not in ['image', 'dtype']:
         raise ValueError('Incorrect value for `source_range` argument: {}'.format(source_range))
     if source_range == 'image':
-        image_min = np.min(image).astype(np.int64)
-        image_max = np.max(image).astype(np.int64)
+        image_min = int(np.min(image).astype(np.int64))
+        image_max = int(np.max(image).astype(np.int64))
```
jakirkham: Are `np.min(image)` and `np.max(image)` returning Dask Arrays when `image` is a Dask Array? If so, casting to `int` will `compute`.

GenevieveBuckley: Yes, that's my intention, sorry that wasn't clearer. 

The problem is: we can't use dask arrays as the values of keyword arguments without causing an error, even when the function itself can take dask arrays as the main input (we see this with `np.bincount()` a few lines later). That's something that ultimately would be very worthwhile to fix, this is a hack that works around it.

I wouldn't say it's *ideal* to have two dask computations triggered when you use the skimage histogram function, but I do think it's a heck of a lot better than not being able to use it at all. 

Finally, there is also another option that won't force computation. The scikit-image histogram function allows `source_range='dtype'` - which will guess an appropriate min/max based on the array type, no computation necessary.

emmanuelle:
Thank you @GenevieveBuckley ! Your PR made me think that we don't have any example in the gallery showcasing the use of dask. Would you have ideas for such an example? (this is independent of this PR).

emmanuelle:
Also, the CI build fails for minimal dependencies.

hmaarrfk:
Old dask probably didn't automatically cast list to numpy arrays. I think changing the test is appropriate.

GenevieveBuckley:
> Thank you @GenevieveBuckley ! Your PR made me think that we don't have any example in the gallery showcasing the use of dask. Would you have ideas for such an example? (this is independent of this PR).

Do you have a good idea for how to generate large example data (ie: not just a huge array of random numbers)?

Once we have that, the gallery example could be something very similar to the image segmentation pipeline I made for the [image processing notebook](https://github.com/dask/dask-examples/tree/master/applications)  in the dask-example docs.

GenevieveBuckley:
> Old dask probably didn't automatically cast list to numpy arrays. I think changing the test is appropriate.

Done :)

hmaarrfk:
Sounds like you also need to bump numpy

https://travis-ci.org/scikit-image/scikit-image/jobs/513700548#L2663

Min didn't always take the out argument

hmaarrfk:
I think that this would have to come in a minor release now that we require a new version of numpy. @jni probably already release 0.15.

hmaarrfk:
@GenevieveBuckley fun fun, seems like there was some disagreement about supporting diffs for np.bool

https://github.com/numpy/numpy/issues/9251

Maybe lets try bumping to numpy 1.14.

hmaarrfk:
I think dask didn't have the `out` argument until 0.15 (found that by trying things locally).

we can try again with numpy 1.12 after too.

GenevieveBuckley:
Thanks @hmaarrfk for putting in effort to work out what combination of dependencies we need for this!

hmaarrfk:
no problem. I think bumping minimum versions is good for progress, but not good if we do it for no good reason.

PR #3824, opened 2019-03-30
J-invariant denoising

batson:
## Description

This PR implements a method for converting any denoising function into a "J-invariant" denoiser for which hyperparameters can be calibrated in a principled way. The theoretical basis and experimental evidence for this calibration method is described in the [preprint](https://arxiv.org/abs/1901.11365) _Noise2Self: Blind Denoising by Self-Supervision_. This allows the user to choose the best denoiser and the best parameters for that denoiser for a given noisy image, without access to clean data.

## Checklist

  existing benchmark
- [X] Unit tests

## For reviewers

  later.


emmanuelle:
Hi @batson thank you for your PR! I had a quick look at the preprint and the method looks very interesting: parameter selection is always hard and help in selecting parameters would be very welcome. However, I'm afraid the method is too bleeding-edge to be included in scikit-image soon: we only include well-established methods, corresponding in general to academic papers which have been well cited over the years. It can be frustrating not to be able to benefit from the latest developments of the scientific community, but it's likely that your paper will inspire other researchers or your own team, who will propose improvements or alternative methods, so that a consolidated algorithm will emerge gradually. In order to keep the API of scikit-image short, we must limit ourselves to these consolidated version only, otherwise we would have to include a very large number of variants, which would be close to the state-of-the-art, but also very hard to navigate for non-specialist users.

@scikit-image/core we could have a ``skimage-externals`` or ``skimage-next`` repo in the scikit-image organization for interesting contributions like this PR. Of course we would also need the resources to maintain such a repo...

jni:
@emmanuelle and @scikit-image/core, sorry that we didn't make this clear sooner (ie with the PR), but @stefanv and I actually invited @batson to submit a PR after hearing about his work in person. We both feel that this is a paradigm shift in thinking about denoising, and are willing to make an exception to our usual "wait for it to be established" policy. The other aspect of this PR that we like is that the API is generic, so that we can swap out any kind of parameter optimiser in the future (rather than the current grid search) as long as we have a way to evaluate how well the denoiser is doing — which the J-invariance framework gives us.

Anyway, wait for @batson's gallery example, which is coming, before passing any further judgement, it pretty much speaks for itself! =)

emmanuelle:
Then I'm waiting for the gallery example :-). I understand that this method is really cool and maybe even a change of paradigm. However, until now we've been happy to delegate to the academic community the enormous workload of selecting the best and more robust methods. It's not something that we can do consistently ourselves, I think (@jni you have much more bandwidth now for skimage, but still...). I worry most about consistency and what we say to academic colleagues who are eager to have their algorithms within scikit-image (I get asked all the time at conferences, and sometimes when it's a close colleague it can be hard to say no unless it's well established in the guidelines of the package). But, OK, let's wait for a terrific demo and no more doubts until then ;-).

batson:
Hi all,

I have added a gallery example to demonstrate the use and accuracy of the self-supervised calibration framework. This is my first PR, so feedback is very welcome.

With respect to the appropriateness of adding `calibrate_denoiser` to `scikit-image`, I am sympathetic to the concerns of all involved. Being new to this community, I am looking forward to the discussion.

There have been [two](https://arxiv.org/abs/1901.10277) [papers](https://arxiv.org/abs/1811.10980) by other authors using the self-supervised loss to train deep neural nets, which for me was good confirmation of the basic principle that J-invariance allows for accurate calibration. I expect that there will be innovation in the future on similar methods but they may sit in the same framework, and potentially API. Since the self-supervised loss provably estimates the ground-truth loss, it would be straightforward to include arbitrary new methods exploiting J-invariance in the future.

One thing which occurs to me is that we can separate the function (calibration) from the metric (self-supervised loss) and the method (interpolation to produce J-invariance). Perhaps just like `denoise_wavelet` is, under the hood, many algorithms functioning on the same principle (eg, `wavelet` can be `db1`, `haar`, etc, and `method` can be `BayesShrink` or `VisuShrink`), `calibrate_denoiser` could support other self-supervised methods or even other methods entirely as they develop.

Thank you all for working to make this a robust, stable, and useful package for scientific imaging, and for the opportunity to contribute.

jni:
@batson would you mind putting up some screenshots of the figures generated by the example?

jni:
@emmanuelle I am very sympathetic to the importance of having a consistent policy about when we include algorithms, and I too have had the experience of having people ask to include their algorithm. But my policy is that we *generally* don't accept algorithms before they are established, ie, there is room for rare exceptions. (I think that virtually all rules should have some wiggle room for human judgement and interpretation.) But I have found that that statement is enough to have people accept the decision. And I think it helps that we follow the rule ourselves, e.g. I wrote a separate `skan` package rather than add to skimage.

Anyway, now is a good time to check out @batson's example. =) 

Thanks for the hard work @batson! =)

stefanv:
On Tue, 02 Apr 2019 00:04:54 -0700, Emmanuelle Gouillart wrote:
> Then I'm waiting for the gallery example :-). I understand that this
> method is really cool and maybe even a change of paradigm. However,
> until now we've been happy to delegate to the academic community the
> enormous workload of selecting the best and more robust methods. It's
> not something that we can do consistently ourselves, I think (@jni you
> have much more bandwidth now for skimage, but still...). I worry most
> about consistency and what we say to academic colleagues who are eager
> to have their algorithms within scikit-image (I get asked all the time
> at conferences, and sometimes when it's a close colleague it can be
> hard to say no unless it's well established in the guidelines of the
> package). But, OK, let's wait for a terrific demo and no more doubts
> until then ;-).

I think we have to distinguish between methods and meta-methods, if
which this seems to be the latter.  Not that it means any meta-method
should simply be accepted, just because it falls into that category.
But if a meta-method is an obvious way of doing things, it may be
considered for adding.

I think, since this touches on the machine learning/cross-validation
front, we should get people like Emma, Andy, and/or someone else from
the scikit-learn community to take a look, and tell us if it seems
straightforward and correct.  I find the examples + paper compelling,
but it would be good to have that input from machine learning experts
too, since this is slightly out of our typical expertise.

Let's also ping @amueller, who happens to be a core developer of
scikit-learn and scikit-image, and see if he can recommend a reviewer
for this technique.

batson:
# Calibrating Denoisers Using J-Invariance

In this example, we show how to find an optimally calibrated
version of any denoising algorithm.

fig 1

## The Self-Supervised Loss and J-Invariance

The key to this calibration method is the notion of J-invariance. A denoising
function is J-invariant if the prediction it makes for each pixel does
not depend on the value of that pixel in the original image. The prediction
for each pixel may instead use all the relevant information contained in the
rest of the image, which is typically quite significant. Any function
can be converted into a J-invariant one using a simple masking procedure,
as described in [1].

The pixel-wise error of a J-invariant denoiser is uncorrelated
to the noise, so long as the noise in each pixel is independent.
Consequently, the average difference between the denoised image and the
noisy image, the _self-supervised loss_, is the same as the
difference between the denoised image and the original clean image, the
_ground-truth loss_ (up to a constant).

This means that the best J-invariant denoiser for a given image can
be found using the noisy data alone, by selecting the denoiser minimizing
the self-supervised loss. Below, we demonstrate that this
for a family of wavelet denoisers with varying `sigma` parameter. The
self-supervised loss (solid blue line) and the ground-truth loss (dashed
blue line) have the same shape and the same minimizer.


fig 2

## Conversion to J-invariance

The function `invariant_denoise` acts as a J-invariant version of a
given denoiser. It works by masking a fraction of the pixels, interpolating
them, running the original denoiser, and extracting the values returned in
the masked pixels. Iterating over the image results in a fully J-invariant
output.

For any given set of parameters, the J-invariant version of a denoiser
is different from the original denoiser, but it is not necessarily better
or worse. In the plot below, we see that, for the test image of a cat,
the J-invariant version of a wavelet denoiser is significantly better
than the original at small values of variance-reduction `sigma` and
imperceptibly worse at larger values.


fig 3

## Comparing Different Classes of Denoiser

The self-supervised loss can be used to compare different classes of
denoiser in addition to choosing parameters for a single class.
This allows the user to, in an unbiased way, choose the best parameters
for the best class of denoiser for a given image.

Below, we show this for an image of the hubble deep field with significant
speckle noise added. In this case, the J-invariant calibrated denoiser is
better than the default denoiser in each of three families of denoisers --
Non-local means, wavelet, and TV norm. Additionally, the self-supervised
loss shows that the TV norm denoiser is the best for this noisy image.


fig 4a
fig 4b

emmanuelle:
Adding @glemaitre to the list of people mastering both image processing and machine learning :-) and help us review here.

emmanuelle:
Also pinging @NelleV because the preprint also has applications to gene detection.

emmanuelle:
Thank you @batson for the update and your kind words about the package. @jni I'm not paranoid about rules and I'm fine with exceptions as long as we keep a clear vision. Also the fact mentioned by @stefanv that this is a meta-algorithm makes ad difference to me. 

I also agree that we would value the review from machine learning / stats experts such as @amueller @NelleV or @glemaitre. 

A few questions : 
- at the moment when a set of parameters is provided, the optimum is computed by grid search / brute force. Would it be interesting to use a "real" optimization algorithm such as ``scipy.optimize.minimize`` (with a fairly lenient tolerance)? On a related note, do we expect the loss to be a convex function of hyperparameters or not? If it's a very complicated function, the brute force might be better...
- For the API, the denoising function could allow for tuples for parameters (which would be the bound of an interval) and have a keyword argument  ``j_invariant`` (True or False), if True the calibrated denoiser would be searched for and used. It sounds a bit "magic" and we're always cautious about to much magic stuff, but it could also make it easier to use this method.
- A lot of the for loops are embarrassingly parallel and could be accelerated with dask, see for example https://github.com/scikit-image/scikit-image/blob/master/skimage/restoration/_cycle_spin.py#L166

emmanuelle:
Also, I love the gallery example, but it takes a good 40 seconds on my laptop, it would be good to reduce the time to a few seconds to keep the build time of the documentation small enough. Maybe parallelization can help, and also maybe reducing the number of parameters in places?

doc/examples/filters/plot_j_invariant.py
```diff
@@ -0,0 +1,294 @@
...
+# not depend on the value of that pixel in the original image. The prediction
+# for each pixel may instead use all the relevant information contained in the
+# rest of the image, which is typically quite significant. Any function
+# can be converted into a J-invariant one using a simple masking procedure,
+# as described in [1].
+#
+# The pixel-wise error of a J-invariant denoiser is uncorrelated
+# to the noise, so long as the noise in each pixel is independent.
+# Consequently, the average difference between the denoised image and the
+# noisy image, the _self-supervised loss_, is the same as the
```
emmanuelle: please use stars instead of underscores for italic (*self-supervised loss*); the tests fail here because rst is looking for some reference.

doc/examples/filters/plot_j_invariant.py
```diff
@@ -0,0 +1,294 @@
...
+# difference between the denoised image and the original clean image, the
+# _ground-truth loss_ (up to a constant).
```
emmanuelle: same here

batson: Thank you for the thoughtful comments, @emmanuelle.

> * at the moment when a set of parameters is provided, the optimum is computed by grid search / brute force. Would it be interesting to use a "real" optimization algorithm such as `scipy.optimize.minimize` (with a fairly lenient tolerance)? On a related note, do we expect the loss to be a convex function of hyperparameters or not? If it's a very complicated function, the brute force might be better...

I like this idea. The runtime will just be proportional to the number of parameter values tested, but can also be brought down by working on a representative subset of the image.

For every search I have done, the function has been convex in the ordinal hyperparameters (like`sigma` for wavelets, `h` and `patch_size` and `patch_distance` for NL means), at least in any neighborhood of the optimum. (I'm not sure what happens if you start off wrong by an order of magnitude.) No guarantees about the categorical parameters (which family of wavelets or which thresholding, for example).

> * A lot of the for loops are embarrassingly parallel and could be accelerated with dask, see for example https://github.com/scikit-image/scikit-image/blob/master/skimage/restoration/_cycle_spin.py#L166

That would be cool! I haven't used dask before. To speed up the for loops in `invariant_denoise` and `calibrate_denoiser_search`, would I need to make them into list comprehensions as in the example?

jni:
@emmanuelle @batson 

> Would it be interesting to use a "real" optimization algorithm such as scipy.optimize.minimize

@batson and I discussed that grid search is the worst possible option, but I think that better optimisation can come later, since the function name has no indication of the optimisation method.

> A lot of the for loops are embarrassingly parallel and could be accelerated with dask

This is generally true in many places in skimage, and although I'm happy if @batson includes this logic, I'm also happy to leave this for a later PR because I don't want to repeat the pattern of checking for `dask_available` everywhere — if we want to go this route, we should make a unified wrapper that all parallel loops can use.

emmanuelle:
@jni and what do you think about the API question?

jni:
@emmanuelle I think that magic API could come later. I'm not too fond of it — some parameters are already tuples (such as `sigma`), and as a result the API could get very messy very quickly. Using the calibration API makes more sense to me, because here it is expected that you will provide a parameter set/range from the beginning.

emmanuelle:
@NelleV I heard you have more time now :-p

jni:
@scikit-image/core and @NelleV anyone want to make another pass at this? =)

On the plus side, next week I can *literally lean* on @NelleV for a review. :joy:

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,286 @@
...
+    best_denoise_function = lambda x: invariant_denoise(x,
+                                                        denoise_function,
+                                                        stride=stride,
+                                                        multichannel=multichannel,
+                                                        denoiser_kwargs=best_parameters)
+
+    return best_denoise_function
+
+
+def calibrate_denoiser_search(image, denoise_function, parameter_ranges, *,
```
sciunto: Should this be private? Otherwise, the docstring is not very informative about the difference with `calibrate_denoiser`

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,286 @@
...
+
+    for mask in masks:
+        input_image = image.copy()
+        #input_image = image
+        input_image[mask] = interp[mask]
+        output[mask] = denoise_function(input_image, **denoiser_kwargs)[mask]
+    return output
+
+
+def product_from_dict(dictionary):
```
sciunto: I thinks this is internal only. -> _product_from_dict

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,286 @@
...
+    controlled by `idx`.
+
+    Parameters
+    ----------
+    shape : tuple of int
+        Shape of the mask.
+    idx : int
+        The offset of the grid of ones. Iterating over `idx` will cover the
+        entire array.
+    stride : int
```
sciunto:
```suggestion
    stride : int, optional
```

skimage/restoration/j_invariant.py
sciunto: I had hard time to figure out why we have two functions with close behaviors. If the purpose is to get information about the process, then I think this must be an option to set, not a different function to use.

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,286 @@
...
+        interp = np.zeros(image.shape)
+        for i in range(image.shape[-1]):
+            interp[..., i] = ndi.convolve(image[..., i], conv_filter,
+                                          mode='mirror')
+    else:
+        interp = ndi.convolve(image, conv_filter, mode='mirror')
+    return interp
+
+
+def generate_mask(shape, idx, stride=3):
```
sciunto: Private?

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,286 @@
+from itertools import product
+
+import numpy as np
+from scipy import ndimage as ndi
+
+from ..measure import compare_mse
+from ..util import img_as_float
+
+
+def interpolate_image(image, multichannel=False):
```
sciunto: Private?

skimage/restoration/j_invariant.py
jni: Thanks for raising this. I think the docstring needs updating, and maybe the title, but it's still a useful public function. One function, `calibrate_denoiser`, returns a denoising function, ready to apply to new images, while `calibrate_denoiser_search` returns parameter combinations and corresponding losses.

Probably what we need here is for this API to make it easy to swap out search algorithms in the future, so that we don't have to rely on grid search forever, and if we do add other search strategies, we don't have to have a deprecation cycle...

jni:
@emmanuelle @sciunto I've rebased and pushed some changes on top of @batson's. Hopefully he doesn't mind. =P (All the original commits are intact, so my interloping can be reverted if necessary ;)

I think my updates address almost all of your comments. For example, I've renamed `parameter_ranges` to `denoise_parameters` and made it a keyword-only argument, which should allow us to keep the same API even if we move to more clever sampling/searching strategies. I've also made all the functions private that @sciunto tagged.

So, this is ready for another round of review! I'll ping @NelleV again for good measure. =P

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,309 @@
...
+import functools
+
+import numpy as np
+from scipy import ndimage as ndi
+
+from ..measure import compare_mse
+from ..util import img_as_float
+
+
+def _interpolate_image(image, multichannel=False):
```
sciunto:
```suggestion
def _interpolate_image(image, *, multichannel=False):
```
?

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,309 @@
...
+        denoise_function=denoise_function,
+        stride=stride,
+        multichannel=multichannel,
+        denoiser_kwargs=best_parameters,
+    )
+
+    return best_denoise_function
+
+
+def calibrate_denoiser_search(image, denoise_function, *, denoise_parameters,
```
sciunto: I think it would help if we call this function calibrate_denoiser_search_history. Would this be unacceptably too long?

doc/examples/filters/plot_j_invariant.py
```diff
@@ -0,0 +1,292 @@
...
+
+image = rgb2gray(img_as_float(hubble_deep_field()[100:250, 50:300]))
+
+sigma = 0.4
+noisy = random_noise(image, mode='speckle', var=sigma ** 2)
+
+parameter_ranges_tv = {'weight': np.arange(0.01, 0.3, 0.02)}
+parameters_tested_tv, losses_tv = calibrate_denoiser_search(
+                                    noisy,
+                                    denoise_tv_chambolle,
```
emmanuelle: the example raises an error here `TypeError: calibrate_denoiser_search() takes 2 positional arguments but 3 were given`.

doc/examples/filters/plot_j_invariant.py
```diff
@@ -0,0 +1,292 @@
...
+# The pixel-wise error of a J-invariant denoiser is uncorrelated
+# to the noise, so long as the noise in each pixel is independent.
+# Consequently, the average difference between the denoised image and the
+# noisy image, the *self-supervised loss*, is the same as the
+# difference between the denoised image and the original clean image, the
+# *ground-truth loss* (up to a constant).
+#
+# This means that the best J-invariant denoiser for a given image can
+# be found using the noisy data alone, by selecting the denoiser minimizing
+# the self-supervised loss. Below, we demonstrate that this
```
emmanuelle:
```suggestion
# the self-supervised loss. Below, we demonstrate this
```

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,309 @@
...
+     [0 0 0 0]
+     [0 1 0 1]]
+    """
+    phases = np.unravel_index(offset, (stride,) * len(shape))
+    mask = tuple(slice(p, None, stride) for p in phases)
+
+    return mask
+
+
+def invariant_denoise(image, denoise_function, *, stride=4, multichannel=False,
```
emmanuelle: with the recent merge of #4446 all denoising functions in scikit-image now take a `multichannel` argument so maybe we can rely on `denoiser_kwargs` instead?

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,309 @@
...
+    selections : dicts of values
+        Dicts containing individual combinations of the values in the input
+        dict.
+    """
+    keys = dictionary.keys()
+    for element in itertools.product(*dictionary.values()):
+        yield dict(zip(keys, element))
+
+
+def calibrate_denoiser(image, denoise_function, *, denoise_parameters,
```
emmanuelle:
```suggestion
def calibrate_denoiser(image, denoise_function, denoise_parameters, *
```

emmanuelle: unless `denoise_parameters` should be a kwarg

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,309 @@
...
+    image : ndarray
+        Input data to be denoised (converted using `img_as_float`).
+    denoise_function : function
+        Denoising function to be calibrated.
+    denoise_parameters : dict of list
+        Ranges of parameters for `denoise_function` to be calibrated over.
+    stride : int, optional
+        Stride used in masking procedure that converts `denoise_function`
+        to J-invariance.
+    multichannel : bool, optional
```
emmanuelle: maybe we don't need `multichannel`, see above

emmanuelle:
Hey @jni could you please review this one to see whether we can squeeze it in 0.17?

emmanuelle:
@scikit-image/core I believe that this PR is ready, if someone has the time to review it today to see if it could make it for the release that'd be awesome!!

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,327 @@
...
+    image : ndarray
+        Input data to be denoised (converted using `img_as_float`).
+    denoise_function : function
+        Original denoising function.
+    stride : int, optional
+        Stride used in masking procedure that converts `denoise_function`
+        to J-invariance.
+    multichannel : bool, optional
+        Whether the last axis of the image is to be interpreted as multiple
+        channels or another spatial dimension.
```
rfezzani, line 105:
```suggestion
```

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,327 @@
...
+
+    Returns
+    -------
+    interp : ndarray
+        Interpolated version of `image`.
+    """
+    spatialdims = image.ndim if not multichannel else image.ndim - 1
+    conv_filter = ndi.generate_binary_structure(spatialdims, 1).astype(float)
```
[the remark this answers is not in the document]
emmanuelle: Yeah, that's how the algorithm works in the paper, you take the average of 4 neighbors in 2D or 6 neighbors in 3D.

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,317 @@
...
```
rfezzani, line 28:
```suggestion
    conv_filter = ndi.generate_binary_structure(spatialdims, 1).astype(image.dtype)
```

skimage/restoration/j_invariant.py
```diff
@@ -0,0 +1,317 @@
...
+    conv_filter.ravel()[conv_filter.size // 2] = 0
+    conv_filter /= conv_filter.sum()
+
+    if multichannel:
+        interp = np.zeros(image.shape)
```
rfezzani, line 33:
```suggestion
        interp = np.zeros_like(image)
```

emmanuelle:
argh there's more CI failure when building the gallery tutorial for python 3.8, with the `psnr` function...

emmanuelle:
Hurray the CI is green ! Thanks @jni for your help.

rfezzani:
:tada: Thank you @emmanuelle

emmanuelle:
Thanks @rfezzani ! Kudos should go to @batson for the implementation ;-)

rfezzani:
Of course! Thank you @batson for this awesome contribution ;-)

batson:
Thanks to all of you, and especially @emmanuelle for getting this integrated and @jni for walking me through the process of contributing to skimage!

PR #3825, opened 2019-03-30
[WIP] Modified example comments to try to fix CI

emmanuelle:
Travis is broken at the moment, this is an attempt to fix it...

### DIFF
--- a/doc/examples/color_exposure/plot_rgb_to_hsv.py
+++ b/doc/examples/color_exposure/plot_rgb_to_hsv.py
@@ -16,13 +16,9 @@ channels.
 
 .. [1] https://en.wikipedia.org/wiki/HSL_and_HSV
 
+We first load the RGB image and extract the Hue and Value channels:
 """
-
-##############################################################################
-# We first load the RGB image and extract the Hue and Value channels:
-
 import matplotlib.pyplot as plt
-
 from skimage import data
 from skimage.color import rgb2hsv
 
--- a/doc/examples/color_exposure/plot_rgb_to_hsv.py
+++ b/doc/examples/color_exposure/plot_rgb_to_hsv.py
@@ -7,16 +7,17 @@ This example illustrates how RGB to HSV (Hue, Saturation, Value) conversion
 [1]_ can be used to facilitate segmentation processes.
 
 Usually, objects in images have distinct colors (hues) and luminosities, so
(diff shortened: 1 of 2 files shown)


emmanuelle:
Still failing... I find it very hard to understand what's going on, given the fact that the build passes for Python 3.6 and not Python 3.7 and it's a doc/sphinx problem... Will keep on investigating.

hmaarrfk:
Travis is probably in rough shape. If you would like, you can try to move to azure.

hmaarrfk:
Travis the company, not our CI script.

emmanuelle:
@hmaarrfk do you think it could be a problem of Travis ? The error is generated by the doc build and is reproducible (at least on Travis, I can't reproduce on my machine but I didn't try to have exactly the same environment).

hmaarrfk:
Ah you are right. It seems like an issue with the syntax formatting in the rst. 

Seems unrelated to the reported Travis outages.

soupault:
Would there be any interest in having the doctests in Azure Pipelines? I can work on this

hmaarrfk:
@soupault I'm pretty convinced that azure is the way to go now. I just didn't want to jump on too soon. These kind of "low profit" projects tend to get abandoned by larger companies (e.g. google).

soupault:
@hmaarrfk well... MS owns GitHub now, so I'm rather confident about Azure Pipelines being up and running for a while.
However, I think there are some limitations to our subscription: as far as I understood, we don't have a possibility to store artifacts(/built docs) on Azure. That can, probably, be easily overcome using external storage, but I have to double check.

hmaarrfk:
I think we fixed this in @stefanv's PR.

stefanv:
On Wed, 03 Apr 2019 08:09:10 -0700, Egor Panfilov wrote:
> @hmaarrfk well... MS owns GitHub now, so I'm rather confident about Azure Pipelines being up and running for a while.
> However, I think there are some limitations to our subscription: as
> far as I understood, we don't have a possibility to store
> artifacts(/built docs) on Azure. That can, probably, be easily
> overcome using external storage, but I have to double check.

For what it's worth, we don't currently store artifacts on *any* of our
CIs, and our docs get pushed to GitHub upon successful build.  It's
useful to have docs uploaded somewhere after builds for checking, but
it's not been a deal breaker so far.

PR #3829, opened 2019-04-07
Masked marching cubes

Leguark:
## Description

Added functionality to mask the marching cube algorithm. We need this functionality for GemPy to get finite surfaces (https://github.com/cgre-aachen/gempy). 

image

Also gives a nice speed boost if the localtion of the interfaces is known withing the regular mesh. However, passing a full Trues masking array  added an overhead of around+40% computing time for a 1M voxels). That is why I left the original cython function untouched and only if a mask array is passed we call the funtion with the new functionality. probably there is a more elegant way to do this so we do not need to keep both functions but my experience with cython is limited.

## Checklist

  existing benchmark
- [x] Unit tests 

## For reviewers

  later.


skimage/measure/_marching_cubes_lewiner.py
```diff
@@ -57,6 +57,10 @@ def marching_cubes_lewiner(volume, level=None, spacing=(1., 1., 1.),
         produce a topologically correct result. The results with using
         this option are *not* generally the same as the
         ``marching_cubes_classic()`` function.
+    mask : (M, N, P) array
+        Bool array. The marching cube algorith will be computed only on
+        True elements. This may save computational time when interfaces
+        are located and also may work to compute finite surfaces.
```
emmanuelle: Could you please explain what is a finite surface?

emmanuelle:
Hi @Leguark ! Could you please tell us a bit more about why you need this feature? I'm not sure I understand what a finite surface is, and why you can't get one with the current function. Some illustrations could really help :-). Also, which performance gain do you get typically (say, for a simple surface like a sphere) ?

Leguark:
Hi @emmanuelle! Thank you for the fast answer. With finite surface I mean something like the green and purple layers in the image above. Marching cubes always assume close surfaces (within or outside the regular grid) so if for example you have an array like: 
[[3,3, False, False],
 [2, 2.5, 3, False],
 [1, 1, False, False]]

and the level we are looking is 3 for examples. The surface will bend around the boundary 3 - False, following the voxels. You can see it here in the red layer (it is not the clearest picture to show the problem but it is just an issue in our library :) ).
 
image

Regarding the performance improvements depends on the percentage of Trues respect the total number of voxels. The blue horizontal layer of the prototype I was doing (which is the model I showed in the main message):

- Normal marching cubes 1M voxels: 14 ms
- Masked marching cubes 1M voxels all True: 20ms (this adds an overhead I mentioned thats why I left the original funtion)
- Masked marching cubes 1M voxels/200k True: 8 ms
- Masked marching cubes 1M voxels/2000 True: 6 ms

In this case the layer is infinite i.e. the number of triangles are the same. The speed simply comes of not having to evaluate if the level is contained in a voxel. I assume that when we have finite layers (green and purple) the speed up will be a bit higher. I can do a notebook in detail if you want.

Also, I do not understand the error that Travis raises. Do you think it is because the changes I did?

emmanuelle:
Thank you for your answer @Leguark. 

So, you define a finite surface as a surface which is not closed, am I right? Maybe a notebook would help, yes :-). 

Regarding Travis, I don't think the failure is due to your PR. If we decide to integrate this feature, we'll need you to add new tests for the new feature (and maybe extend the existing gallery example for marching cubes to illustrate the mask feature). But we first need to understand more about the use case :-). I'm having mixed feelings about the speed gain, since the mask also needs to be computed in some way, so I'm not sure the speed gain is so large...

Leguark:
hey @emmanuelle  did you have time to take a look at the notebook? I am about to make a major release of my software and although by now I will point to my fork in the pip installation, I would really like to know if the feature will be part of scikit image eventually or if I will have to find a creative way to inject the functionality otherwise.

By the way another useful case for masked marching cubes is if you do octtree refinement since you know that the surface will be within the smallest voxels,

Cheers

metazool:
@Leguark I couldn't find the masked marching cube example notebook at the link above, should it be this? https://github.com/cgre-aachen/gempy/blob/master/notebooks/tutorials/ch1-8b-MaskedMarchinCube_Onlap_relations.ipynb

What would be a good value for a mask to extend the test here? https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/tests/test_marching_cubes.py

Leguark:
Yes, thank you for checking @metazool, I just broke the link by rename the notebook sorry. I just updated it with the 3D mesh plot. If you think it is a useful feature I will create the test (and probably convert the GemPy tutorial to a more general tutorial to put in the scikit image repo) next Monday when I go back to the office

Leguark:
@metazool, I just created a test and also a simpler notebook: https://github.com/Leguark/Sandbox/blob/master/Masked%20marching%20cubes.ipynb I am not sure if this would fit somewhere.

Is anything missing?
Thank you!

skimage/measure/_marching_cubes_lewiner.py
```diff
@@ -154,6 +154,7 @@ def marching_cubes_lewiner(volume, level=None, spacing=(1., 1., 1.),
     # Apply algorithm
     # Check if a mask array is passed
     if mask is not None:
+        assert mask.shape == volume.shape, 'volume and mask must have the same shape.'
```
stefanv: We typically prefer raising a `ValueError` instead of using asserts.

skimage/measure/_marching_cubes_lewiner.py
```diff
@@ -57,6 +57,11 @@ def marching_cubes_lewiner(volume, level=None, spacing=(1., 1., 1.),
         produce a topologically correct result. The results with using
         this option are *not* generally the same as the
         ``marching_cubes_classic()`` function.
+    mask : (M, N, P) array
+        Boolean array. The marching cube algorithm will be computed only on
+        True elements. This will save computational time when interfaces
+        are located within certain region of the volume M, N, P and also
+        allow to compute finite surfaces.
```
stefanv: I am not sure what a "finite surface" is: could you use a different phrasing or, if this is a technical term, please explain it in the docstring?

stefanv: I think the story here is unclear: "when interfaces are located within certain regions"—what regions?

skimage/measure/_marching_cubes_lewiner_cy.pyx
```diff
@@ -999,6 +997,79 @@ def marching_cubes(float [:, :, :] im not None, double isovalue, LutProvider lut
     return cell.get_vertices(), cell.get_faces(), cell.get_normals(), cell.get_values()
 
 
+def marching_cubes_masked(float [:, :, :] im not None, double isovalue, LutProvider luts, int st=1, int classic=0,
```
stefanv: Is there any way to avoid this code duplication?  I.e., have a check inside of the existing marching cubes function instead?

skimage/measure/_marching_cubes_lewiner.py
Leguark: I just added an example of what I mean by certain region and added an explanation of the finite surface concept

skimage/measure/_marching_cubes_lewiner.py
Leguark: Refactored

skimage/measure/_marching_cubes_lewiner_cy.pyx
Leguark: yes, I was able to do it. Sorry I am not too familiar with cython so I just went to the easy route. Now is much better

Leguark:
Is there any decision about this pull request? Some of my users are having troubles to install my forked skimage cgre-aachen/gempy#216

If there is not plans to merge this pull request I will try to inject the functionality directly into my library.

Thank you for your time!

jni:
@Leguark we definitely want this functionality in scikit-image but we have a bit of a backlog, and it's not a trivial PR to review, unfortunately. (I myself don't know anything about the marching cubes algorithm so it'll take me a while to go over this, for one. And we need two people to give the thumbs up before merging.) We also just released 0.16 so this would not be out for another ~6mo at the earliest. So I think you probably want to find a workaround for your users.

In scikit-image, we usually roll our own workarounds, along with a TODO comment to remove them, when we want a feature not yet out in a dependent package. For example, we had our own `pad`, which was just a copy of the code for `numpy.pad` from the PR, until we moved our dependencies to a recent-enough NumPy.

Apologies again for the delay!

Leguark:
Alright now problem. Thank you very much for the insight. I totally understand the limitations of a project as skimage in terms of humans resources.

 I will see if I find a way to hack it for the following months. 

Keep up with the great work!

bistek:
Hello, I am one of the users above. The problem is, I cannot use the standard scikit-image, neither the one with the additions done by @Leguark, so I cannot use some important tools in the geomodelling project by @Leguark.

@Leguark suggested to install his own custom version of scikit, with:

`pip install git+git://github.com/Leguark/scikit-image@master`

but for some reason I get a long list of errors.

A possible reason for this is that I am on Windows and @Leguark is on Linux (or Mac?).

I captured the lines where the error comes out in my terminal, and I attach a file with the complete log that appears on the terminal.

```
(base) C:\Users\bistek>pip install git+git://github.com/Leguark/scikit-image@master
Collecting git+git://github.com/Leguark/scikit-image@master
Cloning git://github.com/Leguark/scikit-image (to revision master) to c:\users\bistek\appdata\local\temp\pip-req-build-xie73893
Running command git clone -q git://github.com/Leguark/scikit-image 'C:\Users\bistek\AppData\Local\Temp\pip-req-build-xie73893'
Running command git submodule update --init --recursive -q
Requirement already satisfied: networkx>=2.0 in c:\anaconda3\lib\site-packages (from scikit-image==0.16.dev0) (2.3)
Requirement already satisfied: pillow>=4.3.0 in c:\anaconda3\lib\site-packages (from scikit-image==0.16.dev0) (6.1.0)
Requirement already satisfied: imageio>=2.0.1 in c:\anaconda3\lib\site-packages (from scikit-image==0.16.dev0) (2.5.0)
Requirement already satisfied: PyWavelets>=0.4.0 in c:\anaconda3\lib\site-packages (from scikit-image==0.16.dev0) (1.0.3)
Requirement already satisfied: decorator>=4.3.0 in c:\anaconda3\lib\site-packages (from networkx>=2.0->scikit-image==0.16.dev0) (4.4.0)
Requirement already satisfied: numpy in c:\anaconda3\lib\site-packages (from imageio>=2.0.1->scikit-image==0.16.dev0) (1.16.5)
Building wheels for collected packages: scikit-image
Building wheel for scikit-image (setup.py) ... error
ERROR: Command errored out with exit status 1:
```

then a lot of errors in red here, then:

```
ERROR: Failed building wheel for scikit-image
Running setup.py clean for scikit-image
Failed to build scikit-image
Installing collected packages: scikit-image
Found existing installation: scikit-image 0.15.0
Uninstalling scikit-image-0.15.0:
Successfully uninstalled scikit-image-0.15.0
Running setup.py install for scikit-image ... error
ERROR: Command errored out with exit status 1:
```

then again a lot of error messages in red, then:

```
Rolling back uninstall of scikit-image
Moving to c:\anaconda3\lib\site-packages\scikit_image-0.15.0-py3.7.egg-info from c:\anaconda3\lib\site-packages\~cikit_image-0.15.0-py3.7.egg-info
Moving to c:\anaconda3\lib\site-packages\skimage from c:\anaconda3\lib\site-packages\~kimage
Moving to c:\anaconda3\scripts\skivi-script.py from C:\Users\bistek\AppData\Local\Temp\pip-uninstall-9rzj1x6_\skivi-script.py
Moving to c:\anaconda3\scripts\skivi.exe from C:\Users\bistek\AppData\Local\Temp\pip-uninstall-9rzj1x6_\skivi.exe
```

Do you have suggestions?

Thanks very much!

[scikit_error.txt](https://github.com/scikit-image/scikit-image/files/3735351/scikit_error.txt)

jni:
@bistek this is the relevant part of the errors:

```
    gcc -g -shared build\temp.win-amd64-3.7\Release\skimage\restoration\_nl_means_denoising.o -LC:\Anaconda3\libs -LC:\Anaconda3\PCbuild\amd64 -lpython37 -lvcruntime140 -o build\lib.win-amd64-3.7\skimage\restoration\_nl_means_denoising.cp37-win_amd64.pyd -fopenmp
    build\temp.win-amd64-3.7\Release\skimage\restoration\_nl_means_denoising.o: In function `__pyx_f_7skimage_11restoration_19_nl_means_denoising_patch_distance_2d':
    C:\Users\bistek\AppData\Local\Temp\pip-req-build-tzsnvq7a/skimage/restoration/_nl_means_denoising.c:2891: undefined reference to `fast_exp'
    build\temp.win-amd64-3.7\Release\skimage\restoration\_nl_means_denoising.o: In function `__pyx_f_7skimage_11restoration_19_nl_means_denoising_patch_distance_2dmultichannel':
    C:\Users\bistek\AppData\Local\Temp\pip-req-build-tzsnvq7a/skimage/restoration/_nl_means_denoising.c:3088: undefined reference to `fast_exp'
    build\temp.win-amd64-3.7\Release\skimage\restoration\_nl_means_denoising.o: In function `__pyx_f_7skimage_11restoration_19_nl_means_denoising_patch_distance_3d':
    C:\Users\bistek\AppData\Local\Temp\pip-req-build-tzsnvq7a/skimage/restoration/_nl_means_denoising.c:3278: undefined reference to `fast_exp'
    build\temp.win-amd64-3.7\Release\skimage\restoration\_nl_means_denoising.o: In function `__pyx_pf_7skimage_11restoration_19_nl_means_denoising_4_fast_nl_means_denoising_2d':
    C:\Users\bistek\AppData\Local\Temp\pip-req-build-tzsnvq7a/skimage/restoration/_nl_means_denoising.c:8140: undefined reference to `fast_exp'
    build\temp.win-amd64-3.7\Release\skimage\restoration\_nl_means_denoising.o: In function `__pyx_pf_7skimage_11restoration_19_nl_means_denoising_6_fast_nl_means_denoising_3d':
    C:\Users\bistek\AppData\Local\Temp\pip-req-build-tzsnvq7a/skimage/restoration/_nl_means_denoising.c:9340: undefined reference to `fast_exp'
```

My guess is that if you (or @Leguark) cherry-pick these commits to that branch, things will work again!

https://github.com/scikit-image/scikit-image/pull/3883/commits

Leguark:
of course, I did not pull from master since a while. How I could not think of that! Thanks for the advice @jni I will merge today and see if that solves the problem

skimage/measure/tests/test_marching_cubes.py
```diff
@@ -30,21 +30,29 @@ def test_marching_cubes_anisotropic():
...
     surf_calc = mesh_surface_area(verts, faces)
     # Test within 1.5% tolerance for anisotropic. Will always underestimate.
     assert surf > surf_calc and surf_calc > surf * 0.985
 
+    # Test marching cube with mask
+    with pytest.raises(AttributeError):
+        verts, faces = marching_cubes_lewiner(
+            ellipsoid_anisotropic, 0., spacing=spacing,
+            mask=np.array([]))[:2]
+        surf_calc = mesh_surface_area(verts, faces)
```
rfezzani: These two lines are never processed since an error is raised just before.

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -10,8 +10,43 @@ ctypedef np.float64_t IMGDTYPE
 
 cdef double DISTANCE_CUTOFF = 5.0
 
-cdef extern from "fast_exp.h":
-    double fast_exp(double y) nogil
+# A fast approximation of the exponential function.
```
rfezzani: Shouldn't this be part of an other PR?

rfezzani: Ah, sorry, I just noticed that it comes from another PR :smile: !!! It's fine then...

skimage/measure/_marching_cubes_lewiner_cy.pyx
```diff
@@ -973,27 +975,49 @@ def marching_cubes(float [:, :, :] im not None, double isovalue, LutProvider lut
...
+                        if nt > 0:
+                            cell.add_triangles(luts.CASESCLASSIC, cell.index, nt)
+                    else:
+                        # Get case, if non-nul, enter the big switch
+                        case = luts.CASES.get2(cell.index, 0)
+                        if case > 0:
+                            config = luts.CASES.get2(cell.index, 1)
+                            the_big_switch(luts, cell, case, config)
+
+                elif mask[z_st, y_st, x_st]:
```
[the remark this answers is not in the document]
rfezzani: I see in [`find_contours`](https://github.com/scikit-image/scikit-image/blob/v0.16.1/skimage/measure/_find_contours_cy.pyx#L103) that the neighborhood of the points are also checked for not being masked. Shouldn't this be also done here?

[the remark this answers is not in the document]
Leguark: okay after testing here https://github.com/Leguark/Sandbox/blob/master/Masked%20marching%20cubes.ipynb

I do not see necessary that condition. I guess the behaviour changes a bit. Since marching cubes acts on the edges, masking a voxel is a bit ambiguous if the edge should be masked or not. For my taste is better without the condition.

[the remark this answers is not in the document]
rfezzani: In a `if A or B` test, **B** is not evaluated if **A** is true. To illustrate this, see this snippet:
```python
>>> if True or 1/0: print('yes')
... 
yes
```
This will not cause any overhead and avoid code duplication.

It's OK to not check voxels neighbors, It was just a small remark ;-)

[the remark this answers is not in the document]
Leguark: That is good information, then I am going to add your suggestion

skimage/measure/tests/test_marching_cubes.py
```diff
@@ -30,21 +30,29 @@ def test_marching_cubes_anisotropic():
...
+
     # Lewiner
     verts, faces = marching_cubes_lewiner(
         ellipsoid_anisotropic, 0., spacing=spacing)[:2]
     surf_calc = mesh_surface_area(verts, faces)
     # Test within 1.5% tolerance for anisotropic. Will always underestimate.
     assert surf > surf_calc and surf_calc > surf * 0.985
 
+    # Test marching cube with mask
+    with pytest.raises(AttributeError):
```
rfezzani: Now it's a `ValueError` ;-)

Leguark: I had not seen that test, thank for the catch! I just sent the fix, lets hope this pass now

skimage/measure/_marching_cubes_lewiner_cy.pyx
```diff
@@ -929,13 +929,15 @@ cdef class LutProvider:
 
         self.SUBCONFIG13 = Lut(SUBCONFIG13)
 
-
-def marching_cubes(float [:, :, :] im not None, double isovalue, LutProvider luts, int st=1, int classic=0):
+def marching_cubes(float [:, :, :] im not None, double isovalue, LutProvider luts, int st=1, int classic=0,
+                   int [:, :, :] mask=None):
```
rfezzani: If you declare `mask` as `bint`, you can avoid its conversion to `int32` in *_marching_cube_lewiner.py* line 153 (see this [ref](https://github.com/cython/cython/wiki/DifferencesFromPyrex) for example).

Leguark: Thank you very much for the advice! It is the first time I wrote something on cython, I remember to be stuck on this point for a while and I was only able to make it run with the hack. I just pushed the suggested changes

rfezzani: Oops, looks like I misled you! Sorry...
Looking at one of my previous PR, I realized that a better way is:
```python
def marching_cubes(float [:, :, :] im not None, double isovalue, LutProvider luts, int st=1, int classic=0,
                   np.ndarray[np.npy_bool, ndim=3, cast=True] mask=None):
```

Leguark: and that compiles in C? I remember that at some point I had it some python comparison and it slowed down the whole thing by a lot

lagru: Cython's docs recommend [memoryviews](https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html) over the old `np.array` API. 

I haven't looked through the rest of the PR but if `mask` is a boolean array you can use it without conversion by constructing a view `mask.view(np.uint8)` beforehand which is a compatible dtype to `bint[:, :, :]` (or `np.uint8_t[:, :, :]`, same thing memorywise!). Cython's memoryview API can't yet deal with boolean arrays automatically. That should be the most efficient way.

If you need even faster indexing and your input arrays are contiguous take a look at [this section](https://cython.readthedocs.io/en/latest/src/userguide/memoryviews.html#c-and-fortran-contiguous-memoryviews).

rfezzani: Hi @lagru, thank you for this ref...
But the main advantage of the NumPy array buffer syntax here is that it explicitly deal with data cast to bool.
Cython MemoryView doesn't give much performance here since we do not enforce `mask` to be memory contiguous.

rfezzani: @Leguark, yes this is the numpy array buffer syntax, supported by Cython ;-)

lagru: > But the main advantage of the NumPy array buffer syntax here is that it explicitly deal with data cast to bool.

Note: using cast will be no longer necessary once we upgrade our dependency to Cython 0.29.14 wich contains a [fix added support for bool](https://github.com/cython/cython/pull/2676). 

> Cython MemoryView doesn't give much performance here since we do not enforce mask to be memory contiguous.

Have you benchmarked that? 

I'd guess that `mask` is a contiguous array most of the time so we could enforce it? Using `np.asarray(mask, dtype=bool, order="C").view(np.uint8)` together with the memoryview API might yield noticeable improvements for the common case considering that we check that array deep inside 3 nested loops.

Otherwise I'd still prefer not to mix both APIs here and go with the recommended one. Although I grant you that in this case the old API has an advantage over the new one and I won't complain if there is no performance difference. :wink:

rfezzani: I did not benchmarked, just guessing :-p

> I'd guess that `mask` is a contiguous array most of the time so we could enforce it? Using `np.asarray(mask, dtype=bool, order="C").view(np.uint8)` together with the memoryview API might yield noticeable improvements for the common case considering that we check that array deep inside 3 nested loops.

Good suggestion! :wink: 

Both options are OK for me, so @Leguark, its better to implement @lagru suggestion, and sorry once again that my proposition gave you extra work.

rfezzani:
@Leguark, you have to add the `mask` parameter to the `marching_cubes` function and pass it to `_marching_cubes_lewiner` [here](https://github.com/scikit-image/scikit-image/pull/3829/files#diff-814ab49575e140281fef9598f6cca201R124) and [here](https://github.com/scikit-image/scikit-image/pull/3829/files#diff-814ab49575e140281fef9598f6cca201R120). This is what is causing CI failures...

skimage/measure/_marching_cubes_lewiner.py
```diff
@@ -119,11 +126,11 @@ def marching_cubes(volume, level=None, *, spacing=(1., 1., 1.),
...
         return _marching_cubes_lewiner(volume, level, spacing,
                                        gradient_direction, step_size,
-                                       allow_degenerate, use_classic=False)
+                                       allow_degenerate, use_classic=False, mask=mask)
     elif method == 'lorensen':
         return _marching_cubes_lewiner(volume, level, spacing,
                                        gradient_direction, step_size,
-                                       allow_degenerate, use_classic=True)
+                                       allow_degenerate, use_classic=True, mask=mask)
     elif method == '_lorensen':
```
rfezzani: When mask is not None, may be add a warning in this case stipulating that mask is not supported with the '_lorensen' method.

Leguark: That is a good idea

skimage/measure/_marching_cubes_lewiner.py
```diff
@@ -119,12 +126,14 @@ def marching_cubes(volume, level=None, *, spacing=(1., 1., 1.),
...
+        if mask is not None:
+            warnings.warn('Attribute mask is not implemented for _lorensen method.')
```
lagru:
```suggestion
            warnings.warn(
                'Parameter `mask` is not implemented for method "_lorensen" '
                'and will be ignored.', stacklevel=2
            )
```
This ensures that the warning points the user at the place where the function `marching_cubes` is called and not at its internals. I also took the liberty to tweak the wording a little bit because "Attribute" as a different meaning in the Python space.

Leguark: oh of course I mixed it with argument but yes, parameter also works. Thank you very much

skimage/measure/_marching_cubes_lewiner.py
```diff
@@ -119,12 +126,17 @@ def marching_cubes(volume, level=None, *, spacing=(1., 1., 1.),
...
-                                       allow_degenerate, use_classic=False)
+                                       allow_degenerate, use_classic=False, mask=mask)
     elif method == 'lorensen':
         return _marching_cubes_lewiner(volume, level, spacing,
                                        gradient_direction, step_size,
-                                       allow_degenerate, use_classic=True)
+                                       allow_degenerate, use_classic=True, mask=mask)
     elif method == '_lorensen':
+        if mask is not None:
+            warnings.warn(
```
rfezzani: I think it is may be better to raise a `NotImplementedError` here...

skimage/measure/tests/test_marching_cubes.py
rfezzani: Can you please remove these two lines ;-)

skimage/measure/tests/test_marching_cubes.py
```diff
@@ -42,6 +43,14 @@ def test_marching_cubes_anisotropic():
     # Test within 1.5% tolerance for anisotropic. Will always underestimate.
     assert surf > surf_calc and surf_calc > surf * 0.985
 
+    # Test marching cube with mask
+    with pytest.raises(ValueError):
+        verts, faces = marching_cubes(
+            ellipsoid_anisotropic, 0., spacing=spacing,
+            mask=np.array([]))[:2]
+        surf_calc = mesh_surface_area(verts, faces)
+        assert surf > surf_calc and surf_calc > surf * 0.985
```
rfezzani:
```suggestion
```

skimage/measure/tests/test_marching_cubes.py
```diff
@@ -135,3 +144,42 @@ def test_both_algs_same_result_donut():
...
+    area = mesh_surface_area(ver, faces)
+
+    np.testing.assert_allclose(area, 299.56878662109375, rtol=.01)
+
+
+def test_masked_marching_cubes_empty():
+    with pytest.raises(ValueError):
+        ellipsoid_scalar = ellipsoid(6, 10, 16, levelset=True)
+        mask = np.array([])
+        ver, faces, _, _ = marching_cubes(ellipsoid_scalar, 0, mask=mask)
```
rfezzani:
```suggestion
def test_masked_marching_cubes_empty():
    ellipsoid_scalar = ellipsoid(6, 10, 16, levelset=True)
    mask = np.array([])
    with pytest.raises(ValueError):
        _ = marching_cubes(ellipsoid_scalar, 0, mask=mask)
```

skimage/measure/tests/test_marching_cubes.py
```diff
@@ -135,3 +144,42 @@ def test_both_algs_same_result_donut():
...
+
+
+def test_masked_marching_cubes_lewiner():
+    with pytest.raises(NotImplementedError):
+        ellipsoid_scalar = ellipsoid(6, 10, 16, levelset=True)
+        mask = np.array([])
+        ver, faces, _, _ = marching_cubes(ellipsoid_scalar, 0, mask=mask, method='_lorensen')
```
rfezzani:
```suggestion
def test_masked_marching_cubes_old_lewiner():
    ellipsoid_scalar = ellipsoid(6, 10, 16, levelset=True)
    mask = np.array([])
    with pytest.raises(NotImplementedError):
        ver, faces, _, _ = marching_cubes(ellipsoid_scalar, 0, mask=mask, method='_lorensen')
```

Leguark:
Thank you for the review and the code cleaning @rfezzani ! Hopefully we are able to merge this for the 0.17

metazool:
😘

Leguark:
You mean because mask is a python object and does not get compiled? Or because a and condition is faster?

it would be `cdef bint use_mask = mask is not None` though, isnt it

skimage/measure/_marching_cubes_lewiner_cy.pyx
```diff
@@ -929,13 +929,15 @@ cdef class LutProvider:
 
         self.SUBCONFIG13 = Lut(SUBCONFIG13)
 
-
-def marching_cubes(float [:, :, :] im not None, double isovalue, LutProvider luts, int st=1, int classic=0):
+def marching_cubes(float [:, :, :] im not None, double isovalue, LutProvider luts, int st=1, int classic=0,
```
rfezzani: Can I ask you @Leguark for a last small modification before merging? Please wrap these few lines to 80 characters length.. Thank you in advance ;-)

skimage/measure/_marching_cubes_lewiner_cy.pyx
```diff
@@ -929,13 +929,15 @@ cdef class LutProvider:
...
+                    np.ndarray[np.npy_bool, ndim=3, cast=True] mask=None):
```
rfezzani:
```suggestion
def marching_cubes(float[:, :, :] im not None, double isovalue,
                   LutProvider luts, int st=1, int classic=0,
                   np.ndarray[np.npy_bool, ndim=3, cast=True] mask=None):
```

skimage/measure/_marching_cubes_lewiner_cy.pyx
```diff
@@ -929,13 +929,15 @@ cdef class LutProvider:
...
     """ marching_cubes(im, double isovalue, LutProvider luts, int st=1, int classic=0)
-    This is the main entry to apply marching cubes.
+    This is the main entry to compute marching cubes.
+    Masked version of marching cubes. This function will check a masking array (same size as im) to decide if the
+    algorithm must be computed for a given voxel. This adds a small overhead that rapidly gets compensated by the
+    fewer computed cubes
```
rfezzani:
```suggestion
    Main entry to apply marching cubes.

    Masked version of marching cubes. This function will check a
    masking array (same size as im) to decide if the algorithm must be
    computed for a given voxel. This adds a small overhead that
    rapidly gets compensated by the fewer computed cubes
```

skimage/measure/_marching_cubes_lewiner_cy.pyx
Leguark: It was the optimization that @jni suggested in order to check is mask is None only once and not once per voxel. Regarding the wrap, sure no problem. I have it at 120 but if the whole project is at 80 makes sense

Leguark: Thank you for the code suggestions. Makes it much easier since I am on the road :)

Leguark:
I am not sure why Travis got stuck. @jni or @rfezzani can you trigger the testing again? Is there anything left or should I ping someone to merge the pull request?

Thank you again!

rfezzani:
:tada: CI finally all green. Merging! Thank you again @Leguark for your contribution.

PR #3836, opened 2019-04-12
[New feature] Majority voting filter

rfezzani:
## Description

This PR is related to a feature request in [scipy](https://github.com/scipy/scipy/issues/9873).
Based on the [answer](https://github.com/scipy/scipy/issues/9873#issuecomment-467726092) of @stefanv, I had some explorations.

A first implementation could be:

```python
import numpy as np

img = np.random.randint(255, (1024, 1024), dtype=np.uint8)
selem = np.ones((5, 5), dtype=bool)

res = rank.windowed_histogram(img, selem).argmax(axis=-1)
```

This provides good results: I measured 0.97sec average time over 10 runs.

I Tried to improve this using Cython (See my code).

```python
res = rank.majority_filter(img, selem)
```

On the same data, with the same settings, I measured an average time of 0.32 seconds.

I hope that this can be of any help.

## Checklist

  existing benchmark
- [X] Unit tests

## For reviewers

  later.


stefanv:
Your first version, the one-liner, is actually pretty sweet!  But, I would be OK with adding a new rank filter for this.

stefanv:
(Travis-CI failures are unrelated.)

jni:
@rfezzani thanks, very clean!

As a side note:
- we don't actually use the reference result added to the .npz file, as far as I can tell, so maybe we can remove it? (I like the test as it is.)
- I think the kernels should all have docstrings, but since this PR matches the standard of the original file, I'm happy to have it added.

Thanks again!

rfezzani:
Thank you @stefanv and @jni for your feed back. I am really happy if this can help!
Cheers.

soupault:
I suggest to rename this from `majority_filter` to `majority`. Opinions?

stefanv:
Yes, that is a good suggestion.

rfezzani:
yes, `majority` is better indeed.
@jni, the .npz file is still in use, `test_rank_filter` was causing CI failure, that's why I added a reference result for the majority filter.
I also agree that the kernels should have docstrings, it wasn't that obvious to understand the framework of rank filters.

soupault:
See https://github.com/scikit-image/scikit-image/pull/3839

PR #3837, opened 2019-04-12
Force imageio reader to return NumPy arrays

stefanv:
In newer versions of `imageio`, they now return an ndarray subclass with a `.meta` attribute.  We moved away from this kind of representation many years ago (yes, we used to have an Image object!), because these subclassed arrays do not survive computation through a typical pipeline.

### DIFF
--- a/skimage/io/_plugins/imageio_plugin.py
+++ b/skimage/io/_plugins/imageio_plugin.py
@@ -4,12 +4,6 @@ from functools import wraps
 import numpy as np
 from imageio import imread as imageio_imread, imsave
 
-
-def array_return(f):
-    @wraps(f)
-    def wrapper(*args, **kwargs):
-        return np.asarray(f(*args, **kwargs))
-    return wrapper
-
-
-imread = array_return(imageio_imread)
+@wraps(imageio_imread)
+def imread(*args, **kwargs):
+    return np.asarray(imageio_imread(*args, **kwargs))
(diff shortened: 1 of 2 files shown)


jni:
@stefanv btw do you know what *we* changed between 0.14 and 0.15 that meant that we used to return an array even with new imageio versions?

jni:
@scikit-image/core This is probably a critical bug fix (see report in [zulip](https://skimage.zulipchat.com/#narrow/stream/175598-general/topic/type.20changes.20in.200.2E15.2E0/near/163042890), so please review ASAP!

stefanv:
Can you think of a simpler wrapper that preserves the docstring?  I'd be happy to switch.

hmaarrfk:
Basically, using wraps on a function definition is simpler.

jni:
@stefanv, @hmaarrfk provided the full alternative implementation above. I'm inclined to agree with him that it's simpler and better *unless* we envision adding lots of plugins *and* that many will need the decorator. I hope we don't get there!

stefanv:
That's fine with me; especially once you fix the capitalized R in return ;)  Go ahead and push to my branch.

hmaarrfk:
Thanks for the input everybody. Merging.

hmaarrfk:

soupault:

soupault:
@Carreau Asking for backporting twice is, probably, stupid, but you might be interested to take a look at the logs :).

Carreau:
Sorry missed the notification. 
Feel free to email me next time.

PR #3838, opened 2019-04-12
Documentation formatting / compilation fixes

stefanv:
Seems to work with Sphinx 2.0

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,6 +1,4 @@
-# Sphinx 1.7.8 has a bug when building on cached envs
-# https://github.com/sphinx-doc/sphinx/issues/5361
-sphinx>=1.3,!=1.7.8
+sphinx>=2.0
 numpydoc>=0.6
 sphinx-gallery
 sphinx-copybutton
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,5 +1,5 @@
 sphinx>=2.0
-numpydoc>=0.7
+numpydoc==0.7
 sphinx-gallery
 sphinx-copybutton
 pytest-runner
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,6 +1,6 @@
 sphinx>=2.0
 numpydoc>=0.7
-sphinx-gallery>=0.3
+sphinx-gallery
 sphinx-copybutton
 pytest-runner
 scikit-learn
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,6 +1,6 @@
 sphinx>=2.0
-numpydoc>=0.6
-sphinx-gallery
+numpydoc>=0.7
+sphinx-gallery>=0.3
 sphinx-copybutton
 pytest-runner
 scikit-learn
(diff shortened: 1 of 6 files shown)


jni:
@stefanv does this fix the current rendering issues with arg names and types being smushed together?

stefanv:
I didn't change anything, but it renders correctly for me.

stefanv:
I can't replicate the Travis-CI failure on my system; anyone else?

jni:
@stefanv I think you can download a docker image of the travis machine? That's probably the way to go. (I might be able to give that a go in a couple of weeks, but not till then.)

hmaarrfk:
I get a similar error.
```
pickling environment... /home/mark2/git/scikit-image/doc/source/auto_examples/color_exposure/plot_rgb_to_hsv.rst:26: WARNING: Explicit markup ends without a blank line; unexpected unindent.
/home/mark2/git/scikit-image/doc/source/auto_examples/segmentation/plot_thresholding.rst:21: WARNING: Explicit markup ends without a blank line; unexpected unindent.
```
but the warning isn't being treated as an error.

hmaarrfk:
The first comment in the file cannot be immediately after the file docstring...

requirements/docs.txt
```diff
@@ -1,7 +1,5 @@
-# Sphinx 1.7.8 has a bug when building on cached envs
-# https://github.com/sphinx-doc/sphinx/issues/5361
-sphinx>=1.3,!=1.7.8
-numpydoc>=0.6
+sphinx>=2.0
+numpydoc==0.7
```
hmaarrfk: Do we know if these changes are still necessary

jni: We don't, but I don't think it's a big deal — docs are mostly for dev work.

jni: Happy to approve a different PR with just your last commit, if you want to issue such a thing. =)

hmaarrfk: I'll undo this if it fails. It probably will. I'm mostly worried about the pin on numpydocs

stefanv:
There's absolutely no reason to run on an older version of numpydoc!

stefanv:
Also, should this be considered a bug in sphinx gallery?

jni:
@stefanv no, but no reason to force devs to update their existing environment if it's not needed, either. The requirements have a `>=` so it's fine to keep the older version supported if nothing is breaking. I'm going to go ahead and merge, we can move up any time we deem necessary. 

Re sphinx-gallery, might as well report it, but it seems like a design choice. But the error message could have been better!

hmaarrfk:
Yeah, my greater fear is the "pin", setting `==0.7` would have made it more difficult to detect changes in numpydoc where we would need to adapt.

stefanv:
`= 0.7` would be fine; but numpydoc gets fairly crucial fixes.  A lot of reference generation bugs have been fixed between 0.6 and 0.7, for example.

stefanv:
@hmaarrfk I'm a bit confused; it doesn't look like my patch was proposing pinning to a specific version.  So if you're OK with `>= 0.7`, let's do that.

hmaarrfk:
Yeah. Greater or equal is better.

I think you had `==` maybe by accident

hmaarrfk:

hmaarrfk:

PR #3850, opened 2019-04-18
Add mask support for slic

### LINKED ISSUE #1569: Consider adding mask support for SLIC
I would like to know if is possible to perform the SLIC segmentation on a masked array, it means to give the mask as a parameter to the method. 

This is important when you want to segment within a region of interest. 

Thank you in advance

### LINKED ISSUE #3722: Consider returning segments starting from 1 from SLIC
We are getting close to being fully consistent in considering 0 to be the background in a label image. However, SLIC returns segments starting from 0, resulting in problems like this SO question:

https://stackoverflow.com/questions/54506079/slic-segmentation-in-skimage/54526084

where SLIC returns a full segmentation of the image, but segment 0 is ignored by regionprops. (Note: maybe the background segment should be measured by regionprops anyway!)

## Way to reproduce

Input image:

example-image

```python
import imageio
import numpy as np
import matplotlib.image as mpimg
from skimage.measure import regionprops
from skimage.segmentation import slic
from skimage.segmentation import mark_boundaries

def rgb2gray(rgb):
    return np.dot(rgb[..., :3], [0.2126, 0.7152, 0.0722])

img_file_path = 'example-image.jpg'
image = imageio.imread(img_file_path)
segments_slic = slic(image, n_segments=250, compactness=100)
regions = regionprops(segments_slic, intensity_image=rgb2gray(image))
for props in regions:
    cy, cx = props.centroid
    plt.plot(cx, cy, 'ro')

plt.imshow(mark_boundaries(image, segments_slic))
plt.show()
```

Output:

figure_1

Note the missing segment annotation on the top left. Adding `1 + ` to the SLIC call fixes this.

```python
3.6.3 |Anaconda custom (x86_64)| (default, Nov  3 2017, 12:34:11)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]
Darwin-18.2.0-x86_64-i386-64bit
```

rfezzani:
## Description

Closes #1569, Closes #3722

To address #1569 and based on the work of @benjaminirving on [maskSLIC](https://github.com/benjaminirving/maskSLIC), I added the mask support to slic.

The method is outlined [here](https://arxiv.org/abs/1606.09518).

~~Masked area is label `-1` in the returned labels. This adds some confusion to the discussion #3722 :-).~~

## Checklist

  existing benchmark
- [X] Unit tests

## For reviewers

  later.


skimage/segmentation/_slic.pyx
```diff
@@ -78,8 +82,8 @@ def _slic_cython(double[:, :, :, ::1] image_zyx,
     step_z, step_y, step_x = [int(s.step if s.step is not None else 1)
                               for s in slices]
 
-    cdef Py_ssize_t[:, :, ::1] nearest_segments \
-        = np.empty((depth, height, width), dtype=np.intp)
+    cdef int[:, :, ::1] nearest_segments \
+        = np.full((depth, height, width), -1, dtype=np.int32)
```
jni: Why did you change this dtype?

rfezzani: @jni , first of all, thank you for your review. I tried to stick as much as possible to the reference implementation, so some of the implementation decisions come from this.

1. I agree with you, introducing this feature can be the occasion to also close #3722 by considering that masked areas are `0` labelled.
2. segments dtype did not change, it is still `double`. `bint` for the mask seems a better choice. The `mask.size` test is indeed cleaner.
3. The `only_dist` parameter is for updating centroids positions without considering pixels color (it is the step 2 in the maskSLIC algorithm). It makes me think that renaming the parameter `ignore_color` is may be more clear.
4. `regular_grid` reimplementation is just a matter of code style and code lines count reduction :-).

Concerning the last dtype modification that you point, labeling masked area with `-1` in the current implementation prevents the use of the unsigned `Py_size_t`.

I will tak eyour remarks into consideration and come with an other proposition soon.
Thanks again.

hmaarrfk: FYI: Py_ssize_t = signed size type

https://www.python.org/dev/peps/pep-0353/#specification

rfezzani: THX, the double `s` makes sens then :-)

hmaarrfk: If that makes sense to you, maybe you can help me make sense of these 
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-s-snprintf-s-l-snwprintf-s-snwprintf-s-l?view=vs-2019

skimage/segmentation/_slic.pyx
```diff
@@ -129,16 +140,22 @@ def _slic_cython(double[:, :, :, ::1] image_zyx,
                         dy = sy * (cy - y)
                         dy *= dy
                         for x in range(x_min, x_max):
+
+                            if mask_size > 0 and mask[z, y, x] == 0:
```
hmaarrfk: I'm pretty sure C will be able to compile out this condition, but maybe not for all compile options.

hmaarrfk:
Could you please add a benchmark? I'm not saying that "I will only accept things that don't slow down scikit-image", but it would be good to at least understand how much "we will slow people down" if we add these conditions in the tight loop.

skimage/segmentation/_slic.pyx
rfezzani: :-)

rfezzani:
The last commit addresses the issue #3722 by deprecating the current behavior when no mask is given as suggested by @jni

skimage/segmentation/_slic.pyx
rfezzani: I agree. I added some brackets.

rfezzani:
@hmaarrfk, I added a benchmark. I did not observed any significant performance drop if no mask is provided.
If a mask is provided, computing seeds positions may be slow, because a distance map over the mask is computed for each seed.
If the performance is an issue for this feature, one may consider computing seeds positions at lower scale, but this is not in the reference implementation of @benjaminirving.

doc/examples/segmentation/plot_mask_slic.py
```diff
@@ -0,0 +1,69 @@
...
+
+# SLIC result
+slic = segmentation.slic(img, n_segments=200)
+
+# maskSLIC result
+m_slic = segmentation.slic(img, n_segments=100, mask=mask)
+
+# Display result
+fig = plt.figure(figsize=(10, 10))
+ax1, ax2, ax3, ax4 = fig.subplots(2, 2, sharex=True, sharey=True).ravel()
```
emmanuelle: this lines makes test fail (I think it does not work with older version of matplotlib). Could you please use 
```
ax, ... = plt.subplots(...., figsize=(10, 10))
```
?

rfezzani: It's fixed, THX for the review @emmanuelle

rfezzani:
Hi @jni and @hmaarrfk , any decision on the integration of this feature? I simply would know if more work is needed on it ;-)
Thx

doc/examples/segmentation/plot_mask_slic.py
```diff
@@ -0,0 +1,67 @@
+"""
+======================
+maskSLIC Demonstration
+======================
+
+This example is about comparing the segmentations obtained using the
+plain SLIC method [1]_ and its masked version maskSLIC [2]_.
+
```
emmanuelle: Could you please add one sentence here explaining the difference between the two implementations, like the fact that there are less border effects in the maskSLIC? We don't expect users to go and read the publication :-), so let's help them as much as possible.

doc/examples/segmentation/plot_mask_slic.py
```diff
@@ -0,0 +1,67 @@
...
+# Display result
+fig, ax_arr = plt.subplots(2, 2, sharex=True, sharey=True, figsize=(10, 10))
+ax1, ax2, ax3, ax4 = ax_arr.ravel()
+
+ax1.imshow(img)
+ax1.set_title("Origin image")
+
+ax2.imshow(mask, cmap="gray")
+ax2.set_title("Considered mask")
```
emmanuelle: Just "mask" is fine I think.

skimage/segmentation/slic_superpixels.py
```diff
@@ -60,6 +144,9 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
         in most of the cases.
     slic_zero: bool, optional
         Run SLIC-zero, the zero-parameter mode of SLIC. [2]_
+    mask : 2D ndarray, optional
+        if provided, seed points are placed following the strategy
```
emmanuelle: if provided, superpixels are computed only for ``mask=True``, and seed points are placed following the strategy described in _[3], mitigating border effects at the border of the mask.

emmanuelle:
Thank you for your changes @rfezzani. I didn't have the time to review all the code but I added two comments about the example and the documentation, I'll finish the review tonight hopefully.

skimage/segmentation/slic_superpixels.py
```diff
@@ -1,17 +1,101 @@
...
+    coord = np.asarray(np.nonzero(mask))
+    bbox = coord.min(-1), coord.max(-1) + 1
+    roi = mask[bbox[0][0]: bbox[1][0],
+               bbox[0][1]: bbox[1][1],
+               bbox[0][2]: bbox[1][2]].copy()
+
+    centroids = np.repeat([bbox[0]], n_centroids, axis=0)
+
+    for idx in range(n_centroids):
+        dist_map = ndi.distance_transform_edt(roi, sampling=spacing)
```
emmanuelle: I did a quick profiling of the gallery example, and this line is taking 75% of the total CPU time. Would it be OK to replace it by ``distance_transform_cdt`` which is slightly less accurate but would still be OK for the algorithm (I think)? On the gallery example the result looks as good, and it allows to gain a 2.5 factor (from 2.6 s to 1s on my box).

emmanuelle: or we could use euclidean when spacing is not None, and cdt otherwise for better performance.

emmanuelle:
I left another comment / question regarding performance, but otherwise it looks good to me and I think it will be a useful addition.

skimage/segmentation/slic_superpixels.py
rfezzani: Good suggestion, using `distance_transform_cdt` should not affect the quality of the result.

rfezzani:
Thank you @emmanuelle for your review, I made the suggested changes.

skimage/segmentation/slic_superpixels.py
```diff
@@ -1,17 +1,111 @@
...
+        Spacing of elements along each dimension.
+
+    Returns
+    -------
+    centroids : 2D ndarray
+        The coordinates of the centroids with shape (n_centroids, 3).
+    steps : 1D ndarray
+        The approximate distance between two seeds in all dimensions.
+
+    """
```
jni: Would you mind explaining the sampling strategy in more detail here? I don't actually understand what this function is doing. (I do intend to read the paper, but I think for our users, it would be nice if the strategy was explained somewhere. Probably in the argument for `mask=` in the public-facing `slic` function.

rfezzani: I added some comments to the `_get_mask_centroids` function explaining the algorithm and modified the doc of the `mask` argument.
I also added a note in the doc explaining the behavior of the function according to the `spacing` argument value introduced by the [suggestion](https://github.com/scikit-image/scikit-image/pull/3850#discussion_r289618688) of @emmanuelle

skimage/segmentation/slic_superpixels.py
```diff
@@ -1,17 +1,119 @@
...
+
+    # Get tight ROI around th emask to optimize
```
emmanuelle: the mask

skimage/segmentation/slic_superpixels.py
```diff
@@ -1,17 +1,119 @@
...
+        # Exact euclidean distance transform.
+        dist_map = np.empty_like(roi, dtype=np.float64)
+        spacing = np.ascontiguousarray(spacing, dtype=np.float64)
+        update_dist_map = functools.partial(ndi.distance_transform_edt,
+                                            distances=dist_map,
+                                            sampling=spacing)
+
+    centroids = np.repeat([bbox[0]], n_centroids, axis=0)
+
+    # Iteratively place centroids in teh mask
```
emmanuelle: the mask

skimage/segmentation/slic_superpixels.py
```diff
@@ -1,17 +1,119 @@
...
+
+    # Get tight ROI around th emask to optimize
+    coord = np.asarray(np.nonzero(mask))
+    bbox = coord.min(-1), coord.max(-1) + 1
+    roi = mask[bbox[0][0]: bbox[1][0],
+               bbox[0][1]: bbox[1][1],
+               bbox[0][2]: bbox[1][2]].copy()
+
+    if spacing is None:
+        # Chamfer distance transform.
```
emmanuelle: # Chamfer distance transform (faster).

emmanuelle:
LGTM apart from two typos.

benjaminirving:
@rfezzani, thanks for the work proposing this code for merging into scikit-image. Let me know if any support is needed from the method development side. It's been a while since I worked on this method.

rfezzani:
Hi @benjaminirving, thanks for your proposition. If you followed the previous discussion, there is a kind of concern about computational time.
An optimization proposed by @emmanuelle is the replacement of the Euclidean distance transform by the Chamfer distance.
I also proposed to estimate seeds position at lower scale, _ie_ by reducing mask size, detecting seeds position and transforming these positions again to the original mask scale. But this strategy may need some more investigations.
You are welcome if you want to try this, or if you have any other suggestion ;-)

benjaminirving:
@rfezzani, Thanks :). It's been a busy period so I haven't kept up with the discussion. 

Chamfer distance could definitely be an option. Scaling might cause problems in very small regions. this method has been applied to functional images where the region is made up of only a small number of voxels.

rfezzani:
@benjaminirving, you are right, scaling causes problems for small regions for sure!
Scaling can be implemented as an option, and a warning may be raised if it is activated  using a too small mask.
The option for using Chamfer distance is already implemented in this PR.

skimage/segmentation/slic_superpixels.py
```diff
@@ -122,10 +237,36 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
...
+            raise ValueError("image and mask should have the same shape.")
+        # Step 1 of the algorithm [3]_
+        centroids, steps = _get_mask_centroids(mask, n_segments, spacing)
+        update_centroids = True
+        start_label = 1
+    else:
+        centroids, steps = _get_grid_centroids(image, n_segments)
+        mask = np.ones((0, 1, 1), dtype=np.bool)
+        start_label = 0
+        warnings.warn("Labels' indexing start from 0. " +
```
hmaarrfk: I don't think this should be the way to handle deprecations.

A user should be able to write code on the day the warning is issued that will do two things:
1. Silence the warning.
2. Ensure their code stays correct when the scikit-image version gets upgraded.

skimage/segmentation/tests/test_slic.py
```diff
@@ -177,6 +187,7 @@ def test_enforce_connectivity():
     assert_equal(segments_connected_low_max, result_connected)
 
 
+@pytest.mark.filterwarnings("ignore:labels")
```
hmaarrfk: Thanks for ensuring that the warnings stay silenced during tests, but these do highlight the fact that even we all feel that warnings are obtrusive. I would rather there be a more user friendly way to silence these warnings, and suggest it in the warning itself.

skimage/segmentation/slic_superpixels.py
rfezzani: Hello @hmaarrfk, I am not sure to understand your comment here.
1. to silence the warning, the user may use `warnings.catch_warnings`
2. the warning message is made for that! The user is informed and he is responsible of making his code correct when skimage gets upgraded. "We are all consenting adults" :-D
Now, the warning message has to be as clear as possible and it is may be actually not. It basically misses the version number when the change will be made.
We can also think about exposing the `start_label` as a parameter of the function set to `0` by default (to keep the actual behavior) and change it to `1` when we decide to address #3722 
What do you think?

skimage/segmentation/tests/test_slic.py
rfezzani: Warnings are obtrusive, but if we decide to change the behavior of a function, it is important to inform the user ;-)

skimage/segmentation/slic_superpixels.py
rfezzani: My opinion is close to what you think about the spirit of warnings:
1. Receive a warning in 0.16, silence the warning and if you mind about it, take an action that make your code valid anyway.
2. That action is valid in 0.17 and the warning won't appear
3. No future action is necessary in 0.18 to fix the code.

:smile: 

By the way, it is somehow strange to introduce an additional parameter, but it's fine.

hmaarrfk:
Yeah, but you can add a note as to why it was done.

Keep in mind, that the nice thing about `start_label=0` means that
subsequent steps in the analysis pipeline of old code stay valid.

Forcing people to use `start_label=1` might be better for the **future**,
but it breaks past, working code of people who are likely advocates of
scikit-image. Given them an **easy way** to silence warnings **and** to
ensure it will continue working in 1 year (or two) is really important to
the library.

skimage/segmentation/slic_superpixels.py
jni: @rfezzani,

> to silence the warning, the user may use warnings.catch_warnings

This is a very verbose way to handle warnings, and most Python users won't be familiar with this. It's particularly cumbersome to use in interactive settings. A keyword argument is much easier for people to adapt to.

> "We are all consenting adults"

Even adults get annoyed at having to do extra work. =) I think @hmaarrfk's principles are exactly correct: warnings should result in a changed function call that still does what the user wants and that will work "forever". We haven't kept to this in the past but I really think it's the best user experience.

rfezzani:
I made the requested changes, I added a note explaining the introduction of `start_label` parameter and requested it to be 0 or 1 and not only positive. Do you think that we should give much freedom to the user by accepting any positive value?

skimage/segmentation/slic_superpixels.py
rfezzani: :sweat_smile: I still believe that retro-compatibility is most of the time a lot of effort impacting a very small portion of user.
Code versioning is made to handle that, and people needing such stability usually fix software versions.
Now I understand your point of view, It is really a matter of choice.

skimage/segmentation/tests/test_slic.py
rfezzani: Do you went me to set `start_label` to 0 explicitly in the tests or do we update test results when the behavior changes?
Do we need a test for the validation of error raise if `start_label` is not 0 or 1?

rfezzani:
I identified an issue in slic (#4274). I would be happy to work on it, but it will be in conflict with this PR. @hmaarrfk, @jni, do you think this PR still needs more refinement?

jni:
@rfezzani I'm sorry, I don't have time to review this properly right now. It's a great addition but I am worried that it imposes a significant performance penalty on non-masked SLIC (because a mask array needs to be instantiated either way), and I think there's probably a way to achieve this without needing that array.

So my suggestion would be to fix #4274 first and then rebase/merge this on that. I hope that the fix will not be so onerous that a rebase would be hard...

rfezzani:
Hi @jni, thank you for taking time to answer. The instantiated mask is of shape (0, 1, 1):
```python
mask = np.ones((0, 1, 1), dtype=np.bool)
```
so there is no memory allocation here.

jni:
Fascinating! As I said, clearly I need to take a closer look at this! :joy:

rfezzani:
> Hi @jni, thank you for taking time to answer. The instantiated mask is of shape (0, 1, 1):
> 
> ```python
> mask = np.ones((0, 1, 1), dtype=np.bool)
> ```
> 
> so there is no memory allocation here.

@jni, I removed this weired initialization ;-)

rfezzani:
I fixed merge conflicts, if any one wants to review :wink:

rfezzani:
@scikit-image/core, this almost one year old PR is MRG+1 for a while, if by any chance someone can find some time for a review, thx :slightly_smiling_face:

emmanuelle:
@jni would you have the time to review this one? I can read the PR again to talk about it with you if you wish.

skimage/segmentation/_slic.pyx
```diff
@@ -200,18 +236,21 @@ def _slic_cython(np_floats[:, :, :, ::1] image_zyx,
...
 def _enforce_label_connectivity_cython(Py_ssize_t[:, :, ::1] segments,
                                        Py_ssize_t min_size,
-                                       Py_ssize_t max_size):
+                                       Py_ssize_t max_size,
+                                       Py_ssize_t start_label=1):
     """ Helper function to remove small disconnected regions from the labels
 
     Parameters
     ----------
-    segments : 3D array of int, shape (Z, Y, X)
```
stefanv, line 208: I think all these are wrong now.

skimage/segmentation/_slic.pyx
```diff
@@ -183,7 +212,14 @@ def _slic_cython(np_floats[:, :, :, ::1] image_zyx,
...
 
-                            k = nearest_segments[z, y, x]
+                            if use_mask:
+                                if not mask[z, y, x]:
+                                    continue
+
+                                if nearest_segments[z, y, x] == mask_label:
+                                    continue
+
+                            k = nearest_segments[z, y, x]-start_label
```
stefanv, line 222:
```suggestion
                            k = nearest_segments[z, y, x] - start_label
```

skimage/segmentation/_slic.pyx
```diff
@@ -200,18 +236,21 @@ def _slic_cython(np_floats[:, :, :, ::1] image_zyx,
...
-    segments : 3D array of int, shape (Z, Y, X)
+    segments: 3D array of int, shape (Z, Y, X)
         The label field/superpixels found by SLIC.
     min_size: int
         Minimum size of the segment
     max_size: int
         Maximum size of the segment. This is done for performance reasons,
         to pre-allocate a sufficiently large array for the breadth first search
+    start_label: int
+        label indexing start value.
```
stefanv, line 253: Full, capitalized sentence here, please, and some more clarity on what it means (I understand, but not sure other readers would).

skimage/segmentation/_slic.pyx
```diff
@@ -200,18 +236,21 @@ def _slic_cython(np_floats[:, :, :, ::1] image_zyx,
...
+                                       Py_ssize_t start_label=1):
     """ Helper function to remove small disconnected regions from the labels
 
     Parameters
     ----------
-    segments : 3D array of int, shape (Z, Y, X)
+    segments: 3D array of int, shape (Z, Y, X)
         The label field/superpixels found by SLIC.
     min_size: int
         Minimum size of the segment
```
stefanv, line 248: Let's fix these to full sentences too, while we're at it.

skimage/segmentation/slic_superpixels.py
```diff
@@ -1,17 +1,119 @@
...
 from ..color import rgb2lab
 
 
+def _get_mask_centroids(mask, n_centroids, spacing=None):
+    """Find regularly spaced centroids on a mask.
+
+    Parameters
+    ----------
+    mask : 3D ndarray
+        The mask where the centroids must be positioned.
```
stefanv, line 19:
```suggestion
        The mask within which the centroids must be positioned.
```

skimage/segmentation/slic_superpixels.py
```diff
@@ -1,17 +1,119 @@
...
+    """
+
+    # Get tight ROI around the mask to optimize
+    coord = np.asarray(np.nonzero(mask))
+    bbox = coord.min(-1), coord.max(-1) + 1
+    roi = mask[bbox[0][0]: bbox[1][0],
+               bbox[0][1]: bbox[1][1],
+               bbox[0][2]: bbox[1][2]].copy()
+
+    if spacing is None:
```
stefanv, line 41: Why not calculate default spacing and then use same distance metric for both?

skimage/segmentation/slic_superpixels.py
```diff
@@ -1,17 +1,119 @@
...
+from scipy.spatial.distance import pdist, squareform
 
+from ._slic import (_slic_cython, _enforce_label_connectivity_cython)
 from ..util import img_as_float, regular_grid
-from ..segmentation._slic import (_slic_cython,
-                                  _enforce_label_connectivity_cython)
 from ..color import rgb2lab
 
 
+def _get_mask_centroids(mask, n_centroids, spacing=None):
```
stefanv, line 13: Where did you find this technique?  Are you sure it's close to somewhat optimal?  Have you investigated Lloyd's algorithm and others?

skimage/segmentation/slic_superpixels.py
```diff
@@ -60,6 +162,13 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
         in most of the cases.
     slic_zero : bool, optional
         Run SLIC-zero, the zero-parameter mode of SLIC. [2]_
+    start_label: int, optional
+        The labels' index start. Should be 0 or 1.
+    mask : 2D ndarray, optional
+        if provided, superpixels are computed only where mask is True,
```
stefanv, line 168:
```suggestion
        If provided, superpixels are computed only where mask is True,
```

skimage/segmentation/slic_superpixels.py
```diff
@@ -124,12 +248,46 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
...
+        if image.shape[-1] != 3 and convert2lab:
+            raise ValueError("Lab colorspace conversion requires a RGB image.")
+        elif image.shape[-1] == 3:
+            image = rgb2lab(image)
+
+    if start_label is None:
+        if use_mask:
+            start_label = 1
+        else:
+            warnings.warn("Labels' indexing start from 0. " +
```
stefanv, line 261:
```suggestion
            warnings.warn("skimage.measure.label's indexing starts from 0. " +
```

skimage/segmentation/slic_superpixels.py
```diff
@@ -124,12 +248,46 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
...
+                          "In future version it will start from 1. " +
+                          "To disable this warning, explicitely " +
+                          "specify the start_label parameter.",
```
stefanv, line 264:
```suggestion
                          "set the `start_label` parameter to 1.",
```
?

skimage/segmentation/tests/test_slic.py
```diff
@@ -208,5 +212,254 @@ def test_more_segments_than_pixels():
...
+    result_disconnected = np.array([[-1, 0, 0, 1, 1, -1],
+                                    [-1, 0, 0, 1, 1, -1],
+                                    [-1, 0, 0, 1, 1, -1]], np.float) + 1
+
+    assert_equal(segments_connected, result_connected)
+    assert_equal(segments_disconnected, result_disconnected)
+    assert_equal(segments_connected_low_max, result_connected)
+
+
+def test_slic_zero_mask():
```
stefanv, line 370: Do you have a test that compares slic to maskSLIC for a pass-through mask?

skimage/segmentation/slic_superpixels.py
rfezzani: This is the strategy used in maskSLIC for initializing the centroids. I did not made more investigation but you are right, K-means clustering approaches may be appliyed here also...

skimage/segmentation/slic_superpixels.py
rfezzani: Sorry, I am not sure to understand what you mean here. `spacing` is an input to `slic` function.

skimage/segmentation/tests/test_slic.py
rfezzani: The two implementations uses different approaches for the centroid initialization, so I guess that the output will be different...

skimage/segmentation/_slic.pyx
rfezzani: Do you mean the white space before the colons ` : `?

skimage/segmentation/slic_superpixels.py
rfezzani: @stefanv, I made some tests, and `scipy.cluster.vq.kmeans2` is an excellent alternative, thank you again for this suggestion. I will come with some more experimental results later ;-)

stefanv: Nice!

skimage/segmentation/slic_superpixels.py
stefanv: I mean, you can calculate `spacing` if it is not defined, and then use the same code path, whether it was specified or not.

skimage/segmentation/_slic.pyx
stefanv: Exactly.

skimage/segmentation/slic_superpixels.py
rfezzani: The use of `distance_transform_cdt` instead of `distance_transform_edt` was [suggested](https://github.com/scikit-image/scikit-image/pull/3850#discussion_r289618606) by @emmanuelle to reduce the computation time.
`spacing` is in the original API of `slic` and it is usable only with `distance_transform_edt`, that why I assumed that if the user doesn't specify `spacing`, the fast path must be used (`distance_transform_cdt`) and otherwise the path without approximation is used (`distance_transform_edt`).

skimage/segmentation/slic_superpixels.py
rfezzani: My investigation on this topic exposed a bug in the current strategy. I ran the following lines:
```python

In [1]: from skimage.segmentation.slic_superpixels import _get_mask_centroids                                          

In [2]: import numpy as np                                                                                             

In [3]: mask = np.zeros((1, 100, 100), dtype=bool)                                                                     

In [4]: mask[0, 20:60, 30:90] = True                                                                                   

In [5]: c, _ = _get_mask_centroids(mask, 100)

In [6]: P.imshow(mask[0])                                                                                              
Out[6]: <matplotlib.image.AxesImage at 0x7f14e803b150>

In [7]: P.plot(c[:, 2], c[:, 1], 'rx')                                                                                 
Out[7]: [<matplotlib.lines.Line2D at 0x7f14e8e81390>]
```

With the current implementation I obtained
bug
I fixed the bug, here is what I obtained:
fix
Not really good :slightly_frowning_face:!
I implemented the suggestion of @stefanv, and here is the result:
kmeans
Much better :slightly_smiling_face: !
Moreover, the timing performances are improved:
- old:
```python
In [5]: %timeit _get_mask_centroids(mask, 100)                                                                         
17.9 ms ± 192 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
- new:
```python
In [8]: %timeit _get_mask_centroids(mask, 100)                                                                         
14.2 ms ± 107 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

emmanuelle:
About the kmeans algorithm to get the centroids, probably you can accelerate this part (15 ms on a 100*100 array is not negligible) by using a very small number of iterations (the default number is 10 in scipy's algo, 2 or 3 should be enough I believe). Disclaimer: there is a somewhat mysterious sentence in the docstring which I don't fully understand, maybe @stefanv would have a better understanding of it: 
```
iter : int, optional
    Number of iterations of the k-means algorithm to run. Note
    that this differs in meaning from the iters parameter to
    the kmeans function.
```

stefanv:
> About the kmeans algorithm to get the centroids, probably you can accelerate this part (15 ms on a 100*100 array is not negligible) by using a very small number of iterations (the default number is 10 in scipy's algo, 2 or 3 should be enough I believe). Disclaimer: there is a somewhat mysterious sentence in the docstring which I don't fully understand, maybe @stefanv would have a better understanding of it:
> 
> ```
> iter : int, optional
>     Number of iterations of the k-means algorithm to run. Note
>     that this differs in meaning from the iters parameter to
>     the kmeans function.
> ```

I found this docstring:

```
iter : int, optional
   The number of times to run k-means, returning the codebook
   with the lowest distortion. This argument is ignored if
   initial centroids are specified with an array for the
   ``k_or_guess`` parameter. This parameter does not represent the
   number of iterations of the k-means algorithm.
```

I think it refers to the fact that kmeans is randomly initialized, so you can run kmeans to convergence several times over for different starting conditions, and then see which codebook is best?

emmanuelle:
@stefanv thank you for the explanation. So maybe this parameter was not what I thought and we can leave optimization of this kmeans for future work.

rfezzani:

emmanuelle:
I already approved so maybe someone can take a look ?

skimage/segmentation/_slic.pyx
```diff
@@ -137,25 +152,31 @@ def _slic_cython(np_floats[:, :, :, ::1] image_zyx,
...
                             dist_color = 0
                             for c in range(3, n_features):
                                 t = image_zyx[z, y, x, c - 3] - segments[k, c]
                                 dist_color += t * t
+
                             if slic_zero:
-                                dist_center += dist_color / max_dist_color[k]
-                            else:
+                                dist_color /= max_dist_color[k]
+
```
grlee77, line 169: It looks like this color distance computation can be moved within the `if not ignore_color` block below

skimage/segmentation/_slic.pyx
```diff
@@ -137,25 +152,31 @@ def _slic_cython(np_floats[:, :, :, ::1] image_zyx,
...
+                            if not ignore_color:
                                 dist_center += dist_color
 
                             if distance[z, y, x] > dist_center:
-                                nearest_segments[z, y, x] = k
+                                nearest_segments[z, y, x] = k+start_label
```
grlee77, line 174:
```suggestion
                                nearest_segments[z, y, x] = k + start_label
```

skimage/segmentation/_slic.pyx
```diff
@@ -166,7 +187,15 @@ def _slic_cython(np_floats[:, :, :, ::1] image_zyx,
...
-                        k = nearest_segments[z, y, x]
+
+                        if use_mask:
+                            if not mask[z, y, x]:
+                                continue
+
+                            if nearest_segments[z, y, x] == mask_label:
+                                continue
+
+                        k = nearest_segments[z, y, x]-start_label
```
grlee77, line 198:
```suggestion
                        k = nearest_segments[z, y, x] - start_label
```

skimage/segmentation/slic_superpixels.py
```diff
@@ -88,12 +174,23 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
       interpret them as 3D with the last dimension having length 3, use
       `multichannel=False`.
 
+    * If spacing is None and mask is provided, chamfer distance
+      transform is used to homogeneously place seed points over the
+      mask, otherwise, the exact euclidian distance transform is used.
```
grlee77, line 179: isn't this text outdated? I think you switched to using k-means?

skimage/segmentation/slic_superpixels.py
```diff
@@ -88,12 +174,23 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
...
+    * `start_label` is introduced to handle the issue [4]_. The labels
+      indexing starting at 0 will be deprecated in future versions. If
+      `mask` is not `None` labels indexing starts at 1 and masked area
+      is set to 0.
+
     References
     ----------
     .. [1] Radhakrishna Achanta, Appu Shaji, Kevin Smith, Aurelien Lucchi,
         Pascal Fua, and Sabine Süsstrunk, SLIC Superpixels Compared to
         State-of-the-art Superpixel Methods, TPAMI, May 2012.
```
grlee77, line 190:
```suggestion
        State-of-the-art Superpixel Methods, TPAMI, May 2012.
        :DOI:`10.1109/TPAMI.2012.120`
```

doc/examples/segmentation/plot_mask_slic.py
```diff
@@ -0,0 +1,71 @@
...
+overcome border problems that affects SLIC method, particularely in
+case of irregular mask.
+
+.. [1] Radhakrishna Achanta, Appu Shaji, Kevin Smith, Aurelien Lucchi,
+    Pascal Fua, and Sabine Suesstrunk, SLIC Superpixels Compared to
+    State-of-the-art Superpixel Methods, TPAMI, May 2012.
+
+.. [2] Irving, Benjamin. "maskSLIC: regional superpixel generation
+    with application to local pathology characterisation in medical
+    images." arXiv preprint arXiv:1606.09518 (2016).
```
grlee77, line 20:
```suggestion
    images.", 2016, , :arXiv:`1606.09518`
```

doc/examples/segmentation/plot_mask_slic.py
```diff
@@ -0,0 +1,71 @@
...
+plain SLIC method [1]_ and its masked version maskSLIC [2]_.
+
+The maskSLIC method is an extension of the SLIC method for the
+generation of superpixels in a region of interest. maskSLIC is able to
+overcome border problems that affects SLIC method, particularely in
+case of irregular mask.
+
+.. [1] Radhakrishna Achanta, Appu Shaji, Kevin Smith, Aurelien Lucchi,
+    Pascal Fua, and Sabine Suesstrunk, SLIC Superpixels Compared to
+    State-of-the-art Superpixel Methods, TPAMI, May 2012.
```
grlee77, line 16:
```suggestion
    State-of-the-art Superpixel Methods, TPAMI, May 2012.
    :DOI:`10.1109/TPAMI.2012.120`
```

skimage/segmentation/slic_superpixels.py
```diff
@@ -88,12 +174,23 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
...
     .. [2] http://ivrg.epfl.ch/research/superpixels#SLICO
+    .. [3] https://arxiv.org/pdf/1606.09518.pdf
```
grlee77, line 192:
```suggestion
    .. [3] Irving, Benjamin. "maskSLIC: regional superpixel generation with
           application to local pathology characterisation in medical images.",
           2016, :arXiv:`1606.09518`
```

skimage/segmentation/slic_superpixels.py
```diff
@@ -88,12 +174,23 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
...
+      indexing starting at 0 will be deprecated in future versions. If
+      `mask` is not `None` labels indexing starts at 1 and masked area
+      is set to 0.
+
     References
     ----------
     .. [1] Radhakrishna Achanta, Appu Shaji, Kevin Smith, Aurelien Lucchi,
         Pascal Fua, and Sabine Süsstrunk, SLIC Superpixels Compared to
         State-of-the-art Superpixel Methods, TPAMI, May 2012.
     .. [2] http://ivrg.epfl.ch/research/superpixels#SLICO
```
grlee77, line 191:
```suggestion
    .. [2] https://www.epfl.ch/labs/ivrl/research/slic-superpixels/#SLICO
```
This link was no longer pointing to the expected info. I have updated it

emmanuelle:
@rfezzani can you confirm that you addressed all comments raised by @grlee77 ? Given the time zone difference and the fact that I want to release ASAP I'm going to merge this one but I wanted to be sure.

grlee77:
@emanuelle: I am okay with the remaining two comments being addressed in a followup PR after release if you want to get this in now.

One unresolved comment was for a minor efficiency improvement (no change in behavior). The other is an outdated docstring statement regarding the method of seed point initialization used.

Let me know and I can change my review to "Approved" if needed.

emmanuelle:
thanks @grlee77 I was not sure of your time zone or schedule :-). Yes it'd be helpful if you're willing to approve.

rfezzani:
Thank you @grlee77 for your review, I will open a PR ASAP to address your two last comments ;)
And thank you all for your valuable suggestions. :champagne:

PR #3854, opened 2019-04-21
Build and test the documentation in Azure Pipelines

soupault:
## Description

Implement documentation building and testing in Azure Pipelines, similarly to https://github.com/scikit-image/scikit-image/blob/master/tools/travis/script.sh.  

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_haar_extraction_selection_classification.py
+++ b/doc/examples/applications/plot_haar_extraction_selection_classification.py
@@ -41,136 +41,139 @@ from skimage.feature import haar_like_feature
 from skimage.feature import haar_like_feature_coord
 from skimage.feature import draw_haar_like_feature
 
-###############################################################################
-# The usual feature extraction scheme
-###############################################################################
-# The procedure to extract the Haar-like feature for an image is quite easy: a
-# region of interest (ROI) is defined for which all possible feature will be
-# extracted. The integral image of this ROI will be computed and all possible
-# features will be computed.
 
+if __name__ == '__main__':
+    ###########################################################################
+    # The usual feature extraction scheme
+    ###########################################################################
+    # The procedure to extract the Haar-like features for an image is quite
+    # easy. Firstly, a region of interest (ROI) is defined for which all
+    # possible features will be extracted. Secondly, the integral image of
+    # this ROI will be computed and used to compute the features.
 
-@delayed
-def extract_feature_image(img, feature_type, feature_coord=None):
-    """Extract the haar feature for the current image"""
-    ii = integral_image(img)
-    return haar_like_feature(ii, 0, 0, ii.shape[0], ii.shape[1],
-                             feature_type=feature_type,
-                             feature_coord=feature_coord)
+    @delayed
+    def extract_feature_image(img, feature_type, feature_coord=None):
+        """Extract the haar feature for the current image"""
+        ii = integral_image(img)
+        return haar_like_feature(ii, 0, 0, ii.shape[0], ii.shape[1],
+                                 feature_type=feature_type,
+                                 feature_coord=feature_coord)
 
+    ###########################################################################
+    # We will use a subset of the CBCL which is composed of 100 face images and
+    # 100 non-face images. Each image has been resized to a ROI of 19 by 19
+    # pixels. We will keep 75 images from each group to train a classifier and
+    # check which extracted features are the most salient, and use the
+    # remaining 25 from each class to check the performance of the classifier.
 
-###############################################################################
-# We will use a subset of the CBCL which is composed of 100 face images and 100
-# non-face images. Each image has been resized to a ROI of 19 by 19 pixels. We
-# will keep 75 images from each group to train a classifier and check which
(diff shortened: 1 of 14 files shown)


soupault:
`applications/plot_haar_extraction_selection_classification.py` example requires using "if-name-main" construction on Windows, otherwise the testing falls into the infinite loop. See Python 3.7 build here - https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=322, or SO -
https://stackoverflow.com/questions/20222534/python-multiprocessing-on-windows-if-name-main.

soupault:
Alright! The documentation testing seems to work! I had to implement several workarounds:
1) See above https://github.com/scikit-image/scikit-image/pull/3854#issuecomment-485239125. Unfortunately, the fix breaks the rendering of the example, so I'm open to any suggestions here.
2) `doc/source/random_gallery.py` used platform-dependent path delimiter.
3) Many tests in the example section of the docstrings failed due to unmatching `np.printoptions`: for some specific dtypes, `dtype` field is presented in the output, for others - not. Maybe `conftest.py` is skipped in Azure for some reason, I'm not yet sure.

Otherwise, the PR is ready for review.

skimage/feature/brief.py
```diff
@@ -93,17 +93,17 @@ class BRIEF(DescriptorExtractor):
     >>> extractor.extract(square2, keypoints2)
     >>> descriptors2 = extractor.descriptors
     >>> matches = match_descriptors(descriptors1, descriptors2)
-    >>> matches
+    >>> matches  # doctest: +SKIP
```
Borda: why these tests are skipped?

doc/examples/applications/plot_haar_extraction_selection_classification.py
```diff
@@ -41,136 +41,139 @@
...
-time_full_feature_comp = time() - t_start
-y = np.array([1] * 100 + [0] * 100)
-X_train, X_test, y_train, y_test = train_test_split(X, y, train_size=150,
-                                                    random_state=0,
-                                                    stratify=y)
-
-# Extract all possible features to be able to select the most salient.
-feature_coord, feature_type = \
+
+if __name__ == '__main__':
```
hmaarrfk: i'm a big -1 on this. This really confuses new users, who become confused since they don't knkow what `__main__` means. why do we need to do this?

skimage/feature/brief.py
```diff
@@ -64,18 +64,18 @@ class BRIEF(DescriptorExtractor):
     >>> import numpy as np
     >>> square1 = np.zeros((8, 8), dtype=np.int32)
     >>> square1[2:6, 2:6] = 1
-    >>> square1
+    >>> square1  # doctest: +SKIP
```
hmaarrfk: I think you should get linux setup, and run tehse doctests on linux. Linux will be able to handle the doctests and all.

doc/examples/applications/plot_haar_extraction_selection_classification.py
soupault: @hmaarrfk I'm still trying to figure out a better solution. The issue is described here - https://github.com/scikit-image/scikit-image/pull/3854#issuecomment-485239125, specifically, in the StackOverflow post.
In short, on Windows multiprocessing imports the full contents of the file when spawning extra processes. This leads to starting an infinite number of processes.
So, basically, the multiprocessing code has to be wrapped somehow.

skimage/feature/brief.py
soupault: @Borda I think `conftest.py` file is not being discovered by pytest on Windows. In that file we force numpy to use legacy printoptions. Therefore, the builds expect the newer notation (that is numerically equivalent, but visually different).

soupault: @hmaarrfk on Azure you mean? I'm not sure if anyone from the core team is willing to switch from Travis to Azure at the moment. Of course, it shouldn't be an issue to have a Linux instance running, but I really want to have Windows build with the full coverage.
Unfortunately, googling "pytest, windows, conftest" doesn't yield any info. I'm going to check if forcing conftest could be done with the env.variables...

doc/examples/applications/plot_haar_extraction_selection_classification.py
hmaarrfk: sigh, ok.

skimage/feature/brief.py
hmaarrfk: We could also just test the doctests on numpy 1.14 and higher

or just bunp the minimum reqs all the way up to 1.14 ;)

hmaarrfk: @soupault, it might be worthwhile too post an issue on pytest's github page. I couldn't figure this out either, it could very well be a bug on windows.

I understand that you want full test coverage on windows, but I would argue that having the doctests be readable is more important. I've often seen my comments copied verbatim in other's code. I wouldn't watn to pollute their dropbox with # doctest: +SKIP   ;)

soupault:
The work is continued in https://github.com/scikit-image/scikit-image/pull/3873.

PR #3862, opened 2019-04-25
Fix URL for stain deconvolution reference

stefanv:


soupault:
Doesn't look like a well-established website. Is there an alternative?
image

stefanv:
This is the redirect URL provided by the previous location. I don't think 
the content is hosted elsewhere, but could be wrong.

PR #3865, opened 2019-04-25
Documentation for color images in histogram matching

Borda:
## Description

cleaning unused variables and adding a note about using different colour space than RGB

## Checklist

## For reviewers

  later.

### DIFF
--- a/skimage/transform/histogram_matching.py
+++ b/skimage/transform/histogram_matching.py
@@ -23,7 +23,8 @@ def match_histograms(image, reference, multichannel=False):
     """Adjust an image so that its cumulative histogram matches that of another.
 
     The adjustment is applied separately for each channel.
-    Using RGB image, assume converting input image to some "continues" color space.
+    Using RGB image, assume converting input image to some "continuous" color 
+    space [2]_.
 
     Parameters
     ----------
--- a/skimage/transform/histogram_matching.py
+++ b/skimage/transform/histogram_matching.py
@@ -6,9 +6,9 @@ def _match_cumulative_cdf(source, template):
     Return modified source array so that the cumulative density function of
     its values matches the cumulative density function of the template.
     """
-    src_values, src_unique_indices, src_counts = np.unique(source.ravel(),
-                                                           return_inverse=True,
-                                                           return_counts=True)
+    _, src_unique_indices, src_counts = np.unique(source.ravel(),
+                                                  return_inverse=True,
+                                                  return_counts=True)
     tmpl_values, tmpl_counts = np.unique(template.ravel(), return_counts=True)
 
     # calculate normalized quantiles for each array
@@ -23,8 +23,12 @@ def match_histograms(image, reference, multichannel=False):
     """Adjust an image so that its cumulative histogram matches that of another.
 
     The adjustment is applied separately for each channel.
-    Using RGB image, assume converting input image to some "continuous" color 
-    space [2]_.
+
+    For color image assume to use another colour space then RGB since the colour
+    transitions are not very smooth. For example having red source image
+    and green template image will not produce expected result. Therefor convert
+    your images to "continuous" color space like HSV first and after histogram
+    matching back to RGB. [2]
 
     Parameters
     ----------
@@ -63,8 +67,8 @@ def match_histograms(image, reference, multichannel=False):
 
         matched = np.empty(image.shape, dtype=image.dtype)
         for channel in range(image.shape[-1]):
-            matched_channel = _match_cumulative_cdf(image[..., channel], reference[..., channel])
-            matched[..., channel] = matched_channel
+            matched[..., channel] = _match_cumulative_cdf(image[..., channel],
+                                                          reference[..., channel])
     else:
         matched = _match_cumulative_cdf(image, reference)
 
--- a/skimage/transform/histogram_matching.py
+++ b/skimage/transform/histogram_matching.py
@@ -49,7 +49,8 @@ def match_histograms(image, reference, multichannel=False):
 
     References
     ----------
-    .. [1] http://paulbourke.net/miscellaneous/equalisation/
+     .. [1] http://paulbourke.net/miscellaneous/equalisation/
+     .. [2] https://www.researchgate.net/post/Histogram_matching_for_color_images
 
     """
     if image.ndim != reference.ndim:
--- a/skimage/transform/histogram_matching.py
+++ b/skimage/transform/histogram_matching.py
@@ -24,7 +24,6 @@ def match_histograms(image, reference, multichannel=False):
 
     The adjustment is applied separately for each channel.
     Using RGB image, assume converting input image to some "continues" color space.
-    See: https://www.researchgate.net/post/Histogram_matching_for_color_images
 
     Parameters
     ----------
--- a/skimage/transform/histogram_matching.py
+++ b/skimage/transform/histogram_matching.py
@@ -19,11 +19,12 @@ def _match_cumulative_cdf(source, template):
     return interp_a_values[src_unique_indices].reshape(source.shape)
 
 
-
 def match_histograms(image, reference, multichannel=False):
     """Adjust an image so that its cumulative histogram matches that of another.
 
     The adjustment is applied separately for each channel.
+    Using RGB image, assume converting input image to some "continues" color space.
+    See: https://www.researchgate.net/post/Histogram_matching_for_color_images
 
     Parameters
     ----------
@@ -51,9 +52,6 @@ def match_histograms(image, reference, multichannel=False):
     .. [1] http://paulbourke.net/miscellaneous/equalisation/
 
     """
-    shape = image.shape
-    image_dtype = image.dtype
-
     if image.ndim != reference.ndim:
         raise ValueError('Image and reference must have the same number of channels.')
 
--- a/skimage/transform/histogram_matching.py
+++ b/skimage/transform/histogram_matching.py
@@ -26,9 +26,9 @@ def match_histograms(image, reference, multichannel=False):
 
     For color image assume to use another colour space then RGB since the colour
     transitions are not very smooth. For example having red source image
-    and green template image will not produce expected result. Therefor convert
-    your images to "continuous" color space like HSV first and after histogram
-    matching back to RGB. [2]
+    and green template image may not produce expected result. Therefore you may
+    get better results by first converting your images to "continuous" color
+    space like HSV and after histogram matching converting back to RGB. [2]
 
     Parameters
     ----------


skimage/transform/histogram_matching.py
```diff
@@ -19,11 +19,12 @@ def _match_cumulative_cdf(source, template):
     return interp_a_values[src_unique_indices].reshape(source.shape)
 
 
-
 def match_histograms(image, reference, multichannel=False):
     """Adjust an image so that its cumulative histogram matches that of another.
 
     The adjustment is applied separately for each channel.
+    Using RGB image, assume converting input image to some "continues" color space.
```
hmaarrfk:
```suggestion
    Using RGB image, assume converting input image to some "continuous" color 
    space [2]_.
```

skimage/transform/histogram_matching.py
```diff
@@ -19,11 +19,12 @@ def _match_cumulative_cdf(source, template):
...
+    See: https://www.researchgate.net/post/Histogram_matching_for_color_images
```
hmaarrfk:
```suggestion

```

skimage/transform/histogram_matching.py
```diff
@@ -51,9 +52,6 @@ def match_histograms(image, reference, multichannel=False):
     .. [1] http://paulbourke.net/miscellaneous/equalisation/
```
hmaarrfk:
```suggestion
    .. [1] http://paulbourke.net/miscellaneous/equalisation/
    .. [2] https://www.researchgate.net/post/Histogram_matching_for_color_images
```

hmaarrfk:
sigh, sphinx broke things for you.....

Borda:
@hmaarrfk I know, but it came from the actual master... :-/

hmaarrfk:
I don't really understand what you are trying to say:

> Using RGB image, assume converting input image to some "continuous" color space [2]_.

Can you explain in more words

Borda:
@hmaarrfk can you look at the updated version?

skimage/transform/histogram_matching.py
```diff
@@ -6,9 +6,9 @@ def _match_cumulative_cdf(source, template):
     Return modified source array so that the cumulative density function of
     its values matches the cumulative density function of the template.
     """
-    src_values, src_unique_indices, src_counts = np.unique(source.ravel(),
-                                                           return_inverse=True,
-                                                           return_counts=True)
+    _, src_unique_indices, src_counts = np.unique(source.ravel(),
```
hmaarrfk: I'm -1 on these 3 changed lines. These kinda feel like "flavour of the week" kinda of changes, which are only side steps, and not clear "improvements".

skimage/transform/histogram_matching.py
```diff
@@ -48,12 +53,10 @@ def match_histograms(image, reference, multichannel=False):
 
     References
     ----------
-    .. [1] http://paulbourke.net/miscellaneous/equalisation/
+     .. [1] http://paulbourke.net/miscellaneous/equalisation/
```
hmaarrfk:
```suggestion
    .. [1] http://paulbourke.net/miscellaneous/equalisation/
```
I'm sorry, I don't know why my previous suggestion added that space :S

skimage/transform/histogram_matching.py
```diff
@@ -48,12 +53,10 @@ def match_histograms(image, reference, multichannel=False):
...
+     .. [2] https://www.researchgate.net/post/Histogram_matching_for_color_images
```
hmaarrfk:
```suggestion
    .. [2] https://www.researchgate.net/post/Histogram_matching_for_color_images
```

skimage/transform/histogram_matching.py
```diff
@@ -19,12 +19,17 @@ def _match_cumulative_cdf(source, template):
...
 def match_histograms(image, reference, multichannel=False):
     """Adjust an image so that its cumulative histogram matches that of another.
 
     The adjustment is applied separately for each channel.
 
+    For color image assume to use another colour space then RGB since the colour
+    transitions are not very smooth. For example having red source image
+    and green template image will not produce expected result. Therefor convert
+    your images to "continuous" color space like HSV first and after histogram
+    matching back to RGB. [2]
```
hmaarrfk: This could use some rewording:

> Note that for images in RGB space, the colour transitions are not very smooth. For example, having a red source image and a green template image may not produce the expected results as the per-channel histograms are unrelated. You may have better results by first converting your images to a continuous color space such as HSV prior to calling the function match_instograms [2]_.

Basically, I wouldn't make any hard statements about the performance of the algorithm for a specific use case.

skimage/transform/histogram_matching.py
Borda: sorry, I do not understand "flavour of the week"

skimage/transform/histogram_matching.py
Borda: that why I did not d any code change just add a warning that the standards RGB is not the best colour space for histogram matching per channel

skimage/transform/histogram_matching.py
Borda: I accepted your earlier suggestion and did not do any change in this line, so it seems you are correcting your earlier suggestion... :]

skimage/transform/histogram_matching.py
Borda: I accepted your earlier suggestion and did not do any change in this line, so it seems you are correcting your earlier suggestion... :]

skimage/transform/histogram_matching.py
Borda: Btw, what reworking do you have in mind? I am also fine to remove it completely...

skimage/transform/histogram_matching.py
hmaarrfk: flavour of the week is a statement I created that means that the motivations for this modification change every week depending. In a similar fashion that a desert store will have a new flavour every week.

These kinds of changes make the git history very hard to read, and don't really add too much value. I could agree that `src_values` reminds us what the `_` refers to, in this not so known function.

skimage/transform/histogram_matching.py
hmaarrfk: `assume to use another colour space` isn't very clear. I think you mean `RGB is not the best color space`.

`will not` -> `may not`

`Therefor` -> `Therefor`

`convert` -> `you may get better results by first` <----- key addition here is the word "may".

skimage/transform/histogram_matching.py
```diff
@@ -49,11 +54,9 @@ def match_histograms(image, reference, multichannel=False):
     References
     ----------
     .. [1] http://paulbourke.net/miscellaneous/equalisation/
+    .. [2] https://www.researchgate.net/post/Histogram_matching_for_color_images
```
sciunto: I would prefer to see a paper rather than a researchgate thread.

skimage/transform/histogram_matching.py
Borda: by my understanding, this change is recommended from codacy.com and codefactor.io and it is the same kind of change as yours moving `not` or removing `results`, https://github.com/scikit-image/scikit-image/pull/3224

skimage/transform/histogram_matching.py
Borda: I have not found any paper about it yet, but see the following example
hist-matching

Borda:
@sciunto just to clarify, `skimage` does not support python 2.x any more, right?

jni:
@Borda that is correct, you don't need to worry about Py2.7 support and can use syntax for Py3.5+, including the matmul symbol, `@`.

skimage/transform/histogram_matching.py
Borda: `Therefor` -> `Therefor` ?

skimage/transform/histogram_matching.py
Borda: I found a similar topic here - https://github.com/mapbox/rio-hist/issues/3

skimage/transform/histogram_matching.py
hmaarrfk: gotta add that e ;)

hmaarrfk:
the wording still isn't quite right:

> assume to use another colour space

ins't entirely grammatically correct.

You should say something more like:

> If multichannel is set to True, images are analyzed on a per channel basis.

This is a fact.

> Results for images stored in RGB color space may not behave as expected since transitions are not always smooth in each channel.

General statement that encourages more thought.

Finally, if you want to make this really awesome, I would encourage you to make an example with the images you used in the comment.

Borda:
@hmaarrfk example sounds reasonable, but would I include the image in this repo or download it online when needed? is there any guidence on how to create an example with new/own images?

hmaarrfk:
I think the first PR would be to add the image(s) to the dataset of images shipped with scikit-image.

That PR would include:
1. A license for the image compatible with scikit-image. see discussion about this here: https://github.com/scikit-image/scikit-image/issues/3384
2. Adding them to the data directory
https://github.com/scikit-image/scikit-image/blob/master/skimage/data/__init__.py
3. Ensure it shows up on the gallery (I forget how to do this at the moment).

It seems like you would include two images, the reference, and the source.
Generally here the challenge is to ensure we have the right license for the images.

Not sure how you got that data, but it seems that even lower res images might work to make your point, but high res images would be fine too. I'll push to include your full images.

After you get the data in scikit-image, we can have a more focused discussion on how to improve the example and whatnot.

skimage/transform/histogram_matching.py
sciunto: If you had the example, you can remove this link. The example will speak by itself.

Borda:
I need to check the condition of sharing first and try to remember who is the source...

skimage/transform/histogram_matching.py
```diff
@@ -19,12 +19,17 @@ def _match_cumulative_cdf(source, template):
     return interp_a_values[src_unique_indices].reshape(source.shape)
 
 
-
 def match_histograms(image, reference, multichannel=False):
     """Adjust an image so that its cumulative histogram matches that of another.
 
     The adjustment is applied separately for each channel.
 
+    For color image assume to use another colour space then RGB since the colour
```
stefanv: Hey, @Borda, thank you for helping us clarify the documentation; that's always valuable.

I am struggling to understand the new text.  Why do you say HSV is "continuous", and RGB is not?  I can help wordsmith this a bit, but I don't have a good understanding of the issue at hand.

skimage/transform/histogram_matching.py
stefanv: Woops, I think all the futzing with this led to incorrect indentation.

Borda:
it seems that with setting `multichannel=True` the change is not so drastically... 
https://github.com/Borda/BIRL/raw/devel/figures/Rat-Kidney_histogram-matching.jpg

hmaarrfk:
Can you add the full code?

Borda:
```python
import os
from skimage.io import imread
from skimage.transform import match_histograms
from skimage.color import rgb2hsv, hsv2rgb, rgb2lab, lab2rgb, lab2lch, lch2lab

rgb2lch = lambda im: rgb2lab(lab2lch(im))
lch2rgb = lambda im: lab2rgb(lch2lab(im))

path_imgs = 'https://raw.githubusercontent.com/Borda/BIRL/master/data_images/rat-kidney_/scale-5pc/'
path_img_ref = path_imgs + 'Rat-Kidney_HE.jpg'
path_img_move = path_imgs + 'Rat-Kidney_PanCytokeratin.jpg'

os.system('wget %s' % path_img_ref)
os.system('wget %s' % path_img_move)

img_ref = imread(os.path.basename(path_img_ref))
img_move = imread(os.path.basename(path_img_move))
use_multichannel = True

fig, axarr = plt.subplots(ncols=3, nrows=2, figsize=(12, 6))
axarr[0, 1].set_title('Reference')
axarr[0, 1].imshow(img_ref)
axarr[0, 0].set_title('Source')
axarr[0, 0].imshow(img_move)

axarr[0, 2].set_title('Hist. matching using RGB')
axarr[0, 2].imshow(match_histograms(img_move, img_ref, multichannel=use_multichannel))
axarr[1, 0].set_title('Hist. matching using HSV')
axarr[1, 0].imshow(hsv2rgb(match_histograms(rgb2hsv(img_move), rgb2hsv(img_ref), multichannel=use_multichannel)))
axarr[1, 1].set_title('Hist. matching using LAB')
axarr[1, 1].imshow(lab2rgb(match_histograms(rgb2lab(img_move), rgb2lab(img_ref), multichannel=use_multichannel)))
axarr[1, 2].set_title('Hist. matching using LCH')
axarr[1, 2].imshow(lch2rgb(match_histograms(rgb2lch(img_move), rgb2lch(img_ref), multichannel=use_multichannel)))
```

Borda:
@hmaarrfk do you think that it should/could be included or drop this PR?

hmaarrfk:
maybe you should rephrase your conclusion, and tell users to use multichannel=True for RGB images?

Borda:
Sure and in the doc of the function `match_histograms` or still add an example?
Rat-Kidney_histogram-matching

Borda:
@hmaarrfk ^^

hmaarrfk:
If you can add that to the docs, then just adding it there is fine. Otherwise, this should be a separate example. The documentation will find the example automatically I think

hmaarrfk:
I think showing it in an email example gallery will go a long way to helping others. Especially with syntax

Issue #3866, opened 2019-04-26
fast_exp missing when compiled in Alpine OS


hongquan:
## Description

When I use pip to install scikit-image inside Alpine OS 3.9 (which uses musl for standard lib C), the produced scikit-image is missing `fast_exp`

The normal import code:

```py
import skimage
```
leads to this error:

```
appareil/tasks.py:5: in <module>
    import skimage
/root/.cache/pypoetry/virtualenvs/manwa-fys_VRJo-py3.6/lib/python3.6/site-packages/skimage/__init__.py:135: in <module>
    from .data import data_dir
/root/.cache/pypoetry/virtualenvs/manwa-fys_VRJo-py3.6/lib/python3.6/site-packages/skimage/data/__init__.py:16: in <module>
    from ._binary_blobs import binary_blobs
/root/.cache/pypoetry/virtualenvs/manwa-fys_VRJo-py3.6/lib/python3.6/site-packages/skimage/data/_binary_blobs.py:2: in <module>
    from ..filters import gaussian
/root/.cache/pypoetry/virtualenvs/manwa-fys_VRJo-py3.6/lib/python3.6/site-packages/skimage/filters/__init__.py:3: in <module>
    from .edges import (sobel, sobel_h, sobel_v,
/root/.cache/pypoetry/virtualenvs/manwa-fys_VRJo-py3.6/lib/python3.6/site-packages/skimage/filters/edges.py:17: in <module>
    from ..restoration.uft import laplacian
/root/.cache/pypoetry/virtualenvs/manwa-fys_VRJo-py3.6/lib/python3.6/site-packages/skimage/restoration/__init__.py:10: in <module>
    from .non_local_means import denoise_nl_means
/root/.cache/pypoetry/virtualenvs/manwa-fys_VRJo-py3.6/lib/python3.6/site-packages/skimage/restoration/non_local_means.py:2: in <module>
    from ._nl_means_denoising import (
E   ImportError: Error relocating /root/.cache/pypoetry/virtualenvs/manwa-fys_VRJo-py3.6/lib/python3.6/site-packages/skimage/restoration/_nl_means_denoising.cpython-36m-x86_64-linux-gnu.so: fast_exp: symbol not found
```

## Way to reproduce

I prepared a Docker images, where I installed build tools, libraries. I also built some Python packages to whl files, to help install faster. 

```
docker run -it quay.io/quan/docker-ansible-libav-django /bin/bash
```

Inside the Docker container, run

```
[...]

PR #3868, opened 2019-04-27
Fix fast_exp for gcc 8.3 with certain flags

hmaarrfk:
## Description

Not sure exactly what combination of things this helps
Closes https://github.com/scikit-image/scikit-image/issues/3866

But I think these flags are all necessary for some compilers with certain flags....

Who knows.... But it definitely helps one person
## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/fast_exp.h
+++ b/skimage/_shared/fast_exp.h
@@ -15,7 +15,7 @@
 // All these inline directives are necessary for us to stay compatible with
 // as many compilers as possible.
 // See discussion https://github.com/scikit-image/scikit-image/issues/3866
-__inline double inline __attribute__((always_inline)) fast_exp (double y)
+double inline __attribute__((always_inline)) fast_exp (double y)
 {
     union
     {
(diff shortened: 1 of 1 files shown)


jni:
@hmaarrfk before this PR is merged (thank you for attempting this!), I think a summary of the discussion with links to relevant documentation needs to go in that comment. The code should be much more self-explanatory than it is currently, and the discussion in #3866 has a lot of irrelevant info. Ideally, the comments should link to relevant gcc, MSVC, and Cython docs (as appropriate).

hmaarrfk:
Yeah, i"m not really happy with this PR. It seems to fragile.

I think the appropriate solution will be for somebody to rewrite this in Cython maybe?

Also, I'm not sure if C actually supports inline in headers.

https://stackoverflow.com/questions/12151168/does-ansi-c-not-know-the-inline-keyword

To my knowledge, this kind of stuff isn't really part of the standard.


Alternatively, since nl_mean_denoising is the only one that uses fast_exp, it could be moved inside that cython module/???

This definitely isn't the right solution PR.

PR #3872, opened 2019-04-29
Adding `threshold_multiotsu`.

alexandrejaguar:
## Description

Adds `threshold_multiotsu`. A followup to #3385, #2311.
Thanks @emmanuelle for the cython functions!

## Checklist

- [X] Gallery example in `./doc/examples` (new features only)
- [X] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
- [X] Unit tests

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_filters.py
+++ b/benchmarks/benchmark_filters.py
@@ -1,7 +1,7 @@
 # See "Writing benchmarks" in the asv docs for more information.
 # https://asv.readthedocs.io/en/latest/writing_benchmarks.html
 import numpy as np
-from skimage import filters
+from skimage import data, filters
 
 
 class FiltersSuite:
@@ -12,6 +12,35 @@ class FiltersSuite:
         self.image[3000:, 3000] += 0.5
 
     def time_sobel(self):
-        result = filters.sobel(self.image)
+        filters.sobel(self.image)
 
 
+class MultiOtsu(object):
+    """Benchmarks for MultiOtsu threshold."""
+    param_names = ['classes']
+    params = [(3, 5)]
+
+    def setup(self, *args):
+        self.image = data.camera()
+
+    def time_threshold_multiotsu(self, classes):
+        filters.threshold_multiotsu(self.image, classes=classes)
+
+    def peakmem_reference(self, *args):
+        """Provide reference for memory measurement with empty benchmark.
+
+        Peakmem benchmarks measure the maximum amount of RAM used by a
+        function. However, this maximum also includes the memory used
+        during the setup routine (as of asv 0.2.1; see [1]_).
+        Measuring an empty peakmem function might allow us to disambiguate
+        between the memory used by setup and the memory used by target (see
+        other ``peakmem_`` functions below).
+
+        References
+        ----------
+        .. [1]: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory
+        """
+        pass
+
+    def peakmem_threshold_multiotsu(self, classes):
+        filters.threshold_multiotsu(self.image, classes=classes)
--- a/benchmarks/benchmark_filters.py
+++ b/benchmarks/benchmark_filters.py
@@ -18,7 +18,7 @@ class FiltersSuite:
 class MultiOtsu(object):
     """Benchmarks for MultiOtsu threshold."""
     param_names = ['classes']
-    params = [(3, 5)]
+    params = [3, 5]
 
     def setup(self, *args):
         self.image = data.camera()
--- a/benchmarks/benchmark_filters.py
+++ b/benchmarks/benchmark_filters.py
@@ -18,7 +18,7 @@ class FiltersSuite:
 class MultiOtsu(object):
     """Benchmarks for MultiOtsu threshold."""
     param_names = ['classes']
-    params = [3, 5]
+    params = [3, 4, 5]
 
     def setup(self, *args):
         self.image = data.camera()
--- a/doc/examples/segmentation/plot_multiotsu.py
+++ b/doc/examples/segmentation/plot_multiotsu.py
@@ -35,7 +35,7 @@ image = data.camera()
 # three classes.
 thresholds = threshold_multiotsu(image)
 
-# Using the values on thresh, we generate the three regions.
+# Using the threshold values, we generate the three regions.
 regions = np.digitize(image, bins=thresholds)
 
 fig, ax = plt.subplots(nrows=1, ncols=3, figsize=(10, 3.5))
--- a/doc/examples/segmentation/plot_multiotsu.py
+++ b/doc/examples/segmentation/plot_multiotsu.py
@@ -3,9 +3,9 @@
 Multi-Otsu Thresholding
 =======================
 
-Multi-Otsu threshold [1]_ is used to separate the pixels of an input
-image into several different classes, each one obtained according to the
-intensity of the gray levels within the image.
+The multi-Otsu threshold [1]_ is a thresholding algorithm thatused to separate
+the pixels of an input image into several different classes, each one obtained
+according to the intensity of the gray levels within the image.
 
 Multi-Otsu calculates several thresholds, determined by the number of desired
 classes. The default number of classes is 3: for obtaining three classes, the
@@ -36,7 +36,7 @@ image = data.camera()
 thresholds = threshold_multiotsu(image)
 
 # Using the values on thresh, we generate the three regions.
-regions = np.digitize(image, bins=thresh)
+regions = np.digitize(image, bins=thresholds)
 
 fig, ax = plt.subplots(nrows=1, ncols=3, figsize=(10, 3.5))
 
--- a/doc/examples/segmentation/plot_multiotsu.py
+++ b/doc/examples/segmentation/plot_multiotsu.py
@@ -3,7 +3,7 @@
 Multi-Otsu Thresholding
 =======================
 
-The multi-Otsu threshold [1]_ is a thresholding algorithm thatused to separate
+The multi-Otsu threshold [1]_ is a thresholding algorithm that is used to separate
 the pixels of an input image into several different classes, each one obtained
 according to the intensity of the gray levels within the image.
 
@@ -49,8 +49,8 @@ ax[0].axis('off')
 # multi-Otsu.
 ax[1].hist(image.ravel(), bins=255)
 ax[1].set_title('Histogram')
-for i in range(len(thresholds)):
-    ax[1].axvline(thresholds[i], color='r')
+for thresh in thresholds:
+    ax[1].axvline(thresh, color='r')
 
 # Plotting the Multi Otsu result.
 ax[2].imshow(regions, cmap='Accent')
--- a/doc/examples/segmentation/plot_multiotsu.py
+++ b/doc/examples/segmentation/plot_multiotsu.py
@@ -4,7 +4,7 @@ Multi-Otsu Thresholding
 =======================
 
 Multi-Otsu threshold [1]_ is used to separate the pixels of an input
-image into three different classes, each one obtained according to the
+image into several different classes, each one obtained according to the
 intensity of the gray levels within the image.
 
 Multi-Otsu calculates several thresholds, determined by the number of desired
@@ -25,7 +25,7 @@ import numpy as np
 from skimage import data
 from skimage.filters import threshold_multiotsu
 
-# Stablishing the font size for all plots.
+# Setting the font size for all plots.
 matplotlib.rcParams['font.size'] = 9
 
 # The input image.
@@ -33,7 +33,7 @@ image = data.camera()
 
 # Applying multi-Otsu threshold for the default value, generating
 # three classes.
-thresh = threshold_multiotsu(image)
+thresholds = threshold_multiotsu(image)
 
 # Using the values on thresh, we generate the three regions.
 regions = np.digitize(image, bins=thresh)
@@ -49,8 +49,8 @@ ax[0].axis('off')
 # multi-Otsu.
 ax[1].hist(image.ravel(), bins=255)
 ax[1].set_title('Histogram')
-for i in range(len(thresh)):
-    ax[1].axvline(thresh[i], color='r')
+for i in range(len(thresholds)):
+    ax[1].axvline(thresholds[i], color='r')
 
 # Plotting the Multi Otsu result.
 ax[2].imshow(regions, cmap='Accent')
--- a/doc/examples/segmentation/plot_multiotsu.py
+++ b/doc/examples/segmentation/plot_multiotsu.py
@@ -1,0 +1,62 @@
+"""
+=======================
+Multi-Otsu Thresholding
+=======================
+
+Multi-Otsu threshold [1]_ is used to separate the pixels of an input
+image into three different classes, each one obtained according to the
+intensity of the gray levels within the image.
+
+Multi-Otsu calculates several thresholds, determined by the number of desired
+classes. The default number of classes is 3: for obtaining three classes, the
+algorithm returns two threshold values. They are represented by a red line in
+the histogram below.
+
+.. [1] Liao, P-S., Chen, T-S. and Chung, P-C., "A fast algorithm for multilevel
+       thresholding", Journal of Information Science and Engineering 17 (5):
+       713-727, 2001. Available at:
+       <http://ftp.iis.sinica.edu.tw/JISE/2001/200109_01.pdf>.
+"""
+
+import matplotlib
+import matplotlib.pyplot as plt
+import numpy as np
+
+from skimage import data
+from skimage.filters import threshold_multiotsu
+
+# Stablishing the font size for all plots.
+matplotlib.rcParams['font.size'] = 9
+
+# The input image.
+image = data.camera()
+
+# Applying multi-Otsu threshold for the default value, generating
+# three classes.
+thresh = threshold_multiotsu(image)
+
+# Using the values on thresh, we generate the three regions.
+regions = np.digitize(image, bins=thresh)
+
+fig, ax = plt.subplots(nrows=1, ncols=3, figsize=(10, 3.5))
+
+# Plotting the original image.
+ax[0].imshow(image, cmap='gray')
+ax[0].set_title('Original')
+ax[0].axis('off')
+
+# Plotting the histogram and the two thresholds obtained from
+# multi-Otsu.
+ax[1].hist(image.ravel(), bins=255)
+ax[1].set_title('Histogram')
+for i in range(len(thresh)):
+    ax[1].axvline(thresh[i], color='r')
+
+# Plotting the Multi Otsu result.
+ax[2].imshow(regions, cmap='Accent')
+ax[2].set_title('Multi-Otsu result')
+ax[2].axis('off')
+
+plt.subplots_adjust()
+
+plt.show()
--- a/skimage/filters/_multiotsu.pyx
+++ b/skimage/filters/_multiotsu.pyx
@@ -23,32 +23,15 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
         number of classes to be segmented
     bins : int
         number of bins used in the histogram
-    aux_thresh : array
-        thresholds to be returned
     """
     cdef Py_ssize_t idd
     cdef Py_ssize_t sh0, sh1
     cdef double part_sigma = 0
     # max_sigma is the maximum variance between classes.
     cdef double max_sigma = 0
-    #cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
-    #    combinations(range(1, bins - 2), classes - 1)))
 
-#    sh0 = _tmp.shape[0]
-#    sh1 = _tmp.shape[1]
-#    cdef cnp.intp_t[:, ::1] idx_tuples = np.zeros((sh0, sh1 + 2),
-#                                                   dtype=np.intp)
-#    idx_tuples[:, 1:sh1 + 1] = _tmp[:]
-#    idx_tuples[:, sh1 + 1] = bins - 1
-#    cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes-1, dtype=np.intp)
-
-    # np.intp is the same dtype as Py_ssize_t. I'm not sure how to
-    # create numpy arrays of Py_ssize_t otherwise.
-    # I think you could alternatively use
-    # cdef cnp.intp[::1] array = ...
     py_aux_thresh = np.empty(classes - 1, dtype=np.intp)
     cdef Py_ssize_t[::1] aux_thresh = py_aux_thresh
-
     cdef Py_ssize_t[::1] idx_tuple = np.zeros(classes+1, dtype=np.intp)
     idx_tuple[classes] = bins - 1
 
@@ -65,7 +48,9 @@ cdef double _find_best_rec(double[:, ::1] var_btwcls, cnp.intp_t min_val,
                            cnp.intp_t max_val, Py_ssize_t[::1] idx_tuple,
                            cnp.intp_t divisions, cnp.intp_t depth,
                            double max_sigma, Py_ssize_t[::1] aux_thresh) nogil:
-    """"""
+    """
+    Recursive function for finding max_sigma.
+    """
     cdef cnp.intp_t idx, idd
     cdef double part_sigma
 
--- a/skimage/filters/_multiotsu.pyx
+++ b/skimage/filters/_multiotsu.pyx
@@ -1,0 +1,53 @@
+#cython: cdivision=True
+#cython: boundscheck=False
+#cython: nonecheck=False
+#cython: wraparound=False
+import numpy as np
+
+cimport numpy as cnp
+cimport cython
+from itertools import combinations
+
+def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
+                              Py_ssize_t classes,
+                              Py_ssize_t bins,
+                              double [::1] aux_thresh):
+    """
+    Cython utility function for finding thresholds in multi-Otsu algorithm.
+    This function is only called by filters.threshold_multiotsu.
+
+    Parameters
+    ----------
+
+    var_btwcls : 2-d array
+        array of variance between classes
+    classes : int
+        number of classes to be segmented
+    bins : int
+        number of bins used in the histogram
+    aux_thresh : array
+        thresholds to be returned
+    """
+    cdef Py_ssize_t idd
+    cdef Py_ssize_t sh0, sh1
+    cdef double part_sigma = 0
+    cdef double max_sigma = 0
+    cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
+                                combinations(range(1, bins - 2), classes - 1)))
+    sh0 = _tmp.shape[0]
+    sh1 = _tmp.shape[1]
+    cdef Py_ssize_t [:, :] idx_tuples = np.zeros((sh0, sh1 + 2), 
+                                                  dtype=np.int)
+    idx_tuples[:, 1:sh1 + 1] = _tmp[:]
+    idx_tuples[:, sh1 + 1] = bins - 1
+    cdef Py_ssize_t [::1] idx_tuple
+    for idx_tuple in idx_tuples:
+        part_sigma = 0
+        for idd in range(0, classes):
+            part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
+        if max_sigma < part_sigma:
+            for idd in range(0, classes - 1):
+                aux_thresh[idd] = idx_tuple[idd + 1]
+                max_sigma = part_sigma
+
+    return aux_thresh
--- a/skimage/filters/_multiotsu.pyx
+++ b/skimage/filters/_multiotsu.pyx
@@ -18,11 +18,16 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
     Parameters
     ----------
     var_btwcls : 2-d array
-        array of variance between classes
+        Array of variance between classes.
     classes : int
-        number of classes to be segmented
+        Number of classes to be segmented.
     bins : int
-        number of bins used in the histogram
+        Number of bins used in the histogram.
+
+    Returns
+    -------
+    py_aux_thresh : array
+        Thresholds returned by the multi-Otsu algorithm.
     """
     cdef Py_ssize_t idd
     cdef Py_ssize_t sh0, sh1
@@ -49,7 +54,32 @@ cdef double _find_best_rec(double[:, ::1] var_btwcls, cnp.intp_t min_val,
                            cnp.intp_t divisions, cnp.intp_t depth,
                            double max_sigma, Py_ssize_t[::1] aux_thresh) nogil:
     """
-    Recursive function for finding max_sigma.
+    Recursive function for calculating max_sigma.
+
+    Parameters
+    ----------
+    var_btwcls : 2-d array
+        Array of variance between classes.
+    min_val : int
+        Minimum value of the checked intervals.
+    max_val : int
+        Maximum value of the checked intervals.
+    idx_tuple : array
+        number of bins used in the histogram
+    divisions : int
+        Number of divisions required to generate the desired classes.
+    depth : int
+        Controls the iterations the algorithm had, expanding the interval
+        when _find_best_rec() is called.
+    max_sigma : float
+        Maximum variance between classes.
+    aux_thresh : array
+        Values for multi-Otsu threshold.
+
+    Returns
+    -------
+    max_sigma : float
+        Maximum variance between classes.
     """
     cdef cnp.intp_t idx, idd
     cdef double part_sigma
@@ -65,7 +95,6 @@ cdef double _find_best_rec(double[:, ::1] var_btwcls, cnp.intp_t min_val,
                 if max_sigma < part_sigma:
                     aux_thresh[:idd] = idx_tuple[1:idd + 1]
                     max_sigma = part_sigma
-
     else:
         for idx in range(min_val, max_val-divisions+depth):
             idx_tuple[depth+1] = idx
--- a/skimage/filters/_multiotsu.pyx
+++ b/skimage/filters/_multiotsu.pyx
@@ -36,19 +36,20 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
                                 combinations(range(1, bins - 2), classes - 1)))
     sh0 = _tmp.shape[0]
     sh1 = _tmp.shape[1]
-    cdef Py_ssize_t [:, :] idx_tuples = np.zeros((sh0, sh1 + 2), 
-                                                  dtype=np.int)
+    cdef Py_ssize_t [:, ::1] idx_tuples = np.zeros((sh0, sh1 + 2),
+                                                   dtype=np.intp)
     idx_tuples[:, 1:sh1 + 1] = _tmp[:]
     idx_tuples[:, sh1 + 1] = bins - 1
-    cdef Py_ssize_t [::1] idx_tuple
+    cdef cnp.int64_t [::1] idx_tuple = np.zeros(classes-1, dtype=np.int64)
+
     for idx_tuple in idx_tuples:
         part_sigma = 0
-        for idd in range(0, classes):
+        for idd in range(classes):
             part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
         # checking if partial sigma is higher than maximum sigma
         if max_sigma < part_sigma:
-            for idd in range(0, classes - 1):
+            for idd in range(classes - 1):
                 aux_thresh[idd] = idx_tuple[idd + 1]
-                max_sigma = part_sigma
+            max_sigma = part_sigma
 
     return aux_thresh
--- a/skimage/filters/_multiotsu.pyx
+++ b/skimage/filters/_multiotsu.pyx
@@ -18,7 +18,6 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
 
     Parameters
     ----------
-
     var_btwcls : 2-d array
         array of variance between classes
     classes : int
@@ -31,6 +30,7 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
     cdef Py_ssize_t idd
     cdef Py_ssize_t sh0, sh1
     cdef double part_sigma = 0
+    # max_sigma is the maximum variance between classes.
     cdef double max_sigma = 0
     cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
                                 combinations(range(1, bins - 2), classes - 1)))
@@ -45,6 +45,7 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
         part_sigma = 0
         for idd in range(0, classes):
             part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
+        # checking if partial sigma is higher than maximum sigma
         if max_sigma < part_sigma:
             for idd in range(0, classes - 1):
                 aux_thresh[idd] = idx_tuple[idd + 1]
--- a/skimage/filters/_multiotsu.pyx
+++ b/skimage/filters/_multiotsu.pyx
@@ -1,7 +1,7 @@
 #cython: cdivision=True
-#cython: boundscheck=False
-#cython: nonecheck=False
-#cython: wraparound=False
+##cython: boundscheck=False
+##cython: nonecheck=False
+##cython: wraparound=False
 import numpy as np
 
 cimport numpy as cnp
@@ -10,8 +10,7 @@ from itertools import combinations
 
 def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
                               Py_ssize_t classes,
-                              Py_ssize_t bins,
-                              double [::1] aux_thresh):
+                              Py_ssize_t bins):
     """
     Cython utility function for finding thresholds in multi-Otsu algorithm.
     This function is only called by filters.threshold_multiotsu.
@@ -33,7 +32,7 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
     # max_sigma is the maximum variance between classes.
     cdef double max_sigma = 0
     cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
-                                combinations(range(1, bins - 2), classes - 1)))
+        combinations(range(1, bins - 2), classes - 1)))
     sh0 = _tmp.shape[0]
     sh1 = _tmp.shape[1]
     cdef Py_ssize_t [:, ::1] idx_tuples = np.zeros((sh0, sh1 + 2),
@@ -41,6 +40,9 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
     idx_tuples[:, 1:sh1 + 1] = _tmp[:]
     idx_tuples[:, sh1 + 1] = bins - 1
     cdef cnp.int64_t [::1] idx_tuple = np.zeros(classes-1, dtype=np.int64)
+
+    py_aux_thresh = np.empty(classes - 1)
+    cdef double [::1] aux_thresh = py_aux_thresh
 
     for idx_tuple in idx_tuples:
         part_sigma = 0
@@ -52,4 +54,4 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
                 aux_thresh[idd] = idx_tuple[idd + 1]
             max_sigma = part_sigma
 
-    return aux_thresh
+    return py_aux_thresh
--- a/skimage/filters/_multiotsu.pyx
+++ b/skimage/filters/_multiotsu.pyx
@@ -1,7 +1,7 @@
 #cython: cdivision=True
-##cython: boundscheck=False
-##cython: nonecheck=False
-##cython: wraparound=False
+#cython: boundscheck=False
+#cython: nonecheck=False
+#cython: wraparound=False
 import numpy as np
 
 cimport numpy as cnp
@@ -31,27 +31,49 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
     cdef double part_sigma = 0
     # max_sigma is the maximum variance between classes.
     cdef double max_sigma = 0
-    cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
-        combinations(range(1, bins - 2), classes - 1)))
-    sh0 = _tmp.shape[0]
-    sh1 = _tmp.shape[1]
-    cdef Py_ssize_t [:, ::1] idx_tuples = np.zeros((sh0, sh1 + 2),
-                                                   dtype=np.intp)
-    idx_tuples[:, 1:sh1 + 1] = _tmp[:]
-    idx_tuples[:, sh1 + 1] = bins - 1
-    cdef cnp.int64_t [::1] idx_tuple = np.zeros(classes-1, dtype=np.int64)
+    #cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
+    #    combinations(range(1, bins - 2), classes - 1)))
+
+#    sh0 = _tmp.shape[0]
+#    sh1 = _tmp.shape[1]
+#    cdef cnp.intp_t[:, ::1] idx_tuples = np.zeros((sh0, sh1 + 2),
+#                                                   dtype=np.intp)
+#    idx_tuples[:, 1:sh1 + 1] = _tmp[:]
+#    idx_tuples[:, sh1 + 1] = bins - 1
+#    cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes-1, dtype=np.intp)
 
     py_aux_thresh = np.empty(classes - 1)
-    cdef double [::1] aux_thresh = py_aux_thresh
+    cdef double[::1] aux_thresh = py_aux_thresh
 
-    for idx_tuple in idx_tuples:
-        part_sigma = 0
-        for idd in range(classes):
-            part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
-        # checking if partial sigma is higher than maximum sigma
-        if max_sigma < part_sigma:
-            for idd in range(classes - 1):
-                aux_thresh[idd] = idx_tuple[idd + 1]
-            max_sigma = part_sigma
+    cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes+1, dtype=np.intp)
+    idx_tuple[classes] = bins - 1
+
+    _find_best_rec(var_btwcls=var_btwcls, min_val=0, max_val=bins-1, idx_tuple=idx_tuple, divisions=classes-1, depth=0, max_sigma=0, aux_thresh=aux_thresh)
 
     return py_aux_thresh
+
+
+cdef _find_best_rec(double[:, ::1] var_btwcls, cnp.intp_t min_val, cnp.intp_t max_val, cnp.intp_t[::1] idx_tuple, cnp.intp_t divisions, cnp.intp_t depth, double max_sigma, double[::1] aux_thresh):
+    """"""
+    cdef cnp.intp_t idx, idd
+    cdef double part_sigma
+
+    if divisions-1 == depth:
+        for idx in range(min_val, max_val-divisions+depth):
+            # starting calculations
+            idx_tuple[depth+1] = idx
+            part_sigma = 0
+            for idd in range(divisions+1):
+                part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
+                # checking if partial sigma is higher than maximum sigma
+                if max_sigma < part_sigma:
+                    for idd in range(divisions):
+                        aux_thresh[idd] = idx_tuple[idd + 1]
+                    max_sigma = part_sigma
+
+    else:
+        for idx in range(min_val, max_val-divisions+depth):
+            idx_tuple[depth+1] = idx
+            max_sigma = _find_best_rec(var_btwcls=var_btwcls, min_val=idx+1, max_val=max_val, idx_tuple=idx_tuple, divisions=divisions, depth=depth+1, max_sigma=max_sigma, aux_thresh=aux_thresh)
+
+    return max_sigma
--- a/skimage/filters/tests/test_thresholding.py
+++ b/skimage/filters/tests/test_thresholding.py
@@ -14,6 +14,7 @@ from skimage.filters.thresholding import (threshold_local,
                                           threshold_mean,
                                           threshold_triangle,
                                           threshold_minimum,
+                                          threshold_multiotsu,
                                           try_all_threshold,
                                           _mean_std,
                                           _cross_entropy)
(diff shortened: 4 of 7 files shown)


doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,62 @@
+"""
+=======================
+Multi-Otsu Thresholding
+=======================
+
+Multi-Otsu threshold [1]_ is used to separate the pixels of an input
+image into three different classes, each one obtained according to the
```
stefanv: three, or several?

doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+
+import matplotlib
+import matplotlib.pyplot as plt
+import numpy as np
+
+from skimage import data
+from skimage.filters import threshold_multiotsu
+
+# Stablishing the font size for all plots.
```
stefanv: Set instead of "establishing"?

doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,62 @@
...
+matplotlib.rcParams['font.size'] = 9
+
+# The input image.
+image = data.camera()
+
+# Applying multi-Otsu threshold for the default value, generating
+# three classes.
+thresh = threshold_multiotsu(image)
```
stefanv: thresh -> thresholds ?

doc/examples/segmentation/plot_multiotsu.py
alexandrejaguar: Way better, right? :sweat_smile:

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,53 @@
...
+                              Py_ssize_t classes,
+                              Py_ssize_t bins,
+                              double [::1] aux_thresh):
+    """
+    Cython utility function for finding thresholds in multi-Otsu algorithm.
+    This function is only called by filters.threshold_multiotsu.
+
+    Parameters
+    ----------
+
```
sciunto: Minor: remove this empty line.

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,88 @@ def apply_hysteresis_threshold(image, low, high):
...
+    """
+    # receiving minimum and maximum values for the image type.
+    type_min, type_max = dtype_limits(image)
+
+    # calculating the histogram and the probability of each gray level.
+    hist, _ = np.histogram(image.ravel(), bins=bins,
+                           range=(type_min, type_max))
+    prob = hist / image.size
+
+    momP, momS, var_btwcls = [np.zeros((bins, bins)) for n in range(3)]
```
stefanv: I think @sciunto's question was: what does `momP`, `momS`, `var_btwcls` stand for.  So a comment here would be helpful.

sciunto: No, I was on the cython function ;)

but +1 with this comment as well.

alexandrejaguar: Yeah, you asked this before. Sorry about that, I'll work on it.

doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,62 @@
+"""
+=======================
+Multi-Otsu Thresholding
+=======================
+
+Multi-Otsu threshold [1]_ is used to separate the pixels of an input
```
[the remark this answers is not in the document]
alexandrejaguar: When I wrote this, I was trying to have something more general. I think that seems better, thanks.

skimage/filters/thresholding.py
```diff
@@ -1027,8 +1027,9 @@ def apply_hysteresis_threshold(image, low, high):
 
 
 def threshold_multiotsu(image, classes=3, bins=255):
-    """Generates multiple thresholds for an input image.
-    Based on the Multi-Otsu approach by Liao, Chen and Chung.
+    """Generates multiple thresholds for an input image, chosen maximizing the
```
alexandrejaguar: Maybe "chosen maximizing" was poorly written. Does it make any sense in this context?

alexandrejaguar:
This PR (which started as #2076) is making 3 years today! :cake::birthday::fireworks:

skimage/filters/thresholding.py
stefanv: "chosen to maximize" is probably more common

alexandrejaguar: Done.

doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,62 @@
+"""
+=======================
+Multi-Otsu Thresholding
+=======================
+
+The multi-Otsu threshold [1]_ is a thresholding algorithm thatused to separate
```
sciunto:
```suggestion
The multi-Otsu threshold [1]_ is a thresholding algorithm that is used to separate
```

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,92 @@ def apply_hysteresis_threshold(image, low, high):
     connected_to_high = sums > 0
     thresholded = connected_to_high[labels_low]
     return thresholded
+
+
+def threshold_multiotsu(image, classes=3, bins=255):
```
sciunto:
```suggestion
def threshold_multiotsu(image, classes=3, bins=256):
```

The number of bins must be 256 as in the rest of the lib, otherwise, on int8 images, you will end up with 255/256 division.

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,92 @@ def apply_hysteresis_threshold(image, low, high):
...
+
+    Parameters
+    ----------
+    image : (N, M) ndarray
+        Grayscale input image.
+    classes : int, optional
+        Number of classes to be thresholded, i.e. the number of resulting
+        regions. Accepts an integer from 2 to 5. Default is 3.
+    bins : int, optional
+        Number of bins used to calculate the histogram. Default is 255.
```
sciunto:
```suggestion
        Number of bins used to calculate the histogram. Default is 256.
```

sciunto:
I noticed an issue with the choice of the default bin value.

I do not understand the test failure...

alexandrejaguar:
> I noticed an issue with the choice of the default bin value.
> 
> I do not understand the test failure...

Changed the `bin` value. Let's see what happens now.

sciunto:
@scikit-image/core Any idea on the tests failure? a particular change in the CI?

alexandrejaguar:
> @scikit-image/core Any idea on the tests failure?

Hey @sciunto,
from the failure:

```
python\3.5.4\x64\lib\site-packages\skimage\filters\tests\test_thresholding.py:520: 
python\3.5.4\x64\lib\site-packages\skimage\filters\thresholding.py:1110: in threshold_multiotsu
    aux_thresh)

>   ???
E   ValueError: Buffer dtype mismatch, expected 'Py_ssize_t' but got 'long'
```
It's something from the cython function, which I can't understand that much.
Any hints on how we could solve this type error?

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+    cdef Py_ssize_t idd
+    cdef Py_ssize_t sh0, sh1
+    cdef double part_sigma = 0
+    # max_sigma is the maximum variance between classes.
+    cdef double max_sigma = 0
+    cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
+                                combinations(range(1, bins - 2), classes - 1)))
+    sh0 = _tmp.shape[0]
+    sh1 = _tmp.shape[1]
+    cdef Py_ssize_t [:, :] idx_tuples = np.zeros((sh0, sh1 + 2),
```
jni: @alexandrejaguar I wonder if the issue is actually here rather than in the function call (where the types are usually cast implicitly). Try setting the dtype to `Py_ssize_t` or `np.intp_t`?

But I think @stefanv has a better grasp on the different C types and where mismatches might occur...

alexandrejaguar: Hey @jni, thanks for your help!
I don't know what's happening there actually, know nothing of cython! :smile: Which dtype should I change? If you point where, I'd appreciate as well 😊

hmaarrfk: it should be `np.intp` i think

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+    cdef Py_ssize_t sh0, sh1
+    cdef double part_sigma = 0
+    # max_sigma is the maximum variance between classes.
+    cdef double max_sigma = 0
+    cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
+                                combinations(range(1, bins - 2), classes - 1)))
+    sh0 = _tmp.shape[0]
+    sh1 = _tmp.shape[1]
+    cdef Py_ssize_t [:, :] idx_tuples = np.zeros((sh0, sh1 + 2), 
+                                                  dtype=np.int)
```
hmaarrfk:
```suggestion
                                                  dtype=np.intp)
```

skimage/filters/_multiotsu.pyx
hmaarrfk: might be a little more than that. Let me try to have a go at it.

hmaarrfk: I think i'm hitting this on my local machine
https://bugs.python.org/issue32259

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,92 @@ def apply_hysteresis_threshold(image, low, high):
...
+            if (momP[u, v] != 0):
+                var_btwcls[u, v] = momS[u, v]**2 / momP[u, v]
+            else:
+                var_btwcls[u, v] = 0
+
+    # finding max threshold candidates, depending on classes.
+    # number of thresholds is equal to number of classes - 1.
+    aux_thresh = np.zeros(classes - 1)
+    aux_thresh = _find_threshold_multiotsu(var_btwcls, classes, bins,
+                                           aux_thresh)
```
hmaarrfk: I think somebody once said that functions should either "mutate data" or "return results", this function seems to do both. I would discourage this kind of style.

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+    cdef Py_ssize_t [::1] idx_tuple
+    for idx_tuple in idx_tuples:
+        part_sigma = 0
+        for idd in range(0, classes):
+            part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
+        # checking if partial sigma is higher than maximum sigma
+        if max_sigma < part_sigma:
+            for idd in range(0, classes - 1):
+                aux_thresh[idd] = idx_tuple[idd + 1]
+                max_sigma = part_sigma
```
hmaarrfk: The assignment to `max_sigma` can be moved out of the loop.

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+    sh0 = _tmp.shape[0]
+    sh1 = _tmp.shape[1]
+    cdef Py_ssize_t [:, :] idx_tuples = np.zeros((sh0, sh1 + 2), 
+                                                  dtype=np.int)
+    idx_tuples[:, 1:sh1 + 1] = _tmp[:]
+    idx_tuples[:, sh1 + 1] = bins - 1
+    cdef Py_ssize_t [::1] idx_tuple
+    for idx_tuple in idx_tuples:
+        part_sigma = 0
+        for idd in range(0, classes):
```
hmaarrfk: We are writing "python", so I would omit the `0, `. ans imply opt for `range(classes)`

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+            part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
+        # checking if partial sigma is higher than maximum sigma
+        if max_sigma < part_sigma:
+            for idd in range(0, classes - 1):
```
hmaarrfk: `range(classes-1)`

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+                aux_thresh[idd] = idx_tuple[idd + 1]
```
hmaarrfk: Because you are doing this assignmetn repeatidly, `aux_thresh` should be of the same type as `idx_tuple`.
That is, `aux_thresh` should be a memoryview of `Py_ssize_t`

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+import numpy as np
+
+cimport numpy as cnp
+cimport cython
+from itertools import combinations
+
+def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
+                              Py_ssize_t classes,
+                              Py_ssize_t bins,
+                              double [::1] aux_thresh):
```
hmaarrfk: In your function, you are doing alot of allocations inside cython. I suggest you allocate `aux_thresh` in cython too!

skimage/filters/_multiotsu.pyx
hmaarrfk:
```suggestion
    cdef Py_ssize_t [:, ::1] idx_tuples = np.zeros((sh0, sh1 + 2), 
```
You should specify that this memory view is `C contiguous`

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+    cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
+                                combinations(range(1, bins - 2), classes - 1)))
+    sh0 = _tmp.shape[0]
+    sh1 = _tmp.shape[1]
+    cdef Py_ssize_t [:, :] idx_tuples = np.zeros((sh0, sh1 + 2), 
+                                                  dtype=np.int)
+    idx_tuples[:, 1:sh1 + 1] = _tmp[:]
+    idx_tuples[:, sh1 + 1] = bins - 1
+    cdef Py_ssize_t [::1] idx_tuple
+    for idx_tuple in idx_tuples:
```
hmaarrfk:
```suggestion
    with nogil:
        for idx in range(sh0):
            idx_tuple = idx_tuples[idx]
```

Don't accept this change as you need to indent everything else, but sadly this is cython, and you should use `nogil` and c style looping.

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+    cdef double max_sigma = 0
+    cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
+                                combinations(range(1, bins - 2), classes - 1)))
+    sh0 = _tmp.shape[0]
+    sh1 = _tmp.shape[1]
+    cdef Py_ssize_t [:, :] idx_tuples = np.zeros((sh0, sh1 + 2), 
+                                                  dtype=np.int)
+    idx_tuples[:, 1:sh1 + 1] = _tmp[:]
+    idx_tuples[:, sh1 + 1] = bins - 1
+    cdef Py_ssize_t [::1] idx_tuple
```
hmaarrfk: You can add your allocation for 
```
    py_aux_thresh = np.zeros(classes-1, dtype=np.intp)
    cdef Py_ssize_t [::1] aux_thresh = py_aux_thresh
```
here

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,54 @@
...
+        part_sigma = 0
+        for idd in range(0, classes):
+            part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
+        # checking if partial sigma is higher than maximum sigma
+        if max_sigma < part_sigma:
+            for idd in range(0, classes - 1):
+                aux_thresh[idd] = idx_tuple[idd + 1]
+                max_sigma = part_sigma
+
+    return aux_thresh
```
hmaarrfk:
```suggestion
    return py_aux_thresh
```

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,92 @@ def apply_hysteresis_threshold(image, low, high):
...
+    for u in range(1, bins):
+        for v in range(u+1, bins):
+            if (momP[u, v] != 0):
+                var_btwcls[u, v] = momS[u, v]**2 / momP[u, v]
+            else:
+                var_btwcls[u, v] = 0
+
+    # finding max threshold candidates, depending on classes.
+    # number of thresholds is equal to number of classes - 1.
+    aux_thresh = np.zeros(classes - 1)
```
hmaarrfk: if you do take the suggestion to do all the allocations in cython, then this line isn't needed..

skimage/filters/thresholding.py
hmaarrfk: And you don't need to pass `aux_thresh` in as a parameter.

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,92 @@ def apply_hysteresis_threshold(image, low, high):
...
+    aux_thresh = _find_threshold_multiotsu(var_btwcls, classes, bins,
+                                           aux_thresh)
+
+    # correcting values according to minimum and maximum values.
+    idx_thresh = np.asarray(aux_thresh) * (type_max-type_min) / bins
```
hmaarrfk: by returniing `py_aux_thresh` you avoid returning a memoryview, and you don't need to cast the result back as a `Pythonic numpy array`

skimage/filters/tests/test_thresholding.py
```diff
@@ -501,6 +502,25 @@ def test_niblack_sauvola_pathological_image():
     assert not np.any(np.isnan(threshold_niblack(src_img)))
 
 
+def test_bimodal_multiotsu_hist():
+    image = data.camera()
+    thr_otsu = threshold_otsu(image)
+    thr_multi = threshold_multiotsu(image, classes=2)
+    assert thr_otsu == thr_multi
```
hmaarrfk: I'm getting float to int comparison errors on this line.

skimage/filters/tests/test_thresholding.py
```diff
@@ -501,6 +502,25 @@ def test_niblack_sauvola_pathological_image():
...
+
+
+def test_check_multiotsu_results():
```
hmaarrfk: this test takes for ever on my laptop. Is it entirely necessary?

hmaarrfk:
Alright, I think i gave my comments with respect to this low level programming business.

That test was burning up my computer.

Here is the diff that I have on my local version compared to yours
<details>

```patch
diff --git a/skimage/filters/_multiotsu.pyx b/skimage/filters/_multiotsu.pyx
index f1e325e36..6e273576d 100644
--- a/skimage/filters/_multiotsu.pyx
+++ b/skimage/filters/_multiotsu.pyx
@@ -10,8 +10,7 @@ from itertools import combinations
 
 def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
                               Py_ssize_t classes,
-                              Py_ssize_t bins,
-                              double [::1] aux_thresh):
+                              Py_ssize_t bins):
     """
     Cython utility function for finding thresholds in multi-Otsu algorithm.
     This function is only called by filters.threshold_multiotsu.
@@ -27,28 +26,36 @@ def _find_threshold_multiotsu(double [:, ::1] var_btwcls,
     aux_thresh : array
         thresholds to be returned
     """
-    cdef Py_ssize_t idd
-    cdef Py_ssize_t sh0, sh1
     cdef double part_sigma = 0
     # max_sigma is the maximum variance between classes.
     cdef double max_sigma = 0
-    cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
-                                combinations(range(1, bins - 2), classes - 1)))
+    cdef Py_ssize_t idd
+    cdef Py_ssize_t sh0, sh1
+    cdef Py_ssize_t [:, ::1] _tmp = np.array(
+        list(combinations(range(1, bins - 2), classes - 1)),
+        dtype=np.intp)
+
     sh0 = _tmp.shape[0]
     sh1 = _tmp.shape[1]
-    cdef Py_ssize_t [:, :] idx_tuples = np.zeros((sh0, sh1 + 2), 
-                                                  dtype=np.int)
+    cdef Py_ssize_t [:, ::1] idx_tuples = np.zeros(
+        (sh0, sh1 + 2), dtype=np.intp)
+
     idx_tuples[:, 1:sh1 + 1] = _tmp[:]
     idx_tuples[:, sh1 + 1] = bins - 1
     cdef Py_ssize_t [::1] idx_tuple
-    for idx_tuple in idx_tuples:
-        part_sigma = 0
-        for idd in range(0, classes):
-            part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
-        # checking if partial sigma is higher than maximum sigma
-        if max_sigma < part_sigma:
-            for idd in range(0, classes - 1):
-                aux_thresh[idd] = idx_tuple[idd + 1]
+
+    py_aux_thresh = np.zeros(classes-1, dtype=np.intp)
+    cdef Py_ssize_t [::1] aux_thresh = py_aux_thresh
+
+    with nogil:
+        for idx in range(sh0):
+            idx_tuple = idx_tuples[idx]
+            part_sigma = 0
+            for idd in range(classes):
+                part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
+            # checking if partial sigma is higher than maximum sigma
+            if max_sigma < part_sigma:
+                aux_thresh[:classes-1] = idx_tuple[1:classes]
                 max_sigma = part_sigma
 
-    return aux_thresh
+    return py_aux_thresh
diff --git a/skimage/filters/tests/test_thresholding.py b/skimage/filters/tests/test_thresholding.py
index 61efdf916..94a90c313 100644
--- a/skimage/filters/tests/test_thresholding.py
+++ b/skimage/filters/tests/test_thresholding.py
@@ -506,7 +506,7 @@ def test_bimodal_multiotsu_hist():
     image = data.camera()
     thr_otsu = threshold_otsu(image)
     thr_multi = threshold_multiotsu(image, classes=2)
-    assert thr_otsu == thr_multi
+    assert thr_otsu == int(round(thr_multi[0]))
 
 
 def test_check_multiotsu_results():
diff --git a/skimage/filters/thresholding.py b/skimage/filters/thresholding.py
index 956446071..e342b16c9 100644
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1105,11 +1105,9 @@ def threshold_multiotsu(image, classes=3, bins=256):
 
     # finding max threshold candidates, depending on classes.
     # number of thresholds is equal to number of classes - 1.
-    aux_thresh = np.zeros(classes - 1)
-    aux_thresh = _find_threshold_multiotsu(var_btwcls, classes, bins,
-                                           aux_thresh)
+    aux_thresh = _find_threshold_multiotsu(var_btwcls, classes, bins)
 
     # correcting values according to minimum and maximum values.
-    idx_thresh = np.asarray(aux_thresh) * (type_max-type_min) / bins
+    idx_thresh = aux_thresh * (type_max-type_min) / bins
 
     return idx_thresh
```

</details>

skimage/filters/tests/test_thresholding.py
hmaarrfk: Is the fact that this test is taking so long speaking that the implementation might be inefficient? I really don't know, but it might be something to ponder about.

skimage/filters/_multiotsu.pyx
jni: @hmaarrfk I actually tend to prefer all allocations to happen in Cython. That makes it clearer where gil is needed vs not needed, and whether your Cython is optimal (since you should have exactly 0 yellow lines inside the function).

skimage/filters/_multiotsu.pyx
jni: @hmaarrfk if code should not be accepted as is, just use code blocks and not suggestion blocks!

jni:
@hmaarrfk please see the discussion in https://github.com/scikit-image/scikit-image/pull/3385 regarding elegance vs speed. If you can implement a fast itertools.combinations in Cython, this might run much faster! None of us have succeeded so far, but it should definitely be possible.

skimage/filters/_multiotsu.pyx
hmaarrfk: Sure. Do the allocations how you want, but don't Pass and return the same thing.

sciunto:
When I asked for help, I was puzzled by this: https://travis-ci.org/scikit-image/scikit-image/jobs/531950361#L2419

hmaarrfk:
I think that is a timeout!

hmaarrfk:
yeah, it seems `np.combinations` isn't a thing. itertools is definiteley slowing things down beyond belief.

I'm wondering if we can cache this matrix for users that repeatidly this algorithm.

hmaarrfk:
We could create a little function that takes `nbins` and `classes`, and decorate it with python's `lru_cache`
https://docs.python.org/3/library/functools.html#functools.lru_cache

skimage/filters/tests/test_thresholding.py
hmaarrfk: I'm not sure if it is a timeout, but I definitely thought my computer was frozen while running these tests. If you use pytest parametrize like the test below, it will give the test suite 3 outputs which will give Travis a sense of progress and might not time out.

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,92 @@ def apply_hysteresis_threshold(image, low, high):
...
+    prob = hist / image.size
+
+    # defining arrays to store the zeroth (momP, cumulative probability)
+    # and first (momS, mean) moments, and the variance between classes
+    # (var_btwcls).
+    momP, momS, var_btwcls = [np.zeros((bins, bins)) for n in range(3)]
+
+    # building the lookup tables.
+    # step 1: calculating the diagonal.
+    for u in range(1, bins):
```
hmaarrfk: This looks alot like c code. It feel natural toove these loops into Cython, so that you can use nogil and likely faster indexing.

skimage/filters/_multiotsu.pyx
stefanv: We should not be mixing types here.  It is likely that `Py_ssize_t` happens to be the right type, but you can write this as:

```
cdef cnp.int64_t [::1] idx_tuple = np.zeros(..., dtype=np.int64)
```

Then the match is exactly right.

hmaarrfk: Thanks @stefanv, This is one of my big pet peves with data types. Everybody likes to rename them.....

stefanv: :crown: I crown you Guardian of the Types.  (@jni is the Guardian of the Dimensions, so it's good to we can separate our pet peeves so well).

alexandrejaguar:
@jni @hmaarrfk I found something! Don't know if it alone would solve everything, though :smile:

alexandrejaguar:
> @hmaarrfk please see the discussion in #3385 regarding elegance vs speed. If you can implement a fast itertools.combinations in Cython, this might run much faster!

> yeah, it seems np.combinations isn't a thing. itertools is definiteley slowing things down beyond belief.

@jni @hmaarrfk Could we take advantage of [this SO answer](https://stackoverflow.com/questions/16003217/n-d-version-of-itertools-combinations-in-numpy), maybe?

hmaarrfk:
Is there a particular answer in that post you want to discuss?

alexandrejaguar:
@hmaarrfk I was thinking maybe the accepted one could help us.
I now implemented some ideas you and the crew gave me, but I didn't understand some of them. I'll ask you about them.

hmaarrfk:
So the only difference about that one is the fact that they are using a dedicated function `fromiter` that can help you. You avoid creating the numpy list, but you are still traversing a slow generator.

It will definitely help, the question is how much.

Do you know about the `%timeit` magic for ipython, it can be a good tool to estimate how much speedup can you can expect.
```
In [2]: import numpy as np                                                                                                                            

In [3]: np.array(list(0 for i in range(10)))                                                                                                          
Out[3]: array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0])

In [4]: %timeit np.array(list(0 for i in range(1000)))                                                                                                
98.9 µs ± 427 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [5]: %timeit np.fromiter(list(0 for i in range(1000)), dtype=int, count=-1)                                                                        
62.9 µs ± 453 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [6]: %timeit np.fromiter(list(0 for i in range(1000)), dtype=int, count=1000)                                                                      
63 µs ± 1.01 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [7]: N = 10000                                                                                                                                     

In [8]: %timeit np.fromiter(list(0 for i in range(N)), dtype=int, count=N)                                                                            
612 µs ± 7.46 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,57 @@
...
+        number of bins used in the histogram
+    aux_thresh : array
+        thresholds to be returned
+    """
+    cdef Py_ssize_t idd
+    cdef Py_ssize_t sh0, sh1
+    cdef double part_sigma = 0
+    # max_sigma is the maximum variance between classes.
+    cdef double max_sigma = 0
+    cdef Py_ssize_t [:, ::1] _tmp = np.array(list(
```
alexandrejaguar: @jni @sciunto @hmaarrfk I think the memory thief is here!

alexandrejaguar:
Got a help today from @seberg with the cython functions, and now it's working!
Looks ugly @jni, but we'll work on that as well :)

alexandrejaguar:
@sciunto the default value for `bins` should be 255 instead of 256, considering the current implementation.

sciunto:
I need an argument for that... This is not intuitive. 

I'm also wondering if bins shouldn't be removed from the parameters and instead, its value can be defined from the dtype_limits of the image. Is it costly to increase the number of bins for int16 or int32 images?

seberg:
The complexity at least for more classes is probably something in the very rough direction of `nbins**(classes-1)/(classes-1)! * classes**2` so reducing the `nbins` should be very relevant. (I wondered if there are not better approaches then brute forcing the best solution, but I guess such methods could be added later probably, and this optimization is probably very difficult).

skimage/filters/tests/test_thresholding.py
hmaarrfk:
```patch
-def test_check_multiotsu_results():
+@pytest.mark.parametrize("classes", range(3, 6))
+def test_check_multiotsu_results(classes):
     image = 0.25 * np.array([[0, 1, 2, 3, 4],
                              [0, 1, 2, 3, 4],
                              [0, 1, 2, 3, 4],
                              [0, 1, 2, 3, 4]])
 
-    for idx in range(3, 6):
-        thr_multi = threshold_multiotsu(image,
-                                        classes=idx)
-        assert len(thr_multi) == idx-1
+    thr_multi = threshold_multiotsu(image,
+                                    classes=classes)
+    assert len(thr_multi) == classes - 1
```

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,79 @@
...
+        for idx in range(min_val, max_val-divisions+depth):
+            # starting calculations
+            idx_tuple[depth+1] = idx
+            part_sigma = 0
+            for idd in range(divisions+1):
+                part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
+                # checking if partial sigma is higher than maximum sigma
+                if max_sigma < part_sigma:
+                    for idd in range(divisions):
+                        aux_thresh[idd] = idx_tuple[idd + 1]
```
hmaarrfk: These two should be the same type., then you can assign with
```patch
-                    for idd in range(divisions):
-                        aux_thresh[idd] = idx_tuple[idd + 1]
+                    aux_thresh[:divisions] = idx_tuple[1:divisions +1]

```

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,79 @@
...
+
+    cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes+1, dtype=np.intp)
+    idx_tuple[classes] = bins - 1
+
+    _find_best_rec(var_btwcls=var_btwcls, min_val=0, max_val=bins-1, idx_tuple=idx_tuple, divisions=classes-1, depth=0, max_sigma=0, aux_thresh=aux_thresh)
+
+    return py_aux_thresh
+
+
+cdef _find_best_rec(double[:, ::1] var_btwcls, cnp.intp_t min_val, cnp.intp_t max_val, cnp.intp_t[::1] idx_tuple, cnp.intp_t divisions, cnp.intp_t depth, double max_sigma, double[::1] aux_thresh):
```
hmaarrfk: this function should be declared with nogil
```cython
cdef double _find_best_rec( [...] ) nogil
```

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,79 @@
...
+#    idx_tuples[:, sh1 + 1] = bins - 1
+#    cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes-1, dtype=np.intp)
+
+    py_aux_thresh = np.empty(classes - 1)
+    cdef double[::1] aux_thresh = py_aux_thresh
+
+    cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes+1, dtype=np.intp)
+    idx_tuple[classes] = bins - 1
+
+    _find_best_rec(var_btwcls=var_btwcls, min_val=0, max_val=bins-1, idx_tuple=idx_tuple, divisions=classes-1, depth=0, max_sigma=0, aux_thresh=aux_thresh)
```
hmaarrfk: Should be called `with nogil:`
```patch
-    _find_best_rec([...])
+    with nogil:
+        _find_best_rec([...])
```

skimage/filters/thresholding.py
hmaarrfk: Is this a bug: https://github.com/scikit-image/scikit-image/blob/b573f9c092732b4beb3cc06d5fa37db5c7454c6f/skimage/io/_plugins/_histograms.pyx#L52

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,90 @@ def apply_hysteresis_threshold(image, low, high):
...
+                var_btwcls[u, v] = momS[u, v]**2 / momP[u, v]
+            else:
+                var_btwcls[u, v] = 0
+
+    # finding max threshold candidates, depending on classes.
+    # number of thresholds is equal to number of classes - 1.
+    aux_thresh = _find_threshold_multiotsu(var_btwcls, classes, bins)
+
+    # correcting values according to minimum and maximum values.
+    idx_thresh = aux_thresh * (type_max-type_min) / bins
```
hmaarrfk: Once you accept @sciunto's suggestion, you will need to change this
```suggestion
    idx_thresh = aux_thresh * (type_max-type_min) / (bins - 1)
```

skimage/filters/_multiotsu.pyx
hmaarrfk: @alexandrejaguar it seems that from this line, you are already consider `max_val = bins -1`.

For an 8 bit image, maximum value is 255, and in that case, bins=256. Therefore, your algorithm is written with `bins = 256` in mind right?

alexandrejaguar:
@hmaarrfk I'm having several compiling errors following your ideas on `nogil` and the variable substitution:
```
    ERROR: Complete output from command python setup.py egg_info:
    ERROR: Partial import of skimage during the build process.
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
        cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes+1, dtype=np.intp)
        idx_tuple[classes] = bins - 1
    
        with nogil:
            _find_best_rec(var_btwcls=var_btwcls, min_val=0, max_val=bins-1,
                           idx_tuple=idx_tuple, divisions=classes-1, depth=0,
                                                                ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:53:61: Cannot convert 'Py_ssize_t' to memoryviewslice
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
    cdef _find_best_rec(double[:, ::1] var_btwcls, cnp.intp_t min_val, cnp.intp_t max_val, cnp.intp_t[::1] idx_tuple, double[::1] divisions, cnp.intp_t depth, double max_sigma, double[::1] aux_thresh):
        """"""
        cdef cnp.intp_t idx, idd
        cdef double part_sigma
    
        if divisions-1 == depth:
                   ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:65:16: Invalid operand types for '-' (double[::1]; long)
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
        """"""
        cdef cnp.intp_t idx, idd
        cdef double part_sigma
    
        if divisions-1 == depth:
            for idx in range(min_val, max_val-divisions+depth):
                                            ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:66:41: Invalid operand types for '-' (intp_t; double[::1])
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
        if divisions-1 == depth:
            for idx in range(min_val, max_val-divisions+depth):
                # starting calculations
                idx_tuple[depth+1] = idx
                part_sigma = 0
                for idd in range(divisions+1):
                                         ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:70:38: Invalid operand types for '+' (double[::1]; long)
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
                    part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
                    # checking if partial sigma is higher than maximum sigma
                    if max_sigma < part_sigma:
                        #for idd in range(divisions):
                        #    aux_thresh[idd] = idx_tuple[idd + 1]
                        aux_thresh[:divisions] = idx_tuple[1:divisions + 1]
                                                                      ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:76:67: Invalid operand types for '+' (double[::1]; long)
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
                    part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
                    # checking if partial sigma is higher than maximum sigma
                    if max_sigma < part_sigma:
                        #for idd in range(divisions):
                        #    aux_thresh[idd] = idx_tuple[idd + 1]
                        aux_thresh[:divisions] = idx_tuple[1:divisions + 1]
                                                         ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:76:54: Different base types for memoryviews (intp_t, double)
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
                        #    aux_thresh[idd] = idx_tuple[idd + 1]
                        aux_thresh[:divisions] = idx_tuple[1:divisions + 1]
                        max_sigma = part_sigma
    
        else:
            for idx in range(min_val, max_val-divisions+depth):
                                            ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:80:41: Invalid operand types for '-' (intp_t; double[::1])
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
    
        cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes+1, dtype=np.intp)
        idx_tuple[classes] = bins - 1
    
        with nogil:
            _find_best_rec(var_btwcls=var_btwcls, min_val=0, max_val=bins-1,
                         ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:52:22: Discarding owned Python object not allowed without gil
    
    Error compiling Cython file:
    ------------------------------------------------------------
    ...
    
        cdef cnp.intp_t[::1] idx_tuple = np.zeros(classes+1, dtype=np.intp)
        idx_tuple[classes] = bins - 1
    
        with nogil:
            _find_best_rec(var_btwcls=var_btwcls, min_val=0, max_val=bins-1,
                         ^
    ------------------------------------------------------------
    
    skimage/filters/_multiotsu.pyx:52:22: Calling gil-requiring function not allowed without gil
      File "/home/alex/sources/scikit-image/setup.py", line 243, in <module>
        **extra
      File "/home/alex/anaconda3/lib/python3.6/site-packages/numpy/distutils/core.py", line 137, in setup
        config = configuration()
      File "/home/alex/sources/scikit-image/setup.py", line 156, in configuration
        config.add_subpackage('skimage')
      File "/home/alex/anaconda3/lib/python3.6/site-packages/numpy/distutils/misc_util.py", line 1036, in add_subpackage
        caller_level = 2)
      File "/home/alex/anaconda3/lib/python3.6/site-packages/numpy/distutils/misc_util.py", line 1005, in get_subpackage
        caller_level = caller_level + 1)
      File "/home/alex/anaconda3/lib/python3.6/site-packages/numpy/distutils/misc_util.py", line 942, in _get_configuration_from_setup_py
        config = setup_module.configuration(*args)
      File "skimage/setup.py", line 16, in configuration
        config.add_subpackage('filters')
      File "/home/alex/anaconda3/lib/python3.6/site-packages/numpy/distutils/misc_util.py", line 1036, in add_subpackage
        caller_level = 2)
      File "/home/alex/anaconda3/lib/python3.6/site-packages/numpy/distutils/misc_util.py", line 1005, in get_subpackage
        caller_level = caller_level + 1)
      File "/home/alex/anaconda3/lib/python3.6/site-packages/numpy/distutils/misc_util.py", line 942, in _get_configuration_from_setup_py
        config = setup_module.configuration(*args)
      File "skimage/filters/setup.py", line 20, in configuration
        '_multiotsu.pyx'], working_path=base_path)
      File "/home/alex/sources/scikit-image/skimage/_build.py", line 66, in cython
        compiler_directives={'language_level': 3})
      File "/home/alex/anaconda3/lib/python3.6/site-packages/Cython/Build/Dependencies.py", line 1097, in cythonize
        cythonize_one(*args)
      File "/home/alex/anaconda3/lib/python3.6/site-packages/Cython/Build/Dependencies.py", line 1220, in cythonize_one
        raise CompileError(None, pyx_file)
    Cython.Compiler.Errors.CompileError: /home/alex/sources/scikit-image/skimage/filters/_multiotsu.pyx
    Compiling /home/alex/sources/scikit-image/skimage/filters/_multiotsu.pyx because it changed.
    [1/1] Cythonizing /home/alex/sources/scikit-image/skimage/filters/_multiotsu.pyx
    ----------------------------------------
ERROR: Command "python setup.py egg_info" failed with error code 1 in /home/alex/sources/scikit-image/
```

hmaarrfk:
It thinks divisions is a memory view. can you provide your function declaration?

alexandrejaguar:
@hmaarrfk sure (not actually mine; credits to @seberg):

```
cdef _find_best_rec(double[:, ::1] var_btwcls, cnp.intp_t min_val, cnp.intp_t max_val, cnp.intp_t[::1] idx_tuple, cnp.intp_t divisions, cnp.intp_t depth, double max_sigma, double[::1] aux_thresh):
    """"""
    cdef cnp.intp_t idx, idd
    cdef double part_sigma

    if divisions-1 == depth:
        for idx in range(min_val, max_val-divisions+depth):
            # starting calculations
            idx_tuple[depth+1] = idx
            part_sigma = 0
            for idd in range(divisions+1):
                part_sigma += var_btwcls[1 + idx_tuple[idd], idx_tuple[idd+1]]
                # checking if partial sigma is higher than maximum sigma
                if max_sigma < part_sigma:
                    for idd in range(divisions):
                        aux_thresh[idd] = idx_tuple[idd + 1]
                    max_sigma = part_sigma

    else:
        for idx in range(min_val, max_val-divisions+depth):
            idx_tuple[depth+1] = idx
            max_sigma = _find_best_rec(var_btwcls=var_btwcls, min_val=idx+1, max_val=max_val, idx_tuple=idx_tuple, divisions=divisions, depth=depth+1, max_sigma=max_sigma, aux_thresh=aux_thresh)

    return max_sigma
```

hmaarrfk:
maybe you can push what you have and i can just look at it locally tonight?

alexandrejaguar:
@hmaarrfk everything I have is there, Mark. Sorry about that

hmaarrfk:
i kinda meant even your attempt at nogil. it would be easier to see the bugs on my own computer.

alexandrejaguar:
@hmaarrfk I deleted it, sorry :disappointed: I use this code for production, you know :disappointed_relieved:

hmaarrfk:
Thats fine.

I'll make a PR to your repo with the suggestions in that case.

hmaarrfk:
xref: https://github.com/alexandrejaguar/scikit-image/pull/4

alexandrejaguar:
@hmaarrfk thanks for your help Mark!

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,92 @@
...
+
+    Parameters
+    ----------
+    var_btwcls : 2-d array
+        array of variance between classes
+    classes : int
+        number of classes to be segmented
+    bins : int
+        number of bins used in the histogram
+    aux_thresh : array
```
sciunto: Can be removed :)

hmaarrfk:
I think this needs a correctness test for the mult case. Currently, only the shape of the output is tested.

alexandrejaguar:
> I think this needs a correctness test for the mult case. Currently, only the shape of the output is tested.

@hmaarrfk could you be more specific? Isn't it that `test_check_multiotsu_results()` does?

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,88 @@ def apply_hysteresis_threshold(image, low, high):
...
+    variance between the desired classes. Based on the Multi-Otsu approach by
+    Liao, Chen and Chung.
+
+    Parameters
+    ----------
+    image : (N, M) ndarray
+        Grayscale input image.
+    classes : int, optional
+        Number of classes to be thresholded, i.e. the number of resulting
+        regions. Accepts an integer from 2 to 5. Default is 3.
```
sciunto:
```suggestion
        regions. Accepts an integer from 2 to 5.
```

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,88 @@ def apply_hysteresis_threshold(image, low, high):
...
+    nbins : int, optional
+        Number of bins used to calculate the histogram. Default is 256.
```
sciunto:
```suggestion
        Number of bins used to calculate the histogram.
```

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1024,88 @@ def apply_hysteresis_threshold(image, low, high):
     connected_to_high = sums > 0
     thresholded = connected_to_high[labels_low]
     return thresholded
+
+
+def threshold_multiotsu(image, classes=3, nbins=256):
+    """Generates multiple thresholds for an input image, chosen to maximize the
```
sciunto:
```suggestion
    """Generates multiple thresholds for an input image.
    
    These thresholds are chosen to maximize the
```

A one line short summary is preferred.

doc/examples/segmentation/plot_multiotsu.py
```diff
@@ -0,0 +1,62 @@
...
+matplotlib.rcParams['font.size'] = 9
+
+# The input image.
+image = data.camera()
+
+# Applying multi-Otsu threshold for the default value, generating
+# three classes.
+thresholds = threshold_multiotsu(image)
+
+# Using the values on thresh, we generate the three regions.
```
emmanuelle: Using the threshold values

alexandrejaguar: @seberg helped with that Emma! But without your work we couldn't do it :smile: Thank you very much!

alexandrejaguar:
@seberg helped with that Emma! But without your functions we wouldn't be able to do it :smile: Thank you very much!

skimage/filters/thresholding.py
alexandrejaguar: I inserted the rest of the sentence in the Notes section.

skimage/filters/_multiotsu.pyx
```diff
@@ -0,0 +1,77 @@
...
+
+    return py_aux_thresh
+
+
+cdef double _find_best_rec(double[:, ::1] var_btwcls, cnp.intp_t min_val,
+                           cnp.intp_t max_val, Py_ssize_t[::1] idx_tuple,
+                           cnp.intp_t divisions, cnp.intp_t depth,
+                           double max_sigma, Py_ssize_t[::1] aux_thresh) nogil:
+    """
+    Recursive function for finding max_sigma.
```
jni: Recursion! Brilliant! =D

@alexandrejaguar even though it is a private function, would you mind adding a docstring explaining what it does and what all the arguments are? I find it very hard to read these long argument lists.

benchmarks/benchmark_filters.py
```diff
@@ -12,6 +12,35 @@ def setup(self):
...
+        other ``peakmem_`` functions below).
+
+        References
+        ----------
+        .. [1]: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory
+        """
+        pass
+
+    def peakmem_threshold_multiotsu(self, classes):
+        filters.threshold_multiotsu(self.image, classes=classes)
```
jni: Very nice! Could you show the results of `asv dev` on this? I'm just curious to see how it compares to your earlier attempts.

alexandrejaguar:
> Awesome persistence! =)

Where I come from it's called "stubbornness"! :smile:

benchmarks/benchmark_filters.py
alexandrejaguar: This version is way better:
```
[ 18.75%] ··· benchmark_filters.MultiOtsu.peakmem_reference                  ok
[ 18.75%] ··· ========= =======
               classes         
              --------- -------
                  3      71.4M 
                  4      71.4M 
                  5      71.4M 
              ========= =======

[ 20.83%] ··· ...lters.MultiOtsu.peakmem_threshold_multiotsu                 ok
[ 20.83%] ··· ========= =======
               classes         
              --------- -------
                  3      74.7M 
                  4      74.7M 
                  5      74.7M 
              ========= =======

[ 22.92%] ··· ..._filters.MultiOtsu.time_threshold_multiotsu                 ok
[ 22.92%] ··· ========= ==========
               classes            
              --------- ----------
                  3      82.7±0ms 
                  4      148±0ms  
                  5      1.38±0s  
              ========= ==========
```

jni: @alexandrejaguar awesome!

Watch that exponential explode LOL. But yes, even with 5 classes it's pretty damn good! I'm so happy. =)

alexandrejaguar:
@jni I was testing in some larger images, and it goes to around 1 minute with six classes!On May 27, 2019 4:34 AM, Juan Nunez-Iglesias <notifications@github.com> wrote:@jni commented on this pull request.



In benchmarks/benchmark_filters.py:
> +
+        Peakmem benchmarks measure the maximum amount of RAM used by a
+        function. However, this maximum also includes the memory used
+        during the setup routine (as of asv 0.2.1; see [1]_).
+        Measuring an empty peakmem function might allow us to disambiguate
+        between the memory used by setup and the memory used by target (see
+        other ``peakmem_`` functions below).
+
+        References
+        ----------
+        .. [1]: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#peak-memory
+        """
+        pass
+
+    def peakmem_threshold_multiotsu(self, classes):
+        filters.threshold_multiotsu(self.image, classes=classes)

@alexandrejaguar awesome!
Watch that exponential explode LOL. But yes, even with 5 classes it's pretty damn good! I'm so happy. =)

benchmarks/benchmark_filters.py
alexandrejaguar: @jni I was testing in some larger images (2000 X 2000), and it goes to around 1 minute with six classes!

jni:
:tada:

alexandrejaguar:
:tada: :smile: Thanks everyone! Learned A LOT in the process. Thanks @jni @sciunto @soupault @emmanuelle @hmaarrfk @seberg for the ideas, help and reviews! I'm also grateful to @emmanuelle @hmaarrfk @seberg for the cython code. Thanks again!

benchmarks/benchmark_filters.py
seberg: I think Stéfan suggested to put the complexity in the docs somewhere. I think it is approximated by (for the brute forcing step, not the preparation):
```
combinations = nbins**(classes-1)/(classes-1)!
variance_calc = classes + 1
complexity = O(combinations * variance_calc)
```

hmaarrfk:
Hey all, this is great work.

I think that the "correctness test" is still incomplete. Opening an issue is enough at this stage, but the contents of `thr_multi` are never checked in any test.

alexandrejaguar:
@hmaarrfk we compare it with the Otsu output:
```
def test_bimodal_multiotsu_hist():
    thr_multi = threshold_multiotsu(image, classes=2)
    assert thr_otsu == thr_multi
```
Maybe we could get `data.camera()` and obtain multi-Otsu's thresholds from Fiji's version to try to compare them. Is that what you want?

hmaarrfk:
Right, so you compare the case of 2 classes, but that doesn't guarantee
that you've added new functionality.

The only use case for multi-otsu is to set classes!=2, for which there is
no real correctness test. Just an idirect guarantee from:
1. THe classes = 2 test working
2. Other classes returning the right size
3. Assuming code paths are similar.

stefanv:
Maybe generate an image, drawing from a mixture of Gaussians, and then see whether you get roughly the means of those classes out under the threshold labels?

stefanv:
Thanks for the complexity calculation, @seberg. @alexandrejaguar It doesn't look as though that made it in; another quick update PR?

alexandrejaguar:
> 2. Other classes returning the right size

@hmaarrfk this is exactly what the other test calculates:
```
def test_check_multiotsu_results():
    image = 0.25 * np.array([[0, 1, 2, 3, 4],
                             [0, 1, 2, 3, 4],
                             [0, 1, 2, 3, 4],
                             [0, 1, 2, 3, 4]])

    for idx in range(3, 6):
        thr_multi = threshold_multiotsu(image,
                                        classes=idx)
        assert len(thr_multi) == idx-1
```
Am I missing something? Also, please clarify 3. Thanks!

hmaarrfk:
Right, this is a really "indirect" way of testing that `thr_multi` is correct.

I can totally see somebody optimizing `threshold_multiotsu(classes=2)` to be a thin wrapper over `otsu`, meaning that the correctness codepath of `classes=3` would never get tested.

none of the values of `thr_multi` are tested. What should their values be?

alexandrejaguar:
>> Maybe we could get data.camera() and obtain multi-Otsu's thresholds from Fiji's version to try to compare them. Is that what you want?

> none of the values of thr_multi are tested. What should their values be?

@hmaarrfk again, is this what you want? Would you like something more hands-on...? Maybe the math involved...? :confused:

hmaarrfk:
> Maybe we could get data.camera() and obtain multi-Otsu's thresholds from Fiji's version to try to compare them. Is that what you want?

This is OK. Sorry for missing that earlier.
camera guy is hard to interpret, and relying on Fiji is ok as a development check, though not ideal for testing. Better than nothing though.

Something similar is done in the rank filters.
Due to some corner cases, these kinds of tests often have to be tuned when things are optimized.
So sometimes this isn't ideal, especially with complicated images.


@stefanv's comment is likely good.
> Maybe generate an image, drawing from a mixture of Gaussians, and then see whether you get roughly the means of those classes out under the threshold labels?

Either is OK.

PR #919, opened 2014-03-12
Bradley Thresholding added

mailjenil:
Hello all,

I have implemented Bradley Thresholding algorithm as cited in Adaptive Thresholding Using the Integral Image.(http://people.scs.carleton.ca/~roth/iit-publications-iti/docs/gerh-50002.pdf)

PR #3877, opened 2019-04-30
Adding `threshold_bradley`.

alexandrejaguar:
## Description

Revisiting #919.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -7,7 +7,7 @@ from collections.abc import Iterable
 from ..exposure import histogram
 from .._shared.utils import assert_nD, warn, deprecated
 from ..transform import integral_image
-from ..util import crop, dtype_limits
+from ..util import crop, dtype_limits, img_as_float
 
 
 __all__ = ['try_all_threshold',
@@ -21,6 +21,7 @@ __all__ = ['try_all_threshold',
            'threshold_niblack',
            'threshold_sauvola',
            'threshold_triangle',
+           'threshold_bradley',
            'apply_hysteresis_threshold']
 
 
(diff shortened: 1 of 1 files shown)


alexandrejaguar:
Hi everyone,
first of all, thanks @mailjenil for all your work and code!
The main issue with this function is that it does not follow our API: it returns an image instead of a threshold value. Any ideas on how to adapt that, @jni @hmaarrfk and the others? Thanks!

skimage/filters/thresholding.py
```diff
@@ -1022,3 +1023,68 @@ def apply_hysteresis_threshold(image, low, high):
...
+    mask[0, 0], mask[-1, -1], mask[0, -1], mask[-1, 0] = 1, 1, -1, -1
+
+    # calculating the integral image and convolving with the given mask
+    int_image = integral_image(image)
+    int_image = ndi.convolve(int_image, mask)
+
+    # checking condition for Bradley threshold
+    image = image * block_size ** 2
+    thresh_image = image - int_image * (100 - bright_perc) / 100
+    thresh_image = thresh_image > 0
```
jni: @alexandrejaguar a bit of rearrangement shows that the resulting threshold image should be `integral_image * (1 - q)`, where q is a quantile:

```python
result = (image - integral_image * (1 - q) > 0)
result = (image > integral_image * (1 - q))
```

The lower form is the format we want for applying a threshold, ie normally we would say `image > t` for some threshold t.

However, rearranging also shows that this is equivalent to niblack thresholding with k=0, so I'm not sure this addition needs to happen. Maybe a note on that function is sufficient. Otherwise, the implementation should just call that function, since it is n-dimensional rather than 2D only.

@scikit-image/core any thoughts on this?

jni: Note also that the implementation above has a mistake: the integral image needs to be divided by the window size, otherwise we are looking at the *sum* of the surrounding window, not the *mean* as the docstring suggests.

alexandrejaguar: @jni, @stefanv and I were discussing if this should be implemented. Based on your first comment, maybe it shouldn't. +1 on the note on `niblack` thresholding.

stefanv: Better of we don't have to. Perhaps add a line to niblack stating the equivalence?

jni: Sorry, I spoke too quickly: the `q`/`bright_perc` muddles things up. It's actually equivalent to Niblack *times* (1-q).

alexandrejaguar: I don't see a problem on putting that on niblack's notes, and maybe an example on the docs. Is it okay?

jni: yep!

alexandrejaguar:
Closing this as it feels redundant (a particular case of `threshold_niblack`; see [[1]](https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_niblack_sauvola.html)).

PR #3879, opened 2019-05-01
Fix networkx for v0.14.x and python 2

hmaarrfk:
## Description
Sigh, networkx needs to release things in non-zip format.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,7 +1,8 @@
 numpy>=1.11
 scipy>=0.17.0
 matplotlib>=2.0.0
-networkx>=1.8
+networkx>=1.8; python_version >= '3.5'
+networkx>=1.8,<2.3; python_version < '3.5'
 six>=1.10.0
 pillow>=4.3.0
 PyWavelets>=0.4.0


stefanv:
Probably something @jarrodmillman would like to know about.

jni:
@hmaarrfk can you point to the issue that this is addressing?

hmaarrfk:
i think it is just the `--pre` build in 2.7 that fails due to some strangeness in packages dropping 2.7 this year.

jarrodmillman:
NetworkX dropped support for Python 2 (we did it 8 days after you guys).  I am not sure what issue you are running into with the zip format.  Can you be a little more specific about what problem that is causing?

hmaarrfk:
I don't think it is networkx in particular, or the zip format (though networkx is the first package I encountared that ships via zip)

For example, see this build that is failing with a strange error message
https://travis-ci.org/scikit-image/scikit-image/jobs/526723143#L973

Note that it passes the `--pre` flag to pip during the builds. I think it might be time to drop the `--pre` flag, but the `--pre` flag allows us to anticipate breakage with new versions of numpy/scipy

jarrodmillman:
I just deleted `2.3rc1` and `2.3rc2` from PyPI.  I added 
```
python_requires='>=3.5'
```
in https://github.com/networkx/networkx/commit/dca6362a2bde91a4b228d9f7a58d69b146a2614f
after `2.3rc2`, but before `2.3rc3`.  So now if you are running Python 2.7 and type
```
$ pip install --pre networkx
```
it should still install `2.2`.  Would you mind rerunning the Travis job to see if the problem is resolved?  You shouldn't need to remove the `--pre` flag or add anything to your requirements file.

stefanv:
Thanks, Jarrod!

hmaarrfk:
@jarrodmillman thanks a bunch !!! this really simplifies our pinnings and our tests are now able to get back to green without too many hacks! I never knew you could actually supply pip packages via zip formats. so many things!

PR #3893, opened 2019-05-10
Pin numpydoc to pre-0.8 to fix dev docs formatting

jni:
See https://github.com/numpy/numpydoc/issues/215#issuecomment-487664426
for more.

## Description

Our documentation rendering for the dev branch is broken: the types are all smushed up together with the parameter names, see e.g. https://scikit-image.org/docs/dev/api/skimage.feature.html#skimage.feature.canny

This pin fixes that until https://github.com/numpy/numpydoc/issues/215 is closed.

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,7 +1,7 @@
 # Sphinx 1.7.8 has a bug when building on cached envs
 # https://github.com/sphinx-doc/sphinx/issues/5361
 sphinx>=1.3,!=1.7.8
-numpydoc>=0.6
+numpydoc>=0.7,<0.8
 sphinx-gallery
 sphinx-copybutton
 pytest-runner


hmaarrfk:
Can you add this monitoring to the todo?

Otherwise, the selector

```
!=0.8.0,!=0.9.0,!=0.9.1
```
will probably do.

requirements/docs.txt
```diff
@@ -1,7 +1,7 @@
 # Sphinx 1.7.8 has a bug when building on cached envs
 # https://github.com/sphinx-doc/sphinx/issues/5361
 sphinx>=1.3,!=1.7.8
-numpydoc>=0.6
+numpydoc>=0.7,<0.8
```
hmaarrfk:
```suggestion
numpydoc>=0.7,!=0.8.0,!=0.9.0,!=0.9.1
```

Just a suggestion, my approval holds for both version selectors.

stefanv:
The docs are fixed; thanks a lot!

jni:
@hmaarrfk I've subscribed to the issue, feel free to do the same. Your alternative pin suggestion is good but this way we won't get bitten again if there is another release where this isn't fixed. We don't seem to need any new features or fixes, correct me if I'm wrong.

hmaarrfk:
Yeah. Different philosophy.  Both are ok. I feel like it is more apparent if we get a ping when things change, than things being forced not to change. 

Somebody (who shall not be named) suggested forcibly catching all warnings in the docs, and that has gotten tedious, especially during release phases

stefanv:
Me, me! You'll thank me later (maybe).

Also, numpydoc 0.9 has some nice reference numbering fixes that we probably 
want.

PR #3899, opened 2019-05-11
Add location for Sphinx 2.0.1 search results; clean up templates

stefanv:
In Sphinx 2.0.1, they search for the "role=main" specifier to know where
to put search results.  This is a strange hardcoded choice, but adding
it causes us no difficulties, and should fix some of the issues we've been
seeing with search results not showing up.

### DIFF
--- a/doc/source/themes/scikit-image/layout.html
+++ b/doc/source/themes/scikit-image/layout.html
@@ -1,4 +1,4 @@
-{#
+{#-
     scikit-image/layout.html
     ~~~~~~~~~~~~~~~~~
 
@@ -6,16 +6,14 @@
     Johannes Schönberger.
 
 #}
-
 {%- set url_root = pathto('', 1) %}
-{# XXX necessary? #}
+{#- XXX necessary? #}
 {%- if url_root == '#' %}{% set url_root = '' %}{% endif %}
 {%- if not embedded and docstitle %}
   {%- set titlesuffix = " &mdash; "|safe + docstitle|e %}
 {%- else %}
   {%- set titlesuffix = "" %}
 {%- endif %}
-
 {%- macro script() %}
     <script src="https://code.jquery.com/jquery-latest.js"></script>
     <script src="{{ pathto('_static/', 1) }}js/bootstrap.min.js"></script>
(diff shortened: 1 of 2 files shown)


hmaarrfk:
The download link seems off, it points me to `file:///download.html`

hmaarrfk:
oh man, you guys do some werid stuff with the docs, it seems that not everythign is purely generated by the make docs.

stefanv:
Yes, you're browsing the local build. Not all the links will work there, 
like that one which refers to skimage.org/download

hmaarrfk:
Ok, well it looks good to me then.

stefanv:
What do you mean not generated by make docs?

hmaarrfk:
where is that download page stored for example??? seems like it is not on github pages.

stefanv:
That is part of the main scikit-image website, not the docs. That is hosted in a separate repo (https://github.com/scikit-image/scikit-image-web), but they both use ghpages.

hmaarrfk:
Sounds logical, though I just merged many of my own internal repos into one, because it got too confusing to manage so many.

hmaarrfk:
Is this considered a "non-user facing maintenance PR" seems borderline.

stefanv:
This is urgent: it fixes search on the website for the dev docs.

stefanv:
On May 11, 2019 20:29:40 Mark Harfouche <notifications@github.com> wrote:
> Sounds logical, though I just merged many of my own internal repos into 
> one, because it got too confusing to manage so many.
The main website doesn't live inside the source for scikit-image because it 
a) doesn't belong there and b) is not tied to a specific version of the 
software. This is the simplest separation of concerns, as I see it.

hmaarrfk:
We could pull, though i wouldn't know how to fix anything if it breaks thinsg even more. I waas able to test as much as I could. And its pretty late for me on the east coast.

stefanv:
We can always revert, but I can't see this making things much worse.

hmaarrfk:
nice it works!

jni:
Did this PR cause the extra line spacing in the API reference? Or was it the pinning of numpydoc?

stefanv:
It is the combination. Sphinx now renders that list completely differently.


Please let me know any adjustments you'd like to see.

stefanv:
@jni Can you let me know which changes you observed that you'd like fixed?

stefanv:
Hrm, this renders differently online than on my machine :face_with_head_bandage: 

Sphinx has now made it so hard to theme this correctly.

stefanv:
Maybe revert until I can come up with a proper fix?

stefanv:
:face_with_head_bandage: Looks like I reverted the wrong PR.  This one was necessary to fix the search; the *other* one is the one that upgrades to newer Sphinx + NumpyDoc.

PR #3900, opened 2019-05-11
Fix CSS styling of Sphinx 2.0.1 + numpydoc 0.9 rendered docs

stefanv:
Sphinx 2.0.1 renders parameter listings as description lists instead of
tables.  This CSS, while not perfect, brings us close to what we were
rendering before.

xref: https://github.com/numpy/numpydoc/issues/215

### DIFF
--- a/doc/source/themes/scikit-image/static/css/custom.css
+++ b/doc/source/themes/scikit-image/static/css/custom.css
@@ -1,7 +1,5 @@
 body {
     font-family: "Lato", sans-serif;
-    letter-spacing: 0.02em;
-    font-size: 102%;
     line-height: 23px;
 }
 a {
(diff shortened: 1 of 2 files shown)


hmaarrfk:
image
image

Was this part of the compromise?

jni:
@stefanv can you post screenshots of the result of this PR? The API reference list and a function signature would be good to see. =)

doc/source/themes/scikit-image/static/css/custom.css
```diff
@@ -1,7 +1,5 @@
 body {
     font-family: "Lato", sans-serif;
-    letter-spacing: 0.02em;
-    font-size: 102%;
     line-height: 23px;
```
sciunto:
```suggestion
    line-height: 20px;
```

sciunto: Function signatures look a bit more condensed this way.

stefanv:
The reference rendering should also be double checked.

jni:
don't forget to check regionprops, as those appear to have a different level of nesting and thus faulty styling

sciunto:
References and regionprops look good to me.
Module measure — skimage v0 16 dev0 docs
Module measure — skimage v0 16 dev0 docs(1)

Issue #3890, opened 2019-05-07
skimage.transform.resize: anti_aliasing_sigma default value


etiennelndr:
## Description
I was reading the documentation of the [**skimage.transform.resize**](https://github.com/scikit-image/scikit-image/blob/51d4fd9294955deee48491a5ae4c01c1dfe0f48d/skimage/transform/_warps.py#L19) function when I found that, when `anti_aliasing_sigma` is set to `None`, this parameter should have the default value ["(1 - s) / 2"](https://github.com/scikit-image/scikit-image/blob/51d4fd9294955deee48491a5ae4c01c1dfe0f48d/skimage/transform/_warps.py#L68). But, [line 107](https://github.com/scikit-image/scikit-image/blob/51d4fd9294955deee48491a5ae4c01c1dfe0f48d/skimage/transform/_warps.py#L107) this parameter is set to `np.maximum(0, (factors - 1) / 2)`. Shouldn't it be `np.maximum(0, (1 - factors) / 2)` ?

## Version information
```python
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 03:37:03) [MSC v.1900 64 bit (AMD64)]
Windows-7-6.1.7601-SP1
scikit-image version: 0.14.2
numpy version: 1.14.5
```

This error is also in the current version ([see here](https://github.com/scikit-image/scikit-image/blob/51d4fd9294955deee48491a5ae4c01c1dfe0f48d/skimage/transform/_warps.py#L107)).

hmaarrfk:
He thanks for reporting this apparent discrepancy.

I don't think there is a real difference between

```
(factor - 1) / 2
```
and
```
(1 - factor) /2
```
The reason is that `sigma` is typically considered as the `standard deviation` which is the `square root of the variance`, a positive real number.

I think the documentation should be updated to reflect the implementation, which assumes `factor > 1`, making `(factors - 1) / 2` a positive number.

Would you like to submit the pull request making this change to the documentation?

etiennelndr:
First of all I'm sorry for the late reply. To return to the main subject, I totally agree with you and I'll submit a pull request today.

hmaarrfk:
Fixed in #3911

PR #3911, opened 2019-05-15
DOC: Improve the documentation of transform.resize with respect to the anti_aliasing_sigma parameter

etiennelndr:
## Description
Issue #3890 : fix the documentation of the [`skimage.transform.resize`](https://github.com/scikit-image/scikit-image/blob/51d4fd9294955deee48491a5ae4c01c1dfe0f48d/skimage/transform/_warps.py#L19) method.

## For reviewers
  later.

### DIFF
--- a/skimage/transform/_warps.py
+++ b/skimage/transform/_warps.py
@@ -65,8 +65,8 @@ def resize(image, output_shape, order=1, mode='reflect', cval=0, clip=True,
         avoid aliasing artifacts.
     anti_aliasing_sigma : {float, tuple of floats}, optional
         Standard deviation for Gaussian filtering to avoid aliasing artifacts.
-        By default, this value is chosen as (1 - s) / 2 where s is the
-        down-scaling factor.
+        By default, this value is chosen as (s - 1) / 2 where s is the
+        down-scaling factor, assuming s > 1.
 
     Notes
     -----
--- a/skimage/transform/_warps.py
+++ b/skimage/transform/_warps.py
@@ -66,7 +66,7 @@ def resize(image, output_shape, order=1, mode='reflect', cval=0, clip=True,
     anti_aliasing_sigma : {float, tuple of floats}, optional
         Standard deviation for Gaussian filtering to avoid aliasing artifacts.
         By default, this value is chosen as (s - 1) / 2 where s is the
-        down-scaling factor, assuming s > 1. If s < 1 (up-size), no
+        down-scaling factor, where s > 1. For the up-size case, s < 1, no
         anti-aliasing is performed prior to rescaling.
 
     Notes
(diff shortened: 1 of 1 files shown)


skimage/transform/_warps.py
```diff
@@ -65,8 +65,8 @@ def resize(image, output_shape, order=1, mode='reflect', cval=0, clip=True,
         avoid aliasing artifacts.
     anti_aliasing_sigma : {float, tuple of floats}, optional
         Standard deviation for Gaussian filtering to avoid aliasing artifacts.
-        By default, this value is chosen as (1 - s) / 2 where s is the
-        down-scaling factor.
+        By default, this value is chosen as (s - 1) / 2 where s is the
+        down-scaling factor, assuming s > 1.
```
hmaarrfk: otherwise what value does it take?

etiennelndr: What do you mean ? For the non-default value ?

hmaarrfk: for the default value. so s=1 means no scaling, s>1 means shrinkage, and s<1 means enlargement. What happends when we enlarge photos.

etiennelndr: As far as I know we're not supposed to use the `anti_aliasing` and `anti_aliasing_sigma` parameters when we enlarge images. So `s` should always be superior or equal to 1 but I might be wrong...

hmaarrfk: Maybe an other sentence like 
> If `s < 1` (up-size), no anti-aliasing is performed prior to rescaling.

hmaarrfk: It is more that you now added a condition `s > 1` to the first sentence. So it begs the question 

> what happens when s < 1???

etiennelndr: You're right, I'm adding this sentence right now.

skimage/transform/_warps.py
```diff
@@ -65,8 +65,9 @@ def resize(image, output_shape, order=1, mode='reflect', cval=0, clip=True,
         avoid aliasing artifacts.
     anti_aliasing_sigma : {float, tuple of floats}, optional
         Standard deviation for Gaussian filtering to avoid aliasing artifacts.
-        By default, this value is chosen as (1 - s) / 2 where s is the
-        down-scaling factor.
+        By default, this value is chosen as (s - 1) / 2 where s is the
+        down-scaling factor, assuming s > 1. If s < 1 (up-size), no
```
hmaarrfk:
```suggestion
        down-scaling factor, where s > 1. For the up-size case, s < 1, no
```

ahojnnes:
Looks great, thanks!

etiennelndr:
@ahojnnes You're welcome!

PR #3915, opened 2019-05-17
Fixup test for RANSAC: don't pick duplicate samples #3901

hmaarrfk:
## Description

Seems that residuals needs to be the size of data.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/tests/test_fit.py
+++ b/skimage/measure/tests/test_fit.py
@@ -361,7 +361,7 @@ def test_ransac_sample_duplicates():
             return True
 
         def residuals(self, data):
-            return 1.0
+            return np.ones_like(data, dtype=np.double)
 
     # Create dataset with four unique points. Force 10 iterations
     # and check that there are no duplicated data points.
--- a/skimage/measure/tests/test_fit.py
+++ b/skimage/measure/tests/test_fit.py
@@ -361,7 +361,7 @@ def test_ransac_sample_duplicates():
             return True
 
         def residuals(self, data):
-            return np.ones_like(data, dtype=np.double)
+            return np.ones(len(data), dtype=np.double)
 
     # Create dataset with four unique points. Force 10 iterations
     # and check that there are no duplicated data points.


jni:
Why was it working before?

hmaarrfk:
This totally has to to with a numpy 1.12 only bug. 

```
E               IndexError: in the future, 0-d boolean arrays will be interpreted as a valid boolean index
```

This error was found when I trigged the builds in the wheels, some of which are built with numpy 1.12

https://travis-ci.org/scikit-image/scikit-image-wheels/builds/533603305?utm_source=github_status&utm_medium=notification

I guess this is somewhere where dynamically typed languages fail to help us identify errors at build time.

skimage/measure/tests/test_fit.py
```diff
@@ -361,7 +361,7 @@ def estimate(self, data):
             return True
 
         def residuals(self, data):
-            return 1.0
+            return np.ones_like(data, dtype=np.double)
```
jni: shouldn't this be `np.ones(len(data), dtype=np.double)`?

hmaarrfk:
The test was mostly designed to test for duplicates, but didn't really ensure that residuals was the correct return

These 3 are the offending lines you should look at
https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/fit.py#L832
https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/fit.py#L847
https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/fit.py#L861
This is the line in particular

skimage/measure/tests/test_fit.py
hmaarrfk: ohhh, yeah you are right.

hmaarrfk:
```suggestion
            return np.ones(len(data), dtype=np.double)
```

hmaarrfk:

hmaarrfk:
Merging was done pre-emptively to speed up the release process of 0.14.3 after local testing was made to ensure relevant tests were fixed.

PR #3918, opened 2019-05-19
remove unused dtype argument in IO plugins

sciunto:
## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/io/_plugins/fits_plugin.py
+++ b/skimage/io/_plugins/fits_plugin.py
@@ -14,16 +14,13 @@ except ImportError:
             "for further instructions.")
 
 
-def imread(fname, dtype=None):
+def imread(fname):
     """Load an image from a FITS file.
 
     Parameters
     ----------
     fname : string
         Image file name, e.g. ``test.fits``.
-    dtype : dtype, optional
-        For FITS, this argument is ignored because Stefan is planning on
-        removing the dtype argument from imread anyway.
 
     Returns
     -------
--- a/skimage/io/_plugins/gdal_plugin.py
+++ b/skimage/io/_plugins/gdal_plugin.py
@@ -8,7 +8,7 @@ except ImportError:
                       "for further instructions.")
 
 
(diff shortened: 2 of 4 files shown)


hmaarrfk:
My fear here is that some user uses the dtype argument, and we pass it along, and drop it. but by doing this, they get some error instead of silently ignoring their request.

https://github.com/scikit-image/scikit-image/blob/87830b081516b93dcb2e4a883483b4d895232066/skimage/io/_io.py#L57

This might need a deprecation cycle....

hmaarrfk:
Maybe something like 

```patch
diff --git a/skimage/io/_io.py b/skimage/io/_io.py
index f0b60fb75..fcd8a0103 100644
--- a/skimage/io/_io.py
+++ b/skimage/io/_io.py
@@ -58,6 +58,13 @@ def imread(fname, as_gray=False, plugin=None, flatten=None,
         if fname.lower().endswith(('.tiff', '.tif')):
             plugin = 'tifffile'
 
+    if 'dtype' in plugin_args:
+        warn('The dtype argument was always silently ignored. It will be '
+             'removed from scikit-image version 0.17. To avoid this '
+             'warning, do not specify it in your function call.',
+             UserWarning, stacklevel=2)
+        plugin_args.pop('dtype')
+
     with file_or_url_context(fname) as fname:
         img = call_plugin('imread', fname, plugin=plugin, **plugin_args)
```

sciunto:
You are right. The deprecation cycle sounds reasonable here. At least, I'll keep this PR so that we can point to it in the TODO list to avoid reinventing the wheel that day.

hmaarrfk:
I doubt anybody is going in and digging into the io modules. Aren't they all supposed to be using the "common interface" we provide?

The 3 lines of code I added enable those that use the common interface we provide to get a nice warning. For those that are using internal functions, well they did so at their own peril.

Issue #3889, opened 2019-05-07
canny in 0.15.x cannot deal with all types of image properly


yingfhu:
## Description
`canny` sets percentile threshold if `low_threshold` and `high_threshold` are being set, otherwise, it sets  0.1 * dtype_max. 

Therefore, when inputting a uint8 image with no thresholds for instance, the new `fsmooth` function convert the image to float but the thresholds do not match the data type, so the canny returns all False.

##Version Info
3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0]
Linux-4.18.0-18-generic-x86_64-with-Ubuntu-18.04-bionic
scikit-image version: 0.15.0
numpy version: 1.16.3

hmaarrfk:
Thank you for reporting this @yingfhu 

I think this is very likely a real issue. Would you like to help getting fixed?

I think the general steps would be:
1. Adapt the example to use an image of dtype `uint8` 
https://github.com/scikit-image/scikit-image/blob/master/doc/examples/edges/plot_canny.py
2. Making it into a test case that causes the automated tests to fail.
3. Modify the implementation to ensure that the range is respected in the canny function after smoothing.

We are glad to help you through this process if you want to be involved!

soupault:
Looks like an issue!
We say that input `can be of any dtype.`, but the function is never tested against [u]int images and default `*_threshold` values.

yingfhu:
Sorry for the late response, just finished my finals.
Sure, I would like to fix this issue, but I am not pretty familiar to the steps.
So, from my understandings, what I am going to do would be:
1. Change the dtype of image to `uint8`  in the example of `plot_canny.py`.
2. Add a test case such as `test_image_dtype` in the `test_canny.py`.
3. Modify the `canny.py` to make the test pass.
Are these correct steps? Please tell me if anything is inappropriate! Thanks a lot! @hmaarrfk

hmaarrfk:
@yingfhu sorry for my late reply. 

i guess somebody started to address this
https://github.com/scikit-image/scikit-image/pull/3919

Thanks for reporting this though!

sciunto:
Fixed by #3919

PR #3919, opened 2019-05-19
Canny: same output regardless of dtype

### LINKED ISSUE #3913: `canny` dtype issue
This behavior changed between 0.14 and 0.15.

```
from skimage import data, feature
out = feature.canny(data.coins())
```

`out` is a blank image!  Works fine if you first do `img_as_float`.  Likely has to do with switching to our own implementation of gaussian filtering in canny.

johanjeppsson:
## Description

The same image will produce different outputs depending on the dtype of the input image. This is due to the thresholds, which are not scaled to the dtype of the image, but is compared with the magnitude which is always converted to a float image. This seems correct to me, but I'm not an expert at this, so please let me know if I'm missing some implications of this change.

I added a unit test which asserts that the output should be the same for the same input image regardless if the dtype is uint8 or float. I don't know if there is a better way to test this?

 I think this closes #3913

## For reviewers

  later.

### DIFF
--- a/skimage/feature/tests/test_canny.py
+++ b/skimage/feature/tests/test_canny.py
@@ -104,3 +104,13 @@ class TestCanny(unittest.TestCase):
 
         self.assertRaises(ValueError, F.canny, image, use_quantiles=True,
                           low_threshold=0.5, high_threshold=-100)
+
+    def test_dtype(self):
+        """Check that the same output is produced regardless of image dtype."""
+        image_uint8 = data.camera()
+        image_float = img_as_float(image_uint8)
+
+        result_uint8 = F.canny(image_uint8)
+        result_float = F.canny(image_float)
+
+        assert_equal(result_uint8, result_float)
--- a/skimage/feature/_canny.py
+++ b/skimage/feature/_canny.py
@@ -157,12 +157,12 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
     dtype_max = dtype_limits(image, clip_negative=False)[1]
 
     if low_threshold is None:
-        low_threshold = 0.1 * dtype_max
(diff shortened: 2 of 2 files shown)


johanjeppsson:
I just realized that this looks to be the same thing that @yingfhu discovered in #3889 . Sorry, didn't mean to jump on the same issue!

skimage/feature/tests/test_canny.py
```diff
@@ -104,3 +104,13 @@ def test_invalid_use_quantiles(self):
 
         self.assertRaises(ValueError, F.canny, image, use_quantiles=True,
                           low_threshold=0.5, high_threshold=-100)
+
+    def test_dtype(self):
+        """Check that the same output is produced regardless of image dtype."""
+        image_uint8 = data.camera()[::50, ::50]
```
hmaarrfk: Does it make sense to test with such a small image?

johanjeppsson: No, you're right, It might be better to test with a full image instead. I used this as it was used in another test.

johanjeppsson: I updated the test to use the full image instead.

hmaarrfk: nice. Yeah, I guess for these correctness test it is important to have a meaningful result.

johanjeppsson:
> Looks good to me. Eventually we would want a `preserve_range` parameter, but that an be kept for an other PR. Thanks for your contribution.

Isn't the result always a binary image? Or do you mean to preserve the range for the thresholds?

hmaarrfk:
I think you are right. the result is a binary image so it doesn't matter. I'm just confused today.

hmaarrfk:

PR #3926, opened 2019-05-23
Geometry Transforms: avoid division by zero & some degenerate cases

cv3d:
## Description
Minor update to avoid returning nans due to division by zeros & some degenerate cases

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -549,9 +549,12 @@ class ProjectiveTransform(GeometricTransform):
         src = np.vstack((x, y, np.ones_like(x)))
         dst = src.T @ matrix.T
 
+        # below, we will divide by the last dimension of the homogeneous
+        # coordinate matrix. In order to avoid division by zero,
+        # we replace exact zeros in this column with a very small number.
+        dst[dst[:, 2] == 0, 2] = np.finfo(float).eps
         # rescale to homogeneous coordinates
-        dst[:, 0] /= dst[:, 2]
-        dst[:, 1] /= dst[:, 2]
+        dst[:, :2] /= dst[:, 2:3]
 
         return dst[:, :2]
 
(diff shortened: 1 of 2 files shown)


stefanv:
Hey, @cv3d, thank you for submitting this PR.

Could you please expand on the problem you ran into, and why this is the correct fix?

cv3d:
Hi @stefanv. I am using these geometry transforms in a RANSAC iteration for model estimation, where some degenerative cases might be the cause of such errors.

The basic idea of this PR is to return some numerical values (even if they are very large) rather than NaNs, which might break the user code.

cv3d:
@stefanv The failing test is NOT related to this PR. Shall I do anything?

jni:
@cv3d I think we would like some comments around those lines to explain their purpose, as well as a test case that would fail without the fixes (so that if anyone changes the logic or the algorithm in the future, the new algorithm still has the same stability).

cv3d:
@jni How about now?

skimage/transform/_geometric.py
```diff
@@ -549,9 +549,10 @@ def _apply_mat(self, coords, matrix):
         src = np.vstack((x, y, np.ones_like(x)))
         dst = src.T @ matrix.T
 
+        # avoid division by zero
+        dst[dst[:, 2] == 0, 2] = np.finfo(float).eps
         # rescale to homogeneous coordinates
-        dst[:, 0] /= dst[:, 2]
-        dst[:, 1] /= dst[:, 2]
+        dst[:, :2] /= dst[:, 2:3]
```
emmanuelle: why not ``dst[:, :2] /= dst[:, 2]``? Broadcasting should work here I think?

cv3d: @emmanuelle Not really. `dst[:, :2]` dimension is `(n, 2)`, `dst[:, 2:3]` has a dimension of `(n,1)`, while `dst[:, 2]` dimension is just `n`.

cv3d:
@jni All done!

skimage/transform/_geometric.py
```diff
@@ -549,9 +549,12 @@ def _apply_mat(self, coords, matrix):
         src = np.vstack((x, y, np.ones_like(x)))
         dst = src.T @ matrix.T
 
+        # below, we will divide by the last dimension of the homogeneous
+        # coordinate matrix. In order to avoid division by zero,
+        # we replace exact zeros in this column with a very small number.
```
jni: :clap: good stuff!

jni:
@cv3d thanks for the update! Looks great!

@scikit-image/core imho this is merge-ready!

stefanv:
Thanks, @cv3d!

PR #3928, opened 2019-05-24
ImageCollection is able to receive lists of patterns

alexdesiqueira:
## Description

Users can pass a list of patterns to ImageCollection, instead of a string.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


alexandrejaguar:
OK, I'll consider that the implemented tests think on `load_pattern` as being images themselves, already.

skimage/io/collection.py
```diff
@@ -104,6 +98,9 @@ class ImageCollection(object):
 
     Notes
     -----
+    Note that files are always stored in alphabetical order. Also note that
```
jni: This is false, it's *alphanumeric* order, `im1.jpg` will come before `im10.jpg`. My second contribution ever to scikit-image. =D

skimage/io/collection.py
```diff
@@ -147,21 +144,22 @@ def imread_convert(f):
...
-
+    >>> ic = io.ImageCollection(['/tmp/work/*.png', '/tmp/other/*.jpg'])
     """
 
     def __init__(self, load_pattern, conserve_memory=True, load_func=None,
                  **load_func_kwargs):
         """Load and manage a collection of images."""
-        if isinstance(load_pattern, str):
+        try:
+            if not isinstance(load_pattern, str):
```
jni: I think this check should be more stringent: it should not be a string, but it should be a sequence.

Also, `try/except` clauses should contain as little content as possible. I think this one encompasses too many statements.

skimage/io/collection.py
```diff
@@ -73,19 +73,13 @@ def alphanumeric_key(s):
...
-    slicing returns a new ImageCollection, *not* a view into the data.
-
     Parameters
     ----------
     load_pattern : str or list
         Pattern glob or filenames to load. The path can be absolute or
-        relative.  Multiple patterns should be separated by os.pathsep,
-        e.g. '/tmp/work/*.png:/tmp/other/*.jpg'.  Also see
-        implementation notes below.
+        relative. Also see implementation notes below.
```
jni: I think we should include all possible inputs here, including the pathsep. As far as I know we don't intend to deprecate that? We could keep the original message but include the statement about windows vs *nix, and also the option to pass in a list of patterns.

alexandrejaguar: They still work; I tried to state that in the documentation. Is it good?

skimage/io/collection.py
alexandrejaguar: Tried to fix that with:
```
if isinstance(load_pattern, (list, tuple)):
```
IMHO this `try/except` is as good as it gets. It receives the three types of inputs we'd like to put there, and it's rather efficient when compared to an `if isinstance(load_pattern, (str, list, tuple)`. That could also bring us issues in a more difficult type of input:
```
load_pattern = [(1, 'one'), (2, 'two')]
In []: isinstance(, (list, tuple))                     
Out[]: True
```

skimage/io/collection.py
```diff
@@ -104,6 +98,13 @@ class ImageCollection(object):
 
     Notes
     -----
+    Note that files are always stored in alphanumerical order. Also note
+    that slicing returns a new ImageCollection, *not* a view into the data.
+
+    If multiple patterns are entered as a single string, they should be
+    separated by os.pathsep, e.g. '/tmp/work/*.png:/tmp/other/*.jpg' in *nix,
+    or '/tmp/work/*.png;/tmp/other/*.jpg' in Windows.
```
jni: very nice!

jni:
@alexandrejaguar the problem with the TypeError syntax is that there's lots of statements there, any of which could potentially raise a TypeError that is not actually the error we're looking for.

There is also a design philosophy that I mostly subscribe to that "exceptions are for exceptional circumstances", and you shouldn't use them for "normal" and "expected" logic flow. 

My suggestion: create a helper function, `is_multipattern`, that checks the following condition:

```python
import typing

((isinstance(load_pattern, str) and os.pathsep in load_pattern) or
 (not isinstance(load_pattern, str) and
  isinstance(load_pattern, typing.Sequence) and
  all(isinstance(pattern, str) for pattern in load_pattern))
)
```

Then: 

```python
if is_multipattern(load_pattern):
    if isinstance(load_pattern, str):
        load_pattern = load_pattern.split(os.pathsep)
    ...
elif isinstance(load_pattern, str):
    ...  # single pattern logic
else:
    # Raise TypeError
```

This might also be a good opportunity for us to start using progressive typing:

```python
from typing import Sequence, Union as U

def imread_collection(load_pattern : U[str, Sequence[str]], ...):
```

Then people's IDEs, at least, will tell them that [(1, 'one')] is invalid input for the function.

alexandrejaguar:
@jni sounds good, let's see what I got.

stefanv:
On Fri, 24 May 2019 18:12:33 -0700, Juan Nunez-Iglesias wrote:
> >      Parameters
>      ----------
>      load_pattern : str or list
>          Pattern glob or filenames to load. The path can be absolute or
> -        relative.  Multiple patterns should be separated by os.pathsep,
> -        e.g. '/tmp/work/*.png:/tmp/other/*.jpg'.  Also see
> -        implementation notes below.
> +        relative. Also see implementation notes below.
> 
> I think we should include all possible inputs here, including the
> pathsep. As far as I know we don't intend to deprecate that? We could
> keep the original message but include the statement about windows vs
> *nix, and also the option to pass in a list of patterns.

This is a tricky point: on the one hand you want the docstring to give
an accurate description of the function, but on the other you want it to
encourage best behavior.

It may be best to be honest: "For backward compatibility, we also accept
a string of patterns delimited by `os.pathsep`."

jni:
> It may be best to be honest: "For backward compatibility, we also accept a string of patterns delimited by `os.pathsep`."

I agree with this wording, thanks @stefanv! I would add only "... delimited by ``os.pathsep`` (which is ":" on Mac and Linux platforms, and ";" on Windows)."

I think it's important to be clear on this in case someone is trying to figure out what's wrong with the code that was written on one platform and is now broken on another.

alexandrejaguar:
> > It may be best to be honest: "For backward compatibility, we also accept a string of patterns delimited by `os.pathsep`."
> 
> I agree with this wording, thanks @stefanv! I would add only "... delimited by `os.pathsep` (which is ":" on Mac and Linux platforms, and ";" on Windows)."
> 
> I think it's important to be clear on this in case someone is trying to figure out what's wrong with the code that was written on one platform and is now broken on another.

I'm more into the wordy version as well, with a little modification:  `"... delimited by `os.pathsep` (";" on Windows, ":" on Mac and Linux)."`

jni:
@alexandrejaguar nice job making it more concise! I'm all for it!

alexandrejaguar:
> Then people's IDEs, at least, will tell them that [(1, 'one')] is invalid input for the function.

@jni I understand the concern with `TypeError`, and thanks for your example! Let's be clear then: we'll not accept entries as `[(1, 'one'), (2, 'two')]` in `ImageCollection()` anymore?
If not, we'll have to remove one of the tests as well.

jni:
> Let's be clear then: we'll not accept entries as [(1, 'one'), (2, 'two')] in ImageCollection() anymore?

Wait, what? We accept those? What does it mean?

alexandrejaguar:
There's a test that check for input like that, maybe a list of previously input images... I was trying to deal with that actually =DYep, maybe it's time to stop considering that...On May 27, 2019 5:30 AM, Juan Nunez-Iglesias <notifications@github.com> wrote:
Let's be clear then: we'll not accept entries as [(1, 'one'), (2, 'two')] in ImageCollection() anymore?

Wait, what? We accept those? What does it mean?

stefanv:
If it is supported API *with a test* I would be very hesitant to remove it without further investigation.  Was this perhaps to support "frame generators" that output frame number + image file combos?  Or perhaps a way to provide explicit ordering?  Is it used internally that way, without ever having been externally advertised?  In that case, we could potentially just change it.

alexandrejaguar:
> If it is supported API with a test I would be very hesitant to remove it without further investigation.

@stefanv @jni that's what I'm afraid of. This is the test:

```
================================== FAILURES ===================================
____________________ TestImageCollection.test_custom_load _____________________
self = <skimage.io.tests.test_collection.TestImageCollection testMethod=test_custom_load>
    def test_custom_load(self):
        load_pattern = [(1, 'one'), (2, 'two')]
    
        def load_fn(x):
            return x
    
>       ic = ImageCollection(load_pattern, load_func=load_fn)
python35-x64\lib\site-packages\skimage\io\tests\test_collection.py:83: 
self = <skimage.io.collection.ImageCollection object at 0x000000B388F8FF60>
load_pattern = [(1, 'one'), (2, 'two')], conserve_memory = True
load_func = <function TestImageCollection.test_custom_load.<locals>.load_fn at 0x000000B3885900D0>
load_func_kwargs = {}
    def __init__(self, load_pattern, conserve_memory=True, load_func=None,
                 **load_func_kwargs):
        """Load and manage a collection of images."""
        self._files = []
        if _is_multipattern(load_pattern):
            if isinstance(load_pattern, str):
                load_pattern = load_pattern.split(os.pathsep)
            for pattern in load_pattern:
                self._files.extend(glob(pattern))
            self._files = sorted(self._files, key=alphanumeric_key)
            self._numframes = self._find_images()
        elif isinstance(load_pattern, str):
            self._files.extend(glob(load_pattern))
            self._files = sorted(self._files, key=alphanumeric_key)
            self._numframes = self._find_images()
        else:
>           raise TypeError('Invalid pattern as input.')
E           TypeError: Invalid pattern as input.
python35-x64\lib\site-packages\skimage\io\collection.py:184: TypeError
```

>  Is it used internally that way, without ever having been externally advertised?

I'll ckeck it. :mag:

alexandrejaguar:
Got it (`skimage/io/tests/test_fits.py`):

```
@testing.skipif(not pyfits_available, reason="pyfits not installed")
def test_imread_collection_single_MEF():
    io.use_plugin('fits')
    testfile = os.path.join(data_dir, 'multi.fits')
    ic1 = io.imread_collection(testfile)
    ic2 = io.ImageCollection(
        [(testfile, 1), (testfile, 2), (testfile, 3)],
        load_func=fplug.FITSFactory)
    assert _same_ImageCollection(ic1, ic2)


@testing.skipif(not pyfits_available, reason="pyfits not installed")
def test_imread_collection_MEF_and_simple():
    io.use_plugin('fits')
    testfile1 = os.path.join(data_dir, 'multi.fits')
    testfile2 = os.path.join(data_dir, 'simple.fits')
    ic1 = io.imread_collection([testfile1, testfile2])
    ic2 = io.ImageCollection([(testfile1, 1), (testfile1, 2),
                             (testfile1, 3), (testfile2, 0)],
                             load_func=fplug.FITSFactory)
    assert _same_ImageCollection(ic1, ic2)
```
OK, it seems that this was implemented only for FITS support.

stefanv:
Here is the relevant but extremely cryptic section of the docstring:

> For files with multiple images, the images will be flattened into a list and added to the list of available images.  In this case, ``load_func`` should accept the keyword argument ``img_num``.

How does the image loader know that there are multiple frames inside of the image?   Hrmmm... I feel like this functionality may need to be removed: too complex and confusing.

skimage/io/collection.py
```diff
@@ -147,24 +160,28 @@ def imread_convert(f):
     >>> coll[0].shape
     (200, 200)
 
-    >>> ic = io.ImageCollection('/tmp/work/*.png:/tmp/other/*.jpg')
+    >>> ic = io.ImageCollection(['/tmp/work/*.png', '/tmp/other/*.jpg'])
```
hmaarrfk: IMO, this should be the only supported way to setup multiple patterns in Scikit-Image 0.18.

To many options makes maintaining this kind of code super complicated.

skimage/io/collection.py
```diff
@@ -72,20 +73,25 @@ def alphanumeric_key(s):
     return k
 
 
-class ImageCollection(object):
+def _is_multipattern(load_pattern):
+    """Helping function. Returns True if load_pattern contains a tuple, list,
+    or a string separated with os.pathsep."""
+    pattern = ((isinstance(load_pattern, str) and os.pathsep in load_pattern)
+               or (not isinstance(load_pattern, str) and
+                   isinstance(load_pattern, typing.Sequence) and
```
hmaarrfk: I odn't think you should be checking against `typing.Sequence` rather `collections.abc.Sequence` https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence

alexandrejaguar: @jni can't remember why you suggested that, sorry. Could you back me up here? :smile:

skimage/io/collection.py
hmaarrfk: To do this, you would have to add a deprecation warning. In that case, I see this PR more as a "simplification" of the API, since lists of strings are much more intuitive for python users than a strange seperator. I would approve ;)

skimage/io/collection.py
jni: @alexandrejaguar probably ignorance. =P I think @hmaarrfk is right here. :joy:

alexandrejaguar: @hmaarrfk `typing.Sequence` says it's a generic version of `collections.abc.Sequence`. `typing.Sequence`:
```
This represents a generic version of type 'origin' with type arguments 'params'.
There are two kind of these aliases: user defined and special. The special ones
are wrappers around builtin collections and ABCs in collections.abc. These must
have 'name' always set. If 'inst' is False, then the alias can't be instantiated,
this is used by e.g. typing.List and typing.Dict.
```
I guess both of you are right, but I'll call `collections.abc.Sequence` anyway.

hmaarrfk: Interesting. 

I guess after 6 months of trying to use the typing module I still can't really use it well. I'm going to wait until it settles more before bringing it into my personal work.

alexandrejaguar: Yeah, maybe we should stick to `collections.abc` right now.

skimage/io/collection.py
```diff
@@ -16,18 +17,18 @@
            'imread_collection_wrapper']
 
 
-def concatenate_images(ic):
+def concatenate_images(image_collection):
```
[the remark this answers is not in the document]
alexdesiqueira: :grimacing:

skimage/io/collection.py
```diff
@@ -37,13 +38,18 @@ def concatenate_images(ic):
...
+    Notes
+    -----
+    ``concatenate_images`` receives any iterable object containing images,
+    including ImageCollection and MultiImage, and returns a NumPy array.
     """
-    all_images = [img[np.newaxis, ...] for img in ic]
+    all_images = [image[np.newaxis, ...] for image in ic]
     try:
-        ar = np.concatenate(all_images)
+        array_cat = np.concatenate(all_images)
```
hmaarrfk: why is stack not used?

```python
In [7]: np.stack([a, b])                                                                                 
Out[7]: 
array([[[0., 0., 0.],
        [0., 0., 0.],
        [0., 0., 0.]],

       [[1., 1., 1.],
        [1., 1., 1.],
        [1., 1., 1.]]])

In [8]: np.stack([a, b]).shape                                                                           
Out[8]: (2, 3, 3)
```

hmaarrfk:
I'm really the wrong person to ask about this PR. Personally, I don't think it should exist. imageio provides enough basic functionality. The "always sorted order" basically throws away a ton of metadata from the user. What if the user had images numbered zero through 10, but had 5 loaded images. They would never know which 5 were loaded using this function.

This function is "cute" for the examples you showed, but doesn't really extend well to any real world example I can think of where metadata is just as important for an image collection than the binary numbers themselves.

Because of the above, I don't think you will ever get a thumbs up from me. :( You shouldn't count me in as the 1/2 reviewers needed to merge this in.

That said, if you do want to push through with this, I strongly encourage you to deprecate the `os.pathsep` feature here, because:
1. I feel it is really unpythonic, adding a special character to a string in order to later turn it into a list seems like a `bashism`.
2. It puts undue complexity indefinitely in the scikit-image library.

For now, you still have a thumbs down since it seems to miss a test for the new functionality you added.

skimage/io/collection.py
alexdesiqueira: This really was only refactoring. I didn't think about it.

alexdesiqueira:
> I'm really the wrong person to ask about this PR. Personally, I don't think it should exist.
> (...)
> Because of the above, I don't think you will ever get a thumbs up from me. :( You shouldn't count me in as the 1/2 reviewers needed to merge this in.

What happened with this @hmaarrfk from-the-past? :smile:

> IMO, this should be the only supported way to setup multiple patterns in Scikit-Image 0.18.
> (...)
>  I would approve ;)

I'll take care of the other points.

hmaarrfk:
> What happened with this @hmaarrfk from-the-past? 

Maybe an other way to think about it is:

Loading and saving data imposes a certain permanent organization on experimental setups.

Since you don't provide a "save collection", I just don't see how this function helps anybody.

1. Acquisition
2. Saving and loading
3. Analysis
4. Visualization

Probably best to leave scikit image focused on the later two.

Whoever deals with acquisition can deal with saving and loading.

jni:
@hmaarrfk none of your points provide to me compelling evidence for deprecating working functionality. I'm perfectly ok with soft-deprecating it (removing any mention on it from the docs), but I think the functionality should remain, rather than cause unnecessary churn for downstream users.

hmaarrfk:
ok, dont't deprecate. A test definitely needs to be added.

stefanv:
@hmaarrfk It is true that we have a policy of not implementing one-liner's on the user's behalf.  But ImageCollection is quite useful, unless you want to introduce a heavier dependency such as dask for managing delayed reading.

Use case: I have a directory with 2000 images in it (this happens surprisingly frequently).  I want to load these up, take a look at some of them.  Yes, I can go pick out filenames individually, but often it's handy to just load the whole bunch into a collection and to then grab items from that list-like object.

So, I'd be -1 on deprecating the ImageCollection, unless we provided explicit directions in the user documentation on how to do this with, e.g., dask.

hmaarrfk:
I'm not saying deprecating all of image collection. Scikit image is many voices and you should all push forward with this PR. I'm simply suggesting deprecating the behavior that uses `split(os.pathsep)`.

alexdesiqueira:
@hmaarrfk:

> Analysis
> Visualization
> Probably best to leave scikit image focused on the later two.

I'm humbly against you on that. In my mind we're supposed to be (or, at least, offer) a nice solution  from the start (which is, of course, reading the image). I'm even against our policy that @stefanv brought up:

> It is true that we have a policy of not implementing one-liner's on the user's behalf.

However, as you said, we're many voices.

skimage/io/collection.py
jni: stack did not exist when this code was written.

jni:
I actually do agree with @hmaarrfk that we should "divest" skimage from IO things, and further, also visualisation things. But this is a very long term goal and should not affect us improving existing functionality and fixing bugs. I would just be very reluctant to *add* any IO functionality.

skimage/io/collection.py
```diff
@@ -104,6 +121,13 @@ class ImageCollection(object):
 
     Notes
     -----
+    Note that files are always stored in alphanumerical order. Also note
```
jni:
```suggestion
    Note that files are always returned in alphanumerical order. Also note
```

skimage/io/collection.py
jni: As mentioned in the comments, let's just soft-deprecate the old behaviour. This means removing the documentation for the old syntax, including the second paragraph above. @stefanv do you agree with this approach?

skimage/io/tests/test_fits.py
```diff
@@ -23,44 +23,10 @@ def teardown():
...
-                             load_func=fplug.FITSFactory)
-    assert _same_ImageCollection(ic1, ic2)
-
-
 def _same_ImageCollection(collection1, collection2):
-    """Ancillary function to compare two ImageCollection objects, checking
-       that their constituent arrays are equal.
+    """
+    Ancillary function to compare two ImageCollection objects, checking that
+    their constituent arrays are equal.
```
jni: If the FITS plugin is now broken, this should *definitely* be noted in the release notes.

jni:
@alexdesiqueira LOL, timing! :wave:

alexdesiqueira:
@jni I'm sprinting!

skimage/io/tests/test_fits.py
alexdesiqueira: Indeed. I'll check

alexdesiqueira:
> I actually do agree with @hmaarrfk that we should "divest" skimage from IO things, and further, also visualisation things. But this is a very long term goal and should not affect us improving existing functionality and fixing bugs. I would just be very reluctant to add any IO functionality.

I feel I should again express my concern of we _taking_ things away from skimage. At some point we will restrict the package to a bunch of hard-coding people, instead of all nice people around that want something somewhat easier to use and to get things done with (myself kinda included here).
But it's only my 2 BRL cents.

jni:
> I feel I should again express my concern of we taking things away from skimage. At some point we will restrict the package to a bunch of hard-coding people,

no no: the idea is not to stop supporting ease of use, but to separate concerns into different packages. That way, people who need only an image processing function don't need to download matplotlib, imageio, freetype, and a bunch of other things. scikit-image would retain its ease of use through its easy interoperability with the ecosystem.

Either way, I'm not advocating removing functionality, at least not for a long time. I'm just aiming to be careful to not add a bunch of extra functionality that is not precisely aligned with our vision.

skimage/io/collection.py
```diff
@@ -72,20 +78,31 @@ def alphanumeric_key(s):
...
 
-    Note that files are always stored in alphabetical order. Also note that
-    slicing returns a new ImageCollection, *not* a view into the data.
+
+class ImageCollection(object):
+    """Load and manage a collection of image files.
 
     Parameters
     ----------
     load_pattern : str or list
```
stefanv: This description is not complete.

alexdesiqueira: I thought a little and I think you're right. Let's maintain the old behavior but without the documentation, to not break stuff for a while but not encouraging it.

skimage/io/collection.py
```diff
@@ -72,38 +78,52 @@ def alphanumeric_key(s):
...
     Parameters
     ----------
-    load_pattern : str or list
-        Pattern glob or filenames to load. The path can be absolute or
-        relative.  Multiple patterns should be separated by os.pathsep,
-        e.g. '/tmp/work/*.png:/tmp/other/*.jpg'.  Also see
-        implementation notes below.
+    load_pattern : str or list of str
+        Pattern string or list of strings to load. The filename path can be
+        absolute or relative.
```
jni: @alexdesiqueira based on my reading of the notes from the last meeting, this description should include more detail, then we can merge!

jni: I'll point out that my approval still stands, and @stefanv could merge when ready. =)

hmaarrfk:
Curious to learn how people use this! Keep me updated!

PR #3515, opened 2018-10-27
Extension of Frangi filter to 3D data and other ridge detection filters

DavidBreuer:
## Description

Duplicate/correction of PR #2762 in which I encountered rebase problems.

Correct and extend available 2D Frangi (vesselness) filter [1] to 3D data. Together with the Hybrid Hessian filter, move it to a ridge detection module with two new, additional filters. The new Meijering (neuriteness) filter [2] is applicable to nD data and the new Sato (tubeness) filter [3] is applicable to
2D and 3D data. Both are similar to the Frangi filter but rely on different combinations of eigenvalues of the Hessian matrix to detect ridge structures.

Implement the following changes:

- Correct available 2D Frangi filter by using Hessian eigenvalues which are sorted by their absolute values (not by value), as described in [1].
- Change keyword arguments 'beta1' and 'beta2' in the available Frangi and Hessian filters to 'beta' and 'gamma' for better agreement with the original description in [1]. Add deprecation warnings for the old keywords.
- Correct effect of keyword argument 'black_ridges' in the available 2D Frangi filter, as noted in issue #2700. As proposed in the comments, solve this issue by inverting the image to detect light ridges on dark backgrounds.
- Complete partial references for the available 2D Frangi and Hybrid Hessian filters.
- Remove third, erroneous reference in the doc string of the available 2D Frangi filter concerning the Gabor filter.
- Move the available 2D Frangi and Hybrid Hessian filters to a ridge detection module, similar to that for edge detection.
- Extend Frangi (vesselness) filter to 3D data as described in [1].
- Add Meijering (neuriteness) filter for nD data [2].
- Add Sato (tubeness) filter for 2D and 3D data [3].
- Extend available test suite (test_frangi.py) by tests for the new filters as well as 3D data (test_ridges.py).
- Extend available gallery example (plot_frangi.py) by the new filters as well as a comparison of the detection of dark and light ridges, respectively (plot_ridge_filter.py).
- Add new identity filter which accepts arbitrary keywords and returns the input image. This is very convenient for automated plotting of different filters and comparison to the original image, see gallery example.

Additional note for future contributions:

- The available Hybrid Hessian filter does not agree with its description in [4]. In particular, it should not simply be a thresholded version of the Frangi filter but involves different smoothing before the computation of the Hessian eigenvalues (cf. equations (2)-(6) in [4]). Moreover, the implemented thresholding does not agree with the described one (cf. equation (15) in [4]).

Change files:

[...]

PR #3933, opened 2019-05-25
ENH: doc of ridge functions

sciunto:
## Description

This PR aims at improving the doc of #3515   as these functions are designed for processing on vessels, I changed the image for a crop of retina. I added see also sections as well.
Figure_1

Cc @DavidBreuer 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/edges/plot_ridge_filter.py
+++ b/doc/examples/edges/plot_ridge_filter.py
@@ -40,7 +40,8 @@ References
        :DOI:`10.1007/978-3-319-16811-1_40`
 """
 
-from skimage.data import page
+from skimage import data
+from skimage import color
 from skimage.filters import meijering, sato, frangi, hessian
 import matplotlib.pyplot as plt
 
@@ -50,7 +51,7 @@ def identity(image, **kwargs):
     return image
 
 
-image = page()
+image = color.rgb2gray(data.retina())[300:700, 700:900]
 cmap = plt.cm.gray
 
 kwargs = {}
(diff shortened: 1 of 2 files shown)


emmanuelle:
Thanks @sciunto ! The contrast of frangi is flattened because the higher values are at the border of the image, could you please increase the contrast (eg setting vmax) for this filter, since the result is quite good but it doesn't show here.

emmanuelle:
We could also have more explanations about the differences between the different filters, but this can come in another PR because this PR is already an improvement.

sciunto:
Do you think it is better to set a vmax for a particular case or crop all images of 4 pixels on each edge and add a note saying that Frangi leads to an edge effect removed in this example?

PS: I'm discovering these filters and I'm not familiar at all with these algorithms.

jni:
I'm confused — doesn't matplotlib automatically stretch the contrast for each image?

jni:
Oh, sorry, misread @emmanuelle's comment. Hmmm. I'm not sure what the best answer is here. @sciunto go with whatever you think is prettiest I reckon. =)

emmanuelle:
@sciunto I would stretch the contrast for this particular case and explain in the text that Frangi filter does this special thing at the border (which will be visible because these pixels will be saturated). But I also leave it to your personal taste, the important thing is that this subplot is not pitch black and users think frangi is worthless.

sciunto:
I forgot about this PR. I put a crop to improve the rendering. 
sphx_glr_plot_ridge_filter_001

jni:
Looks great, thanks @sciunto!!!

PR #3935, opened 2019-05-27
Adding complexity and new tests for filters.threshold_multiotsu

alexandrejaguar:
## Description

Adding complexity and possible new tests for `filters.threshold_multiotsu`, as discussed in #3872.

## Checklist

~~- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)~~
~~- [ ] Gallery example in `./doc/examples` (new features only)~~
~~- [ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark~~
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/tests/test_thresholding.py
+++ b/skimage/filters/tests/test_thresholding.py
@@ -3,6 +3,7 @@ import numpy as np
 from scipy import ndimage as ndi
 import skimage
 from skimage import data
+from skimage.draw import circle
 from skimage._shared._warnings import expected_warnings
 from skimage.filters.thresholding import (threshold_local,
                                           threshold_otsu,
@@ -519,6 +520,17 @@ def test_check_multiotsu_results():
         thr_multi = threshold_multiotsu(image,
                                         classes=idx)
         assert len(thr_multi) == idx-1
+
+
+def test_multiotsu_output():
+    image = np.zeros((100, 100), dtype='int')
+    coords = [(25, 25), (50, 50), (75, 75)]
+    values = [64, 128, 192]
+    for coor, val in zip(coords, values):
+        rr, cc = circle(coor[1], coor[0], 20)
+        image[rr, cc] = val
+    thresholds = [64, 128]
+    assert np.array_equal(thresholds, threshold_multiotsu(image))
 
 
 @pytest.mark.parametrize("thresholding, lower, upper", [
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1051,7 +1051,8 @@ def threshold_multiotsu(image, classes=3, nbins=256):
         Number of classes to be thresholded, i.e. the number of resulting
         regions.
     nbins : int, optional
-        Number of bins used to calculate the histogram.
+        Number of bins used to calculate the histogram. This value is ignored
+        for integer arrays.
 
     Returns
     -------
@@ -1093,6 +1094,8 @@ def threshold_multiotsu(image, classes=3, nbins=256):
                                   nbins=nbins,
                                   source_range='image')
     prob = hist / image.size
+    # histogram ignores nbins for integer arrays.
+    nbins = len(bin_centers)
 
     # defining arrays to store the zeroth (momP, cumulative probability)
     # and first (momS, mean) moments, and the variance between classes
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1064,6 +1064,17 @@ def threshold_multiotsu(image, classes=3, nbins=256):
     between the desired classes. Based on the Multi-Otsu approach by
     Liao, Chen and Chung.
 
+    This implementation relies on a Cython function whose complexity
+    is O(combinations * variance_calc), and depends on the number of
+    bins and classes:
+
+    >>> import math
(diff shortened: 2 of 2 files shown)


skimage/filters/thresholding.py
```diff
@@ -1064,6 +1064,17 @@ def threshold_multiotsu(image, classes=3, nbins=256):
...
+    is O(combinations * variance_calc), and depends on the number of
+    bins and classes:
+
+    >>> import math
+    >>> combinations = nbins**(classes-1)/math.factorial(classes-1)
+    >>> variance_calc = classes + 1
+
+    Then, the required calculation time grows exponentially according
+    to the number of classes desired.
+
```
jni: I think this should be straight-up math notation rather than Python. I suggest

```
whose complexity is :math:`O\left(\frac{Ch^{C-1}}{(C-1)!}\right)` where :math:`h` is
the number of histogram bins and :math:`C` is the number of classes desired.
```

(If I remember my complexity analysis correctly, the +1 can be safely left out of the expression.)

jni: You might need to double-check my sphinx-math expression. =)

alexandrejaguar:
Awesome! I'll do that. (Never learned complexity analysis :laugh:)On May 27, 2019 5:58 PM, Juan Nunez-Iglesias <notifications@github.com> wrote:@jni commented on this pull request.



In skimage/filters/thresholding.py:
> @@ -1064,6 +1064,17 @@ def threshold_multiotsu(image, classes=3, nbins=256):
     between the desired classes. Based on the Multi-Otsu approach by
     Liao, Chen and Chung.
 
+    This implementation relies on a Cython function whose complexity
+    is O(combinations * variance_calc), and depends on the number of
+    bins and classes:
+
+    >>> import math
+    >>> combinations = nbins**(classes-1)/math.factorial(classes-1)
+    >>> variance_calc = classes + 1
+
+    Then, the required calculation time grows exponentially according
+    to the number of classes desired.
+

I think this should be straight-up math notation rather than Python. I suggest
whose complexity is :math:`O\left(\frac{Ch^{C-1}}{(C-1)!}\right)` where :math:`h` is
the number of histogram bins and :math:`C` is the number of classes desired.

(If I remember my complexity analysis correctly, the +1 can be safely left out of the expression.)

skimage/filters/thresholding.py
alexandrejaguar: Awesome! I'll do that. (never learned complexity analysis :laugh:)

jni:
Ah, thanks PEP8speaks! @alexandrejaguar I forgot: to include the latex, you need to prepend `r` to the docstring, as in `r"""Return ...`, so that the backslashes are not incorrectly interpreted.

alexandrejaguar:
Thanks @jni! Didn't see that coming :smile:

alexandrejaguar:
Alright, @jni, I think the complexity part is now fine. Now, let's solve the tests issues @hmaarrfk pointed.
Mark, since the ideas given before weren't that great, what do you have in mind?

hmaarrfk:
I liked stefan's idea of drawing values from a few distributions and using the filter to find the centroid of those distributions.

Even simpler, you could set 3 different values with draw circle
https://scikit-image.org/docs/dev/api/skimage.draw.html#skimage.draw.circle

and ensure that you get the 2 thresholds you need.

alexandrejaguar:
@hmaarrfk you got it.

skimage/filters/thresholding.py
```diff
@@ -1088,6 +1094,8 @@ def threshold_multiotsu(image, classes=3, nbins=256):
                                   nbins=nbins,
                                   source_range='image')
     prob = hist / image.size
+    # histogram ignores nbins for integer arrays.
+    nbins = len(bin_centers)
```
hmaarrfk: why?

skimage/filters/tests/test_thresholding.py
```diff
@@ -521,6 +522,17 @@ def test_check_multiotsu_results():
...
 
+def test_multiotsu_output():
+    image = np.zeros((100, 100), dtype='int')
+    coords = [(25, 25), (50, 50), (75, 75)]
+    values = [64, 128, 192]
+    for coor, val in zip(coords, values):
+        rr, cc = circle(coor[1], coor[0], 20)
+        image[rr, cc] = val
+    thresholds = [64, 128]
+    assert np.array_equal(thresholds, threshold_multiotsu(image))
```
hmaarrfk: This is a cool example!

alexandrejaguar: At last! :smile:

skimage/filters/thresholding.py
alexandrejaguar: Because of our [histogram implementation](https://github.com/scikit-image/scikit-image/blob/master/skimage/exposure/exposure.py#L93):
```
    nbins : int, optional
        integer arrays.
```

hmaarrfk: gah sorry, thanks! I was held up on merging this because I didn't fully understand why you used `len(bin_centers)` instead of `len(hist)` but thye are the same length.

PR #3945, opened 2019-06-02
Pooch -- on the fly download of datasets from github

hmaarrfk:
## Description

This PR builds up the infrastructure for Pooch. It would allow users not to have to download the datasets through pypi.
The main advantage would be to allow contributors to contribute larger datasets without massively increasing the size of the downloaded wheel.

xref: https://github.com/scikit-image/scikit-image/issues/3605 #3323 #3324

Proposed solution:
1. The old datasets are considered legacy the ability to load them is transparent to this change.
2. Pooch is used to download future user facing datasets. The first one of these is `data.cell`.
3. Pooch is used to download test datasets.
4. Testing from a git development build will not require pooch.
5. Pooch is an optional dependency, not a core dependency.
6. Tests that need datasets are skipped, and don't fail.
7. Data is shipped in the sdist
8. Chosen design does not require any particular organization in the git repo.
9. `data_dir` remains this string-like object that points to a path on the user's local computer. It contains at least the minimum data that we choose to include.
10. If github sets a throttling limit, we can release a new patch release which would fix issues to users.

The most important aspect is that, it allows us to add **new**, **larger** datasets in the future. Something that is **impossible** now.

Future enhancements:
1. Lazy load `data` to avoid creating and hashing datasets when the user doens't want to use sample data. This is currently being proposed for a particularly costly numpy import https://github.com/numpy/numpy/pull/14097

challenges encountered:
1. Somewhat breaks `data_dir`. The `data_dir` now contains a minimal subset of images, that we can control.
2. We had to hack utilizing the git directory as a secondary cache from pooch.
3. Depending on pooch is hard because it depends on all alot of the web stack. Not really fun to depend on and may negate the benefits for many users.

### Results
* On master, `python setup.py bdist_wheel` creates a 41 MB wheel archive
* On this branch, it creates a 27 MB archive.

The 27 MB breaks down into 2 important contributions:
1. The legacy datasets, approximately 7MB
2. Cython stuff that gets stripped away in our distributions by the wheel builder.

From this, we may shave off somewhere close to 27MB of the wheel, which would create archives on the order of 10-15MB instead of 30-35 MB.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


hmaarrfk:
pooch requires requests do we really want to depend on the whole http stack?

emmanuelle:
is the plan to have pooch as an optional dependency or a required one?

hmaarrfk:
Not too sure, I think optional dependency is nice for this case. I think it really depends on how complicated we want to make "installing scikit-image"

Our optional requirements are all very heavy.

Would we start to recommend everybody install

```
pip install scikit-image[optional]
```

skimage/data/__init__.py
```diff
@@ -32,18 +37,110 @@
            'horse',
            'hubble_deep_field',
            'immunohistochemistry',
-           'lbp_frontal_face_cascade_filename',
```
hmaarrfk: was this ever meant to be a public image? I kinda removed this from the API. Seemed to be important for tests only.

skimage/data/__init__.py
```diff
@@ -32,18 +37,110 @@
...
            'lfw_subset',
            'logo',
            'microaneurysms',
            'moon',
            'page',
+           'prefetch',
```
hmaarrfk: new function added to the API. Probably needs thinking and discussion.

skimage/data/__init__.py
```diff
@@ -32,18 +37,110 @@
...
+           'quantitative_phase_cells',
```
hmaarrfk: 2MB data set.

skimage/data/__init__.py
```diff
@@ -60,8 +157,43 @@ def load(f, as_gray=False):
         Image loaded from ``skimage.data_dir``.
     """
     use_plugin('pil')
-    return imread(_os.path.join(data_dir, f), as_gray=as_gray)
-
+    return imread(fetch(f), as_gray=as_gray)
+
+
+def quantitative_phase_cells():
+    """Image of two cells retrieved from a digital hologram.
```
hmaarrfk: This docstring definitely needs work, but I wanted to get the framework down and find the issue with the framework, rather than format docstrings for something that may never happen.

skimage/data/__init__.py
```diff
@@ -32,18 +37,110 @@
...
+    # even dev versions should use the online repo since users might be
+    # using the nightly builds
+    base_url = "https://github.com/hmaarrfk/scikit-image/raw/pooch/data/"
+    version = None
+else:
+    base_url = "https://github.com/hmaarrfk/scikit-image/raw/{version}/data/"
+    version = __version__
+
+# Create a new friend to manage your sample data storage
+image_fetcher = pooch.create(
```
hmaarrfk: this should probably be prefixed with an `_` to indicate it is internal???

skimage/data/__init__.py
```diff
@@ -32,18 +37,110 @@
...
+    path=pooch.os_cache("scikit-image"),
+    base_url=base_url,
+    version=version,
+    env="SCIKIT_IMAGE_DATA_DIR",
```
hmaarrfk: This will allow developers to use:
```
export SCIKIT_IMAGE_DATA_DIR=/home/mark2/git/scikit-image/data
```
nto to fetch online.

skimage/data/__init__.py
```diff
@@ -32,18 +37,110 @@
...
+        "test/disk-matlab-output.npz": "8a39d5c866f6216d6a9c9166312aa4bbf4d18fab3d0dcd963c024985bde5856b",
+        "test/diamond-matlab-output.npz": "02fca68907e2b252b501dfe977eef71ae39fadaaa3702ebdc855195422ae1cc2",
+        "test/bw_text.png": "308c2b09f8975a69b212e103b18520e8cbb7a4eccfce0f757836cd371f1b9094",
+        "test/bw_text_skeleton.npy": "9ff4fc23c6a01497d7987f14e3a97cbcc39cce54b2b3b7ee33b84c1b661d0ae1",
+        "test/_blobs_3d_fiji_skeleton.tif": "5182a2a94f240528985b8d15ec2aebbd5ca3c6b9214beff1eb6099c431e12b7b",
+        "test/checker_bilevel.png": "2e207e486545874a2a3e69ba653b28fdef923157be9017559540e65d1bcb8e28",
+    }
+)
+
+fetch = image_fetcher.fetch
```
hmaarrfk: This shortcut was really useful in migrating all the tests.

skimage/data/__init__.py
```diff
@@ -7,15 +7,20 @@
...
 from ..io import imread, use_plugin
 from .._shared._warnings import expected_warnings, warn
 from ..util.dtype import img_as_bool
 from ._binary_blobs import binary_blobs
 from ._detect import lbp_frontal_face_cascade_filename
 
+from .. import __version__
+
 import os.path as osp
 data_dir = osp.abspath(osp.dirname(__file__))
```
hmaarrfk: data_dir is  basically useless now.

The only migration path I can see, is to create some really hacky class, that implements the 

```
class _data_dir:
    def __init__(self, image_fetcher):
         self.image_fetcher = image_fetcher
    def __str__(self):
          warn(....)
          prefetch()
          return str(self.image_fetcher.path)
    def __add__(self, filename):
         if filename in self.image_fetcher.registry:
                warn(.....)
                return self.image_fetcher.fetch(filename)
         else:  # Globbing???
                prefetch()
                return str(self.image_fetcher.path) + filename

data_dir = _data_dir(image_fetcher)
```

skimage/io/tests/test_tifffile.py
```diff
@@ -20,32 +20,32 @@ def teardown():
...
+    img = imread(fetch('tests/multipage.tif'), pages=[1],
                  multifile=False, multifile_close=True, fastij=True,
                  is_ome=True)
     assert img.shape == (15, 10), img.shape
 
 def test_imread_handle():
-    expected = np.load(fetch('test/chessboard_GRAY_U8.npy'))
-    with open(fetch('test/chessboard_GRAY_U16.tif'), 'rb') as fh:
+    expected = np.load(fetch('tests/chessboard_GRAY_U8.npy'))
+    with open(fetch('tests/chessboard_GRAY_U16.tif'), 'rb') as fh:
```
jni: Why the renamed directory? So much line churn for an `s`...

hmaarrfk: Because the discrepancy with the regular tree would have driven me madd

hmaarrfk: Especially since I would have been the one that introduced it.

skimage/io/_plugins/fits_plugin.py
```diff
@@ -41,7 +41,7 @@ def imread(fname, dtype=None):
     lazy loading) to get all the extensions at once.
 
     """
-    if 'dtype' is not None:
+    if dtype is not None:
```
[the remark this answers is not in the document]
jni: Thanks for your fixup in a separate PR, just saw that.

skimage/io/collection.py
```diff
@@ -139,9 +139,9 @@ def imread_convert(f):
     Examples
     --------
     >>> import skimage.io as io
-    >>> from skimage import data_dir
+    >>> from skimage.data import fetch
 
-    >>> coll = io.ImageCollection(data_dir + '/chess*.png')
+    >>> coll = io.ImageCollection([fetch('chessboard_GRAY.png'), fetch('coins.png')])
```
jni: I feel like this is a problem, and maybe the hacky class is needed, if pooch can't handle glob expansions. Which would make sense.

hmaarrfk:
> @hmaarrfk thanks for making this happen! I agree that it's much easier to comment on a PR than in the abstract.

Yeah. Originally, I wanted to keep this PR small and a way to stop the data from growing, but then I figured it wouldn't be hard to see where it goes to see all the ramifications of the switch.

I'm good to reverting to a slower release too. We should have a mechanism to experiment with different technologies.

> Why is it useless? You think we shouldn't write to that directory ever?

In relation to datadir, by using pooch, the user is nolongwr interacting with the directory itself and operating on specific files, but they have to use the pooch object. I'm not sure it behaves like a full fledged directory, if it did, it would be really cool, in thinking like Google drive stream. By using pooch, we are really make the data available, and taking away the directory structure.

jni:
> In relation to datadir, by using pooch, the user is nolongwr interacting with the directory itself and operating on specific files, but they have to use the pooch object. I'm not sure it behaves like a full fledged directory, 

This is worth investigating in-depth. My hope is that using `download_all` will mean that users can interact with *some* directory on their hard drive as normal, if they so desire. (This also could minimize the impact on our tests, as we could run download-all before any testing, and then be fine as long as we redirect data_dir.

stefanv:
On June 2, 2019 23:39:41 Juan Nunez-Iglesias <notifications@github.com> wrote:
> In relation to datadir, by using pooch, the user is nolongwr interacting 
> with the directory itself and operating on specific files, but they have to 
> use the pooch object. I'm not sure it behaves like a full fledged directory,
>
> This is worth investigating in-depth. My hope is that using download_all 
> will mean that users can interact with some directory on their hard drive 
> as normal, if they so desire. (This also could minimize the impact on our 
> tests, as we could run download-all before any testing, and then be fine as 
> long as we redirect data_dir.
I strongly support this point: files, real files with known locations. No 
weird objects, just files read as NumPy arrays.

hmaarrfk:
> My hope is that using download_all will mean that users can interact with some directory on their hard drive as normal, if they so desire. 

Pooch claims to do this on a oer version level. Should be nice. 

> This also could minimize the impact on our tests, as we could run download-all before any testing, and then be fine as long as we redirect data_dir.

Our tests should be able to utilize the environment variable to avoid the need for any downloading.

If we do want to keep the ability to interact with datadir as before, it might be worth developing a hacky class ourselves, or working with pooch on this issue. 

The one thing I will warm about is pooch dropped 3.5 support. So we would either have to version them for a bit, at least until we drop 3.5

stefanv:
@leouieda How hard would it be to patch Pooch to be 3.5 compatible again?

leouieda:
@stefanv not too hard. There isn't really any patching that needed to the code itself. I just checked on my machine and all tests still pass on 3.5. It's mainly that conda-forge stopped building 3.5 packages so we figured we could drop it. Do you have a timeline for scikit-image dropping 3.5?

leouieda:
> My hope is that using download_all will mean that users can interact with some directory on their hard drive as normal, if they so desire.

This should be possible without any changes. Pooch maintains a directory with the downloaded files (see the [`Pooch.abspath`](https://www.fatiando.org/pooch/dev/api/generated/pooch.Pooch.html#pooch.Pooch) attribute). This can be mapped to `datadir` and all the data should be there after running a `download_all` function.

> OR, we can put the version string in the default directory to avoid such conflicts.

By providing `version`  to `pooch.create` the data directory already has the version in it. So different versions of scikit-image can coexist on the OS without clashing data files.

hmaarrfk:
> see the Pooch.abspath attribute

Cool thanks! 

> By providing version to pooch.create the data directory already has the version in it. So different versions of scikit-image can coexist on the OS without clashing data files.

Unfortunately, we don't include a `+` sign, but rather use the `.dev[0-9]{0, 5}` convention to denote development version numbers. Anyway we can modify Pooch to add a custom regex to detect development versions?

I think the current plan is to ship the datasets in the repo. It would be nice to also provide a secondary local cache in case users cloned the whole git repo.

leouieda:
> Anyway we can modify Pooch to add a custom regex to detect development versions?

For sure. Our current code is very simplistic and could use some upgrading: https://github.com/fatiando/pooch/blob/master/pooch/utils.py#L78

>  It would be nice to also provide a secondary local cache in case users cloned the whole git repo.

Do you mean to be able to have released versions use one local cache folder (e.g. `~/.cache/scitki-image/1.4.5`) and the *master* branch use another (e.g. `~/src/scikit-image/data`)? 

If so, this would probably be a bit more complicated refactor since the current implementation doesn't really know much about versions beyond appending them to the directory. I see two options that would work right now:

1. Link or copy the cloned repository data to the local cache so it won't be downloaded
2. In the library, check if the version relates to the *master* branch and set a different data directory in that case.

hmaarrfk:
> For sure. Our current code is very simplistic and could use some upgrading: https://github.com/fatiando/pooch/blob/master/pooch/utils.py#L78

Cool, we can discuss it in Pooch's issue tracker. I think the current logic will work for us, for now, but it would be nice to be able to move it into Pooch.

> Do you mean to be able to have released versions use one local cache folder (e.g. ~/.cache/scitki-image/1.4.5) and the master branch use another (e.g. ~/src/scikit-image/data)?

It wouldn't even need to be that complicated:
1. Check in the cache
2. Check in some hard coded directory. For scikit-image, this would be something like: `Path('__file__/../../data').resolve()`
3. Fetch from the URL online like the regular logic does.

hmaarrfk:
> I strongly support this point: files, real files with known locations

 @stefanv @jni I don't see why this is important

We want to teach people to analyze their data, not load our sample images with fancy reg-exps and file manipulations.

None of the examples used the `directory` functionality you both allude to.

Data storage and organization seems beyond the goals of scikit-image.

I think we should encourage people to use the functions for our sample data, enabling the addition of docstrings, online storage, and caching of data in a straightforward manner.


I will very likely submit an other PR cleaning up the data used in the example to add functions with docsrings for all the used data. Once that is in, the examples will no longer even mention `data_dir`. This won't be because we to create obscure objects, but more because we want to make our data start as numpy arrays, as opposed to "png" or "jpg" files.

skimage/data/__init__.py
```diff
@@ -7,20 +7,25 @@
...
-from ._detect import lbp_frontal_face_cascade_filename
+
+from .. import __version__
 
 import os.path as osp
 data_dir = osp.abspath(osp.dirname(__file__))
 
 __all__ = ['data_dir',
            'load',
+           'prefecth',
```
stefanv: typo

stefanv:
@hmaarrfk You wrote: "they have to use the pooch object".  I don't want users to even have to know about Pooch.  They should just know that they can either download the images to disk, or get them through our `data.*` methods.  Why complicate things by using intermediate objects?

doc/examples/applications/plot_morphology.py
```diff
@@ -191,7 +190,7 @@ def plot_comparison(original, filtered, filter_name):
 #*single-pixel wide skeleton*. It is important to note that this is
 #performed on binary images only.
 
-horse = io.imread(os.path.join(data_dir, "horse.png"), as_gray=True)
+horse = horse()
```
sciunto: Wouldn't it preferable to have variable names != functions?

hmaarrfk: Most likely, but we should establish the bigger picture problems first.

leouieda:
> @hmaarrfk You wrote: "they have to use the pooch object". I don't want users to even have to know about Pooch. They should just know that they can either download the images to disk, or get them through our `data.*` methods. Why complicate things by using intermediate objects?

This is the ideal case. Pooch is only used in the background to download and return a file name in the users computer. Here is example (see [here](https://github.com/fatiando/rockhound/blob/master/rockhound/etopo1.py):

```python
def fetch_etopo1(version, *, load=True, **kwargs):
    """
    Fetch the ETOPO1 global relief model.

    Parameters
    ----------
    version : str
        Which version of the dataset to load. Can be ``"ice"`` for the ice surface
        version, ``'bedrock'`` for the bedrock version.
    load : bool
        Wether to load the data into an :class:`xarray.Dataset` or just return the
        path to the downloaded data.

    Returns
    -------
    grid : :class:`xarray.Dataset` or str
        The loaded grid or the file path to the downloaded data.
    """
    available = {
        "ice": "ETOPO1_Ice_g_gmt4.grd.gz",
        "bedrock": "ETOPO1_Bed_g_gmt4.grd.gz",
    }
    # REGISTRY is a Pooch object
    fname = REGISTRY.fetch(available[version], processor=pooch.Decompress())
    if not load:
        return fname
    grid = xr.open_dataset(fname, **kwargs)
    return grid
```

So the user can get a file name if they want to load it themselves or get the grid loaded into xarray. For scikit-image, users could get the option of getting a numpy array.

skimage/exposure/tests/test_exposure.py
```diff
@@ -564,7 +564,7 @@ def test_is_low_contrast():
 # =======================
 
 def test_dask_histogram():
-    pytest.importorskip('dask', reason="dask python library is not installed")
+    pytest.importorskip('dask')
```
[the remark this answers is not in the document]
lagru: Same change [here](https://github.com/scikit-image/scikit-image/pull/3945/files#diff-58d8dd8e1e3c809cbd43d1820fe9f06aL533).

hmaarrfk:
@jni @stefanv  if we do implement Pooch, which I think we should, are we OK returning a `data_dir` path (well string) that points to an empty directory?

We would simply document that the user would have to call `fetch` or whatever we call it prior to use their desired datasets.

sciunto:
> @jni @stefanv if we do implement Pooch, which I think we should, are we OK returning a `data_dir` path (well string) that points to an empty directory?
> 
> We would simply document that the user would have to call `fetch` or whatever we call it prior to use their desired datasets.

yes

stefanv:
I'm fine with that too.

leouieda:
> can we initialize this directory with a readme file explaining this?

Right now, the README could be hosted in the repository and downloaded by Pooch as "dataset" on first import (when the folder will likely be created). This is kind of hacky. We could implement a new argument to pass in text for a README to `pooch.create` that can be will be created along with the folder. Is this something that would be good to have?

hmaarrfk:
I wonder if it is just easier to give it multiple locations to look for things in. Creating a Readme file would definitely be cool, though it seems like 2 lines in whatever library implements it. It may save a user some confusion.

hmaarrfk:
@jni @stefanv I think such a big change taught me a few things:

1. The important aspect of this change, is to be able to add bigger datasets, rather than shrinking the size of the skimage image. Therefore, we can likely roll out pooch in master (and in 0.16 if desired) without such intruisive changes.
2. The challenge with rolling out pooch is that the `skimage/data` directory is used for two things:
    a. Holding datasets that the users should interact with.
    b. Holding datasets only used for tests

To make the pooch PR less intrusive, a good cleanup would have to be use to move files as part of 2b into their own directory. Happy to do that. It seems to be a good idea irrespective of pooch.

In terms of design decisions that were wrong with this PR:

1. The pooch base url should simply point to the `skimage` subdirectory in this repo. This would allow us to use pooch on any file here with the file structure as is. For example, the registry entry for `astronaut` would be `"data/astronaut.png"`. This would help us move any file in the skimage outside of distribution and simply fetched from github on demand. It also helps us avoid moving files around simply to please pooch. To do this, we have to remove the `config.add_data_dir` in each `setup.py` we wish to transition.

2. We should probably only transition the images in the `skimage/data` directory for now. One the data in `skimage/data` is cleaned up and moved to pooch, we should be able to immediately start adding much larger datasets.

3. Specific data file that we will want to include as part of the standard distribution can be added using `add_data_file` from numpy instead of `add_data_dir`

https://docs.scipy.org/doc/numpy/reference/distutils.html#numpy.distutils.misc_util.Configuration.add_data_files

4. I think we should ask for help from upstream to enable the addition of an other source directory (Hi @leouieda ) to check before going online. This directory would point to
```
# skimage/data/__init__.py
from pathlib import Path

Path(__file__).parent / '..'
```
and allow pooch to search for things in the skimage installed directory. This would really help developers that have cloned the repo and testing locally.

5. We probably should also provide an environment variable that would ask all data file to be included in the distribution. (This is really where 4 is needed). I could see downstream packagers willing to ship with our large datasets if they feel it is better for their users.

skimage/data/__init__.py
```diff
@@ -50,6 +53,172 @@
...
+        if directory is not None:
+            image_fetcher.path = directory
+
+        for filename in image_fetcher.registry:
+            fetch(filename)
+    finally:
+        image_fetcher.path = old_dir
+
+
+def lbp_frontal_face_cascade_filename():
```
hmaarrfk: this was moved in because i decided it didn't serve its own file for one function.

skimage/data/__init__.py
```diff
@@ -78,7 +247,7 @@ def camera():
     camera : (512, 512) uint8 ndarray
         Camera image.
     """
-    return load("camera.png")
+    return load("data/camera.png")
```
hmaarrfk: prefixing things with `data` allows all datasets to be relative to the root `skimage` directory. this means that we can now put any file we want in pooch.

skimage/feature/setup.py
```diff
@@ -11,7 +11,8 @@ def configuration(parent_package='', top_path=None):
 
     config = Configuration('feature', parent_package, top_path)
     config.add_data_dir('tests')
-
+    config.add_data_files('orb_descriptor_positions.txt')
```
hmaarrfk: orb file was moved to the directory where it belongs. These are more like coefficient positions anyway as opposed to data.

skimage/io/tests/test_collection.py
```diff
@@ -26,14 +26,19 @@ def test_string_sort():
...
 
-    pattern = [os.path.join(data_dir, pic)
-               for pic in ['camera.png', 'color.png']]
-
-    pattern_matched = [os.path.join(data_dir, pic)
-                       for pic in ['camera.png', 'moon.png']]
-
     def setUp(self):
         reset_plugins()
+        data.fetch('data/color.png')
```
hmaarrfk: This was moved into the setup function so we can deterministically call `data.fetch`. I guess there are other ways of cleaning things up too.

hmaarrfk:
@scikit-image/core this is ready for a second review in terms of usability.

By default, scikit-image will ship with what I refer to as legacy data. This contains old data that was made available in 0.15. I don't see an easy way to deprecate this without really depending on pooch.

Test data, and new datasets can be added to the pooch registry with this framework.

In this PR, I believe we still ship with much of the test data that isn't located in the `skimage/data` directory.
We can make future PRs to slowly stop shipping test data and start slimming down the wheel once this is pulled in (large changes to many files are really annoying to rebase).

To ensure that scikit-image can continue being tested without pooch, we can skip the tests that require pooch.

sciunto:
Tremendous work. I'm appreciative.

I would like to express some concerns regarding the file hosting.
In this PR, github links are used. What will happen if:
* github is offline
* github decides to limit the bandwidth
* we are forced to change the URL (for any reason), especially regarding the hardcoded url that will be shipped with the future releases.

Wouldn't be safer to have some sort of redirection (based on scikit-image.org) to be more resilient and evolutive (a CDN for example)?

hmaarrfk:
A few answers:
* github is offline: wait, there will be serious other problems in the world
* github decides to limit the bandwidth: make a new patch release
* we are forced to change the URL: make a new patch release

We could provide a way to override the URL with an environment variable, but in truth, that will be much harder to teach users how to use than simply releasing a new point release.

Hard coded is only hard coded in the loose sense of the word in python at the end of the day. Users that know how to use environment variables and really need to could patch scikit-image themselves.

skimage/data/__init__.py
```diff
@@ -48,6 +53,159 @@
...
+
+
+def _fetch(data_filename):
+    # potential_local_file is the location of the file that would be shipped
+    # with scikit-image.
+    #
+    # Two things can cause the file to already be in the distribution:
+    # 1. The user is running it from the github repo, in which case
+    #    we should avoid a downloading from the internet
+    # 2. We chose to ship this file with skimage by default
```
jni: Make this a docstring? =)

skimage/data/__init__.py
```diff
@@ -48,6 +53,159 @@
...
+        raise ValueError(
+            "The resquested file {data_filename} is not part of the "
+            "scikit-image distribution."
+            "".format(data_filename=data_filename))
+    return resolved_path
+
+
+if has_pooch:
+    # Pooch expects a `+` to exist in development versions.
+    # Since scikit-image doesn't follow that convetion, we have to manually
```
jni:
```suggestion
    # Since scikit-image doesn't follow that convention, we have to manually
```

jni:
@sciunto in my (admittedly quick) review I didn't spot the github URL hardcoding?

I agree that legacy datasets should continue to be distributed with skimage for now. We can move them in a subsequent PR. The main goal right now is to open up the possibility of using bigger and better datasets in the documentation.

For storage, I suggest the Open Science Framework, which has a "keep the lights on" fund to last 50 years. https://osf.io . I put the 200+MB dataset for the skan paper there and haven't had any issues.

skimage/data/__init__.py
```diff
@@ -48,6 +53,159 @@
...
+    # provide it with the URL and set the version to None
+    if 'dev' in Version(__version__).version:
+        pooch_version = __version__.replace('.dev', '+')
+    else:
+        pooch_version = __version__
+    url = "https://github.com/scikit-image/scikit-image/raw/{version}/skimage/"
```
sciunto: @jni Over here :)

hmaarrfk:
> For storage, I suggest the Open Science Framework, which has a "keep the lights on" fund to last 50 years. https://osf.io . I put the 200+MB dataset for the skan paper there and haven't had any issues.

There are currently many test data sets (15 MB) that are included in the GitHub repo. Should we move those too?

For new datasets, it is possible to add specific links to the OSF. Deleting the files from the GitHub repo won't reduce the size of the repo unless you want to rewrite histor (which I don't recommend). I could include a link for the new cell data set there that isn't part of the legacy data (for now).

skimage/data/__init__.py
hmaarrfk: Why? This is an implementation detail. A docgring might be justified though explaining how we check local cache too and the design choices behind the two errors that can be raised

hmaarrfk:
Well, here is a short demo of how custom urls are integrated into pooch. The OSF urls are not the most friendly ones...

sciunto:
I would not move the test data set for now. As you mention, there is not gain from the git side. It can be left as a future decision, I would prefer to see this merged asap.

Perfect for the custom url demo.

Why is it failing for the min_requirement test cases?

hmaarrfk:
> Perfect for the custom url demo.

should we just comment it out for now?

> Why is it failing for the min_requirement test cases?

Not too sure, I remember that happened before too..... lets see.

skimage/data/tests/test_data.py
```diff
@@ -102,6 +102,9 @@ def test_lfw_subset():
 
 
 def test_cell():
-    """ Test that "cell" image can be loaded. """
+    """ Test that "cell" image can be loaded."""
+    # Needs pooch, or git development. using testing.fetch will
+    # cause the test to be skipped instead of failing if neither of
+    # those two conditions are met.
+    fetch('data/cells.png')
```
hmaarrfk:
```suggestion
    fetch('data/cell.png')
```

hmaarrfk:
I think the issue is that old Cython probably didn't define some constant correctly. Weird that it is only showing up now.

hmaarrfk:
I think it is some cython bug where `__file__` isn't super well defined. I moved the code to python, which seems a little more reliable.

Note I tried cython 0.27.3 on python 3.7 and it was still giving the same errors.

hmaarrfk:
I'm not sure if this should be a blocker for 0.16. This touches quite a bit of infrastructure, test, download, and build, and should probably be broken up into smaller diffs. That, or somebody really needs to review in detail that there aren't any unintended consequences (like downloading files twice from github during builds like was being done before).

skimage/data/__init__.py
hmaarrfk: the usefulness of datadir has been restored

skimage/data/__init__.py
```diff
@@ -32,18 +37,120 @@
...
+           'quantitative_phase_cells',
            'text',
            'retina',
            'rocket',
            'stereo_motorcycle']
 
 
+# Pooch expects a `+` to exist in development versions.
+# Since scikit-image doesn't follow that convetion, we have to manually provide
+# it with the URL and set the version to None
```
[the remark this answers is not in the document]
hmaarrfk: Not in this PR.

hmaarrfk:
@jni, @scikit-image/core any interest in continuing this work? scikit-image turns out to be one of the larger downloads in the scientific stack. It would be nice to be able to provide a path to keep that size under control.

stefanv:
Definitely still interested; sorry if you haven't received enough feedback. Where are things standing right now?

Wrt multiple downloads,we can use the Travis cache to store images. For testing, do we want to mock out the actual fetching?

.travis.yml
```diff
@@ -135,7 +151,9 @@ install:
         pip uninstall cython -y;
         pip install dist/scikit-image-*.tar.gz;
       else
-        pip install --no-build-isolation .;
+        # installing in editable mdoe avoids many downloads from pooch
```
stefanv: These Travis changes are confusing to me. Why does this avoid many factors downloads?

Also note typo.

skimage/data/__init__.py
```diff
@@ -49,6 +54,159 @@
...
+try:
+    import pooch
+    from pooch.utils import file_hash
+    has_pooch = True
+except ImportError:
+    has_pooch = False
+    import hashlib
+
+    def file_hash(fname):
+        # Versioned from pooch.utils
```
stefanv: Unclear what this means

skimage/data/__init__.py
```diff
@@ -49,6 +54,159 @@
...
+            "Loading the file {data_filename} requires the dependency "
+            "'pooch' to be installed. Install it with the command "
+            "'pip install pooch' or 'conda install pooch' as appropriate."
+            "".format(data_filename=data_filename))
+    else:
+        # This is only expected to happen if the user calls
+        # ``skimage.data.load`` with a filename that isn't part of the
+        # skimage distribution
+        raise ValueError(
+            "The resquested file {data_filename} is not part of the "
```
stefanv: Requested

skimage/data/__init__.py
```diff
@@ -49,6 +54,159 @@
...
+    return resolved_path
+
+
+if has_pooch:
+    # Pooch expects a `+` to exist in development versions.
+    # Since scikit-image doesn't follow that convention, we have to manually
+    # remove `.dev` with a `+` if it exists.
+    # This helps pooch understand that it should look in master
+    # to find the required files
+    pooch_version = __version__.replace('.dev', '+')
```
stefanv: I'm confused now: are the image files going to live in this repository, or elsewhere?

.travis.yml
hmaarrfk: Test datasets aren't shipped anymore. 

The pooch logic will search for the data next to the `py` files, so in editable mode, they are all present.

skimage/data/__init__.py
hmaarrfk: Copied as a local version.

skimage/data/__init__.py
hmaarrfk: thanks.

hmaarrfk:
There is little incentive to remove the current binary files from the github repository. They already take the space in the repo, and we cannot delete from everybody's computer without rewriting git history (and in the process invalidating all pending PRs).

The idea here is to:
1. Store current data set in **this** repo. Continue shipping them for now.
2. Store test datasets in **this**repo, do not ship them anymore.
3. Store other "small" datasets in **this** repo, do not ship them in future.
4. Store large datasets in other archival locations as shown in an example.

1 and 2 allow tests to run VERY quickly when using editable mode.

2 means that users would have to install pooch to run the full test suite. The partial test suite still has decent coverage and should help users validate their install.

3. and 4. open up the possibility for contributors to store new datasets where they find appropriate.

hmaarrfk:
Honestly, the registry isn't fully necessary, but it does give a good check for the files.

Other things to consider:
1. Generating the hashes is tedious, and I've already done it.
2. Validating the hashes should give users some ease of mind.
3. If we do decide to remove certain datasets from the distribution, it becomes an easy one liner.

Test datasets are fetched on the fly when installing from pypi, so the hashes are useful there.

stefanv:
We definitely want hashes. If those go in a registry, then by all means.

Were the "this"s in your comment above meant to be bold or links to other repos?

I know we've already committed files to this repo, so we can't get that space back. But for consistency in our code it may be easier to host our images all in an external repo (sdists, e.g., would never have the images included, so the space saving argument isn't valid there).

hmaarrfk:
*this* means the main scikit-image repo.

I've kinda already done the work to freeze the contents of the sdist in the setup.py files 

I guess we need to update the `check_sdist.py`. But for now, maybe we can leave this as a future problem?

With regards to hosting in a different repository: Multiple repos is a little bit of a pain. With pooch, storing in the same repo means that we automatically create the right tag in the repository that matches the version of the library. Development versions point to master.

With regards to saving storage space, I'm not so worried about development computers. I'm very worried about end user space (ie. I have 20+ computers where scikit-image is installed, but only 2 where I do skimage development). Not to mention everytime I run CI tests.

jni:
I agree with Mark's assessment. There is not much value in ripping out files from the history now. There is a lot of value in ripping out files from wheel and source distributions in the future.

GitHub is actually not super generous with file bandwidth, so we might find that we need to store the files elsewhere in the not-so-far future. But we can cross that bridge after we get to it.

skimage/data/__init__.py
```diff
@@ -49,6 +54,159 @@
...
+def _fetch(data_filename):
+    # potential_local_file is the location of the file that would be shipped
+    # with scikit-image.
+    #
+    # Two things can cause the file to already be in the distribution:
+    # 1. The user is running it from the github repo, in which case
+    #    we should avoid a downloading from the internet
+    # 2. We chose to ship this file with skimage by default
+    potential_local_file = osp.join(skimage_distribution_dir, data_filename)
+    resolved_path = osp.join(data_dir, '..', data_filename)
```
stefanv: The variable naming here is a bit confusing.  Can we perhaps go for something like `gh_repository_path` and `cache_path` to make it clear what is being referred to?

The logic below I'd also perhaps write some one-liner functions to make intent clear:

```
if in_cache(filename) and has_hash(filename, expected_hash):
    ...
```

skimage/data/__init__.py
```diff
@@ -49,6 +54,159 @@
...
+            "'pooch' to be installed. Install it with the command "
+            "'pip install pooch' or 'conda install pooch' as appropriate."
+            "".format(data_filename=data_filename))
+    else:
+        # This is only expected to happen if the user calls
+        # ``skimage.data.load`` with a filename that isn't part of the
+        # skimage distribution
+        raise ValueError(
+            "The requested file {data_filename} is not part of the "
+            "scikit-image distribution."
```
stefanv: Perhaps "The requested file {} is not known to scikit-image."

sciunto: Why this is not an f-string?

jni: Because the PR predates our decision to go py36+. But yes, that is definitely the right solution now!

stefanv:
@hmaarrfk Looks like we're quite close on this one?

emmanuelle:
I fixed the merge conflict, don't forget to pull first :-).

skimage/data/_registry.py
```diff
@@ -0,0 +1,143 @@
...
+    "feature/tests/data/TransformedX75Y75.png": "a1e9ead5f8e4a0f604271e1f9c50e89baf53f068f1d19fab2876af4938e695ea",
+    # Data below here isn't distributed in github, and individual URLs must
+    # be added to the urls dictionary below
+    # For files on osf.io, the SHA-2 seems to be providing the SHA256 hash
+    "data/schizonts.zip": "6bec7ae303423164c54ad7144d5700c9beae69e06ada3b9bc351ea1a3f0ca1ae",
+}
+
+registry_urls={
+    # TODO: Remove the schizonts dataset before merging into master
+    "data/schizonts.zip": "https://files.osf.io/v1/resources/svpfu/providers/osfstorage/5a162bbf9ad5a10269cc1d4f?action=download&version=2&direct"
```
emmanuelle: what is this?

hmaarrfk: we wanted an example of a dataset located somewhere else that is more ammenable to larger download volumnes

emmanuelle: OK thanks :-)

skimage/data/__init__.py
```diff
@@ -49,6 +54,159 @@
...
+            "Loading the file {data_filename} requires the dependency "
+            "'pooch' to be installed. Install it with the command "
+            "'pip install pooch' or 'conda install pooch' as appropriate."
+            "".format(data_filename=data_filename))
+    else:
+        # This is only expected to happen if the user calls
+        # ``skimage.data.load`` with a filename that isn't part of the
+        # skimage distribution
+        raise ValueError(
+            "The requested file {data_filename} is not part of the "
```
hmaarrfk:
```suggestion
            f"The requested file {data_filename} is not part of the "
```

skimage/data/__init__.py
```diff
@@ -49,6 +54,159 @@
...
+            "scikit-image distribution."
+            "".format(data_filename=data_filename))
```
hmaarrfk:
```suggestion
```

skimage/data/__init__.py
hmaarrfk:
```suggestion
            "scikit-image.")
```

emmanuelle:
This will be a great addition, thanks @hmaarrfk ! I hope we can merge it soon ;-).

skimage/data/__init__.py
hmaarrfk: ok i added a docstring that talks about what the function *does* but not *how* it does it.

skimage/data/__init__.py
hmaarrfk: Thanks for the guidance. Please see rework.

skimage/data/__init__.py
hmaarrfk: The image files **can** be added to this repository, but they will no longer be added to the the pypi distribution.

I think we should avoid adding more binary files to the github repo, since this will enable us to add much larger files elsewhere.

hmaarrfk:
@emanuelle, please see additions. This may also require testing from a nightly build to see how the user experience is.

skimage/data/__init__.py
```diff
@@ -49,6 +54,209 @@
...
+            buff = fin.read(chunksize)
+            while buff:
+                hasher.update(buff)
+                buff = fin.read(chunksize)
+        return hasher.hexdigest()
+
+
+def _hash_hash(path, expected_hash):
+    """Check if the provided path has the expected hash."""
+    if not osp.exists(resolved_path):
```
emmanuelle: looks like this line makes the CI fail.

emmanuelle: should this be `path` instead?

emmanuelle:
@hmaarrfk ok, so I should get the build from the Azure CI, is this correct?

skimage/data/__init__.py
hmaarrfk:
```suggestion
    if not osp.exists(path):
```

hmaarrfk:
Get the build? I think you can build locally. But all CIs should pass here.

emmanuelle:
Oh sure I can get your branch and work from it, it's just that you had suggested to get the "nightly build".

skimage/data/__init__.py
```diff
@@ -49,6 +54,209 @@
...
+
+    """
+    resolved_path = osp.join(data_dir, '..', data_filename)
+    expected_hash = registry[data_filename]
+
+    # Case 1:
+    # The file may already be in the data_dir regardless of if Pooch is
+    # installed. We may have decided to ship it in the scikit-image
+    # distribution
+    if _hash_hash(resolved_path, expected_hash):
```
emmanuelle:
```suggestion
    if _has_hash(resolved_path, expected_hash):
```

skimage/data/__init__.py
```diff
@@ -49,6 +54,209 @@
...
+        hasher = hashlib.sha256()
+        with open(fname, "rb") as fin:
+            buff = fin.read(chunksize)
+            while buff:
+                hasher.update(buff)
+                buff = fin.read(chunksize)
+        return hasher.hexdigest()
+
+
+def _hash_hash(path, expected_hash):
```
emmanuelle:
```suggestion
def _has_hash(path, expected_hash):
```

skimage/data/__init__.py
```diff
@@ -49,6 +54,209 @@
...
+        return resolved_path
+
+
+    # Case 2:
+    # The user is using a cloned version of the github repo, which
+    # contains both the publically shipped data, and test data.
+    # In this case, the file would be located relative to the
+    # skimage_distribution_dir
+    gh_repository_path = osp.join(skimage_distribution_dir, data_filename)
+    if _has_hash(gh_repository_path):
```
emmanuelle:
```suggestion
    if _has_hash(gh_repository_path, expected_hash):
```

skimage/data/__init__.py
```diff
@@ -49,6 +54,209 @@
...
+
+
+def _hash_hash(path, expected_hash):
+    """Check if the provided path has the expected hash."""
+    if not osp.exists(path):
+        return False
+    return file_hash(path) == expected_hash
+
+
+def _fetch(data_filename):
```
emmanuelle: so, don't we want to expose this function at all? The question is whether we want to return arrays for all datasets, or whether in some cases we are just happy downloading the image and letting the user to choose what to do (for large datasets, loading in memory only a chunck of the array etc.). Anyway this can be decided later on and should not block the PR, it's easier to add a function to the API than removing it ;-).

skimage/data/_registry.py
```diff
@@ -0,0 +1,143 @@
...
+    "color/tests/data/luv_array_d65_2.npy": "962ce180132c6c11798cbc423b2b204d1d10187670f6eb5dec1058eaad301e0e",
+    "color/tests/data/luv_array_d75_10.npy": "e1cc70d56eb6789633d4c2a4059b9533f616a7c8592c9bd342403e41d72f45e4",
+    "color/tests/data/luv_array_d75_2.npy": "07db3bd59bd89de8e5ff62dad786fe5f4b299133495ba9bea30495b375133a98",
+    "color/tests/data/luv_array_e_2.npy": "41b1037d81b267305ffe9e8e97e0affa9fa54b18e60413b01b8f11861cb32213",
+    "color/tests/ciede2000_test_data.txt": "2e005c6f76ddfb7bbcc8f68490f1f7b4b4a2a4b06b36a80c985677a2799c0e40",
+    "data/astronaut.png": "88431cd9653ccd539741b555fb0a46b61558b301d4110412b5bc28b5e3ea6cb5",
+    "data/brick.png": "7966caf324f6ba843118d98f7a07746d22f6a343430add0233eca5f6eaaa8fcf",
+    "data/cell.png": "8d23a7fb81f7cc877cd09f330357fc7f595651306e84e17252f6e0a1b3f61515",
+    "data/camera.png": "361a6d56d22ee52289cd308d5461d090e06a56cb36007d8dfc3226cbe8aaa5db",
+    "data/cells_qpi.npz": "a9c5212894bd4de8fddebd679500aff67f04b4e25e41c3f347f3e876ce648252",
```
emmanuelle: This one is causing an `HTTPError` when I call `download_all`.

skimage/data/__init__.py
```diff
@@ -49,6 +54,209 @@
...
+try:
+    import pooch
+    from pooch.utils import file_hash
+    has_pooch = True
+except ImportError:
+    has_pooch = False
+    import hashlib
+
+    def file_hash(fname):
+        # Versioned (i.e. copied from) from pooch.utils
```
emmanuelle:
```suggestion
        # Vendored (i.e. copied from) from pooch.utils
```

skimage/data/_registry.py
emmanuelle:
```suggestion
```

emmanuelle:
@hmaarrfk I finally took the time to play with your branch :-). There are a couple of bugs for which I left change suggestions, you can just accept them if you're ok (I could not push to your branch). Then we can see if tests pass.

emmanuelle:
Tests were passing on my machine, so :crossed_fingers:

skimage/data/_registry.py
hmaarrfk: Not too sure why. How did the data set change recently?

skimage/data/__init__.py
sciunto: If I'm correct, the current behavior (download only) has the advantage to reduce the bandwidth consumption (or I misunderstood your comment).

emmanuelle: Maybe my comment was not clear :-). I was just wondering if we should expose this `_fetch` function in the public API, since it allows to download one specific file. At the moment we have access to the `download_all` function, and to `load` which calls `imread` on the downloaded file. For specific file formats not compatible with `imread` this might not work. But it's ok, this can be added later.

emmanuelle:
@jni did you have the time to take a look at this one? For me it is ready.

skimage/data/__init__.py
```diff
@@ -49,6 +54,209 @@
...
+    Parameters
+    ----------
+    directory: path-like, optional
+        The directory where the dataset should be stored.
+
+    Notes
+    -----
+    scikit-image will only search for images stored in the default directory.
+    Only specify the directory if you wish to download the images to your own
+    folder for a particular reason. You may access the location of the data
```
jni: of the *default* data directory?

skimage/data/__init__.py
```diff
@@ -49,6 +54,209 @@
...
+def lbp_frontal_face_cascade_filename():
+    """
+    Returns the path to the XML file containing information about the weak
+    classifiers of a cascade classifier trained using LBP features. It is part
+    of the OpenCV repository [1]_.
+
+    References
+    ----------
+    .. [1] OpenCV lbpcascade trained files
+           https://github.com/Itseez/opencv/tree/master/data/lbpcascades
```
jni: Should we fix this to a specific commit, rather than master?

skimage/data/_registry.py
```diff
@@ -0,0 +1,142 @@
+# flake8: noqa
+
+# This minimal dataset was available as part of
+# scikit-image 0.15 and will be retained until
+# further notice.
+# Testing data and additional datasets should only
+# be made available by pooch
```
jni: What does this mean exactly? ie how does one make a dataset available via pooch?

skimage/data/_registry.py
```diff
@@ -0,0 +1,142 @@
...
+    "feature/tests/data/TransformedX-130Y130.png": "1cda90ed69c921eb7605b73b76d141cf4ea03fb8ce3336445ca08080e40d7375",
+    "feature/tests/data/TransformedX130Y130.png": "bb10c6ae3f91a313b0ac543efdb7ca69c4b95e55674c65a88472a6c4f4692a25",
+    "feature/tests/data/TransformedX75Y75.png": "a1e9ead5f8e4a0f604271e1f9c50e89baf53f068f1d19fab2876af4938e695ea",
+    # Data below here isn't distributed in github, and individual URLs must
+    # be added to the urls dictionary below
+    # For files on osf.io, the SHA-2 seems to be providing the SHA256 hash
+    "data/schizonts.zip": "6bec7ae303423164c54ad7144d5700c9beae69e06ada3b9bc351ea1a3f0ca1ae",
+}
+
+registry_urls={
```
jni:
```suggestion
registry_urls = {
```

skimage/data/_registry.py
jni: Do we necessarily want to remove this? Maybe it's fine to leave it, I can work on an example with this once this is merged. =)

skimage/data/setup.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
+from _registry import legacy_datasets
```
jni:
```suggestion
from ._registry import legacy_datasets
```
?

skimage/data/tests/test_data.py
```diff
@@ -102,6 +102,9 @@ def test_lfw_subset():
 
 
 def test_cell():
-    """ Test that "cell" image can be loaded. """
+    """ Test that "cell" image can be loaded."""
+    # Needs pooch, or git development. using testing.fetch will
+    # cause the test to be skipped instead of failing if neither of
+    # those two conditions are met.
+    fetch('data/cell.png')
```
jni: Why fetch this image but not others?

skimage/feature/_orb_descriptor_positions.py
```diff
@@ -0,0 +1,11 @@
...
+import numpy as np
+
+# Putting this in cython was giving strange bugs for different versions
+# of cython which seemed to indicate troubles with the __file__ variable
+# not being defined. Keeping it in pure python makes it more reliable
+this_dir = os.path.dirname(__file__)
+POS = np.loadtxt(os.path.join(this_dir, "orb_descriptor_positions.txt"),
+                 dtype=np.int8)
+POS0 = np.ascontiguousarray(POS[:, :2])
+POS1 = np.ascontiguousarray(POS[:, 2:])
```
jni: Nice! I love moving Cython stuff out to Python. =)

skimage/filters/rank/tests/test_rank.py
```diff
@@ -78,14 +85,12 @@ def setup(self):
         # Set again the seed for the other tests.
         np.random.seed(0)
         self.selem = morphology.disk(1)
-        self.refs = np.load(os.path.join(skimage.data_dir,
-                                         "rank_filter_tests.npz"))
```
jni: I think it would be less intrusive here to use `np.load(fetch(` here and forget about the pytest fixture. I like fixtures, but they seem unnecessary when using a UnitTest-style class.

skimage/filters/tests/test_thresholding.py
```diff
@@ -304,7 +304,10 @@ def test_li_constant_image_with_nan():
 
 
 def test_li_arbitrary_start_point():
-    cell = data.cell()
+    try:
+        cell = data.cell()
+    except RuntimeError:
+        pytest.skip('pooch or git repository needed to load data.cell')
```
jni: Why not add pooch to our test requirements?

jni: (and again, then the test needs to fail gracefully when pooch can't reach the internet.)

skimage/data/setup.py
```diff
@@ -0,0 +1,32 @@
...
+
+def configuration(parent_package='', top_path=None):
+    from numpy.distutils.misc_util import Configuration
+
+    config = Configuration('data', parent_package, top_path)
+    # This minimal dataset was available as part of
+    # scikit-image 0.15 and will be retained until
+    # further notice.
+    # Testing data and additional datasets should only
+    # be made available by pooch
```
stefanv:
```suggestion
    # be made available via pooch
```

skimage/data/tests/test_data.py
stefanv: This fetch should catch the local version, no?

skimage/filters/tests/test_thresholding.py
stefanv: Why "git repository" needed?

stefanv: And why not pooch a required dep?

hmaarrfk:
@jni, @emmanuelle this is now working again after a rebase.

skimage/data/_registry.py
stefanv: A bit big; I think we said the `cell.tif` image would be a good fit.

stefanv: (That's the one used in our skimage-tutorials repo.)

stefanv:
@hmaarrfk I left some comments during last review; take a look when you have a minute.

skimage/data/_registry.py
```diff
@@ -0,0 +1,138 @@
...
+    "feature/tests/data/OriginalX130Y130.png": "7fdd4c06d504fec35ee0703bd7ed2c08830b075a74c8506bae4a70d682f5a2db",
+    "feature/tests/data/OriginalX75Y75.png": "c5cd58893c93140df02896df80b13ecf432f5c86eeaaf8fb311aec52a65c7016",
+    "feature/tests/data/TransformedX-130Y130.png": "1cda90ed69c921eb7605b73b76d141cf4ea03fb8ce3336445ca08080e40d7375",
+    "feature/tests/data/TransformedX130Y130.png": "bb10c6ae3f91a313b0ac543efdb7ca69c4b95e55674c65a88472a6c4f4692a25",
+    "feature/tests/data/TransformedX75Y75.png": "a1e9ead5f8e4a0f604271e1f9c50e89baf53f068f1d19fab2876af4938e695ea",
+    "data/cells.tif": "2120cfe08e0396324793a10a905c9bbcb64b117215eb63b2c24b643e1600c8c9",
+}
+
+registry_urls = {
+    "data/cells.tif": "https://github.com/scikit-image/skimage-tutorials/raw/master/images/cells.tif",
```
hmaarrfk: @stefanv cherry-picked from @jni 

the power of collaboration!

jni:
Hmmm, are we now hitting our GitHub bandwidth limits? Or...?

```pytb
================================== FAILURES ===================================
________________________ test_imread_uint16_big_endian ________________________
    @testing.skipif(si is None, reason="skimage not installed")
    def test_imread_uint16_big_endian():
        expected = np.load(fetch('data/chessboard_GRAY_U8.npy'))
>       img = imread(fetch('data/chessboard_GRAY_U16B.tif'))
python37-x64\lib\site-packages\skimage\external\test_tifffile.py:30: 
python37-x64\lib\site-packages\skimage\_shared\testing.py:216: in fetch
    return data._fetch(data_filename)
python37-x64\lib\site-packages\skimage\data\__init__.py:155: in _fetch
    resolved_path = image_fetcher.fetch(data_filename)
python37-x64\lib\site-packages\pooch\core.py:399: in fetch
    downloader(url, tmp.name, self)
python37-x64\lib\site-packages\pooch\downloaders.py:121: in __call__
    response.raise_for_status()
self = <Response [401]>
    def raise_for_status(self):
        """Raises stored :class:`HTTPError`, if one occurred."""
    
        http_error_msg = ''
        if isinstance(self.reason, bytes):
            # We attempt to decode utf-8 first because some servers
            # choose to localize their reason strings. If the string
            # isn't utf-8, we fall back to iso-8859-1 for all other
            # encodings. (See PR #3538)
            try:
                reason = self.reason.decode('utf-8')
            except UnicodeDecodeError:
                reason = self.reason.decode('iso-8859-1')
        else:
            reason = self.reason
    
        if 400 <= self.status_code < 500:
            http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
    
        elif 500 <= self.status_code < 600:
            http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
    
        if http_error_msg:
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 401 Client Error: UNAUTHORIZED for url: https://github.com/scikit-image/scikit-image/raw/master/skimage/data/chessboard_GRAY_U16B.tif
python37-x64\lib\site-packages\requests\models.py:941: HTTPError
---------------------------- Captured stderr call -----------------------------
Downloading data file 'data/chessboard_GRAY_U16B.tif' from remote data store 'https://github.com/scikit-image/scikit-image/raw/master/skimage/data/chessboard_GRAY_U16B.tif' to 'C:\Users\appveyor\AppData\Local\scikit-image\scikit-image\Cache\master'.
```

from https://ci.appveyor.com/project/scikit-image/scikit-image/builds/31198015/job/1b53bj1uccu1i9y2#L8947

Thanks for the cherry-picking, @hmaarrfk! Plus all the original work of course :joy:

jni:
The Travis 3.7 failure appears to have absolutely nothing to do with us, but it is currently reproducible... Looks like a Python 3.7 image got moved...

```
Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/14.04/x86_64/python-3.7.tar.bz2

0.04s$ curl -sSf --retry 5 -o python-3.7.tar.bz2 ${archive_url}

curl: (22) The requested URL returned error: 404 Not Found

Unable to download 3.7 archive. The archive may not exist. Please consider a different version.
```

jni:
Still worried about our bandwidth:

```pytb
================================== FAILURES ===================================
_________________________ TestColorconv.test_lab2xyz __________________________
self = <skimage.color.tests.test_colorconv.TestColorconv testMethod=test_lab2xyz>
    def test_lab2xyz(self):
        assert_array_almost_equal(lab2xyz(self.lab_array),
                                  self.xyz_array, decimal=3)
    
        # Test the conversion with the rest of the illuminants.
        for I in ["d50", "d55", "d65", "d75"]:
            for obs in ["2", "10"]:
                fname = "color/tests/data/lab_array_{0}_{1}.npy".format(I, obs)
>               lab_array_I_obs = np.load(fetch(fname))
python37-x64\lib\site-packages\skimage\color\tests\test_colorconv.py:345: 
python37-x64\lib\site-packages\skimage\_shared\testing.py:216: in fetch
    return data._fetch(data_filename)
python37-x64\lib\site-packages\skimage\data\__init__.py:155: in _fetch
    resolved_path = image_fetcher.fetch(data_filename)
python37-x64\lib\site-packages\pooch\core.py:399: in fetch
    downloader(url, tmp.name, self)
python37-x64\lib\site-packages\pooch\downloaders.py:121: in __call__
    response.raise_for_status()
self = <Response [504]>
    def raise_for_status(self):
        """Raises stored :class:`HTTPError`, if one occurred."""
    
        http_error_msg = ''
        if isinstance(self.reason, bytes):
            # We attempt to decode utf-8 first because some servers
            # choose to localize their reason strings. If the string
            # isn't utf-8, we fall back to iso-8859-1 for all other
            # encodings. (See PR #3538)
            try:
                reason = self.reason.decode('utf-8')
            except UnicodeDecodeError:
                reason = self.reason.decode('iso-8859-1')
        else:
            reason = self.reason
    
        if 400 <= self.status_code < 500:
            http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
    
        elif 500 <= self.status_code < 600:
            http_error_msg = u'%s Server Error: %s for url: %s' % (self.status_code, reason, self.url)
    
        if http_error_msg:
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 504 Server Error: Gateway Time-out for url: https://github.com/scikit-image/scikit-image/raw/master/skimage/color/tests/data/lab_array_d55_10.npy
```

from https://ci.appveyor.com/project/scikit-image/scikit-image/builds/31359220/job/ati0g8v56gik640v#L8912

Otherwise this is ready...! Should we just pull the trigger and risk it? Or find somewhere else for storage?

jni:
Also, whoever merges this, please make sure to "Squash and Merge", not Merge directly.

alexdesiqueira:
> Or find somewhere else for storage?

:grimacing: do we have options @jni? Any ideas?

jni:
@alexdesiqueira I briefly suggested the Open Science Framework (osf.io), but @hmaarrfk wasn't too happy about their URLs. Another option is to make a separate repo + gh-pages service with cloudflare CDN. I have this set up for my blog (don't ask me how :joy:), and it seems to work pretty well. That is at least as reliable as using our main repo github, but it requires a bit of extra coordination.

jni:
Anyway, the failure seems to be rare enough, so I'm happy to merge and cross that bridge when we need to.

hmaarrfk:
Why were the last builds cancelled

jni:
@hmaarrfk I've noticed entire builds disappearing in other PRs. I think there must have been a glitch with Travis. At any rate, this passes locally. Given approvals by @stefanv and @emmanuelle, I'm inclined to squash and merge. Thoughts?

sciunto:
At some point, we must jump and trust the paragliding wing :)

leouieda:
> At some point, we must jump and trust the paragliding wing :)

As someone who's built parts of the paragliding wing, this makes me slightly nervous :slightly_smiling_face:

hmaarrfk:
the rebase messed up some travis build stuff. Hopefully this one will work.

sciunto:
Lights are all green, let's pull the trigger.

jni:
LOL that merge hash :point_up_2: is ominous! :sweat_smile: 

At any rate, this is amazing! We can start using real scientific datasets in our docs! :tada:

hmaarrfk:
That's crazy!

Super happy that this is in!

PR #3949, opened 2019-06-02
don't cast integer to float before using it as integer in numpy logspace

hmaarrfk:
by putting everything in an array, `num` is cast to an float, then it gets passes as the `num` parameter which expects to be an integers. Numpy is warning us that this behaviour is deprecated.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -417,9 +417,7 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
     if log_scale:
         start = np.log10(min_sigma)
         stop = np.log10(max_sigma)
-        sigma_list = np.stack([np.logspace(_start, _stop, num_sigma)
-                               for _start, _stop in zip(start, stop)],
-                              axis=1)
+        sigma_list = np.logspace(start, stop, num_sigma)
     else:
         scale = np.linspace(0, 1, num_sigma)[:, None]
         sigma_list = scale * (max_sigma - min_sigma) + min_sigma
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -417,7 +417,12 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
     if log_scale:
         start = np.log10(min_sigma)
         stop = np.log10(max_sigma)
-        sigma_list = np.logspace(start, stop, num_sigma)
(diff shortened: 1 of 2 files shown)


jni:
Erm, or maybe not? Tests are actually failing. :man_facepalming:

hmaarrfk:
Hmm I swear I was running the exact test that is failing locally. I guess I'll keep working on this one

skimage/feature/blob.py
```diff
@@ -415,10 +415,9 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
     max_sigma = np.asarray(max_sigma, dtype=float)
 
     if log_scale:
-        start, stop = np.log10(min_sigma)[:, None], np.log10(max_sigma)[:, None]
-        space = np.concatenate(
-            [start, stop, np.full_like(start, num_sigma)], axis=1)
-        sigma_list = np.stack([np.logspace(*s) for s in space], axis=1)
+        start = np.log10(min_sigma)
+        stop = np.log10(max_sigma)
+        sigma_list = np.logspace(start, stop, num_sigma)
```
jni: my mind is blown!

hmaarrfk: right. numpy does so much magic. I'm obsessed with this single copy paradigm, I had to try it here.

hmaarrfk: Unfortunately, it isn't realy implemented in numpy 1.14
```
        delta = stop - start
        if num > 1:
            step = delta / div
>           if step == 0:
E           ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

In recent versions of numpy, they are now using `any`
https://github.com/numpy/numpy/blob/master/numpy/core/function_base.py#L150

I guess I'll just leave a comment there that it can be optimized when numpy gets bumped

Numpy 1.14, 1.15 still uses the logic that doesn't work for our usecase
https://github.com/numpy/numpy/blob/maintenance/1.14.x/numpy/core/function_base.py#L124
https://github.com/numpy/numpy/blob/maintenance/1.15.x/numpy/core/function_base.py#L124

Its only been fixed to numpy 1.16
https://github.com/numpy/numpy/blob/maintenance/1.16.x/numpy/core/function_base.py#L150

jni: Damn. It sucks not being on the bleeding edge. Thanks for the investigation!

jni:
@scikit-image/core anyone up for the second review?

PR #3951, opened 2019-06-02
avoid potential divide by zero in meijering filter.

hmaarrfk:
## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/ridges.py
+++ b/skimage/filters/ridges.py
@@ -213,7 +213,11 @@ def meijering(image, sigmas=range(1, 10, 2), alpha=None,
             auxiliary = np.max(auxiliary, axis=0)
 
             # Rescale image intensity
-            filtered = np.abs(auxiliary) / np.abs(np.max(auxiliary))
+            denum = np.abs(np.max(auxiliary))
+            if denum == 0.:
+                filtered = np.zeros_like(auxiliary)
+            else:
+                filtered = np.abs(auxiliary) / denum
 
             # Remove background
             filtered = np.where(filtered > 0, filtered, 0)


skimage/filters/ridges.py
```diff
@@ -213,7 +213,11 @@ def meijering(image, sigmas=range(1, 10, 2), alpha=None,
             auxiliary = np.max(auxiliary, axis=0)
 
             # Rescale image intensity
-            filtered = np.abs(auxiliary) / np.abs(np.max(auxiliary))
+            denum = np.abs(np.max(auxiliary))
+            if denum == 0.:
+                filtered = np.zeros_like(auxiliary)
+            else:
+                filtered = np.abs(auxiliary) / denum
```
jni: Are both branches covered by a test? Also, denum -> denom.

But anyway, maybe I would rework this as:

```python
# use max(..., 1) to avoid division by 0
filtered = np.abs(auxiliary) / max(np.abs(np.max(auxiliary)), 1)
```

I also wonder whether the order of abs and max is correct... It's a bit weird to take the abs of the max rather than the max of the abs...

hmaarrfk: I think auxiliary was a float

jni: what I mean is that `np.abs(np.max([-1, -2]))` is 1, while `np.abs(np.max([-1, -2, 0.5]))` is 0.5. The converse is 2 in both cases. I don't know which is correct in this case, but I've encountered `np.max(np.abs())` more in my life. =P

stefanv: I think almost always you rather want `max(abs(...))`.  Depends on what you want, of course, but the other version is likely to give you something close to zero for negative values—seldom what you want.

jni: @stefanv if it's something like curvature, I can imagine you want negative values to be "clipped" to be close to 0, but then you probably want `clip(max())` rather than `abs(max())`. Probably best to ask the original author. @DavidBreuer?

DavidBreuer:
Hi everyone, thanks for noticing this and including me:

- In principle, the denominator is "only" a number/proportionality constant that scales image intensity, but of course it should avoid division-by-zero and agree with the original definition [1].

- Looking into this again, I found that my original implementation is actually wrong! :( 
Figure_1
The corrected filtering yields results that are largely similar but with visible differences, see attached image. 

- Here are the relevant line changes in the meijering() function:

old + wrong [current implementation]:
```
if not black_ridges:
    image = invert(image)
eigenvalues = compute_hessian_eigenvalues(image, sigma, sorting='val')
auxiliary = np.max(auxiliary, axis=0)
filtered = np.abs(auxiliary) / np.abs(np.max(auxiliary))
filtered = np.where(filtered > 0, filtered, 0)
```
new + correct [1]:
```
if black_ridges:
    image = invert(image)
eigenvalues = compute_hessian_eigenvalues(image, sigma, sorting='abs')
auxiliary = auxiliary[1]
filtered = _divide_nonzero(auxiliary / np.min(auxiliary))
filtered = np.where(auxiliary < 0, filtered, 0)
```
- So, not the maximum eigenvalues are used but those of maximum magnitude and they are not divided by their overall maximum but their minimum value. Because sorting by eigenvalue magnitude is already used for the Frangi filter, this can be done using compute_hessian_eigenvalues defined in ridges.py.

- Moreover, in the original paper [1], the authors are aware that the denominator may be zero but state that it "in practice will always be smaller than zero". However, to avoid any division-by-zero error, I suggest using _divide_nonzero which is also already defined in ridges.py.

- @ all Does that sound reasonable to you?

- @hmaarrfk Do you want to adjust this PR accordingly or should I start a separate one to fix these issues?

[1] https://onlinelibrary.wiley.com/doi/pdf/10.1002/cyto.a.20022 (page 176, bottom left)

hmaarrfk:
Thanks for chiming in!

Please start a new one. Writing the tests will require more intimate knowledge of the filter than I have

hmaarrfk:
Closed in favour of #3957

PR #3952, opened 2019-06-02
Avoid numpy warning for bool labels. 

hmaarrfk:
## Description
Probably needs a test???

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -127,6 +127,8 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
     # operating on each label separately
     if labels is not None:
         label_values = np.unique(labels)
+        if np.issubdtype(label_values.dtype, np.bool):
+            label_values = label_values.astype(np.int8)
         # Reorder label values to have consecutive integers (no gaps)
         if np.any(np.diff(label_values) != 1):
             mask = labels >= 1


skimage/feature/peak.py
```diff
@@ -127,6 +127,8 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
     # operating on each label separately
     if labels is not None:
         label_values = np.unique(labels)
+        if np.issubdtype(label_values.dtype, np.bool):
+            label_values = label_values.astype(np.int8)
```
jni: I wonder if bool dtype here means they expect a labeling... Is this related to a bug report or failing/warning test?

hmaarrfk: This generates a warning. They implement diff using the subtraction operator so they claim that subtracting book types is no good.

jni: Sure but do we know of cases where bool labels have been passed here? it might be safer to just raise on bool

hmaarrfk: It was in a doc test. A binary image was constructed and passed in as a label

jni: I imagine it's probably used that way by mistake...

hmaarrfk: Hmm. Ok I'll find the reference and let you decide.

hmaarrfk:
I can't find the warning I was trying to avoid before. But if somebody does find it, the yshould probably use `scipy.ndimage.label` to label the boolean image first.

PR #3960, opened 2019-06-09
Set minimum version for imread so that it compiles from source on linux in test builds

hmaarrfk:
## Description

Seems  like imread is broken on python 3.5.

I have a feeling that this is because imread releases wheels for manylinux2010
https://www.python.org/dev/peps/pep-0571/

I don't understand why pip is choosing the manylinux2010 wheels

https://pypi.org/project/imread/#files

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/requirements/extras.txt
+++ b/requirements/extras.txt
@@ -1,1 +1,1 @@
-imread
+imread>=0.5.1


hmaarrfk:
I think this is because of the issue discussed here
https://github.com/luispedro/imread/issues/35

sciunto:
Can we move this testcase on another version? To make sure that we still have a test suite in the former configuration.

hmaarrfk:
minimum + optional + extra?

sciunto:
yes

hmaarrfk:
hmm setting the minimum version of imread might also solve the issue.

hmaarrfk:
The reason this works is that old versions of imread don't have `manylinux2010` wheels uploaded. I think this hack is fine for now. Until we stop to support manylinux1 that is.

PR #3966, opened 2019-06-13
Fix threshold_li (again) and add developer example

### LINKED ISSUE #3605: threshold_li changed from version 0.14.0 to version 0.14.1

jni:
## Description

This PR addresses the issues explained in https://github.com/scikit-image/scikit-image/issues/3605#issuecomment-494622300 . Specifically:

- Li thresholding optimisation is not guaranteed to be convex, and can get stuck in local minima or even *local maxima*, as long as they are stationary points.
- This happens in the cases one might expect: when histograms have three or more modes: each gap between the modes provides an opportunity for a local minimum.
- In order to combat this, we add an `initial_guess` parameter that takes in either a float or a function `[Array[float]] -> float` to obtain an initial guess from the image intensity values.
- I also add new tests, thanks to @hmaarrfk's new test image.
- Finally, I add a long developer example explaining Li thresholding in detail.

The example could use some simplification, and some of the layouts need improvement, but I think it's ready for an initial review.

Fixes #3605. 

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


doc/examples/developers/plot_threshold_li.py
```diff
@@ -0,0 +1,233 @@
...
+# guess from the image intensities, just like :func:`numpy.mean` does by
+# default. This might be a good option when many images with different ranges
+# need to be processed.
+
+def quantile_95(image):
+    return np.quantile(image, 0.95)
+
+iter_thresholds5 = []
+opt_threshold5 = filters.threshold_li(cell, initial_guess=quantile_95,
+                                      iter_callback=iter_thresholds5.append)
```
jni: @paulmueller would you be happy to use the form above? `initial_guess=lambda x: np.quantile(x, q=0.95)` for example, or defining the function as above. Since Li thresholding, it turns out, can get stuck in local optima, I think it is actually the cleanest solution for your images.

sciunto:
I haven't made a full expertise, but you did an awesome work of investigation!

doc/examples/developers/plot_threshold_li.py
paulmueller: Yes, sure thank you. Just a question (not a request): What happened to the idea of allowing multiple initial guesses (such as ``np.linspace(min, max, 10)`` and then take the best result?

jni: @paulmueller I played around with that but ultimately thought that it was too disruptive to the code (we would need an additional *nested* for-loop), not really in-line with the original method *or* with our "we don't do magic" policy, and would result in a significantly less-performant method. (10 times, say, ~6 iterations per optimization gives 60 iterations over the image, which is approaching the cost of brute-force.) I think the `quantile_95` method will work well for most datasets, once you know the characteristics of your data, and anyway the more sophisticated multi-start method is still available, as you can manually use an *external* loop to supply multiple initial guesses. (Though it is true that we should make `_cross_entropy` public for this last bit to work. I could be persuaded to do that.)

hmaarrfk:
```
  File "doc/examples/developers/plot_threshold_li.py", line 221, in <module>
    iter_callback=iter_thresholds5.append)
  File "/home/travis/venv/lib/python3.5/site-packages/skimage/filters/thresholding.py", line 605, in threshold_li
    t_next = initial_guess(image)
  File "doc/examples/developers/plot_threshold_li.py", line 217, in quantile_95
    return np.quantile(image, 0.95)
AttributeError: module 'numpy' has no attribute 'quantile'
The command "tools/travis/script.sh" exited with 1.
```

Time to bump numpy: to.... 1.15
https://docs.scipy.org/doc/numpy/reference/generated/numpy.quantile.html

hmaarrfk:
I guess percentile can be used, I would just add a comment as to why you chose to implement it with percentile as opposed to quantile.
https://docs.scipy.org/doc/numpy/reference/generated/numpy.percentile.html

jni:
Why are the docs getting built with minimum requirements?

hmaarrfk:
Not sure, that was a policy for a long time. I think the idea is that anybody that installed scikit-image should be able to run the examples.

stefanv:
They aren't built with minimum requirements. They need, e.g., scikit-learn for the facial classifier.

hmaarrfk:
I guesss they are built with the minimum requirements, even the optional ones. At least the ones we have listed.

https://github.com/scikit-image/scikit-image/blob/master/requirements/docs.txt

https://github.com/scikit-image/scikit-image/blob/master/tools/travis/script.sh#L32

doc/examples/developers/plot_threshold_li.py
```diff
@@ -0,0 +1,233 @@
...
+
+###############################################################################
+# In addition to allowing users to provide a number as an initial guess,
+# :func:`skimage.filters.threshold_li` can receive a function that makes a
+# guess from the image intensities, just like :func:`numpy.mean` does by
+# default. This might be a good option when many images with different ranges
+# need to be processed.
+
+def quantile_95(image):
+    return np.quantile(image, 0.95)
```
hmaarrfk: Maybe this could be replaced with
```
def quantile_95(image):
    # numpy 1.14, which scikit-image still supports, does not implement 
    # quantile, but rather percentile
    return np.percentile(image, 0.95 * 100)
```

skimage/filters/thresholding.py
```diff
@@ -530,6 +531,21 @@ def threshold_li(image, *, tolerance=None):
         is less than this value. By default, this is half the smallest
         difference between intensity values in ``image``.
 
+    initial_guess : float or Callable[[array[float]], float], optional
```
hmaarrfk: This is so complicated. I know you want to make a generic function, but you could just call your callable before calling `threshold_li`

IMO, initial_guess should just be `float`

doc/examples/developers/plot_threshold_li.py
```diff
@@ -0,0 +1,233 @@
...
+
+iter_thresholds5 = []
+opt_threshold5 = filters.threshold_li(cell, initial_guess=quantile_95,
```
hmaarrfk:
```suggestion
opt_threshold5 = filters.threshold_li(cell, initial_guess=np.quantile(cell, 0.95),
```

Following my other comment, this acheives the same thing, and really reduces the scope of the API.

Users that want to make things more generic, will know how to make their own function wrapper.

hmaarrfk:
Well our build infrastructure allows you to specify different minimum doc requirements and build requirements.
Build
https://travis-ci.org/scikit-image/scikit-image/jobs/547528654#L1009

Test:
https://travis-ci.org/scikit-image/scikit-image/jobs/547528654#L2174

Doc:
https://travis-ci.org/scikit-image/scikit-image/jobs/547528654#L3757

If this is what you want, then we can do this.

doc/examples/developers/plot_threshold_li.py
jni: @hmaarrfk this is fine for one image, but dramatically increases the code complexity for a large image collection.

jni: (and doesn't solve the problem that we do a lot of sanitising of the image, which the user won't benefit from here)

hmaarrfk: Ah. Ok. The sanitization point makes sense.

skimage/filters/thresholding.py
```diff
@@ -580,9 +596,26 @@ def threshold_li(image, *, tolerance=None):
     image -= image_min
     tolerance = tolerance or np.min(np.diff(np.unique(image))) / 2
 
-    # Initial estimate
-    t_curr = np.mean(image)
-    t_next = t_curr + 2 * tolerance
+    # Initial estimate for iteration. See "initial_guess" in the parameter list
```
sciunto: Why one of these case would be decreased by image_min (due to line 596) and not the others? I would have expected that all of them should be reduced.

skimage/filters/thresholding.py
```diff
@@ -580,9 +596,26 @@ def threshold_li(image, *, tolerance=None):
...
+    if initial_guess is None:
+        t_next = np.mean(image)
+    elif callable(initial_guess):
+        t_next = initial_guess(image)
+    elif np.isscalar(initial_guess):  # convert to new, positive image range
+        t_next = initial_guess - image_min
```
sciunto: What happens if this is negative?

skimage/filters/thresholding.py
jni: @sciunto at this point, `image` is already in the new range (with the original `image_min` removed), so typically the mean/function should apply to the pixels in the new range. However, if the input is a float, it needs to be rescaled to the new range, because typically users will give the float value in the original data range.

it's important to note that the purpose here is to *increase* the image range in cases where it is negative, so `np.log` makes no sense.

skimage/filters/thresholding.py
jni: Nothing good! Actually any value outside of the range of the initial data will cause the loop to fail. I'll add a check for that!

skimage/filters/thresholding.py
sciunto: "image is already in the new range", You are correct; my remark was stupid.

jni:
@sciunto @scikit-image/core this is ready for another review. Technically, we should be testing that neither the foreground nor the background are empty, but I think that is too expensive for a very remote edge case.

skimage/data/__init__.py
```diff
@@ -132,6 +132,37 @@ def checkerboard():
...
 
 
+def cell():
+    """Cell floating in saline.
+
+    This is a quantitative phase image retrieved from a digital hologram using
+    the Python library ``qpformat``. The image shows a cell with high phase
+    value, above the background phase.
+
+    Because of a banding pattern artifact in the background, this image is a
```
hmaarrfk: @jni is this the image I preprocessed? If so, maybe add the preprocessing code I applied to the original github image in some multiline string in the function itself.

hmaarrfk: just as a reference to other developers.

hmaarrfk:
@jni a few minor questions, feel free to merge if you feel like you answered these in a satisfactory manner: 
1. why is this a developer example, and not a standard example?
2. I don't agree with bumping the minimum doc version of numpy just for `quantile` instead of percentile. Seems like additional maintenance burden.

jni:
@hmaarrfk We might need to draw more distinctions between our example types, but as far as I see it, things that discuss mostly the internals of a function, rather than how to use our API, go in the developer docs. Currently, though, the most compelling reason for making this a developer doc is that the example uses a private API. I've added a note to that effect.

We can revisit the questions of where to place the tutorial and whether to surface this API in a future PR.

I've changed the function to using percentile and reverted the new NumPy requirement for docs.

hmaarrfk:
lol, I didn't even notice you did that. I'm sure somebody will end up using it if you are for debugging purposes. We should just make it public. 

Happy to merge this as is though. but now that you explosed the underscore method, if 0.16 gets pulled in before, you will have to add a deprecation to the underscore method.

jni:
> if 0.16 gets pulled in before, you will have to add a deprecation to the underscore method.

no. The whole point of the underscore is that you shouldn't use it. Hence the warning. (Same way NumPy pulled `_validate_lengths` without warning, and we had zero right to complain.)

hmaarrfk:
That's fair. But if it is useful for debugging how an algorithm is supposed to work, it seems natural to share. 

Maybe I've just listened to too many Linus talks where every observable behaviour is a feature and hence can't be removed without thinking it will break people's system.

hmaarrfk:
Thanks for the discussion jni

jni:
We are not the Linux kernel, and this is in developer docs. We have tons of other private functions we break all the time. And we can even add an alias for this once we expose it publicly. 

In short, I don’t think this is an issue. Yes we might expose it eventually, maybe in evaluate. I don’t think now is the time.

jni:
Oh good LOL. 👍

PR #3967, opened 2019-06-14
Modify ProjectiveTransform.__nice__ formatting

Erotemic:
Logical extension to #3525

In the previous merged PR there was more support for an alternate matrix format. This PR implements that format. 

I also fixed a small inconsistency in the docstrings and added a doctest. I originally created the doctest for debugging, but I think doctests generally improve code quality --- even for simple things like this --- so I opted to keep it in.  

Advertisement: This doctest can be run using the [xdoctest](https://github.com/Erotemic/xdoctest) CLI:  `xdoctest -m skimage.transform._geometric ProjectiveTransform.__nice__`. Any doctest can theoretically be run this way: e.g. `xdoctest -m skimage.morphology.selem ellipse`.

### DIFF
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -1,6 +1,7 @@
 import math
 import numpy as np
 from scipy import spatial
+import textwrap
 
 from .._shared.utils import get_bound_method_class, safe_as_int
 
@@ -728,12 +729,7 @@ class ProjectiveTransform(GeometricTransform):
         >>> print(ProjectiveTransform(np.random.rand(3, 3)))
         """
         npstring = np.array2string(self.params, separator=', ')
-        lines = npstring.split('\n ')
-        # Initial indentation
-        lines = ['    ' + p for p in lines]
-        # Add an extra space to every non-first line for alignment
-        lines[1:] = [' ' + p for p in lines[1:]]
-        paramstr = 'matrix=\n' + '\n'.join(lines)
+        paramstr = 'matrix=\n' + textwrap.indent(npstring, '    ')
         return paramstr
 
     def __repr__(self):
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -721,13 +721,7 @@ class ProjectiveTransform(GeometricTransform):
                             "types.")
 
     def __nice__(self):
-        """common 'paramstr' used by __str__ and __repr__
-
-        Examples
-        --------
-        >>> print(ProjectiveTransform())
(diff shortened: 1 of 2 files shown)


skimage/transform/_geometric.py
```diff
@@ -720,24 +720,31 @@ def __add__(self, other):
...
+        >>> print(ProjectiveTransform())
+        >>> print(ProjectiveTransform(np.random.rand(3, 3)))
+        """
         npstring = np.array2string(self.params, separator=', ')
-        # add a newline and indentation after the first square braket to ensure
-        # the columns of the params matrix are aligned in the repr
-        lines = npstring[1:].split('\n ')
-        indented_lines = ['    ' + p for p in lines]
-        paramstr = 'matrix=[\n' + '\n'.join(indented_lines)
+        lines = npstring.split('\n ')
```
hmaarrfk:
```suggestion
        lines = npstring.split('\n')
```

skimage/transform/_geometric.py
```diff
@@ -720,24 +720,31 @@ def __add__(self, other):
...
+        # Initial indentation
+        lines = ['    ' + p for p in lines]
+        # Add an extra space to every non-first line for alignment
+        lines[1:] = [' ' + p for p in lines[1:]]
```
hmaarrfk:
```suggestion

```

skimage/transform/_geometric.py
```diff
@@ -720,24 +720,31 @@ def __add__(self, other):
...
+        paramstr = 'matrix=\n' + '\n'.join(lines)
```
hmaarrfk: Can't you avoid most of this with
```
    return "matrix=\n" + textwrap.indent(np.array2string(params), '    ')
```

Erotemic: Yup, I was overthinking it.

Erotemic:
I ended up removing the doctest in `__nice__` because the builtin Python doctest module forces you to specify a "want" string whenever you write anything to stdout (unless you use a specific inline directive). As an additional note, this behavior wouldn't happen if xdoctest --- which allows for a less strict and more expressive syntax -- was used as the backend for running doctests.

skimage/transform/_geometric.py
```diff
@@ -1,6 +1,7 @@
 import math
 import numpy as np
 from scipy import spatial
+import textwrap
```
jni: Whoa, Python stdlib ftw! =) Had never seen this before, super nice.

Issue #1103, opened 2014-08-15
poly2mask


dpereira84:
Hi everybody,

I need a function poly2mask equal to function in Maltlab, someone could tell me if there is already some implementation in Python ?

Regards
Danilo

zhangda82:
Hello jni,

Can you help clarify the meaning of parameters for draw.polygon() ? From your example the 1st and 2nd parameters are list of row and col coordinates of the vertex, but from the official document of skimage.draw.polygon(), they are y and x coordinates defined elsewhere. Thanks!

Da

MalikAsif:
i am getting this issue.
from .._shared.utils import warn

ImportError: cannot import name 'warn'
anyone can help

hmaarrfk:
@MalikAsif this issue is 4 years old. Please open a new issue and describe the issue you are having from the beginning.

stefanv:
Yes, why not.

MalikAsif:
Thanks, got it

hadim:
I confirm this is a useful function.

For the record, I have been experimenting with a matplotlib-based similar function. I compared it to the one provided by @jni.

Both give similar results and the skimage's one seems more efficient. The speed is probably dependent on the shape of the polygon and the size of the image so you should benchmark with your own data both implementations (if speed is a concern for you).

See more details here: https://gist.github.com/hadim/fa89b50bbd240c486c61787d205d28a6

jni:
I'm going to reopen this to note that we want to have this function added to skimage. Thanks for chiming in, @hadim!

hadim:
I felt like I needed to submit this PR after making you reopen the issue xD

See https://github.com/scikit-image/scikit-image/pull/3971

soupault:
Added in #3971 .

PR #3971, opened 2019-06-19
Add poly2mask to convert a polygon to a mask.

hadim:
See https://github.com/scikit-image/scikit-image/issues/1103 for a background.

In short, this PR add a `poly2mask` function that converts a polygon to a mask given an image shape.

I added two different implementations. I have no idea if you guys are ok to have two implementations or if you prefer to stick with only one. Let me know.

## Checklist

  existing benchmark
- [X] Unit tests

## For reviewers

  later.


skimage/transform/_poly2mask.py
```diff
@@ -0,0 +1,53 @@
...
+        return _poly2mask_mpl(image_shape, polygon)
+
+    raise ValueError(
+        "Wrong backend selected. Choose either 'default' or 'matplotlib'.")
+
+
+def _poly2mask_skimage(image_shape, polygon):
+    polygon = np.array(polygon)
+    fill_row_coords, fill_col_coords = draw.polygon(
+        polygon[:, 1], polygon[:, 0], image_shape)
```
jni: These should not be inverted. I suggest using the same syntax as in my original comments, as it is more self-documenting:

```python
vertex_row_coords, vertex_col_coords = polygon.T
fill_row_coords, fill_col_coords = draw.polygon(vertex_row_coords, vertex_col_coords, image.shape)
```

(accounting for PEP8 of course — hard to do on the github editor)

skimage/transform/_poly2mask.py
```diff
@@ -0,0 +1,53 @@
...
+        return _poly2mask_skimage(image_shape, polygon)
+    elif backend == 'matplotlib':
+        return _poly2mask_mpl(image_shape, polygon)
+
+    raise ValueError(
+        "Wrong backend selected. Choose either 'default' or 'matplotlib'.")
+
+
+def _poly2mask_skimage(image_shape, polygon):
+    polygon = np.array(polygon)
```
jni: Use `np.asarray` to prevent a copy when `polygon` is already an array.

skimage/transform/_poly2mask.py
```diff
@@ -0,0 +1,53 @@
...
+    fill_row_coords, fill_col_coords = draw.polygon(
+        polygon[:, 1], polygon[:, 0], image_shape)
+    mask = np.zeros(image_shape, dtype=np.bool)
+    mask[fill_row_coords, fill_col_coords] = True
+    return mask
+
+
+def _poly2mask_mpl(image_shape, polygon):
```
jni: Remove this backend. We are actually aiming to remove the mpl dependency for non-GUI code in the medium-term. At any rate we don't need two backends, it just adds to the complexity for little gain.

skimage/transform/_poly2mask.py
hadim: Wow, I didn't know `np.array` were doing a copy!

sciunto:
I think I would prefer `polygon2mask` instead of  `poly2mask`

hadim:
@sciunto: do you mean an example in the doc or only in the docstring?

hadim:
I have added an example in the docstring. Let me know whether it's enough.

jni:
Great! So simple now! I'd like to ask for two extra things:

1. I think this belongs in draw rather than transform. Could you move it there? Sorry that I didn't catch this earlier.
2. Could you also add a Notes section before the Examples section, which notes that this function does *not* do any border checking, so that all the vertices need to be within the given shape?

hadim:
Done.

skimage/draw/_polygon2mask.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Notes
+    -----
+    This function does not do any border checking, so that all
+    the vertices need to be within the given shape.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from skimage.transform import polygon2mask
```
sciunto: you forgot that one @hadim  :)

hadim: Wait. Are the examples in docstrings executed? Wow!

sciunto:
yes, it's a [doctest](https://docs.python.org/3.5/library/doctest.html)

hadim:
I am coding in Python for almost 10 years now, and still, I am learning new stdlib features all the time xD

jni:
Thanks for the PR, @hadim! And yes, I have found that contributing to open source is the best way to learn new stuff! =)

ayushnoori:
> Great! So simple now! I'd like to ask for two extra things:
> 
> 1. I think this belongs in draw rather than transform. Could you move it there? Sorry that I didn't catch this earlier.
> 2. Could you also add a Notes section before the Examples section, which notes that this function does _not_ do any border checking, so that all the vertices need to be within the given shape?

Hi all - is there an available Python implementation of polygon2mask() that _does_ do border checking, so that vertices outside of the given shape are acceptable?

jni:
@ayushnoori actually, can you test this particular function? My reading of the code, and the call to `draw.polygon`, is that it *does* do border checking, by using `draw.polygon` and passing in the image shape. See:

https://scikit-image.org/docs/dev/api/skimage.draw.html#polygon

So, if you confirm this, I think that we would appreciate a PR removing that statement from the documentation!

ayushnoori:
Sure @jni - I'll take a look. Tagging my colleague @cmagdamo as well.

ayushnoori:
I'm also interested in understanding the differences between `scikit-image polygon2mask()` and MATLAB's `poly2mask()', see: https://blogs.mathworks.com/steve/2006/12/05/poly2mask-and-roipoly-part-1/.

PR #3615, opened 2018-12-19
[WIP] Make dtype consistent in filters.rank fonctions

Henley13:
## Description
Boolean inputs return uint8 output with `filters.rank.maximum` or `filters.rank.mean`for example. We want to keep the input and output dtypes identical.

## Checklist
- [ ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Test for boolean inputs
- [ ] Test for `_handle_input` in filters/rank/generic.py
- [x] Be sure to (re)cast the output properly (after the cython computation ?)

## References
Closes  #3577
Closes #1010

## For reviewers
- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

Issue #1010, opened 2014-05-10
Increase coverage of skimage.filters.rank.*


ahojnnes:
This module has insufficient test coverage. Also, the tests should be restructured.

PR #3981, opened 2019-06-27
 Make dtype consistent in filters.rank functions

### LINKED ISSUE #3577: rank filter maximum incorrect doc of output dtype
## Description

The documentation of `skimage.filters.rank.maximum` says that the returned array has the same dtype as the input:

> Returns: | out : 2-D array (same dtype as input image)

However this is not the case for `bool`.

## Way to reproduce
```python
import numpy as np
from skimage.filters.rank import maximum

arr = np.random.randint(2, size=(10, 10)).astype(bool)
print(arr.dtype)
result = maximum((arr), selem=np.ones((3, 3)))
print(result.dtype)
```

Output:
> bool
> uint8

Tested on skimage `0.14.1`.

Related to #466, #1244, #1010

clementkng:
## Description
Supersedes gh-3615. Fixed test behavior locally and added new tests to ensure that `dtypes` remain the same in the output of the `filter.rank` functions. For now, the `test_input_boolean_dtype` in `test_rank` is currently commented out because a `TypeError: No matching signature found` error is raised when the `filter.rank.maximum` function is called w/ a `dtype np.bool_` arguments. I have no experience w/ Cython, but I assume that means we need to add Cython support somehow for `np.bool_` dtypes. Before I explore this further, does anyone know off the top of their heads how to do this? 

Closes #3577
May address some, but perhaps not all, issues in #1010

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/rank/generic.py
+++ b/skimage/filters/rank/generic.py
@@ -1,4 +1,4 @@
-r"""
+"""
 
 General Description
 -------------------
@@ -25,11 +25,10 @@ histogram computation. By default the entire image is filtered.
 
 This implementation outperforms grey.dilation for large structuring elements.
 
-Input image can be 8-bit or 16-bit, for 16-bit input images, the number of
-histogram bins is determined from the maximum value present in the image.
-
-Result image is 8-/16-bit or double with respect to the input image and the
-rank filter operation.
+Input images will be cast in unsigned 8-bit integer or unsigned 16-bit integer
+if necessary. The number of histogram bins is then determined from the maximum
+value present in the image. Eventually, the output image is cast in the desired
+dtype.
 
 To do
 -----
@@ -64,9 +63,49 @@ __all__ = ['autolevel', 'bottomhat', 'equalize', 'gradient', 'maximum', 'mean',
 
 
 def _handle_input(image, selem, out, mask, out_dtype=None, pixel_size=1):
+    """Preprocess and verify input for filters.rank methods.
 
+    Parameters
+    ----------
+    image : 2-D array (integer, float or boolean)
+        Input image.
+    selem : 2-D array (integer, float or boolean)
+        The neighborhood expressed as a 2-D array of 1's and 0's.
+    out : 2-D array (integer, float or boolean)
+        If None, a new array is allocated.
+    mask : ndarray (integer, float or boolean)
+        Mask array that defines (>0) area of the image included in the local
+        neighborhood. If None, the complete image is used (default).
+    out_dtype : data-type
+        Desired output data-type. Default is None, which means we cast output
+        in input dtype.
+    pixel_size : int
+        Dimension of each pixel. Default value is 1.
+
+    Returns
+    -------
+    image : 2-D array (np.uint8 or np.uint16)
+    selem : 2-D array (np.uint8)
+        The neighborhood expressed as a binary 2-D array.
+    out : 3-D array (same dtype out_dtype or as input)
+        Output array. The two first dimensions are the spatial ones, the third
+        one is the pixel vector (length 1 by default).
+    mask : 2-D array (np.uint8)
+        Mask array that defines (>0) area of the image included in the local
+        neighborhood.
+    n_bins : int
+        Number of histogram bins.
+    out_dtype : data-type
+        Output data-type.
+
+    """
+    # check `image` is 2D
     assert_nD(image, 2)
-    if image.dtype not in (np.uint8, np.uint16):
+
+    # if `image` is not an integer 8-bit/16-bit or a boolean, we cast it as
+    # an unsigned 8-bit image.
+    input_dtype = image.dtype
+    if image.dtype not in (np.uint8, np.uint16, np.bool_):
         message = ('Possible precision loss converting image of type {} to '
                    'uint8 as required by rank filters. Convert manually using '
                    'skimage.util.img_as_ubyte to silence this warning.'
@@ -74,9 +113,13 @@ def _handle_input(image, selem, out, mask, out_dtype=None, pixel_size=1):
         warn(message, stacklevel=2)
         image = img_as_ubyte(image)
 
-    selem = np.ascontiguousarray(img_as_ubyte(selem > 0))
+    # make `image` contiguous in memory
     image = np.ascontiguousarray(image)
 
+    # cast `selem` as an unsigned 8-bit array and make it contiguous in memory
+    selem = np.ascontiguousarray(img_as_ubyte(selem > 0))
+
+    # idem with `mask`
     if mask is None:
         mask = np.ones(image.shape, dtype=np.uint8)
     else:
@@ -86,17 +129,20 @@ def _handle_input(image, selem, out, mask, out_dtype=None, pixel_size=1):
     if image is out:
         raise NotImplementedError("Cannot perform rank operation in place.")
 
+    # allocate a new output array if necessary
     if out is None:
-        if out_dtype is None:
-            out_dtype = image.dtype
-        out = np.empty(image.shape+(pixel_size,), dtype=out_dtype)
+        out = np.empty(image.shape + (pixel_size,))
     else:
         if len(out.shape) == 2:
             out = out.reshape(out.shape+(pixel_size,))
+    out = out.astype(image.dtype)
 
-    is_8bit = image.dtype in (np.uint8, np.int8)
+    # cast output in the desired dtype
+    if out_dtype is None:
+        out_dtype = input_dtype
 
-    if is_8bit:
+    # compute the number of bins in the histogram from `image` dtype
+    if image.dtype in (np.uint8, np.int8):
         n_bins = 256
     else:
         # Convert to a Python int to avoid the potential overflow when we add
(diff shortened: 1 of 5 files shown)


skimage/filters/rank/generic.py
```diff
@@ -64,16 +63,54 @@
...
+    image : 2-D array (integer, float or boolean)
+        Input image.
+    selem : 2-D array (integer, float or boolean)
+        The neighborhood expressed as a 2-D array of 1's and 0's.
+    out : 2-D array (integer, float or boolean)
+        If None, a new array is allocated.
+    mask : ndarray (integer, float or boolean)
+        Mask array that defines (>0) area of the image included in the local
+        neighborhood. If None, the complete image is used (default).
+    out_dtype : data-type
```
sciunto:
```suggestion
    out_dtype : data-type, optional
```

skimage/filters/rank/generic.py
```diff
@@ -64,16 +63,54 @@
...
+        Desired output data-type. Default is None, which means we cast output
+        in input dtype.
+    pixel_size : int
```
sciunto:
```suggestion
    pixel_size : int, optional
```

skimage/filters/rank/generic.py
```diff
@@ -126,11 +192,51 @@ def _apply_scalar_per_pixel(func, image, selem, out, mask, shift_x, shift_y,
...
+        Mask array that defines (>0) area of the image included in the local
+        neighborhood. If None, the complete image is used (default).
+    shift_x, shift_y : int
+        Offset added to the structuring element center point. Shift is bounded
+        to the structuring element sizes (center must be inside the given
+        structuring element).
+    out_dtype : data-type, optional
+        Desired output data-type. Default is None, which means we cast output
+        in input dtype.
+    pixel_size : int
```
sciunto:
```suggestion
    pixel_size : int, optional
```

skimage/filters/rank/generic.py
```diff
@@ -126,11 +192,51 @@ def _apply_scalar_per_pixel(func, image, selem, out, mask, shift_x, shift_y,
...
+        Dimension of each pixel. Default value is 1.
```
sciunto:
```suggestion
        Dimension of each pixel.
```
We don't have to specify default values for pure python functions if no precision is needed.

skimage/filters/rank/generic.py
```diff
@@ -164,18 +270,18 @@ def func_out(image, selem=None, *args, **kwargs):
 def autolevel(image, selem, out=None, mask=None, shift_x=False, shift_y=False):
     """Auto-level image using local histogram.
 
-    This filter locally stretches the histogram of greyvalues to cover the
+    This filter locally stretches the histogram of grey values to cover the
```
sciunto:
```suggestion
    This filter locally stretches the histogram of gray values to cover the
```
We use US spelling.

skimage/filters/rank/generic.py
```diff
@@ -164,18 +270,18 @@ def func_out(image, selem=None, *args, **kwargs):
...
     entire range of values from "white" to "black".
 
     Parameters
     ----------
-    image : 2-D array (uint8, uint16)
+    image : 2-D array (integer, float or boolean)
         Input image.
-    selem : 2-D array
+    selem : 2-D array (integer, float or boolean)
```
sciunto: Please, add the missing ", optional" in this docstring.

sciunto: Can you do it also in the other docstrings below?

skimage/filters/rank/tests/test_rank.py
```diff
@@ -720,3 +691,19 @@ def test_majority(self):
...
+        elem = np.ones((3, 3), dtype=np.uint8)
+        rank.maximum(image=image, selem=elem, out=out, mask=mask)
+        assert_equal(image.dtype, out.dtype)
+
+    def test_input_boolean_dtype(self):
+        image = (np.random.rand(100, 100) * 256).astype(np.bool_)
+        out = np.empty_like(image)
+        mask = np.ones(image.shape, dtype=np.bool_)
+        elem = np.ones((3, 3), dtype=np.bool_)
+        # rank.maximum(image=image, selem=elem, out=out, mask=mask)
```
sciunto: WHy this is commented? (2 lines)

clementkng: This currently fails w/ a `TypeError: No matching signature found` in the `skimage/filters/rank/generic_cy.pyx` file. I suspect it's b/c I need to do something explicitly in the cython to support `np.bool_`, but can't figure out how to do that or if that is the actual cause.

clementkng: I have no idea on where to start for cython debugging, so any help is appreciated

clementkng: I have no idea on where to start for cython debugging, so any help is appreciated

skimage/filters/rank/tests/test_rank.py
```diff
@@ -41,8 +41,7 @@ def test_otsu_edge_case():
     assert result[1, 1] in [141, 172]
 
 
-
```
hmaarrfk: @clementkng do you have an editor that lets you pick out the lines of code you commit? I know you are "cleaning" this function, but this is something I consider a "flavour of the week" and has a chance to detract from your other contribution.

hmaarrfk:
@scikit-image/core how do we feel about programatically mangling docstrings? @clementkng made a herioic effort over 250 lines, but most of it could have been avoided if we do something with decorators.

lagru:
I'm not a fan. It's an elegant and "clever" way to avoid duplication but it comes at the cost of breaking IDE inspection tools and makes life harder for developers reading the docstring directly (faster in my experience) rather than having the HTML documentation open at all times. 

Concerning docstring parts that describe the in- and output of a function (Parameters, Returns) I'd argue that being explicit is worth it, especially because those type hints are used by IDEs. This "rule" would apply here.

Though I grant that other sections like notes or examples are more targeted at HTML documentation anyway and are better candidates for injection.

sciunto:
+1 with @lagru  I think we already had this discussion somewhere.

skimage/filters/rank/tests/test_rank.py
clementkng: I adopted this PR from @Henley13, who originally made these changes. I can try reversing them.

hmaarrfk: no not worth it.

skimage/filters/rank/generic.py
```diff
@@ -64,16 +63,54 @@
...
+
+    """
     assert_nD(image, 2)
-    if image.dtype not in (np.uint8, np.uint16):
+    if image.dtype not in (np.uint8, np.uint16, np.bool_):
         message = ('Possible precision loss converting image of type {} to '
                    'uint8 as required by rank filters. Convert manually using '
                    'skimage.util.img_as_ubyte to silence this warning.'
                    .format(image.dtype))
         warn(message, stacklevel=2)
```
hmaarrfk: This stack level should be 3.

hmaarrfk:
```suggestion
        warn(message, stacklevel=3)
```

hmaarrfk: Actually, it could be as high as 5!

hmaarrfk: To get the warnings at the correct level, you also need to remvoe the `default_selem` decorator.

```python
if selem is None:
        selem = ndi.generate_binary_structure(image.ndim, image.ndim)
```
and just that line where the decorator used to be used.

hmaarrfk:
Your test now assumes we support bool as a valid output data type. That is an interesting question.

bool is  rather a hybrid data type. The "cheapest" computational way to obtain the correct results for a bool array is to create a view of the array that is uint8. This is technically supported since `bool` is defined as being implemented with `uint8`

That said, you can get quite strange behavior:

```python
In [1]: import numpy as np                                                                               

In [2]: a = np.arange(5, dtype='uint8')                                                                  

In [3]: a.view('bool')                                                                                   
Out[3]: array([False,  True,  True,  True,  True])

In [4]: a.view('bool').view('uint8')                                                                     
Out[4]: array([0, 1, 2, 3, 4], dtype=uint8)

In [5]: a.view('bool') + 1                                                                               
Out[5]: array([1, 2, 2, 2, 2])
```

Regardless, I'll point you to where you have to do the change

```python
def _handle_input(image, selem, out, mask, out_dtype=None, pixel_size=1):
    [...]
    if out is None:
        if out_dtype is None:
            out_dtype = image.dtype
        out = np.empty(image.shape + (pixel_size,), dtype=out_dtype)
    else:
        if out.dtype == np.dtype(np.bool_):
            # So much hackery,
            # out = out.view(np.uint8)
            raise ValueError('out dtype cannot be bool.')
        if len(out.shape) == 2:
            out = out.reshape(out.shape+(pixel_size,))
```

skimage/filters/rank/generic.py
clementkng: I'm not sure which fields should be marked optional. Are you referring to all the `selem` fields or all the fields below?

skimage/filters/rank/generic.py
clementkng: When you say remove the decorator, do you mean to fully delete it (as in deleting the entire `default_selem` method) and Line 602 (`@default_elem`) as the line where the decorator used to be used?

clementkng:
Should I remove my test then, since supporting `bool` opens the door to this strange behavior?

skimage/filters/rank/generic.py
hmaarrfk: the ones that typically have `=None` where in the function call, they get assigned default values.

hmaarrfk:
I think it is OK.

We are using `out` as an empty array.

The strange behaviour comes when you do `our_bool_array.view('uint8')` and use that as an input array. That behaviour is undefined because any non-zero value evaluates to `True` or `1` for bool arrays, but the `view` might reveal otherwise.

IMO doing math on bool arrays with computers is only marginally beneficial. Virtually no computer architecture gives you any benefits.

skimage/filters/rank/generic.py
clementkng: By that logic, should the `selem` arg have a default `=None` in the function call as well?

clementkng:
I'm not sure I fully understand the nature of the fix, b/c when I incorporate it, I get a `TypeError: Function call with ambiguous argument types`. Just to be clear, we're coercing the `out.dtype` and `out_dtype` to be an `np.uint8` in `_handle_input` so operations are supported when we pass the arguments into the cython functions. I am able to confirm that both the `out.dtype` and `out_dtype` are `np.uint8` at the end of the function, so I'm still not sure why I'm still getting a `TypeError`.

skimage/filters/rank/generic.py
hmaarrfk: in some of them it seems so.

skimage/filters/rank/generic.py
hmaarrfk: yes, because the decorator adds a "stack level" that you would have to indicate in the function.

hmaarrfk:
hmm, yeah, i guess I had also had an other fix that I didn't know was necessary:

```python
    if image.dtype not in (np.uint8, np.uint16, np.bool_):
        message = ('Possible precision loss converting image of type {} to '
                   'uint8 as required by rank filters. Convert manually using '
                   'skimage.util.img_as_ubyte to silence this warning.'
                   .format(image.dtype))
        warn(message, stacklevel=5)

    if image.dtype not in (np.uint8, np.uint16):    # This is a new line.
        image = img_as_ubyte(image)
```

hmaarrfk:
Personally, I wouldn't bother with boolean types and just throw an error. We don't document it as an image dtype

https://scikit-image.org/docs/stable/user_guide/data_types.html
image

clementkng:
Sounds good, I'll change the unit test to reflect that.

hmaarrfk:
That said, I'm not too sure that that page reflects the current thinking among many of the core devs. I've seen quite a few PRs aimed at allowing uint8s to be treated as images in places. Maybe those were justs `masks` or `labels`.

clementkng:
But w/ no discussion for boolean types, I'm willing to pull back that functionality if there's really no gain as you mentioned above.

hmaarrfk:
i was just trying to help since you said there was an issue with bools. We've explored it, and for now, I think it is better to just leave it undefined. Setting a new behaviour takes alot more decision making power than what you originally anticiated doing. We can add the test and the error message later.

skimage/filters/rank/generic.py
```diff
@@ -203,24 +290,25 @@ def autolevel(image, selem, out=None, mask=None, shift_x=False, shift_y=False):
...
     This filter computes the morphological closing of the image and then
     subtracts the result from the original image.
 
     Parameters
     ----------
-    image : 2-D array (uint8, uint16)
+    image : 2-D array (integer, float or boolean)
         Input image.
-    selem : 2-D array
+    selem : 2-D array (integer, float, boolean or optional)
```
hmaarrfk: Bottom hat doesn't seem to specify default `selem`. Did you want to add a default to all rank filters? if so, if might be appropriate to put it in the `_handle_input` function.

skimage/filters/rank/tests/test_rank.py
```diff
@@ -713,10 +684,25 @@ def test_median_default_value(self):
...
+        image = (np.random.rand(100, 100) * 256).astype(np.uint8)
+        out = np.empty_like(image)
+        mask = np.ones(image.shape, dtype=np.uint8)
+        elem = np.ones((3, 3), dtype=np.uint8)
+        rank.maximum(image=image, selem=elem, out=out, mask=mask)
+        assert_equal(image.dtype, out.dtype)
+
+    def test_input_boolean_dtype(self):
+        image = (np.random.rand(100, 100) * 256).astype(np.bool_)
+        out = np.empty_like(image)
```
hmaarrfk: Just don't specify out here, it isn't part of the test, that said, don't even specify mask.

skimage/filters/rank/generic.py
```diff
@@ -63,17 +62,59 @@
...
+        The neighborhood expressed as a binary 2-D array.
+    out : 3-D array (same dtype out_dtype or as input)
+        Output array. The two first dimensions are the spatial ones, the third
+        one is the pixel vector (length 1 by default).
+    mask : 2-D array (np.uint8)
+        Mask array that defines (>0) area of the image included in the local
+        neighborhood.
+    n_bins : int
+        Number of histogram bins.
+    out_dtype : data-type
```
hmaarrfk: Did you mean to return this? It doesn't seem to be used anywhere.

hmaarrfk: Looking at the commit history, it seems that there was a point in time where it was useful to return this, now it no-longer seems to be the case. I was able to remove the output, refactor the code, and it passes all rank tests now

skimage/filters/rank/generic.py
```diff
@@ -63,17 +62,59 @@
            'entropy', 'otsu']
 
 
-def _handle_input(image, selem, out, mask, out_dtype=None, pixel_size=1):
+def _handle_input(image, selem, out=None, mask=None, out_dtype=None,
```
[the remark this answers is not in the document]
hmaarrfk: at least for `out` and `mask` that is true.

skimage/filters/rank/generic.py
```diff
@@ -63,17 +62,59 @@
...
+        Mask array that defines (>0) area of the image included in the local
+        neighborhood.
+    n_bins : int
+        Number of histogram bins.
+    out_dtype : data-type
+        Output data-type.
 
+    """
     assert_nD(image, 2)
+    if (image.dtype in (bool, np.bool, np.bool_)
```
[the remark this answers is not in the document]
hmaarrfk: Just note that the suggestion to have this raise an error might still be up for discussion. We might just have to issue a warning for now.

https://github.com/scikit-image/scikit-image/issues/3577#issuecomment-509004726

Or maybe just leave the behaviour as is.

skimage/filters/rank/generic.py
clementkng: I may have overdone it trying to address feedback above to add `optional` to the docstrings for `selem`. I'm not sure what's the best policy for this.

skimage/filters/rank/generic.py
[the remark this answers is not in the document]
clementkng: I think I'll hold on this change for now and wait for more discussion.

skimage/filters/rank/generic.py
clementkng: @hmaarrfk What is the best policy to add the defaults? I'm not sure what makes the most sense in this context.

hmaarrfk: Generally it comes in an as needed basis. People make requests as to what makes sense. It is often hard to impose a "most useful" default and one that makes the "most untiuitive sense"

clementkng: Ok, I'm going to try adding them into `_handle_input` then.

clementkng:
Hmm, I tried rebasing to pick up the Tifffile fix, but I'm not exactly sure what happened.

sciunto:
@soupault Would you like to finalize your review? This looks good to me and it would be great to merge this soon.

hmaarrfk:
this one is really annoy8ing to fix. You will want to rebase, keep your changes, than convert al lthe assert_nD to check_nD

sciunto:
@scikit-image/core Who wants to prove that s/he has super rebasing skills? This PR needs a rebase and must be ready for 0.16.

lagru:
@sciunto Seeing that this PR already contains a merge commit from master wouldn't another one be the best option to resolve the conflicts?

jni:
@clementkng, @hmaarrfk did a nice rebase of these changes in #4289. All your contributions are nicely preserved there, and it's almost ready to go in. In the meantime, we'll close this one. Future discussion of this issue should happen in #4289

jni:
@clementkng I forgot to say: thank you!!

PR #3983, opened 2019-06-28
[New feature] Add TV-L1 optical flow algorithm

rfezzani:
## Description

As I previously proposed on the mailing list (sorry for the delay), this PR is about the implementation of the TV-L1 algorithm for the estimation of optical flow **[1]**.

I integrated it in the sub-package `skimage.registration` because I did not found a better place for it. I hope that there is no conflict with the Meta issue #3810. 

Concerning the tests, I actually implemented only the zero motion case. I also can implement a non regression test by saving the output of the algorithm in the `data` sub-package.

For the examples, I used the `stereo_motorcycle` sequence that is already included in skimage, but I can also add any of the train sequences from the [middlebury dataset](http://vision.middlebury.edu/flow/) : I already asked the maintainer of the project if there is any copyright restriction concerning the distribution of the Middlebury dataset, and there is not.

**[1]**: Zach, C., Pock, T., & Bischof, H. (2007, September). A duality based approach for realtime TV-L 1 optical flow. In Joint pattern recognition symposium (pp. 214-223). Springer, Berlin, Heidelberg.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


jni:
@rfezzani thanks for this contribution! I apologise that I don't have time to do a proper review right now, but one concern I have is that we are aiming to be a fully n-dimensional library, minimising the amount of 2D-only code as much as possible. From what I can tell, this algorithm generalises quite readily to n-dimensions. Do you think that would be something you would be comfortable doing? Otherwise we can certainly take over at this point!

hmaarrfk:
@jni do we really want to "ndify" the whole library at the start of every contribution? This seems quite intense and I feel like it make make the higher dimension stuff rushed.

jni:
@hmaarrfk 

> do we really want to "ndify" the whole library at the start of every contribution?

yes. Anything else is taking on debt that we generally can't afford to pay back. I put merging a non-nD request (except where the feature is intrinsically 2D) on almost the same level as merging something without tests with the idea that we'll add them later. The truth is that none of us have time for that. A large proportion of my work on the library for the past 6 years has been patching 2D-only stuff to support nD.

This is not to say that the contributor needs to do it, which is why I offered to take over. But I think the 2D tests should be passing before I do. (Again, no time for a full review right now, but if no one else comes in before then, I can help after the SciPy conference.)

JDWarner:
While I am also firmly in the belief that n-D is the right path forward, it does result in us not having nice things, or getting them significantly late at times.  As just one example, I had perfectly functional and performant 2D code for flood filling in 2015 - simple Cython, still have the gist published - but we wanted it n-D.  Flood fill itself didn't get merged until 3 years later.  There's a lot we could have done with a 2D or 2D/3D version in the meantime.  I'm mostly playing the Devil's Advocate here, but food for thought.

jni:
@JDWarner :wave: and looking forward to seeing you next week! =) =) =)

The flip side of that coin is that we have nD flood fill now, which I believe might not be true if we'd had 2D already. Not only that, but I think the code ended up much neater in the end. In the current stage of scikit-image's development, I think long-term maintainability is a much greater concern than short term development speed, and that balance is only going to become more lopsided.

2D implementations can live in branches, gists, and WIP PRs and still be useful to the community while nD support is worked out...

rfezzani:
Hi all and thanks for your feedback.
Optical flow in n-D looks affordable, but it needs much more work and even more validation. I even don't know any data set or challenge concerning volumetric optical flow.
I don't believe that I will be able to develop it in the near future :-(.

hmaarrfk:
Is there a just middle we can hit? Where we ensure the API is compatible with ND, but not actually implement it until followup PRs?

I think the most valuable aspect is being able to contribute a small useful chunk at the same time. That ensures that we don't loose interest from the many parties involved in a contribution.

The desire to go to ND has for example made it difficult to move the color module over because it is hard to ensure backward compatibility with the majic that was decided long ago.

For example, I was able to do something like this in numpy where I was able to improve the np.block algorithm without having to write it in C. For a certain usecase, it was actually slower than the naive implementation. It was decided that python workarounds (falling back on the previous implementation) were allowed and that the C Port could be delayed until further notice.

jni:
> Is there a just middle we can hit? Where we ensure the API is compatible with ND, but not actually implement it until followup PRs?

This might be possible. Already this PR doesn't meet that bar (at first glance) because it uses warp, which uses reversed coordinates, again, partly because warp assumed 2D images when it was written long ago.

> The desire to go to ND has for example made it difficult to move the color module over because it is hard to ensure backward compatibility with the majic that was decided long ago.

For me this makes the complete opposite point: moving forward has been extremely difficult because of code that assumed 2D inputs. There's many examples like this.

> I think the most valuable aspect is being able to contribute a small useful chunk at the same time. That ensures that we don't loose interest from the many parties involved in a contribution.

This is definitely a huge concern, and thanks for keeping it in focus! But it is one among several. My proposed solution to "fast contribution experiences" is that maintainers should help contributors by pushing commits to the PR branch, *not* that we lower our standards for long-term maintainability.

@rfezzani sorry that the discussion has gone off track a little bit. I just saw that your tests are passing, so at this point you can take a break. =) Thanks for your contribution! We'll do some review, maybe argue a little bit =P, and then help push it to nD, but again, from me at least this has to happen after SciPy.

rfezzani:
>This might be possible. Already this PR doesn't meet that bar (at first glance) because it uses warp, which uses reversed coordinates, again, partly because warp assumed 2D images when it was written long ago.

`warp` does not assume 2D images, here is a perfectly working demo toy:

```python
from skimage.transform import warp
import numpy as np
img = np.random.randn(256, 256, 256)
I, J, K = np.meshgrid(np.arange(256), np.arange(256), np.arange(256))
img = warp(img, np.array([I, J, K]))
```

I may be don't understood what you mean, but even if `warp` was not supporting nD, this doesn't prevent the API from being compatible with nD as suggested by @hmaarrfk.

> @rfezzani sorry that the discussion has gone off track a little bit.

Don't worry @jni, this discussion is a big concern: is the nD support a condition for the inclusion of a new feature in skimage? In my opinion, this is a huge requirement, difficult to reach in many cases.

By the way, listing all the actual skimage functions only supporting 2D and creating the corresponding issues for extending their support to nD can be interesting. What do you think?

rfezzani:
Hi everyone,
after some efforts, I finally managed to implement the nD support for the TV-L1 algorithm ;-).
I would be happy if someone finds some time to review my code. Thx in advance.

doc/examples/registration/plot_opticalflow.py
```diff
@@ -0,0 +1,78 @@
...
+from skimage.color import rgb2gray
+from skimage.data import stereo_motorcycle
+from skimage.transform import warp
+from skimage.registration import tvl1
+
+# --- Load the sequence
+I0, I1, disp = stereo_motorcycle()
+
+# --- Convert the images to gray level: color is not supported.
+I0 = rgb2gray(I0)
```
jni: Rename these "image0" and "image1" for consistency with PEP8

doc/examples/registration/plot_opticalflow.py
```diff
@@ -0,0 +1,78 @@
...
+I1 = rgb2gray(I1)
+
+# --- Compute the optical flow
+v, u = tvl1(I0, I1)
+
+# --- Use the estimated optical flow for registeration
+
+nl, nc = I0.shape
```
jni: nr, nc (rows/columns)

doc/examples/registration/plot_opticalflow.py
```diff
@@ -0,0 +1,78 @@
...
+
+y, x = np.meshgrid(np.arange(nl), np.arange(nc), indexing='ij')
```
jni: Here I suggest row_coords, col_coords

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+        if (flow0*flow0).sum() < tol:
+            break
+
+        flow0 = flow
+
+    return flow
+
+
+def tvl1(I0, I1, dt=0.2, lambda_=15, tau=0.3, nwarp=5, niter=10,
+         tol=1e-4, prefilter=False):
```
jni: So, after reading through all the code, I still don't know what "TV-L1" stands for, and I suspect most people won't either. I suggest that we rename this function to "optical_flow", then specify in the docstring as an implementation detail that we use the TV-L1 solver internally.

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+    """Coarse to fine TV-L1 optical flow estimator.
+
+    TV-L1 ia popular algorithm for optical flow estimation intrudced
+    by Zack et al. [1]_, improved in [2]_ and detailed in [3]_.
+
+    Parameters
+    ----------
+    I0 : ~numpy.ndarray
```
jni: Just "array" or "ndarray" (no numpy, and definitely no tilde~)

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+        flow0 = flow
+
+    return flow
+
+
+def tvl1(I0, I1, dt=0.2, lambda_=15, tau=0.3, nwarp=5, niter=10,
+         tol=1e-4, prefilter=False):
+    """Coarse to fine TV-L1 optical flow estimator.
+
+    TV-L1 ia popular algorithm for optical flow estimation intrudced
```
jni:
```suggestion
    TV-L1 is a popular algorithm for optical flow estimation introduced
```

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+    Parameters
+    ----------
+    I0 : ~numpy.ndarray
+        The first gray scale image of the sequence.
+    I1 : ~numpy.ndarray
+        The second gray scale image of the sequence.
+    dt : float
+        Time step of the numerical scheme. Convergence is proved for
+        values dt < 0.125, but it can be larger for faster
+        convergence.
```
jni: Should we use a default value in the proven convergence regime? =)

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+        flow0 -= flow
+        if (flow0*flow0).sum() < tol:
+            break
+
+        flow0 = flow
+
+    return flow
+
+
+def tvl1(I0, I1, dt=0.2, lambda_=15, tau=0.3, nwarp=5, niter=10,
```
jni: Since we are in Python 3.5+, use `def optical_flow(image0, image1, *, dt=0.2, ...)` to indicate that arguments after image1 *must* be given as keywords.

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+    I0 : ~numpy.ndarray
+        The first gray scale image of the sequence.
+    I1 : ~numpy.ndarray
+        The second gray scale image of the sequence.
+    dt : float
+        Time step of the numerical scheme. Convergence is proved for
+        values dt < 0.125, but it can be larger for faster
+        convergence.
+    lambda_ : float
+        Attachement parameter. The smaller this parameter is,
```
jni:
```suggestion
        Attachment parameter. The smaller this parameter is,
```

I wonder if we should name these "attachment" and "tightness", and specify in the docstring that they correspond to "lambda" and "tau" from the paper. What do you think?

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+        the smoother is the solutions.
```
jni:
```suggestion
        the smoother the found solution will be.
```

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+    tau : float
+        Tightness parameter. It should have a small value in order to
+        maintain attachement and regularization parts in
```
jni:
```suggestion
        maintain attachment and regularization parts in
```

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+        correspondence.
+    nwarp : int
+        Number of times I1 is warped.
+    niter : int
+        Number of fixed point iteration.
+    tol : float
+        Tolerance used as stopping criterion based on the L² distance
+        between two consecutive values of (u, v).
+    prefilter : bool
+        whether to prefilter the estimated optical flow before each
```
jni:
```suggestion
        Whether to prefilter the estimated optical flow before each
```

A few more words about this parameter would be nice (when is it useful?)

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+        image warp.
+
+    Returns
+    -------
+    flow : tuple[~numpy.ndarray]
```
jni: tuple of arrays

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+        The estimated optical flow.
```
jni: The estimated optical flow *for each axis*?

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+    g = np.zeros((I0.ndim, ) + I0.shape)
+    proj = np.zeros((I0.ndim, I0.ndim, ) + I0.shape)
+
+    s_g = [slice(None), ] * g.ndim
+    s_p = [slice(None), ] * proj.ndim
+    s_d = [slice(None), ] * (proj.ndim-2)
+
+    for _ in range(nwarp):
+        if prefilter:
+            flow = ndi.filters.median_filter(flow, [1]+I0.ndim*[3])
```
jni: You can just use `ndi.median_filter` directly (no need for `.filters`)

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+            flow_ = flow
+
+            flow_[:, idx] -= rho[idx]*grad[:, idx]/NI[idx]
+
+            idx = ~idx
+            srho = f0 * np.sign(rho[idx])
+            flow_[:, idx] -= srho*grad[:, idx]
+
+            # Regularization term
+            flow = flow_.copy()
```
jni: I'm a bit confused by `flow`, `flow0`, `flow_`. Could you name these more verbosely? e.g. `flow_reference`, `flow_current`, `flow_update`?

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+
+            for idx in range(flow.shape[0]):
+                s_p[0] = idx
+                for _ in range(2):
```
jni: Why is this block run twice exactly?

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+
+from functools import partial
+import numpy as np
+from scipy import ndimage as ndi
+from skimage.transform import warp
+
+from .utils import coarse_to_fine
+
+
+def _tvl1(I0, I1, flow0, dt, lambda_, tau, nwarp, niter, tol, prefilter):
```
jni: As mentioned below, use `*` after `image1` to make later arguments keyword-only.

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+    """TV-L1 solver for optical flow estimation.
+
+    Parameters
+    ----------
+    I0 : ~numpy.ndarray
+        The first gray scale image of the sequence.
+    I1 : ~numpy.ndarray
+        The second gray scale image of the sequence.
+    flow0 : ~numpy.ndarray
+        Vector field initialization.
```
jni: You can write array shapes here, eg "image1 : array, shape (M, N[, P[, ...]])", then "flow0 : array, shape (D, M, N[, P[, ...]])"

skimage/registration/_tvl1.py
jni: Actually it's not a tuple here, it's an array, shape D, M, N[, ...].

skimage/registration/_tvl1.py
```diff
@@ -0,0 +1,206 @@
...
+    >>> # --- Convert the images to gray level: color is not supported.
+    >>> I0 = rgb2gray(I0)
+    >>> I1 = rgb2gray(I1)
+    >>> flow = tvl1(I1, I0)
+
+    """
+
+    solver = partial(_tvl1, dt=dt, lambda_=lambda_, tau=tau,
+                     nwarp=nwarp, niter=niter, tol=tol,
+                     prefilter=prefilter)
```
jni: I love this functional approach. =)

skimage/registration/_tvl1.py
alexdesiqueira: I wonder if it'd be possible to have more "explanatory" variable names here and throughout the code. For instance, if `I0` is the first grayscale image, couldn't we use something as `first_image`, or `first_frame` (as in the paper? This way, we would have `second_image`, `vector_init`, `step`, `attach_par`, and `tight_par. It'd be very valuable for someone who doesn't know the algorithm (myself included) :)

skimage/registration/_tvl1.py
alexdesiqueira: Could we set a default for `flow0`? The same for the other variables (except the images). It would be easier to test the function if there's some default values around.

skimage/registration/_tvl1.py
alexdesiqueira: I second that :)

skimage/registration/_tvl1.py
alexdesiqueira: Could this function be named more relevantly as well? Something like `optical_flow_tvl1`, maybe.

skimage/registration/_tvl1.py
jni: imho we should go for `image0` and `image1` or `reference` and `target`/`moving` to be consistent with other registration methods.

Also, no need to add a `_par` suffix to parameters! `attachment`, `tightness` is sufficient.

skimage/registration/_tvl1.py
jni: I think `optical_flow` by itself is enough. Optical flow is the name of what we are computing. TV-L1 is an implementation detail. If we add other methods in the future, we can add a `method=` keyword argument.

skimage/registration/_tvl1.py
rfezzani: The proof of convergence is provided in Chambolle's paper [[1]](https://www.math.ucla.edu/~lvese/285j.1.09f/Chambolle.pdf), but in the article he commented that "_In practice, it appears that the optimal constant for the stability and convergence of the algorithm is not 1/8 but 1/4. We do not know the reason for this_".
In `skimage.restoration._denoise._denoise_tv_chambolle_nd`, this parameter is defined as a variable `tau = 1. / (2.*ndim)`. We can opt for this solution if you agree and remove this confusing sentence.

skimage/registration/_tvl1.py
rfezzani: I would vote for `optical_flow_tvl1` since if we implement an other algorithm for optical flow, the function's parameters will be different.

skimage/registration/_tvl1.py
rfezzani: I opted for `...the smoother the returned result will be`. What do you think?

skimage/registration/_tvl1.py
rfezzani: It should better be `The estimated optical flow components for each axis`. What do you think?

skimage/registration/_tvl1.py
rfezzani: I defined `flow_current`, `flow_previous`, `flow_axilliary` and kept `flow0` for the initial flow. Is it OK for you?

skimage/registration/_tvl1.py
rfezzani: You are right. This is much like a parameter. In the original paper [1] it is set to 1 and in the improved version [2] it is set to 5. In this revision I set it as a variable in the code `reg_niter`, do you think that I should expose it as a parameter of the function?

skimage/registration/_tvl1.py
rfezzani: `_tvl1` is supposed to be hidden and only called internally as a routine, so I think that there is no need for such precautions.

skimage/registration/_tvl1.py
rfezzani: @alexdesiqueira, `flow0` is updated  in the coarse to fine schema: at each scale move the flow estimated at small resolution is rescaled to the higher resolution and used as initialization.
Zeros is a reasonable default initial value for  `flow0`  at coarsest level.

skimage/registration/_tvl1.py
rfezzani: Thx, ;-)

skimage/registration/_tvl1.py
rfezzani: This option is proposed in the improved version of the algorithm [2]. It is not really necessary if the attachment parameter is low enough to produce a smooth result.

rfezzani:
Thank you both, @jni and @alexdesiqueira, for your review and comments. The code looks much better now.
Concerning the tests, I agree with you @jni, it is far from being perfect. I don't have any satisfactory idea of tests to implement. As I mentioned in the PR description, I can implement a non regression test by saving the output of the algorithm in the data sub-package, but I found no other option.
There is effectively some 2D datasets providing ground truth solutions, [middlebury](http://vision.middlebury.edu/flow/) is one of them and I already asked its maintainer, and there is no restriction to the distribution of any sequence from this dataset.
The `stereo_motorcycle` sequence that is already included in `skimage` comes from the stereo challenge of _Middlebury_, What about adding a sequence from the optical flow challenge of _Middlebury_?
I unfortunately never heard about any volumetric dataset for optical flow problem, and I don't really know how to test it :-(.

skimage/registration/tests/test_tvl1.py
```diff
@@ -0,0 +1,29 @@
...
+def test_no_motion_2d():
+    rnd = np.random.RandomState(0)
+    img = rnd.normal(size=(256, 256))
+
+    flow = optical_flow_tvl1(img, img)
+
+    assert np.all(flow == 0)
+
+
+def test_no_motion_3d():
```
jni: Great! This will at least exercise the code. =)

Another test would be to use ndimage transforms that are expected to produce a particular pattern of flow. For example, wouldn't shrinking the image about its center produce a radially shrinking field with predictable flow? Or, more simply, a translation of x, y, z pixels should produce that flow uniformly across the image, right?

rfezzani: Yes, good suggestion, but the resulting optical flow will not be equal to the ground truth, all we can do is expect a result reasonably close to the true solution. I will give this a try ASAP ;-)

rfezzani:
Hi @jni, I finally managed to add non trivial tests for the optical flow estimation. Is it OK for you?
Thx again.

skimage/registration/_optical_flow.py
```diff
@@ -0,0 +1,215 @@
...
+
+    return flow_current
+
+
+def optical_flow_tvl1(image0, image1, *, attachment=15, tightness=0.3,
+                      nwarp=5, niter=10, tol=1e-4, prefilter=False,
+                      dtype='float32'):
+    """Coarse to fine optical flow estimator.
+
+    The TV-L1 solver is applyed at each level of the image
```
jni:
```suggestion
    The TV-L1 solver is applied at each level of the image
```

skimage/registration/_optical_flow.py
```diff
@@ -0,0 +1,215 @@
...
+    pyramid. TV-L1 is a popular algorithm for optical flow estimation
+    introduced by Zack et al. [1]_, improved in [2]_ and detailed in [3]_.
+
+    Notes
```
jni: This section should be moved to be after "returns"

skimage/registration/_optical_flow.py
```diff
@@ -0,0 +1,215 @@
...
+
+        flow_previous = flow_current
+
+    return flow_current
+
+
+def optical_flow_tvl1(image0, image1, *, attachment=15, tightness=0.3,
+                      nwarp=5, niter=10, tol=1e-4, prefilter=False,
+                      dtype='float32'):
+    """Coarse to fine optical flow estimator.
```
jni:
```suggestion
    r"""Coarse to fine optical flow estimator.
```

This is needed because the docstring contains math (`\lambda` etc), which can be ambiguous in non-raw strings.

skimage/registration/_optical_flow.py
```diff
@@ -0,0 +1,215 @@
...
+"""TV-L1 optical flow algorithm implementation.
+
+"""
+
+from functools import partial
+import numpy as np
+from scipy import ndimage as ndi
+from skimage.transform import warp
+
+from .utils import coarse_to_fine
```
jni: These should be in a private module `_utils` to make it clear that they should not be used publicly. I also might call the module `_optical_flow_utils` to make it clearer what the file is when browsing the repo.

rfezzani:
Thank you again @jni for your review, this PR is much better now then its first implementation ;-).

alexdesiqueira:
Hey @rfezzani, I really think this could have better name variables, once I read it and don't understand that much yet. However, how I liked the algorithm! This seems a lot useful (and I'm seeing applications for that right now). But @jni pointed several points and you dealed with them all, therefore I won't hold it.
Thank you very much for your hard work, we really appreciate it!

rfezzani:
Hi @alexdesiqueira, thank you again for your review. I am really happy if this new feature can help you! I am sorry if you feel that variable names can be more explicit, this may be addressed in a documentation improvement issue? Let me know if you have any idea ;-)

alexdesiqueira:
Hey @rfezzani,
I like to think that the code by itself should be as readable as it could be, without referring too much on documentation on a first moment (no one reads the manual, right? :smile:).
Thanks again for your work!

Issue #3974, opened 2019-06-23
Proposal to speed up skimage.feature.peak_local_max


brownmk:
## Description
The current implementation of peak_local_max is slow, when labels argument is provided.
The reason is when it loops over each object in the labels image, local peaks are searched again using the mask of the object of interest.  The input image used for each recursive peak search uses an image the same size as the original image, which is unnecessary as all the pixels outside the object are zero.

This can be extremely slow when there are hundreds of objects and/or the input image is 3D.
I proposal to only pass a small image enclosing each object to the recursive call, the modified code attached (peak.txt) is shown to provide a significant speed up.

## Way to reproduce
```python
# Place the full code we need to recreate your issue here

import numpy as np
import time
import warnings
warnings.simplefilter("ignore")

from scipy import ndimage as ndi
from skimage.io import imsave
from skimage.feature import peak_local_max
from peak import peak_local_max as peak_local_max_v2

mask=np.zeros([500,500], dtype=bool)
x,y=np.indices((500,500))
x_c=x//20*20+10
y_c=y//20*20+10
mask[(x-x_c)**2+(y-y_c)**2<8**2]=True
# create a mask, label each disk, create distance image for peak searching
labels, num_objs=ndi.label(mask)
dist=ndi.distance_transform_edt(mask)
print("Number of objects: {}".format(num_objs))
imsave("mask.png", mask.astype(np.uint8)*255)
imsave("label.png", labels.astype(np.uint16)*80)
imsave("dist.png", dist.astype(np.uint8)*31)

start_time = time.time()
local_max=peak_local_max(dist, min_distance=20, indices=False, exclude_border=False, labels=labels)
print("--- %s seconds ---" % (time.time() - start_time))
print(local_max.dtype)
imsave("max.png", local_max.astype(np.uint8)*255)

start_time = time.time()
local_max2=peak_local_max_v2(dist, min_distance=20, indices=False, exclude_border=False, labels=labels)
print("--- %s seconds ---" % (time.time() - start_time))
print("Find the same set of maxima?", np.array_equal(local_max, local_max2))

# upload all necessary images to github too!
```
The above code "example.txt" first constructs a mask "mask.png", labels each disk as "label.png", derives a distance map "dist.png" from the mask, it then uses label.png to help identify the center of each disk.

Although labels for peak_local_max is not important for this made-up example, it is here to illustrate performance issues encountered when labels argument is used.

example.txt contains the test code pasted above.
[...]

PR #3984, opened 2019-06-30
Speed up skimage.feature.peak_local_max

clementkng:
## Description

@brownmk's proposal to modify the `peak_local_max` algorithm to recursively pass a small image enclosing each object rather than an image of the same size, improving algorithm performance.

Fixes gh-3974

## Checklist

~- [ ] Gallery example in `./doc/examples` (new features only)~
  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -44,7 +44,7 @@ def _get_peak_mask(image, min_distance, footprint, threshold_abs,
 
 def _exclude_border(mask, footprint, exclude_border):
     """
-    Remove peaks round the borders
+    Remove peaks near the borders
     """
     # zero out the image borders
     for i in range(mask.ndim):
@@ -189,10 +189,16 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
             mask[nd_indices] = True
             out[obj] += mask
 
+        if not indices and np.isinf(num_peaks):
+            return out
+
         coordinates = _get_high_intensity_peaks(image, out, num_peaks)
-        if indices is True:
+        if indices:
             return coordinates
         else:
+            out.fill(False)
+            nd_indices = tuple(coordinates.T)
+            out[nd_indices] = True
             return out
 
     # Non maximum filter
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -25,7 +25,7 @@ def _get_high_intensity_peaks(image, mask, num_peaks):
 def _get_peak_mask(image, min_distance, footprint, threshold_abs,
                    threshold_rel):
     """
-    Return the mask containing all peak candidates above thresholds
+    Return the mask containing all peak candidates above thresholds.
     """
     if footprint is not None:
         image_max = ndi.maximum_filter(image, footprint=footprint,
@@ -44,7 +44,7 @@ def _get_peak_mask(image, min_distance, footprint, threshold_abs,
 
 def _exclude_border(mask, footprint, exclude_border):
     """
-    Remove peaks round the borders
+    Remove peaks round the borders.
     """
     # zero out the image borders
     for i in range(mask.ndim):
@@ -152,9 +152,6 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
     array([[10, 10, 10]])
 
     """
-    if type(exclude_border) == bool:
-        exclude_border = min_distance if exclude_border else 0
-
     out = np.zeros_like(image, dtype=np.bool)
 
     # no peak for a trivial image
@@ -166,8 +163,10 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
 
     threshold_abs = threshold_abs if threshold_abs is not None else image.min()
 
-    # In the case of labels, recursively build and return an output
-    # operating on each label separately
+    if type(exclude_border) == bool:
+        exclude_border = min_distance if exclude_border else 0
+
+    # In the case of labels, call ndi on each label
     if labels is not None:
         label_values = np.unique(labels)
         # Reorder label values to have consecutive integers (no gaps)
@@ -181,10 +180,8 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
             inner_mask = _exclude_border(np.ones_like(labels, dtype=bool),
                                          footprint, exclude_border)
 
-        for i, obj in enumerate(ndi.find_objects(labels)):
-            label = i + 1
-            # print("Label>", label)
-            img = image[obj] * (labels[obj] == label)
+        for label_idx, obj in enumerate(ndi.find_objects(labels)):
+            img = image[obj] * (labels[obj] == label_idx + 1)
             mask = _get_peak_mask(img, min_distance, footprint, threshold_abs,
                                   threshold_rel)
             if exclude_border:
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -158,6 +158,13 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
     if type(exclude_border) == bool:
         exclude_border = min_distance if exclude_border else 0
 
+    # no peak for a trivial image
+    if np.all(image == image.flat[0]):
+        if indices is True:
+            return np.empty((0, 2), np.int)
+        else:
+            return out
+
     # In the case of labels, call ndi on each label
     if labels is not None:
         label_values = np.unique(labels)
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -180,14 +180,17 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
             inner_mask = _exclude_border(np.ones_like(labels, dtype=bool),
                                          footprint, exclude_border)
 
+        # For each label, extract a smaller image enclosing the object of
+        # interest, identify num_peaks_per_label peaks and mark them in
+        # variable out.
         for label_idx, obj in enumerate(ndi.find_objects(labels)):
-            img = image[obj] * (labels[obj] == label_idx + 1)
-            mask = _get_peak_mask(img, min_distance, footprint, threshold_abs,
+            img_object = image[obj] * (labels[obj] == label_idx + 1)
+            mask = _get_peak_mask(img_object, min_distance, footprint, threshold_abs,
                                   threshold_rel)
             if exclude_border:
                 # remove peaks fall in the exclude region
                 mask &= inner_mask[obj]
-            coordinates = _get_high_intensity_peaks(img, mask,
+            coordinates = _get_high_intensity_peaks(img_object, mask,
                                                     num_peaks_per_label)
             nd_indices = tuple(coordinates.T)
             mask.fill(False)
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -185,8 +185,8 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
         # variable out.
         for label_idx, obj in enumerate(ndi.find_objects(labels)):
             img_object = image[obj] * (labels[obj] == label_idx + 1)
-            mask = _get_peak_mask(img_object, min_distance, footprint, threshold_abs,
-                                  threshold_rel)
+            mask = _get_peak_mask(img_object, min_distance, footprint,
+                                  threshold_abs, threshold_rel)
             if exclude_border:
                 # remove peaks fall in the exclude region
                 mask &= inner_mask[obj]
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -22,6 +22,39 @@ def _get_high_intensity_peaks(image, mask, num_peaks):
     return coord[::-1]
 
 
+def _get_peak_mask(image, min_distance, footprint, threshold_abs, threshold_rel):
+    """
+    Return the mask containing all peak candidates above thresholds
+    """
+    if footprint is not None:
+        image_max = ndi.maximum_filter(image, footprint=footprint,
+                                       mode='constant')
+    else:
+        size = 2 * min_distance + 1
+        image_max = ndi.maximum_filter(image, size=size, mode='constant')
+    mask = image == image_max
+    if threshold_rel is not None:
+        threshold=max(threshold_abs, threshold_rel * image.max())
+    else:
+        threshold=threshold_abs
+    mask &= image > threshold
+    return mask
+
+
+def _exclude_border(mask, footprint, exclude_border):
+    """
+    Remove peaks round the borders
+    """
+    # zero out the image borders
+    for i in range(mask.ndim):
+        mask = mask.swapaxes(0, i)
+        remove = (footprint.shape[i] if footprint is not None
+                  else 2 * exclude_border)
+        mask[:remove // 2] = mask[-remove // 2:] = False
+        mask = mask.swapaxes(0, i)
+    return mask
+
+
 def peak_local_max(image, min_distance=1, threshold_abs=None,
                    threshold_rel=None, exclude_border=True, indices=True,
                    num_peaks=np.inf, footprint=None, labels=None,
@@ -123,6 +156,15 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
 
     out = np.zeros_like(image, dtype=np.bool)
 
+    # no peak for a trivial image
+    if np.all(image == image.flat[0]):
+        if indices is True:
+            return np.empty((0, 2), np.int)
+        else:
+            return out
+
+    threshold_abs = threshold_abs if threshold_abs is not None else image.min()
+
     # In the case of labels, recursively build and return an output
     # operating on each label separately
     if labels is not None:
@@ -134,19 +176,35 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
         labels = labels.astype(np.int32)
 
         # New values for new ordering
-        label_values = np.unique(labels)
-        for label in label_values[label_values != 0]:
-            maskim = (labels == label)
-            out += peak_local_max(image * maskim, min_distance=min_distance,
-                                  threshold_abs=threshold_abs,
-                                  threshold_rel=threshold_rel,
-                                  exclude_border=exclude_border,
-                                  indices=False, num_peaks=num_peaks_per_label,
-                                  footprint=footprint, labels=None)
+        # for label in label_values[label_values != 0]:
+        #    maskim = (labels == label)
+        #    out += peak_local_max(image * maskim, min_distance=min_distance,
+        #                          threshold_abs=threshold_abs,
+        #                          threshold_rel=threshold_rel,
+        #                          exclude_border=exclude_border,
+        #                          indices=False, num_peaks=num_peaks_per_label,
+        #                          footprint=footprint, labels=None)
+        # modify to speed it up, as the original code passes the full image for each label
+        # too slow, more so for 3D images and many objects
+        if exclude_border:
+            # create a mask for the non-exclude region
+            inner_mask = _exclude_border(np.ones_like(labels, dtype=bool), footprint, exclude_border)
 
-        # Select highest intensities (num_peaks)
+        for i, obj in enumerate(ndi.find_objects(labels)):
+            label = i + 1
+            # print("Label>", label)
+            img = image[obj] * (labels[obj] == label)
+            mask = _get_peak_mask(img, min_distance, footprint, threshold_abs, threshold_rel)
+            if exclude_border:
+                # remove peaks fall in the exclude region
+                mask &= inner_mask[obj]
+            coordinates = _get_high_intensity_peaks(img, mask, num_peaks_per_label)
+            nd_indices = tuple(coordinates.T)
+            mask.fill(False)
+            mask[nd_indices] = True
+            out[obj] += mask
+
         coordinates = _get_high_intensity_peaks(image, out, num_peaks)
-
         if indices is True:
             return coordinates
         else:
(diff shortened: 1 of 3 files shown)


sciunto:
@clementkng Thanks for submitting. Is it still a WIP? I see that all the tests are passing...

clementkng:
@sciunto Hi, I believe it may still be a WIP b/c the algorithm has changed a bit since to pass the tests, so I'm not sure if the performance gain is still there. Also, I'm considering putting in new tests that @brownmk used initially to test the code. Or do you think there's enough coverage w/ the existing tests?

skimage/feature/peak.py
```diff
@@ -123,6 +157,15 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
     out = np.zeros_like(image, dtype=np.bool)
 
+    # no peak for a trivial image
+    if np.all(image == image.flat[0]):
+        if indices is True:
+            return np.empty((0, 2), np.int)
+        else:
+            return out
+
+    threshold_abs = threshold_abs if threshold_abs is not None else image.min()
```
sciunto: Minor: I would move the block "if type(exclude_border)..." few lines above just before or after this line, to gather these lines which have the same spirit.

skimage/feature/peak.py
```diff
@@ -133,60 +176,41 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
-        # Select highest intensities (num_peaks)
-        coordinates = _get_high_intensity_peaks(image, out, num_peaks)
+        if exclude_border:
+            # create a mask for the non-exclude region
+            inner_mask = _exclude_border(np.ones_like(labels, dtype=bool),
+                                         footprint, exclude_border)
+
+        for i, obj in enumerate(ndi.find_objects(labels)):
+            label = i + 1
+            # print("Label>", label)
```
sciunto: This line can be removed.

skimage/feature/peak.py
```diff
@@ -22,6 +22,40 @@ def _get_high_intensity_peaks(image, mask, num_peaks):
...
+        threshold = max(threshold_abs, threshold_rel * image.max())
+    else:
+        threshold = threshold_abs
+    mask &= image > threshold
+    return mask
+
+
+def _exclude_border(mask, footprint, exclude_border):
+    """
+    Remove peaks round the borders
```
sciunto:
```suggestion
    Remove peaks round the borders.
```

skimage/feature/peak.py
```diff
@@ -22,6 +22,40 @@ def _get_high_intensity_peaks(image, mask, num_peaks):
     return coord[::-1]
 
 
+def _get_peak_mask(image, min_distance, footprint, threshold_abs,
+                   threshold_rel):
+    """
+    Return the mask containing all peak candidates above thresholds
```
sciunto:
```suggestion
    Return the mask containing all peak candidates above thresholds.
```

skimage/feature/peak.py
```diff
@@ -133,60 +176,41 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
-
-        # Select highest intensities (num_peaks)
-        coordinates = _get_high_intensity_peaks(image, out, num_peaks)
+        if exclude_border:
+            # create a mask for the non-exclude region
+            inner_mask = _exclude_border(np.ones_like(labels, dtype=bool),
+                                         footprint, exclude_border)
+
+        for i, obj in enumerate(ndi.find_objects(labels)):
+            label = i + 1
```
sciunto: This variable seems to be unused... (Am i correct?)

brownmk:
Hi, The performance gain is still there.  If you want to add my test case, use the method below:

```
import numpy as np
from peak import peak_local_max

def test_many_objects():
    mask=np.zeros([500,500], dtype=bool)
    x,y=np.indices((500,500))
    x_c=x//20*20+10
    y_c=y//20*20+10
    mask[(x-x_c)**2+(y-y_c)**2<8**2]=True
    # create a mask, label each disk, create distance image for peak searching
    labels, num_objs=ndi.label(mask)
    dist=ndi.distance_transform_edt(mask)
    local_max=peak_local_max(dist, min_distance=20, indices=True, exclude_border=False, labels=labels)
    assert(len(local_max) == 625)

```

skimage/feature/peak.py
```diff
@@ -123,6 +157,15 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
 
+    # no peak for a trivial image
+    if np.all(image == image.flat[0]):
+        if indices is True:
+            return np.empty((0, 2), np.int)
+        else:
+            return out
+
+    threshold_abs = threshold_abs if threshold_abs is not None else image.min()
+
```
sciunto: The comment just below "    # In the case of labels, recursively build and return an output
    # operating on each label separately" seems to be uncorrect to me now. There is no more recursive call and instead, it's a call to ndi on each label. Can you fix this please?

skimage/feature/peak.py
```diff
@@ -133,60 +176,41 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
-                                  footprint=footprint, labels=None)
-
-        # Select highest intensities (num_peaks)
-        coordinates = _get_high_intensity_peaks(image, out, num_peaks)
+        if exclude_border:
+            # create a mask for the non-exclude region
+            inner_mask = _exclude_border(np.ones_like(labels, dtype=bool),
+                                         footprint, exclude_border)
+
+        for i, obj in enumerate(ndi.find_objects(labels)):
```
sciunto: Perhaps a comment describing what this for loop is doing would help the reader.

skimage/feature/peak.py
```diff
@@ -133,60 +176,41 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
+            label = i + 1
+            # print("Label>", label)
+            img = image[obj] * (labels[obj] == label)
```
sciunto: Can we use a less generic variable name here?

skimage/feature/peak.py
clementkng: @brownmk How would you describe this loop?

skimage/feature/peak.py
clementkng: @brownmk What would you recommend here?

skimage/feature/peak.py
brownmk: sounds good.

skimage/feature/peak.py
brownmk: The variable label is used in the next line:

`img=image[obj]*(labels[obj]==label)`

skimage/feature/peak.py
brownmk: For each label, extract a smaller image enclosing the object of interest, identify num_peaks_per_label peaks and mark them in variable out.

skimage/feature/peak.py
brownmk: call it image_object? Up to you.

skimage/feature/peak.py
sciunto: Only this one to fix, and it will be good for me.

sciunto:
The test failure is due to #3985 Not blocking for this PR.

@jni @hmaarrfk  would you like to have a look as well?

skimage/feature/peak.py
```diff
@@ -22,6 +22,40 @@ def _get_high_intensity_peaks(image, mask, num_peaks):
...
+def _exclude_border(mask, footprint, exclude_border):
+    """
+    Remove peaks round the borders.
+    """
+    # zero out the image borders
+    for i in range(mask.ndim):
+        mask = mask.swapaxes(0, i)
+        remove = (footprint.shape[i] if footprint is not None
+                  else 2 * exclude_border)
+        mask[:remove // 2] = mask[-remove // 2:] = False
```
hmaarrfk: I really dislike swapping axes

```
mask[(slice(None),) * i + slice(None, remove // 2)] = False
mask[(slice(None),) * i + slice(-remove // 2, None)] = False
```

skimage/feature/peak.py
```diff
@@ -118,13 +152,21 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
     array([[10, 10, 10]])
 
     """
+    out = np.zeros_like(image, dtype=np.bool)
+
+    # no peak for a trivial image
+    if np.all(image == image.flat[0]):
```
hmaarrfk: Trial images are definitely an "edge case", is this check really necessary? the call to `np.all` and checking equality for large images is actually quite expensive. Since this is an "unlikely" scenerio, is it better to move this logic to the end or remove it all together.

skimage/feature/peak.py
```diff
@@ -133,60 +175,42 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
+            mask.fill(False)
+            mask[nd_indices] = True
+            out[obj] += mask
 
+        coordinates = _get_high_intensity_peaks(image, out, num_peaks)
         if indices is True:
             return coordinates
         else:
             nd_indices = tuple(coordinates.T)
             out[nd_indices] = True
```
hmaarrfk: coordinates is obtained from `out`. Do you need to set `out` again?

hmaarrfk:
Anyway to add teh benchmark code to the benchmarks dir?

skimage/feature/peak.py
clementkng: Sorry, are you referring to the line above or below or both?

clementkng:
@hmaarrfk I can add the benchmark code (assuming you mean the code sample under Way to Produce in the [original issue](https://github.com/scikit-image/scikit-image/issues/3974#issue-459542733)). I can extrapolate based off of other benchmarks how the benchmark for `peak_local_max` should be written, but since I would need conda to run asv, it would require some significant overhead on my end to get that set up. Would it be faster to push my benchmark code and have someone else confirm that it works, or should I proceed w/ a local install?

skimage/feature/peak.py
brownmk: Replace 
```
        mask = mask.swapaxes(0, i)
        remove = (footprint.shape[i] if footprint is not None
                  else 2 * exclude_border)
        mask[:remove // 2] = mask[-remove // 2:] = False
        mask = mask.swapaxes(0, i)

```
with what @hmaarrfk provided:

```
        remove = (footprint.shape[i] if footprint is not None
                  else 2 * exclude_border)
        mask[(slice(None),) * i + slice(None, remove // 2)] = False
        mask[(slice(None),) * i + slice(-remove // 2, None)] = False
```

brownmk: The code actually should be:
```
def _exclude_border(mask, footprint, exclude_border):
    """
    Remove peaks round the borders
    """
    # zero out the image borders
    for i in range(mask.ndim):
        remove = (footprint.shape[i] if footprint is not None
                  else 2 * exclude_border)
        mask[(slice(None),) * i + (slice(None, remove // 2),)] = False
        mask[(slice(None),) * i + (slice(-remove // 2, None),)] = False
    return mask
```

skimage/feature/peak.py
brownmk: Moving the code to the end does not help, as it will still need to be checked before final results are returned (potentially overwrite peak candidates).
I am okay with removing the logic, just we might need to somehow documente that there could be an incompatibility for edge cases.

Here is an example, where there was no peak before, it will become 25 peaks.  I agree this is a case probably users should check, instead of checking it every time within the library.

```
        image = np.ones((5, 5))
        # return 0, will return 25 if we remove the trivial case check
        print(len(peak.peak_local_max(image, min_distance=0, threshold_abs=0, indices=True)))

```

skimage/feature/peak.py
hmaarrfk: Thanks for making that super explicit

hmaarrfk:
i don't think you need conda for asv do you?

skimage/feature/peak.py
brownmk: The comment should be ("round" was a typo):

Remove peaks near the borders

skimage/feature/peak.py
hmaarrfk: OK, thanks for the explanation. I guess this can be left as a secondary issue.

skimage/feature/peak.py
brownmk: Good point. If num_peaks is provided, out may contain more than num_peaks candidates.  It needs to be regenerated in that case.

I actually see a bug here.  It was missing a statement             out.fill(False).
The code review really helps!!

It should be:

```
        if not indices and np.isinf(num_peaks):
            return out
        coordinates = _get_high_intensity_peaks(image, out, num_peaks)
        if indices:
            return coordinates
        else:
            out.fill(False)
            nd_indices = tuple(coordinates.T)
            out[nd_indices] = True
            return out

```

clementkng:
According to the install [link](https://asv.readthedocs.io/en/stable/installing.html), I need either virtualenv or conda. For now, the virtualenv solution is not working for me, possibly b/c I'm working of the Windows Subsystem for Linux.

skimage/feature/peak.py
clementkng: Where did the first two lines of code come from?

clementkng: Where did the first two lines of code come from?

hmaarrfk:
> Windows Subsystem for Linux

Having never used WSL it seems like a potential challenge. I feel like you are always going to hit these problems and never know if your code is wrong, or if it is a bug in WSL.

skimage/feature/peak.py
brownmk: The first two lines is to deal with the case where user does not provide num_peaks (defaults to all, np.inf) and indices is False, out is already the answer that can be returned.
The code I provided (bold) in within the context below:
```

        for i,obj in enumerate(ndi.find_objects(labels)):
            label=i+1
            #print("Label>", label)
            img=image[obj]*(labels[obj]==label)
            mask=_get_peak_mask(img, min_distance, footprint, threshold_abs, threshold_rel)
            if exclude_border:
                # remove peaks fall in the exclude region
                mask &= inner_mask[obj]
            coordinates =_get_high_intensity_peaks(img, mask, num_peaks_per_label)
            nd_indices = tuple(coordinates.T)
            mask.fill(False)
            mask[nd_indices] = True
            out[obj] += mask

        **if not indices and np.isinf(num_peaks):
            return out
        coordinates = _get_high_intensity_peaks(image, out, num_peaks)
        if indices:
            return coordinates
        else:
            out.fill(False)
            nd_indices = tuple(coordinates.T)
            out[nd_indices] = True
            return out**

    # Non maximum filter
    mask = _get_peak_mask(image, min_distance, footprint, threshold_abs, threshold_rel)

```

hmaarrfk: @brownmk are you sure you don't want to open your own PR and take it over from here? You seem to be going through this quite extensively. `Github Desktop`, and editors like `Atom` or `VS Code` make git much less painful.

brownmk: It's less of git, it's the 500-line guideline seems rather intimidating. All issues are addressed now, I will try to learn it next time.

https://github.com/scikit-image/scikit-image/blob/master/CONTRIBUTING.txt

hmaarrfk: That is a really fair point. I'll open an issue about  this.

Most of that is a reference for us to point people to when they hit problems. We often forget how we setup our own environments....

hmaarrfk:
```
skdev) mark2@xps ~/g/scikit-image speed-up-peak-local-max↑·1|+2                                                                                   
$ asv continuous -b PeakLocalMaxSuite -E conda:3.7 master speed-up-peak-local-max                                                                  
· Creating environments
· Discovering benchmarks
· Running 2 total benchmarks (2 commits * 1 environments * 1 benchmarks)
[  0.00%] · For scikit-image commit 7cd12184 <master> (round 1/2):
[  0.00%] ·· Building for conda-py3.7-cython-numpy-scipy.
[  0.00%] ·· Benchmarking conda-py3.7-cython-numpy-scipy
[ 25.00%] ··· Running (benchmark_peak_local_max.PeakLocalMaxSuite.time_peak_local_max--).
[ 25.00%] · For scikit-image commit 5a06093b <speed-up-peak-local-max> (round 1/2):
[ 25.00%] ·· Building for conda-py3.7-cython-numpy-scipy.
[ 25.00%] ·· Benchmarking conda-py3.7-cython-numpy-scipy
[ 50.00%] ··· Running (benchmark_peak_local_max.PeakLocalMaxSuite.time_peak_local_max--).
[ 50.00%] · For scikit-image commit 5a06093b <speed-up-peak-local-max> (round 2/2):
[ 50.00%] ·· Benchmarking conda-py3.7-cython-numpy-scipy
[ 75.00%] ··· benchmark_peak_local_max.PeakLocalMaxSuite.time_peak_local_max                                                             40.2±0.9ms
[ 75.00%] · For scikit-image commit 7cd12184 <master> (round 2/2):
[ 75.00%] ·· Building for conda-py3.7-cython-numpy-scipy.
[ 75.00%] ·· Benchmarking conda-py3.7-cython-numpy-scipy
[100.00%] ··· benchmark_peak_local_max.PeakLocalMaxSuite.time_peak_local_max                                                                1.75±0s
     [7cd12184]       [5a06093b]
     <master>         <speed-up-peak-local-max>
-         1.75±0s       40.2±0.9ms     0.02  benchmark_peak_local_max.PeakLocalMaxSuite.time_peak_local_max

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE INCREASED.
```

hmaarrfk:
Seems to work too
```
asv continuous -b PeakLocalMaxSuite -E virtualenv:3.7 master speed-up-peak-local-max
```

hmaarrfk:
I think the test for the trivial case should be added back. Unless we decide that this is a behavioural  change we want to make. Which I don't think this PR is advocating for.

 You can test for it in a different way, that isn't O(N) in cost.

```patch
$ git diff skimage/                                                                                                                                
diff --git a/skimage/feature/peak.py b/skimage/feature/peak.py
index 7736bf70c..7aa6ae20a 100644
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -190,6 +190,14 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
             out[obj] += mask
 
         coordinates = _get_high_intensity_peaks(image, out, num_peaks)
+
+        # Test for the case where the image is a constant
+        # Here, we decide to return to the user that there is no local maximum
+        if coordinates.shape[0] == image.size:
+            if indices is True:
+                return np.empty((0, 2), np.int)
+            else:
+                return np.zeros_like(image, dtype=bool)
         if indices is True:
             return coordinates
         else:
@@ -205,6 +213,14 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
     # Select highest intensities (num_peaks)
     coordinates = _get_high_intensity_peaks(image, mask, num_peaks)
 
+    # Test for the case where the image is a constant
+    # Here, we decide to return to the user that there is no local maximum
+    if coordinates.shape[0] == image.size:
+        if indices is True:
+            return np.empty((0, 2), np.int)
+        else:
+            return np.zeros_like(image, dtype=bool)
+
     if indices is True:
         return coordinates
     else:
```

A test like the

```python
In [1]: from skimage.feature.peak import peak_local_max                                                                                            
i
In [2]: import numpy as np                                                                                                                         

In [3]: image = np.ones((5, 5))                                                                                                                    

In [4]: print(len(peak_local_max(image, min_distance=0, threshold_abs=0, indices=True)))                                                           
0
```

Should also be added to test suite so that this behaviour doesn't magically disappear

brownmk:
The change is not right.  One cannot determine if the image is trivial by relying on the number of peaks detected.  The returned number of peaks depends on the settings of threshold and num_peaks.  For instance one can set num_peaks to 1 for a trivial image.  We should stick with the original code here.

> I think the test for the trivial case should be added back. Unless we decide that this is a behavioural change we want to make. Which I don't think this PR is advocating for.
> 
> You can test for it in a different way, that isn't O(N) in cost.

hmaarrfk:
Ah got it. Sorry for misunderstanding that. Yeah, for this PR. the original check should be kept.

skimage/feature/peak.py
clementkng: @brownmk If you want, I can try guiding you through a setup process in exchange for allowing me to create a PR w/ your code.

clementkng:
@hmaarrfk Is there any update in the background on this PR or the [issue](https://github.com/scikit-image/scikit-image/issues/3990) about the trivial check?

hmaarrfk:
background? I think you had the PR mostly complete. I don't think the failures are due to your code.

skimage/feature/tests/test_peak.py
```diff
@@ -445,3 +445,15 @@ def test_input_labels_unmodified(self):
...
+        mask = np.zeros([500, 500], dtype=bool)
+        x, y = np.indices((500, 500))
+        x_c = x // 20 * 20 + 10
+        y_c = y // 20 * 20 + 10
+        mask[(x - x_c) ** 2 + (y - y_c) ** 2 < 8 ** 2] = True
+        labels, num_objs = ndi.label(mask)
+        dist = ndi.distance_transform_edt(mask)
+        local_max = peak.peak_local_max(dist, min_distance=20, indices=True,
+                                        exclude_border=False, labels=labels)
+        assert (len(local_max) == 625)
```
hmaarrfk:
```suggestion
        assert len(local_max) == 625
```

just triggering a build

hmaarrfk: Maybe just apply this to trigger a fresh new build. I think builds are fixed.

hmaarrfk:
I think you can likely remove your WIP too.

sciunto:
Thanks @clementkng That's a great contribution!

PR #3987, opened 2019-07-01
Remove pytest-faulthandler from test dependencies

### LINKED ISSUE #3986: Packaging failures with Pytest 5.0
## Description

Pytest 5.0 was just released. It builds-in pytest-faulthandler and causes errors during packaging because we have pytest-faulthandler in requirements.txt.

Failing build: https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=761&view=logs&jobId=38fa3105-a3d6-58c5-2eec-10e695f0c1d9&taskId=a0d33157-d94d-587b-eae8-ff8bf4466631&lineStart=5880&lineEnd=5881&colStart=1&colEnd=1

We should either pin to 4.x or to 5.0+. I'd rather avoid complicated dependency logic doing different things for different pytest versions.

jni:
pytest-faulthandler is included in pytest as of pytest 5.0 and
including it in requirements raises a packaging error with 5.0.

Fixes #3986

## For reviewers

  later.

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -5,4 +5,3 @@ pytest-cov
 pytest-localserver
 flake8
 codecov
-pytest-faulthandler


rfezzani:
Hi @jni and @hmaarrfk, I first though that `pytest-faulthandler` was causing the CI failures for the PR #3983, but it is more complicated then it looks :-(
Looking at CI logs, it appears that `pad` is not raising the expected `TypeError` :-(. Do you have any idea of what is happening here?

hmaarrfk:
I maintain that we should "pass" when we fail on `--pre`. There is virtually no reason to keep all the tests at `red` when when it virtually no fault of this projects or the contributors for a pre-release.

jni:
@hmaarrfk I'm on board with this for sure, and would support a PR that puts `--pre` into acceptable failures. However that wouldn't have prevented the faulthandler failure.

hmaarrfk:
For reference, this is what we need for travis
https://github.com/hmaarrfk/scikit-image/pull/4/files

and something like this for appveyor

https://github.com/hmaarrfk/scikit-image/pull/3/files

Not sure about travis.

hmaarrfk:
i'm pretty sure we had a similar conversation before, and we had decided to fail on `--pre`. Not sure how to coerce the dask arrays into numpy ones now.

jni:
@hmaarrfk we did have that conversation, and @stefanv in particular was opposed, with the idea that maintainers wouldn't click through to "details" and thus failures would go undetected. If that can be avoided (will the Travis integration show "allowed failures" in the message, for example), then we can go ahead. Either way, though, after what feels like 17 pre-release-caused failures since we discussed this, I think it's legitimate to revisit the conversation.

hmaarrfk:
Well pytest-faulthandler just released 2 which can be installed with pytest 5.0

It is a dummy package that allows the packager to "install" both simulatenously. That said, this cleanup is probably a good thing anyway.

stefanv:
So what is the status of this PR then?

hmaarrfk:
Ready to go. Failures on pre is an other one

PR #4002, opened 2019-07-10
matplotlib: remove interpolation=nearest, none in our examples

### LINKED ISSUE #3998: SKIP: remove interpolation=nearest in our examples
## Description

Since 1.1.x, matplotlib set interpolation to nearest [by default](https://matplotlib.org/3.1.0/api/prev_api_changes/api_changes_1.1.x.html). We have plenty of examples in which the option is set to nearest (and a few to None).

I think there is room for simplifying our examples by removing these options. Any comment?

sciunto:
## Description

Closes #3998

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_coins_segmentation.py
+++ b/doc/examples/applications/plot_coins_segmentation.py
@@ -18,7 +18,7 @@ coins = data.coins()
 hist, hist_centers = histogram(coins)
 
 fig, axes = plt.subplots(1, 2, figsize=(8, 3))
-axes[0].imshow(coins, cmap=plt.cm.gray, interpolation='nearest')
+axes[0].imshow(coins, cmap=plt.cm.gray)
 axes[0].axis('off')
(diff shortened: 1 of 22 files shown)


hmaarrfk:
Does this still output warnings?

sciunto:
Warnings about what? I'm confused...

hmaarrfk:
I think this was added because we changed what interpolation method we used by default.

If a user runs this with 0.16.dev, will they get many deprecation warnings?

sciunto:
I'm still confused. Let me explain and see if it helps. Here we are talking about matplotlib. We have not decided which method to use by default (but mpl dev did). The removal of these option as no impact for users as mpl use nearest by default since a long time. they changed to this default value when they figured out that interpolating is not a good idea by default for people playing with images.

hmaarrfk:
.... good double explaining something that never needed it in the first place.

Thanks for your patience!

PR #4006, opened 2019-07-11
MAINT: Do not use assert in code, raise an exception instead.

sciunto:
## Description

I discovered that some of our functions are using assert statements to do checks.

This should be avoided. assert lines are removed with -O flags, so we may be in trouble to debug issues.

Any comments are welcome!

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


JDWarner:
The change in `predict` causes a test failure where a test for `LineModel1D` expects an AssertionError, but gets a ValueError.

That test needs to be tweaked.  No other blocking issues.

sciunto:
Thanks @JDWarner The test should be fixed.

Issue #3773, opened 2019-02-26
CLAHE return nan on black image


YubinXie:
## Description
When I apply CLAHE to my images, some of which is np.zeros((n,n)) after thresholding, CLAHE gives errors nan, nan. I am thinking if it is more reasonable to return np.zeros(n.n)). Thanks!

## Way to reproduce
```python
import numpy as np
a=np.zeros((10,10))
img_rescale = exposure.equalize_adapthist(a, clip_limit=0.03)
plt.imshow(img_rescale,'gray')
print(img_rescale)
```

```python
array([[nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan],
       [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan]])

```

stefanv:
Agreed.

jni:
In #3402 I made lots of checks for `threshold_li` that could really extend to a bunch of methods, including CLAHE. We should probably pilfer all those checks and make a "sanity check this array" function.

nandahkrishna:
Hello, can I contribute to this if it's still open?

hmaarrfk:
Of course!!! It doesn't seem that anybody has referenced this issue in a PR.

Have you seen the contribution guide?
https://scikit-image.org/docs/stable/contribute.html

let us know if you need any more guidance at this point, somebody will surely chime in.

(i've seen this https://github.com/scikit-image/scikit-image/pull/2761 but it seems to have gotten stale)

nandahkrishna:
Thanks!

nandahkrishna:
The PR #2761 seems to be an enhancement, so shall I incorporate the same idea in a PR and simultaneously solve this issue? Or would it be best to just solve this alone?

[...]

PR #4015, opened 2019-07-13
Edit rescale_intensity to return original image instead of nans

marshuang80:
## Description
This PR was originally addressing issue stated in #3773, which requests CLAHE to return numpy of zeros instead of nans when the input is np.zeros(n,n). 

Further examining the code indicates that the problem stems from the rescale_intensity() function in equalize_adapthist(), which returns np.array of nans if all values in the input image are the same. This PR edits rescale_intensity() to return the original image if the max and min of the image is the same. 

## Checklist

- [X ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [X ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -341,6 +341,9 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     array([  0,  63, 127], dtype=int8)
 
     """
+    if image.min() == image.max():
+        return image
+
     dtype = image.dtype.type
 
     imin, imax = intensity_range(image, in_range)
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -351,6 +351,8 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
 
     image = np.clip(image, imin, imax)
 
+    if imin == imax:
+        return image
     image = (image - imin) / float(imax - imin)
     return np.array(image * (omax - omin) + omin, dtype=dtype)
 
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -348,10 +348,9 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
 
     image = np.clip(image, imin, imax)
 
-    if imin == imax:
-        return np.asarray(image)
-    image = (image - imin) / float(imax - imin)
-    return np.array(image * (omax - omin) + omin, dtype=dtype)
+    if imin != imax:
+        image = (image - imin) / float(imax - imin)
+    return np.asarray(image * (omax - omin) + omin, dtype=dtype)
 
 
 def _assert_non_negative(image):
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -341,9 +341,6 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     array([  0,  63, 127], dtype=int8)
 
     """
-    if image.min() == image.max():
-        return image
-
     dtype = image.dtype.type
 
     imin, imax = intensity_range(image, in_range)
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -349,7 +349,7 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     image = np.clip(image, imin, imax)
 
     if imin == imax:
-        return image
+        return np.asarray(image)
     image = (image - imin) / float(imax - imin)
     return np.array(image * (omax - omin) + omin, dtype=dtype)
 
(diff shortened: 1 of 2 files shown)


skimage/exposure/exposure.py
```diff
@@ -341,6 +341,9 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     array([  0,  63, 127], dtype=int8)
 
     """
+    if image.min() == image.max():
```
hmaarrfk: I feel like you should use `imin` and `imax` computed below. The check should likely be done just before line 354 is computed. Where by you compare the denominator to `0` to check if you are dividing by zero.

marshuang80: That makes sense. Just to clarify, we should return the clipped image instead of the original image right?

hmaarrfk: Yes i believe so. The output would therefore be `omin` after all the math is performed.

I generally think we shouldn't spend too much time special casing edge cases like this, but not returning `nan` makes sense.

marshuang80: Sounds good, I've updated my PR anyways. 

There is a failed test from Travis but is seems unrelated to the fix.

hmaarrfk:
I think i mean that we require a *new* test. One that **would** have failed before your change, but passes now because of your change.

skimage/exposure/exposure.py
hmaarrfk: These two lines should be removed right?

skimage/exposure/exposure.py
```diff
@@ -348,6 +348,8 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
 
     image = np.clip(image, imin, imax)
 
+    if imin == imax:
+        return image
```
hmaarrfk: I would really like it if the only thing that was avoided here was the division by zero.

The code path should go through the same trajectory multiplying by `(omax - omin)` and adding `omin`.

Sound riddiculous, but sometimes not going through that code path can hide nasty bugs due to returning a view of the same image, as opposed to a new array like the code below does.

skimage/exposure/exposure.py
```diff
@@ -348,6 +348,8 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
...
     image = (image - imin) / float(imax - imin)
     return np.array(image * (omax - omin) + omin, dtype=dtype)
```
hmaarrfk: can you please make this `asarray` instead of `np.array`, it may avoid a copy if the image is already a float.

hmaarrfk: Sorry. I was referring to line 354, the final line.

skimage/exposure/exposure.py
marshuang80: Ah I see. So should I do the following check instead of returning an original copy? 
```if imin` != imax: image = (image - imin) / float(imax - imin)```

File: skimage/exposure/exposure.py (hunk 3 above)
Comment (marshuang80) [reply skimage/exposure/exposure.py]: Okay, sounds good. Should I still return the original image if max == min? Or should I just skip the division on line 353 if max != min?

Comment (hmaarrfk) [reply skimage/exposure/exposure.py]: Just skip the division part.

Comment (hmaarrfk) [reply skimage/exposure/exposure.py]: I want to return omin if the image is a constant. Get it?

Comment (marshuang80) [reply skimage/exposure/exposure.py]: Yep, got it

**Comment by marshuang80** [comment]
Sorry can you clarify on what I am testing?

**Comment by hmaarrfk** [comment]
I think that is more nitpicking.

But more that zeros is generally an edge case. I tend not to use it when testing for general "constants"

**Comment by marshuang80** [comment]
Okay sounds good, I just made another test with non-zero constants

File: skimage/exposure/tests/test_exposure.py [hunk 4]
```diff
@@ -267,6 +267,18 @@ def test_rescale_uint14_limits():
     assert_array_almost_equal(out, [0, uint14_max])
 
 
+def test_rescale_all_zeros():
+    image = np.zeros((2, 2), dtype=np.uint8)
+    out = exposure.rescale_intensity(image)
+    assert ~np.isnan(out).all()
```
hmaarrfk: We know what the output shouldn't be. Do we know what the output should be?

marshuang80: Yep, it should be the same

hmaarrfk: it is probably because of the changes I requested. I think due to the logic that you have now, omin and omax are the same, and thus the image will be rescaled to zero. Not sure what the right behavior here is.

PR #4018, opened 2019-07-13
Suppress warnings in test suite regarding measure.label

hmaarrfk:
## Description

Doing this actually revealed a bug. Will open an issue.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/TODO.txt
+++ b/TODO.txt
@@ -21,12 +21,6 @@ Version 0.17
 Version 0.18
 ------------
 
-* In ``skimage/meausre/_ccomp.label_cython`` remove the deprecated parameter
-  ``neighbors``.
-
-Version 0.18
-------------
-
 * Remove deprecated function ``guess_spatial_dimensions`` in ``skimage.color.__init__``.
 * Remove deprecated function ``load`` in ``skimage.data.__init__.py``.
 * The following functions are deprecated and should be removed in 0.18:
@@ -37,6 +31,8 @@ Version 0.18
 * `skimage/measure/tests/test_simple_metrics.py` should also be removed
 * make coordinates='rc' the default in ``segmentation.active_contour``
 * remove deprecated ``bc`` argument in ``segmentation.active_contour``
+* In ``skimage/meausre/_ccomp.label_cython`` remove the deprecated parameter
+  ``neighbors``.
 
 Other
 -----
--- a/TODO.txt
+++ b/TODO.txt
@@ -16,7 +16,7 @@ Version 0.17
 * Remove deprecated arguments ``beta1`` and ``beta2`` from function ``skimage.filters.frangi``
 * Remove deprecated arguments ``beta1`` and ``beta2`` from function ``skimage.filters.hessian``
 * Remove unused arguments ``dtype`` in ``imread`` functions. See #3918.
-    * Remove ``rough_wall`` from ``skimage.data.__init__``.
+* Remove ``rough_wall`` from ``skimage.data.__init__``.
 
 Version 0.18
 ------------
--- a/TODO.txt
+++ b/TODO.txt
@@ -16,7 +16,13 @@ Version 0.17
 * Remove deprecated arguments ``beta1`` and ``beta2`` from function ``skimage.filters.frangi``
 * Remove deprecated arguments ``beta1`` and ``beta2`` from function ``skimage.filters.hessian``
 * Remove unused arguments ``dtype`` in ``imread`` functions. See #3918.
-* Remove ``rough_wall`` from ``skimage.data.__init__``.
+    * Remove ``rough_wall`` from ``skimage.data.__init__``.
+
+Version 0.18
+------------
+
+* In ``skimage/meausre/_ccomp.label_cython`` remove the deprecated parameter
+  ``neighbors``.
 
 Version 0.18
 ------------
--- a/skimage/measure/_ccomp.pyx
+++ b/skimage/measure/_ccomp.pyx
@@ -371,10 +371,6 @@ def label_cython(input_, neighbors=None, background=None, return_num=False,
         # use the full connectivity by default
         connectivity = ndim
     elif neighbors is not None:
-        # not sure why stacklevel should only be 2 not 3. Maybe cython
-        # is stripping away a stacklevel????
-        warn("The argument 'neighbors' is deprecated, use 'connectivity' "
-             "instead", stacklevel=2)
         # backwards-compatible neighbors recalc to connectivity,
         if neighbors == 4:
             connectivity = 1
@@ -383,6 +379,13 @@ def label_cython(input_, neighbors=None, background=None, return_num=False,
         else:
             raise ValueError("Neighbors must be either 4 or 8, got '%d'.\n"
                              % neighbors)
+        # not sure why stacklevel should only be 2 not 3. Maybe cython
+        # is stripping away a stacklevel????
+        warn("The argument 'neighbors' is deprecated and will be removed in "
+             "scikit-image 0.18, use 'connectivity' instead. "
+             "For neighbors={neighbors}, use connectivity={connectivity}"
+             "".format(neighbors=neighbors, connectivity=connectivity),
+             stacklevel=2)
 
     if not 1 <= connectivity <= ndim:
         raise ValueError(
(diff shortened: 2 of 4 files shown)


skimage/morphology/tests/test_ccomp.py
```diff
@@ -66,10 +67,18 @@ def test_diag(self):
     def test_4_vs_8(self):
         x = np.array([[0, 1],
                       [1, 0]], dtype=int)
-        assert_array_equal(label(x, 4),
+        with expected_warnings(["use 'connectivity'"]):
+            assert_array_equal(label(x, 4),
```
emmanuelle: maybe put `ǹeighbors=4` here?

hmaarrfk: It's hard to tell the original meaning of the code. Maybe the test was designed to ensure positional arguments work as expected

skimage/morphology/tests/test_ccomp.py
```diff
@@ -66,10 +67,19 @@ def test_diag(self):
     def test_4_vs_8(self):
         x = np.array([[0, 1],
                       [1, 0]], dtype=int)
-        assert_array_equal(label(x, 4),
+        with expected_warnings(["use 'connectivity'"]):
+            assert_array_equal(label(x, ǹeighbors=4),
```
hmaarrfk:
```suggestion
            assert_array_equal(label(x, neighbors=4),
```

skimage/measure/_ccomp.pyx
```diff
@@ -371,8 +371,10 @@ def label_cython(input_, neighbors=None, background=None, return_num=False,
         # use the full connectivity by default
         connectivity = ndim
     elif neighbors is not None:
-        DeprecationWarning("The argument 'neighbors' is deprecated, use "
-                           "'connectivity' instead")
+        # not sure why stacklevel should only be 2 not 3. Maybe cython
+        # is stripping away a stacklevel????
+        warn("The argument 'neighbors' is deprecated, use 'connectivity' "
+             "instead", stacklevel=2)
```
jni: add `category=DeprecationWarning`?

skimage/morphology/convex_hull.py
```diff
@@ -126,7 +126,8 @@ def convex_hull_object(image, neighbors=8):
     if neighbors != 4 and neighbors != 8:
         raise ValueError('Neighbors must be either 4 or 8.')
 
-    labeled_im = label(image, neighbors, background=0)
+    connectivity = neighbors // 4
+    labeled_im = label(image, connectivity=connectivity, background=0)
```
jni: Happy to leave this to another PR, but we should deprecate neighbors here also. Let me know if you want to leave it for later.

skimage/morphology/tests/test_ccomp.py
```diff
@@ -103,11 +113,17 @@ def test_background_one_region_center(self):
                       [0, 1, 0],
                       [0, 0, 0]])
 
-        assert_array_equal(label(x, neighbors=4, background=0),
+        with expected_warnings(["use 'connectivity'"]):
+            assert_array_equal(label(x, neighbors=4, background=0),
+                               [[0, 0, 0],
+                                [0, 1, 0],
+                                [0, 0, 0]])
```
jni: Why not use connectivity directly here, like you've done with other tests?

skimage/morphology/tests/test_ccomp.py
```diff
@@ -182,8 +198,18 @@ def test_4_vs_8(self):
         x[1, 0, 0] = 1
         label4 = x.copy()
         label4[1, 0, 0] = 2
-        assert_array_equal(label(x, 4), label4)
-        assert_array_equal(label(x, 8), x)
+        with expected_warnings(["use 'connectivity'"]):
+            assert_array_equal(label(x, 4), label4)
+            assert_array_equal(label(x, 8), x)
+
+    def test_connectivity_1_vs_2(self):
```
jni: Maybe 1 vs 3??? :joy:

skimage/morphology/convex_hull.py
hmaarrfk: Let's leave it for later

skimage/measure/_ccomp.pyx
hmaarrfk: No. Deprecation warnings get magically silenced by python. We don't do this consistently elsewhere in the library

skimage/morphology/tests/test_ccomp.py
hmaarrfk: I mostly just duplicated the tests to ensure that both the old and new behaviour give the expected results during the deprecation cycle

skimage/morphology/tests/test_ccomp.py
hmaarrfk: Will do

hmaarrfk:
I think the reason this never got caught before was that deprecation warnings are getting silenced by python. We had an internal function calling with the old parameter, something that should not have happened.

We should figure out how we want to deprecate this? Remove in scikit image 0.18? Seems fair. I can add the message and the Todo when I get to a real computer

skimage/measure/_ccomp.pyx
jni: > get magically silenced by Python

?

```
 $ cat test.py
import warnings
warnings.warn('hello', category=DeprecationWarning)

 $ python test.py 
test.py:2: DeprecationWarning: hello
  warnings.warn('hello', category=DeprecationWarning)
```

> we don't do this consistently

never too late to start. ;)

jni: I think the previous code was simply incorrect, rather than any Python silencing.

jni:
@hmaarrfk sounds good!

skimage/measure/_ccomp.pyx
hmaarrfk: https://docs.python.org/3/library/warnings.html#warnings.warn

I can't send screenshot, but see the explanation there

hmaarrfk:
Unless you want to just edit the changes. I don't really care. I can press "accept" or you can use the online editor ;)

skimage/measure/_ccomp.pyx
jni: @hmaarrfk either I'm insane or that says nothing about DeprecationWarnings being ignored. It says that if `message` is already a DeprecationWarning object, whatever you give to category= will be ignored. But we are passing a string, so that's not an issue...?

jni: Note: by "insane" I mean "sleepy/tired". =P On a dark plane going cross country right now.

hmaarrfk: Deprecation warning: Base category for warnings about deprecated features when those warnings are intended for other Python developers (ignored by default, unless triggered by code in __main__).

skimage/morphology/tests/test_ccomp.py
hmaarrfk:
```suggestion
    def test_connectivity_1_vs_3(self):
```

skimage/measure/_ccomp.pyx
jni: @hmaarrfk :man_facepalming: Yikes. 

However, clicking on the link, I don't see the part in parentheses?

hmaarrfk: I guess it is in their table only
image

was on my phone when i was sending the initial link.

jni: Wow. Just retested with #4035 and indeed the DeprecationWarning is silent. wth. Thanks for the pointer!!!

sciunto: Shouldn't we set a version at which the support for this deprecated argument is over? I don't see anything in TODO.txt.

TODO.txt
```diff
@@ -16,7 +16,7 @@ Version 0.17
 * Remove deprecated arguments ``beta1`` and ``beta2`` from function ``skimage.filters.frangi``
 * Remove deprecated arguments ``beta1`` and ``beta2`` from function ``skimage.filters.hessian``
 * Remove unused arguments ``dtype`` in ``imread`` functions. See #3918.
-* Remove ``rough_wall`` from ``skimage.data.__init__``.
+    * Remove ``rough_wall`` from ``skimage.data.__init__``.
```
sciunto: I think this is not intended

sciunto:
Approved. I'm just waiting for the tests.

Issue #2728, opened 2017-07-23
hough_circle_peaks  not respecting min_xdistance, min_ydistance.


tonypower:
## Description
It looks like hough_circle_peaks  is not respecting min_xdistance, min_ydistance. Below I provide an example that is just a modified version of the example provided in http://scikit-image.org/docs/dev/auto_examples/edges/plot_circular_elliptical_hough_transform.html

## Way to reproduce
~~~~
hough_radii = np.arange(15, 35, 2)
hough_transf = hough_circle(edges, hough_radii)
accums, cx, cy, radii = hough_circle_peaks(hough_transf, hough_radii, total_num_peaks=24,
                                               min_xdistance=50, min_ydistance=50)

image = color.gray2rgb(image)
    
for center_y, center_x, radius in zip(cy, cx, radii):
     circy, circx = draw.circle_perimeter(center_y, center_x, radius)
     image[circy, circx] = (255, 0, 0)
~~~~

![hough_circle_skimage](https://user-images.githubusercontent.com/11573891/28502279-0d52033a-6fe7-11e7-99b1-c351ab7b9a48.png)

It is clear from the image above that, for instance, the first coin (and second) has two circles whose centers are much closer that 50 pixels in any of the dimensions.

Linux Debian
Python 3.5.2
scikit image Version: 0.13.0

Thanks,

jmetz:
So we can see that this is because each radius is being processed independently of the rest (including wrt `min_xdistance`, `min_ydistance`), so I guess the `needs decision` that @soupault flagged here was whether that should be the case or not. 

The fix would require doing 3d peak finding **where `normalize` would need to be set to true to not bias toward larger radii**. 

I'm happy to implement this fix if we get a consensus on this.

jmetz:
UPDATE: I've implemented a specialised 3d peak finding that uses a max-projection along the "radii axis" from the hough transform (i.e. it doesn't need to use a `maximum_filter1d` along that axis). 

It seems to work and fixes the issue above. 

Happy to put in a PR once we get consensus :)

soupault:
>So we can see that this is because each radius is being processed independently of the rest (including wrt min_xdistance, min_ydistance), so I guess the needs decision that @soupault flagged here was whether that should be the case or not.

[...]

PR #4024, opened 2019-07-14
WIP:  Add minimal distance argument to local_maxima

### LINKED ISSUE #3816: Add minimal distance argument to local_maxima
https://github.com/scikit-image/scikit-image/pull/3809#issuecomment-475874038 got me thinking that an argument specifying the minimal allowed euclidean distance between maxima could be very useful. I imagine it to work akin to [this implementation](https://github.com/scipy/scipy/blob/715f3ad659fbe1f8e4b9412bf4433f85cf21dff5/scipy/signal/_peak_finding_utils.pyx#L90-L157) in SciPy. The difficulty for the n-dimensional case is to only evaluate close neighbors as otherwise the complexity grows very unfavorably with the number of maxima. I wonder whether a [KDTree](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.KDTree.html) could help optimizing that particular problem. 

It's only an idea right now but before I investigate I'd like to know if there is an interest in adding this feature.

lagru:
## Description

Closes #3816. 

Adds the optional keyword `distance` to [local_maxima](https://scikit-image.org/docs/dev/api/skimage.morphology.html#skimage.morphology.local_maxima) (and local_minima) to specify the minimal euclidean distance allowed between maxima. In case of a conflict, the maximum with the smaller value is dismissed.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/_extrema_cy.pyx
+++ b/skimage/morphology/_extrema_cy.pyx
@@ -5,7 +5,9 @@
 
 """Cython code used in extrema.py."""
 
+import numpy as np
 cimport numpy as cnp
+from libc.math cimport pow
 
 
 # Must be defined to use QueueWithHistory
@@ -57,7 +59,7 @@ def _local_maxima(dtype_t[::1] image not None,
         An array of flags that is used to store the state of each pixel during
         evaluation and is MODIFIED INPLACE. Initially, pixels that border the
         image edge must be marked as "BORDER_INDEX" while all other pixels
-        should be marked with "NOT_MAXIMUM".
+        should be marked with "NOT_MAXIMUM". Modified in place.
     neighbor_offsets : ndarray
         A one-dimensional array that contains the offsets to find the
         connected neighbors for any index in `image`.
@@ -107,7 +109,7 @@ cdef inline void _mark_candidates_in_last_dimension(
         The raveled view of a n-dimensional array.
     flags :
         An array of flags that is used to store the state of each pixel during
-        evaluation.
+        evaluation. Modified in place.
     
     Notes
     -----
@@ -152,7 +154,7 @@ cdef inline void _mark_candidates_all(unsigned char[::1] flags) nogil:
     ----------
     flags :
         An array of flags that is used to store the state of each pixel during
-        evaluation.
+        evaluation. Modified in place.
     """
     cdef Py_ssize_t i = 1
     while i < flags.shape[0]:
@@ -173,7 +175,7 @@ cdef inline void _fill_plateau(
         The raveled view of a n-dimensional array.
     flags :
         An array of flags that is used to store the state of each pixel during
-        evaluation.
+        evaluation. Modified in place.
     neighbor_offsets :
         A one-dimensional array that contains the offsets to find the
         connected neighbors for any index in `image`.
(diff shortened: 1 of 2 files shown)


lagru:
The relevant implementation is in Cython and uses a rather "naive brute force" algorithm:
- Iterate maxima in order of priority (here maxima with higher image values are prioritized)
   - *continue* if current maximum was removed in an earlier step
   - Find all points of current maxima
   - Find all points within minimal distance of any point in the current maxima
   - Remove all other maxima that reach into this volume

Early benchmarks indicate a slow-down by 400% and more depending on how the `distance`-keyword is used. I still want to benchmark this more thoroughly and see if I can find any optimizations and how it compares to similar functions: e.g. [peak_Local_max](https://scikit-image.org/docs/dev/api/skimage.feature.html?highlight=peak_local_max#skimage.feature.peak_local_max) and [find_peaks](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.find_peaks.html) come to mind.

The current API is not necessarily the best one. It may be useful to make this public or add this keyword to [h_maxima](https://scikit-image.org/docs/dev/api/skimage.morphology.html#skimage.morphology.h_maxima) as well.

lagru:
Can't believe I forgot my earlier attempt https://github.com/scikit-image/scikit-image/issues/3016#issuecomment-405087610. :smile: I'll need to investigate how these two approaches compare...

lagru:
Okay, a quick evaluation hints that the implementation in https://github.com/scikit-image/scikit-image/issues/3016#issuecomment-405087610 is significantly slower and scales very badly with an increasing number of peaks.  So the new approach seems like the way to go.

jni:
@lagru very cool! Sorry for the silence, it's been busy. =)

I haven't had a chance to fully understand the code yet. With my review, I'll probably aim to break the function into some smaller functions so that the flow can be clearer. In the meantime though, it would be great to add some tests to this, and in particular to verify that functions where `peak_local_max` is currently doing a poor job actually work with this approach, e.g. https://github.com/scikit-image/scikit-image/issues/2728

Also, imho the keyword argument should be called `minimum_distance`, rather than just `distance`.

Thanks!

jni:
@lagru another point, about your initial implementation: you were using brute-force to compare the distance to every other peak. Instead, a scipy.spatial.cKDTree could be used to rapidly search for nearby maxima, once you have the coordinates.

lagru:
> I'll probably aim to break the function into some smaller functions so that the flow can be clearer. In the meantime though, it would be great to add some tests to this, and in particular to verify that functions where peak_local_max is currently doing a poor job actually work with this approach

Great suggestions. Definitely something to keep in mind/do next. I'm still tweaking the algorithm so I refrained from adding tests and compartmentalizing the code for now. 

[cKDTree](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html) looks like it could be really helpful. I initially shied away because it works with "unraveled" indices (that can be solved though). I'm guessing that because it's written in C we could use it directly in Cython. Not sure whether the C-API is part of SciPy's official API though. I'll investigate and see if there's a performance improvement to be had.

skimage/morphology/_extrema_cy.pyx
```diff
@@ -222,3 +226,191 @@ cdef inline void _fill_plateau(
...
+                        if flags[neighbor] == MAXIMUM:
+                            queue_push(&current_maximum, &neighbor)
+                            queue_count[neighbor] = 1
+                        else:
+                            queue_push(&to_search, &neighbor)
+                            queue_count[neighbor] = 1
+
+                # Evaluate the space within the minimal distance of the current
+                # maximum
+                while queue_pop(&to_search, &current_index):
```
lagru: As @jni suggested, a cKDTree might be a good tool to speed this while-block up.

sciunto:
I don't resist to xref to this issue: #4048 Perhaps we can solve and gain also on peak_local_max.

lagru:
@jni in the mailing list you mentioned this:

> The padding of a volume to avoid neighbor issues was done for convenience. It can and should be removed, and checking for out-of-bounds access is a mod (%) calculation away.

That'd be great to do however I don't see how this is only a modulo calculation away. As I understand it, one has to do a modulo call for all dimension in the correct order an evaluate the reminder each time (a % b = r, r=0 or r=b-1 -> border, else not border). But that wouldn't solve the problem because we actually care about not crossing the border in the wrong direction (evaluating an index at the border is fine in itself). So we'd have to consider the offsets as well.

That is quite a bit of overhead compared to padding with a border flag. Seems to me that we would be trading memory cost for runtime and complexity. Do you have examples or experience regarding this or (if you find the time) can you elaborate where this statement is coming from?

JDWarner:
Lars is correct, it's not easy or trivial; if one goes back through my flood fill PR my immediate approach was to remove the padding.  That initially seemed to work, but actually let the fill alias off one side and into the opposite side of the image.  Fixing this became such a headache that I reverted to padding.

lagru:
Closed in favor of #4165 which proposes an equally fast (if not faster) solution based on SciPy's cKDTree.

Issue #2345, opened 2016-10-18
Implement quality metrics for image segmentation, classification and object detection


soupault:
## Description

I believe that these days a vast majority of engineers and researchers in the field of image processing are working on image segmentation, classification and object detection problems. That said, I think that it could be benefitial to have some typical quality metrics implemented in `scikit-image`.
Of course, there are already some simple metrics in `scikit-learn`, but none of them are applicable as is to the problems mentioned above. Image processing-related problems are quite specific in that sense.

As a good starting point, I'd suggest to implement some of the metrics used in ImageNet Challenge (see both http://image-net.org/challenges/LSVRC/2012/index and http://image-net.org/challenges/LSVRC/2016/index) and/or PASCAL VOC (see both http://host.robots.ox.ac.uk/pascal/VOC/voc2007/htmldoc/index.html and http://host.robots.ox.ac.uk/pascal/VOC/voc2012/htmldoc/index.html).

Perhaps, we should consider adding several different algorithms for each problem (e.g. for object detection there is a brilliant article `B. Ozdemir, S. Aksoy, S. Eckert, M. Pesaresi, and D. Ehrlich, “Performance measures for object detection evaluation,” Pattern Recognit. Lett., vol. 31, no. 10, pp. 1128–1137, 2010.` with 3 different quality metrics to be used in different cases).

If you have any thoughts on this topic, please, do share.

P.S. Let this GitHub issue be an umbrella topic for the mentioned things.

soupault:
@scikit-image/core any opinions on this?

sciunto:
I don't think it's clear to me what's a quality metric? Is it an analogue of the chi2 for a fit? Do you have a short illustration of the benefits?

soupault:
@sciunto I'm refering to metrics typically used for evaluation of segmentation quality (http://host.robots.ox.ac.uk/pascal/VOC/voc2012/htmldoc/devkit_doc.html#SECTION00064000000000000000, https://en.wikipedia.org/wiki/Jaccard_index), object detection quality (http://host.robots.ox.ac.uk/pascal/VOC/voc2012/htmldoc/devkit_doc.html#SECTION00054000000000000000, the ones from the article above), and multi-class classification quality (accuracy, precision, recall, http://host.robots.ox.ac.uk/pascal/VOC/voc2012/htmldoc/devkit_doc.html#SECTION00044000000000000000).

Here is an example of metric from the aforementioned article:
[...]

PR #4025, opened 2019-07-14
skimage.evaluation module

rbavery:
## Description
This pull request supersedes https://github.com/scikit-image/scikit-image/pull/3354 and addresses https://github.com/scikit-image/scikit-image/issues/2345. It refactors @seanbudd's original PR into a new `skimage.evaluate` module, moving simple metrics (mse, nrmse, psnr) and segmentation metrics (adapted rand error, variation of information) into `evaluate`.

Citations for segmentation metrics:
```
Arganda-Carreras I, Turaga SC, Berger DR, et al. (2015)
           Crowdsourcing the creation of image segmentation algorithms
           for connectomics. Front. Neuroanat. 9:142.
           :DOI:`10.3389/fnana.2015.00142`

Marina Meilă (2007), Comparing clusterings—an information based
           distance,Journal of Multivariate Analysis, Volume 98, Issue 5, Pages
           873-895, ISSN 0047-259X, https://doi.org/10.1016/j.jmva.2006.11.013.
```

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


requirements/docs.txt
```diff
@@ -6,6 +6,7 @@ sphinx-gallery
 sphinx-copybutton
 pytest-runner
 scikit-learn
+matplotlib>=3.1.0
```
hmaarrfk: can you explain the need to go bleeding edge here?

rbavery: Sure, the issue is that `constrained_layout` isn't implemented in matplotlib 2.0 so one of the doc builds was failing. `constrained_layout` is implemented in 2.2 but is experimental. It received a substantial update in matplotlib 3.0 to also constrain text:
https://matplotlib.org/3.0.0/api/api_changes.html
```
Layout tools like Figure.tight_layout, constrained_layout, and fig.savefig('fname.png', bbox_inches="tight") use matplotlib.axes.Axes.get_tightbbox to determine the bounds of each axes on a figure and adjust spacing between axes.

In Matplotlib 2.2 get_tightbbox started to include legends made on the axes, but still excluded some other artists, like text that may overspill an axes. This has been expanded to include all artists.
```
@jni suggested 3.1.0 so I went with that over 3.0.

hmaarrfk: I guess it is the docs so it is fine. I'm cool with bleeding edge anyway. Much more fun :D.

rbavery:
I'm not sure what the python 3.7 build is failing due to this error:
```
Error: Download failed on Cask 'basictex' with message: Download failed: http://mirror.ctan.org/systems/mac/mactex/mactex-basictex-20190410.pkg
+shell_session_update
```
I'll look into this later today.

The Python 3.5 build fails are because of a typo, I need to set the minimum version to `3.0.1` not `3.1.0`

hmaarrfk:
Is the build fixed now? Sometimes mirrors fail. There was a `retry` flag on most pip stuff. Not sure about other dependencies. THe general rule is that at least 1 build will fail everytime everythign is correct.

skimage/evaluate/simple_metrics.py
```diff
@@ -0,0 +1,143 @@
...
+from ..util.dtype import dtype_range
+from .._shared.utils import warn
+
+__all__ = ['mean_squared_error',
+           'normalized_root_mse',
+           'peak_signal_noise_ratio',
+           ]
+
+
+def _assert_compatible(im1, im2):
```
JDWarner: Is this the ideal place for this function?  Would it make more sense for this to live in `_shared`?

Some duplicated effort where we ask for masks in other functions could be simplified with this.  I wouldn't mind removing the leading underscore in that case.

skimage/evaluate/simple_metrics.py
```diff
@@ -0,0 +1,143 @@
...
+        if true_min >= 0:
+            # most common case (255 for uint8, 1 for float)
+            data_range = dmax
+        else:
+            data_range = dmax - dmin
+
+    im_true, im_test = _as_floats(im_true, im_test)
+
+    err = mean_squared_error(im_true, im_test)
+    return 10 * np.log10((data_range ** 2) / err)
```
JDWarner: In `mean_squared_error` above we use `np.square`.  Here we use `** 2`.  I think `** 2` may be slightly faster for small inputs.  I don't have strong feelings either way (both should be appropriately dispatched by `__array_function__`) but we should probably choose one and be consistent.

rbavery:
> Is the build fixed now? Sometimes mirrors fail. There was a `retry` flag on most pip stuff. Not sure about other dependencies. THe general rule is that at least 1 build will fail everytime everythign is correct.

Yep, looks like all builds pass. I think the python 3.7 build that failed last time was a fluke, it's working now.

doc/source/api_changes.rst
```diff
@@ -1,11 +1,25 @@
+Version 0.16
+------------
+- The following functions are deprecated and will be removed in 0.18:
+  `skimage.measure.mse`,
```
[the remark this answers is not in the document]
rbavery: true, true I'll fix that.

skimage/evaluate/simple_metrics.py
```diff
@@ -0,0 +1,143 @@
...
+    return im1, im2
+
+
+def mean_squared_error(im1, im2):
+    """Compute the mean-squared error between two images.
+
+    Parameters
+    ----------
+    im1, im2 : ndarray
+        Image.  Any dimensionality.
```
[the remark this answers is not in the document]
rbavery: Can't hurt. I'll add this note for each of the evaluate functions.

rbavery:
> Thank you for the contribution! This is a good cleanup of the namespace in `measure`, and I like the API for the new `variation_of_information`. I really only have a few minor comments/tweaks, some which relate to the original code and not your changes, with one additional question:
> 
> Is there a reason `compare_ssim` didn't make the move also? It is very much in this class of functions, and seems like `skimage.evaluate` would be the proper home for it as well.

Hey @JDWarner thanks for the review! I missed `compare_ssim` on my first look through of measure, I've moved it to `skimage.evaluate.structural_similarity` and updated `api_changes` and `TODO` accordingly.

skimage/evaluate/simple_metrics.py
rbavery: Agreed, I went through evaluate and changed instances of `np.square` to `** 2`

skimage/evaluate/simple_metrics.py
rbavery: Agreed, moved to `_shared` and renamed `assert_equal_shape`

stefanv:
I don't object to the new submodule, although I wonder if `evaluate` evokes a strong connection with the proposed content. Evaluate exactly what? Why not compare? It holds for segmentation and images equally well.

Sorry to nitpick on such a bike-sheddable item, but since we don't often introduce top-level models I think it is worth getting the name right in this case.

rbavery:
@stefanv I hear you, I was thinking of the term `evaluation metrics` when I chose that name, but some other names I considered for the submodule were `metrics` and `validate`. `compare` sounded a bit too general to me, if I'm looking for metrics that measure error, I'd more expect to find them in a module called `validate` or `metrics` than `compare`. Even though not all the functions here are measuring error I think most folks will use these to compare an image to a true reference image, so choosing a name that reflects this type of comparison would help folks find the functions they need I think.

But I'm not totally opposed to `compare`, curious to hear what you think about the other options and if `compare` is still preferred.

stefanv:
Metrics would feel more familiar to me. What do others think?

JDWarner:
Metrics feels familiar and the colloquial definition fits.  However, while I don't want to get too far into the weeds with this, note that *metric* has a specific mathematical definition and not all of these functions are metrics.

For example, two routines I'll be contributing as soon as this refactor happens are the Dice coefficient and Jaccard index for comparison of binary images/segmented objects against a ground truth.  The Dice is not a metric, while the Jaccard is (see [the Jaccard Wikipedia article](https://en.wikipedia.org/wiki/Jaccard_index) for further details).

rbavery:
I'm leaning towards `metrics`. Maybe we can note in the module description that it also includes semimetrics like the Dice coefficient. I checked out the wiki article and it sounds like `semimetric` is a valid term for measures that don't satisfy the triangle equality. And there are other sets of measures that satisfy some of the rules for being a metric, like `quasimetrics`, and all of them seem to have the suffix `metric`. https://en.wikipedia.org/wiki/Metric_(mathematics)#Generalized_metrics

skimage/measure/simple_metrics.py
```diff
@@ -29,31 +24,34 @@ def compare_mse(im1, im2):
...
-    im1, im2 = _as_floats(im1, im2)
-    return np.mean(np.square(im1 - im2), dtype=np.float64)
+    warn('DEPRECATED: skimage.measure.compare_mse has been moved to '
+         'skimage.evaluation.mean_squared_error. It will be removed from '
+         'skimage.measure in version 0.18.')
+    return mean_squared_error(im1, im2)
 
 
-def compare_nrmse(im_true, im_test, norm_type='Euclidean'):
+def compare_nrmse(im_true, im_test, norm_type='euclidean'):
```
sciunto: Call `.lower()` to do not break legacy user's codes.

skimage/evaluate/_variation_of_information.py
```diff
@@ -0,0 +1,129 @@
...
+    Returns
+    -------
+    vi : ndarray of float, shape (2,)
+        The conditional entropies of im_test|im_true and im_true|im_test.
+
+    References
+    ----------
+    .. [1] Marina Meilă (2007), Comparing clusterings—an information based
+           distance,Journal of Multivariate Analysis, Volume 98, Issue 5, Pages
+           873-895, ISSN 0047-259X, https://doi.org/10.1016/j.jmva.2006.11.013.
```
sciunto:
```suggestion
           873-895, ISSN 0047-259X, :DOI:`10.1016/j.jmva.2006.11.013`.
```

skimage/measure/simple_metrics.py
rbavery: right, `.lower` gets called in the corresponding function in evaluate, and `compare_nrmse` calls it

https://github.com/scikit-image/scikit-image/blob/7bf1da54cd1f91514da7ab6ac7482e9ce64d26c9/skimage/evaluate/simple_metrics.py#L80

skimage/evaluate/_variation_of_information.py
rbavery: done!

skimage/_shared/testing.py
```diff
@@ -52,6 +52,13 @@ def assert_greater(a, b, msg=None):
     assert a > b, message
 
 
+def assert_shape_equal(im1, im2):
+    """Raise an error if the shape and dtype do not match."""
+    if not im1.shape == im2.shape:
```
sciunto: Why not just `assert im1.shape == im2.shape`? Do we really need this function? The gain looks low to me.

sciunto: See also my comments below.

skimage/metrics/_adapted_rand_error.py
```diff
@@ -0,0 +1,78 @@
...
+    Pixels with label 0 in the true segmentation are ignored in the score.
+
+    References
+    ----------
+    .. [1] Arganda-Carreras I, Turaga SC, Berger DR, et al. (2015)
+           Crowdsourcing the creation of image segmentation algorithms
+           for connectomics. Front. Neuroanat. 9:142.
+           :DOI:`10.3389/fnana.2015.00142`
+    """
+    assert_shape_equal(im_true, im_test)
```
sciunto: This comment follows another inline comment I made where this function is defined.

I don't like calling this function assert_something, because it does not call assert. And this is important. Assert statements are removed if python is called with an optimization flag.

So, assert_shape_equal is not written in the right file. I'm not sure where it can fit the best and I'm not certain that we want this sort of functions in scikit-image. @scikit-image/core Thoughts?

skimage/metrics/_structural_similarity.py
```diff
@@ -0,0 +1,226 @@
...
+def structural_similarity(im1, im2, win_size=None, gradient=False,
+                          data_range=None, multichannel=False,
+                          gaussian_weights=False, full=False, **kwargs):
+    """Compute the mean structural similarity index between two images.
+
+    Parameters
+    ----------
+    im1, im2 : ndarray
+        Images. Any dimensionality with same shape.
+    win_size : int or None
```
sciunto:
```suggestion
    win_size : int or None, optional
```

skimage/metrics/_variation_of_information.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # then, compute the entropies
+    hygx = -px @ _xlogx(px_inv @ pxy).sum(axis=1)
+    hxgy = -_xlogx(pxy @ py_inv).sum(axis=0) @ py
+
+    return list(map(np.asarray, [hxgy, hygx]))
+
+
+def _invert_nonzero(arr):
+    """Compute the inverse of the non-zero elements of arr, not changing 0.
```
sciunto: Please add an empty line before section headers.

skimage/metrics/simple_metrics.py
```diff
@@ -0,0 +1,137 @@
...
+    im2 = np.asarray(im2, dtype=float_type)
+    return im1, im2
+
+
+def mean_squared_error(im1, im2):
+    """Compute the mean-squared error between two images.
+
+    Parameters
+    ----------
+    im_true, im_test : ndarray
```
sciunto:
```suggestion
    im1, im2 : ndarray
```

skimage/metrics/simple_metrics.py
```diff
@@ -0,0 +1,137 @@
...
+    """Compute the normalized root mean-squared error (NRMSE) between two
+    images.
+
+    Parameters
+    ----------
+    im_true : ndarray
+        Ground-truth image, same shape as im_test.
+    im_test : ndarray
+        Test image.
+    norm_type : {'Euclidean', 'min-max', 'mean'}
```
sciunto:
```suggestion
    norm_type : {'Euclidean', 'min-max', 'mean'}, optional
```

skimage/metrics/simple_metrics.py
```diff
@@ -0,0 +1,137 @@
...
+def peak_signal_noise_ratio(im_true, im_test, data_range=None):
+    """ Compute the peak signal to noise ratio (PSNR) for an image.
+
+    Parameters
+    ----------
+    im_true : ndarray
+        Ground-truth image, same shape as im_test.
+    im_test : ndarray
+        Test image.
+    data_range : int
```
sciunto:
```suggestion
    data_range : int, optional
```

skimage/metrics/tests/test_segmentation_metrics.py
```diff
@@ -0,0 +1,33 @@
...
+
+def test_contingency_table():
+    im_true = np.array([1, 2, 3, 4])
+    im_test = np.array([1, 1, 8, 8])
+
+    table1 = sparse.coo_matrix((np.full(im_true.size, 1 / im_true.size),
+                                (im_true.ravel(), im_test.ravel())),
+                               dtype=float).tocsr()
+    table2 = contingency_table(im_true, im_test, normalize=True)
+    (table1 != table2).nnz == 0
```
sciunto:
```suggestion
    assert (table1 != table2).nnz == 0
```
?

skimage/metrics/tests/test_simple_metrics.py
```diff
@@ -0,0 +1,78 @@
...
+from skimage._shared import testing
+import numpy as np
+
+import skimage.data
+from skimage.metrics import (peak_signal_noise_ratio, normalized_root_mse,
+                              mean_squared_error)
+
+
+np.random.seed(5)
+cam = skimage.data.camera()
```
sciunto:
```suggestion
cam = data.camera()
```

doc/examples/segmentation/plot_metrics.py
```diff
@@ -0,0 +1,145 @@
+"""
+========================
+Evaluating segmentations
```
sciunto: Segmentation metrics (?)

doc/examples/segmentation/plot_metrics.py
```diff
@@ -0,0 +1,145 @@
...
+
+ax[4].imshow(mark_boundaries(image, im_test2))
+ax[4].set_title('Edge Detection')
+ax[4].set_axis_off()
+
+ax[5].imshow(mark_boundaries(image, im_test3))
+ax[5].set_title('Morphological GAC')
+ax[5].set_axis_off()
+
+plt.show()
```
sciunto: This is how the plot looks like for me.

sphx_glr_plot_metrics_001

skimage/_shared/testing.py
rbavery: We had `assert im1.shape == im2.shape` previously, @JDWarner thought it might be useful to have a function in `_shared`
See https://github.com/scikit-image/scikit-image/pull/4025#discussion_r303270180
If we made this function use an assert statement would that be an improvement? I'm impartial either way.

```
In [7]: def assert_shape_equal(a, b): 
   ...:     message = "%r does not have equal shape with %r" % (a.shape, b.shape) 
   ...:     assert a.shape == b.shape, message 

In [8]: assert_shape_equal(x,y)                                                                                                                                                   
AssertionError                            Traceback (most recent call last)
<ipython-input-8-8237c8a1005b> in <module>
----> 1 assert_shape_equal(x,y)

<ipython-input-7-2933226b4e44> in assert_shape_equal(a, b)
      1 def assert_shape_equal(a, b):
      2     message = "%r does not have equal shape with %r" % (a.shape, b.shape)
----> 3     assert a.shape == b.shape, message
      4 
      5 

AssertionError: (3, 3) does not have equal shape with (4, 4)

```

skimage/metrics/tests/test_segmentation_metrics.py
rbavery: This tests if the `contingency_table` function is working, forgot the assert. I changed the test to be more legible

```
def test_contingency_table():
    im_true = np.array([1, 2, 3, 4])
    im_test = np.array([1, 1, 8, 8])

    table1 = np.array([[0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  ],
       [0.  , 0.25, 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  ],
       [0.  , 0.25, 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  ],
       [0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.25],
       [0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.  , 0.25]])

    sparse_table2 = contingency_table(im_true, im_test, normalize=True)
    table2 = sparse_table2.toarray()
    assert_array_equal(table1, table2)
```

doc/examples/segmentation/plot_metrics.py
rbavery: I changed to `Evaluating segmentation metrics`

doc/examples/segmentation/plot_metrics.py
rbavery: Ah yikes, I changed the figure size and it looks good on my end.

skimage/metrics/_variation_of_information.py
rbavery: Done

jni:
Hi everyone,

apologies that I don't have time for a thorough review right now, but I just wanted to chime in on the module name. imho "metrics" is too precise. Even semi-metrics don't cut it — in some cases, we are measuring similarity, not distance.

imho, we should go with something along the lines of "evaluation", "evaluate", "performance", ... For me, a non-mathematician, "metrics" just sounds too vague. I wouldn't know to look there.

jni:
I'll also note, I don't necessarily want this debate to hold the merge up — I think this is going to be a prolonged debate and we should not clutter the discussion here.

skimage/_shared/testing.py
sciunto: We should not use assert if the check must be done in any situation (including with optimize flags). I would say no.

My opinion on this is to put this function in _shared/utils.py, below `assert_nD`, and rename these functions to do not use the word "assert". Like check_nD, check_shape_equality.

This can be addressed in another PR. These functions are not in the public API, it won't be painful to change this later.

sciunto:
The example looks good for me. I validate this PR as well.

@jni. Please, open an issue for the "evaluation vs metrics" debate.

I'll open another one about the assert thing.

Many thanks @rbavery for your patience!

rbavery:
Thanks for the feedback all!

PR #4026, opened 2019-07-14
fix #4014 gaussian filter output and add test

### LINKED ISSUE #4014: `skimage.filters.gaussian` doesn't use `output` parameter at all
## Description
`skimage.filters.gaussian` doesn't use the value of `output` parameter. It returns only the image of float64 dtype, even if explicitly specify the output dtype. 

## Way to reproduce
Simply run the code snippet:
```python
from skimage.filters import gaussian
import numpy as np

image = np.arange(25, dtype=np.uint8).reshape((5,5))

# output is not specified
filtered_0 = gaussian(image, sigma=0.25)
print(filtered_0.dtype)
# return: float64

# output is specified
filtered_output = gaussian(image, output = np.uint8, sigma=0.25)
print(filtered_output.dtype)
# return: float64
```
This function is a wrapper around `scipy.ndi.gaussian_filter`. But the scikit-image `gaussian` doesn't pass the `output` to `scipy.ndi.gaussian_filter`.

snasibli:
## Description
Closes #4014
`skimage.filters.gaussian` didn't use the value of `output` parameter. It returned only the image of float64 dtype, even if explicitly specify the output dtype. Fix it and add test.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/_gaussian.py
+++ b/skimage/filters/_gaussian.py
@@ -129,6 +129,12 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
             sigma = [sigma] * (image.ndim - 1)
         if len(sigma) != image.ndim:
             sigma = np.concatenate((np.asarray(sigma), [0]))
+    if output is not None and not preserve_range:
+        if isinstance(output, np.ndarray):
+            if issubclass(output.dtype.type, np.integer):
+                preserve_range = True
+        elif issubclass(output, np.integer):
+            preserve_range = True
     image = convert_to_float(image, preserve_range)
-    return ndi.gaussian_filter(image, sigma, mode=mode, cval=cval,
-                               truncate=truncate)
+    return ndi.gaussian_filter(image, sigma, output=output, mode=mode,
+                               cval=cval, truncate=truncate)
--- a/skimage/filters/tests/test_gaussian.py
+++ b/skimage/filters/tests/test_gaussian.py
@@ -68,3 +68,13 @@ def test_4d_ok():
     img[2, 2, 2, 2] = 1
     res = gaussian(img, 1, mode='reflect')
     assert np.allclose(res.sum(), 1)
+
+
+def test_output_type():
+    img = np.arange(16, dtype=np.uint8).reshape((4, 4))
+    output_type = np.uint8
+    gaussian_img = gaussian(img, 1, output=output_type)
+    assert gaussian_img.dtype == output_type
+    output_image = np.zeros_like(img, dtype=output_type)
+    gaussian_img = gaussian(img, 1, output=output_image)
+    assert gaussian_img.dtype == output_type


skimage/filters/_gaussian.py
```diff
@@ -129,6 +129,12 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
             sigma = [sigma] * (image.ndim - 1)
         if len(sigma) != image.ndim:
             sigma = np.concatenate((np.asarray(sigma), [0]))
+    if output is not None and not preserve_range:
+        if isinstance(output, np.ndarray):
+            if issubclass(output.dtype.type, np.integer):
+                preserve_range = True
+        elif issubclass(output, np.integer):
```
hmaarrfk: oh my god, this is so confusing. The fact that output can be both a dtype, or an array is killer.

skimage/filters/_gaussian.py
```diff
@@ -129,6 +129,12 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
             sigma = [sigma] * (image.ndim - 1)
         if len(sigma) != image.ndim:
             sigma = np.concatenate((np.asarray(sigma), [0]))
+    if output is not None and not preserve_range:
```
hmaarrfk: i dont think you need to condition on the value of preserve_range.

skimage/filters/tests/test_gaussian.py
```diff
@@ -67,4 +67,14 @@ def test_4d_ok():
...
-    assert np.allclose(res.sum(), 1)
+    assert np.allclose(res.sum(), 1)
+
+
+def test_output_type():
+    img = np.arange(16, dtype=np.uint8).reshape((4, 4))
+    output_type = np.uint8
+    gaussian_img = gaussian(img, 1, output=output_type)
+    assert gaussian_img.dtype == output_type
+    output_image = np.zeros_like(img, dtype=output_type)
```
hmaarrfk: should you also test the cool logic you gave about perserve range and floats / integer output dtypes?

maybe output as int32 or uint32 instead of the same type of dtype for those?

skimage/filters/_gaussian.py
snasibli: I didn't get why `skimage.filters.gaussian` converting all type images? Is `convert_to_float` really needed for all dtypes? `scipy.ndimage.gaussian_filter` works correctly with all numpy dtypes

hmaarrfk: because images in uint8 format are between 0 and 255 inclusive, but "images" in float32 (and float64) are between 0.0 and 1.0 inclusive.

The only thing that is different is thus the scaling factor. Annoying, but it is pretty standard in most "image" formats.

mixmixmix: @snasibli, are you expecting to finish this PR? I could help out if you're too busy. I'm new to scikit-image but I already found those data types inconsistencies very confusing. Shouldn't there be some common checks for all functions that expect specific types? If every filter has its own type-checking it will end up a big mess, I'm worried.

snasibli: It is needed to understand what it is output.  Is it numpy dtype or allocateed numpy array, or both? In docstring of skimage.filters.gaussian output is expected as an array, but for scipy.ndimage.gaussian_filter it could be as dtype as array.
I agree with that there should be some common checks for all functions that expect specific types. If I will get answers to this questions I finish

snasibli: @hmaarrfk, @mixmixmix can you please, ask the questions on previous comment. I'm going to finish this work)

hmaarrfk: The fact that output can be a dtype or array seems like a strange convention in numpy/scipy functions.. i guess we should follow.

snasibli:
I am closing this PR, because it is no longer relevant. Further work on the issue #4014 I do here #4329

snasibli:
I am closing this PR, because it is no longer relevant.

Issue #3667, opened 2019-01-20
Hide xy/rc warnings when computing regionprops not affected by the change


anntzer:
## Description

As of skimage 0.14.2, the xy-to-rc warning is emitted when computing certain regionprops that are not actually affected by the xy-to-rc change, e.g. the minor/major_axis_length:
```
import numpy as np
from skimage import measure
p, = measure.regionprops(np.array([[0, 1], [1, 0]]))
print(p.major_axis_length)
```
gives
```
$HOME/.local/lib/python3.7/site-packages/skimage/measure/_regionprops.py:250: UserWarning: regionprops and image moments (including moments, normalized moments, central moments, and inertia tensor) of 2D images will change from xy coordinates to rc coordinates in version 0.16.
See http://scikit-image.org/docs/0.14.x/release_notes_and_installation.html#deprecations for details on how to avoid this message.
  warn(XY_TO_RC_DEPRECATION_MESSAGE)
$HOME/.local/lib/python3.7/site-packages/skimage/measure/_regionprops.py:260: UserWarning: regionprops and image moments (including moments, normalized moments, central moments, and inertia tensor) of 2D images will change from xy coordinates to rc coordinates in version 0.16.
See http://scikit-image.org/docs/0.14.x/release_notes_and_installation.html#deprecations for details on how to avoid this message.
  warn(XY_TO_RC_DEPRECATION_MESSAGE)
2.8284271247461903
```
I guess that this is because the computation goes through the computation of moments, which *is* affected by the change, but the computation of the minor/major_axis_length should wrap itself using e.g. warningsfilter to hide that warning.

(I also think that forcing the warning to be emitted *every* time is a misfeature (you can easily end up overwhelmed by the same warning emitted thousands of times...), as argued in https://github.com/scikit-learn/scikit-learn/issues/8182, but that's a separate issue.)

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [X] Code example
- [N/A] Relevant images (if any)
- [X] Operating system and version: Fedora 29
- [X] Python version: 3.7
- [X] scikit-image version (run `skimage.__version__`): 0.14.2

jni:
@anntzer both good points! Would you like to submit a PR to fix this? The two warnings are because one is emitted at the time of creation of the regionprops object, and the other one is emitted (incorrectly, as you point out) at the time of the moments computation. I agree that major/minor axes length computations should filter out the warnings.

anntzer:
[...]

PR #4027, opened 2019-07-14
silence r,c  warnings when property does not depend on r,c

holly-c-gibbs:
Addresses issue #3667 by filtering r,c warning for region properties that don't depend on ordering of r,c (eccentricity, major_axis_length, and minor_axis_length). 
## Checklist

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_regionprops.py
+++ b/skimage/measure/_regionprops.py
@@ -195,9 +195,9 @@ class RegionProperties:
                           for i in range(self._ndim)]).T
 
     @only2d
-    def eccentricity(self): 
+    def eccentricity(self):
         with warnings.catch_warnings():
-            warnings.filterwarnings("ignore","regionprops and image moments")
+            warnings.filterwarnings('ignore', 'regionprops and image moments')
             l1, l2 = self.inertia_tensor_eigvals
         if l1 == 0:
             return 0
@@ -265,13 +265,13 @@ class RegionProperties:
 
     def major_axis_length(self):
         with warnings.catch_warnings():
-            warnings.filterwarnings("ignore","regionprops and image moments")
+            warnings.filterwarnings('ignore', 'regionprops and image moments')
             l1 = self.inertia_tensor_eigvals[0]
         return 4 * sqrt(l1)
 
     def minor_axis_length(self):
         with warnings.catch_warnings():
-            warnings.filterwarnings("ignore","regionprops and image moments")
+            warnings.filterwarnings('ignore', 'regionprops and image moments')
             l2 = self.inertia_tensor_eigvals[-1]
         return 4 * sqrt(l2)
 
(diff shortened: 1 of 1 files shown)


skimage/measure/_regionprops.py
```diff
@@ -194,7 +195,7 @@ def coords(self):
                           for i in range(self._ndim)]).T
 
     @only2d
-    def eccentricity(self):
+    def eccentricity(self): #suppress warning
```
jni: @holly-c-gibbs I think you meant to use this comment as a to-do. The line underneath should also be silenced.

jni:
@scikit-image/core in offline chats with @holly-c-gibbs, I initially thought it was a bit of a waste of time to silence the warnings since the whole xy machinery should go away in 0.16. However I think it's valuable to do it in two steps, and backport to 0.14 and 0.15, and then remove the whole thing altogether.

hmaarrfk:
This seems like it should be addressed in `inertia_tensor_eigvals` no?

jni:
@hmaarrfk Ach, sorry we were too fast. I kinda wondered about that. Are the eigvals indeed invariant to the coordinate convention?

alexandrejaguar:
Thank you @holly-c-gibbs!

hmaarrfk:
seems like that is all it is doing right? finding  when the eigenvalues are computed and going off that.

hmaarrfk:
It is unclear if this is even helping the situation.

I'm not able to parse where the warning is comming from.
https://github.com/scikit-image/scikit-image/blob/v0.15.x/skimage/measure/_regionprops.py#L249

You all see to be in pysical proximity, but this behaviour change didn't come with any tests that would indicate that the underlying challenge is fixed. A test that uses: `with expected_warnings(None)` would have done well here.

jni:
@hmaarrfk thanks for the review. Actually it seems the warning was already removed from master. :man_facepalming: Will fix, thanks for the help!

hmaarrfk:
I would revert on master, and target the PR directly to the 0.14 branch. Personally I don't think 0.15.1 is happening.

If it is happening, we can likely cherrypick the commit from 0.14.x

jni:
:+1:

hmaarrfk:
If you want help, I can try and get things building on 0.15 again. But it is going to require significant backporting I think.

jni:
don't stress. =)

hmaarrfk:
THere isn't any stress, just trying to give constructive critisms as opposed to: "this isn't the right approach", and offering to help where I can.

jni:
I'm happy to push on 0.16 in general. We want to get this Quiet Edition of skimage out ASAP! =)

hmaarrfk:
yeha, but RC is a big change, and will likely break stuff.

hmaarrfk:
no matter how hard we try not to.

Issue #4038, opened 2019-07-21
importing scikit-image is too slow


hmaarrfk:
## Description
It takes about 0.5 s!!!

![image](https://user-images.githubusercontent.com/90008/61595643-ab29dd80-abc7-11e9-8840-36ea6f4aa6ea.png)

## Way to reproduce
```python
import skimage.util
```

```
$ python -X importtime profile.py 2> import.log
$ python -X importtime profile.py 2> import.log
$ tuna import.log
```

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21)
[GCC 7.3.0]
Linux-5.0.0-20-generic-x86_64-with-debian-buster-sid
scikit-image version: 0.15.0
numpy version: 1.16.4
```

sciunto:
I guess after the merge of 4039, we can close this issue, although optimization is an infinite task... :)

hmaarrfk:
Well now it is limited by numpy. I'm working with them to speed up their lib too ;)

Issue #3397, opened 2018-09-10
Audit our code import structure


jni:
## Description
I think we have become a little too lackadaisical in how we reuse our code from different modules *across* modules. Things used across several modules (segmentation, morphology, etc) should probably be moved to "util".

Here is a [user's](https://stackoverflow.com/questions/52242859/file-e-python-lib-re-py-line-229-in-finditer-return-compilepattern-flags) traceback (from SO) after trying to import `feature.greycomatrix`:

```pytb
Traceback (most recent call last):
  File "C:\Users\Martin Ma\Desktop\every day python\test\GLCM\main.py", line 80, in <module>
    from skimage.feature import greycomatrix
  File "E:\python\lib\site-packages\skimage\feature\__init__.py", line 9, in <module>
    from .peak import peak_local_max
  File "E:\python\lib\site-packages\skimage\feature\peak.py", line 3, in <module>
    from ..segmentation import relabel_sequential
  File "E:\python\lib\site-packages\skimage\segmentation\__init__.py", line 1, in <module>
    from .random_walker_segmentation import random_walker
  File "E:\python\lib\site-packages\skimage\segmentation\random_walker_segmentation.py", line 43, in <module>
    from ..filters import rank_order
  File "E:\python\lib\site-packages\skimage\filters\__init__.py", line 3, in <module>
    from .edges import (sobel, sobel_h, sobel_v,
  File "E:\python\lib\site-packages\skimage\filters\edges.py", line 17, in <module>
    from ..restoration.uft import laplacian
  File "E:\python\lib\site-packages\skimage\restoration\__init__.py", line 12, in <module>
    from .inpaint import inpaint_biharmonic
  File "E:\python\lib\site-packages\skimage\restoration\inpaint.py", line 9, in <module>
    from ..measure import label
  File "E:\python\lib\site-packages\skimage\measure\__init__.py", line 6, in <module>
    from ._regionprops import regionprops, perimeter
  File "E:\python\lib\site-packages\skimage\measure\_regionprops.py", line 645, in <module>
    _install_properties_docs()
  File "E:\python\lib\site-packages\skimage\measure\_regionprops.py", line 632, in _install_properties_docs
    prop_doc = _parse_docs()
  File "E:\python\lib\site-packages\skimage\measure\_regionprops.py", line 625, in _parse_docs
    doc, flags=re.DOTALL)
  File "E:\python\lib\re.py", line 229, in finditer
    return _compile(pattern, flags).finditer(string)
[...]

PR #4039, opened 2019-07-21
Improve import time

hmaarrfk:
## Description
This delays almost all imports from the critical import path of scikit image.

Now, on a "good" install, only numpy is imported, all other libraries are now lazy imported and only used when the user needs them in their code.

This decreases the total import time from 0.7 s to 0.13 s. Most of this is due to importing numpy.
Before:
image

After:
image

Ref: #4038

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_import_time.py
+++ b/benchmarks/benchmark_import_time.py
@@ -3,14 +3,18 @@ from sys import executable
 
 class ImportSuite:
     """Benchmark the time it takes to import various modules"""
-    def setup(self):
+    params = [
+        'numpy',
+        'skimage',
+        'skimage.feature',
+        'skimage.morphology',
+        'skimage.color',
+    ]
+    param_names = ["package_name"]
+    def setup(self, package_name):
         pass
 
-    def time_import_numpy(self):
-        results = run(executable + ' -c "import numpy"',
-            stdout=PIPE, stderr=PIPE, stdin=PIPE, shell=True)
-
-    def time_import_skimage(self):
-        results = run(executable + ' -c "import skimage"',
+    def time_import(self, package_name):
+        results = run(executable + ' -c "import ' + package_name + '"',
             stdout=PIPE, stderr=PIPE, stdin=PIPE, shell=True)
 
--- a/benchmarks/benchmark_import_time.py
+++ b/benchmarks/benchmark_import_time.py
@@ -1,0 +1,16 @@
+from subprocess import run, PIPE
+from sys import executable
+
+class ImportSuite:
+    """Benchmark the time it takes to import various modules"""
+    def setup(self):
+        pass
+
+    def time_import_numpy(self):
+        results = run(executable + ' -c "import numpy"',
+            stdout=PIPE, stderr=PIPE, stdin=PIPE, shell=True)
+
+    def time_import_skimage(self):
+        results = run(executable + ' -c "import skimage"',
+            stdout=PIPE, stderr=PIPE, stdin=PIPE, shell=True)
+
--- a/skimage/_shared/_warnings.py
+++ b/skimage/_shared/_warnings.py
@@ -1,7 +1,6 @@
 from contextlib import contextmanager
 import sys
 import warnings
-import inspect
 import re
 
 __all__ = ['all_warnings', 'expected_warnings', 'warn']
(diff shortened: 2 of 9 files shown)


stefanv:
Is this really worth it? We're bound to just do it again.

hmaarrfk:
Thats fine, and I'll gladly make the PR to get it back to this level again. We do all the effort to keep the namespace clean, but we import much of scipy to create optional datasets for examples. It really does take too long to import scikit-image and I am basically having to drop it as one of my dependencies due to that.

I can work on adding a benchmark if that helps.

jni:
Thanks @hmaarrfk! I support this work, and think we should definitely not be so lackadaisical about this sort of stuff. See also #3397.

With regards to @stefanv's point that you will be forever maintaining this, perhaps adding a benchmark for imports would be a good idea?

My main concern is about how unpythonic this all is (PEP8 puts all imports at the top, which I like), and it makes me wonder whether we can do better by fiddling with `skimage/__init__.py` instead.

jni:
(but note that I still might be in favour of merging this as-is, just need to think about it a bit more!)

hmaarrfk:
i totally understand the desire not to have random imports everywhere. Unfortunately, certain things like `platform` take a non-negligeable amount of time to import.

I think you had an issue where we talked about the craziness in the dependencies.

I'm trying to write a benchmark for this, but it isn't as easy as I thought ;)

hmaarrfk:
some of the other ones are: importing dask array for a convenience tutorial function.

That is fine, but that increases the import time of scikit-image by about 200 ms, which is non negligeable. This will happen when people download distributions such as anaconda, and find `scikit-image` as a result of combined packages being installed at once.

In terms of cost to benefit analysis, I think this is well worth it at this point to keep the package light, especially for the `img_as_*` functions, which are pretty core to any analysis.

Here are some results:

master
```
[100.00%] ··· ============== ===========
               package_name
              -------------- -----------
                  numpy       126±0.9ms
                 skimage       498±2ms
              ============== ===========
```

this branch:

```
[ 75.00%] ··· ============== =========
               package_name
              -------------- ---------
                  numpy       126±1ms
                 skimage      128±1ms
              ============== =========
```

```

     [cba1798f]       [3e5f1a08]
     <master>         <improve_import_time~1>
-        579±90ms         389±30ms     0.67  benchmark_import_time.ImportSuite.time_import('skimage.feature') [mark-xps/conda-py3.7-cython-numpy1.16-scipy]
-         597±8ms         363±20ms     0.61  benchmark_import_time.ImportSuite.time_import('skimage.feature') [mark-xps/conda-py3.7-cython-numpy1.15-scipy]
-        564±20ms         326±20ms     0.58  benchmark_import_time.ImportSuite.time_import('skimage.morphology') [mark-xps/conda-py3.7-cython-numpy1.15-scipy]
-        586±70ms         324±10ms     0.55  benchmark_import_time.ImportSuite.time_import('skimage.morphology') [mark-xps/conda-py3.7-cython-numpy1.16-scipy]
-        509±20ms          162±7ms     0.32  benchmark_import_time.ImportSuite.time_import('skimage.color') [mark-xps/conda-py3.7-cython-numpy1.15-scipy]
-        580±40ms          166±3ms     0.29  benchmark_import_time.ImportSuite.time_import('skimage.color') [mark-xps/conda-py3.7-cython-numpy1.16-scipy]
-         570±4ms          131±5ms     0.23  benchmark_import_time.ImportSuite.time_import('skimage') [mark-xps/conda-py3.7-cython-numpy1.16-scipy]
-         606±5ms          132±2ms     0.22  benchmark_import_time.ImportSuite.time_import('skimage') [mark-xps/conda-py3.7-cython-numpy1.15-scipy]

```

hmaarrfk:
Numpy seems to have their own benchmarks, though they don't display quite nicely.
https://pv.github.io/numpy-bench/#bench_import.Import.time_fft

Finally, i don't think their numbers are quite accurate, if it really does take 1 second to start up numpy, that is REALLY slow.

hmaarrfk:
@jni, i wish I knew how to use `Popen`, then we could try `preloading` certain parts of the python standard library and running only the relevant import statement in the benchmark. but as is now, the benchmark shows that we are basically limited by importing numpy, which is fine since this is a `numpy` first library.

hmaarrfk:
https://snarky.ca/lazy-importing-in-python-3-7/

stefanv:
I didn't realize we were up to 0.5s already—that is quite painful.  Pity about the style, but I don't see an easy way around it, other than messing with hacks like lazy import (https://pypi.org/project/lazy-import/).

stefanv:
And, yes, this is a great usage of benchmarks!

sciunto:
I (re?)discovered tuna, thanks for this @hmaarrfk

PR #4042, opened 2019-07-22
Undo lazy import in arraycrop

hmaarrfk:
## Description

@stefanv sorry, for this one, @jni was right that some of these were probably a little frivolous and added much  stylistic noise.

You merged as soon as I tried to push this change ;)

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/util/arraycrop.py
+++ b/skimage/util/arraycrop.py
@@ -7,6 +7,13 @@ import numpy as np
 
 __all__ = ['crop']
 
+from distutils.version import LooseVersion as Version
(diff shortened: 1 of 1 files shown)


jni:
@hmaarrfk Actually, I recommend closing this one, as it has been on my todo list to fix this function: we should *not* be using private NumPy functions from any version. The original implementation had the `_validate_lengths` implementation vendored *as a comment*. imho we should just vendor that and get rid of this complicated version logic, that might still have us putting out 0.14.2-style fires in the future. If you agree, please just close this PR.

PR #4053, opened 2019-07-25
fix mpl future deprecation normed -> density

### LINKED ISSUE #3430: Handle `normed` to `density` deprecations in `histogram`
## Description
See https://github.com/scikit-image/scikit-image/pull/3427 and the related issues.

sciunto:
## Description

I come back on a long standing issue we have with a mpl kwarg. See #3430 and the related issues. At some point, we were in a nightmare with this and we decided to wait.

Here, I try to just change the keyword and see what happen. On my machine, it is passing well.

Closes #3430


jni:
🤞

sciunto:
tests are passing :vic:

PR #4063, opened 2019-07-29
Critical: address internal warnings in test suite related to metrics 4025

hmaarrfk:
## Description
I think this PR should act as proof that we shouldn't hard deprecate the old names for these functions. 

It touches a ridiculous number of lines of code, just because we decided to change the name of a function.
It seems a little "flavor of the week" to me. If it is really just internal reorganization that we want, we shouldn't hard impose this on users...

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/io/tests/test_pil.py
+++ b/skimage/io/tests/test_pil.py
@@ -9,7 +9,6 @@ from .. import imread, imsave, use_plugin, reset_plugins
 from PIL import Image
 from .._plugins.pil_plugin import (
     pil_to_ndarray, ndarray_to_pil, _palette_is_grayscale)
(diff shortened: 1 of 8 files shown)


stefanv:
> Thanks @hmaarrfk!
> 
> Re deprecations, it's a good point. I'd consider function renames for extended deprecations e.g. to 1.0. @scikit-image/core thoughts?

Is it any less painful to have users change the function call a year later? Perhaps the question is rather whether you should ever deprecate after renaming.  Having an alias in place is very cheap for us, as long as we ensure that it is not listed in the documentation.

PR #4064, opened 2019-07-29
Test rank capture warnings in threadsafe manner

### LINKED ISSUE #4052: travis has random failures on rank filters
## Description

See for ex: https://travis-ci.org/scikit-image/scikit-image/jobs/563363217

## Way to reproduce
```python
```

## Version information
```python
```

```python
# your output here

```

hmaarrfk:
## Description
Fixes #4052 

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/testing.py
+++ b/skimage/_shared/testing.py
@@ -235,7 +235,9 @@ def test_parallel(num_threads=2, warnings_matching=None):
         @functools.wraps(func)
         def inner(*args, **kwargs):
             if warnings_matching is None:
-                # dummy context manager
+                # The easiest dummy context manager available for python 3.2+
+                # we still support python older than 3.7
+                # https://stackoverflow.com/questions/45187286/how-do-i-write-a-null-no-op-contextmanager-in-python
                 ctx = memoryview(b'')
             else:
                 ctx = expected_warnings(warnings_matching)
--- a/skimage/_shared/testing.py
+++ b/skimage/_shared/testing.py
@@ -234,14 +234,7 @@ def test_parallel(num_threads=2, warnings_matching=None):
     def wrapper(func):
         @functools.wraps(func)
         def inner(*args, **kwargs):
-            if warnings_matching is None:
-                # The easiest dummy context manager available for python 3.2+
-                # we still support python older than 3.7
-                # https://stackoverflow.com/questions/45187286/how-do-i-write-a-null-no-op-contextmanager-in-python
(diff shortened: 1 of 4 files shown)


skimage/_shared/testing.py
```diff
@@ -220,27 +221,39 @@ def test_parallel(num_threads=2):
...
-            for thread in threads:
-                thread.join()
-
-            return result
+            if warnings_matching is None:
+                # dummy context manager
+                ctx = memoryview(b'')
+            else:
+                ctx = expected_warnings(warnings_matching)
+            with ctx:
```
jni: This works??? Can you add links to documentation about this (the dummy context) in a comment? Because my mind is being blown.

Additionally, perhaps we can make expected warnings behave correctly with None input, which would simplify this code dramatically?

hmaarrfk: https://stackoverflow.com/a/54786924

I swear I benchmarked this and memory view as the fastest nullcontext manager. I don't understand why python 3.7 chose anything else. Maybe they offer more debugging messages??

hmaarrfk: What behaviour do you want? the no-op if None is passed?

skimage/_shared/testing.py
```diff
@@ -220,27 +221,39 @@ def test_parallel(num_threads=2):
...
-                thread.start()
-
-            result = func(*args, **kwargs)
-
-            for thread in threads:
-                thread.join()
-
-            return result
+            if warnings_matching is None:
+                # dummy context manager
```
hmaarrfk:
```suggestion
                # The easiest dummy context manager available for python 3.2+
                # we still support python older than 3.7
                # https://stackoverflow.com/questions/45187286/how-do-i-write-a-null-no-op-contextmanager-in-python
```

skimage/_shared/testing.py
jni: Yes: basically make `expected warnings` return that null memoryview when None is the input.

hmaarrfk: This is in contrast to the behaviour that is done now with `[None]`, i forget what that does.

hmaarrfk: i think it jsut skips over them.

jni:
(The macOS errors are homebrew errors but they seem to be becoming more frequent... :confused:)

soupault:
Let's see if it helps! Thanks Mark!

PR #4079, opened 2019-08-02
docs update for downscale_local_mean and N-dimensional images

i3v:
## Description
Currently:
* `downscale_local_mean` documentation [says](https://github.com/scikit-image/scikit-image/blob/87830b081516b93dcb2e4a883483b4d895232066/skimage/transform/_warps.py#L396):
   >  In contrast to the 2-D interpolation in `skimage.transform.resize`

  which looks like `skimage.transform.resize` only actually works for 2D images. (Actually, higher-dimensional interpolation is possible since about 3b67c8204255757391b3a4f76e4cc778a6ad5072 .)

* `resize` & `rescale` documentation, effectively, only explicitly [mentions](https://github.com/scikit-image/scikit-image/blob/87830b081516b93dcb2e4a883483b4d895232066/skimage/transform/_warps.py#L25-L26) N-dimensional images when sending user to `downscale_local_mean` :
  > For down-sampling N-dimensional images with an integer factor also see `skimage.transform.downscale_local_mean`.  

   (I'm not saying this part is is not correct, but... This proved to be a bit misleading for a new user looking for a 3D image resize method, at the first glance.)

The minor changes, suggested here aim to resolve these ambiguities.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_warps.py
+++ b/skimage/transform/_warps.py
@@ -20,9 +20,9 @@ def resize(image, output_shape, order=1, mode='reflect', cval=0, clip=True,
            preserve_range=False, anti_aliasing=True, anti_aliasing_sigma=None):
     """Resize image to match a certain size.
 
-    Performs interpolation to up-size or down-size images. Note that anti-
+    Performs interpolation to up-size or down-size N-dimensional images. Note that anti-
     aliasing should be enabled when down-sizing images to avoid aliasing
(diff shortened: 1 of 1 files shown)


i3v:
@jni , thanks for merging!

Issue #4043, opened 2019-07-23
task: unify FFT usage across scikit-image


grlee77:
## Description

There are a few issues related to the usage of FFTs in scikit-image that could use some maintainance. I will try to make PR(s) to address these in the near future. There are a few separate issues:

**1.)  unify FFT modules used**
FFTs are called from a variety of sources across the library.:

Functions from `scikit.fftpack` are called in:
    `skimage/transform/radon_transform.py`

Functions from `numpy.dual` are called in:
    `skimage/filters/lpi_filter.py`

(These `numpy.dual` call SciPy's fft when scipy.fftpack has been imported. Otherwise they fall back to the numpy implementations)

A number of other files/functions just call the functions from `numpy.fft` directly.

Unless there is a compelling reason otherwise, we should try to use a single convention.

**2.) replace any unitary FFT code that is no longer needed**

The unitary FFTs in `skimage.restoration.uft` seem to implement functionality that has since been implemented directly in NumPy / SciPy via the `norm='ortho'` argument (introduced in NumPy 1.10).

**3.) Take advantage of the new scipy.fft module when available**
As a result of a GSOC 2019 project, SciPy 1.4.0 will have a new `scipy.fft` module that has the same API as `scipy.fftpack`, but has a number of accuracy and performance improvements. [More details here](https://github.com/scipy/scipy/wiki/Release-note-entries-for-SciPy-1.4.0#scipyfft-added)

I think we should try to use these `scipy.fft` variants when available because they make it easy for the user to swap out the FFT backend to something like pyFFTW when desired via a simple context manager or a call to `scipy.fft.set_global_backend`. An example of how that would look can be seen at: https://github.com/pyFFTW/pyFFTW/pull/269#issuecomment-512336676

jni:
@grlee77 brilliant! I very much like this task. =)

sciunto:
Might be of interest: #3735

grlee77:
Regarding items 1 and 3 above. I think there are a few possible options. I am interested to hear what other devs here prefer prior to working on this.

[...]

PR #4084, opened 2019-08-05
Make use of FFTs more consistent across the library

grlee77:
## Description

This is an implementation of #4043

This PR prioritizes `scipy.fft` when it is found (SciPy >= 1.4), otherwise `numpy.fft` will be used. I did this by adding a new file: `skimage/_shared/fft.py` where the logic for this is located. I wasn't sure whether to just import a module within that file or to import all of the individual functions used. 

There was one file, `radon_transform.py `, which was previously using `scipy.fftpack` instead of NumPy. I was not entirely certain if that was a deliberate choice, but to minimize the chance of breaking changes, I have that one file fall back to `scipy.fftpack` rather than `numpy.fft` on older SciPy.

**Note:** I don not think we currently include any SciPy weekly development builds on CI so we should not merge until we add that or SciPy has a release candidate for 1.4.  Until scipy/scipy#10586 is merged, there are expected to be a few test failures with current SciPy master. 

## Checklist

- [ N/A ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ N/A ] Gallery example in `./doc/examples` (new features only)
  existing benchmark
- [x] Unit tests  **relies on the existing tests for each function**

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/fft.py
+++ b/skimage/_shared/fft.py
@@ -1,0 +1,20 @@
+"""Prefer FFTs via the new scipy.fft module when available (SciPy 1.4+)
+
+Otherwise fall back to numpy.fft.
+"""
+try:
+    import scipy.fft
+    from scipy.fft import next_fast_len
+    fftmodule = scipy.fft
+except ImportError:
+    import numpy.fft
+    fftmodule = numpy.fft
+
+    # next_fast_len was implemented in scipy 0.18
+    # In case it cannot be imported, we use the id function
+    try:
+        from scipy.fftpack import next_fast_len
+    except ImportError:
+        next_fast_len = lambda size: size
+
+__all__ = ['fftmodule', 'next_fast_len']
--- a/skimage/_shared/fft.py
+++ b/skimage/_shared/fft.py
@@ -1,6 +1,9 @@
 """Prefer FFTs via the new scipy.fft module when available (SciPy 1.4+)
 
 Otherwise fall back to numpy.fft.
+
+Like numpy 1.15+ scipy 1.3+ is also using pocketfft, but a newer 
+C++/pybind11 version called pypocketfft
 """
 try:
     import scipy.fft
--- a/skimage/_shared/fft.py
+++ b/skimage/_shared/fft.py
@@ -12,9 +12,9 @@ except ImportError:
 
     # next_fast_len was implemented in scipy 0.18
     # In case it cannot be imported, we use the id function
-    try:
-        from scipy.fftpack import next_fast_len
-    except ImportError:
-        next_fast_len = lambda size: size
+    def next_fast_len(size):
+        """Dummy next_fast_len that returns size unmodified."""
+        return size
+
 
 __all__ = ['fftmodule', 'next_fast_len']
--- a/skimage/_shared/fft.py
+++ b/skimage/_shared/fft.py
@@ -10,11 +10,13 @@ except ImportError:
     import numpy.fft
     fftmodule = numpy.fft
 
-    # next_fast_len was implemented in scipy 0.18
-    # In case it cannot be imported, we use the id function
-    def next_fast_len(size):
-        """Dummy next_fast_len that returns size unmodified."""
-        return size
-
+    try:
+        from scipy.fftpack import next_fast_len
+    except ImportError:
+        # next_fast_len was implemented in scipy 0.18
+        # In case it cannot be imported, we use the id function
+        def next_fast_len(size):
+            """Dummy next_fast_len that returns size unmodified."""
+            return size
 
 __all__ = ['fftmodule', 'next_fast_len']
--- a/skimage/transform/radon_transform.py
+++ b/skimage/transform/radon_transform.py
@@ -1,9 +1,18 @@
 import numpy as np
-from scipy.fftpack import fft, ifft, fftfreq
+
 from scipy.interpolate import interp1d
 from ._warps_cy import _warp_fast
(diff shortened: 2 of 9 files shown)


skimage/transform/radon_transform.py
```diff
@@ -1,10 +1,19 @@
...
 from scipy.interpolate import interp1d
 from ._warps_cy import _warp_fast
 from ._radon_transform import sart_projection_update
+from .._shared.fft import fftmodule
 from warnings import warn
 
+if fftmodule is np.fft:
+    # fallback from scipy.fft to scipy.fftpack instead of numpy.fft
+    # (fftpack preserves single precision while numpy.fft does not)
+    from scipy.fftpack import fft, ifft
```
hmaarrfk: why not always fallback to this?

skimage/_shared/fft.py
```diff
@@ -0,0 +1,20 @@
+"""Prefer FFTs via the new scipy.fft module when available (SciPy 1.4+)
+
+Otherwise fall back to numpy.fft.
+"""
+try:
+    import scipy.fft
+    from scipy.fft import next_fast_len
```
hmaarrfk: Should we prefer pocketfft from numpy instead?

https://docs.scipy.org/doc/numpy/release.html

grlee77: No, because `scipy.fft` is also now using pocketfft, but a newer C++/pybind11 version called pypocketfft. One difference with NumPy's is that the scipy one does transforms in either `float`, `double` or `long double` precision while NumPy's pocketfft is `double` only.

see some details on the new `fft` module in SciPy here.
https://github.com/scipy/scipy/wiki/Release-note-entries-for-SciPy-1.4.0#scipyfft-added

hmaarrfk: ok. well this is really confusing (but totally unreated to this PR). I wish they could have just harmonized the two.

hmaarrfk:
This is a really minor thing, but I would really like for https://github.com/scikit-image/scikit-image/pull/4082

to be merged in first, then restarting the tests.

skimage/transform/radon_transform.py
grlee77: 1.) NumPy's pypocketfft will be MUCH faster than old `scipy.fftpack`, at least for non-favorable sizes (e.g. not factors of 2, 3, 5, 7 etc).
2.) `scipy.fft` follows numpy's API much more closely than `scipy.fftpack`. e.g. `fftpack`'s `rfft`/`irfft` do not organize data in the same way as NumPy's so the places in other files that were using those cannot just switch to `fftpack` as a drop-in replacement for the existing code.

grlee77: For `radon`/`iradon` I did not look closely enough to see if single-precision is actually being used. If everything in this file gets promoted to `float64` anyways then we probably don't need to special case here and can just fallback to `numpy.fft` as done elsewhere.

hmaarrfk: man, i feel you pain in making this PR. well thought out. Thanks for the explanation. 

I wish we could just bumpt all the way up to scipy 1.4 for you ;)

grlee77:
> This is a really minor thing, but I would really like for #4082 to be merged in first, then restarting the tests.

Sure, that is fine. I have marked this as WIP so it hopefully won't get merged until we get a SciPy 1.4 test case on CI.

hmaarrfk:
I see, yeah, scipy doesn't release pre-releases on pypi, so even the `--pre` doesn't pick them up.
https://travis-ci.org/scikit-image/scikit-image/jobs/567688450#L937

Builds should be up and running soon.

skimage/_shared/fft.py
grlee77: Well, `scipy.fft` should now be harmonized with `numpy.fft` aside from differences in dtype support and the fact that `scipy.fft` offers an additional, optional "overwrite_x" argument. Unfortunately we have to still keep `scipy.fftpack` around indefinitely for all the legacy code that is using it. The FFTPACK Fortran code has actually been removed in 1.4 (the fftpack API is still there, but is implemented via pypocketfft).

peterbell10: >ok. well this is really confusing (but totally unreated to this PR). I wish they could have just harmonized the two.

NumPy doesn't accept C++ code so that's not really possible.

skimage/_shared/fft.py
```diff
@@ -0,0 +1,20 @@
+"""Prefer FFTs via the new scipy.fft module when available (SciPy 1.4+)
```
sciunto: Please, could you add an entry in TODO.txt to update this code when our min req will be 1.4+?

Thank you!

sciunto: In that file, you can also remove one entry
```
"* When minimum supported version of ``scipy`` reaches 0.18, use
  ``scipy.fftpack.next_fast_len`` directly in
  ``skimage.feature.masked_register_translation``."
```

skimage/_shared/fft.py
hmaarrfk: Ah I see. That's really too bad

skimage/_shared/fft.py
grlee77: I restored the import from scipy.fftpack.next_fast_len for scipy < 0.18 and have just updated the path in that entry (see commit 753204c)

hmaarrfk:
@grlee77 seems like builds are passing. Feel free to merge if you are satisfied.

PR #4035, opened 2019-07-20
Deprecate xy coordinates in segmentation.active_contours

jni:
## Description

As I delivering @JDWarner's segmentation tutorial, someone asked how to go from the coordinate outputs to a segmentation masked, and I noticed that the coordinates returned by `active_contour` were actually in x/y format rather than rc. Our brand new poly2mask function produces a transposed mask from the output, for example.

This PR provides a keyword argument to reverse that convention, as well as a deprecation path to make rc the default. I also took the opportunity to clean up the API a little, replacing `bc` with `boundary_condition`. 

## Checklist

- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Gallery example in `./doc/examples` (new features only)
- [ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
- [ ] Unit tests
- [x] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

## For reviewers

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

PR #4092, opened 2019-08-15
Bugfix for documentation example: plot_active_contours.py

dmontagu:
Fixes the example at https://scikit-image.org/docs/dev/auto_examples/edges/plot_active_contours.html.

(I've tested it against 0.15.0; I don't know which version is used to generate the docs but I got the same incorrect result that is shown in the docs using 0.15.0 after fixing the keyword argument names.)

The inputs currently being used to generate the (linked) docs are transposed, causing the examples to look nonsensical at first. It looks like it is probably due to the removal of a `coordinates` argument.

The textual description already used the right argument names (which required changes), so no text needed updating.

## Description

Example bug fix

## Checklist

(Very few changes, so I'm assuming the below can all be checked.)
  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/edges/plot_active_contours.py
+++ b/doc/examples/edges/plot_active_contours.py
@@ -39,11 +39,10 @@ img = rgb2gray(img)
 s = np.linspace(0, 2*np.pi, 400)
 r = 100 + 100*np.sin(s)
 c = 220 + 100*np.cos(s)
-init = np.array([r, c]).T
+init = np.array([c, r]).T
 
 snake = active_contour(gaussian(img, 3),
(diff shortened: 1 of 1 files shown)


dmontagu:
Oh, I guess this was due to a recent change https://github.com/scikit-image/scikit-image/pull/4035; I'll close it.

jni:
Apologies for the confusion, @dmontagu. I've made an issue on our site repo to try to improve our docs to better indicate version differences:

https://github.com/scikit-image/scikit-image-web/issues/61

PR #4095, opened 2019-08-15
Regionprops memory usage

rdgraham:
## Description

I have finally managed to prepare branch with changes related to improving memory usage in regionprops and some other refactoring. This was partially discussed earlier in mailing list thread [1] 'strange style in regionprops' (Apr 2019).

The main changes in this branch:
* Added the ability to cache a packed version of the sliced mask for the region on instantiation when caching is enabled.
* Added method for retrieving full mask in addition to bounded mask (see below)
* Changed all properties to conventional style using the @property decorator, with documentation located in the method docstrings.
* One other minor change. Added a total_intensity property. There were already properties for mean, min and max, this one seems to have been overlooked.

Reasons for the changes and details are as follows:

* As has been raised before [2], code using regionprops often ends up using a lot of memory, more so than other solutions. It seems to me that the main reason for this is that each region continues to maintain a reference to the original labeled_image. In most cases, the labeled image is no longer needed after instantiating the regionprops. With this change, enabling caching will cause the image to be sliced and a mask cached upon instantiation, thus allowing the the labeled image to be de-referenced if it is not used elsewhere. While it does mean a tiny bit of extra memory is used per region, this is minimal (boolean array for the small individual regions is packed) and the overall result is generally a significant reduction in memory usage for the application as a whole.
* One other thing that doesn't help is that measure.label returns a 32-bit data type, which is almost always going to be an overkill. I elected to simply add a note about this to the docstring to suggest manually casting to a smaller type. In the future method might be modified to work natively with smaller data type, but looks like a bit of work to do properly.
* One consequence of this change is that the original labeled image is no longer necessarily available. This was always a hidden/private instance variable anyway, so users shouldn't access it. I searched the internet for any examples of people accessing it despite this. The only example I could find was internally, in the unit tests for regionprops. So I have gone ahead and re-factored the unit tests to work without the labeled image.
* There is however one case I could think of where it would be useful to have the labeled_image. This is for recovering a mask of the region with dimensions of the original image. So I created a new method for this. With now two similar methods for getting the region mask available, I think the original naming is very confusing. So I used the names 'bounded_mask' and 'full_mask' for the bounding box-sized mask (originally just called 'image') and full image respectively. The original name 'image' is also retained for backwards compatibility.
* The other change I made was to go through all the properties and re-factor to use @property decorators to a conventional style with the docstrings for the methods located as usual in the methods. The documentation for each method had previously been located in an add-hoc way inside the documentation for the regionprops function. A separate function used a regex to parse all this and dynamically put it inside the methods __doc__ attribute whenever the module was imported. Weirdly, this same loop was actually used to enable property access to these methods (I assume this code dates to before property decorators were available in python?). Anyway, this caused a nasty bug where if python was executed in optimized  mode (-O flag) then property access would not be available! Almost any code using the class in the 'normal' way would break. All of that is removed now and all doc-strings updated.

[1] https://mail.python.org/archives/list/scikit-image@python.org/thread/RPFPAIWSJQB62W4RCMHKVY6A7AQR73WO/#FM42JSYZVL7FJII3XGICRDSRDAYIBD4N

[2] https://mail.python.org/pipermail/scikit-image/2017-July/005313.html

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


jni:
@rdgraham thanks very much for this contribution! I'm currently on holidays so I don't have time for a thorough review, but others might jump in. Either way on first glance it looks really great!

The main worry I have is whether this affects speed. Ideally, we would have benchmarks to measure these operations, make sure the caching is working properly, etc. @hmaarrfk wrote a fantastic guide on writing and using benchmarks [here](https://scikit-image.org/docs/dev/contribute.html#benchmarks). If you are so inclined, it would be awesome to have some benchmarks as part of this PR, but if it's too much work, don't worry, we'll take it from here.

hmaarrfk:
Regarding the property decorator, this is being addressed in
https://github.com/scikit-image/scikit-image/pull/3934

How the docs are added in, well that is likely a decision for an other day. I would stick to the performance gains in this PR, and avoid the property decorator changes.

The benchmarks will help us understand the use cases you are consider (image sizes, dtypes, dimentionality)

skimage/measure/tests/test_regionprops.py
```diff
@@ -423,104 +434,21 @@ def test_invalid_size():
...
-                   'bbox+2': array([10]), 'bbox+3': array([18])}
-
-
-def test_regionprops_table():
-    out = regionprops_table(SAMPLE)
-    assert out == {'label': array([1]),
-                   'bbox-0': array([0]), 'bbox-1': array([0]),
-                   'bbox-2': array([10]), 'bbox-3': array([18])}
-
-    out = regionprops_table(SAMPLE, properties=('label', 'area', 'bbox'),
```
hmaarrfk: Why are these tests removed?

rdgraham: Ah, that was my mistake. Messed up when merging and lost the recently added code. I have now added all these new tests back (except for old test of parsing of the documentation). All tests now passing.

In doing so, I found a nasty bug. I noticed in the new testers for the regionprops_table and dict that it was asserting the area of the SAMPLE global as 180. This is clearly incorrect and the area should be 72. It turns out this was happening because the old cache tester (which I have replaced) had a bug which swapped out the SAMPLE global (to all ones) when that test was run. So I guess a lesson to be sure to check the actual values independently when making unit tests.

hmaarrfk: Thats really funny, i JUST made a PR for that bug

https://github.com/scikit-image/scikit-image/pull/4099

hmaarrfk: Thats really funny, i JUST made a PR for that bug

https://github.com/scikit-image/scikit-image/pull/4099

hmaarrfk: I'm really sorry that there are 2 PRs in flight that are going to caues you merge issues......

rdgraham:
> @rdgraham thanks very much for this contribution! I'm currently on holidays so I don't have time for a thorough review, but others might jump in. Either way on first glance it looks really great!
> 
> The main worry I have is whether this affects speed. Ideally, we would have benchmarks to measure these operations, make sure the caching is working properly, etc. @hmaarrfk wrote a fantastic guide on writing and using benchmarks [here](https://scikit-image.org/docs/dev/contribute.html#benchmarks). If you are so inclined, it would be awesome to have some benchmarks as part of this PR, but if it's too much work, don't worry, we'll take it from here.

Thanks for your nice reply. I will look into adding a benchmark for CPU and memory in a common usage.

rdgraham:
> Regarding the property decorator, this is being addressed in
> #3934
> 
> How the docs are added in, well that is likely a decision for an other day. I would stick to the performance gains in this PR, and avoid the property decorator changes.
> 
> The benchmarks will help us understand the use cases you are consider (image sizes, dtypes, dimentionality)

Thank you for pointing me to that discussion in #3934. It looks like we independently noticed this issue. Correct me if I am wrong, but I didn't see an independent/conflicting fix proposed in that thread, so it looks to me like this branch should suffice for a fix.

I agree that ideally a pull request should just cover just one self-contained issue. That's what I would normally do. I can separate these out into multiple pull request if that is important, though it is some work. Given that the extra changes are in essence only to documentation and style, not the API, I was hoping it would be ok.

FYI, the reason it came about like this is that I didn't want to write new code with the old style. So I decided to just update everything to be consistent after checking on the mailing list. Fixing the bug with the -O flag came along for the ride.

hmaarrfk:
I know, I would typically just try to review this PR as is, but it seems like you are trying to address a few issues:
1. Doc mangling style
2. Python optimize
3. A hack that was used in the past.
4. Speed
5. A bug in the tests

I think it is just blank luck that you are addressing issues that are already in flight.
To help you get your contribution through, I would pick one of the issues above (that aren't in flight) and focus on it.

It will help your PR stay contained, and allow us to accept it faster.

hmaarrfk:
Breaking it up also helps, but that could be more work than what you want to put in.

skimage/measure/tests/test_regionprops.py
rdgraham: Hah. So I guess I should check the other pull requests frequently. Would have saved me quite a while figuring out what the heck was going on with that bug.

Maybe it would be a good idea to use some sort of immutable container like frozendict for the test data to make sure that this kind of thing can't happen.

hmaarrfk: code inertia is real. Unless there is a real problem with the functionality code (python optimize for example), probably best not to touch it. I was just running the test suite alot, and these bugs were causing parallel tests to fail.

This code was also probably written before the days of the property decorator, and the popularity of these cool containers.

jni:
@rdgraham since we merged #3934, this needs a rebase or merge from master, as it included some of the changes from this PR (using `@property`). Are you ok with doing this or would you like some help?

rdgraham:
> @rdgraham since we merged #3934, this needs a rebase or merge from master, as it included some of the changes from this PR (using `@property`). Are you ok with doing this or would you like some help?

Yes I am OK to do that. Was planning on splitting my branch into two PRs, maybe this weekend.

mkcor:
Dear @rdgraham,

thank you for this PR so far! Would you like to continue working on it? If not, we can help finishing it.

Thanks again!

rdgraham:
Thanks for reminding me of this. It looks like this PR was something I lost track of with everything that was happening mid 2019.

It needed some organizing since as things turned out, there were some other bug fixes and changes that were rolled into it and needed to be separated.

As of now, I am not working with scikit-image very frequently and am not up to speed with the code. I expect there have probably been a lot of changes in the last 2 years. Someone else may be better positioned to get these changes in if they are still relevant. I am happy to answer any questions though.

I can report that I did end up using this commit in my own personal branch for a large analysis project that I was doing, now finished. It was quite essential to reduce memory usage in that project. I can report that the technique of caching a numpy-packed version of the region mask (and then allowing the big original labeled images to be garbage collected) worked fine in that project and I didn't run into any problems.

As was mentioned earlier in the thread, it might be a good idea to add a benchmark test case to establish for reviewers the extent of the performance hit in the case where caching is enabled. My expectation is that there might be a small hit, (it is optional after all), but in the larger context of many larger applications, it will be well worth it due to much lower memory usage.

matthew-brett:

<details>

## Rationale

</details>

Issue #4086, opened 2019-08-06
Interest in a 'register_rotation_scaling' function?


mellertd:
I am interested in possibly contributing to the project, but want to first gauge general interest before I get too far along. I think it would be great to have a cross-correlation-based registration function similar to `feature.register_translation`, but for determining scaling and rotation differences between images. This is possible by including a log-polar transformation as a step preceding the cross-correlation.

I have been able to implement this in the form of a script that mostly uses skimage, but relies on cv2 for the log-polar warping.

So my proposal would be to create two new functions in skimage:

1) a log-polar warping function, which I guess would mostly be a convenience function for the more generally useful `transform.warp` (i.e., `transform.warp_polar` with a log/linear flag). Done in #4097

2) a `feature.register_rotation_scaling` function that leverages the `warp_polar` function.

Would this be a welcome addition?

One caveat here is that I have almost no experience contributing to open source projects, and this would be very much a learning experience for me (although one that I would very enthusiastically undertake). On the plus side:

- I have a pretty good understanding of the instructions for contributing

- I basically already have a draft of the `warp_polar` thing finished in my own fork of the project

Looking forward to hearing back from you all on this. Thanks!

sciunto:
Xref #2496

jni:
@sciunto :joy: You know, I was with @mellertd recently who asked me about this, and I was thinking, I'm *sure* I've seen something like this in skimage! I even remembered your droplet from the paper. But I still couldn't find it! :joy:

@mellertd I'd be very interested in the polar transform. It'd be interesting to compare your version with #2496, and grab the best bits of both.

I'm not sure that `register_rotation_scaling` is needed since (as I understand it) it's a simple composition of `register_translation` and `transform.polar`, so it might be better as an example than a function. But I could certainly be convinced otherwise!

sciunto:
The missing stuffs in #2496 are unittests + checking that the convention is correct. If it is worth it, you can make PR on it.

mellertd:
@sciunto and @jni our approaches to the polar transform look largely similar, although I think it's worth taking a look at mine because I think it is a little more flexible and has semilog scaling built in as an option. What would be the best way to approach this? A PR to this repo (I would still need to write tests and such), or is there a good SOP for comparing my fork with @sciunto's directly?

[...]

PR #2496, opened 2017-02-10
WIP: Polar transform

sciunto:
## Description

This PR is a WIP, I open the PR to avoid duplicate work.

The purpose of this PR is to address an old issue: #140
and to provide a better implementation than #1554 as warping has been introduced in the library in the mean time.

Illustration:

index

## Checklist
[It's fine to submit PRs which are a work in progress! But before they are merged, all PRs should provide:]
- [x] Cartesian -> polar
- [x] Polar -> cartesian
- [ ] TEST Cartesian -> polar
- [ ] TEST Polar -> cartesian
- [ ] Gallery example

PR #4097, opened 2019-08-16
Log polar

mellertd:
## Description

A `warp_polar` function that can be useful for recovering scale and rotation differences between two images, as I show in the gallery example. This PR relates to issue #4086 and should be compared with @sciunto's implementation in #2496.

This is my first ever PR to an open source project! Thanks for the encouragement @jni. I am hoping this is a good learning experience

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/transform/_warps.py
```diff
@@ -898,3 +898,146 @@ def coord_map(*args):
     _clip_warp_output(image, warped, order, mode, cval, clip)
 
     return warped
+
+
+def _linear_polar_mapping(output_coords, k_ang, k_rad, center):
```
anadon: https://github.com/scikit-image/scikit-image/blob/f998c6334501d4724ded875aea04cd8bef4c6c6a/skimage/color/colorconv.py#L1508

skimage/transform/_warps.py
```diff
@@ -898,3 +898,146 @@ def coord_map(*args):
...
+        correspond to the `output_coords` given as input.
+    """
+    angle = output_coords[:, 1] / k_ang
+    rr = ((output_coords[:, 0] / k_rad) * np.sin(angle)) + center[0]
+    cc = ((output_coords[:, 0] / k_rad) * np.cos(angle)) + center[1]
+    coords = np.column_stack((cc, rr))
+    return coords
+
+
+def _log_polar_mapping(output_coords, k_ang, k_rad, center):
```
anadon: It looks like you have a more general implementation than other referenced code, but I'm also willing to bet another part of scikit does this.  The code I linked to should be refactored to use that code as a separate PR, and you should see if that code is better and use it, or if this is better and refactor appropriately.

skimage/transform/_warps.py
```diff
@@ -898,3 +898,146 @@ def coord_map(*args):
...
+        correspond to the `output_coords` given as input.
+    """
+    angle = output_coords[:, 1] / k_ang
+    rr = ((np.exp(output_coords[:, 0] / k_rad)) * np.sin(angle)) + center[0]
+    cc = ((np.exp(output_coords[:, 0] / k_rad)) * np.cos(angle)) + center[1]
+    coords = np.column_stack((cc, rr))
+    return coords
+
+
+def warp_polar(image, center=None, radius=None, output_shape=None,
```
anadon: Consider parameterizing arguments, and adding return type.

skimage/transform/_warps.py
```diff
@@ -898,3 +898,146 @@ def coord_map(*args):
...
+               scaling='linear', **kwargs):
```
anadon: kwargs -> transform_warp_func_kwargs

skimage/transform/_warps.py
```diff
@@ -898,3 +898,146 @@ def coord_map(*args):
...
+        k_rad = width / radius
+        map_func = _linear_polar_mapping
+    elif scaling == 'log':
+        k_rad = width / np.log(radius)
+        map_func = _log_polar_mapping
+
+    k_ang = height / (2 * np.pi)
+    warp_args = {'k_ang': k_ang, 'k_rad': k_rad, 'center': center}
+
+    warped = warp(image, map_func, map_args=warp_args,
```
anadon: I'd just return the result, not store it then return the stored value.

skimage/transform/tests/test_warps.py
```diff
@@ -517,3 +520,78 @@ def test_zero_image_size():
     with testing.raises(ValueError):
         warp(np.zeros((10, 10, 0)),
              SimilarityTransform())
+
+
+def test_linear_polar_mapping():
```
anadon: This looks like an excellent use case for hypothesis.

https://hypothesis.readthedocs.io/en/latest/

skimage/transform/_warps.py
mellertd: good catch! Here it's a private function though and I am hesitant to change that unless it's too much redundancy

doc/examples/transform/plot_register_rotation.py
```diff
@@ -0,0 +1,106 @@
...
+===================================
+
+Rotation differences between two images can be converted to translation
+differences along the angular coordinate (:math:`\\theta`) axis of the
+polar-transformed images. Scaling differences can be converted to translation
+differences along the radial coordinate (:math:`\\rho`) axis if it
+is first log transformed (i.e., :math:`\\rho = \ln\sqrt{x^2 + y^2}`). Thus,
+in this example, we use phase correlation (``feature.register_translation``)
+to recover rotation and scaling differences between two images that share a
+center point.
```
sciunto: Nice example.

I would like to see this image as a fist plot: https://cloud.githubusercontent.com/assets/335370/22819610/26b2cce6-ef73-11e6-989c-997a8d44fb2b.png

This is rather simple to introduce the concept and will be a good thumbnail.
If you don't want to add it, I'll do later on.

skimage/transform/_warps.py
sciunto: Let's keep this for a future issue.

skimage/transform/_warps.py
```diff
@@ -898,3 +898,146 @@ def coord_map(*args):
...
+        the transformation is applied to the first two axes.
+    center : tuple (row, col), optional
+        Point in image that represents the center of the transformation (i.e.,
+        the origin in cartesian space). Values can be of type `float`.
+        If no value is given, the center is assumed to be the center point
+        of the image.
+    radius : float, optional
+        Radius of the circle that bounds the area to be transformed.
+    output_shape : tuple (row, col), optional
+    scaling : {'linear', 'log'}
```
sciunto:
```suggestion
    scaling : {'linear', 'log'}, optional
```

doc/examples/transform/plot_register_rotation.py
sciunto: `data.retina()` is already in skimage.data.

mellertd: not entirely sure what you mean by "fist plot", but I agree that the retina image is more compelling for an example and have used it instead of the astronaut image. This also obviates the need for highlighting the area in the image to be transformed, because the image is naturally circular

sciunto: What I meant is just to reproduce the figure I linked in the previous message and display this figure in the first position. The two other figures that you made can come after.
The first figure is used as a thumbnail on the gallery. 

My point was to get something simple first, before showing some usage on the estimation of geometrical transforms.

But, this is just a suggestion. I'm happy with this PR anyway.

mellertd: Ahhhh, okay I understand now. Yes I agree that could be useful. The information is more or less there in the two plots I have, but those aren't really optimized for demonstrating the basic concept of the transform. I will add later, or you can feel free to drop it in.

skimage/transform/_warps.py
```diff
@@ -898,3 +898,149 @@ def coord_map(*args):
...
+    return coords
+
+
+def warp_polar(image, center=None, radius=None, output_shape=None,
+               scaling='linear', **kwargs):
+    """Remaps image to polor or log-polar coordinates space.
+
+    Parameters
+    ----------
+    image : ndarray (2-D or 3-D)
```
hmaarrfk: Is this API compatible with the desire to have things thinking about ND images when they get merged in?

It seems like we often like the "multichannel" parameter to be passed in when the last index is image channel.

Otherwise, we apply the transformation to the last two dimensions.

Not sure what is best. @jni?

skimage/transform/_warps.py
```diff
@@ -898,3 +898,149 @@ def coord_map(*args):
...
+
+
+def _log_polar_mapping(output_coords, k_ang, k_rad, center):
+    """Inverse mapping function to convert from cartesion to polar coordinates
+
+    Parameters
+    ----------
+    output_coords : ndarray
+        `(M, 2)` array of `(col, row)` coordinates in the output image
+    k_ang : float
```
jni: Let's expand these a little: `k_angle`, `k_radius`. (since `rad` is a common abbreviation for `radians`)

skimage/transform/_warps.py
```diff
@@ -898,3 +898,149 @@ def coord_map(*args):
...
+        correspond to the `output_coords` given as input.
+    """
+    angle = output_coords[:, 1] / k_ang
+    rr = ((np.exp(output_coords[:, 0] / k_rad)) * np.sin(angle)) + center[0]
+    cc = ((np.exp(output_coords[:, 0] / k_rad)) * np.cos(angle)) + center[1]
+    coords = np.column_stack((cc, rr))
+    return coords
+
+
+def warp_polar(image, center=None, radius=None, output_shape=None,
```
jni:
```suggestion
def warp_polar(image, center=None, *, radius=None, output_shape=None,
```

This makes any arguments after `center` keyword-only.

skimage/transform/_warps.py
jni: Oh yeah, 100%, we don't do this kind of guessing anymore. For multichannel images, we should require `multichannel=True` to be given.

skimage/transform/_warps.py
```diff
@@ -898,3 +898,149 @@ def coord_map(*args):
...
+               scaling='linear', **kwargs):
+    """Remaps image to polor or log-polar coordinates space.
```
jni:
```suggestion
    """Remap image to polar or log-polar coordinate space.
```

skimage/transform/tests/test_warps.py
jni: Unfortunately we are not ready to add hypothesis as a test dependency yet, although this is something we could consider in the future. I don't think we should saddle this PR with a new dependency, which will only slow down its time to merge.

skimage/transform/_warps.py
hmaarrfk: To be a little more explicit, jni means that by default, multichannel should be set to False. In your code, by default, images that include RGB channels should be rejected.

They should be accepted only when the multichannel argument is true

mellertd:
One last question on this, about attribution and referencing. I roughly followed the convention of the openCV warpPolar python API (but making it more skimage-y), although I didn't actually try to port the actual code. And the math is just math. Should a reference to openCV be made in the docstring or anything?

skimage/transform/tests/test_warps.py
```diff
@@ -517,3 +520,78 @@ def test_zero_image_size():
...
+    for rad in radii:
+        rr, cc, val = circle_perimeter_aa(25, 25, rad)
+        image[rr, cc] = val
+    warped = warp_polar(image, radius=25)
+    profile = warped.mean(axis=0)
+    peaks = peak_local_max(profile)
+    assert np.alltrue([peak in radii for peak in peaks])
+
+
+def test_log_warp_polar():
```
jni: We need at least one more test, `test_invalid_scaling`, using `with testing.raises(ValueError):`.

jni:
(I don't think anything else is needed in terms of references.)

hmaarrfk:
Yeah, i think the missign coverage is half my fault. I tried to make it stop writing long annoying messages, but it seems to have disabled other stuff too.

sciunto:
All good. Let's merge this one. Many thanks @mellertd

Issue #4100, opened 2019-08-17
Drop Python 3.5 support


hmaarrfk:
According to https://github.com/numpy/numpy/pull/14086
We should basically be dropping Python 3.5 support for scikit-image 0.16

https://github.com/numpy/numpy/pull/14086/files#diff-b37c3c8057e08a90d877a2ba31d8a215R107

JDWarner:
NumPy has thought this through pretty well, I am :+1: on sticking with their cadence here.  Especially as our next release is going to 3.x only, it will simplify future support burden.

sciunto:
This opens the use of f-strings. :tada:

jni:
Woohoo! I'm so happy. =)

hmaarrfk:
Of course this is now more difficult than just removing entries in the build matrix. We should decide on how we want to address the rest of that NEP:

If we aim for a september release date, then we could follow their: "all minor versions of NumPy released in the prior 24 months". We can use this to suggest new minimum versions to set.

1. Numpy: 1.14 as recommended (Released Jan 2018)
2. Scipy: 1.0.0 release Sept 2017
3. Matplotlib: 2.1. 2.1.0 Released in Oct 2017
4. Networkx: 2.0. 2.0 Released in Oct 2017
5. Pillow: 4.3.0 Released in Oct 2017
6. imageio: 2.3. Released in Oct 2018, 2.2 releaesd in March 2017
7. PyWavelets: 1.0. Released in Oct 2018. 0.5 was releaesd in March 2017.

I think all of these should basically be bumped to help modernize contribution too :D.

soupault:
:+1: to @hmaarrfk. I think those versions are fine.

hmaarrfk:
I guess we should think of version bumbs as they are needed. But we should keep in mind that there is no reason to support increasingly aging versions.

PR #4102, opened 2019-08-18
Drop Python 3.5 support

hmaarrfk:
## Description

xref: https://github.com/numpy/numpy/pull/14086
closes https://github.com/scikit-image/scikit-image/issues/4100

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -13,7 +13,7 @@ https://scikit-image.org
 
 
 This release of scikit-image drops support for Python 3.5.
-This release of scikit-image officially supports for Python 3.6 and 3.7.
+This release of scikit-image officially supports Python 3.6 and 3.7.
 
 New Features
 ------------
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -12,6 +12,8 @@ For more information, examples, and documentation, please visit our website:
 https://scikit-image.org
 
 
+This release of scikit-image drops support for Python 3.5.
+This release of scikit-image officially supports for Python 3.6 and 3.7.
 
 New Features
 ------------
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -11,8 +11,10 @@ For more information, examples, and documentation, please visit our website:
 
 https://scikit-image.org
 
-
-This release of scikit-image drops support for Python 3.5.
+Starting from this release, scikit-image will follow the spirit of the recently
+introduced numpy deprecation policy -- NEP 29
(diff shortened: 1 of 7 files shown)


hmaarrfk:
3.8 is too hard to add.

hmaarrfk:
Let me know how you want it to be changed.

sciunto:
Indeed, it's more than a guarantee, 0.16 won't be compatible as we will use f-string in many places very soon.

sciunto:
@hmaarrfk Do you want to propose a solution to the conflict?

soupault:
Are we going to make the releases for Py3.5 still (https://github.com/scikit-image/scikit-image/blob/master/setup.py#L218)? 🤔

hmaarrfk:
No, no releases for 3.5. Thanks for the catch. Hopefully I updated the setup.py correctly.

setup.py
```diff
@@ -36,11 +36,11 @@
 from numpy.distutils.command.build_ext import build_ext
 
 
-if sys.version_info < (3, 5):
+if sys.version_info < (3, 6):
 
     error = """Python {py} detected.
 
-scikit-image 0.15+ support only Python 3.5 and above.
+scikit-image 0.16+ support only Python 3.6 and above.
```
jni:
```suggestion
scikit-image 0.16+ supports only Python 3.6 and above.
```

doc/release/release_dev.rst
```diff
@@ -12,6 +12,8 @@ For more information, examples, and documentation, please visit our website:
 https://scikit-image.org
 
 
+This release of scikit-image drops support for Python 3.5.
+This release of scikit-image officially supports Python 3.6 and 3.7.
```
jni: Perhaps we should state the policy going forward?

hmaarrfk: I thought release notes were about facts, not decision process. Maybe it can go somewhere else?

jni: When the API or supported versions change, affected users typically want to know why. It's the reason why we "link" to the PRs, I think. You could add a "see #4102 for more info", but you can explain the NEP in about the same space.

soupault: How about something like this? :
"Starting from this release, scikit-image will follow the spirit of the recently introduced numpy deprecation policy -- NEP 29 (https://github.com/numpy/numpy/blob/master/doc/neps/nep-0029-deprecation_policy.rst). Accordingly, scikit-image 0.16 drops the support for Python 3.5 [and numpy __]. (#4102)"

soupault:
Thanks, Mark!

PR #4110, opened 2019-08-24
Gallery: use fstrings for better readability

sciunto:


doc/examples/transform/plot_masked_register_translation.py
```diff
@@ -33,23 +33,23 @@
 image = data.camera()
 shift = (-22, 13)
 
-corrupted_pixels = np.random.choice([False, True], 
-                                    size = image.shape, 
+corrupted_pixels = np.random.choice([False, True],
+                                    size = image.shape,
```
soupault: This looks bad 😄. Also, the same issue below. I'll fix that once I review that PR history.

hmaarrfk:
Whats the review process here? Are we expected to look at the resulting doc build? These kind of changes are cosmetic, and often end up trading one set of bugs for an other.

I would be a little less eager to add f-strings to existing code, but since you've done it already, this is fine.
The existing code was fine, and in many cases, only had one or two variables printed out per line, so the curly brace formatting was fine OK.

sciunto:
I did it for the gallery as this is the code people tend to replicate. From my experience, fstring is much easier to explain to students learning python than any other option.

I agree to do not generalize the idea to the existing code of the library.

hmaarrfk:
> fstring is much easier to explain to students learning python

I've found it pretty hard to explain to people transitioning from py 2

JDWarner:
Agree these are harder to explain at a whiteboard, but really easy to get people to catch on to by showing examples and experimenting (actually, this PR's diff would be a decent teaching tool).  The readability is a lot better.

PR #4124, opened 2019-08-29
Gallery hough line transform

sciunto:
## Description

I made some edits to improve the examples:

* sections to split hough line and prob hough line
* change the default value of the hlt, often people have problems with the accuracy and ignore after reading the example that it's possible to change it.
* increase a bit the image resolution. No  significant impact on the processing time.
* Simplify the notations for the origin in the plot. The example is less verbose.
* PEP8

New version below

Straight line Hough transform — skimage v0 16 dev0 docs


soupault:
@sciunto I am thinking about moving all the imports into a separate section, similar to https://scikit-learn.org/stable/auto_examples/cluster/plot_linkage_comparison.html. That way it should be easier for novices to run the second example independently. What is your opinion on this?

sciunto:
>  What is your opinion on this?

Good suggestion. In this particular case, I wanted to make a clear separation between deterministic and probabilistic approach and show that we are using different functions.

Imports on two different lines in a dedicated section could work as well. :+1:

alexdesiqueira:
Very nice, @sciunto. Thanks!

PR #4126, opened 2019-08-29
morphology.rectangle() dimensions swapped, Fixes Issue #4125

### LINKED ISSUE #4125: BUG: morphology.rectangle() dimensions swapped
## Description
The function `rectangle(width, height, dtype=np.uint8)`, from morphology, generates a rectangle structing element with these dimensions swapped. 

## Way to reproduce
```python
skimage.morphology.rectangle(3,1)
```

## Expected output:
```python
[[1 1 1]]
```

## Actual output:
```python
[[1]
 [1]
 [1]]
```

## Version information
```python
# your output here
3.6.8 (default, Jan 14 2019, 11:02:34) 
[GCC 8.0.1 20180414 (experimental) [trunk revision 259383]]
Linux-4.15.0-58-generic-x86_64-with-Ubuntu-18.04-bionic
scikit-image version: 0.15.0
numpy version: 1.16.2
```

aleblucher:
## Description
Issue required that the parameters to build the rectangle were inverted.

Fixes #4125

## Checklist

- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/selem.py
+++ b/skimage/morphology/selem.py
@@ -54,7 +54,7 @@ def rectangle(width, height, dtype=np.uint8):
         pixel belongs to the neighborhood.
 
     """
-    return np.ones((width, height), dtype=dtype)
+    return np.ones((height, width), dtype=dtype)
 
 
 def diamond(radius, dtype=np.uint8):


aleblucher:
[https://github.com/scikit-image/scikit-image/issues/4125](url)

hmaarrfk:
Thanks for the fix. The test should really be fixed too..... once we figure out the correct way forward

The challenge is that this bug goes back 10 years, and that this swap of coordinates will break certain people's code.

image

I'm really not sure how to go about this....

aleblucher:
I didn't realize the problem was so deep rooted. How do you think I should proceed? I could alter some documentation if need be and make some tests. Or if I could at least map which parts of other could would break.

jni:
I think the best way forward is to change the function signature to:

```python
def rectangle(nrows, ncols, dtype=np.uint8, *, height=None, width=None):
    """
    ...
    """
    if height is not None:  # warn that height is deprecated, set ncols to height
    # same for width
    return np.ones((nrows, ncols), dtype=dtype)
```

hmaarrfk:
That transposes things for people that use position arguments

jni:
@hmaarrfk no it doesn't, the point is that we were using `width` as `nrows` already, unless I'm misreading this whole thing.

hmaarrfk:
Oh. I see, then brows and ncols need default vues so one can omit them

jni:
@hmaarrfk you're exactly right there, yes. =)

jni:
@aleblucher do you think you can handle the above comments, or would you like us to take over?

Thanks for your contribution so far!

aleblucher:
@jni I think I can handle it. Thank you for your help!

hmaarrfk:
@aleblucher did you want to follow through with @jni's recommendation?

You would have to detail it out with:
https://scikit-image.org/docs/stable/contribute.html#deprecation-cycle

FirefoxMetzger:
@hmaarrfk @jni A (probably stupid) question: Why can't we just swap the names of the variables? That would leave the functionality as is, but the signature `rectangle(height, width)` would now correspond with what is actually happening and be less confusing.

I would like this much more than `rectangle(nrow, ncol)`, because one can do `rectangle(*some_gray_img.shape)` and it will just work.

rfezzani:
It is fixed now by #4926. Thank you @aleblucher.

Issue #3799, opened 2019-03-09
radon doesn't interpret uint8 dtypes correctly


hmaarrfk:
## Description

it should at least use skimage units.

needs a deprecation cycle
```python
>>> from skimage.transform import radon                                     
>>> import numpy as np                                                      
>>> a = np.ones((2, 2), dtype='uint8')
>>> from numpy.testing import assert_allclose                                                                                                                                                           
>>> assert_allclose(radon(a), radon(a.astype('float64')))
>>> from skimage import img_as_float
>>> assert_allclose(radon(a), radon(img_as_float(a)))                                                                                                                                                   
/home/mark/miniconda3/envs/owl/lib/python3.7/site-packages/skimage/transform/radon_transform.py:58: UserWarning: The default of `circle` in `skimage.transform.radon` will change to `True` in version 0.15.
  warn('The default of `circle` in `skimage.transform.radon` '
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-9-f353f5050d50> in <module>
----> 1 assert_allclose(radon(a), radon(img_as_float(a)))

~/miniconda3/envs/owl/lib/python3.7/site-packages/numpy/testing/_private/utils.py in assert_allclose(actual, desired, rtol, atol, equal_nan, err_msg, verbose)
   1457     header = 'Not equal to tolerance rtol=%g, atol=%g' % (rtol, atol)
   1458     assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
-> 1459                          verbose=verbose, header=header, equal_nan=equal_nan)
   1460 
   1461 

~/miniconda3/envs/owl/lib/python3.7/site-packages/numpy/testing/_private/utils.py in assert_array_compare(comparison, x, y, err_msg, verbose, header, precision, equal_nan, equal_inf)
    794                                 verbose=verbose, header=header,
    795                                 names=('x', 'y'), precision=precision)
--> 796             raise AssertionError(msg)
    797     except ValueError:
    798         import traceback

AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

(mismatch 99.62962962962963%)
 x: array([[2.000000e+00, 1.965247e+00, 1.930810e+00, 1.896699e+00,
        1.862923e+00, 1.829494e+00, 1.796421e+00, 1.763715e+00,
        1.731386e+00, 1.699443e+00, 1.667896e+00, 1.636755e+00,...
[...]

PR #4128, opened 2019-08-29
Allow warping float32 images without upcasting to double

### LINKED ISSUE #4108: resize output data type is not consistent with input data type
## Description

The `resize` function actually always returns `float64` arrays whatever is the data type of the input array to be resized.

## Way to reproduce

```python
from skimage import data, img_as_float32, img_as_float
img_u8 = data.camera()
img_f32 = img_as_float32(img_u8)
img_f64 = img_as_float(img_u8)
print(resize(img_u8, (256, 256)).dtype,
      resize(img_f32, (256, 256)).dtype,
      resize(img_f64, (256, 256)).dtype)

# outputs: float64 float64 float64
```

## Version information

```
3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0]
Linux-4.15.0-58-generic-x86_64-with-debian-buster-sid
scikit-image version: 0.15.0
numpy version: 1.16.4
```

rfezzani:
## Description

This PR fixes #4108 and #4190.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/transform/_warps_cy.pyx
```diff
@@ -119,8 +120,13 @@ def _warp_fast(cnp.ndarray image, cnp.ndarray H, output_shape=None,
 
     """
 
-    cdef double[:, ::1] img = np.ascontiguousarray(image, dtype=np.double)
-    cdef double[:, ::1] M = np.ascontiguousarray(H)
+    cdef floating[:, ::1] img = np.ascontiguousarray(image)
+    cdef floating[:, ::1] M = np.ascontiguousarray(H)
+
+    if floating is float:
```
hmaarrfk: Is float ambiguous here? I'm concerned there is a difference between python floats and c floats.

rfezzani: In Cython context, I think that `float` belongs to `cython.float`. So I suppose that it is not ambiguous...

rfezzani: I took it from [cython doc](http://docs.cython.org/en/latest/src/userguide/fusedtypes.html)  in **Type Checking Specializations** section ;-).

hmaarrfk: The fact that you used the word "think" and "not ambiguous" in the same sentence to me makes it ambiguous.

The challenge comes when having code like this:

image

`float` clearly means different things when used in python+numpy, or in cython.

I would just replace `float` with `np.float32_t`, making it explicit that you mean float32.

skimage/_shared/utils.py
```diff
@@ -226,7 +226,7 @@ def check_random_state(seed):
 
 
 def convert_to_float(image, preserve_range):
-    """Convert input image to double image with the appropriate range.
+    """Convert input image to float image with the appropriate range.
```
hmaarrfk: This needs to say something like:

Floating point types will maintain their precision, while integer types will be converted to `double`.

skimage/_shared/utils.py
```diff
@@ -243,7 +243,8 @@ def convert_to_float(image, preserve_range):
         Transformed version of the input.
     """
     if preserve_range:
-        image = image.astype(np.double)
+        if image.dtype.char not in 'df':
```
hmaarrfk:
```suggestion
        if not np.issubdtype(image.dtype, np.floating):
```

```python
>>> import numpy as np
>>> np.issubdtype(np.float32, np.floating)
True
>>> np.issubdtype(np.float64, np.floating)
True
>>> np.issubdtype(np.float16, np.floating)
True
>>> np.issubdtype(np.uint8, np.floating)
False
```

skimage/transform/tests/test_warps.py
```diff
@@ -331,6 +331,22 @@ def test_resize3d_bilinear():
     assert_almost_equal(resized, ref)
 
 
+def test_resize_dtype():
+    x = np.zeros((5, 5))
+    x_f32 = x.astype(np.float32)
+    x_u8 = x.astype(np.uint8)
+    x_b = x.astype(bool)
```
hmaarrfk: I'm not sure we care about supporting boolean data types. @jni?

skimage/transform/_warps.py
```diff
@@ -841,17 +841,19 @@ def warp(image, inverse_map, map_args={}, output_shape=None, order=1,
...
 
         if matrix is not None:
-            matrix = matrix.astype(np.double)
+            matrix = matrix.astype(image.dtype)
+            ctype = 'float' if image.dtype.char == 'f' else 'double'
             if image.ndim == 2:
-                warped = _warp_fast(image, matrix,
-                                    output_shape=output_shape,
-                                    order=order, mode=mode, cval=cval)
+                warped = _warp_fast[ctype](image, matrix,
```
hmaarrfk: this won't select things automatially? Maybe that was my mistake....

skimage/transform/_warps.py
```diff
@@ -841,17 +841,19 @@ def warp(image, inverse_map, map_args={}, output_shape=None, order=1,
             matrix = np.linalg.inv(inverse_map.__self__.params)
 
         if matrix is not None:
-            matrix = matrix.astype(np.double)
+            matrix = matrix.astype(image.dtype)
+            ctype = 'float' if image.dtype.char == 'f' else 'double'
```
hmaarrfk: I would really like it if you found a way to call these `float32` and `float64`. Mostly `float` is ambiguous, `double` not so much.

skimage/transform/_warps_cy.pyx
rfezzani: OK, I was thinking that you where talking about _computer ambiguity_, not _understanding ambiguity_. Sorry for my poor English speaking/understanding.
On python side, the ambiguity is clearly there, but in cython context I don't think so, the whole test block is explicit:
```python
    if floating is float:
        dtype = np.float32
    else:
        dtype = np.double
```
We can make it even more explicit using:
```python
    if floating is float:
        dtype = np.float32
    elif floating is double:
        dtype = np.double
    else:
        raise TypeError("Only float32 and float64 arrays are supported.")
```

skimage/_shared/utils.py
rfezzani: This also true for al the functions defined in `_warp_cy.pyx` and for `radon`. It is in my TODO list ;-)

skimage/_shared/utils.py
rfezzani: That's definitely more readable ;-). I excluded `float16` and `float128` because I though that these types are not supported by `skimage`.

skimage/transform/_warps.py
rfezzani: Nop, it wasn't working without this specialization...

skimage/transform/_warps.py
rfezzani: I was happy to make it work without the explicit import here to cython: I did not knew that passing strings for specialization works :-D
Well, I will give it a try.

skimage/_shared/utils.py
rfezzani: Sorry, I am not awake yet :-)
I will also fix the docstrings for `radon` and the function in `_warp_cy`.

skimage/_shared/utils.py
rfezzani: This will cause a bug since `float16` is not supported by `cython.floating` (only `float` and `double`) :-(

```python
>>> a = np.ones((5, 5), dtype=np.float16)
>>> resize(a, (10, 10))
  File "skimage/transform/_warps_cy.pyx", line 70, in skimage.transform._warps_cy._warp_fast
    def _warp_fast(floating[:, :] image, floating[:, :] H, output_shape=None,
ValueError: Does not understand character buffer dtype format string ('e')
```
* We can use
  ```python
  if image.dtype not in [float, np.float, np.double, np.float32, np.float64]:
  ```
  but I am not sure that it is more readable...
* The other option is to keep the original test (`if image.dtype.char not in 'df'`) and to add a comment in the code.

What do you think?

skimage/transform/_warps.py
rfezzani: Infortunatly, this doesn't work :-(  This is because `_warp_cy.__signatures__.keys()` only contains `"float"` and `"double"`.

I added a comment trying to alleviate the ambiguity:
```python
# Manage _warp_fast specialization according to input data
# type floating point precision (single or double)
ctype = 'float' if image.dtype.char == 'f' else 'double'
```

skimage/transform/_warps_cy.pyx
rfezzani: The last part of my answer is wrong, since floating can only be `float` or `double`. Sorry :frowning_face:

rfezzani:
Thank you @hmaarrfk for your review, I tried in my last push to address all your concerns.

skimage/_shared/utils.py
rfezzani: According to my [comment](https://github.com/scikit-image/scikit-image/pull/4128#discussion_r319407879), since not all the floating point precision are supported, I added a **Notes** section in the docstring specifying this: _"Input images with `float32` data type are not upcasted."_

skimage/_shared/utils.py
hmaarrfk: Just ignore float16, the whole library is broken with it.

rfezzani: All right, I will have to ignore `np.float128` also:
```python
>>> a = ones((5, 5), dtype=np.float128)
>>> resize(a, (10, 10))
ValueError: Buffer dtype mismatch, expected 'double' but got 'long double'
```
This is because `cython.floating` only supports single and double precision floating points.
Do you prefer `if not (np.issubdtype(image.dtype, np.floating) or image.dtype in [np.float16, np.float128])` instead of `if image.dtype.char not in 'df'`?

hmaarrfk: I once wrote a fused numerics file. You can add things there.

That said, neither of those two datatypes are efficient in hardware, nor are they well tested. 

If you want to, you can add to the fused numerics file maybe. Not exactly sure what will happen to the rest of the library.

hmaarrfk: Honestly, you could just add a comment. That said, you could also add a Todo, and raise an issue that float16 and float128 have yet undefined behaviours in the img_as routines.

rfezzani: I already tried some explorations with `skimage._shared.fused_numerics.np_floats`, but `numpy` doesn't declare `float16_t` and `float128_t` for `cimport` :-(

hmaarrfk: yeah, float16 and float128 are pretty badly supported in general. Adding their definitions to the cimport could be added as an issue upstream.

skimage/transform/_warps.py
hmaarrfk: It needs a sentence like:

> Contrary to Python, Cython `float` follows the meaning of `float` in C, and is equivalent to `float32`.

hmaarrfk:
I think what would make me happy is that if you included this sentence

> Contrary to Python and numpy, Cython `float` follows the meaning of `float` in C, and is equivalent to `float32`.

Whenver you used `float` in cython.  Otherwise, maybe use our fused numerics, which would translate to `np.float32_t` which is much less ambiguous, maybe...

rfezzani:
Using `skimage.__shared.fused_numercs.np_floats` instead of `cython.floating` removes the ambiguity since 
```python
ctype = 'float' if image.dtype.char == 'f' else 'double'
```
becomes
```python
ctype = 'float32_t' if image.dtype is np.float32 else 'float64_t'
```
and
```python
   if floating is float:
        dtype = np.float32
    else:
        dtype = np.double
```
becomes

```python
    if np_floats is cnp.float32_t:
        dtype = np.float32
    else:
        dtype = np.float64
```
I think that there is no need then for more comments. Thx @hmaarrfk ;-)

skimage/transform/radon_transform.py
```diff
@@ -62,6 +63,8 @@ def radon(image, theta=None, circle=True):
     if theta is None:
         theta = np.arange(180)
 
+    image = convert_to_float(image, preserve_range=True)
```
hmaarrfk: This needs a new parameter, and documentation. I think our default is to have `preserver_range=False`.

You can consider leaving this for a different PR and creating an issue about it.

skimage/transform/radon_transform.py
```diff
@@ -107,13 +110,14 @@ def radon(image, theta=None, circle=True):
 
     def build_rotation(theta):
         T = np.deg2rad(theta)
-        R = np.array([[np.cos(T), np.sin(T), 0],
-                      [-np.sin(T), np.cos(T), 0],
+        c, s = np.cos(T), np.sin(T)
```
hmaarrfk: These marginal gains should be left to the PR that includes the test for preserve range.

hmaarrfk:
The cython stuff seems fine for me. I think the radon fix should be more fleshed out, and ideally left to a different PR if it can be.

skimage/transform/radon_transform.py
rfezzani: Should I create an Issue about something that is not yes in `skimage` code?

rfezzani: I though about leaving `preserve_range` to its default value, but this would potentially break users code, but well, I can open an issue about the fact that the arrays are implicitly casted to `double` by `_warp_cy` in `radon` ;-)

hmaarrfk:
You know, come to think of it, I think I had opened an issue about this in
the past. Will have to find it. Please leave this out for now. if you can.

rfezzani:
Isn't it #3799?

hmaarrfk:
Thats the one!

Although that was a bug, I think that behaviour definitely needs to be deprecated too.

If you try to add the fix for that in here, then the review time will be multiplicative, instead of additive, and will delay this getting merged.

rfezzani:
Fixing the `radon` function is a one line fix + doc update: just add a call to `convert_to_float` :-)
I am about to complete filling the issue.

rfezzani:
> Thats the one!
> 
> Although that was a bug, I think that behaviour definitely needs to be deprecated too.
> 
> If you try to add the fix for that in here, then the review time will be multiplicative, instead of additive, and will delay this getting merged.

So what do you recommend? Should we fix #4130 directly in this PR and close both related issues?

hmaarrfk:
Just fix the issue of rescale.

rfezzani:
OK, but this will cause `test_iradon_bias_circular_phantom` to fail in this PR :-(

hmaarrfk:
So you need the preserve range to be set to true? If so, that is fine, add a note #3799

rfezzani:
> So you need the preserve range to be set to true? If so, that is fine, add a note #3799

Not necessarly, only the failing test has to be fixed ;-)

skimage/_shared/utils.py
```diff
@@ -237,13 +237,21 @@ def convert_to_float(image, preserve_range):
         using img_as_float. Also see
         https://scikit-image.org/docs/dev/user_guide/data_types.html
 
+    Notes:
+    ------
+    * Input images with `float32` data type are not upcasted.
```
jni:
```suggestion
    * Input images with `float32` data type are not upcast.
```

See https://www.usingenglish.com/reference/irregular-verbs/cast.html

=)

skimage/_shared/utils.py
```diff
@@ -237,13 +237,21 @@ def convert_to_float(image, preserve_range):
...
     image : ndarray
         Transformed version of the input.
+
     """
     if preserve_range:
-        image = image.astype(np.double)
+        # Convert image to double only if it is not single or double
+        # precision float
+        if image.dtype.char not in 'df':
+            image = image.astype(np.double)
```
jni:
```suggestion
            image = image.astype(float)
```

(This should be the default floating point type for the platform, ie whatever np.array([1.0]).dtype returns on that platform. I think using `float` ensures that.)

skimage/transform/_warps.py
jni: 🎉 very nice, thanks @rfezzani for this discovery!

skimage/transform/tests/test_warps.py
jni: Eh. If it works, why not keep it working?

rfezzani:
Thank you for the review @jni ;-)

jni:
@stefanv any chance you can review this? I think it's a pretty major improvement, which would be nice to get into 0.16.

stefanv:
This looks good to me, merging.  If there are any other minor updates, @hmaarrfk, feel free to add a refinement PR.

PR #4131, opened 2019-08-30
Deprecate undefined data type management in radon function

### LINKED ISSUE #4130: radon function accepts any data type and implicitly converts arrays to float
## Description

It is stipulated in the `radon` function doc that only `float` input arrays are supported but  `radon` actually accepts any data type.

There is no explicit conversion of the array type in `radon` function code. This conversion is made implicitely in the cython `_warp_cy` function.

This is may be related to #3799 and impacts #4128 

## Way to reproduce

```python
import numpy as np
from skimage.transform import radon

pixels = 128
xy = np.arange(-pixels / 2, pixels / 2) + 0.5
x, y = np.meshgrid(xy, xy)
image = x**2 + y**2 <= (pixels/4)**2  # array.dtype is bool
radon(image, theta=theta)
radon(image.astype(np.uint8), theta=theta)
radon(image.astype(np.float32), theta=theta)
radon(image.astype(np.float64), theta=theta)
```

## Version information
```python
3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21) 
[GCC 7.3.0]
Linux-4.15.0-58-generic-x86_64-with-debian-buster-sid
scikit-image version: 0.15.0
numpy version: 1.16.4
```

rfezzani:
## Description

Fixes #4130 and #3799.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/transform/radon_transform.py
```diff
@@ -62,6 +67,8 @@ def radon(image, theta=None, circle=True):
     if theta is None:
         theta = np.arange(180)
 
+    image = convert_to_float(image, preserve_range)
```
hmaarrfk: I think I would feel better with a deprecation cycle. This kind of unit change is REALLY annoying pin down, and a warning can go a long way.

rfezzani: This deprecation warning should only be raised when a non float array is provided, right? But this is strange since the doc says that only float array is supported.

hmaarrfk: The challenges encountered here are why this deserves its own PR. I would identify the cases where the behavior is changed, and warn only on those.

Finally, update the documentation specifying the support for uint8 if you want. If you don't, keep it as is.

rfezzani: Yes, I agree, the addition of a new parameter can't be harmless. I will write a test to assert the multiple data types support.
The use of `convert_to_float` allows to use any data type, so I think that it is enough to simply remove the restriction to float in the doc.

rfezzani:
When `float32` support will be added to `_warp_fast`, we will have to ensure that `build_rotation` function defined in `radon` return a matrix with the same dtype as `padded_image`, otherwise an error will be raised.
Should we create a specific issue to Add `float32` support to `_warp_fast` and make the changes there? Any function calling `_warp_fast` is potentially impacted.

skimage/transform/radon_transform.py
```diff
@@ -18,23 +19,27 @@
 __all__ = ['radon', 'order_angles_golden_ratio', 'iradon', 'iradon_sart']
 
 
-def radon(image, theta=None, circle=True):
+def radon(image, theta=None, circle=True, preserve_range=False):
```
hmaarrfk: To keep the old behavior going, this should be set to `None` with a warning for users that it will change in a future version.

@jni, this function's documentation only included support for floating point types, but actually accepted integer types. Should we follow the general deprecation procedure?

rfezzani: Actually, setting `preserve_range` to `True` keeps the actual behavior unchanged since with the actual version of `convert_to_float` directly convert any array to `double` when `preserve_range` is `True`.
When `preserve_range` is False, `img_as_float` is called, and in this case, non float arrays are rescaled betweeen 0 and 1. The backward compatibility may be broken here.
Should we raise a deprecation warning when `preserve_range` is `False` and `dtype` is not floating?

hmaarrfk: Right, so I was suggesting we go through a deprecation cycle.

1. Set `preserve_range=None` by default.
2. If not set, and we have integer types, raise a warning, and set it to True
3. In 0.18, make it `False` by default.

skimage/transform/radon_transform.py
```diff
@@ -67,6 +67,11 @@ def radon(image, theta=None, circle=True, preserve_range=False):
     if theta is None:
         theta = np.arange(180)
 
+    if preserve_range is None and np.issubdtype(image.dtype, np.integer):
+        warn('Image dtype is not float. To keep old behavior, '
```
hmaarrfk: This needs to be fleshed out a little more.

> Imge dtype is not float. By default radon will assume you want to preserve the range of your image (preserve_range=True). In scikit-image 0.18 this behavior will change to preserve_range=False. To avoid this warning, explictiely specify the preserve_range parameter.

skimage/transform/radon_transform.py
```diff
@@ -18,23 +19,27 @@
 __all__ = ['radon', 'order_angles_golden_ratio', 'iradon', 'iradon_sart']
 
 
-def radon(image, theta=None, circle=True):
+def radon(image, theta=None, circle=True, preserve_range=None):
```
hmaarrfk:
```suggestion
def radon(image, theta=None, circle=True, *, preserve_range=None):
```

These kind of "behavior" parameters should definitely be made keyword only (something that wasn't possible in python 2.7)

skimage/transform/radon_transform.py
rfezzani: Done

skimage/transform/radon_transform.py
rfezzani: Done

rfezzani:
> Should we create a specific issue to Add `float32` support to `_warp_fast` and make the changes there? Any function calling `_warp_fast` is potentially impacted.

`_warp_fast` is only called by `resize` and `radon`. What about making a two step resolution of #4128 :

1. Adding `float32` support to `_warp_fast`,
2. Fix `convert_to_float` to not upcast `float32` arrays.

This way #4128 issue and its equivalent for `radon` will be solved. What do you think @hmaarrfk ?

skimage/transform/radon_transform.py
```diff
@@ -62,6 +67,17 @@ def radon(image, theta=None, circle=True):
     if theta is None:
         theta = np.arange(180)
 
+    if preserve_range is None and np.issubdtype(image.dtype, np.integer):
+        warn('Image dtype is not float. By default radon will assume'
+             'you want to preserve the range of your image'
+             '(preserve_range=True). In scikit-image 0.18 this behavior will'
+             'change to preserve_range=False. To avoid this warning,'
+             'explictiely specify the preserve_range parameter.',
+             DeprecationWarning)
```
hmaarrfk: I don't like `DeprecationWarning`, it tends to get hidden if it isn't called directly by `__main__`, and users won't get the warning. Let it be a standard warning, or `FutureWarning`.

```suggestion
             stacklevel=2)
```

For us, Deperecations will actually break things for users in 2 versions (1 year).

skimage/transform/radon_transform.py
```diff
@@ -62,6 +67,17 @@ def radon(image, theta=None, circle=True):
     if theta is None:
         theta = np.arange(180)
 
+    if preserve_range is None and np.issubdtype(image.dtype, np.integer):
+        warn('Image dtype is not float. By default radon will assume'
+             'you want to preserve the range of your image'
+             '(preserve_range=True). In scikit-image 0.18 this behavior will'
```
hmaarrfk: you need spaces just before each closing quote.

skimage/transform/radon_transform.py
rfezzani: Done

skimage/transform/radon_transform.py
```diff
@@ -107,13 +123,14 @@ def radon(image, theta=None, circle=True):
 
     def build_rotation(theta):
         T = np.deg2rad(theta)
-        R = np.array([[np.cos(T), np.sin(T), 0],
-                      [-np.sin(T), np.cos(T), 0],
+        c, s = np.cos(T), np.sin(T)
```
hmaarrfk:
```suggestion
        cos_T, sin_T = np.cos(T), np.sin(T)
```

skimage/transform/radon_transform.py
```diff
@@ -107,13 +123,14 @@ def radon(image, theta=None, circle=True):
...
+        R = np.array([[c, s, 0],
```
hmaarrfk:
```suggestion
        R = np.array([[cos_T, sin_T, 0],
```

skimage/transform/radon_transform.py
```diff
@@ -107,13 +123,14 @@ def radon(image, theta=None, circle=True):
...
+                      [-s, c, 0],
```
hmaarrfk:
```suggestion
                      [-sin_T, cos_T, 0],
```

skimage/transform/tests/test_radon_transform.py
```diff
@@ -80,21 +81,31 @@ def check_radon_center(shape, circle):
 
 shapes_for_test_radon_center = [(16, 16), (17, 17)]
 circles_for_test_radon_center = [False, True]
+dtype_for_test_radon_center = [np.float64, np.float32, np.uint8, bool]
+preserve_range_for_test_radon_center = [False, True]
 
 
-@testing.parametrize("shape, circle",
+@testing.parametrize("shape, circle, dtype, preserve_range",
```
hmaarrfk: this is extremely confusing. You can just have multiple `parametrize` decorators which becomes much more readable. One decorator per parameter. itertools product is pretty obscur for science programming.

skimage/transform/tests/test_radon_transform.py
```diff
@@ -80,21 +81,31 @@ def check_radon_center(shape, circle):
...
 rectangular_shapes = [(32, 16), (33, 17)]
+dtype_for_test_radon_center = [np.float64, np.float32, np.uint8, bool]
+preserve_range_for_test_radon_center = [True, False]
 
 
-@testing.parametrize("shape", rectangular_shapes)
-def test_radon_center_rectangular(shape):
-    check_radon_center(shape, False)
+# @testing.parametrize("shape", rectangular_shapes)
+@testing.parametrize("shape, dtype, preserve_range",
```
hmaarrfk: same thing for this use of itertools product.

skimage/transform/tests/test_radon_transform.py
```diff
@@ -80,21 +81,31 @@ def check_radon_center(shape, circle):
...
+                                       circles_for_test_radon_center,
+                                       dtype_for_test_radon_center,
+                                       preserve_range_for_test_radon_center))
+def test_radon_center(shape, circle, dtype, preserve_range):
+    check_radon_center(shape, circle, dtype, preserve_range)
 
 
 rectangular_shapes = [(32, 16), (33, 17)]
+dtype_for_test_radon_center = [np.float64, np.float32, np.uint8, bool]
+preserve_range_for_test_radon_center = [True, False]
```
hmaarrfk: once you stop using itertools product, these definitions are likely useless too.

skimage/transform/tests/test_radon_transform.py
rfezzani: I agree, just wanted to keep the same test structure as before. Can I also rename the parameter lists :smile: ?

hmaarrfk: you can just remove the fact that they are defined in the first place.
The list is easily defined inline

jni:
Great stuff! Thanks @rfezzani (again!) for your work and thanks @hmaarrfk for the detailed reviews, which made my job very easy. #pushthegreenbutton =P

PR #4134, opened 2019-09-02
enable multichannel images in drawing module using shape

k-dominik:
## Description

Most of the `skimage.drawing` functions allow to pass a `shape` parameter in order to confine the output coordinates to a given shape. These functions assume 2D images. All functions, except `rectangle` would also gracefully handle images with a channel dimension.

In this PR I clarified how the `shape` parameter is interpreted in the drawing docstrings. Furthermore, this PR adds the same behavior to `skimage.drawing.rectangle`, such that it will properly handle 2D images with channel (but still, 3D images are handled by this as well. Dimensinality of the drawing area is determined via the `start` parameter).

I didn't bother opening an issue about this, since the fix is so small

## Checklist

- [x] Unit tests

Not applicable checks:

  existing benchmark

## For reviewers

  later.

### DIFF
--- a/skimage/draw/_draw.pyx
+++ b/skimage/draw/_draw.pyx
@@ -20,7 +20,8 @@ def _coords_inside_image(rr, cc, shape, val=None):
         Indices of pixels.
     shape : tuple
         Image shape which is used to determine the maximum extent of output
-        pixel coordinates.
+        pixel coordinates.  Must be at least length 2. Only the first two values
+        are used to determine the extent of the input image.
     val : (N, D) ndarray of float, optional
         Values of pixels at coordinates ``[rr, cc]``.
 
--- a/skimage/draw/draw.py
+++ b/skimage/draw/draw.py
@@ -792,8 +792,9 @@ def rectangle(start, end=None, extent=None, shape=None):
     tl, br = _rectangle_slice(start=start, end=end, extent=extent)
 
     if shape is not None:
-        br = np.minimum(shape, br)
-        tl = np.maximum(np.zeros_like(shape), tl)
+        n_dim = len(start)
+        br = np.minimum(shape[0:n_dim], br)
+        tl = np.maximum(np.zeros_like(shape[0:n_dim]), tl)
     coords = np.meshgrid(*[np.arange(st, en) for st, en in zip(tuple(tl),
                                                                tuple(br))])
     return coords
--- a/skimage/draw/draw.py
+++ b/skimage/draw/draw.py
@@ -13,7 +13,8 @@ def _ellipse_in_shape(shape, center, radii, rotation=0.):
     Parameters
     ----------
     shape :  iterable of ints
-        Shape of the input image.  Must be length 2.
+        Shape of the input image.  Must be at least length 2. Only the first
+        two values are used to determine the extent of the input image.
     center : iterable of floats
         (row, column) position of center inside the given shape.
     radii : iterable of floats
@@ -53,7 +54,8 @@ def ellipse(r, c, r_radius, c_radius, shape=None, rotation=0.):
     shape : tuple, optional
         Image shape which is used to determine the maximum extent of output pixel
         coordinates. This is useful for ellipses which exceed the image size.
-        By default the full extent of the ellipse are used.
+        By default the full extent of the ellipse are used. Must be at least
+        length 2. Only the first two values are used to determine the extent.
     rotation : float, optional (default 0.)
         Set the ellipse rotation (rotation) in range (-PI, PI)
         in contra clock wise direction, so PI/2 degree means swap ellipse axis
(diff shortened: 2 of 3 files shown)


hmaarrfk:
Thanks for your contribution.

This seems to hit at a pretty core issue in the direction scikit-image is trying to go. Specifically, that is supporting higher dimentional images.

One challenge currently being faced is that we have to "guess" at the last dimension means. 

* It is a "channel" dimension (think RGBA) <- this is the case you are considering.
* It is the width of the image <- the more common case in very high dimensional images.

The current *trend* has been to move toward the later. I guess this is because many at scikit-image are interested in microscopy, where the number of channels are many more than the standard RGB images.

One example why this is important is the color module. It would be really nice to convert large multi-dimensional images to grayscale using `rgb2gray`
<details>

```
In [8]: rgb2gray(np.zeros((4,) * 3 + (3,)))                                                    
Out[8]: 
array([[[0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]],

       [[0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]],

       [[0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]],

       [[0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.],
        [0., 0., 0., 0.]]])

In [9]: rgb2gray(np.zeros((4,) * 4 + (3,)))                                                    
ValueError                                Traceback (most recent call last)
<ipython-input-9-257482b830ff> in <module>
----> 1 rgb2gray(np.zeros((4,) * 4 + (3,)))

~/miniconda3/envs/mcam_dev/lib/python3.7/site-packages/skimage/color/colorconv.py in rgb2gray(rgb)
    797         return np.ascontiguousarray(rgb)
    798 
--> 799     rgb = _prepare_colorarray(rgb[..., :3])
    800     coeffs = np.array([0.2125, 0.7154, 0.0721], dtype=rgb.dtype)
    801     return rgb @ coeffs

~/miniconda3/envs/mcam_dev/lib/python3.7/site-packages/skimage/color/colorconv.py in _prepare_colorarray(arr)
    150         msg = ("the input array must be have a shape == (.., ..,[ ..,] 3)), " +
    151                "got (" + (", ".join(map(str, arr.shape))) + ")")
--> 152         raise ValueError(msg)
    153 
    154     return dtype.img_as_float(arr)

ValueError: the input array must be have a shape == (.., ..,[ ..,] 3)), got (4, 4, 4, 4, 3)
```
</details>

When scaling up your data, to use GPUs, or distributed arrays, it would be nice to be able to continue to use the work you've done with scikit-image. That said, it isn't possible due to the heavy desire to maintain backward compatibility.

I emphasize the word trend, as the library isn't fully ported to this new way of thinking. 

We are currently asking that new features that want to support channels in the last dimension do so explicitely through the `multichiannel` parameter. By default, this parameter should be `False`.

Search through the library for `multichannel=` parameters. They have docstrings that can help you figure out how to correctly implement this.

k-dominik:
Hi @hmaarrfk, thank you very much for the thorough considerations here. And apologies for this longish answer. I just want to make sure I understand the direction you want to go with this.

In the context of `rgb2gray`, in the first case it's clear to me that you have just a 3D+RGB volume, so it works. The second case should illustrate a batch of four of such images? so 4x 3D+RGB and `rgb2gray` should then work on each of those batch images, did I understand this correctly? When channel is last, it seems trivial to scale this up to arbitrary higher dimensions. The only problem would be if someone relied on it raising a `ValueError` (as a sort of shape check)...

I've looked at various examples where the `multichannel`  parameter is used. Now, in the context of drawing the api should support similarly drawing 2D in batches? More concretely, let's say we have a 3D+RGB image (still tangible, so I'll stick to this example for now). With the current codebase I could do the following

```python
image = np.zeros((10, 9, 8, 3))
rr, cc = skimage.draw.rectangle(start=(0, 0), extent=(5, 5))

if multichannel:
    image[..., rr, cc, :] = 1
else:
    image[..., rr, cc] = 1
```

so the only question I wanted to address is, whether one can supply a shape and how it should be interpreted. For the above case this would yield different result arrays, depending on the value of `multichannel` in case of clipping, so

```python
image = np.zeros((10, 9, 8, 3))
rr, cc = skimage.draw.rectangle(
    start=(0, 0), extent=(5, 5), shape=image.shape, multichannel=False)
```
would clip, while

```python 
rr, cc = skimage.draw.rectangle(
    start=(0, 0), extent=(5, 5), shape=image.shape, multichannel=True)
```
would not. The drawing function would _return 2D indices_ no matter how high dimensional the data is It would be then up to the user to apply those indices accordingly. Is this how you would want it to be implemented?

jni:
@k-dominik actually, and surprisingly, I disagree with @hmaarrfk here. This is an enhancement already, and doesn't break anyone's existing code. Almost all of our draw functions are 2D only so it's fine to keep them 2D only. (I only object to adding *new* 2D-only functionality ;) (And then only when it's not *intrinsically* 2D, like e.g. `circle` is.) So I'm :+1: for merging this! Thanks @k-dominik!

hmaarrfk:
Alright, I'm convinced. Thanks!

PR #4136, opened 2019-09-04
Refactor radon transform module

rfezzani:
## Description

The major changes introduced by this PR are:

- use `warp` instead of `_warp_fast` to ease the maintenance when an enhancement/fix is introduced to `_warp_cy` (`_warp_fast` is now called only by the `warp` function),
- remove unnecessary inner functions (`build_rotation` in `radon` and `angle_distance` in `order_angles_golden_ratio`),
- define `_get_fourier_filter` function to improve `iradon` readability,
- replace double dot product by a direct array definition for the creation of the rotation matrix in `radon`,
- simplify some `if` statements,
- use `scipy.ndimage.measurements.find_objects` to find reconstruction circle area instead of a hand made loop,
- use `functools.partial` to make `np.interp` call consistent with `scipy.interpolate.interp1d` call in `iradon`

This results in some lines drop and no API change is introduced.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/radon_transform.py
+++ b/skimage/transform/radon_transform.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 from scipy.interpolate import interp1d
-from ._warps_cy import _warp_fast
+from ._warps import warp
 from ._radon_transform import sart_projection_update
 from .._shared.fft import fftmodule
 from warnings import warn
@@ -62,26 +62,25 @@ def radon(image, theta=None, circle=True):
     if theta is None:
         theta = np.arange(180)
 
+    shape_min = min(image.shape)
+
     if circle:
-        radius = min(image.shape) // 2
-        c0, c1 = np.ogrid[0:image.shape[0], 0:image.shape[1]]
-        reconstruction_circle = ((c0 - image.shape[0] // 2) ** 2
-                                 + (c1 - image.shape[1] // 2) ** 2)
-        reconstruction_circle = reconstruction_circle <= radius ** 2
+        img_shape = np.array(image.shape)
+        radius = shape_min // 2
+        coords = np.array(np.ogrid[:image.shape[0], :image.shape[1]])
+        dist = ((coords - img_shape // 2) ** 2).sum(0)
+        reconstruction_circle = dist <= radius ** 2
         if not np.all(reconstruction_circle | (image == 0)):
             warn('Radon transform: image must be zero outside the '
                  'reconstruction circle')
         # Crop image to make it square
-        slices = []
-        for d in (0, 1):
-            if image.shape[d] > min(image.shape):
-                excess = image.shape[d] - min(image.shape)
-                slices.append(slice(int(np.ceil(excess / 2)),
-                                    int(np.ceil(excess / 2)
-                                        + min(image.shape))))
+        slices = ()
+        for val in image.shape:
+            if val > shape_min:
+                start = int(np.ceil((val - shape_min) / 2))
+                slices += (slice(start, start + shape_min), )
             else:
-                slices.append(slice(None))
-        slices = tuple(slices)
+                slices += (slice(None), )
         padded_image = image[slices]
     else:
         diagonal = np.sqrt(2) * max(image.shape)
@@ -95,25 +94,21 @@ def radon(image, theta=None, circle=True):
     # padded_image is always square
     if padded_image.shape[0] != padded_image.shape[1]:
         raise ValueError('padded_image must be a square')
-    radon_image = np.zeros((padded_image.shape[0], len(theta)))
     center = padded_image.shape[0] // 2
 
-    shift0 = np.array([[1, 0, -center],
-                       [0, 1, -center],
-                       [0, 0, 1]])
-    shift1 = np.array([[1, 0, center],
-                       [0, 1, center],
-                       [0, 0, 1]])
+    shift0 = np.eye(3)
+    shift0[:2, 2] = -center
+    shift1 = np.eye(3)
+    shift1[:2, 2] = center
+    R = np.eye(3)
 
-    def build_rotation(theta):
-        T = np.deg2rad(theta)
-        R = np.array([[np.cos(T), np.sin(T), 0],
-                      [-np.sin(T), np.cos(T), 0],
-                      [0, 0, 1]])
-        return shift1.dot(R).dot(shift0)
-
-    for i in range(len(theta)):
-        rotated = _warp_fast(padded_image, build_rotation(theta[i]))
+    radon_image = np.zeros((padded_image.shape[0], len(theta)))
+    for i, angle in enumerate(theta):
+        T = np.deg2rad(angle)
+        cos_T, sin_T = np.cos(T), np.sin(T)
+        R[:2, :2] = [[cos_T, sin_T],
+                     [-sin_T, cos_T]]
+        rotated = warp(padded_image, shift1.dot(R).dot(shift0))
         radon_image[:, i] = rotated.sum(0)
     return radon_image
 
--- a/skimage/transform/radon_transform.py
+++ b/skimage/transform/radon_transform.py
@@ -324,28 +324,29 @@ def order_angles_golden_ratio(theta):
     """
     interval = 180
 
-    def angle_distance(a, b):
-        difference = a - b
-        return min(abs(difference % interval), abs(difference % -interval))
-
-    remaining = list(np.argsort(theta))   # indices into theta
+    remaining_indices = list(np.argsort(theta))   # indices into theta
     # yield an arbitrary angle to start things off
-    index = remaining.pop(0)
-    angle = theta[index]
-    yield index
+    angle = theta[remaining_indices[0]]
+    yield remaining_indices.pop(0)
     # determine subsequent angles using the golden ratio method
     angle_increment = interval * (1 - (np.sqrt(5) - 1) / 2)
-    while remaining:
+    while remaining_indices:
+        remaining_angles = theta[remaining_indices]
         angle = (angle + angle_increment) % interval
-        insert_point = np.searchsorted(theta[remaining], angle)
-        index_below = insert_point - 1
-        index_above = 0 if insert_point == len(remaining) else insert_point
-        distance_below = angle_distance(angle, theta[remaining[index_below]])
-        distance_above = angle_distance(angle, theta[remaining[index_above]])
+        index_above = np.searchsorted(remaining_angles, angle)
+        index_below = index_above - 1
+        index_above %= len(remaining_indices)
+
+        diff_below = abs(angle - remaining_angles[index_below])
+        distance_below = min(diff_below % interval, diff_below % -interval)
+
+        diff_above = abs(angle - remaining_angles[index_above])
+        distance_above = min(diff_above % interval, diff_above % -interval)
+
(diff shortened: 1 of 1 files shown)


skimage/transform/radon_transform.py
```diff
@@ -128,6 +113,62 @@ def _sinogram_circle_to_square(sinogram):
     return np.pad(sinogram, pad_width, mode='constant', constant_values=0)
 
 
+def _get_fourier_filter(size, filter_name):
+    """Construct the Fourier filter
```
sciunto:
```suggestion
    """Construct the Fourier filter.
```

skimage/transform/radon_transform.py
```diff
@@ -128,6 +113,62 @@ def _sinogram_circle_to_square(sinogram):
...
+
+    This computation lessens artifacts and removes a small bias as
+    explained in [1], Chap 3. Equation 61
```
sciunto:
```suggestion
    explained in [1], Chap 3. Equation 61.
```

skimage/transform/radon_transform.py
```diff
@@ -128,6 +113,62 @@ def _sinogram_circle_to_square(sinogram):
...
+
+    Parameters
+    ----------
+    size: int
+        filter size.
+    filter_name: str, optional
```
sciunto:
```suggestion
    filter_name: str
```

skimage/transform/radon_transform.py
```diff
@@ -128,6 +113,62 @@ def _sinogram_circle_to_square(sinogram):
...
+        Filter used in frequency domain filtering. Ramp filter used by
+        default.  Filters available: ramp, shepp-logan, cosine,
```
sciunto:
```suggestion
        default. Filters available: ramp, shepp-logan, cosine,
```

skimage/transform/radon_transform.py
```diff
@@ -128,6 +113,62 @@ def _sinogram_circle_to_square(sinogram):
...
+        hamming, hann.  Assign None to use no filter.
```
sciunto:
```suggestion
        hamming, hann. Assign None to use no filter.
```

skimage/transform/radon_transform.py
```diff
@@ -128,6 +113,62 @@ def _sinogram_circle_to_square(sinogram):
...
+
+    Returns
+    -------
+        fourier_filter: ndarray
```
sciunto:
```suggestion
    fourier_filter: ndarray
```

skimage/transform/radon_transform.py
```diff
@@ -128,6 +113,62 @@ def _sinogram_circle_to_square(sinogram):
...
+            The computed Fourier filter.
```
sciunto:
```suggestion
        The computed Fourier filter.
```

skimage/transform/radon_transform.py
```diff
@@ -185,101 +226,66 @@ def iradon(radon_image, theta=None, output_size=None,
...
+    if angles_count != radon_image.shape[1]:
         raise ValueError("The given ``theta`` does not match the number of "
                          "projections in ``radon_image``.")
+
     interpolation_types = ('linear', 'nearest', 'cubic')
     if interpolation not in interpolation_types:
         raise ValueError("Unknown interpolation: %s" % interpolation)
-    if not output_size:
+
+    filter_types = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann', None]
```
sciunto:
```suggestion
    filter_types = ('ramp', 'shepp-logan', 'cosine', 'hamming', 'hann', None)
```

rfezzani:
Thank you @sciunto for your kind comments and for your review. I already made some tests and I observed no performance drop:

## Setup

```python
In [1]: from skimage.transform.radon_transform import order_angles_golden_ratio, radon, iradon          

In [2]: from skimage.data import shepp_logan_phantom                                                    

In [3]: image = shepp_logan_phantom()                                                                   

In [4]: theta = np.linspace(0., 180., max(image.shape), endpoint=False)                                 

In [5]: sinogram = radon(image, theta=theta, circle=True)  
```

## Actual version

```python
In [6]: %timeit -n10 -r10 iradon(sinogram, theta=theta, circle=True)                                    
572 ms ± 9.51 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)

In [7]: %timeit -n10 -r10 radon(image, theta=theta, circle=True)                                        
1.21 s ± 4.71 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)

In [8]: %timeit -n10 -r10 [x for x in order_angles_golden_ratio(theta)]                                 
6.96 ms ± 312 µs per loop (mean ± std. dev. of 10 runs, 10 loops each)
```

## Refactored version

```python
In [6]: %timeit -n10 -r10 iradon(sinogram, theta=theta, circle=True)                                    
572 ms ± 769 µs per loop (mean ± std. dev. of 10 runs, 10 loops each)

In [7]: %timeit -n10 -r10 radon(image, theta=theta, circle=True)                                        
1.2 s ± 3.18 ms per loop (mean ± std. dev. of 10 runs, 10 loops each)

In [8]: %timeit -n10 -r10 [x for x in order_angles_golden_ratio(theta)]                                 
6.66 ms ± 285 µs per loop (mean ± std. dev. of 10 runs, 10 loops each)
```

skimage/transform/radon_transform.py
```diff
@@ -128,6 +113,62 @@ def _sinogram_circle_to_square(sinogram):
...
+
+    This computation lessens artifacts and removes a small bias as
+    explained in [1], Chap 3. Equation 61.
+
+    Parameters
+    ----------
+    size: int
+        filter size.
+    filter_name: str
+        Filter used in frequency domain filtering. Ramp filter used by
```
sciunto: Sorry, I forgot to ask you to remove "Ramp filter used by default."

sciunto:
Perfect. we just need a second approval by another core dev.

rfezzani:
Thank you @sciunto, my last commit is just about replacing `np.pi * np.linspace(0, 1, size, endpoint=False )` by  `np.linspace(0, np.pi, size, endpoint=False )`.

rfezzani:
Given the golden ratio  γ. It's inverse `g = 1/ γ` verify `g² = (1 - g)` (See the appendix in [[1]](http://citeseerx.ist.psu.edu/viewdoc/download;jsessionid=00A7E960E348DFB7137C60702B6F134F?doi=10.1.1.666.3807&rep=rep1&type=pdf).
The actual implementation make use of `(1-g)` with `g` computed as `(np.sqrt(5) - 1) / 2`.
I used `scipy.constants.golden_ratio` to simplify the expression and replaced `(1-g) = 1 - (np.sqrt(5) - 1) / 2` by `g² = 1/golden_ratio**2`

hmaarrfk:
is this going to conflict with: https://github.com/scikit-image/scikit-image/pull/4131

which one do you want us to look at first?

rfezzani:
The will be merge conflicts, but fixing is not difficult. I tested both directions (#4131 in #4136 and #4136 in #4131), both are easy. So I let you decide about the priorities ;-)

rfezzani:
This PR makes #4128 pass the CI tests, while #4131 not.
So considering this PR first allows both #4131 and #4128 to be eligible for merge.

skimage/transform/radon_transform.py
```diff
@@ -63,25 +65,17 @@ def radon(image, theta=None, circle=True):
...
-        for d in (0, 1):
-            if image.shape[d] > min(image.shape):
-                excess = image.shape[d] - min(image.shape)
-                slices.append(slice(int(np.ceil(excess / 2)),
-                                    int(np.ceil(excess / 2)
-                                        + min(image.shape))))
-            else:
-                slices.append(slice(None))
-        slices = tuple(slices)
+        slices = find_objects((~outside_reconstruction_circle).astype(int))[0]
```
emmanuelle: calling `find_objects` is more elegant, but it takes a few ms for a 2000x2000 image. This is not a big deal since the whole function takes several seconds to complete, but if someday we accelerate massively this function (the loop on angles could easily be parallelized) it could be significant. Even then it would only represent around 10% of the execution time so I'm ok with this change, I just wanted to raise this point in case somebody wants to comment.

emmanuelle:
I just left one comment, but it looks good to me so I'll merge tomorrow if there's nothing to add. Note that it would be great to parallelize the loop on angles (if it is the bottleneck which I did not check, but it's likely).

skimage/transform/radon_transform.py
hmaarrfk: Seems silly to analyze the data instead of the metadata when looking for a slice....

rfezzani: I agree with you, the use of `find_object` requires also negating `outside_reconstruction_circle` and casing it to `int`, which is a lot of processing to find the desired slice...
My last commit removes the call to `find_object`.

rfezzani:
Thank you @emmanuelle for your feedback. I also think that the `radon` function could be optimized by parallelizing the loop on angles but this will be at the cost of a larger memory use... I will try this in an other PR once this one is merged ;-). Thank you again.

skimage/transform/radon_transform.py
```diff
@@ -92,28 +88,19 @@ def radon(image, theta=None, circle=True):
...
-                       [0, 1, center],
-                       [0, 0, 1]])
-
-    def build_rotation(theta):
-        T = np.deg2rad(theta)
-        R = np.array([[np.cos(T), np.sin(T), 0],
-                      [-np.sin(T), np.cos(T), 0],
+    for i, T in enumerate(np.deg2rad(theta)):
+        cos_T, sin_T = np.cos(T), np.sin(T)
+        R = np.array([[cos_T, sin_T, -center * (cos_T + sin_T - 1)],
```
soupault: Please, use lowercase for the variable names.

skimage/transform/radon_transform.py
```diff
@@ -185,107 +227,71 @@ def iradon(radon_image, theta=None, output_size=None,
...
         raise ValueError("The given ``theta`` does not match the number of "
                          "projections in ``radon_image``.")
+
     interpolation_types = ('linear', 'nearest', 'cubic')
     if interpolation not in interpolation_types:
         raise ValueError("Unknown interpolation: %s" % interpolation)
-    if not output_size:
+
+    filter_types = ('ramp', 'shepp-logan', 'cosine', 'hamming', 'hann', None)
+    if filter not in filter_types:
```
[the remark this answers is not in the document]
emmanuelle: oh yes thank you for catching this @soupault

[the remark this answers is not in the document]
rfezzani: I fully agree with you, but this is the original argument name!
I also  opened an equivalent [issue](https://github.com/scipy/scipy/issues/10010) in `scipy` where `input` is used as argument name in many functions. This issue was closed and will not be fixed for retro-compatibility reasons :-(
I proposed this [gist](https://gist.github.com/rfezzani/002181c8667ec4c671421a4d938167eb) as a solution for this problem. I don't know if `skimage` accepts such a solution...

skimage/transform/radon_transform.py
rfezzani: Done. Thank you @soupault for your review.

PR #4137, opened 2019-09-04
Forward-port release notes for 0.14.4

jni:
## Description

0.14.4 is out, but the release notes, I think, should be on master, so that they appear on the latest website. (@stefanv correct me if I'm wrong.)

### DIFF
--- a/doc/release/release_0.14.rst
+++ b/doc/release/release_0.14.rst
@@ -1,3 +1,106 @@
+Announcement: scikit-image 0.14.4
+=================================
+
+We're happy to announce the release of scikit-image v0.14.4!
+
+As a reminder, 0.14.x is the final version of scikit-image with support for
+Python 2.7, and will receive critical bug fixes until Jan 1, 2020. If you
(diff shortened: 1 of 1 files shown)


doc/release/release_0.14.rst
```diff
@@ -1,3 +1,106 @@
...
+- Backport PR #3943 on branch v0.14.x (Update the joblib link in tutorial_parallelization.rst) (#3944)
+
+3 authors added to this release [alphabetical by first name or login]
+---------------------------------------------------------------------
+- Alexandre de Siqueira
+- Juan Nunez-Iglesias
+- Mark Harfouche
+
+
+Announcement: scikit-image 0.14.3
```
hmaarrfk: opps, we never added these..

hmaarrfk:
I think this should really push us to release 0.16. It is strange that users can use 0.14.4 to overcome such a big bug. If only we had time

stefanv:
How much longer will it take to release 0.16 than 0.14.4?

jni:
I intend to do this in the next couple of days also. ;) @NelleV expressed interest in looking at our release process. =)

PR #4021, opened 2019-07-13
Enable packagers to disable failures on warnings.

hmaarrfk:
## Description
The `SKIMAGE_TEST_STRICT_WARNINGS` now enables packagers to avoid failures due to warnings. This often happens when warnings pop up due to newer versions of dependencies.

I added a few tests for this functionality too.

As somebody who maintains the conda-forge package, I would like this backported to both 0.15 and 0.14.

I couldn't find where the `expected_warnings` context manager was tested, so I created new tests that test this added functionality.

## Checklist

- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Gallery example in `./doc/examples` (new features only)
- [ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
- [ ] Unit tests
- [ ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

## For reviewers

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

PR #4144, opened 2019-09-07
Restore deprecated coordinates arg to regionprops

### LINKED ISSUE #4029: Bring back coordinates='rc' parameter
## Description

While we were trying to fix #3667, @holly-c-gibbs, @hmaarrfk, @alexdesiqueira and I noticed that `coordinates='rc'` is actually gone as a keyword argument for regionprops. I think that is the incorrect approach for the deprecation: we should change the default value to 'rc', and maybe warn that 'xy' support will disappear, *but* we should not break user code that worked *only* on 0.14 and 0.15 when we release 0.16. Specifically, the *recommended*, *correct* way to silence the warnings was to pass `coordinates='rc'` to the function, but that code will now break in 0.16.

## Way to reproduce
```python
```

## Version information
```python
```

```python
# your output here

```

jni:
Fixes #4029

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_regionprops.py
+++ b/skimage/measure/_regionprops.py
@@ -622,7 +622,7 @@ def regionprops_table(label_image, intensity_image=None, cache=True,
 
 def regionprops(label_image, intensity_image=None, cache=True,
                 coordinates=None):
-    """Measure properties of labeled image regions.
+    r"""Measure properties of labeled image regions.
 
     Parameters
     ----------
@@ -645,6 +645,17 @@ def regionprops(label_image, intensity_image=None, cache=True,
     coordinates : DEPRECATED
         This argument is deprecated and will be removed in a future version
         of scikit-image.
+
+        See :ref:`Coordinate conventions <numpy-images-coordinate-conventions>`
+        for more details.
+
+        .. deprecated:: 0.16.0
+            Use "rc" coordinates everywhere. It may be sufficient to call
+            ``numpy.transpose`` on your label image to get the same values as
+            0.15 and earlier. However, for some properties, the transformation
+            will be less trivial. For example, the new orientation is
+            :math:`\frac{\pi}{2}` plus the old orientation.
+
 
     Returns
     -------
--- a/skimage/measure/_regionprops.py
+++ b/skimage/measure/_regionprops.py
@@ -1,3 +1,4 @@
+from warnings import warn
 from math import sqrt, atan2, pi as PI
 import numpy as np
 from scipy import ndimage as ndi
@@ -619,7 +620,8 @@ def regionprops_table(label_image, intensity_image=None, cache=True,
     return _props_to_dict(regions, properties=properties, separator=separator)
 
 
-def regionprops(label_image, intensity_image=None, cache=True):
+def regionprops(label_image, intensity_image=None, cache=True,
+                coordinates=None):
     """Measure properties of labeled image regions.
 
     Parameters
@@ -640,6 +642,9 @@ def regionprops(label_image, intensity_image=None, cache=True):
         Determine whether to cache calculated properties. The computation is
         much faster for cached properties, whereas the memory consumption
         increases.
+    coordinates : DEPRECATED
+        This argument is deprecated and will be removed in a future version
+        of scikit-image.
 
     Returns
     -------
@@ -818,6 +823,22 @@ def regionprops(label_image, intensity_image=None, cache=True):
 
     if not np.issubdtype(label_image.dtype, np.integer):
         raise TypeError('Label image must be of integer type.')
+
+    if coordinates is not None:
+        if coordinates == 'rc':
+            msg = ('The coordinates keyword argument to skimage.measure.'
+                   'regionprops is deprecated. All features are now computed '
+                   'in rc (row-column) coordinates. Please remove '
+                   '`coordinates="rc"` from all calls to regionprops before '
+                   'updating scikit-image.')
+            warn(msg, stacklevel=2)
+        else:
+            msg = ('Values other than "rc" for the "coordinates" argument '
+                   'to skimage.measure.regionprops are no longer supported. '
+                   'You should update your code to use "rc" coordinates and '
+                   'stop using the "coordinates" argument, or use skimage '
+                   'version 0.15.x or earlier.')
+            raise ValueError(msg)
 
     regions = []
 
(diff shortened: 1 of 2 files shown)


skimage/measure/_regionprops.py
```diff
@@ -819,6 +824,22 @@ def regionprops(label_image, intensity_image=None, cache=True):
...
+                   'in rc (row-column) coordinates. Please remove '
+                   '`coordinates="rc"` from all calls to regionprops before '
+                   'updating scikit-image.')
+            warn(msg, stacklevel=2)
+        else:
+            msg = ('Values other than "rc" for the "coordinates" argument '
+                   'to skimage.measure.regionprops are no longer supported. '
+                   'You should update your code to use "rc" coordinates and '
+                   'stop using the "coordinates" argument, or use skimage '
+                   'version 0.15.x or earlier.')
```
hmaarrfk:
```suggestion
                   'version 0.15.x or earlier. The functions"transpose" or "moveaxis"'
                   'may help you transition your code.')
```

emmanuelle: `np.transpose` and `np.moveaxis` ?

jni: Honestly I'm not super in-love with this suggestion. It seems to me to be too vague to be useful. We could add an "xy to rc transition guide", but I don't think it should hold up this PR or release...

hmaarrfk: I agree it shouldn't hold up this PR, but the most frustrating thing to hit is:

> Aw man my code isn't working, let me type `pip install scikit-image=0.14` 

and users now get stuck on 0.14 for much longer than they should. Pointhing them to `np.moveaxis` or `np.transpose` alleviates that.

hmaarrfk: or my favorite

`sudo pip install scikit-image=0.14`

which will totally break their modern systems in 1 year.

skimage/measure/_regionprops.py
```diff
@@ -640,6 +642,9 @@ def regionprops(label_image, intensity_image=None, cache=True):
         Determine whether to cache calculated properties. The computation is
         much faster for cached properties, whereas the memory consumption
         increases.
+    coordinates : DEPRECATED
+        This argument is deprecated and will be removed in a future version
+        of scikit-image.
```
lagru: Maybe remove this from the docstring? Is that current library practice? The warning message gives all the information that is needed in my opinion and is likely the way users will notice this depreciation anyway.

skimage/measure/tests/test_regionprops.py
```diff
@@ -2,6 +2,8 @@
 
 import numpy as np
 from numpy import array
+
+from skimage._shared._warnings import expected_warnings
```
lagru: What is the current opinion on this vs `pytest.warns`?

skimage/measure/_regionprops.py
```diff
@@ -819,6 +824,22 @@ def regionprops(label_image, intensity_image=None, cache=True):
...
         raise TypeError('Label image must be of integer type.')
 
+    if coordinates is not None:
+        if coordinates == 'rc':
+            msg = ('The coordinates keyword argument to skimage.measure.'
+                   'regionprops is deprecated. All features are now computed '
+                   'in rc (row-column) coordinates. Please remove '
+                   '`coordinates="rc"` from all calls to regionprops before '
+                   'updating scikit-image.')
+            warn(msg, stacklevel=2)
```
lagru:
```suggestion
            warn(msg, stacklevel=2, category=DeprecationWarning)
```

lagru: Might be worth updating the example in the contribution guide with the parameters `stacklevel` and `category` as well...

skimage/measure/_regionprops.py
hmaarrfk: We've been trying to use markers like `.. deprecated :` to make it more explicit and to help people avoid warnings before their code is run
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=deprecated#directive-deprecated

skimage/measure/_regionprops.py
hmaarrfk: `DeprecationWarning` are siliced if things are not invoked from `main` directly. I'm not sure that is what we want. Maybe `FutureWarning`

skimage/measure/tests/test_regionprops.py
hmaarrfk: Our warnings can be made optional

https://github.com/scikit-image/scikit-image/pull/4021

lagru: Thanks for the pointer. I don't know the full history behind that function but was `pytest.mark.skipif(os.environ.get('SKIMAGE_TEST_STRICT_WARNINGS', '1'))` discussed as an alternative? Seems to me that the behavior could be entirely be replaced with pytest.

skimage/measure/_regionprops.py
lagru: Agree, after [looking that up](https://docs.python.org/3.7/library/exceptions.html#FutureWarning) `FutureWarning` (for end users) seems like a better option than `DeprecationWarning` (for developers).

skimage/measure/_regionprops.py
lagru: In that case isn't that directive missing here? Not that the directive seems to have any special styling in the current theme but that may change in the future.

skimage/measure/tests/test_regionprops.py
hmaarrfk: Yeah, that might work. There are places though were it is still nice to have the test run even without the strict warning test.

These cleanups have a hard time getting through the review process. so i doubt anything sweeping will get done.

lagru: I guess there are more pressing matters right now. :+1:

skimage/measure/_regionprops.py
jni: My feeling is that you have to know exactly what you need to properly transition your code. Calling `np.transpose` on your image *may* fix your xy/rc problem, but it may not. For example, the orientation will change sign compared to the previous convention.

skimage/measure/_regionprops.py
jni: Added the directive, thanks for the info!

skimage/measure/tests/test_regionprops.py
jni: I've been in favour of fully embracing pytest, but it's still pretty contentious so let's leave it for another time...

skimage/measure/_regionprops.py
jni: I've made this a FutureWarning.

jni:
@lagru thanks for the review! I believe I've addressed your comments, and also @hmaarrfk's.

lagru:
I hope merging this as a squashed commit was the right thing to do. I see both approaches used here...

PR #4146, opened 2019-09-08
Remove duplicates from perimeter functions in draw module

### LINKED ISSUE #4141: Duplicate coordinates produced by draw.circle_perimeter
## Description

Hi, I've noticed that there are (4 or 8) duplicate coordinates produced by `draw.circle_perimeter`. Although this is probably not a bug, I was surprised nevertheless.

It can lead to unexpected behavior, depending on how the coordinates are used for indexing (see below). For example, I think [`transform._hough_circle`](https://github.com/scikit-image/scikit-image/blob/master/skimage/transform/_hough_transform.pyx#L75-L96) is accumulating evidence twice from these duplicates (I doubt it matters in practice though). 

Maybe it would be a good idea to mention this in the docstring? Or remove the duplicates?

## Way to reproduce
```python
>>> import numpy as np
>>> from skimage.draw import circle_perimeter

>>> rr,cc = circle_perimeter(1,1, radius=1)
>>> print(sorted(zip(rr,cc)))
[(0, 1), (0, 1), (1, 0), (1, 0), (1, 2), (1, 2), (2, 1), (2, 1)]

>>> img = np.zeros((3,3),int)
>>> img[rr,cc] += 1
>>> print(img)
[[0 1 0]
 [1 0 1]
 [0 1 0]]

>>> img = np.zeros((3,3),int)
>>> for r,c in zip(rr,cc):
>>>     img[r,c] += 1
>>> print(img)
[[0 2 0]
 [2 0 2]
 [0 2 0]]

>>> for m in ('bresenham', 'andres'):
>>>     coords = [list(zip(*circle_perimeter(0,0, radius=r, method=m))) for r in range(1,100)]
>>>     duplicates = [len(c) - len(set(c)) for c in coords]
>>>     print(m.rjust(9), np.unique(duplicates, return_counts=True))
bresenham (array([4, 8]), array([50, 49]))
   andres (array([4, 8]), array([27, 72]))
```

## Version information
```python
```

```
3.6.4 |Anaconda, Inc.| (default, Mar 13 2018, 01:15:57) 
[GCC 7.2.0]
Linux-3.10.0-957.27.2.el7.x86_64-x86_64-with-centos-7.6.1810-Core
scikit-image version: 0.14.2
numpy version: 1.14.2
```

lagru:
## Description

- Fixes #4141. As described, the functions `circle_perimeter`, `circle_perimeter_aa` and `ellipse_perimeter` might return duplicate coordinates. This is likely unintuitive. 
- Furthermore I took the chance to slightly reduce the Cython footprint for the `skimage.draw` module.

Another point to discuss: as you may see the test `skimage.draw.tests.test_draw.Test_circle_perimeter_aa.test_normal` fails. The cause for the mismatching pixels is that the function returns duplicates with conflicting intensity values. I'm am not sure how to handle this. Perhaps this hints at a bug inside `circle_perimeter_aa`?

## Checklist

- <strike>[ ] Gallery example in `./doc/examples` (new features only)</strike>
- <strike>[ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark</strike>
- [x] Unit tests

## For reviewers

  later.


lagru:
I took the liberty to hide the PEP8 related comment as resolved as it doesn't like the formatting of the reference arrays in the tests.

skimage/draw/tests/test_draw.py
```diff
@@ -577,125 +596,129 @@ def test_ellipse_rotated():
...
-         [0, 0, 0, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
-         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]]
-    )
-    assert_array_equal(img, img_)
+class Test_ellipse_perimeter:
```
hmaarrfk: whats the value in adding this? IMO it kills the git history for little value. But that is just me.

lagru: Oh, I honestly thought GitHub's diff viewer would handle this more gracefully. Consolidating related tests in classes is nice to have but I agree in this case it's not worth the review burden. :wink:

hmaarrfk: THe diff can handle it, you can hide whitespace, but the `git blame` will be pretty bad is my assumption.

lagru: Good point.

skimage/draw/draw.py
```diff
@@ -520,7 +554,20 @@ def circle_perimeter(r, c, radius, method='bresenham', shape=None):
            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
     """
-    return _circle_perimeter(r, c, radius, method, shape)
+    if not np.can_cast(r, np.integer) or not np.can_cast(c, np.integer):
+        raise TypeError("r and c must be integers")
+
+    rr, cc = _circle_perimeter(radius, method)
+    rr, cc = np.unique(np.stack([rr, cc]), axis=1)
```
hmaarrfk: Is this the key line in all of this?

skimage/draw/draw.py
```diff
@@ -520,7 +554,20 @@ def circle_perimeter(r, c, radius, method='bresenham', shape=None):
...
+
+    rr += r
+    cc += c
+    if shape is not None:
+        keep = _inside_image(rr, cc, shape)
+        rr = rr[keep]
+        cc = cc[keep]
```
jni: This pattern is repeated over and over. If you're going to change the `_inside_image` logic (which seems unnecessary, but ok), perhaps we should let the function do the work of filtering as well? ie:

```python
rr, cc = _inside_image((rr, cc), shape)
(rr, cc), values = _inside_image((rr, cc), shape, values=values)
(pp, rr, cc), values = _inside_image((pp, rr, cc), shape, values=values)
```

If we want to avoid changing the return based on the input, we can always return values, returning None if none were provided.

skimage/draw/_draw.pyx
```diff
@@ -398,18 +340,20 @@ def _circle_perimeter_aa(Py_ssize_t r_o, Py_ssize_t c_o,
         val.extend([1 - dceil, dceil] * 8)
```
hmaarrfk: Is there a better way to get unique coordinates than to check the whole array for uniqueness?

jni (commented):
@lagru I don't know anything about anti-aliased circle generation. My git sleuthing suggests that this was originally implemented by @sciunto, who perhaps has ideas about the duplicate but inconsistent values there?

I definitely don't think we need classes for tests, fwiw. Glad that bit's gone... =)

Other than that, I made a grand total of *one* useful suggestion (I think), which is to implement the coordinate filtering within `_inside_image` to avoid so much repeated boilerplate code.

skimage/draw/draw.py
sciunto: I have not idea ether. That's why I made a +1 on @lagru 's comment in the issue. At that time, I basically followed the implementation linked in the references.

skimage/draw/draw.py
lagru: Yes, this as well as the other two similar lines are the ones addressing the duplicates returned by the Cython implementations. The rest is just maintenance that's not strictly necessary but, as I felt, simplifies the Cython side a little.

skimage/draw/_draw.pyx
lagru: Maybe. I don't fully understand the algorithms that generate the duplicates. It's easy to see why some duplicates are generated (e.g. in `_circle_perimeter` if one of the variables `r` or `c` is 0 the leading signs become meaningless).

Do you feel that performance is a problem here? I did a quick test and using `np.unique` slows down `circle_perimeter` by the factor 4 (performance seems similar when commenting that out). I'm surprised that `np.unique` has that big of an impact.

skimage/draw/draw.py
lagru: > This pattern is repeated over and over.

Yes but considering how stupidly simple that snippet is one could argue that the code is actually better. But that's not something I feel strongly about. :wink:

skimage/draw/draw.py
```diff
@@ -520,7 +554,20 @@ def circle_perimeter(r, c, radius, method='bresenham', shape=None):
            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
     """
-    return _circle_perimeter(r, c, radius, method, shape)
+    if not np.can_cast(r, np.integer) or not np.can_cast(c, np.integer):
+        raise TypeError("r and c must be integers")
```
lagru: Removing these parameters from the Cython signature made this type checking necessary in order to keep the old behavior. Is this the best way to go about this?

skimage/draw/draw.py
hmaarrfk: Got it. That makes sense. I just wanted to make sure I didn't miss anything else

skimage/draw/draw.py
hmaarrfk: Why do r and c need to be integers?

lagru: The old Cython signature on master declared these as `Py_ssize_t`. So technically we need to 
So raising a TypeError continues the old behavior. An alternative would be to cast/round them to integers but that would technically change the API.

hmaarrfk: I think the `index` function will do what you want. It was introduced in python to define things that behave like Pyssize

skimage/draw/_draw.pyx
hmaarrfk: i guess you could check if r == c i many cases, but in the aa case it is a little trickier.

lagru: That's an idea. But we have to check for 0 as well. This

```python
if r == c == 0:
    rr.append(r)
    cc.append(c)
elif r == 0:
    rr.extend([r, r, c, -c])
    cc.extend([c, -c, r, r])
elif c == 0:
    rr.extend([r, -r, c, c])
    cc.extend([c, c, r, -r])
elif r == c:
    rr.extend([r, -r, r, -r])
    cc.extend([c, c, -c, -c])
else:
    rr.extend([r, -r, r, -r, c, -c, c, -c])
    cc.extend([c, c, -c, -c, r, r, -r, -r])
```

seems to work. Did a few quick tests and this is nearly the same as the old implementation. The `np.unique` approach is consistently 3 to 4x slower.

lagru: That solution would work for `circle_perimeter_aa` as well but the cases would get kind of tedious because we'd have to consider `r - 1` and `1 - r` as well...

skimage/draw/draw.py
lagru: Thanks, using the `index` function now.

skimage/draw/_draw.pyx
```diff
@@ -288,17 +248,12 @@ def _circle_perimeter(Py_ssize_t r_o, Py_ssize_t c_o, Py_ssize_t radius,
...
     cdef list cc = list()
-
     cdef Py_ssize_t c = 0
     cdef Py_ssize_t r = radius
     cdef Py_ssize_t d = 0
 
-    cdef double dceil = 0
-    cdef double dceil_prev = 0
-
     cdef char cmethod
```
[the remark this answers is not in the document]
lagru: I would agree with you if there wasn't also a calculation and assignment to `d` in both branches. Recalculating `d` to save an elif-branch seems more confusing to me.

lagru:
[`test_circle_perimeter_aa`](https://github.com/scikit-image/scikit-image/actions/runs/3118531960/jobs/5057887541#step:5:658) fails for because a9fc1dced48834b94e8e08675bd728d99dd3085f exposes the bug documented in #6541.

stefanv:
It should be possible to avoid generating duplicates by simply comparing each coordinate as it gets generated. As well as then checking first + last coords.

src/skimage/draw/_draw.pyx
```diff
@@ -398,18 +340,20 @@ def _circle_perimeter_aa(Py_ssize_t r_o, Py_ssize_t c_o,
         val.extend([1 - dceil, dceil] * 8)
```
lagru: @stefanv, regarding your [comment](https://github.com/scikit-image/scikit-image/pull/4146#issuecomment-4503616017): Did you see this thread. 

It certainly is possible to generate without duplicates. However, this will complicate the cython functions somewhat with more logic. Shall I try?

lagru: Maybe we can use a set-based solution for the more complicated algorithms such as `_circle_perimeter_aa` or `_ellipse_perimeter`?
 
Some quick testing with `_circle_parameter` suggests, that it is significantly faster than the approach using `np.unique`, but still slower than without filtering duplicates.

<details><summary>Patch</summary>

```patch
Subject: [PATCH] Avoid duplicate coords with set in `_circle_perimeter`
---
Index: src/skimage/draw/_draw.pyx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/skimage/draw/_draw.pyx b/src/skimage/draw/_draw.pyx
--- a/src/skimage/draw/_draw.pyx	(revision 26d68ff0d1de07ada286020fbac9af1d6d057ba3)
+++ b/src/skimage/draw/_draw.pyx	(revision e467b74259046767898e1731947803b4837e3631)
@@ -251,8 +251,7 @@
     .. [2] E. Andres, "Discrete circles, rings and spheres", Computers &
            Graphics, 18 (1994) 695-706.
     """
-    cdef list rr = list()
-    cdef list cc = list()
+    cdef set coords = set()
     cdef Py_ssize_t c = 0
     cdef Py_ssize_t r = radius
     cdef Py_ssize_t d = 0
@@ -268,8 +267,9 @@
         raise ValueError('Wrong method')
 
     while r >= c:
-        rr.extend([r, -r, r, -r, c, -c, c, -c])
-        cc.extend([c, c, -c, -c, r, r, -r, -r])
+        coords |= {
+            (r, c), (-r, c), (r, -c), (-r, -c), (c, r), (-c, r), (c, -r), (-c, -r)
+        }
 
         if cmethod == b'b':
             if d < 0:
@@ -290,7 +290,7 @@
                 r = r - 1
                 c = c + 1
 
-    return np.array(rr, dtype=np.intp), np.array(cc, dtype=np.intp)
+    return tuple(np.array(cc, dtype=np.intp) for cc in zip(*coords))
 
 
 def _circle_perimeter_aa(Py_ssize_t radius):
Index: src/skimage/draw/draw.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/skimage/draw/draw.py b/src/skimage/draw/draw.py
--- a/src/skimage/draw/draw.py	(revision 26d68ff0d1de07ada286020fbac9af1d6d057ba3)
+++ b/src/skimage/draw/draw.py	(revision e467b74259046767898e1731947803b4837e3631)
@@ -615,7 +615,6 @@
            [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
     """
     rr, cc = _circle_perimeter(radius, method)
-    rr, cc = np.unique(np.stack([rr, cc]), axis=1)
 
     try:
         rr += operator.index(r)

```

</details> 

```python
%timeit circle_perimeter(400, 500, 300)
```
With `set`:
476 μs ± 1.16 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

With `np.unique`:
1.25 ms ± 4.11 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

On `main`:
166 μs ± 49.7 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

stefanv: Your solution here looks correct, @lagru, catching those zero and `r == c` duplicates.

For the anti-aliasing case, that becomes a fair number of cases. At least there is no case where `1 - r == r - 1`, but we do have to deal with `r == 0`, `c == 0`, and `1 - r == -c`, `r - 1 == c` etc. (edit: reading the code below, realize that these are the same case, which helps!)

You'd have to double-check obviously, but Gemini's attempt at this makes me think it is perfectly viable:

```python
if radius == 0:
    rr = [0]
    cc = [0]
    val = [1.0]
else:
    # Initial cardinal points: (r, 0), (-r, 0), (0, r), (0, -r)
    # These are unique since radius > 0
    rr = [radius, -radius, 0, 0]
    cc = [0, 0, radius, -radius]
    val = [1.0] * 4

while r > c + 1:
    c += 1
    dceil = sqrt(radius * radius - c * c)
    dceil = ceil(dceil) - dceil
    if dceil < dceil_prev:
        r -= 1
    
    # Outer pixel (radius r)
    # 8-way symmetry collapses to 4 points if r == c
    if r > c:
        rr.extend([ r,  r,  c,  c, -r, -r, -c, -c])
        cc.extend([ c, -c,  r, -r,  c, -c,  r, -r])
        val.extend([1.0 - dceil] * 8)
    else:
        rr.extend([ r,  r, -r, -r])
        cc.extend([ c, -c,  c, -c])
        val.extend([1.0 - dceil] * 4)

    # Inner pixel (radius r-1)
    # 8-way symmetry collapses to 4 points if r-1 == c
    if r - 1 > c:
        rr.extend([ r - 1,  r - 1,  c,  c, -(r - 1), -(r - 1), -c, -c])
        cc.extend([ c, -c,  r - 1, -(r - 1),  c, -c,  r - 1, -(r - 1)])
        val.extend([dceil] * 8)
    elif r - 1 == c:
        rr.extend([ c,  c, -c, -c])
        cc.extend([ c, -c,  c, -c])
        val.extend([dceil] * 4)
        
    dceil_prev = dceil
```

src/skimage/draw/_draw.pyx
```diff
@@ -249,31 +220,21 @@ def _polygon(r, c, shape):
...
+    """Generate circle perimeter coordinates around point (0, 0).
 
     Parameters
     ----------
-    r_o, c_o : int
-        Centre coordinate of circle.
     radius : int
-        Radius of circle.
+        Radius of the circle.
     method : {'bresenham', 'andres'}
```
stefanv, line 230: Should we switch to andres as default in 2.0?

src/skimage/draw/_draw.pyx
lagru: Noted, I'll attempt a solution making heavy use of if–else like you suggest. That's likely the one with the most performance, but probably not the easiest to follow/maintain.

matthew-brett:

Read on for the rationale, and more detailed instructions.

<details>

## Rationale

</details>

PR #4150, opened 2019-09-11
Add mask parameter to find_contours

### LINKED ISSUE #3006: Provide a nodata parameter to measure.find_contours
## Provide a nodata parameter to measure.find_contours
At the moment, contours found by `skimage.measure.find_contours` which intersect the array edge are left open and others are closed ([docs](http://scikit-image.org/docs/stable/api/skimage.measure.html#skimage.measure.find_contours)).

Where the data of interest is masked but the array can't be cropped to this data (e.g. the area of interest is rotated or an unusual shape) and surrounding 'pixels' are set to NaN, contours are closed where they meet the NaNs. There's an [illustration in this question](https://gis.stackexchange.com/questions/278516/how-could-i-avoid-extracting-incorrectly-closed-contours-from-a-raster-with-angl) on Stack Exchange.

Could a `nodata` parameter (e.g. `nodata=np.NaN` or `nodata=-9999`) be provided to `measure.find_contours` which treats the `nodata` value in the same way as the edge of the array, returning an open contour when it leaves the data of interest and meets a `nodata` pixel?

acarapetis:
## Description

Resolves #3006: Adds a `nodata` parameter to `measure.find_contours` which draws no contours in the region where the input array is equal to `nodata`.

`find_contours` currently misbehaves when handling input arrays containing `NaN` values: it attempts to linearly interpolate these values and ends up returning contours containing `NaN` coordinates. This PR allows avoiding this by passing `nodata=numpy.nan`. For now I've left the previous behaviour as the default (`nodata=None`) for backwards compatibility, in case anyone was relying on it.

The implementation is very simple - no lines are added in any squares with at least one `NaN` corner. This could be cleverer - for example, the square `[[1,NaN],[0,1]]` should probably result in the same segment as `[[1,1],[0,1]]` - but I didn't have time to think through all the possible combinations.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -118,6 +118,8 @@ def find_contours(array, level,
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
     if mask is not None:
+        if not np.can_cast(mask.dtype, bool, casting='safe'):
+            raise TypeError('Parameter "mask" must be a binary array.')
         mask = np.asarray(mask, dtype=np.uint8)
         if mask.shape != array.shape:
             raise ValueError('Parameters "array" and "mask" must have same shape.')
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -120,7 +120,7 @@ def find_contours(array, level,
     if mask is not None:
         if not np.can_cast(mask.dtype, bool, casting='safe'):
             raise TypeError('Parameter "mask" must be a binary array.')
-        mask = np.asarray(mask, dtype=np.uint8)
+        mask = mask.view(np.uint8)
         if mask.shape != array.shape:
             raise ValueError('Parameters "array" and "mask"'
                              ' must have same shape.')
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -117,9 +117,7 @@ def find_contours(array, level,
     if array.ndim != 2:
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
-    if mask is None:
-        mask = np.ones_like(array, dtype=np.uint8)
-    else:
+    if mask is not None:
         mask = np.asarray(mask, dtype=np.uint8)
         if mask.shape != array.shape:
             raise ValueError('Parameters "array" and "mask" must have same shape.')
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -122,7 +122,8 @@ def find_contours(array, level,
             raise TypeError('Parameter "mask" must be a binary array.')
         mask = np.asarray(mask, dtype=np.uint8)
         if mask.shape != array.shape:
-            raise ValueError('Parameters "array" and "mask" must have same shape.')
+            raise ValueError('Parameters "array" and "mask"'
+                             ' must have same shape.')
     if (fully_connected not in _param_options or
        positive_orientation not in _param_options):
         raise ValueError('Parameters "fully_connected" and'
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -118,9 +118,9 @@ def find_contours(array, level,
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
     if mask is None:
-        mask = np.ones_like(array, dtype=bool)
+        mask = np.ones_like(array, dtype=np.uint8)
     else:
-        mask = np.asarray(mask, dtype=bool)
+        mask = np.asarray(mask, dtype=np.uint8)
         if mask.shape != array.shape:
             raise ValueError('Parameters "array" and "mask" must have same shape.')
     if (fully_connected not in _param_options or
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -8,7 +8,7 @@ _param_options = ('high', 'low')
 
 def find_contours(array, level,
                   fully_connected='low', positive_orientation='low',
-                  nodata=None):
+                  mask=None):
     """Find iso-valued contours in a 2D array for a given level value.
 
     Uses the "marching squares" method to compute a the iso-valued contours of
@@ -31,11 +31,10 @@ def find_contours(array, level,
          contours will wind counter- clockwise around elements below the
          iso-value. Alternately, this means that low-valued elements are always
          on the left of the contour. (See below for details.)
-    nodata : float
-        Value to treat as missing data. No contours will be drawn where *array*
-        has values equal to *nodata* (or where values are ``NaN`` if nodata is
-        ``NaN``).  Default value is ``None``, which disables this behaviour
-        entirely for backwards-compatibility.
+    mask : 2D ndarray of bool, or None
+        A boolean mask, True where we want to draw contours.
+        Note that NaN values are always excluded from the considered region (
+        i.e. mask=False is forced where where array is NaN).
 
     Returns
     -------
@@ -64,9 +63,10 @@ def find_contours(array, level,
     with the 'fully_connected' parameter.
 
     Output contours are not guaranteed to be closed: contours which intersect
-    the array edge or a region of missing data will be left open. All other
-    contours will be closed. (The closed-ness of a contours can be tested by
-    checking whether the beginning point is the same as the end point.)
+    the array edge or a masked-off region (either where mask is False or where
+    array is NaN) will be left open. All other contours will be closed. (The
+    closed-ness of a contours can be tested by checking whether the beginning
+    point is the same as the end point.)
 
     Contours are oriented. By default, array values lower than the contour
     value are to the left of the contour and values greater than the contour
@@ -117,15 +117,19 @@ def find_contours(array, level,
     if array.ndim != 2:
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
-    if nodata is not None:
-        nodata = np.double(nodata)
+    if mask is None:
+        mask = np.ones_like(array, dtype=bool)
+    else:
+        mask = np.asarray(mask, dtype=bool)
+        if mask.shape != array.shape:
+            raise ValueError('Parameters "array" and "mask" must have same shape.')
     if (fully_connected not in _param_options or
        positive_orientation not in _param_options):
         raise ValueError('Parameters "fully_connected" and'
         ' "positive_orientation" must be either "high" or "low".')
     point_list = _find_contours_cy.iterate_and_store(array, level,
                                                      fully_connected == 'high',
-                                                     nodata=nodata)
+                                                     mask=mask)
     contours = _assemble_contours(_take_2(point_list))
     if positive_orientation == 'high':
         contours = [c[::-1] for c in contours]
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -117,7 +117,8 @@ def find_contours(array, level,
     if array.ndim != 2:
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
-    if nodata is not None: nodata = float(nodata)
+    if nodata is not None:
+        nodata = float(nodata)
     if (fully_connected not in _param_options or
        positive_orientation not in _param_options):
         raise ValueError('Parameters "fully_connected" and'
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -118,7 +118,7 @@ def find_contours(array, level,
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
     if nodata is not None:
-        nodata = float(nodata)
+        nodata = np.double(nodata)
     if (fully_connected not in _param_options or
        positive_orientation not in _param_options):
         raise ValueError('Parameters "fully_connected" and'
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -7,7 +7,8 @@ _param_options = ('high', 'low')
 
 
 def find_contours(array, level,
-                  fully_connected='low', positive_orientation='low'):
+                  fully_connected='low', positive_orientation='low',
+                  nodata=None):
     """Find iso-valued contours in a 2D array for a given level value.
 
     Uses the "marching squares" method to compute a the iso-valued contours of
@@ -30,6 +31,11 @@ def find_contours(array, level,
          contours will wind counter- clockwise around elements below the
          iso-value. Alternately, this means that low-valued elements are always
          on the left of the contour. (See below for details.)
+    nodata : float
+        Value to treat as missing data. No contours will be drawn where *array*
+        has values equal to *nodata* (or where values are ``NaN`` if nodata is
+        ``NaN``).  Default value is ``None``, which disables this behaviour
+        entirely for backwards-compatibility.
 
     Returns
     -------
@@ -58,9 +64,9 @@ def find_contours(array, level,
     with the 'fully_connected' parameter.
 
     Output contours are not guaranteed to be closed: contours which intersect
-    the array edge will be left open. All other contours will be closed. (The
-    closed-ness of a contours can be tested by checking whether the beginning
-    point is the same as the end point.)
+    the array edge or a region of missing data will be left open. All other
+    contours will be closed. (The closed-ness of a contours can be tested by
+    checking whether the beginning point is the same as the end point.)
 
     Contours are oriented. By default, array values lower than the contour
     value are to the left of the contour and values greater than the contour
@@ -111,12 +117,14 @@ def find_contours(array, level,
     if array.ndim != 2:
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
+    if nodata is not None: nodata = float(nodata)
     if (fully_connected not in _param_options or
        positive_orientation not in _param_options):
         raise ValueError('Parameters "fully_connected" and'
         ' "positive_orientation" must be either "high" or "low".')
     point_list = _find_contours_cy.iterate_and_store(array, level,
-                                                     fully_connected == 'high')
+                                                     fully_connected == 'high',
+                                                     nodata=nodata)
     contours = _assemble_contours(_take_2(point_list))
     if positive_orientation == 'high':
         contours = [c[::-1] for c in contours]
--- a/skimage/measure/_find_contours_cy.pyx
+++ b/skimage/measure/_find_contours_cy.pyx
@@ -3,9 +3,8 @@
 #cython: nonecheck=False
 #cython: wraparound=False
 import numpy as np
-cimport numpy as np
+cimport numpy as cnp
 
-ctypedef np.uint8_t DTYPE_BOOL_t
 cdef extern from "numpy/npy_math.h":
     bint npy_isnan(double x)
     double NAN "NPY_NAN"
@@ -19,7 +18,7 @@ cdef inline double _get_fraction(double from_value, double to_value,
 
 def iterate_and_store(double[:, :] array,
                       double level, Py_ssize_t vertex_connect_high,
-                      np.ndarray[DTYPE_BOOL_t, cast=True, ndim=2] mask):
+                      cnp.uint8_t[:, :] mask):
     """Iterate across the given array in a marching-squares fashion,
     looking for segments that cross 'level'. If such a segment is
     found, its coordinates are added to a growing list of segments,
--- a/skimage/measure/_find_contours_cy.pyx
+++ b/skimage/measure/_find_contours_cy.pyx
@@ -7,7 +7,6 @@ cimport numpy as cnp
 
 cdef extern from "numpy/npy_math.h":
     bint npy_isnan(double x)
-    double NAN "NPY_NAN"
 
 cdef inline double _get_fraction(double from_value, double to_value,
                                  double level):
--- a/skimage/measure/_find_contours_cy.pyx
+++ b/skimage/measure/_find_contours_cy.pyx
@@ -28,10 +28,9 @@ def iterate_and_store(double[:, :] array,
     if array.shape[0] < 2 or array.shape[1] < 2:
         raise ValueError("Input array must be at least 2x2.")
 
-    cdef bint _nodata_enabled
+    cdef bint _nodata_enabled = nodata is not None
     cdef double _nd = 0
 
-    _nodata_enabled = nodata is not None
     if _nodata_enabled:
         _nd = <double>nodata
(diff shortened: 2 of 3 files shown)


skimage/measure/_find_contours.py
```diff
@@ -111,12 +117,15 @@ def find_contours(array, level,
     if array.ndim != 2:
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
+    if nodata is not None:
+        nodata = float(nodata)
```
lagru:
```suggestion
        nodata = np.double(nodata)
```
We should probably use exactly the same dtype as the `array` here.

skimage/measure/_find_contours_cy.pyx
```diff
@@ -25,6 +28,13 @@ def iterate_and_store(double[:, :] array,
     if array.shape[0] < 2 or array.shape[1] < 2:
         raise ValueError("Input array must be at least 2x2.")
 
+    cdef bint _nodata_enabled
+    cdef double _nd = 0
+
+    _nodata_enabled = nodata is not None
```
lagru: Can you merge this line with the declaration above? Should be a little easier to read.

skimage/measure/tests/test_find_contours.py
```diff
@@ -9,6 +9,9 @@
 a[1:-1, 1] = 0
 a[1, 1:-1] = 0
 
+b = np.copy(a)
+b[7:, :] = np.nan
```
lagru: Maybe move this into the `test_nodata` function if it is only used there?

skimage/measure/tests/test_find_contours.py
```diff
@@ -45,6 +48,36 @@ def test_binary():
...
+           [1. ,  0.5],
+           [2. ,  0.5],
+           [3. ,  0.5],
+           [4. ,  0.5],
+           [5. ,  0.5],
+           [6. ,  0.5]]
+    contours = find_contours(b, 0.5, positive_orientation='high',
+                             nodata=np.nan)
+    assert len(contours) == 1
+    assert_array_equal(contours[0][::-1], ref)
```
lagru: Could you perhaps change the order of `ref` above? Than you can drop the inversion here.

skimage/measure/_find_contours_cy.pyx
```diff
@@ -93,6 +103,13 @@ def iterate_and_store(double[:, :] array,
             coords[0] += 1
             coords[1] = 0
 
+        if _nodata_enabled:
+            if npy_isnan(_nd):
+                if (npy_isnan(ul) or npy_isnan(ur) or npy_isnan(ll) or npy_isnan(lr)):
```
lagru: Why the enclosing brackets here?

skimage/measure/_find_contours.py
```diff
@@ -30,6 +31,11 @@ def find_contours(array, level,
          contours will wind counter- clockwise around elements below the
          iso-value. Alternately, this means that low-valued elements are always
          on the left of the contour. (See below for details.)
+    nodata : float
```
lagru: I don't really find this name descriptive. What do you think about `ignore_value` or `ignore_as_missing`?

skimage/measure/tests/test_find_contours.py
acarapetis: I was following the same pattern as the existing tests in `test_find_contour.py` here: all the test data is constructed at the top of the file. (For example, `a`, despite only being used in `test_binary`.)

skimage/measure/tests/test_find_contours.py
acarapetis: As in the previous comment, this is done in the same way as the existing tests; but I agree with you that this one in particular is unnecessary.

I'd be happy to make your suggested tweak (for both this and the previous item), but I think that if I do, I should also make the same changes to the existing test `test_binary` - what do you think?

skimage/measure/_find_contours_cy.pyx
acarapetis: I originally wrote this whole block as one big boolean expression before thinking better of it, so the parens are a hangover from that. I'll get rid of them.

skimage/measure/_find_contours.py
acarapetis: I followed the suggestion of `nodata` from #3006, but I'm not particularly attached to it - something like  `ignore_as_missing` or `missing_data_value` would indeed be more descriptive.

skimage/measure/_find_contours_cy.pyx
acarapetis: Fixed by 6bd1ae2.

skimage/measure/_find_contours_cy.pyx
acarapetis: Good catch. Fixed in 6bd1ae2.

acarapetis:
Hi @lagru, thanks for your review! I've left some replies to your comments.

Regarding `_iterate_and_store`, it's an implementation of the algorithm described [here](https://en.wikipedia.org/wiki/Marching_squares#Isoline), which just iterates over the input data and builds a list of line segments. These are then assembled into contour lines by the pure python code in `find_contours.py`.

emmanuelle:
For the consistency of the API, couldn't we have instead a `mask` parameter for which only the pixels at which `mask=True` are considered? This is a pattern we have in a lot of scikit-image functions, and it's a bit more generic than excluding only one value. I'm not sure it would do exactly what you are proposing, though. What do you think about this API ?

acarapetis:
@emmanuelle: That sounds sensible to me - the call to exclude a single value would then just look like `find_contours(array, level, mask=(array != nodata))`, which is very readable.

While it would change the behaviour of previously-written code, I think it's also worth always masking out NaN values even when the `mask` parameter is not provided, since the current behaviour is just broken:

```python
>>> from skimage.measure import find_contours
>>> import numpy as np
>>> data = np.array([[1,0,0],[1,1,np.nan],[1,1,1]])
>>> find_contours(data, 0.5)
[array([[0. , 0.5],
        [0.5, 1. ],
        [1. , nan]]), array([[ 1., nan],
        [nan,  2.]])]
```
What do you think?

skimage/measure/_find_contours_cy.pyx
```diff
@@ -13,14 +18,17 @@ cdef inline double _get_fraction(double from_value, double to_value,
 
 
 def iterate_and_store(double[:, :] array,
-                      double level, Py_ssize_t vertex_connect_high):
+                      double level, Py_ssize_t vertex_connect_high,
+                      np.ndarray[DTYPE_BOOL_t, cast=True, ndim=2] mask):
```
lagru: This is the first time I encountered this syntax. Is there a specific reason/advantage over using a memoryview like `np.uint8_t[:, :] mask` here?

skimage/measure/_find_contours_cy.pyx
```diff
@@ -3,7 +3,12 @@
 #cython: nonecheck=False
 #cython: wraparound=False
 import numpy as np
+cimport numpy as np
```
lagru: My personal preference is to differentiate the cimport by using `cimport numpy as cnp`. But using `np` is frequently used as well so you can leave it if you want.

skimage/measure/_find_contours_cy.pyx
```diff
@@ -93,13 +103,65 @@ def iterate_and_store(double[:, :] array,
...
+            # If a square has 2 or more missing values, we cannot correctly
+            # infer the presence of any contour line segments within it;
+            # so just move to the next square.
+            continue
+        elif nan_count == 1:
+            # There is (up to symmetry) one square arrangement containing a
+            # missing value in which we can unambiguously draw an isoline
+            # segment:
+            # The arrangement +-
+            #                 -x (where x denotes a missing value)
```
lagru: Is this true though? The way I understand this, the missing value above might either be + or - corresponding to the cases 1 and 9. So we actually don't know what the correct iso line might be.

Also could you elaborate what you mean with "after symmetry"?

skimage/measure/_find_contours_cy.pyx
```diff
@@ -93,13 +103,65 @@ def iterate_and_store(double[:, :] array,
...
+            # should have the same contour line as case 1.
+            # After symmetry, there are 8 cases, which are enumerated here.
```
lagru: Is this true though? The way I understand this, the missing value above might either be + or - corresponding to the cases 1 and 9. So we actually don't know what the correct iso line might be.

Also could you elaborate what you mean with "after symmetry"?

skimage/measure/_find_contours_cy.pyx
lagru: The [official docs](https://cython.readthedocs.io/en/latest/src/tutorial/numpy.html) seem to recommend using memoryviews for their performance advantage.

acarapetis: I'm new to Cython, so I appreciate the help here! `np.uint8_t[:, :]` was what I tried first, but I ran into the error `ValueError: Does not understand character buffer dtype format string`, and this approach was what I found googling for help.

Since iterate_and_store isn't a public API anyway, I guess the right solution is to make it take a uint8 memoryview as you suggest, and cast mask to `dtype=np.uint8` in `find_contours`?

skimage/measure/_find_contours_cy.pyx
acarapetis: By symmetry, I mean the fact that cases 1,2,4,8 are just rotations of each other, and furthermore these are the same as 14,13,11,7 with + and - swapped (you might think of this as reflection in the Z coordinate). Whatever we decide is the correct treatment for one of these cases can be transferred to the others by applying the correct symmetry.

My thinking was that the outputs in case 1 (top-left segment) and case 9 (top-left and bottom-right segments) both include the top-left line segment, so we could unambiguously output this segment; but I see now that I might have been too hasty: 9 is one of the ambiguous "saddle points", so it could instead output top-right and bottom-left.

I think there are probably still some cases where this really isn't ambiguous if we take `vertex_connect_high` in to account; but I also might just be trying to get too clever here - since contours are going to be cut off by the mask edge anyway, losing a single pixel probably isn't a big deal.

skimage/measure/_find_contours_cy.pyx
```diff
@@ -93,6 +100,14 @@ def iterate_and_store(double[:, :] array,
             coords[0] += 1
             coords[1] = 0
 
+        # Skip this square if any of the four input values are masked out.
+        if not (mask[r0, c0] and mask[r0, c1] and
+                mask[r1, c0] and mask[r1, c1]):
+            continue
+
+        # Skip this square if any of the four input values are NaN.
+        if npy_isnan(ul) or npy_isnan(ur) or npy_isnan(ll) or npy_isnan(lr):
```
acarapetis: Instead of checking for nans here, we could make a copy of `mask` and do `mask &= ~np.isnan(array)` before we start, as suggested by @jni. I imagine this would be slightly faster (since we'd read each array value for nan checks once instead of four times) but use slightly more memory to store the copy. I have no idea whether this is likely to matter much - I'm new around here, so happy to go with whatever you all think is best.

hmaarrfk: The best way to figure out which is faster is to benchmark

https://scikit-image.org/docs/dev/contribute.html#benchmarks

hmaarrfk:
I don't think scikit-image uses many vectorized instructructions, but I want to caution against making things `nansafe`. 
`nan`s can typically limited to specific regions in your code, often after a division, which in many cases is better organized as a multiplication.


The reason is simply performance. Not sure the true cause of the slowdown, but `np.nansum` is more than 4x slower than the normal summation.
<details>

```
import numpy as np
a = np.ones(int(1E6))
np.sum(a)
Out[14]: 1000000.0
np.nansum(a)
Out[15]: 1000000.0

%timeit np.nansum(a)
1.48 ms ± 17 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit np.sum(a)
313 µs ± 12 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
</details>
Not sure where this is coming from, but it is a real slowdown.

skimage/measure/_find_contours_cy.pyx
hmaarrfk: I think you'll also find that what is faster for the small arrays, will not be faster for large arrays.

lagru: To get a quick ballpark estimate you can also use [ipython's %timeit command](https://ipython.org/ipython-doc/3/interactive/magics.html#magic-timeit) (in case you didn't know already).

lagru:
@hmaarrfk Your impulse to benchmark this before deciding is probably the right one. In case there is a real performance disadvantage I'd be fine with a solution that ignores nans as long as the docstring mentions prominently that nans _need_ to be masked to get sensible output.

skimage/measure/_find_contours_cy.pyx
```diff
@@ -93,6 +99,14 @@ def iterate_and_store(double[:, :] array,
             coords[0] += 1
             coords[1] = 0
 
+        # Skip this square if any of the four input values are masked out.
+        if not (mask[r0, c0] and mask[r0, c1] and
+                mask[r1, c0] and mask[r1, c1]):
```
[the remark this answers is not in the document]
acarapetis: Good point, I'll make this change.

acarapetis:
Ran a quick and dirty benchmark to see the performance impact of the nan checks when dealing with nan-free data:
```python
# Once, to generate fixed test data:
for s in [20,50,100,500,1000,2000,3000]:
    np.save(f'{s}.npy', np.random.random((s,s)))

# Then once for each version of the code:
import numpy as np
data = [(s, np.load(f'{s}.npy')) for s in [20,50,100,500,1000,2000,3000]]
from skimage.measure import find_contours
for s, arr in data:
    print(f'{s}x{s} ', end='')
    %timeit find_contours(np.random.random((s,s)), 0.5)
```

Not super controlled conditions, but on my hardware at least it looks like roughly a 2% hit:

| size | before | after | difference |
| --- | --- | --- | --- |
| 20x20 | 714 µs ± 1.87 µs | 727 µs ± 6.15 µs | ~2% slower |
| 50x50 | 4.47 ms ± 23.2 µs | 4.52 ms ± 10.5 µs | ~1% slower |
| 100x100 | 19.2 ms ± 244 µs | 19.2 ms ± 299 µs | ~no change |
| 500x500 | 539 ms ± 3.91 ms | 549 ms ± 10.5 ms | ~2% slower |
| 1000x1000 | 2.23 s ± 15.2 ms | 2.27 s ± 27.3 ms | ~2% slower |
| 2000x2000 | 9.38 s ± 194 ms | 9.35 s ± 51.8 ms | ~0.3% faster |
| 3000x3000 | 21.7 s ± 328 ms | 22.2 s ± 965 ms | ~2% slower |

lagru:
@acarapetis Thank's for the benchmarks. I get similar results. Looks to me like the performance impact of checking for nans is negligible and in this case well worth it. I'm in favor of keeping it.

skimage/measure/_find_contours_cy.pyx
```diff
@@ -13,14 +16,17 @@ cdef inline double _get_fraction(double from_value, double to_value,
 
 
 def iterate_and_store(double[:, :] array,
-                      double level, Py_ssize_t vertex_connect_high):
+                      double level, Py_ssize_t vertex_connect_high,
+                      cnp.uint8_t[:, :] mask):
```
lagru: I'm surprised that we don't get a `ValueError: Does not understand character buffer dtype format string ('?')` here when passing in a mask of dtype('bool') like you mentioned previously in https://github.com/scikit-image/scikit-image/pull/4150#discussion_r324168938. I still get this error when using the `cnp.uint8_t[::1]` notation in another PR. Perhaps that's why...

See also https://github.com/cython/cython/pull/2676#issuecomment-529916887.

skimage/measure/_find_contours.py
```diff
@@ -111,12 +117,17 @@ def find_contours(array, level,
     if array.ndim != 2:
         raise ValueError('Only 2D arrays are supported.')
     level = float(level)
+    if mask is not None:
+        mask = np.asarray(mask, dtype=np.uint8)
```
lagru: Ah, silly me, this is the reason why we don't get the buffer error in https://github.com/scikit-image/scikit-image/pull/4150/files#r326102540.

Using asarray here also implies that passing in a mask of dtype('int64') works as well. Do we want to be this forgiving with the input?

skimage/measure/_find_contours.py
```diff
@@ -111,12 +117,17 @@ def find_contours(array, level,
...
+        if mask.shape != array.shape:
+            raise ValueError('Parameters "array" and "mask" must have same shape.')
```
lagru: Could you add a test that tests this case?

skimage/measure/_find_contours.py
acarapetis: I think something definitely needs to change here - the current downcast to uint8 doesn't correspond to python truthiness (for example `np.asarray([256], dtype=np.uint8)` results in `[0]`), which is certainly not intuitive to me.

Two suggestions:
- A forgiving API that accepts any mask dtype and treats non-zero as True, zero as False. I think we could implement this cleanly by making the cython function take the mask as `cython.numeric[:, :]`, and casting to uint8 in the python wrapper only if the input dtype is bool.
- Stick with the current approach and check that mask has dtype bool or uint8.

Thoughts?

skimage/measure/_find_contours.py
acarapetis: Done! See 362dfb982

skimage/measure/_find_contours.py
lagru: I think `np.asarray(mask, dtype=bool).view(np.uint8)` would work as intended (treating any value != 0 as True).

However after thinking about this a bit more I'd say we don't do users any favors by being more forgiving than the docstring indicates. I think I read something about "In the face of ambiguity..." somewhere. :smile:  So maybe something like this is better:

```python
if mask is not None and not np.can_cast(mask.dtype, bool, casting="safe"):
    raise TypeError("mask is not a binary array")
```

This might be a little annoying if the user has a `mask` of 1s and 0s (`dtype('uint8')`) but then I'd recommend to just call `.view(bool)` before passing it in.

lagru: @acarapetis Thanks for addressing this. You can even remove the line `mask = np.asarray(mask, dtype=np.uint8)` (and combine the nested if-statements) here and instead pass `mask.view(np.uint8)` to `iterate_and_store` later on. That way we can be sure that no unnecessary copy is performed on `mask`.

emmanuelle:
This looks good to me. @lagru I'll let you merge if you're happy with the PR now. 

Regarding the performance of `find_contours`, I ran the line profiler on it and the majority of the time is actually spent in the Python function `_assemble_contours` (here https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/_find_contours.py#L120). Efforts to accelerate `find_contours` should therefore concentrate on `_assemble_contours`. Any ideas about the possibility to parallelize this function @zpincus @stefanv ?

zpincus:
Unfortunately I don't see a whole lot of low-hanging fruit for optimizing `_assemble_contours`. The main problem is that it has to deal with potentially many line segments, and can only do that by iterating through them in python. Not a lot of opportunity for data parallelism. Cythonizing might help a little, and perhaps a slightly better data structure than a dict could be used for looking up the heads and tails by spatial position (KD tree? Quadtree?)...

acarapetis:
> the majority of the time is actually spent in the Python function _assemble_contours
I suspected that might be the case.

Regarding parallelism, I feel like splitting things up spatially should be quite effective; for example:

1. Split the domain into e.g. four quadrants
2. Treating each quadrant as a sub-image, compute and assemble the contours in this quadrant.
3. Join the threads and re-run contour assembly on the combined dataset.

The final step should be quite fast, since (in the average case) there will just be a few long-ish contour lines to assemble, not a soup of line segments.

skimage/measure/_find_contours.py
lagru: > You can even remove the line `mask = np.asarray(mask, dtype=np.uint8)` (and combine the nested if-statements) here and instead pass `mask.view(np.uint8)` to `iterate_and_store` later on.

@acarapetis sorry to prod you on this but could you address this part? Calling `np.asarray(mask, dtype=np.uint8)` on a boolean array will make an unnecessary copy. 

```python
>>> a = np.array([False, False], dtype=bool)
>>> b = np.asarray(a, dtype=np.uint8)
>>> c = a.view(np.uint8)
>>> a[0] = True
>>> a, b, c
(array([ True, False]), array([0, 0], dtype=uint8))
```

We can avoid that by using `.view(np.uint8)` instead. 

I'll merge once this is resolved.

lagru:
Oh and there's a small PEP issue:

> In the file skimage/measure/_find_contours.py:
[Line 125:80](https://github.com/scikit-image/scikit-image/blob/52af77c49be2753ce81187d3cfb34ecdcbf48035/skimage/measure/_find_contours.py#L125): E501 line too long (83 > 79 characters)

skimage/measure/_find_contours.py
acarapetis: I've made this change to mask. Is the same problem not true of `array`, for example if the user passes single-precision float? 

Also, regarding "combining the nested if statements", this would require something like `iterate_and_store(..., mask=mask.view(np.uint8) if mask is not None else None)`, so I just stuck with the nested if.

lagru: > Is the same problem not true of array, for example if the user passes single-precision float?

Assuming I understand you correctly I'd argue that this is a special case due to a "quirk" in NumPy. Each element in a boolean array actually takes up one full byte (same as uint8). Because we made sure that mask is a boolean array we can safely reinterpret that by creating a view showing us `True` as `1` and `False` as `0` which is just how those values are actually stored in memory.

`array` is a different case because `view` behaves differently for numeric dtypes. Creating views between dtypes with a different itemsize only works by changing the number of elements in the array which we clearly don't want here. Furthermore interpreting say an int64 as a float64 will get interesting but in this case not really useful results. E.g. `np.array([3], "i8").view("f8")` will return `array([1.5e-323])`.

> Also, regarding "combining the nested if statements", this would require something like iterate_and_store(..., mask=mask.view(np.uint8) if mask is not None else None), so I just stuck with the nested if.

You're right. I forgot that we need to accommodate `mask` being `None`.

acarapetis:
Happy to contribute! Thanks @lagru and everyone else for the feedback.

(and a last note - after looking up what `np.view` actually does, you're totally right.)

Issue #4152, opened 2019-09-11
Do not use reserved Python keywords as variable names


soupault:
## Description
This requires searching throught the complete codebase and renaming/deprecating old function/variable names.

The most commonly misused are `filter` and `slice`.

rfezzani:
I wrote the PR #4158 to demonstrate how this issue can be managed using a decorator. Let me know if this kind of solution is acceptable or not ;-)

rfezzani:
I also found `next` used in `skimage/morphology/greyreconstruct.py` :smile:

rfezzani:
Do we consider that `array` is an acceptable argument name? I found an occurrence in `graph.mcp.route_through_array` for example.

jni:
I would say yes, array is not reserved.

soupault:
I agree with Juan. Although, it is probably not the best choice for the
variable name, since it is not descriptive.

Issue #4154, opened 2019-09-12
Define standard common argument's names


rfezzani:
## Description

It is important for a library to be consistent in arguments naming. The issue #4003 points out such problem.
I noticed for example that the argument for the _"maximum number of iterations"_ has at least 5 names over `skimage` functions:
- `max_iter` in `slic`,
- `max_iterations` in `active_contour`,
- `iterations` in `richardson_lucy`,
- `n_iter_max` in `denoise_tv_chambolle`,
- `niter` in `optical_flow_tvl1`.

What about defining standards for such common argument names and fixing the code base in the same spirit as #4152?

rfezzani:
I didn't knew `wabisabi`, thank you @hmaarrfk for pointing it. You also are right, the decorator defined in this [gist](https://gist.github.com/rfezzani/002181c8667ec4c671421a4d938167eb) can be used for this issue. The PR #4158 is an example of function's signature modification using such a decorator.
@jni, I also vote for `max_num_iter` and `num_iter` ;-)

rfezzani:
For [ref](https://github.com/scikit-image/scikit-image/pull/4209#discussion_r355546849).

rfezzani:
Concerning the `selem`  argument name  consistency identified in #4209, I found:

- `selem` in `skimage.filters.rank` module's functions,
- `selem` in `skimage.morphology` module's functions,
- `structure` in `skimage.morphology._util._offsets_to_raveled_neighbors`,
- `footprint` in `skimage.feature.peak.peak_local_max` and `skimage.feature.peak._get_peak_mask`

jni:
Thanks for the investigation!

Since we have all those uses, I think it makes sense to find the one we like most and use it, since something will have to be deprecated whatever we choose. (Even though of course some deprecations are more panful than others.) Again, imho selem is really the worst of them.

anntzer:
fwiw you may be interested in using https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/cbook/deprecation.py#L258 (a renaming decorator that handles the deprecation `@rename_parameter(version_number, old_name, new_name)`) rather than manually going through all of them.
It uses some other matplotlib internal deprecation machinery but should be easy enough to adapt for skimage...

rfezzani:
Thank you @anntzer, I didn't knew this one! I already proposed #4158 a decorator in the same spirit ;-)

rfezzani:
[...]

Issue #4003, opened 2019-07-10
API inconsistancy between remove_small_holes and remove_small_objects?


sciunto:
## Description

In 0.16, in remove_small_holes, min_size has been replaced definitely by area_threshold. [See here](https://scikit-image.org/docs/stable/api/skimage.morphology.html#remove-small-holes).

But for remove_small_objects, it is still min_size. Why that? The first one call the other internally...

jni:
@sciunto personally I disagreed with the change to `area_threshold`, but didn't catch that PR in time. Either way, it's likely that `remove_small_objects` was simply overlooked. Thanks for flagging this. We should have a long discussion about the best name for these parameters.

mkcor:
I think that `min_size` is/was fine as long as documentation is well written.

My second favourite option would be https://github.com/scikit-image/scikit-image/pull/4761#issuecomment-635850530 by @jni because it reads so well.

lagru:
Having dug through the past issues and discussions on this, it seems to me that each discussed option will be a trade-off between consistency and matching the context of each function. 

Since `remove_small_objects` and `remove_small_holes` are so similar, I definitely give consistency more weight. Ignoring past deprecations, `min_size` seems to be the least controversial option. The main argument against it was that we'd have to revert a previous deprecation. 

That's not ideal, but addressing previous mistakes is precisely what `skimage2` is about. So, I propose to deprecate `area_threshold` in `remove_small_holes` back to `min_size` and only finalize the deprecation in `2.0.0`.

I think we can keep using `area_threshold` in `area_opening` and `area_closing`, since `area` is a prominent term and I anticipate less user confusion.

I'll make a PR to get this going again. Please, feel very welcome to suggest an alternate concrete action to address this issue. :wink:

PR #4158, opened 2019-09-13
Deprecate filter argument in iradon due to clash with python keyword

rfezzani:
## Description

This PR is about deprecating the reserved Python keyword `filter` used as argument for `iradon` function. It is related to the issue #4152.

`filter` is replaced by `filter_name` in the definition and documentation of `iradon` and the decorator `deprecate_arg` is defined in `skimage._shared.utils` and used to raise a `DeprecationWarning` if the old argument is explicitly used by the user and replace it by the new argument name.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/_shared/utils.py
```diff
@@ -20,6 +20,44 @@ class skimage_deprecation(Warning):
...
+        @functools.wraps(func)
+        def fixed_func(*args, **kwargs):
+            for old_arg, new_arg in self.arg_mapping.items():
+                if old_arg in kwargs:
+                    #  warn that the function interface has changed:
+                    warnings.warn(self.warning_msg %
+                                  (old_arg, func.__name__, new_arg),
+                                  DeprecationWarning)
+                    # Substitute new_arg to old_arg
+                    kwargs[new_arg] = kwargs.pop(old_arg)
```
hmaarrfk: I'm not exactly sure it is safe to change what you are iterating over. I've had a few issues there in other circumstances

skimage/transform/radon_transform.py
```diff
@@ -171,8 +172,9 @@ def _get_fourier_filter(size, filter_name):
     return fourier_filter[:, np.newaxis]
 
 
+@deprecate_arg({'filter': 'filter_name'})
 def iradon(radon_image, theta=None, output_size=None,
-           filter="ramp", interpolation="linear", circle=True):
+           filter_name="ramp", interpolation="linear", circle=True):
```
hmaarrfk: Generally you need to add warnings so the old behavior can be taken out.

skimage/_shared/utils.py
```diff
@@ -20,6 +20,44 @@ class skimage_deprecation(Warning):
...
+
+    def __init__(self, arg_mapping, removed_version=None):
+        self.arg_mapping = arg_mapping
+        self.warning_msg = "'%s' is a deprecated argument name for %s. "
+        if removed_version is not None:
+            self.warning_msg += "It will be removed in version %s" % (
+                removed_version)
+        self.warning_msg += "Use '%s' instead."
+
+    def __call__(self, func):
```
hmaarrfk: I think we would want tests for this.

skimage/_shared/utils.py
rfezzani: The iteration is not on `kwargs` but on the `arg_mapping`, it is safe then.

hmaarrfk: nice one!

skimage/_shared/utils.py
rfezzani: I did not implement tests because I found that the `deprecated` decorator is not tested neither, but I can manage to do it, the way it is made in `wabisabi` ;-)

skimage/transform/radon_transform.py
rfezzani: I am sorry, but don't understand what you mean here :-S. The warning is raised by the decorator.

skimage/_shared/utils.py
```diff
@@ -20,6 +20,44 @@ class skimage_deprecation(Warning):
...
+        @functools.wraps(func)
+        def fixed_func(*args, **kwargs):
+            for old_arg, new_arg in self.arg_mapping.items():
+                if old_arg in kwargs:
+                    #  warn that the function interface has changed:
+                    warnings.warn(self.warning_msg %
+                                  (old_arg, func.__name__, new_arg),
+                                  DeprecationWarning)
```
hmaarrfk: Gah sorry, reading this form my phone is hard. You should set stack level to 2 otherwise the warning will be very confusing

skimage/_shared/utils.py
```diff
@@ -20,6 +20,44 @@ class skimage_deprecation(Warning):
...
+    arg_mapping: dict
+        mapping between the function's old argument names and the new ones.
+    removed_version : str
+        The package version in which the deprecated argument will be removed.
+
+    """
+
+    def __init__(self, arg_mapping, removed_version=None):
+        self.arg_mapping = arg_mapping
+        self.warning_msg = "'%s' is a deprecated argument name for %s. "
```
hmaarrfk: Can you add names for the the %s

skimage/transform/radon_transform.py
hmaarrfk: You are right 0!

skimage/transform/radon_transform.py
```diff
@@ -191,7 +193,7 @@ def iradon(radon_image, theta=None, output_size=None,
         between 0 and 180 (if the shape of `radon_image` is (N, M)).
     output_size : int, optional
         Number of rows and columns in the reconstruction.
-    filter : str, optional
+    filter_name : str, optional
```
hmaarrfk: I think we are trying to add notes, that something has been removed in a specific version. Versionchanged or versionadded

hmaarrfk:
You don't need to pass all tests from wabisabi. Generally, Ive Found that the tests help me find bugs in my own code that might be difficult to spot. You should also consider what you want you to happen when the user specifies the argument with both names

skimage/_shared/utils.py
```diff
@@ -35,21 +35,22 @@ class deprecate_arg:
...
         def fixed_func(*args, **kwargs):
             for old_arg, new_arg in self.arg_mapping.items():
                 if old_arg in kwargs:
                     #  warn that the function interface has changed:
-                    warnings.warn(self.warning_msg %
-                                  (old_arg, func.__name__, new_arg),
-                                  DeprecationWarning)
+                    warnings.warn(self.warning_msg.format(
+                        old_arg=old_arg, func_name=func.__name__,
+                        new_arg=new_arg), DeprecationWarning, stacklevel=2)
```
hmaarrfk: DeprecationWarnings are really easy to silence. please use `FutureWarning`

rfezzani:
I am not sure that we should consider the case where the user specifies the argument with both names: the deprecation mechanism is made to manage old users code. Using both arguments name is a user bug not an `skimage` bug :-)

skimage/_shared/utils.py
rfezzani: Done

skimage/_shared/utils.py
rfezzani: Done

skimage/_shared/utils.py
rfezzani: Are the implemented tests good enough?

skimage/_shared/utils.py
```diff
@@ -20,6 +20,45 @@ class skimage_deprecation(Warning):
...
+
+    def __call__(self, func):
+        @functools.wraps(func)
+        def fixed_func(*args, **kwargs):
+            for old_arg, new_arg in self.arg_mapping.items():
+                if old_arg in kwargs:
+                    #  warn that the function interface has changed:
+                    warnings.warn(self.warning_msg.format(
+                        old_arg=old_arg, func_name=func.__name__,
+                        new_arg=new_arg), FutureWarning, stacklevel=2)
```
hmaarrfk: Is this stack level correct?

rfezzani: Good question! I made this simple test:

```python
from functools import wraps
import warnings


def my_decorator(f):
    @wraps(f)
    def wrapper(*args, **kwargs):
        warnings.warn("Oooooops", stacklevel=2)
        return f(*args, **kwargs)
    return wrapper


@my_decorator
def func():
    print("Here I am")


if __name__ == '__main__':
    func()
```
I saved this code snippet to a script `test.py` and got after execution
```
python test.py 
test.py:19: UserWarning: Oooooops
  func()
Here I am
```
So yes, the stack level is correct ;-)

hmaarrfk: Thanks!

lagru: > I think it is, as long as this decorator is the last one that was applied. Maybe that should go into the notes section unless we want to make this a parameter for the decorator.

Didn't notice that you already replied but I think it's till relevant.

skimage/_shared/utils.py
```diff
@@ -20,6 +20,45 @@ class skimage_deprecation(Warning):
     pass
 
 
+class deprecate_arg:
```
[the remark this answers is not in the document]
rfezzani: You are absolutely right, thx ;-)

skimage/_shared/utils.py
```diff
@@ -20,6 +20,45 @@ class skimage_deprecation(Warning):
...
+    ----------
+    arg_mapping: dict
+        mapping between the function's old argument names and the new ones.
+    removed_version : str
+        The package version in which the deprecated argument will be removed.
+
+    """
+
+    def __init__(self, arg_mapping, removed_version=None):
+        self.arg_mapping = arg_mapping
```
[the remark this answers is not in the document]
rfezzani: After the warning message setting `removed_version` is no more used, that's why I did not store as attribute.

rfezzani:
Thank you @lagru for your review and comments. The [functools.warps](https://docs.python.org/2/library/functools.html#functools.wraps) decorator is the generic tool to manage the docstring and function name and signature stuff. So there is no impact on that side with this decorator, but I can add a test for that ;-)

lagru:
> Thank you @lagru for your review and comments. The functools.warps decorator is the generic tool to manage the docstring and function name and signature stuff. So there is no impact on that side with this decorator, but I can add a test for that ;-)

Just tested locally. At least `help()` and IPyton's "?" aren't affected. So you're right. I don't think you need to add a test for that.

rfezzani:
@lagru , I already implemented the test, it doesn't hurt ;-)

hmaarrfk:
@rfezzani if you are going to get into docstring mangling, then maybe it becomes worthwhile to take the code (tests or implementation) from: https://github.com/hmaarrfk/wabisabi shameless plug again, but it was developed for these usecases. Ultimately, I decided it wasn't worth imposing the custom decorator on others. But you are soon approaching some of the other features of the wabisabi decorator.

rfezzani:
@hmaarrfk, I am not planning to go into docstring mangling, unless you think it is a good idea...
I would prefer to keep it simple: deprecating and argument consists only in decorating the target function and replacing old argument by the new one where ever it is necessary (code and doc).

hmaarrfk:
The tests need to be updated, with:
1. Using the new keyword arguments.
2. Optionally, and preferrably, a single test to see if when specifying the parameter with the old name, it still has the desired effect.
      - The warning generated by this test should be asserted in the test suite.

rfezzani:
@hmaarrfk, I implemented your recommendations but unfortunately a timeout makes Travis-CI fail on osX/python3.7 :cry:

rfezzani:
I added to the decorator the option of displaying a custom warning message.

skimage/transform/tests/test_radon_transform.py
```diff
@@ -179,6 +180,15 @@ def test_radon_iradon(interpolation_type, filter_type):
     check_radon_iradon(interpolation_type, filter_type)
 
 
+@pytest.mark.parametrize("filter_type", filter_types)
```
hmaarrfk: didn't need to parameterize, but ok!

rfezzani:
Hi everyone, do you think that this PR still needs more work or is it fine?
If it is merged, I would be happy to go on fixing #4152 using the `deprecate_kwarg` decorator ;-)

jni:
@rfezzani as mentioned on my comment in #4161, I don't have time to review right now but I hope to get more time back in the coming weeks. Hold, please! =) (Or maybe someone else will review before me, but just wanted to make sure you're not just left feeling hanging. =)

rfezzani:
No problem @jni, I am just searching some more issues to work on ;-)

jni:
> I am just searching some more issues to work on ;-)

Oh, *that* we can provide in spades! 😂 Did you see my post on the mailing list about n-dimensional edge filters, for example? =)

rfezzani:
Yes, I saw it, very cool! But I also saw on your message that you *will  submit a PR soon* on that :smile:

hmaarrfk:
I think the only thing missing from this is a `..versionchanged` in the docstring.

rfezzani:
I added it @hmaarrfk ;-)

hmaarrfk:
whoa, so many failures.

rfezzani:
Thank you very much @lagru, fixing #4003, #4154 and #4152 will be much more handy now :wink:

lagru:
@rfezzani Thanks!

PR #4159, opened 2019-09-13
modified foerstner to return response image in line with other corner…

### LINKED ISSUE #4000: Return type of corner_foerstner different from other corner_* functions
## Description

The return type of [corner_foerstner](https://scikit-image.org/docs/stable/api/skimage.feature.html#corner-foerstner) is a tuple : 

    (Error ellipse sizes. Roundness of error ellipse.)

while other corner_* functions ([shi-tomasi](https://scikit-image.org/docs/stable/api/skimage.feature.html#corner-shi-tomasi), [harris](https://scikit-image.org/docs/stable/api/skimage.feature.html#corner-harris), etc) return a response image.

The example of the docstring indicates to do:

`foerstner = (q > roundness_thresh) * (w > accuracy_thresh) * w`

where foerstner is the response image.

zagerpatrick:
corner_foerstner was modified to return the Foerstner corner measure response image instead of the Foerstner corner measures which are currently returned. Fixes #4000 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/corner.py
+++ b/skimage/feature/corner.py
@@ -586,7 +586,7 @@ def corner_shi_tomasi(image, sigma=1):
     return response
 
 
-def corner_foerstner(image, sigma=1):
+def corner_foerstner(image, accuracy_thresh, roundness_thresh, sigma=1):
     """Compute Foerstner corner measure response image.
 
     This corner detector uses information from the auto-correlation matrix A::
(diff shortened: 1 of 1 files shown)


jni:
@zagerpatrick thank you very much for helping with scikit-image! Unfortunately, we can't simply change the API here, since many users might be affected. Instead, we need to go through a deprecation cycle, as explained in our contributing document here:

https://scikit-image.org/docs/dev/contribute.html#deprecation-cycle

Please let us know if you would like more guidance in how to proceed.

skimage/feature/corner.py
```diff
@@ -586,7 +586,7 @@ def corner_shi_tomasi(image, sigma=1):
     return response
 
 
-def corner_foerstner(image, sigma=1):
+def corner_foerstner(image, accuracy_thresh, roundness_thresh, sigma=1):
```
alexdesiqueira: I would put some default values for these two variables. How's the example values (`0.5`, `0.3`) sound?

alexdesiqueira: Also, it would make more sense globally to have `image, sigma, accuracy_thresh, roundness_thresh`, since it'd maintain the model of the other functions. Then, we wouldn't need to deprecate the API. Is that right @jni?

jni: > is that right @jni?

The arguments should indeed come after *and* be made keyword-only by using `sigma=1, *, accuracy_thresh=...`, but the API still needs to go through a deprecation cycle somehow, because the return value is being altered. I don't know what the best way to do this is...

PR #4161, opened 2019-09-13
Add the iterative Lucas-Kanade (iLK) optical flow method

rfezzani:
## Description

Here is an implementation of the iterative Lucas-Kanade (iLK) method for optical flow estimation [[1]](https://www.researchgate.net/profile/Frederic_Champagnat/publication/4186265_Dense_optical_flow_by_iterative_local_window_registration/links/5d25b1ef92851cf44074e8c0/Dense-optical-flow-by-iterative-local-window-registration.pdf?_sg%5B0%5D=1yoRq9BYNnmR7MxEIliAWmFkzqb_O0Up4HC-spZQEr_bBYa0D_mxim0SmX4BGkA3R_KbZTRIcCEcUVTLcKIa0w.KvokhKD_TUxsZAFghKUe0ncSXglVVa3e-oRvoDpvhZhKbzKFgWRkMZTyocilR3uc9CJfBxG5eTC6plXGY7D6hA&_sg%5B1%5D=HLuynwMvvrLHI0Iu9UJ5unG3BxalBTQkxbCHbsnc0Ql5Pt5akK-CBNbpkuU6Lpq5i0Q8Q4_rpsazLtrt7Pz7yuQlkSoy_fcQ-tXhqrYYt1mg.KvokhKD_TUxsZAFghKUe0ncSXglVVa3e-oRvoDpvhZhKbzKFgWRkMZTyocilR3uc9CJfBxG5eTC6plXGY7D6hA&_iepl=).
The [Lucas-Kanade method](https://en.wikipedia.org/wiki/Lucas%E2%80%93Kanade_method) is a local approach for the estimation of optical flow (as opposed to global approaches such TV-L1 for example). It is fast, robust and requires basically only one parameter that is the size of the neighborhood considered for the estimation.

**[1]**: Le Besnerais, G., & Champagnat, F. (2005, September). Dense optical flow by iterative local window registration. In IEEE International Conference on Image Processing 2005 (Vol. 1, pp. I-137). IEEE.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


jni:
Ha! I didn't realise the question of how to present multiple optical flow algorithms in the API would become so relevant so soon! :joy: Thanks @rfezzani! I also notice that it's nD out of the box, a very nice touch. ;)

I don't have time for a deep review right this second but it looks pretty great on first glance!

rfezzani:
:joy: Thank you @jni, the code was almost done in [pyimof](https://github.com/rfezzani/pyimof), I just had to implement the nD support ;-) I hope you like it.
I had a look on Travis CI, but I don't understand what makes it Fail :-(

lagru:
> I had a look on Travis CI, but I don't understand what makes it Fail :-(

Both failures are due to [`optical_flow_ilk` returning float64 when float32 is expected](https://travis-ci.org/scikit-image/scikit-image/jobs/588914793#L2631). Does that help or did you mean you don't understand *why* that function returns the wrong dtype?

skimage/registration/tests/test_ilk.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_optical_flow_dtype():
+    # Generate synthetic data
+    rnd = np.random.RandomState(0)
+    image0 = rnd.normal(size=(256, 256))
+    gt_flow, image1 = _sin_flow_gen(image0)
+    # Estimate the flow at double precision
+    flow_f64 = optical_flow_ilk(image0, image1, dtype='float64')
+
+    assert flow_f64.dtype == 'float64'
```
lagru: First time seeing that way to check dtypes. Seems a bit brittle and doesn't seem to work for cases such as `np.float64 == "float64"` which will return `False`. So maybe use `np.float64` here (and similar at other places) instead?

skimage/registration/tests/test_ilk.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_3d_motion():
+    # Generate synthetic data
+    rnd = np.random.RandomState(0)
+    image0 = rnd.normal(size=(128, 128, 128))
+    gt_flow, image1 = _sin_flow_gen(image0)
+    # Estimate the flow
+    flow = optical_flow_ilk(image0, image1)
+    # Assert that the average absolute error is less then half a pixel
+    assert abs(flow - gt_flow) .mean() < 0.5
```
lagru:
```suggestion
    assert abs(flow - gt_flow).mean() < 0.5
```

skimage/registration/tests/test_ilk.py
```diff
@@ -0,0 +1,108 @@
...
+
+def test_2d_motion():
+    # Generate synthetic data
+    rnd = np.random.RandomState(0)
+    image0 = rnd.normal(size=(256, 256))
+    gt_flow, image1 = _sin_flow_gen(image0)
+    # Estimate the flow
+    flow = optical_flow_ilk(image0, image1)
+    # Assert that the average absolute error is less then half a pixel
+    assert abs(flow - gt_flow) .mean() < 0.5
```
lagru:
```suggestion
    assert abs(flow - gt_flow).mean() < 0.5
```

skimage/registration/tests/test_ilk.py
rfezzani: good catch!!! ;-)

skimage/registration/tests/test_ilk.py
rfezzani: You are right, `np.float64 == "float64"` returns `False`, but `np.dtype(np.float64) == 'float64'` is `True`. Here I compare data type object...
But I will check if this can be done differently ;-)

rfezzani:
> > I had a look on Travis CI, but I don't understand what makes it Fail :-(
> 
> Both failures are due to [`optical_flow_ilk` returning float64 when float32 is expected](https://travis-ci.org/scikit-image/scikit-image/jobs/588914793#L2631). Does that help or did you mean you don't understand _why_ that function returns the wrong dtype?

I actually don't understand why that function returns the wrong dtype :smile:. Since the tests pass in the other checks...

rfezzani:
OK, I finally reproduced the CI failure: It comes from `numpy` version 1.13.3!

rfezzani:
The CI failure is related to the issue #4189.

rfezzani:
Hi everyone, I added some comments to the code to ease code review, if anyone is interested ;-)

jni:
@rfezzani it's late here but this is on my to-do! =) Thank you!

rfezzani:
Just to point that this PR actually implements a version of the FOLKI algorithm, on which the [Gefolki](https://w3.onera.fr/medusa/gefolki) projects is based.

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+    dtype = reference_image.dtype
+    ndim = reference_image.ndim
+    size = 2 * radius + 1
+
+    if gaussian:
+        sigma = (0, ) + ndim * (size / 4, )
+        filter_func = partial(ndi.gaussian_filter, sigma=sigma, mode='mirror')
+    else:
+        filter_func = partial(ndi.uniform_filter, size=(1, ) + ndim * (size, ),
+                              mode='mirror')
```
jni: Is there ever a good reason to use uniform over gaussian?

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+    grid = np.meshgrid(*[np.arange(n, dtype=dtype)
+                         for n in reference_image.shape], indexing='ij')
+
+    for _ in range(num_warp):
+        if prefilter:
+            flow = ndi.filters.median_filter(flow, (1, ) + ndim * (3, ))
+
+        moving_image_warp = warp(moving_image, grid + flow, mode='nearest')
+        grad = np.array(np.gradient(moving_image_warp))
+        It = (grad * flow).sum(0) + reference_image - moving_image_warp
```
jni:
```suggestion
        It = (grad * flow).sum(axis=0) + reference_image - moving_image_warp
```

I also don't love the name `It`. Could you change it to something more descriptive?

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+    coef = np.zeros((int((ndim * (ndim + 1)) / 2 + ndim), )
+                    + reference_image.shape, dtype=dtype)
+    # For each pixel location (i, j), the optical flow X = flow[:, i, j]
+    # is the solution of the ndim x ndim linear system
+    # A[i, j] * X = b[i, j]
+    A = np.zeros(reference_image.shape + (ndim, ndim), dtype=dtype)
+    b = np.zeros(reference_image.shape + (ndim, ), dtype=dtype)
+
+    grid = np.meshgrid(*[np.arange(n, dtype=dtype)
+                         for n in reference_image.shape], indexing='ij')
```
jni: Could you add a comment here and in the function docstring that the space required for this computation is (as far as I can tell) ndim * (ndim + 2) of the original image size, in addition to the image itself?

jni: Actually, that's just from the above three lines, there's lots of other volumes being created willy-nilly... LOL

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+        grad = np.array(np.gradient(moving_image_warp))
+        It = (grad * flow).sum(0) + reference_image - moving_image_warp
+
+        # Local linear systems creation
+        k = 0
+        for i in range(ndim):
+            for j in range(i, ndim):
+                coef[k] = grad[i] * grad[j]
+                k += 1
+            coef[i - ndim] = grad[i] * It
```
jni: I kinda like the below better, as I am allergic to nested loops and to external counters. =)

```python
from itertools import combinations_with_replacement as comb_with_replacement


for k, (i, j) in enumerate(comb_with_replacement(range(ndim, 2))):
    coef[k] = grad[i] * grad[j]
    if j == ndim - i:
        coef[i - ndim] = grad[i] * It
```

Can you explain in a comment the intuition behind the `coef[i - ndim]` line?

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+        filter_func(coef, output=coef)
+
+        k = 0
+        for i in range(ndim):
+            A[..., i, i] = coef[k]
+            b[..., i] = coef[i - ndim]
+            k += 1
+            for j in range(i + 1, ndim):
+                A[..., i, j] = A[..., j, i] = coef[k]
+                k += 1
```
jni: As above.

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+        The first gray scale image of the sequence.
+    moving_image : ndarray, shape (M, N[, P[, ...]])
+        The second gray scale image of the sequence.
+    radius : int, optional
+        Radius of the window considered around each pixel.
+    num_warp : int, optional
+        Number of times moving_image is warped.
+    gaussian : bool, optional
+        if True, a gaussian kernel is used for the local
+        intagration. Otherwise, a uniform kernel is used.
```
jni:
```suggestion
        integration. Otherwise, a uniform kernel is used.
```

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+
+def optical_flow_ilk(reference_image, moving_image, *,
+                     radius=7, num_warp=10, gaussian=False,
+                     prefilter=False, dtype=np.float32):
+    """Coarse to fine optical flow estimator.
+
+    The iterative Lucas-Kanade (iLK) solver is applied at each level
+    of the image pyramid. iLK is a fast and robust algorithm
+    developped by Le Besnerais and Champagnat [4]_ and improved in
+    [5]_..
```
jni:
```suggestion
    [5]_.
```

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+    radius : int, optional
+        Radius of the window considered around each pixel.
+    num_warp : int, optional
+        Number of times moving_image is warped.
+    gaussian : bool, optional
+        if True, a gaussian kernel is used for the local
+        intagration. Otherwise, a uniform kernel is used.
+    prefilter : bool, optional
+        whether to prefilter the estimated optical flow before each
+        image warp. This helps to remove the potential outliers.
```
jni:
```suggestion
        image warp. This helps to remove potential outliers.
```

skimage/registration/tests/test_ilk.py
jni: `flow.dtype == np.float64` should work?

rfezzani:
Thank you @jni for your review.
The iLK method is really cheaper then the tvl1 method: it mainly consists in building and solving a nD×nD symmetric linear system per pixel for a given couple of nD images.

skimage/registration/_optical_flow.py
rfezzani: uniform is supposed to be more robust then gaussian for images of particles for example.

skimage/registration/tests/test_ilk.py
rfezzani: Yes, absolutely :-)

rfezzani:
OK @jni, I see. Your concern is about the memory use... You are right, the linear system building requires nD×(nD+1)×image.size memory allocation.
If symmetry was exploited, it can be reduced to nD×(nD +3)/2×image.size. But as far as I know, scipy doesn't provide a direct solver for symmetric matrice, nor an efficient data storage for it :cry:

skimage/registration/_optical_flow.py
```diff
@@ -216,3 +216,164 @@ def optical_flow_tvl1(reference_image, moving_image,
                      tol=tol, prefilter=prefilter)
 
     return coarse_to_fine(reference_image, moving_image, solver, dtype=dtype)
+
+
+def _ilk(reference_image, moving_image, flow0, radius, num_warp, gaussian,
+         prefilter):
```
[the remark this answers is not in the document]
rfezzani: This function is private... The keyword-only args is already used in the public function `optical_flow_ilk`.
Do you want to me to use it here as well?

rfezzani:
Thank you @jni for your comments. I removed the `coef` array allocation that was used to vectorize the calls to `scipy.ndimage` filters.
This simplifies the code and saves memory, but it removes this small optimization (`scipy.ndimage` filters are now called in a small python loop).
The memory use in `ilk` is now equivalent to `tvl1`.

skimage/registration/_optical_flow.py
grlee77: You don't have to store the full dense grid if you use
```Python
grid = np.meshgrid(*[np.arange(n, dtype=float) for n in (256, 256, 192)], indexing='ij', sparse=True)
```
It appears `grid` is only ever used in an addition (`grid + flow`) below. In the sparse grid case, an equivalent addition would be:
```Python
np.stack([g + f for g, f in zip(grid, flow)], axis=0)
```
which is a bit less readable, but does allow not storing the full grid. A more verbose, sparse grid addition helper would be:
```Python
def _add_flow(grid, flow):
    out = flow.copy()
    for n, g in enumerate(grid):
        out[n, ...] += g
    return out
```

rfezzani: Thank you @grlee77 for this suggestion, I vote for defining a helper function which can also be used in `_tvl1` :wink:.

emmanuelle:
Hi @rfezzani I had missed this nice PR, it good that the 0.18 label attracted my attention :-). Did you think about a possible gallery example ? I took a look at https://www.mathworks.com/matlabcentral/fileexchange/48745-lucas-kanade-tutorial-example-2 which is a very nice example. Like for example if we had two images of cells at different times would we be able to compute their velocity field? Or another kind of example?

rfezzani:
Hi @emmanuelle, I will try to find a pretty illustration. Any suggestion is welcome :wink:

emmanuelle:
I as thinking that maybe @royerloic @batson or @GenevieveBuckley could have nice data with a CC0 or public domain license. If I understand well what the algorithm does, suitable data could be two consecutive images in a timeseries where objects (like cells for examples) are moving but not all in the same direction and the iLK algorithm would detect these different directions. 

@rfezzani is this an algorithm which you are using for your own applications ?

rfezzani:
> @rfezzani is this an algorithm which you are using for your own applications ?

I worked with this algo during my PhD/post-doc... It's been a while now :smile:

rfezzani:
I see in SIPI dataset the [Toy Vehicle](http://sipi.usc.edu/database/database.php?volume=sequences&image=60#top) that can be a good illustration for this algo. I will try to come soon with an example :wink:.

GenevieveBuckley:
> I as thinking that maybe @royerloic @batson or @GenevieveBuckley could have nice data with a CC0 or public domain license. 

None that I own outright, sorry

rfezzani:
I ended up using a sequence from the [PIV challenge 2001](http://www.pivchallenge.org/pub/index.html#b)
ilk
I hope you like it :wink:

emmanuelle:
The figure is great! It's just too bad the PIV challenge did not explicitly give a license to their data (or did they?). Is it ok to use these files if we just download them? I'd say it but it's a bit tangent (@jni?). Maybe we can ask the author of the image if they can release the image in the public domain or as CC0?

rfezzani:
We are publishing results using these data (as any paper citing the PIV challenge) I am not sure that we have to care about license here since we are citing the source of the data...

emmanuelle:
The example has been discussed during the dev meeting and the consensus has been that the source of the images should not block the PR. Nevertheless, asking the authors whether they are willing to license their data is a good idea.

rfezzani:
I see on the [PIV challenge](http://www.pivchallenge.org/) webpage:

> This is the official web-site for PIV Challenge. All of the information, including the test images, are found on this web-site. Please freely link this web-site. 

and at the bottom

> Copyright 2013-2016 All Right Reserved. International PIV Challenge 

Does it answer the License question @emmanuelle?

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +218,152 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+    flow : ndarray, shape ((reference_image.ndim, M, N[, P[, ...]])
+        The estimated optical flow components for each axis.
+
+    Notes
+    -----
+    Color images are not supported.
+
+    References
+    ----------
+    .. [4] Le Besnerais, G., & Champagnat, F. (2005, September). Dense
```
grlee77, line 344: These references should be numbered [1] and [2] rather than starting from [4].

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +218,152 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+    radius : int, optional
+        Radius of the window considered around each pixel.
+    num_warp : int, optional
+        Number of times moving_image is warped.
+    gaussian : bool, optional
+        if True, a gaussian kernel is used for the local
+        integration. Otherwise, a uniform kernel is used.
+    prefilter : bool, optional
+        whether to prefilter the estimated optical flow before each
+        image warp. This helps to remove potential outliers.
```
grlee77, line 327: It would be nice to specify exactly what type of prefiltering this involves. Perhaps adding: `When True, a median filter with window size 3 along each axis is applied.`.
I think this text was copied from the existing TV-L1 variant, so perhaps change it there as well.

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +218,152 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+    reference_image : ndarray, shape (M, N[, P[, ...]])
+        The first gray scale image of the sequence.
+    moving_image : ndarray, shape (M, N[, P[, ...]])
+        The second gray scale image of the sequence.
+    radius : int, optional
+        Radius of the window considered around each pixel.
+    num_warp : int, optional
+        Number of times moving_image is warped.
+    gaussian : bool, optional
+        if True, a gaussian kernel is used for the local
```
grlee77, line 323:
```suggestion
        If True, a Gaussian kernel is used for the local
```

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +218,152 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+                         for n in reference_image.shape],
+                       indexing='ij', sparse=True)
+
+    for _ in range(num_warp):
+        if prefilter:
+            flow = ndi.filters.median_filter(flow, (1, ) + ndim * (3, ))
+
+        moving_image_warp = warp(moving_image, get_warp_points(grid, flow),
+                                 mode='nearest')
+        grad = np.array(np.gradient(moving_image_warp))
```
grlee77, line 280:
```suggestion
        grad = np.stack(np.gradient(moving_image_warp), axis=0)
```
Nitpick: array stacking style consistent with #4826

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +218,152 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+        It = (grad * flow).sum(axis=0) + reference_image - moving_image_warp
```
grlee77, line 281: can we make a more descriptive lowercase name for this `It` term? (I am fine leaving the Hessian matrices term, `A`, capitalized)

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +218,152 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+
+        # Local linear systems creation
+        for i, j in combinations_with_replacement(range(ndim), 2):
+            A[..., i, j] = A[..., j, i] = filter_func(grad[i] * grad[j])
```
grlee77, line 285: If I am reading the referenced Paley2016 publication section 2.2 correctly, this computation as well as the `grad` computation should occur outside of this `num_warps` loop.
They mention that the Hessians, `A` (**H** in the publication), are computed "during initialization". Specifically, the following should be placed prior to the `num_warps` loop and then the `grad` and `A` computations removed from the loop. (Note also the change to "reference_image" as the input to the gradient computation)

```Python
    grad = np.array(np.gradient(reference_image))
    # Local linear systems creation
    for i, j in combinations_with_replacement(range(ndim), 2):
        A[..., i, j] = A[..., j, i] = filter_func(grad[i] * grad[j])
```


So, the loop will only update `flow`, `b` and `moving_image_warp` warp at each iteration, but not `grad` or `A`.

The test suite still passes for me with this change.

grlee77 (changes requested):
Overall this seems to work pretty well and is close, but there are a few issues left to address.

Note that there is a general question regarding whether the algorithm should always precompute `grad` and `A` outside of the `num_warps` loop (see other review comments)

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +218,152 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+    flow : ndarray, shape ((reference_image.ndim, M, N[, P[, ...]])
+        The estimated optical flow components for each axis.
+
+    """
+    dtype = reference_image.dtype
+    ndim = reference_image.ndim
+    size = 2 * radius + 1
+
+    if gaussian:
+        sigma = (0, ) + ndim * (size / 4, )
```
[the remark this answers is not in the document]
rfezzani: Good catch, that was when this operation was vectorized :clap:

skimage/registration/_optical_flow.py
rfezzani: Is `time_diff` OK?

skimage/registration/_optical_flow.py
rfezzani: @grlee77, you are perfectly reading section 2.2 of [[2]](https://www.researchgate.net/profile/Aurelien_Plyer/publication/262948097_Massively_Parallel_Lucas_Kanade_Optical_Flow_for_Real-Time_Video_Processing_Applications/links/54b4db3e0cf2318f0f96f01c.pdf) :wink:, but the algorithm described there solves the *inverse* form fo Lucas-Kanade algo (see section 3.2 [here](http://ncorr.com/download/publications/bakerunify.pdf) as ref for example). This form is briefly discussed in section 3.2 of [[1]](https://www.researchgate.net/profile/Frederic_Champagnat/publication/4186265_Dense_optical_flow_by_iterative_local_window_registration/links/5d25b1ef92851cf44074e8c0/Dense-optical-flow-by-iterative-local-window-registration.pdf).
Current implementation is described in Table 2 of [1].

rfezzani:
Thank you very much @grlee77 for your review :wink:

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +218,152 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+        time compared to double precision.
+
+    Returns
+    -------
+    flow : ndarray, shape ((reference_image.ndim, M, N[, P[, ...]])
+        The estimated optical flow components for each axis.
+
+    Notes
+    -----
+    Color images are not supported.
```
grlee77, line 340: Can you specify either here or in the description above that the algorithm being implemented corresponds to Table 2 of [1]_?

skimage/registration/_optical_flow.py
grlee77: That seems okay. I see Table 2 of [1] refers to this as the "error image" so `error_image` also seems reasonable

skimage/registration/_optical_flow.py
grlee77: Okay, thanks. So, this "MIWS" version is a little slower, but should have better stability. I just left a new comment about adding a Note in the docstring guiding the user interested in details to the specific algorithm within the publications that was used.

rfezzani:
Thank you again @grlee77 for your excellent review, I think I addressed all your requests now :wink:.

grlee77 (approved):
Thanks @rfezzani for addressing the comments. This seems good to go.

emmanuelle (commented):
just a few comments as I was reading

skimage/registration/_optical_flow.py
```diff
@@ -217,3 +220,156 @@ def optical_flow_tvl1(reference_image, moving_image,
...
+        for i in range(ndim):
+            b[..., i] = filter_func(grad[i] * error_image)
+
+        # Don't consider badly conditioned linear systems
+        idx = abs(np.linalg.det(A)) < 1e-14
+        A[idx] = np.eye(ndim, dtype=dtype)
+        b[idx] = 0
+
+        # Solve the local linear systems
+        flow = np.transpose(np.linalg.solve(A, b),
```
emmanuelle, line 299: nitpicking here, but I think `np.moveaxis` would be clearer here than `np.transpose`, unless I'm missing something and it's something different.

emmanuelle:
@rfezzani one thing which I've been missing either in docstrings or in the gallery example (maybe both) is when to use which optical flow algorithm. From what I understand, one should use TV-l1 to allow for discontinuities in the flow, but when the flow is smooth one should prefer ilk which is faster. Is this correct? Could you please add a few explanations on how to select an optical flow algorithm? Algorithm selection is maybe as hard as parameter selection for our users :-).

emmanuelle:
@rfezzani thank you very much for the PR! I have finished reading through the Plyer et al. paper, and the PR is ready to go for me. The only caveat would be to add a few explanations about when to use which algorithm, please tell us if you have the time to add it now or not. But there should not be other comments from me!

rfezzani:
Sorry @emmanuelle, I didn't found time yesterday :/ I will try to manage adding notes in the doc today.

emmanuelle:
Thanks @rfezzani :-). Still in time for the release, don't worry.

rfezzani:
@emmanuelle, I modified `optical_flow_ilk` docstring's long description:
>    The iterative Lucas-Kanade (iLK) solver is applied at each level
>    of the image pyramid. iLK [1]_ is a fast and robust alternative to
>    TVL1 algorithm although less accurate for rendering flat surfaces
 >   and object boundaries (see [2]_).

Is it enough or should I elaborate more?

emmanuelle:
Thanks @rfezzani . Let's wait for CI and then I'll merge!

rfezzani:
I think CI failure is not related...

emmanuelle:
Merged :-) !

rfezzani:
:tada: Thank you @emmanuelle !

PR #4162, opened 2019-09-14
Fix blob dog/log for anisotropic data

### LINKED ISSUE #3777: feature.blob_(log|dog) bug with 3D images
## Description

When using 3D images with```feature.blog_dog``` or ```feature.blob_log``` and a sequence argument for min and max sigma, the sigmas are converted from indexes back to their original values (as in [here](https://github.com/scikit-image/scikit-image/blob/2962429237988cb60b9b317aa020ca3bab100b7f/skimage/feature/blob.py#L301)) and stacked horizontally (as in [here](https://github.com/scikit-image/scikit-image/blob/2962429237988cb60b9b317aa020ca3bab100b7f/skimage/feature/blob.py#L308)) to give an array with 3 spatial coordinates and 3 sigma values.  This is then passed to ```_prune_blobs``` which selects all but the last column for the KDTree construction (```spatial.cKDTree(blobs_array[:, :-1])```) meaning that the tree is considering 3 spatial coordinates and 2 out of 3 sigma values for distance calculations, which seems to be a mistake.

## Way to reproduce
NA

## Version information
master branch

jni:
## Description

Fixes #3777 

Pruning was being done incorrectly when the input data was anisotropic, as introduced in #3690. 

Computing the overlap between two ellipses appears to be hard in general (every solution I found online relies on numerical integration), so instead I rescaled the input coordinates to turn the ellipses into spheres. This is only possible because the sigmas are guaranteed to be proportional across dimensions, even for different scales, since they all come from the same `linspace` call.

Thanks to @eric-czech for reporting this. @ambrosejcarr I would very much appreciate your help in reviewing, since this will probably affect your downstream usage, I guess in starfish?

One last note: our pruning code has always been broken for ndim > 3, now it remains broken but in a more systematic way — effectively no pruning is done at all for higher dimensions.

## Checklist

  existing benchmark~~
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -437,16 +437,14 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
     max_sigma = np.asarray(max_sigma, dtype=float)
 
     if log_scale:
-        start = np.log10(min_sigma)
-        stop = np.log10(max_sigma)
-        sigma_list = np.stack([np.logspace(_start, _stop, num_sigma)
-                               for _start, _stop in zip(start, stop)],
-                              axis=1)
-        # The line below may only be used with numpy 1.16 and above
-        # https://github.com/numpy/numpy/commit/58ebb6a7d77cf89afeb888a70aff23e03d213788
-        # sigma_list = np.logspace(start, stop, num_sigma)
+        # for anisotropic data, we use the "highest resolution/variance" axis
+        standard_axis = np.argmax(min_sigma)
+        start = np.log10(min_sigma[standard_axis])
+        stop = np.log10(max_sigma[standard_axis])
+        scale = np.logspace(start, stop, num_sigma)[:, np.newaxis]
+        sigma_list = scale * min_sigma / np.max(min_sigma)
     else:
-        scale = np.linspace(0, 1, num_sigma)[:, None]
+        scale = np.linspace(0, 1, num_sigma)[:, np.newaxis]
         sigma_list = scale * (max_sigma - min_sigma) + min_sigma
 
     # computing gaussian laplace
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -451,8 +451,8 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
 
     # computing gaussian laplace
     # average s**2 provides scale invariance
-    gl_images = [-gaussian_laplace(image, s) * s ** 2
-                 for s in np.mean(sigma_list, axis=1)]
+    gl_images = [-gaussian_laplace(image, s) * np.mean(s) ** 2
+                 for s in sigma_list]
 
     image_cube = np.stack(gl_images, axis=-1)
 
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -81,10 +81,11 @@ def _compute_sphere_overlap(d, r1, r2):
     return vol / (4./3 * math.pi * min(r1, r2) ** 3)
 
 
-def _blob_overlap(blob1, blob2):
+def _blob_overlap(blob1, blob2, *, sigma_dim=1):
     """Finds the overlapping area fraction between two blobs.
 
-    Returns a float representing fraction of overlapped area.
+    Returns a float representing fraction of overlapped area. Note that 0.0
+    is *always* returned for dimension greater than 3.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


jni:
@scikit-image/core even though I tagged @ambrosejcarr here I would still appreciate a review from any of you! ;) This is tagged as holding up 0.16 and I think it's important enough to hold that up. I'll be going through more of those tags tomorrow.

skimage/feature/blob.py
```diff
@@ -429,8 +451,8 @@ def blob_log(image, min_sigma=1, max_sigma=50, num_sigma=10, threshold=.2,
 
     # computing gaussian laplace
     # average s**2 provides scale invariance
-    gl_images = [-gaussian_laplace(image, s) * s ** 2
-                 for s in np.mean(sigma_list, axis=1)]
+    gl_images = [-gaussian_laplace(image, s) * np.mean(s) ** 2
+                 for s in sigma_list]
```
lagru: Did you intentionally not apply `np.mean` to the `s` given to `gaussian_laplace`? Or should this rather be `gaussian_laplace(image, np.mean(s))`?

ambrosejcarr:
I don't have time for a proper review right now, but the changes here look good to me. The scaling is a clever solution and should work fine with our use of this function in `starfish`. Thanks for the Fix @jni and the catch @eric-czech.

skimage/feature/blob.py
jni: Yes, this was a bug in the original implementation, see `blob_dog` for the equivalent code. It makes no sense to have this accept multiple sigma if you're just going to take the mean of them!

stefanv:
@jni Worth adding some documentation on what is broken and should be fixed for N>=3 still?

jni:
@stefanv >3, I think, which is why I'm not too worried... =) But, yes.

jni:
@stefanv the docstring specifies that `image` must be 2D or 3D... I think that is probably sufficient...?

stefanv:
Fine by me!

jni:
Ach, one sec, this code does not guarantee that `blob_log` has proportional spacing across sigmas:

```python
    if log_scale:
        start = np.log10(min_sigma)
        stop = np.log10(max_sigma)
        sigma_list = np.stack([np.logspace(_start, _stop, num_sigma)
                               for _start, _stop in zip(start, stop)],
                              axis=1)
        # The line below may only be used with numpy 1.16 and above
        # https://github.com/numpy/numpy/commit/58ebb6a7d77cf89afeb888a70aff23e03d213788
        # sigma_list = np.logspace(start, stop, num_sigma)
    else:
        scale = np.linspace(0, 1, num_sigma)[:, None]
        sigma_list = scale * (max_sigma - min_sigma) + min_sigma
```

`blob_dog` does:

```python
    # k such that min_sigma*(sigma_ratio**k) > max_sigma
    k = int(np.mean(np.log(max_sigma / min_sigma) / np.log(sigma_ratio) + 1))
    # a geometric progression of standard deviations for gaussian kernels
    sigma_list = np.array([min_sigma * (sigma_ratio ** i)
                           for i in range(k + 1)])
```

So, somehow, we need to ensure that the sigmas have the same anisotropy (ratio between different axes) along every scale in `blob_log`. Will have to figure that one out... =\

jni:
Done. I don't know how to test for this, before you all ask. =P

skimage/feature/blob.py
emmanuelle: oh yes that was a bug! For the scaling factor, is taking the `mean` rather than another statistical value (median, harmonic mean, etc.) arbitrary? I guess it's not very important.

skimage/feature/blob.py
```diff
@@ -141,22 +156,27 @@ def _prune_blobs(blobs_array, overlap):
...
 
     Returns
     -------
     A : ndarray
         `array` with overlapping blobs removed.
     """
-    sigma = blobs_array[:, -1].max()
-    distance = 2 * sigma * sqrt(blobs_array.shape[1] - 1)
-    tree = spatial.cKDTree(blobs_array[:, :-1])
+    sigma = blobs_array[:, -sigma_dim].max()
```
emmanuelle: shouldn't this be `sigma = blobs_array[:, -sigma_dim:].max()` ?

emmanuelle:
There was only one point I was not sure about, otherwise looks good to me. Minor doc point: should explain more clearly that when sigma is given as lists, they all increase together, ie it's not the cross-product of the ranges which is used?

skimage/feature/blob.py
jni: Oops! Yes! Fixed!

jni: Very nice catch btw. ;)

jni:
@emmanuelle it's a fair point about the documentation, but it's too messy to document properly right now... I think the solution is to clean up how the scaling is worked out so that min means min and max means max. I don't want to do that in this PR though. I'll make an issue to keep track of a future fix.

PR #4165, opened 2019-09-16
Add function to remove near objects in nd-image

### LINKED ISSUE #3816: Add minimal distance argument to local_maxima

lagru:
## Description

Replaces #4024 with an implementation based on [SciPy's cKDTree](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html).

I propose to add a function that iterates over objects (connected pixels that are not zero) inside a <strike>binary</strike> image and removes neighboring objects until all remaining ones are more than a minimal euclidean distance from each other.

<strike>Furthermore it moves the functions `_resolve_neighborhood`, `_set_edge_values_inplace` and `_fast_pad` into `morphology._util`. They are used for several nd-algorithms throughout the submodule so that seems like a sensible place to keep them.</strike> Done in #4209

This function might be used to implement a distance "filtering" for functions like `local_maxima` as well (closes #3816).

## Checklist

- A descriptive but concise pull request title
- [Unit tests](https://scikit-image.org/docs/dev/development/contribute.html#testing)
- A gallery example in `./doc/examples` for new features
- [Contribution guide](https://scikit-image.org/docs/dev/development/contribute.html) is followed

## Release note

For maintainers and optionally contributors, please refer to the [instructions](https://scikit-image.org/docs/stable/development/contribute.html#documenting-changes) on how to document this PR for the release notes.

```release-note
{label="New feature"} Add `skimage.morphology.remove_objects_by_distance`,
which removes labeled objects, ordered by size (default), until the remaining objects are a given distance apart.
```

```release-note
{label="Documentation"} Add a new gallery example on "Removing objects"
based on their size or distance.
```


jni:
@lagru thanks for this! Sorry that I don't have time for a full review just yet, but this is something I've wanted for a long time! =)

My three comments from a brief overview:

- couldn't you use flood fill directly rather than reimplement it?
- Should we maybe set the default priority to object size rather than C-order priority? Whatever we end up choosing, I would prefer that it be independent of axis direction! If size is too brittle (how to break ties?), then label ID might be a better choice. They might coincide sometimes, but in other times (e.g. watershed), it might work well. =)
- Does SciPy's cKDtree have a C API we can use directly? And is it worth it anyway. ;)

Will do a proper review soon!

lagru:
> couldn't you use flood fill directly rather than reimplement it?

I think it might be possible to use [_flood_fill_equal](https://github.com/scikit-image/scikit-image/blob/master/skimage/morphology/_flood_fill_cy.pyx#L40-L86). However there are two disadvantages compared to the current solution:

- The current approach reuses a single queue. This queue reallocates its internal buffer when it grows to small. Reusing that queue allows us to get away with a minimal number of reallocations.
- _flood_fill_equal releases the gil. Generally that's a good thing. But I think I read somewhere that this can actually be slower if the time spent in the released state is very short. That's relevant for the current implementation because there might be many maxima which are only 1 sample large which would lead to releasing and acquiring the GIL very often in short order. 

But I'll try it out and see what the benchmarks say so that we can make a more informed decision.

> Should we maybe set the default priority to object size rather than C-order priority? Whatever we end up choosing, I would prefer that it be independent of axis direction! If size is too brittle (how to break ties?), then label ID might be a better choice. They might coincide sometimes, but in other times (e.g. watershed), it might work well. =)

Good idea! Using the object size sounds more intuitive. Currently the priority actually corresponds to the label ID which is assigned by C-order. I'll see what I can do.

> Does SciPy's cKDtree have a C API we can use directly? And is it worth it anyway. ;)

That's something I'd like to know as well and a weak point of the current implementation. `query_ball_point` is called for *every* object sample that remains in the final image. Calling Python from Cython for something that's executed that often feels just wrong. :sweat_smile: If we could somehow use a C/Cython interface I'd expect significant gains performance wise. Furthermore we could consider releasing the GIL for the whole function. I already looked into doing this earlier and had the following thoughts:

- [query_ball_point](https://github.com/scipy/scipy/blob/master/scipy/spatial/ckdtree.pyx#L835-L995) is actually implemented in Cython but only has a Python interface (`def` and not `cpdef`). Perhaps we could ask SciPy to change that but then we'd have to wait until we can rely on that feature (minimal required SciPy version).
- Internally query_ball_point calls a function of the same name that is declared in C++ [here](https://github.com/scipy/scipy/blob/master/scipy/spatial/ckdtree/src/ckdtree_decl.h#L105-L113). I think we could potentially use that function directly (I doubt its considered part of SciPy's public API) but I couldn't figure out how to include that header file / use that function. The header file is not available unless SciPy is installed from source. But I'm not sure if we actually need that header and could use the shipped binary of `ckdtree` directly ([ckdtree.pyx](https://github.com/scipy/scipy/blob/master/scipy/spatial/ckdtree.pyx#L121-L128) also contains a declaration of that function). 

However because a KDTree is such a nice data structure for this problem :heart_eyes:  the performance is already on par with the "brute force" solution in #4024 and the code is way shorter!

lagru:
Note to self: Find out if there is an efficient way to find the surface of objects. That should make significant performance gains possible because only the surface between two objects needs to be evaluated to find their distance to each other (the inside of objects can be ignored)!

lagru:
> couldn't you use flood fill directly rather than reimplement it?

@jni There is another reason why I wouldn't recommend using flood fill here: The current approach uses labels to avoid crossing the image edge in the raveled view and can forgo padding / creating an array of flags (we have to label the image anyway).

jni:
@lagru thank you for picking this up again! =) One point: how does this function decide the order in which to traverse the objects?

lagru:
@jni I'm not sure whether you are referring to the traversal order of samples belonging to one object or the traversal order when considering separate objects for removal. 
- In short the former is controlled by the index order as returned by `_offsets_to_raveled_neighbors` but to my knowledge shouldn't have any impact on the final results of this function.
- The latter is controlled by the `priority` parameter which orders objects by their bincount if not provided. It's theoretically possible to assign different priority values to the samples of one and the same object. In that case the algorithm will start at the highest priority sample of an object and find all other samples of the object by iterating the neigborhood.

Does that answer your question?

jni:
> Does that answer your question?

it does, thanks, plus, I just re-read my earlier comment and realised I suggested the new ordering. :facepalm: Sorry, just getting into the rhythm again! :sweat_smile:

lagru:
@scikit-image/core If I may ask: This has been ready for review for some time now. If you can spare the time I'd be very grateful. :slightly_smiling_face:

skimage/morphology/_near_objects_cy.pyx
```diff
@@ -0,0 +1,168 @@
...
+
+
+# Must be defined to use QueueWithHistory
+ctypedef Py_ssize_t QueueItem
+
+
+include "_queue_with_history.pxi"
+
+
+ctypedef fused dtype_t:
```
rfezzani, line 21: The fused type `np_real_numeric` (already defined in ``skimage._shared.fused_numerics.pxd) can be used here :)

lagru: Oh, nice. Didn't know we had that!

grlee77:
I have not reviewed the code in detail, but can you make it clearer in the docstring and the text at the top of the example what distance is being measured? It appears to be a boundary-to-boundary distance of the objects, but this should be stated explicitly. Otherwise a user might assume centroid-to-centroid distance or some other definition.

Is the approach proposed here a well-known approach to the problem, or is it novel? I was just wondering if there is a literature reference we could cite (or precedent for the approach in existing software packages).

lagru:
@grlee77 very good points. I'll update the docstring description.

Concerning the approach, I took my motivation from this https://github.com/scikit-image/scikit-image/issues/3816#issuecomment-476741911 and just took what I did in 1D for scipy.signal.find_peaks and generalized it to N dimensions. Right now I feel a little stupid that you had to remind me to do a proper literature research on this; I'll remedy that. :see_no_evil:

Concerning the centroids: that's a use case I'll think about more. I think part of what would be required to implement that is already covered by this function. Hopefully I have a clearer picture after some research.

jni (approved):
I'm happy with this, with the only caveat that I wonder whether we should *only* support integer-labeled images, and ask users to use `ndi.label` or `measure.label` if they have boolean images.

Thanks for picking it up yet again @lagru! 🎉

jni:
Looks like there's a segfault with optional deps... I doubt that it's due to this PR but would be great to check!

lagru:
> whether we should only support integer-labeled images, and ask users to use ndi.label or measure.label if they have boolean images.

In a vacuum I'd be totally fine with that suggestion. The simpler the function the better. However, I'd like to keep this reasonable consistent with [`remove_small_objects`](https://github.com/scikit-image/scikit-image/blob/00177e14097237ef20ed3141ed454bc81b308f82/skimage/morphology/misc.py#L64-L66) and would therefore at least like to support arbitrary int and bool arrays.

However, I'm fine with deprecating that behavior in `remove_small_objects` as well to simplify our API. `remove_small_objects` calls `scipy.ndimage.label` under the hood as well.

stefanv (approved):
Approved if @lagru agrees it's ready.

lagru:
Thanks @stefanv! I'm happy with this now. :blush:

jni:
I'm happy too, but just want to raise another question: I find it somewhat unintuitive that the priority is given by an *image*, rather than, say, a dictionary of labels->priority, or an array for which the number at `idx` has the priority for label `idx`. (Like the output of bincount.) Like, what happens if a label's region contains more than one priority value? Reading the code, I think it takes the maximum value, but is that documented?

I'm agnostic as to what to decide here because it's easily updated in the future, but just wanted to check what you think.

jarrodmillman (approved):

jarrodmillman:
@lagru This looks good to me (pending your response to @jni). I am thinking about releasing 0.20.0rc0 tomorrow and wanted to check whether you thought we should merge this before or after 0.20. If before, feel free to merge.

lagru:
> I find it somewhat unintuitive that the priority is given by an image, rather than, say, a dictionary of labels->priority, or an array for which the number at idx has the priority for label idx.

That's a very good point and I agree with your intuition. Right now `labels` isn't an input argument but derived from the parameter `image` whose objects can have any number of values. So it's difficult to use unkown label ids in `priority`. But we should probably go the route your suggest and more closely emulate `remove_small_objects` here! I'll look into it as soon as possible.

> wanted to check whether you thought we should merge this before or after 0.20. If before, feel free to merge.

With @jni pointing out that the API can still be improved, I'm hesitant to include it in v0.20.0rc in this state just to have users deal with API changes later on. I don't feel like waiting for another feature release but I don't think we want to backport something like this after v0.20.0rc? I think our policy is to only backport bug fixes is it not or would you be more lenient for a release candidate? If not my vote would be to merge it after v0.20.

lagru:
Sounds good. :blush:

stefanv:
@lagru Let's definitely get this in for 0.21, though!

stefanv:
OK, looks like we still need to follow up on https://github.com/scikit-image/scikit-image/pull/4165#issuecomment-1399423776 here

lagru:
> OK, looks like we still need to follow up on https://github.com/scikit-image/scikit-image/pull/4165#issuecomment-1399423776 here

Yes, I've been exploring an option that relies solely on sorting by priority to discover all the pixels of an object. This allows removing the whole spatial iteration of neighboring pixels. I had high hopes for it being more faster but it seems slower. Doesn't have to be a bad thing though if it makes the algorithm simpler. I'll try to push in the next few days.

lagru:
Ready to merge IMO! :tada: I'm somewhat proud of this one now. :blush:  https://github.com/scikit-image/scikit-image/pull/4165#issuecomment-1399423776 is addressed and I figured out how to include only objects surface in the KD-tree, while still including the inside of objects when they are removed. That shortens up the critical loop of this algorithm significantly. It even works with anistropic data now; https://github.com/scikit-image/scikit-image/issues/7318 gave me the idea.

I included a basic outline of the algorithm in the docstring and documented complicated parts extensively with comments and tests. Hope that makes reviewing easier! :pray:

skimage/morphology/_misc_cy.pyx
```diff
@@ -0,0 +1,122 @@
...
+    remembered_ids = set()
+    for i_indices in range(border_indices.shape[0]):
+        i_out = border_indices[i_indices]
+        object_id = out[i_out]
+        # Skip if sample is part of a removed object
+        if object_id == 0:
+            continue
+
+        neighborhood = kdtree.query_ball_point(
+            kdtree.data[i_indices, ...],
```
tylerjereddy, line 64: Just cross-reviewing from SciPy since there were some concerns expressed about the `list` return type here (which we need because of the ragged data structures that are possible) for performance reasons.

A few questions:

1) Do you have any sample timings/examples demonstrating the bottleneck/% time spent on this line for a typical use case?
2) Any chance you could use the `workers` argument to speed up via concurrency? This might be particularly viable if you could accumulate the query points externally first.
3) How many points are you querying against in a typical use case? The reason I ask is that you seem to be querying the tree against itself, which may benefit from using `query_ball_tree` instead.

lagru: For the following case

```python
import numpy as np
import skimage as ski

image = ski.color.rgb2gray(ski.data.hubble_deep_field())
foreground1 = image > ski.filters.threshold_li(image)  # many small objects
foreground2 = np.ones_like(foreground1)
foreground2[400, :] = 0  # two very large objects
foreground = np.concatenate([foreground1, foreground2])
objects = ski.measure.label(foreground)

%timeit ski.morphology.remove_near_objects(objects, min_distance=5)
%timeit ski.morphology.remove_near_objects(objects, min_distance=50)
%timeit ski.morphology.remove_near_objects(objects, min_distance=100)
%timeit ski.morphology.remove_near_objects(objects, min_distance=300)
```

Querying and entire object's indices at once with `kdtree.query_ball_point(kdtree.data[start:stop], ...]`,

<details><summary>Patch</summary>

```patch
Subject: [PATCH] Query indices with the same object ID at once
---
Index: skimage/morphology/_misc_cy.pyx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/skimage/morphology/_misc_cy.pyx b/skimage/morphology/_misc_cy.pyx
--- a/skimage/morphology/_misc_cy.pyx	(revision c8756cae67078c1cfb05763a503792e05e842aef)
+++ b/skimage/morphology/_misc_cy.pyx	(date 1713872591799)
@@ -46,34 +46,45 @@
         The shape of the unraveled `image`.
     """
     cdef:
-        Py_ssize_t i_indices, j_indices  # Loop variables to index `indices`
+        Py_ssize_t i_indices, j_indices, start, stop  # Loop variables to index `indices`
         Py_ssize_t i_out  # Loop variable to index `out`
-        np_anyint object_id, other_id
-        list neighborhood
+        np_anyint start_id, stop_id
         set remembered_ids
+        list n
 
     remembered_ids = set()
-    for i_indices in range(border_indices.shape[0]):
-        i_out = border_indices[i_indices]
-        object_id = out[i_out]
-        # Skip if sample is part of a removed object
-        if object_id == 0:
+    start = 0
+    start_id = out[border_indices[start]]
+    for stop in range(border_indices.shape[0]):
+        i_out = border_indices[stop]
+        stop_id = out[i_out]
+
+        if start_id == stop_id:
+            continue
+        elif start_id == 0:
+            start = stop
+            start_id = stop_id
             continue
 
         neighborhood = kdtree.query_ball_point(
-            kdtree.data[i_indices, ...],
+            kdtree.data[start:stop, ...],
             r=min_distance,
             p=p_norm,
         )
-        for j_indices in neighborhood:
-            # Check object IDs in neighborhood
-            other_id = out[border_indices[j_indices]]
-            if other_id != 0 and other_id != object_id:
-                # If neighbor ID wasn't already removed or is the current one
-                # remove the boundary and remember the ID
-                _remove_object(out, border_indices, j_indices)
-                remembered_ids.add(other_id)
+
+        for n in neighborhood:
+            for j_indices in n:
+                # Check object IDs in neighborhood
+                other_id = out[border_indices[j_indices]]
+                if other_id != 0 and other_id != start_id:
+                    # If neighbor ID wasn't already removed or is the current one
+                    # remove the boundary and remember the ID
+                    _remove_object(out, border_indices, j_indices)
+                    remembered_ids.add(other_id)
 
+        start = stop
+        start_id = out[border_indices[start]]
+
     # Delete inner parts of remembered objects
     for j_indices in range(inner_indices.shape[0]):
         object_id = out[inner_indices[j_indices]]

```

</details> 
 
is faster for `min_distance=5` and `50` but a lot slower once `min_distance` get's larger. Without profiling I'm not really sure about the reason. The implementation that queries single points only, still seems to be the one that's fastest overall.

Using `workers` seems to introduce an overhead that, for most cases, is slower regardless of the specific implementation.

doc/examples/features_detection/plot_remove_objects.py
```diff
@@ -0,0 +1,44 @@
+"""
+================
+Removing objects
+================
+
+scikit-image supports several ways to remove objects inside N-dimensional
```
stefanv, line 6:
```suggestion
scikit-image has several ways of removing objects from N-dimensional
```

doc/examples/features_detection/plot_remove_objects.py
```diff
@@ -0,0 +1,44 @@
...
+images. In this context, "objects" (and "holes") are defined as groups of
```
stefanv, line 7:
```suggestion
images. Here, "objects" (and "holes") are defined as groups of
```

doc/examples/features_detection/plot_remove_objects.py
```diff
@@ -0,0 +1,44 @@
...
+connected samples that are distinct from the background. A binary image can
```
stefanv, line 8: Maybe say what "connected means".

Presumably, not only binary images can contain several disconnected objects.

doc/examples/features_detection/plot_remove_objects.py
```diff
@@ -0,0 +1,44 @@
...
+contain several objects that are not connected to each other.
+
+The code snippet below demonstrates two ways to remove objects inside an image,
+either by removing objects
+
+- based on the number of samples that make up each object
+- or based on their distance of each object to each other.
```
stefanv, line 15:
```suggestion
The example shows how to remove objects based on their size, or their distances from other objects.
```

skimage/morphology/_misc_cy.pyx
```diff
@@ -0,0 +1,122 @@
...
+def _remove_near_objects(
+    np_anyint[::1] out not None,
+    Py_ssize_t[::1] border_indices not None,
+    Py_ssize_t[::1] inner_indices not None,
+    kdtree,
+    cnp.float64_t p_norm,
+    cnp.float64_t min_distance,
+    tuple shape,
+):
+    """Remove objects until a minimal distance is ensured.
```
stefanv, line 22:
```suggestion
    """Remove objects, starting with the smallest, until objects are a minimum distance apart.
```

skimage/morphology/_misc_cy.pyx
```diff
@@ -0,0 +1,122 @@
...
+
+    Iterates over all objects (pixels that aren't zero) inside an image and
```
stefanv, line 24: What does this mean in the context of a color image?

skimage/morphology/_misc_cy.pyx
```diff
@@ -0,0 +1,122 @@
...
+    removes "nearby" objects until all remaining ones are spaced more than a
+    given minimal distance from each other.
+
+    Parameters
+    ----------
+    out :
```
stefanv, line 30: type not given?

stefanv: Also, why is "out" the first parameter?

skimage/morphology/_misc_cy.pyx
```diff
@@ -0,0 +1,122 @@
...
+        An array with labels for each object in `image` matching it in shape.
+    border_indices, inner_indices :
+        Indices into `out` for the border of objects (`border_indices`) and
```
stefanv, line 33: I cannot derive, from this description, what these are or what they do.

skimage/morphology/_misc_cy.pyx
```diff
@@ -0,0 +1,122 @@
...
+        the inner part of objects (`inner_indices`). `border_indices`
+        determines the iteration order; objects that are indexed first are
+        preserved. Indices must be sorted such, that indices pointing to the
+        same object are next to each other.
+    kdtree : scipy.spatial.cKDTree
```
stefanv, line 38: Ah, realized that this couldn't possibly be a public function, so take the above with a grain of salt.

skimage/morphology/meson.build
```diff
@@ -12,6 +12,7 @@ extensions = [
   '_flood_fill_cy',
   '_grayreconstruct',
   '_max_tree',
+  '_misc_cy',
```
stefanv, line 15: Best name?

skimage/morphology/misc.py
```diff
@@ -90,6 +94,10 @@ def remove_small_objects(ar, min_size=64, connectivity=1, *, out=None):
     out : ndarray, same shape and type as input `ar`
         The input array with small connected components removed.
 
+    See Also
+    --------
+    skimage.morphology.remove_near_objects
```
stefanv, line 99: I don't have a better name, so I'll just put this out there in case you have other ideas: but `remove near objects` isn't super intuitive, since we're not removing objects near to something else. The *types* of names that might capture it, but that doesn't fit nicely into the `remove_*_objects` mold, are:

- `reduce_object_density`
- `filter_objects` (could enclose both by size and by distance?)
- `sparsify_objects`

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+    min_distance,
+    *,
+    priority=None,
+    p_norm=2,
+    spacing=None,
+    out=None,
+):
+    """Remove objects until a minimal distance is ensured.
+
+    Iterates over all objects (pixels that aren't zero) inside an image and
```
stefanv, line 259: As above, I think this needs a more intuitive explanation. Are neighboring values that are equal considered an object?

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+    removes "nearby" objects until all remaining ones are spaced more than a
+    given minimal distance from each other.
+
+    Parameters
+    ----------
+    label_image : ndarray of integers
+        An n-dimensional array containing object labels, e.g. as returned by
+        :func:`~.label`. A value of zero is considered background, all other
```
stefanv, line 267: As an aside, have you looked into the working of label? It builds a graph of the graph, and creates unions of those graphs. It sounds a lot like what you are doing in the internal function above.

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+        object IDs must be positive integers.
+    min_distance : int or float
+        Remove objects with lower priority whose distance is not greater than
```
stefanv, line 270: Priority is undefined at this point; try and avoid forward references, unless you make them explicit.

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+        this positive value.
+    priority : ndarray, optional
+        Defines the priority with which objects that are too close to each other
+        are removed. Expects a 1-dimensional array of length
+        :func:`np.amax(label_image) + 1 <numpy.amax>` that contains the priority
+        for each object ID at the respective index. Objects with a lower value
+        are removed first until all remaining objects fulfill the distance
+        requirement. If not given, priority is given to objects with a higher
```
stefanv, line 278: Why is ID involved? And is ID used to break ties only?

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+        number of samples and their ID second.
+    p_norm : int or float, optional
+        The Minkowski p-norm, used to calculate the distance between objects.
```
stefanv, line 281:
```suggestion
        The Minkowski distance of order p, used to calculate the distance between objects.
```

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+    See Also
+    --------
+    skimage.morphology.remove_small_objects
+        Remove objects smaller than the specified size.
+
+    Notes
+    -----
+    The basic steps of this algorithm work as follows:
+
+    1. If `priority` is not given, use :func:`numpy.bincount` as a fallback.
```
stefanv, line 306: I think this is a note, rather than "step 1".

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+    2. Find the indices for of all given objects and separate them depending on
+       if they point to an object's border or not.
+    3. Sort indices by their object ID, ensuring all indices pointing to the
```
stefanv, line 309:
```suggestion
    3. Sort indices by their object ID, ensuring that indices which point to the
```

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+       same object are next to each other. This optimization allows finding
+       all parts of an object, simply by stepping to the neighboring indices.
+    4. Sort boundary indices by priority. Use a stable-sort to preserve the
+       contiguousness from the previous sorting step.
```
stefanv, line 313:
```suggestion
       ordering from the previous sorting step.
```

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+    5. Construct a kd-tree with :func:`scipy.spatial.cKDTree` from the
```
stefanv, line 314:
```suggestion
    5. Construct a :func:`scipy.spatial.cKDTree <kd-tree>` from the
```

Not sure if this is the right syntax, but you get the idea.

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+       boundary indices.
+    6. Iterate all boundary indices in the sorted order, and query the kd-tree
```
stefanv, line 316:
```suggestion
    6. Iterate across boundary indices in priority sorted order, and query the kd-tree
```

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+       for objects that are too close. Remove ones that are and skip them on
```
stefanv, line 317: "skip them on"? is the KD-tree updated upon removal?

skimage/morphology/misc.py
```diff
@@ -235,3 +243,208 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+       during the iteration.
+
+    The performance of this algorithm benefits significantly from reducing the
```
stefanv, line 320: What is the actionable tip you want to provide here?

stefanv (commented):
Nice implementation!

I left mostly comments around documentation. And a general question: would it make sense to combine object removal by size into a more general `filter_objects`?

stefanv:
Emscripten failure: https://github.com/scikit-image/scikit-image/actions/runs/9453937550/job/26040373784?pr=4165#step:8:548

That is NumPy 1.26, which I think we also test on in other targets :thinking:

lagru:
@stefanv, I think I addressed most of your concerns. Sadly I don't have a clue what the [pyodide specific failures](https://github.com/scikit-image/scikit-image/actions/runs/9457279531/job/26050781729?pr=4165#step:8:1845) are about. Also don't know an easy way to debug this locally yet.

lagru:
Hey @agriyakhetarpal, could I perhaps summon you in case you have an idea at a first glance? I/we can't figure out, why numpy is acting up for the emscripten job but nowhere else.

agriyakhetarpal:
Hi, @lagru – thanks for mentioning me, and please feel free to mention me on further issues that can arise; I fully intend to continue maintenance and help out after establishing this workflow. :)

I had a cursory look and the issue seems to be something that I have encountered before. A stopgap-like fix is to convert the data to `int32` before passing it to `np.bincount()` (maybe under `is_wasm()` only too). The bitness of the Pyodide interpreter is 32 bits, which is the most likely producer of the problem.

stefanv:
```ipython
Cell In[3], line 2
      1 x = np.ones((3,), dtype=np.int64)
----> 2 np.bincount(x)

TypeError: Cannot cast array data from dtype('int64') to dtype('int32') according to the rule 'safe'
```

@seberg I believe this is a bug in NumPy on 32-bit platforms? In the mean time, we can work around it with a special `is_wasm` clause.

seberg:
Not sure it is a bug, but `bincount` is indexing, so 64bit integers don't fit anyway, since you are limited to 32bit array sizes.  Would it make sens to use `intp` for the array?  Or just add `.astype(np.intp, copy=False)`, I guess if you cannot change that.

stefanv:
> Not sure it is a bug, but `bincount` is indexing, so 64bit integers don't fit anyway, since you are limited to 32bit array sizes. Would it make sens to use `intp` for the array? Or just add `.astype(np.intp, copy=False)`, I guess if you cannot change that.

Thanks for looking at this, Sebastian. The fix isn't hard, but NumPy does not document the intp requirement as far as I can tell, nor does it throw a helpful error message.

skimage/morphology/misc.py
```diff
@@ -235,3 +243,212 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+    # simple iteration up or down the index!
+    border_indices = border_indices[np.argsort(out_raveled[border_indices])]
+    inner_indices = inner_indices[np.argsort(out_raveled[inner_indices])]
+
+    if priority is None:
+        if out_raveled.dtype.itemsize > np.intp().itemsize:
+            # bincount expects intp (e.g. 32-bit on WASM), so down-cast to that
+            priority = np.bincount(out_raveled.astype(np.intp, copy=False))
+        else:
+            priority = np.bincount(out_raveled)
```
lagru, line 408: Okay, I tried to address the WASM-bincount failure in https://github.com/scikit-image/scikit-image/pull/4165/commits/e6446d6fd2b2674ab361de344b2172323f673fd1. Not sure if this is the best approach though. I wanted to avoid creating unnecessary up-casting of int8 etc.

lagru: Thanks for chiming in @agriyakhetarpal & @seberg.

lagru: That didn't work. https://github.com/scikit-image/scikit-image/pull/4165/commits/d749dd316d40fd7d1ca71c74d170fd16b2c5143d works though.

stefanv: Out of curiosity, what if the values are too large to be casts down to int32?

jni: I'm confused — aren't there non-wasm platforms where this would be an issue? Or is it because we no longer support 32-bit win or linux that we can get away with just is_wasm?

skimage/morphology/misc.py
```diff
@@ -235,3 +244,212 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+
+    if spacing is not None:
+        spacing = np.array(spacing)
+        if spacing.shape != (out.ndim,) or spacing.min() <= 0:
+            raise ValueError(
+                "`spacing` must contain exactly one positive factor "
+                "for each dimension of `label_image`"
+            )
+
+    indices = np.nonzero(out_raveled)[0]
```
jni, line 376:
```suggestion
    indices = np.flatnonzero(out_raveled)
```

Unless this is deprecated in np2, but I don't think so...?

skimage/morphology/misc.py
seberg: Yeah, this just force casts which is unfortunately something that indexing-like operations do in NumPy, so if you manage to end up with a too large value, that could silently do the wrong (although I guess it's likely that even after the force cast you have huge values that cause an out of memory error).

If floats might come around using `casting="same_kind"` would be better; you can do this unconditionally really, that is what the `copy=False` is for.

skimage/morphology/_misc_cy.pyx
```diff
@@ -0,0 +1,122 @@
...
+                # If neighbor ID wasn't already removed or is the current one
+                # remove the boundary and remember the ID
+                _remove_object(out, border_indices, j_indices)
+                remembered_ids.add(other_id)
+
+    # Delete inner parts of remembered objects
+    for j_indices in range(inner_indices.shape[0]):
+        object_id = out[inner_indices[j_indices]]
+        if object_id != 0 and object_id in remembered_ids:
+            _remove_object(out, inner_indices, j_indices)
```
jni, line 81: Is this fast?

I have an alternate implementation for this section that may be worth exploring:

(1) build a coo matrix -> csr matrix mapping object ids (rows) to raveled indices (columns). This can be done with:

```python
sz = label_image.size
coo = sparse.coo_matrix(
    (np.ones(sz), (label_image.ravel(), np.arange(sz))),
    shape=(np.max(label_image), sz + 1),
)
csr = coo.tocsr()
```

(Bonus: `csr.sum(axis=1)` is fast and gives you the bincount)

(2) Finding all the indices belonging to label i is then O(1) and they are adjacent in memory: `idxs = csr.indices[csr.indptr[i]:csr.indptr[i+1]]`

jni: (Indeed I expect you should be able to iterate over removed indices in Python rather than Cython without much performance cost, since typically the size of the indices is much greater than the number of labels to be removed.)

jni: Indeed re-reading the code, it seems you are iterating over all indices and checking whether they are in the remembered objects — even if the object is a set and O(1) lookup, this seems less efficient than only iterating over remembered object ids and instantly recalling the relevant indices. I think?

skimage/morphology/misc.py
```diff
@@ -235,3 +244,212 @@ def remove_small_holes(ar, area_threshold=64, connectivity=1, *, out=None):
...
+    if (spacing is None or np.all(spacing[0] == spacing)) and p_norm <= 2:
+        # For unity spacing we can make the borders more sparse by using a
+        # lower connectivity
+        footprint = ndi.generate_binary_structure(out.ndim, 1)
+    else:
+        footprint = ndi.generate_binary_structure(out.ndim, out.ndim)
+    border = (
+        ndi.maximum_filter(out, footprint=footprint)
+        != ndi.minimum_filter(out, footprint=footprint)
+    ).ravel()[indices]
```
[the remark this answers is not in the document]
lagru: I can't yet. I wasn't sure if there are edge cases for `spacing` and `p_norm` where the using a higher-connected footprint would return the wrong results. So I opted to only use the lower connectivity for cases were I was sure that it was correct.

skimage/morphology/misc.py
lagru: Oh, nice. I think it's not.

skimage/morphology/_misc_cy.pyx
lagru: Sounds interesting! Let me explore this once I find the time!

skimage/morphology/_misc_cy.pyx
lagru: Sorry, I just closed this because @stefanv realized later on that this was an internal function during review (which he commented). Nevertheless, I should have given an update anyway. 

The type is not given, because this is a typed Cython function already and I didn't want them to be redundant and potentially out of date.

"out" is first, because it made sense to me. It's the only value that is modified and all other parameters are just there to help with that.

skimage/morphology/misc.py
lagru: Upon in-person discussion we settled for `remove_objects_by_distance`. It's not ideal but we feel like this makes it reasonably clear that the function deals with distances between objects instead of the property of a single one. 

The idea is to also rename `remove_small_objects` to `remove_objects_by_size` for a similar reason. 1) "small" isn't really a good term to use in this context; what is small? And 2) to keep things consistent.

skimage/morphology/misc.py
lagru: I reworded this. Hope that's what you had in mind.

skimage/morphology/misc.py
lagru: I reworded this. Hope that's what you had in mind.

skimage/morphology/_misc_cy.pyx
lagru: Hmm, is there a typo in the code you are suggesting? Running 

<details><summary>Details</summary>

```python
import numpy as np
import scipy as sp
label_image = np.array(
    [[8, 0, 0, 0, 0, 0, 0, 0, 0, 9, 9],
     [8, 8, 8, 0, 0, 0, 0, 0, 0, 9, 9],
     [0, 0, 0, 0, 0, 0, 0, 0, 9, 0, 0],
     [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
     [0, 0, 3, 0, 0, 0, 1, 0, 0, 0, 0],
     [2, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0],
     [0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0],
     [0, 0, 0, 0, 0, 0, 0, 0, 0, 7, 7]]
)
sz = label_image.size
coo = sp.sparse.coo_matrix(
    (np.ones(sz), (label_image.ravel(), np.arange(sz))),
    shape=(np.max(label_image), sz + 1),
)
csr = coo.tocsr()
```
</details> 

gives 
```
ValueError: axis 0 index 9 exceeds matrix dimension 9
```
when calling `coo_matrix`...?

jni: yes, sorry:

```
shape=(np.max(label_image) + 1, sz),
```

😅 

btw this is the same idea as #4855

lagru: Thanks, I managed to get it working with

<details><summary>Details</summary>

```patch
Subject: [PATCH] Use sparse matrix to remove objects
---
Index: skimage/morphology/_misc_cy.pyx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/skimage/morphology/_misc_cy.pyx b/skimage/morphology/_misc_cy.pyx
--- a/skimage/morphology/_misc_cy.pyx	(revision d749dd316d40fd7d1ca71c74d170fd16b2c5143d)
+++ b/skimage/morphology/_misc_cy.pyx	(date 1718287031000)
@@ -18,6 +18,7 @@
     cnp.float64_t p_norm,
     cnp.float64_t min_distance,
     tuple shape,
+    csr,
 ):
     """Remove objects, in specified order, until remaining are a minimum distance apart.
 
@@ -70,15 +71,9 @@
             other_id = out[border_indices[j_indices]]
             if other_id != 0 and other_id != object_id:
                 # If neighbor ID wasn't already removed or is the current one
-                # remove the boundary and remember the ID
-                _remove_object(out, border_indices, j_indices)
-                remembered_ids.add(other_id)
+                # remove the boundary
+                out.base[csr.indices[csr.indptr[other_id]:csr.indptr[other_id + 1]]] = 0
 
-    # Delete inner parts of remembered objects
-    for j_indices in range(inner_indices.shape[0]):
-        object_id = out[inner_indices[j_indices]]
-        if object_id != 0 and object_id in remembered_ids:
-            _remove_object(out, inner_indices, j_indices)
 
 
 cdef inline _remove_object(
Index: skimage/morphology/misc.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/skimage/morphology/misc.py b/skimage/morphology/misc.py
--- a/skimage/morphology/misc.py	(revision d749dd316d40fd7d1ca71c74d170fd16b2c5143d)
+++ b/skimage/morphology/misc.py	(date 1718286735125)
@@ -2,6 +2,7 @@
 
 import numpy as np
 import functools
+import scipy as sp
 from scipy.spatial import cKDTree
 
@@ -410,9 +411,16 @@
     # `priority` can only be indexed by positive object IDs,
     # `border_indices` contains all unique sorted IDs so check the lowest / first
     smallest_id = out_raveled[border_indices[0]]
+    largest_id = out_raveled[border_indices[-1]]
     if smallest_id < 0:
         raise ValueError(f"found object with negative ID {smallest_id!r}")
 
+    coo = sp.sparse.coo_matrix(
+        (np.ones(out.size), (out_raveled, np.arange(out.size))),
+        shape=(largest_id + 1, out.size)
+    )
+    csr = coo.tocsr()
+
     try:
         # Sort by priority second using a stable sort to preserve the contiguous
         # sorting of objects. Because each pixel in an object has the same
@@ -447,6 +455,7 @@
         min_distance=min_distance,
         p_norm=p_norm,
         shape=label_image.shape,
+        csr=csr,
     )
 
     if out_raveled.base is not out:

```

</details> 

However, benchmarking with 

```python
import skimage as ski
import scipy as sp
image = ski.data.hubble_deep_field()
image = ski.color.rgb2gray(image)
objects = image > 0.18
label_image, _ = sp.ndimage.label(objects)
%timeit ski.morphology.remove_objects_by_distance(label_image, 1)
%timeit ski.morphology.remove_objects_by_distance(label_image, 5)
%timeit ski.morphology.remove_objects_by_distance(label_image, 100)
%timeit ski.morphology.remove_objects_by_distance(label_image, 500)
```

Shows that the current Cython-loop based approach is still faster.

- Current Cython loop
  364 ms ± 12.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
  346 ms ± 5.39 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
  198 ms ± 2.32 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
  236 ms ± 2.83 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
- Sparse-matrix approach
  379 ms ± 6.25 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
  364 ms ± 746 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
  224 ms ± 5.08 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
  263 ms ± 881 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)

:thinking:

lagru: Might still be worth the switch though, as the logic should be way easier to follow...?

jni: > the logic should be way easier to follow...?

Yes, *and*, you're currently doing redundant work, because the csr approach is sorting the indices for you. So you should be able to get rid of the argsort calls. (But it might be a bit tricky to deal with border/inner. But should be doable e.g. with a mask I think.)

lagru: Okay, since this may take me some time to figure out. Do you mind if we added that optimization in another PR? That way we can get this into the next release which we want to do soon. This PR also has kind of a bad track record with getting stalled again because attention shifts again. :sweat_smile: 

cc @jarrodmillman

jni: Sure! Might be worth adding a `# TODO` comment to the code pointing to this discussion, but whatever works for you. 😊

jarrodmillman:
I am merging this and making a 0.24rc0 release. Thanks!

Issue #3980, opened 2019-06-26
threshold_otsu and threshold_multiotsu give different results


Miiiluuu:
## Description
thank you very much for implementing threshold_multiotsu, it is super helpful for my project.
However, I found some inconsistency, using threshold_otsu and threshold_multiotsu with two classes produces different results.

## Way to reproduce
```python
import numpy as np
from skimage.filters import threshold_otsu, threshold_multiotsu

array = np.array([-730.,  371.,  485.,  514.,  640.,  687.,  699.,  825.,  828., 1025., 1082., 1129., 1233.])
thresh = threshold_multiotsu(array, 2)
print(f"Multiotsu: {thresh[0]}")
# 830.43164062

thresh = threshold_otsu(array)
print(f"Singleotsu: {thresh}")
# -726.166015625

```

## Version information

```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.6.3 (v3.6.3:2c5fed8, Oct  3 2017, 18:11:49) [MSC v.1900 64 bit (AMD64)]
Windows-10-10.0.17134-SP0
scikit-image version: 0.16.dev0
numpy version: 1.16.4
```

alexandrejaguar:
Hi @Miiiluuu,
thank you very much for your bug report. I am trying to identify what could cause this effect, and I'll get back to you later, OK?

sciunto:
With our images, I found two that replicate the issue

```
names = ('camera', 'moon', 'coins', 'text', 'clock', 'page')

for name in names:
    img = getattr(data, name)()
    otsu = threshold_otsu(img)
    multi = threshold_multiotsu(img, 2)[0]
    if otsu != multi:
        print(name, otsu, multi)
```

> moon 87 89



```
names = ('chelsea', 'coffee', 'astronaut', 'rocket')

from skimage.color import rgb2gray

for name in names:
    img = rgb2gray(getattr(data, name)())
    otsu = threshold_otsu(img)
    multi = threshold_multiotsu(img, 2)[0]
    if otsu != multi:
        print(name, otsu, multi)
```

> astronaut 0.388671875 0.439453125

hmaarrfk:
[...]

PR #4167, opened 2019-09-16
Refactor/optimize threshold_multiotsu

rfezzani:
## Description

This PR basically consists in getting rid of python for loops in the `threshold_multiotsu` function.

This optimization improves dramatically the performance of this function when the number of classes is less then 5:

### Setup

```python
In [1]: from skimage.filters import threshold_multiotsu, threshold_otsu                                 

In [2]: from skimage.data import moon                                                                   

In [3]: img = moon()                              

In [4]: %timeit threshold_otsu(img)                                                                     
1.21 ms ± 444 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

```

### Before optimization

```python

In [5]: %timeit threshold_multiotsu(img, 2)                                                             
62.4 ms ± 1.21 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [6]: %timeit threshold_multiotsu(img, 3)                                                             
61.8 ms ± 918 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit threshold_multiotsu(img, 4)                                                             
79.9 ms ± 1.51 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit threshold_multiotsu(img, 5)                                                             
1.39 s ± 42.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

### After optimization

```python

In [5]: %timeit threshold_multiotsu(img, 2)                                                             
2.07 ms ± 3.44 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: %timeit threshold_multiotsu(img, 3)                                                             
2.27 ms ± 3.96 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [7]: %timeit threshold_multiotsu(img, 4)                                                             
19.4 ms ± 3.84 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [8]: %timeit threshold_multiotsu(img, 5)                                                             
1.3 s ± 704 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

This optimization make the computation time of `threshold_multiotsu` comparable with `threshold_otsu`. This is may be an opportunity to deprecate the actual implementation of `threshold_otsu` and replace it by a call to  `threshold_multiotsu` with `classes=2` to fix the issue #3980.

## For reviewers

  later.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1158,10 +1158,10 @@ def threshold_multiotsu(image, classes=3, nbins=256):
 
     momP[2:, 2:][upper_tri] = (momP[1, 2:][upper_tri[1]]
                                - np.repeat(momP[1, 1:-2],
-                                           np.arange(nbins-3, 0, -1)))
+                                           np.arange(nbins - 3, 0, -1)))
     momS[2:, 2:][upper_tri] = (momS[1, 2:][upper_tri[1]]
                                - np.repeat(momS[1, 1:-2],
-                                           np.arange(nbins-3, 0, -1)))
+                                           np.arange(nbins - 3, 0, -1)))
 
     # step 4: calculating the between class variance.
     var_btwcls = np.zeros_like(momP)
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1141,35 +1141,32 @@ def threshold_multiotsu(image, classes=3, nbins=256):
     # histogram ignores nbins for integer arrays.
     nbins = len(bin_centers)
 
-    # defining arrays to store the zeroth (momP, cumulative probability)
-    # and first (momS, mean) moments, and the variance between classes
-    # (var_btwcls).
-    momP, momS, var_btwcls = [np.zeros((nbins, nbins)) for n in range(3)]
+    # Compute the zeroth (momP, cumulative probability) and first
+    # (momS, mean) moments
 
     # building the lookup tables.
     # step 1: calculating the diagonal.
-    for u in range(1, nbins):
-        momP[u, u] = prob[u]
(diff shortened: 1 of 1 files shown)


alexdesiqueira:
@rfezzani thank you for your help! I will check this PR in a few, ok? It looks promising!

> This is may be an opportunity to deprecate the actual implementation of threshold_otsu and replace it by a call to threshold_multiotsu with classes=2 to fix the issue #3980.

I guess this would not solve the issue... `threshold_multiotsu` is having some issues on negative images, and I'm studying how to solve it. It would be good to maintain the two algorithms; `threshold_otsu` can be used to fine-tune `threshold_multiotsu`, when we have issues like that.

rfezzani:
Thank you @alexdesiqueira and @sciunto for your quick review. I added a small comment for the step 3 as requested ;-)

sciunto:
Let's just wait for the tests.

rfezzani:
I based my suggestion of deprecating `threshold_otsu` actual implementation on the comment of @hmaarrfk in #3980:
> I found that there often exists two points that give the same result mathematically. The one that would be chosen would be implementation specific. I had to corner case a few of the points in the test.

This suggests that both results are valid...
But it OK to maintain both versions, this just means that it will be hard to close #3980 :disappointed:

sciunto:
I propose to continue the discussion on a possible merge of the two functions in #3980 with the additional information that performances are similar and @hmaarrfk comment.

jni:
I missed the boat on reviewing this one, but it's gorgeous! Thanks @rfezzani!

rfezzani:
Thx @jni ;-)

PR #4172, opened 2019-09-17
Handle case in which NamedTemporaryFile fails

### LINKED ISSUE #3785: IO:  unhandled exception, in case the URL cannot be opened
## Description
image

## Way to reproduce
```python
def url_to_image(url):
    """download image from url and return it"""
    im_mat = io.imread(url)
    image = cv2.imdecode(im_mat, cv2.IMREAD_COLOR)
 
    # return the image
    return image
```

## Version information
```python
3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)]
Windows-10-10.0.17134-SP0
scikit-image version: 0.14.2
```

```python
  File "C:\Python37\lib\site-packages\skimage\io\util.py", line 28, in file_or_url_context
    with tempfile.NamedTemporaryFile(delete=False, suffix=ext) as f:
  File "C:\Python37\lib\tempfile.py", line 547, in NamedTemporaryFile
    (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
  File "C:\Python37\lib\tempfile.py", line 258, in _mkstemp_inner
    fd = _os.open(file, flags, 0o600)
OSError: [Errno 22] Invalid argument: 'C:\\Users\\kelvin\\AppData\\Local\\Temp\\tmpegjt2y8r.png?mi_guid=CFCDF62B5115D31389F2893D3CCFFA5CB42BDE7F8CD696BB57B97B8BB567940F&mi_language=EN'

During handling of the above exception, another exception occurred:

  File "imagecheck.py", line 28, in <module>
    print(url_to_image("http://www.movable-ink-6437.com/p/rp/0fd853751b1edbb0.png?mi_guid=CFCDF62B5115D31389F2893D3CCFFA5CB42BDE7F8CD696BB57B97B8BB567940F&mi_language=EN"))
  File "imagecheck.py", line 9, in url_to_image
    im_mat = io.imread(url)
  File "C:\Python37\lib\site-packages\skimage\io\_io.py", line 61, in imread
    with file_or_url_context(fname) as fname:
  File "C:\Python37\lib\contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "C:\Python37\lib\site-packages\skimage\io\util.py", line 34, in file_or_url_context
    os.remove(f.name)
UnboundLocalError: local variable 'f' referenced before assignment
```

jni:
Fixes #3785

Currently I just re-raise the exceptions, which I think is the least ambiguous way of handling them. Other alternatives welcome!

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/io/tests/test_io.py
+++ b/skimage/io/tests/test_io.py
@@ -67,6 +67,16 @@ def _named_tempfile_func(error_class):
     -------
     named_temp_file : callable
         A function that always raises the desired error.
+
+    Notes
+    -----
+    Although this function has general utility for raising errors, it is
+    expected to be used to raise errors that ``tempfile.NamedTemporaryFile``
+    from the Python standard library could raise. As of this writing, these
+    are ``FileNotFoundError``, ``FileExistsError``, ``PermissionError``, and
+    ``BaseException``. See
+    `this comment <https://github.com/scikit-image/scikit-image/issues/3785#issuecomment-486598307>`__
+    for more information.
     """
     def named_temp_file(*args, **kwargs):
         raise error_class()
--- a/skimage/io/tests/test_io.py
+++ b/skimage/io/tests/test_io.py
@@ -1,4 +1,5 @@
 import os
+import tempfile
 
 import numpy as np
 from skimage import io, data_dir
@@ -52,3 +53,39 @@ def test_imread_http_url(httpserver):
     # by extension
     image = io.imread(httpserver.url + '/test.jpg' + '?' + 's' * 266)
     assert image.shape == (1, 1)
+
+
+def _named_tempfile_func(error_class):
+    """Create a mock function for NamedTemporaryFile that always raises.
+
+    Parameters
+    ----------
+    error_class : exception class
+        The error that should be raised when asking for a NamedTemporaryFile.
+
+    Returns
+    -------
+    named_temp_file : callable
+        A function that always raises the desired error.
+    """
+    def named_temp_file(*args, **kwargs):
+        raise error_class()
+    return named_temp_file
+
+
+@testing.parametrize(
+    'error_class', [
+        FileNotFoundError, FileExistsError, PermissionError, BaseException
+    ]
+)
+def test_failed_temporary_file(monkeypatch, error_class):
+    # tweak data path so that file URI works on both unix and windows.
+    data_path = data_dir.lstrip(os.path.sep)
+    data_path = data_path.replace(os.path.sep, '/')
+    image_url = 'file:///{0}/camera.png'.format(data_path)
+    with monkeypatch.context():
+        monkeypatch.setattr(
+            tempfile, 'NamedTemporaryFile', _named_tempfile_func(error_class)
+        )
+        with testing.raises(error_class):
+            image = io.imread(image_url)
--- a/skimage/io/util.py
+++ b/skimage/io/util.py
@@ -1,5 +1,6 @@
 import urllib.parse
 import urllib.request
+from urllib.error import URLError, HTTPError
 
 import os
 import re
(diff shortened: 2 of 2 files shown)


emmanuelle:
Is it possible / useful to add a test checking that a meaningful error is raised?

jni:
@emmanuelle I really don't know how... Possibly through mocking NamedTemporaryFile? Help appreciated here!

sciunto:
For example, for the network error, the only way I know is to use unittest.mock (pip install mock) with [side_effect](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock) like `side_effect=HTTPError('oops')`

emmanuelle:
@jni no better idea than the one suggested by @sciunto (which I wasn't aware of). My question was genuinely open :-). We can also merge as it is if you don't have the bandwidth for this at the moment.

jni:
@emmanuelle @sciunto fine, you called my lazy bluff, you win. =D

Actually I'm pretty happy to have learned this. =D As always, being part of scikit-image keeps pushing me to be a better developer. :pray:

jni:
@scikit-image/core this is ready for review. Would be nice to get it in before 0.16!

skimage/io/tests/test_io.py
```diff
@@ -52,3 +53,39 @@ def test_imread_http_url(httpserver):
...
+
+    Parameters
+    ----------
+    error_class : exception class
+        The error that should be raised when asking for a NamedTemporaryFile.
+
+    Returns
+    -------
+    named_temp_file : callable
+        A function that always raises the desired error.
```
alexdesiqueira: I would put a list here, shows what kinds of errors we're looking for (which I'm assuming to be `FileNotFoundError`, `FileExistsError`, `PermissionError`, `BaseException`). Is that right @jni? :)

skimage/io/util.py
```diff
@@ -28,7 +29,15 @@ def file_or_url_context(resource_name):
                 f.write(u.read())
             # f must be closed before yielding
             yield f.name
-        finally:
+        except (URLError, HTTPError):
+            # could not open URL
+            os.remove(f.name)
+            raise
+        except (FileNotFoundError, FileExistsError,
```
stefanv: Shouldn't this be a catch-all?

stefanv: Perhaps having BaseException makes it so?

jni: BaseException probably makes it so, but that is only there because it is specifically raised by NamedTemporaryFile. Perhaps you are right that it should be catch-all, as there might be things called by NamedTemporaryFile that might raise in other, unexpected ways. Either way, I think catch-all should be a separate clause.

stefanv: I'd usually agree, except that the existing clause is empty, so it doesn't make much of a difference.  Either way is fine, though.

jni: Actually, here's the reason I don't like a bare clause: in the case of an unexpected exception, it's unclear whether we should `os.remove(f.name)`, because it's unclear whether the exception was raised by `NamedTemporaryFile()`, `urlopen`, or `f.write`. So, I'd rather not add an `except` clause...

One possible option:

```
finally:
    if 'f' in locals() and os.path.exists(f):
        os.remove(f.name)
```

But, this still feels brittle. Overall I think we should just leave this be until someone complains. =)

stefanv: Sounds then like another try/except is required to specialize what is being caught.

stefanv:
@jni Perhaps just wait until the discussion is completed before merging; I know you are in a hurry to get the release out the door :)

jni:
@stefanv Huh? The discussion *was* completed, with two +1s. Your first comment was after the merge...?

jni:
at any rate, what is there now may not be perfect, but it's a heck of a lot better than what was in master, so this merge does not seem like a bad thing to me.

stefanv:
Sorry, got my wires crossed there! I agree that this merge was an improvement.

PR #4178, opened 2019-09-23
Refactor/fix threshold_multiotsu

### LINKED ISSUE #3975: Divergent results in threshold_multiotsu with 16bits images
## Description
Hello,
I was testing the function with tomograph images, 16 bit (uint16), and memory error occurred:

`[...] scikit-image\skimage\filters\thresholding.py", line 1107, in <listcomp>
 momP, momS, var_btwcls = [np.zeros((nbins, nbins)) for n in range(3)]
MemoryError`

One solution found was to modify [these lines](https://github.com/scikit-image/scikit-image/blob/master/skimage/filters/thresholding.py#L1093-L1096) to `hist, bin_centers = np.histogram (image.ravel (), bins = nbins)` and `prob = hist / len (bin_centers)`, that is, use the default function in numpy to calculate the histogram.

Despite this, there is another (more important) problem where the threshold values found differ from those found in the SimpleITK and MatLAB libraries.

An example using a single image, I received the following values (3 classes, 2 thresholds):

SimpleITK: (12443, 29469)
MatLAB: (12491, 29802)
Skimage: (26629, 33177)

I was unable to upload the test image, but it is available at this link: [https://www.digitalrocksportal.org/projects/35/origin_data/85/](https://www.digitalrocksportal.org/projects/35/origin_data/85/)

 (note: although it says that the images are 8-bit, they are actually 16-bit).

## Way to reproduce
```python
# code to reproduce MemoryError

from skimage.filters import threshold_multiotsu
import numpy as np
import pytiff

names = ['AT CARB_0.7 Micron scan_recon_Export0001.tiff']

imgs = np.empty((1, 1008, 980), np.uint16)
for i, fpath in enumerate(names):
    with pytiff.Tiff(fpath) as handle:
        for page in handle.pages:
            imgs[i] = page[:]

print(threshold_multiotsu(imgs))
```

## Version information
```python
```

```python
# your output here
3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)]
Windows-10-10.0.17134-SP0
scikit-image version: 0.16.dev0
numpy version: 1.16.2
```

### LINKED ISSUE #3980: threshold_otsu and threshold_multiotsu give different results

rfezzani:
## Description

Closes #3975, closes #3980.

In this PR, `threshold_multiotsu` is reimplemented and fixed:

- this implementation requires less memory (only `var_btwcls` and the first row of `momP` and `momS` are allocated)
- `threshold_multiotsu` and `threshold_otsu` give the same results on `skimage.data` images (see `test_bimodal_multiotsu_hist` test)
- `uint16` images are now managed by disabling the LUT approach
- a `ValueError` is raised if the input image contains less values then requested classes
- A warning is raised if the input image is suspected to be RGB (as in `threshold_otsu`)

This implementation is fast:

```python
In [1]: from skimage.filters import threshold_multiotsu, threshold_otsu                                  

In [2]: from skimage.data import moon                                                                    

In [3]: img = moon()                                                                                     

In [4]: %timeit threshold_otsu(img)                                                                      
1.21 ms ± 385 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: %timeit threshold_multiotsu(img, 2)                                                              
905 µs ± 2.06 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [6]: %timeit threshold_multiotsu(img, 3)                                                              
1.8 ms ± 856 ns per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [7]: %timeit threshold_multiotsu(img, 4)                                                              
83.4 ms ± 22.7 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit threshold_multiotsu(img, 5)                                                              
5.08 s ± 20.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

Concerning #3975 use case, the output is now `[12649, 29834]`.
For #3980 use case, this implementation outputs `-718.49804688`. which corresponds to the value at index 1of the bins_centers while `threshold_otsu` outputs `-726.166015625` which corresponds to the index 0 value...

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/filters/tests/test_thresholding.py
```diff
@@ -561,11 +571,26 @@ def test_multiotsu_output():
     for coor, val in zip(coords, values):
         rr, cc = circle(coor[1], coor[0], 20)
         image[rr, cc] = val
-    thresholds = [64, 128]
+    thresholds = [1, 128]
```
sciunto: I do not understand why this value changed. I would have thought that the example was designed to get a threshold between 64 and 128, and another one between 128 and 192.

rfezzani: Hi @sciunto, thank you for your review!
The example image contains 4 values: 0, 64, 128 and 192. Both thresholds are valid:
```python
for ax, thresh in zip(plt.subplots(1, 2)[1], [[1, 128], [64, 128]]): 
    regions = np.digitize(image, thresh) 
    ax.imshow(regions, cmap='Accent') 
    ax.set_title("Thresholds = {}".format(thresh)) 
plt.tight_layout()  
```
Figure_1

sciunto: Then, shouldn't increase the number of bins by one?

rfezzani: After some more investigations, I would say that the actual implementation gives the best result: The returned thresholds [t_0, t_1, ..., t_n] separating the classes C_0, C_1, ...., C_(n+1) are defined such as
- C_0 = {x; x ≤ t_0}
- C_1 = {x; t_0 < x ≤ t_1}
- ....
- C_n+1 = {x; x ≤ t_n}

This means that the `right` parameter of `np.digitise` has to be set to `True`:
```python
for ax, thresh in zip(plt.subplots(1, 2)[1], [[1, 128], [64, 128]]): 
    regions = np.digitize(image, thresh, right=True) 
    ax.imshow(regions, cmap='Accent') 
    ax.set_title("Thresholds = {}".format(thresh)) 
plt.tight_layout()  
```
Figure_3

The threshold [64, 128] merges the pixels values 0 and 64 making the first circle disappear!
With the thresholds set to [1, 128] the values 64 and 128 are considered as one class and no circle are lost!

If the number of classes is set to 4, both implementations return [0, 64, 128] and all the circles are separated!

Figure_2

rfezzani: > Then, shouldn't we increase the number of bins by one?

There is no reason for modifying the output since it verify the expected theoretical result.

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,295 @@ import numpy as np
...
+                                    Py_ssize_t nbins,
+                                    float [::1] P,
+                                    float [::1] S) nogil:
+    cdef cnp.intp_t i
+
+    P[0] = prob[0]
+    S[0] = prob[0]
+    for i in range(1, nbins):
+        P[i] = P[i-1] + prob[i]
+        S[i] = S[i-1] + i*prob[i]
```
Borda: spaces around the operator

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,295 @@ import numpy as np
...
+            return (S[j]**2)/P[j]
+    else:
+        Pij = P[j] - P[i-1]
+        if Pij > 0:
+            Sij = S[j] - S[i-1]
+            return (Sij**2)/Pij
+    return 0
+
+
+cdef float _set_thresh_indices(float[::1] P, float[::1] S,
```
Borda: I would rather see a longer (more descriptive) name then single capital letter

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,295 @@ import numpy as np
...
+
+    idx = 1
+    P[0] = prob[0]
+    S[0] = prob[0]
+    for i in range(1, nbins):
+        P[i] = P[i-1] + prob[i]
+        S[i] = S[i-1] + i*prob[i]
+        if P[i] > 0:
+            H[idx] = (S[i]**2)/P[i]
+        idx += 1
```
Borda: `idx` is not used / neede since `idx == i`

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,295 @@ import numpy as np
...
+
+    for i in range(1, nbins):
+        for j in range(i, nbins):
+            Pij = P[j] - P[i-1]
+            if Pij > 0:
+                Sij = S[j] - S[i-1]
+                H[idx] = (Sij**2)/Pij
+            idx += 1
```
Borda: is `idx` restarted or used the last value from the previous cycle?

skimage/filters/tests/test_thresholding.py
```diff
@@ -536,10 +536,20 @@ def test_niblack_sauvola_pathological_image():
 
 
 def test_bimodal_multiotsu_hist():
-    image = data.camera()
-    thr_otsu = threshold_otsu(image)
-    thr_multi = threshold_multiotsu(image, classes=2)
-    assert thr_otsu == thr_multi
+
```
Borda: why so many spaces?

skimage/filters/tests/test_thresholding.py
rfezzani: Ah, sorry @sciunto, I got you wrong... You meant increasing the number of bins in the test, right? Yes, of course it is a good idea ;-)

skimage/filters/_multiotsu.pyx
rfezzani: I kept the notations of the reference paper while developing, but you are right, we can modify those. What do you think of `first_moment` and `second_moment`?

skimage/filters/_multiotsu.pyx
rfezzani: Good catch, I may set `idx = nbins` after this loop instead of incrementing it.

skimage/filters/_multiotsu.pyx
rfezzani: It uses the last value from the previous cycle since it is the current index in `H`

Borda: but the cycle has fix number f iteration (there is no conditional break)

rfezzani: Nop, we loop over the upper triangular part of the matrix, so `idx = nbins * (nbins + 1) / 2`. The difficulty is that `H` here stores only the upper triangular part  of the look up table.

skimage/filters/tests/test_thresholding.py
sciunto: > Ah, sorry @sciunto, I got you wrong... You meant increasing the number of bins in the test, right? Yes, of course it is a good idea ;-)

Yes, that's what I meant :)

rfezzani: It's done ;-)

skimage/filters/tests/test_thresholding.py
rfezzani: Fixed ;-)

skimage/filters/_multiotsu.pyx
Borda: so you know the value of `idx`

rfezzani: Yes I can compute it for any `i` or `j`, it is what the `_get_var_btwclas_lut` does.

rfezzani: I don't know if it is clear, but `H` here is a 1D view of the upper triangular part of the LUT containing the different variance between classes. Since I only store the upper triangular part, I have to match between the 2D coordinates `i, j` an the 1D coordinate `idx`.
When I am setting `H`, I don't really care about the 2D indices (that's why I am simply incrementing `idx`), it is only when I want to access them that I have to "translate" `i, j` into `idx`. That's what `_get_var_btwclas_lut` does.

jni: @rfezzani that's definitely not immediately clear and probably something that should be written in as comments in the code. =)

rfezzani: You are right, I only exposed this in my comments [1](https://github.com/scikit-image/scikit-image/issues/3975#issuecomment-534889441) and [2](https://github.com/scikit-image/scikit-image/issues/3975#issuecomment-537046656) in the issue #3975, sorry.
I also notice that multiple Cython functions have no docstrings. I will fix that.

rfezzani: I added some doc to Cython function ;-)

sciunto:
@rfezzani We merged a PR to prepare the release that impacts a little this PR. Would you mind adopting the modifications in this PR?

Then, we can proceed and finalize the review. @alexdesiqueira Your feedback would be valuable on this code :)

rfezzani:
HI @sciunto, based on our [discussion](https://github.com/scikit-image/scikit-image/pull/4178#discussion_r330873253), I think that the modifications introduced by the merged PR are no more relevant.
I resolved the conflicts, but did not adopted the modifications because it makes the test fail for this PR. Is it OK for you?

alexdesiqueira:
Hey everyone,
@sciunto @rfezzani I am truly sorry, I wasn't around on the last weeks. I'll check it first thing on Monday, ok?

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,426 @@ import numpy as np
...
+        The Upper triangular part of the between classes variance
+        lookup table.
+    hist_idx : int
+        Current index in the histogram.
+    thresh_idx : int
+        Current index in thresh_indices.
+    nbins : int
         number of bins used in the histogram
-    divisions : int
+    thresh_count : int
```
alexdesiqueira: Now that I'm seeing what this variable represents, its name is not clear at all... couldn't it be something like `classes_count`, or `classes_gen`, according to the documentation?

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,426 @@ import numpy as np
...
         Number of divisions required to generate the desired classes.
-    depth : int
-        Controls the iterations the algorithm had, expanding the interval
-        when _find_best_rec() is called.
+    sigma_max : float
+        Current maximum variance between classes.
+    current_indices: (thresh_count, ) 1D array
```
alexdesiqueira: `current_indices : array`

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,426 @@ import numpy as np
...
+    -------
     max_sigma : float
         Maximum variance between classes.
-    aux_thresh : array
-        Values for multi-Otsu threshold.
+
+    Notes
+    -----
+    For any candidate current_indices {t_0, ..., t_n}, sigma equals
+    var_btwcls[0, t_0] + var_btwcls[t_0+1, t_1] +
```
alexdesiqueira: `...` here instead of the next line, maybe?

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,426 @@ import numpy as np
...
+    ... + var_btwcls[t_(i-1) + 1, t_i] + ... var_btwcls[t_n, nbins-1]
```
alexdesiqueira: A `+` is missing :)

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,426 @@ import numpy as np
...
+
+    References
+    ----------
+    .. [1] Liao, P-S., Chen, T-S. and Chung, P-C., "A fast algorithm for
+           multilevel thresholding", Journal of Information Science and
+           Engineering 17 (5): 713-727, 2001. Available at:
+           <http://ftp.iis.sinica.edu.tw/JISE/2001/200109_01.pdf>
+           :DOI:`10.6688/JISE.2001.17.5.1`
+
```
alexdesiqueira: I think this new line is not necessary. Is it, though? :)

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,426 @@ import numpy as np
...
+    return sigma_max
+
+
+def _get_multiotsu_thresh_indices(float [::1] prob, Py_ssize_t thresh_count):
+    """Finds the indices of Otsu thresholds according to the values
+    occurence probabilities.
+
+    Parameters
+    ----------
+    prob: 1D array
```
alexdesiqueira: `prob : array`
You got the idea, right? :) Please, `var_name<space>:<space>datatype`. 1D arrays would be only `array`.

skimage/filters/_multiotsu.pyx
```diff
@@ -6,101 +6,279 @@ import numpy as np
...
     -------
     max_sigma : float
         Maximum variance between classes.
+
+    References
+    ----------
+    .. [1] Liao, P-S., Chen, T-S. and Chung, P-C., "A fast algorithm for
+           multilevel thresholding", Journal of Information Science and
+           Engineering 17 (5): 713-727, 2001. Available at:
+           <http://ftp.iis.sinica.edu.tw/JISE/2001/200109_01.pdf>
```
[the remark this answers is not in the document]
alexdesiqueira: @sciunto :fearful: you found the DOI! I'm searching for it since 2015! In the wrong places, of course :smile:

skimage/filters/thresholding.py
```diff
@@ -1104,16 +1105,24 @@ def threshold_multiotsu(image, classes=3, nbins=256):
 
     Returns
     -------
-    idx_thresh : array
+    thresh_vals : array
```
alexdesiqueira: Please replace it with `thresh` only, if we're changing this :) Is this a deprecation @jni?

skimage/filters/thresholding.py
```diff
@@ -1104,16 +1105,24 @@ def threshold_multiotsu(image, classes=3, nbins=256):
...
         Array containing the threshold values for the desired classes.
 
+    Raises
+    ------
+    ValueError
+         If ``image`` contains less grayscale value then requested
```
alexdesiqueira: `If ``image`` contains less grayscale values than the desired number of classes.`

skimage/filters/_multiotsu.pyx
alexdesiqueira: I understand the importance of better names for our variables, but in this case, I am kindly against it if we don't refer to the "original" names (variables that appear in the paper) somewhere. The paper does not make it clear to understand what/where the "moments" are, citing them only once as the moments (if I recall correctly). All arguments receive P, S and H in the algorithms.
Therefore, @rfezzani, could you please add a note on the pertinent functions? Something like:
```
zeroth_moment, first_moment, and var_btwcls are referred to as P, S, and H in [1]_.
```
Thanks!

skimage/filters/_multiotsu.pyx
rfezzani: Here `thresh_count` is really the number of returned threshold values, so it is the number of classes  - 1.

alexdesiqueira: Could you fix the documentation for this variable, then? It seems we're talking classes, not thresholds.

rfezzani: It's fixed ;-) thank you again for your review.

skimage/filters/thresholding.py
```diff
@@ -1132,49 +1142,38 @@ def threshold_multiotsu(image, classes=3, nbins=256):
...
+        thresh_idx = np.where(prob > 0)[0][:-1]
+    else:
+        # Get threshold indices
+        try:
+            thresh_idx = _get_multiotsu_thresh_indices_lut(prob, classes - 1)
+        except MemoryError:
+            # Don't use LUT if the number of bins is too large (if the
+            # image is uint16 for example): in this case, the
+            # allocated memory is too large.
+            thresh_idx = _get_multiotsu_thresh_indices(prob, classes - 1)
```
sciunto: I'm unclear about this part. What are the similarities and differences between _get_multiotsu_thresh_indices_lut and _get_multiotsu_thresh_indices. The docstrings are not very informative.

Do we expect the same results to be returned? Is *lut just faster with more memory consumption? Then, can we add tests to check that returned values are identical?

Also, is the try except approach suitable here? It means that using this function on multiple large images will fill up the memory, fail, unload the memory, perform the other calculation... It's not so efficient. Based on the answers of the question I pointed out, I think we can find another approach.

Thank you!

rfezzani: Hi @sciunto,
yes, the result is supposed to be the same. The use of LUT is in fact faster at the price of more memory consumption. This can be added to the doc.
I can in fact also assert in a test that the returned values are identical with both approaches.

The memory will not be filled up: the necessary memory is allocated in `_get_multiotsu_thresh_indices_lut` (for `var_btwcls`, `zeroth_moment` and `first_moment`), the OS is asked for memory allocation, and if there is not enough, the `MemoryError` is raised.

The amount of required memory doesn't depend on the size of the input image but on the number of bins `nbins`. This number is automatically set by the `histogram` function when the input image values are integrals. That's why we face problems with `uint16` images (see #3975).

The amount of available RAM is machine dependent, that's why I think that the `try except` is the best approach.

sciunto: Thanks for the explanations! Then :+1: for adding this to the code :)

rfezzani:
@sciunto, multi-Otsu was already Cython, and the PR is more about fixing then optimizing, but well... :smile:

alexdesiqueira:
Ready when you are, @sciunto @jni

rfezzani:
Thank you all for your review and comments ;-)

jni:
@rfezzani thank you for your hard work here!!! It's a massive improvement! :tada:

alexdesiqueira:
Indeed it is @jni! Thank you very much @rfezzani for these improvements! We (and I personally) really appreciate them!

rfezzani:
Thank you @jni and @alexdesiqueira for your kind words, I am really happy to participate to the skimage adventure ;-)

PR #4180, opened 2019-09-23
DOC: Improve downscale_local_mean() docstring

stuarteberg:
The `downscale_local_mean()` function takes an argument `clip` which is unused.  Looks like it was introduced intentionally, years ago in 0ae8c4a74bc99cb9de7229e961eacee1c7da0840.

Assuming we want to keep it, it would be nice to mention it in the docstring.  While I was at it, I also added a note about the returned dtype.

## For reviewers

  later.


jni:
Thanks @stuarteberg!

PR #4181, opened 2019-09-24
setup.py: Fail gracefully when NumPy is not installed

stefanv:


hmaarrfk:
Cant we just have a try catch as the first line?

stefanv:
Well, no, there are cases in which we don't need NumPy.

hmaarrfk:
Have users recently complained about the inability to install nunpy?

stefanv:
No, this is for the case where they try to install skimage and numpy is missing, so that they get a graceful error message.

hmaarrfk:
Will this automatically install numpy in time to build exceptions?

stefanv:
No, as far as I know that cannot be done.

setup.py
```diff
@@ -33,8 +33,6 @@
 from distutils.command.sdist import sdist
 from distutils.errors import CompileError, LinkError
 
-from numpy.distutils.command.build_ext import build_ext
```
hmaarrfk: Why not have `try: .... Except ImportError: raise RuntimeError('numpy needs to be installed to install scikit-image')`

stefanv:
Because some setup.py commands don't require numpy at all, like egg info.

jni:
Any fixes to the setup.py can come in later PRs. This solves a reported problem and passes CI, so in it goes!

I suspect most of these hacks will get fixed when we go back to pyproject.toml. I think the time for giving that another go is coming soon.

PR #4183, opened 2019-09-25
Fix incorrect resolution date on SKIP1

jni:
## Description

In the rush to merge #4182 before @stefanv's bedtime, I committed the wrong date to the resolution tag for SKIP1 - governance. This fixes that mistake.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -39,7 +39,7 @@ Sections
 
        Documentation for the functions included in scikit-image.
 
-     - `Mission statement <values.html>`_
+   * - `Mission statement <values.html>`_
 
        Our mission, vision and values.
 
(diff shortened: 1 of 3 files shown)


jni:
Note: I incorrectly changed in #4182 the index page to 2x5 instead of 5x2. The second commit fixes that.

jni:
Note some more: for some reason, the docs build fine on my machine, which had sphinx 1.8 but now has 2.2, builds fine on both, but not on Travis. Thankfully, the build error was pretty clear, so the next build should pass... :crossed_fingers:

jni:
Looking good! CC @scikit-image/core would be good to get this merged ASAP as @stefanv's link on the mailing list won't work until this is in master!

doc/source/index.rst
```diff
@@ -39,23 +39,23 @@ Sections
...
 
        How decisions are made in scikit-image.
 
-     - `Installation Steps <install.html>`_
+   * - `Installation Steps <install.html>`_
 
        How to install scikit-image.
 
-   * - `User Guide <user_guide.html>`_
+     - `User Guide <user_guide.html>`_
```
soupault: Wait, why no asterisk here? 🤔

jni: @soupault there should be an asterisk every 2nd heading. Asterisk is the start of a new row.

jni: I only just discovered this myself, hence this element of the PR. :joy:

jni:
@soupault thanks for the approval! :pray:

Issue #4117, opened 2019-08-25
DOC: simple backquotes have no special rendering in the doc


sciunto:
# Work plan

In skimage/

* [ ] draw/      
* [ ] external/  
* [x] filters/  #4127
* [ ] graph/       
* [ ] io/       
* [ ] metrics/     
* [ ] novice/       
* [ ] restoration/  
* [ ] segmentation/  
* [ ] _shared/    
* [ ] util/
* [ ] color/     
* [ ] data/       
* [ ] exposure/  
* [ ] feature/   
* [ ] future/   
* [ ] measure/  
* [ ] morphology/  
* [ ] transform/  
* [ ] viewer/

In doc

## Description

Example:

[Source code](https://github.com/scikit-image/scikit-image/blob/v0.15.0/skimage/color/colorconv.py#L70)
There is one word with simple backquotes and one with double.

Here is the rendering:

https://scikit-image.org/docs/stable/api/skimage.color.html#skimage.color.guess_spatial_dimensions

![Module color — skimage v0 15 0 docs](https://user-images.githubusercontent.com/335370/63651201-97d3da00-c752-11e9-812d-6853af6357c9.png)

The html code is

```
 
 <tr class="field-even field"><th class="field-name">Returns:</th><td class="field-body"><dl class="first docutils">
 <dt><strong>spatial_dims</strong> <span class="classifier-delimiter">:</span> <span class="classifier">int or None</span></dt>
 <dd><p class="first last">The number of spatial dimensions of <cite>image</cite>. If ambiguous, the value
 is <code class="docutils literal notranslate"><span class="pre">None</span></code>.</p>

```

Simple backquotes give `<cite>image</cite>`

jni:
This is a sad but expected difference between restructured text and markdown. The correct formatting is with double backticks. Single backticks are for links. See http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links

sciunto:
Thanks for the pointer. This turns into an issue: fix all the single backticks (I know I used them in PRs...).

lagru:
> This is a sad but expected difference between restructured text and markdown. The correct formatting is with double backticks. Single backticks are for links. See http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links

[...]

PR #4184, opened 2019-09-25
Minor docstring fixes for #4150

jni:
## Description

Additionally, I made mask keyword-only.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_find_contours_cy.pyx
+++ b/skimage/measure/_find_contours_cy.pyx
@@ -25,7 +25,7 @@ def iterate_and_store(double[:, :] array,
     nonzero, high-values pixels are considered to be face+vertex
     connected into objects; otherwise low-valued pixels are.
 
-    Positions where the boolean array mask is False are considered
+    Positions where the boolean array ``mask`` is ``False`` are considered
     as not containing data.
     """
     if array.shape[0] < 2 or array.shape[1] < 2:
--- a/skimage/measure/_find_contours_cy.pyx
+++ b/skimage/measure/_find_contours_cy.pyx
@@ -25,7 +25,7 @@ def iterate_and_store(double[:, :] array,
     nonzero, high-values pixels are considered to be face+vertex
     connected into objects; otherwise low-valued pixels are.
 
-    Positions where the boolean array ``mask`` is ``False`` are considered
+    Positions where the boolean array `mask` is ``False`` are considered
     as not containing data.
     """
     if array.shape[0] < 2 or array.shape[1] < 2:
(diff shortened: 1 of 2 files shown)


skimage/measure/_find_contours_cy.pyx
```diff
@@ -25,7 +25,7 @@ def iterate_and_store(double[:, :] array,
     nonzero, high-values pixels are considered to be face+vertex
     connected into objects; otherwise low-valued pixels are.
 
-    Positions where the boolean array mask is False are considered
+    Positions where the boolean array ``mask`` is ``False`` are considered
```
lagru: Do you deliberately choose to format variables as literals in docstrings? The [convention (section Parameters)](https://numpydoc.readthedocs.io/en/latest/format.html#sections) I know is to emphasize with a single surrounding \`, e.g. \`mask\`.

lagru: Not to block this PR. I'm just curious.

jni: You're correct! I've gotten into the habit of double backticks as they are always safe. Fix coming...

jni: Fixed in cde8c3137

sciunto:
But we say that there is not rendering, and I find this problematic... #4181 4117

jni:
Oops! A bug in the numpydoc docs???

sciunto:
@jni It's not what you said here: https://github.com/scikit-image/scikit-image/issues/4117#issuecomment-524660492 :)

jni:
Yes, that's the Sphinx docs. @lagru points to the NumPy docstring conventions page and they explicitly say, use single-ticks.

sciunto:
I don't know whether it is deliberate on the numpy doc side.

jni:
> I don't know whether it is deliberate on the numpy doc side.

That was my implication when I said "bug" combined with a triple question mark. =P

sciunto:
I open an issue on the numpydoc repository, they can probably shed some light on the mystery. :)

sciunto:
See https://github.com/numpy/numpydoc/issues/233

lagru:
> See numpy/numpydoc#233 

The way I see it this is an issue with the configuration and/or the theme (?) of the skimage docs. Correct? E.g. [SciPy's docs render it correctly](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.argrelextrema.html).

As long as we try to be consistent I'm fine with either solution. Double backticks is probably the better option as long as the docs don't render it correctly.

lagru:
Thanks @jni!

lagru:
> Sure. Until that's fixed, double backticks make more sense. PRs welcome!

There's already #4188.

jni:
... Where you suggested it's not the right solution???

lagru:
Oh, perhaps I worded that badly. I'll try address that.

stefanv:
Single backticks for parameter names, double for code quotes.

PR #4187, opened 2019-09-26
API updates before 0.16

jni:
## Description

Using @Carreau's great tool [frappuccino](https://github.com/Carreau/frappuccino), I went through all the new APIs we are releasing in 0.16 and checked that they were where we want them to be. Mostly it was a matter of making some arguments keyword-only.

The only thing missing right now is in the new `metrics` and `registration` modules, and also in the new `util.compare_images` function, some functions use `im1, im2`, others use `im_true, im_test`, while `compare_images` uses `image1, image2` and the new optical flow uses `image0, image1`. Needless to say this is a mess and we should aim to rectify it. Suggestions welcome! My preference is `reference_image, target_image`.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/histogram_matching.py
+++ b/skimage/exposure/histogram_matching.py
@@ -19,7 +19,7 @@ def _match_cumulative_cdf(source, template):
     return interp_a_values[src_unique_indices].reshape(source.shape)
 
 
-def match_histograms(image, reference, multichannel=False):
+def match_histograms(image, reference, *, multichannel=False):
     """Adjust an image so that its cumulative histogram matches that of another.
 
     The adjustment is applied separately for each channel.
--- a/skimage/measure/tests/test_simple_metrics.py
+++ b/skimage/measure/tests/test_simple_metrics.py
@@ -1,9 +1,7 @@
 import numpy as np
 
 import skimage.data
-from skimage.metrics import peak_signal_noise_ratio
-from skimage.metrics import normalized_root_mse
-from skimage.metrics import mean_squared_error
+from skimage.measure import compare_nrmse, compare_psnr, compare_mse
 
 from skimage._shared import testing
 from skimage._shared.testing import assert_equal, assert_almost_equal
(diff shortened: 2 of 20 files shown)


soupault:
>My preference is reference_image, target_image.

I'd maybe prefer `image_reference` and `image_target` (more convenient with auto-completion). For the same-order arguments (e.g. in some of the comparison/metric functions), `image_0`, `image_1` would probably be fine.

skimage/metrics/_contingency_table.py
```diff
@@ -4,7 +4,7 @@
 __all__ = ['contingency_table']
 
 
-def contingency_table(im_true, im_test, ignore_labels=[], normalize=False):
+def contingency_table(im_true, im_test, *, ignore_labels=[], normalize=False):
```
lagru: Wait isn't using mutable default values bad practice? Doesn't seem like an issue for this particular function but better safe than sorry?

skimage/metrics/_variation_of_information.py
```diff
@@ -8,8 +8,7 @@
 
 def variation_of_information(im_true=None, im_test=None, *, table=None,
                              ignore_labels=[], normalize=False):
```
lagru: Same here.

skimage/exposure/histogram_matching.py
```diff
@@ -19,7 +19,7 @@ def _match_cumulative_cdf(source, template):
     return interp_a_values[src_unique_indices].reshape(source.shape)
 
 
-def match_histograms(image, reference, multichannel=False):
+def match_histograms(image, reference, *, multichannel=False):
```
sciunto: This was released with 0.15, so it is an API change https://scikit-image.org/docs/stable/api/skimage.transform.html#match-histograms

skimage/metrics/_structural_similarity.py
```diff
@@ -9,9 +9,11 @@
 __all__ = ['structural_similarity']
 
 
-def structural_similarity(im1, im2, win_size=None, gradient=False,
-                          data_range=None, multichannel=False,
-                          gaussian_weights=False, full=False, **kwargs):
+def structural_similarity(im1, im2,
```
sciunto: I thought we didn't like im or img in signatures and prefered image. Should we fix here?

skimage/util/compare.py
```diff
@@ -28,9 +28,9 @@ def compare_images(image1, image2, method='diff', *, n_tiles=(8, 8)):
 
     Notes
     -----
-    `diff` computes the absolute difference between the two images.
-    `blend` computes the mean value.
-    `checkerboard` makes tiles of dimension `n_tiles` that display
+    ``'diff'`` computes the absolute difference between the two images.
```
sciunto: We must keep single backquotes now... (I'm responsible for the mess...)

rfezzani:
I have a small suggestion concerning the API of optical flow algorithms: since only two values are allowed for the `dtype` argument (`float32` and `float64`), isn't it better to use a bool argument (`single_precision=True` for example) to specify which floating point precision to use for the estimation?

rfezzani:
Otherwise, both names are fine to me (`image_reference, image_target` or `image_0, image_1`), may be the second is shorter, but less explicit...

lagru:
> since only two values are allowed for the dtype argument (float32 and float64), isn't it better to use a bool argument (single_precision=True for example) to specify which floating point precision to use for the estimation?

Hmm, I think I slightly prefer to keep the dtype argument because it is more flexible. The fact that only two values are allowed is not because it is a "true" flag that modifies the logic of the algorithm (please correct me if I'm wrong) but because those are the limits of the technology used. If in the future another dtype became available we'd have to change the API again. But maybe I'm overthinking it. ;)

Why do you think a flag would be better?

rfezzani:
Optical flow will always be estimated as float. Using single or double precision is a matter of treadoff between precision and memory use/computation speed.
Moreover, correct me if I am wrong, but I don't think that `float16` or `float128` are supported by `skimage`.
In my opinion, the use of `dtype` parameter is confusing for the user, he may be tempted to use any other dtype (`int`, `bool` or whatever).
The use of `dtype` also  forces a check and the raise of a `ValueError` if the input `dtype` is not float...
The use of a flag is in my opinion less confusing and prevents some user errors.

skimage/metrics/_contingency_table.py
jni: Yep, need to fix this, good catch!

skimage/exposure/histogram_matching.py
jni: No, skimage.**exposure**.match_histograms is a brand new function! =)

skimage/metrics/_structural_similarity.py
jni: Correct, just waiting for the discussion above to settle...

lagru:
@rfezzani Thanks for the insights. I don't see the check and raise as a disadvantage. Though I see your point that `dtype` might indeed be confusing for users who are used to that parameter from NumPy. I'm fine with either option but your suggestion seems to be the better one.

skimage/exposure/histogram_matching.py
sciunto: sounds good

skimage/util/compare.py
jni: @sciunto I think actually in this case this is better, because it's a literal string, not an argument name.

NelleV:
I've labeled this as critical as it is the only element we need to sort out before releasing 0.16 (which we've delayed by a month now)

jni:
Hi everyone,

After discussing with Stéfan in the community meeting this week, we settled on the following:
- for comparison functions, `image0` and `image1` where the metric is in a strong sense symmetric, and `image_true` and `image_test` where it is not. (though this could be `true_image` and `test_image`, come to think of it)
- for registration, `reference_image` and `moving_image`.

@NelleV has volunteered to be release manager and we've kept her waiting for a while, so I would like everyone who wants to have a say here to have one last review and then (hopefully) click approve. =) If no one has *strong* objections in the next ~24h, we will start pushing out 0.16 this time tomorrow.

Have a nice day! =)

skimage/util/compare.py
sciunto: You're  right, the rendering will be better.

skimage/registration/_optical_flow.py
```diff
@@ -11,15 +11,15 @@
 from ._optical_flow_utils import coarse_to_fine
 
 
-def _tvl1(image0, image1, flow0, attachment, tightness, nwarp, niter,
-          tol, prefilter):
+def _tvl1(reference_image, moving_image, flow0, attachment, tightness, nwarp,
+          niter, tol, prefilter):
```
rfezzani: Referring to your [comment ](https://github.com/scikit-image/scikit-image/issues/4154#issuecomment-531064682) on #4154, I suggest to replace `niter` by `num_iter` before releasing v0.16 to anticipate ;-)

skimage/registration/_optical_flow.py
```diff
@@ -135,9 +136,10 @@ def _tvl1(image0, image1, flow0, attachment, tightness, nwarp, niter,
     return flow_current
 
 
-def optical_flow_tvl1(image0, image1, *, attachment=15, tightness=0.3,
-                      nwarp=5, niter=10, tol=1e-4, prefilter=False,
-                      dtype='float32'):
+def optical_flow_tvl1(reference_image, moving_image,
+                      *,
+                      attachment=15, tightness=0.3, nwarp=5, niter=10,
```
rfezzani: The same here: `niter` -> `num_iter`

soupault:
>for registration, reference_image and moving_image.

How about `fixed_image` and `moving_image` instead? There are several votes for that option besides mine - https://github.com/scikit-image/scikit-image/pull/3544#issuecomment-535516956.

sciunto:
> > for registration, reference_image and moving_image.
> 
> How about `fixed_image` and `moving_image` instead? There are several votes for that option besides mine - [#3544 (comment)](https://github.com/scikit-image/scikit-image/pull/3544#issuecomment-535516956).

I like reference...

jni:
@soupault 

> How about `fixed_image` and `moving_image` instead?

Stéfan preferred reference, and I agree. (And so does @sciunto it turns out! :joy:) Ultimately I think the most important thing here is that it's unambiguous, and I think reference/moving meets that criterion.

soupault:
>it's unambiguous

I agree with that. Let's keep `reference` then. 👍

skimage/measure/tests/test_simple_metrics.py
```diff
@@ -52,12 +52,12 @@ def test_PSNR_errors():
 def test_NRMSE():
     x = np.ones(4)
     y = np.asarray([0., 2., 2., 2.])
-    assert_equal(normalized_root_mse(y, x, 'mean'), 1 / np.mean(y))
-    assert_equal(normalized_root_mse(y, x, 'Euclidean'), 1 / np.sqrt(3))
-    assert_equal(normalized_root_mse(y, x, 'min-max'), 1 / (y.max() - y.min()))
+    assert_equal(normalized_root_mse(y, x), 1 / np.mean(y))
+    assert_equal(normalized_root_mse(y, x), 1 / np.sqrt(3))
+    assert_equal(normalized_root_mse(y, x), 1 / (y.max() - y.min()))
```
grlee77: I think these need the old `norm_type` keyword instead of removing these arguments.

jni: Oops! These were changed automatically by PyCharm during the refactor of the kwarg, but they shouldn't have been using the new API — they should have been testing the existing API! So I've changed them back. I think the new commits should fix the build!

stefanv:
+1; @jni one observation: in comparison functions, we have `f(im_true, im_test)`, but registration is `f(image, reference)`.  This seems like the opposite way around.  Probably nothing we can do about that at this point, but feels inconsistent.

jni:
@stefanv huh? Optical flow uses `reference, moving`?

jni:
Ditto for `feature.register_translation`, which should be moved to `registration` but I think that can wait until the next release.

stefanv:
I just saw this one: https://github.com/scikit-image/scikit-image/pull/4187/files#diff-c6278103295d45c5fea1f5fcb3aff3e0R22

jni:
Ah. Sure but that's not registration. =) I think in that case it makes more sense, because you are modifying the input image to match a reference. The output is a modified version of the first input. In registration, the output is something that relates the first input to the second. So I don't think it's a critical inconsistency. But anyway, thanks for pointing it out! It's something to consider for 1.0 I think. =)

stefanv:
👍 I'd merge, but you said you had one more thing to change?

jni:
Nope, all done. =P

skimage/metrics/_structural_similarity.py
mkcor: Noticing these `im1, im2` as I'm reviewing #6612 and... it hurts my eyes! :cry: 

I'm adding this item to the list of [Pending API changes for 1.0](https://github.com/scikit-image/scikit-image/issues/5439#issuecomment-1268429149).

PR #4175, opened 2019-09-20
Morphological snakes

axarekma:
## Description

This PR contains a fast marching version of morphological_chan_vese based on propagation a vector of edge pixels/voxels. It is implemented as a parallel module
`skimake.segmentation.morphological_chan_vese_fm`
with the same interface as the original.

Credit where credit is due: design is heavily based on the source from:
>    Luis Alvarez, Luis Baumela, Pablo Márquez-Neila, and Pedro Henríquez,
    A Real Time Morphological Snakes Algorithm, Image Processing On Line,
    2 (2012), pp. 1–7. https://doi.org/10.5201/ipol.2012.abmh-rtmsa

This version extends it to 3D and has modified boundary conditions to conform with the scipy defaults for the morphologgical operators.

Extension to the existing version:
All the boolean operators are based on level_set==1 comparison, so passing an initial mask with value \notin {0,1} removes this region from the propagation and averages.

A quick and dirty benchmark on use-cases resulted in:

>  === 2d case ===
data.camera() (512, 512) 35 iterations (scikit image example)
init_ls = checkerboard_level_set(image.shape, 6)
Reference:           calls: 5          elapsed: 7.772 s       ms/call: 1554.4 ms
Edge propagation:    calls: 5          elapsed: 0.431 s       ms/call: 86.1 ms
    Overhead (iterations==0)
Reference:           calls: 5          elapsed: 0.001 s       ms/call: 0.2 ms
Edge propagation:    calls: 5          elapsed: 0.009 s       ms/call: 1.8 ms

> === 3d case ===
confocal.npy (60, 120, 160) 15 iterations (from the [morphsnakes ](https://github.com/pmneila/morphsnakes)repo)
init_ls = circle_level_set(image.shape, (30, 50, 80), 25)
Reference:           calls: 4          elapsed: 22.310 s       ms/call: 5577.6 ms
Edge propagation:    calls: 5          elapsed: 0.627 s       ms/call: 125.4 ms
    Overhead (iterations==0)
Reference:           calls: 5          elapsed: 0.009 s       ms/call: 1.8 ms
Edge propagation:    calls: 5          elapsed: 0.052 s       ms/call: 10.4 ms

## Checklist

- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Gallery example in `./doc/examples` (new features only)
- [ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
- [x] Unit tests
[...]

PR #4193, opened 2019-09-27
Fix conversion to float32 dtype

### LINKED ISSUE #4189: skimage.util.convert not converting to 'float32' with numpy v1.13.3
## Description

During my explorations for PR #4161 CI failure, I found that `skimage.util.convert` doesn't convert double precision images to single precision with `numpy v1.13.3` when called with `dtype='float32'`. Even more strange, it works when `dtype=np.float32`

## Way to reproduce

```python
>>> import numpy as np
>>> from skimage.util.dtype import convert
>>> a = np.random.rand(5, 5)
>>> convert(a, 'float32').dtype
dtype('float64')
>>> convert(a, np.float32).dtype
dtype('float32')
>>> np.__version__
'1.13.3'
```

This doesn't happen with more recent numpy version:

```python
>>> a = np.random.rand(5, 5)
>>> convert(a, 'float32').dtype
dtype('float32')
>>> convert(a, np.float32).dtype
dtype('float32')
>>> np.__version__
'1.16.4'
```

## Version information

```python
3.6.7 | packaged by conda-forge | (default, Jul  2 2019, 02:18:42) 
[GCC 7.3.0]
Linux-4.15.0-64-generic-x86_64-with-debian-buster-sid
scikit-image version: 0.16.dev0
numpy version: 1.13.3

```

rfezzani:
## Description

Closes #4189.

This PR consists in:

- refactoring `skimage.util.convert`: defining annex functions (`_scale`, `_dtype_bits`, `_dtype_itemsize`) outside the body of the `convert` function to make it more readable,
- refactoring of `_dtype_bits`: removing the definition of `compare` function,
- using `np.obj2sctype(dtype)` instead of `type_out` to fix the conversion to `float32` whatever is `numpy` version,
- adding tests for conversion from any combination of floating data type definition (`float32`, np.float, float, etc...)
- adding tests for conversion from a subclass of the desired output datatype (check for no regression).

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_moments.py
+++ b/skimage/measure/_moments.py
@@ -95,7 +95,7 @@ def moments_coords_central(coords, center=None, order=3):
 
     >>> coords2 = np.concatenate((coords, [[17, 17]]), axis=0)
     >>> np.round(moments_coords_central(coords2),
-    ...          decimals=2)  # doctest: +NORMALIZE_WHITESPACE +ELLIPSIS
+    ...          decimals=2)  # doctest: +NORMALIZE_WHITESPACE
     array([[ 17.  ,   0.  ,  22.12,  -2.49],
            [  0.  ,   3.53,   1.73,   7.4 ],
            [ 25.88,   6.02,  36.63,   8.83],
--- a/skimage/measure/_moments.py
+++ b/skimage/measure/_moments.py
@@ -95,7 +95,7 @@ def moments_coords_central(coords, center=None, order=3):
 
     >>> coords2 = np.concatenate((coords, [[17, 17]]), axis=0)
     >>> np.round(moments_coords_central(coords2),
-                 decimals=2)  # doctest: +NORMALIZE_WHITESPACE
+    ...          decimals=2)  # doctest: +NORMALIZE_WHITESPACE
     array([[17.  ,  0.  , 22.12, -2.49],
            [ 0.  ,  3.53,  1.73,  7.4 ],
            [25.88,  6.02, 36.63,  8.83],
--- a/skimage/measure/_moments.py
+++ b/skimage/measure/_moments.py
@@ -95,10 +95,10 @@ def moments_coords_central(coords, center=None, order=3):
 
     >>> coords2 = np.concatenate((coords, [[17, 17]]), axis=0)
     >>> np.round(moments_coords_central(coords2), 2)
-    array([[ 17.  ,   0.  ,  22.12,  -2.49],
-           [  0.  ,   3.53,   1.73,   7.4 ],
-           [ 25.88,   6.02,  36.63,   8.83],
-           [  4.15,  19.17,  14.8 ,  39.6 ]])
+    array([[17.  ,  0.  , 22.12, -2.49],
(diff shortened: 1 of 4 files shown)


sciunto:
I'm +1 with @hmaarrfk : bump numpy + keep the tests.

rfezzani:
I removed inner functions because they are not required by design and they make `convert` function body length from 262 lines to 173. It made my debugging much easier!
I would also refer to @jni [comment](https://github.com/scikit-image/scikit-image/pull/4175#issuecomment-533756432) last item concerning functions length...
But well, if you all agree and request reverting this refactoring, I will do it ;-)

sciunto:
It's not really the fact you separated the function but rather the shortcuts such as `compare_bits_to = bits.__le__ if kind == 'u' else bits.__lt__` that make me less happy with the modifications.

skimage/util/dtype.py
```diff
@@ -54,6 +54,126 @@ def dtype_limits(image, clip_negative=False):
...
+        The data type object element size.
+
+    Returns
+    -------
+    dtype: data type object
+        Data type of `kind` that can store a `bits` wide unsigned int
+
+    """
+
+    compare_bits_to = bits.__le__ if kind == 'u' else bits.__lt__
```
jni: I agree with @sciunto, undo this particular optimisation, as it is much less readable than the inner `compare` function. Something preferable would be to just be explicit in the loop:

```python
s = next(i for i in (itemsize,) + (2, 4, 8)
         if bits < i*8 or bits == i*8 and kind == 'u')
```

rfezzani: I like this ;-) well done @jni

rfezzani:
Thx @sciunto, I am still stuck with Travis failures :-( Even using `  # doctest: +NORMALIZE_WHITESPACE` doesn't solve the problem :disappointed:

sciunto:
Try numpy 0.14.1. Looks like there is a fix backported in this version that might be related to this.

sciunto:
It works.

skimage/measure/_moments.py
```diff
@@ -94,7 +94,8 @@ def moments_coords_central(coords, center=None, order=3):
     point, this no longer holds:
 
     >>> coords2 = np.concatenate((coords, [[17, 17]]), axis=0)
-    >>> np.round(moments_coords_central(coords2), 2)
+    >>> np.round(moments_coords_central(coords2),
+    ...          decimals=2)  # doctest: +NORMALIZE_WHITESPACE +ELLIPSIS
```
sciunto: I don't think ELLIPSIS is need. The output does not contain `...`

jni:
@hmaarrfk this is ready for you to pull the trigger! =) Thanks all!

hmaarrfk:
didn't know that my change request was blocking. I rarely mean to have blocking requests..

jni:
@hmaarrfk typically when I have requests that I don't want to block, I use "comment" rather than "request changes". Otherwise the big red x remains. I can "dismiss" it, but that feels, well, dismissive. =D

PR #4195, opened 2019-09-29
Explain use of num_peaks parameter in corner_peaks

madebyollin:
## Description

Adds a note in the documentation for `skimage.feature.corner_peaks` that explains how the `num_peaks` parameter is interpreted.

At present (without reading the source code) it is ambiguous whether the `num_peaks` limit applies *before* suppression of connected peaks, or *after*.

If a user assumes the latter behavior, then `corner_peaks` will appear broken when the `num_peaks` argument is used; `corner_peaks` will return far fewer than the expected number of results, and adding more peaks to an input image may increase *or decrease* the number of results that `corner_peaks` returns for that image.

This PR adds documentation that clarifies the meaning of the `num_peaks` parameter.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/corner.py
+++ b/skimage/feature/corner.py
@@ -920,10 +920,11 @@ def corner_peaks(image, min_distance=1, threshold_abs=None, threshold_rel=0.1,
 
     This differs from `skimage.feature.peak_local_max` in that it suppresses
     multiple connected peaks with the same accumulator value.
-    
-    Note that the `num_peaks` limit is applied *before* suppression of connected
-    peaks. If you want to limit the number of peaks *after suppression*, you
-    should set `num_peaks=np.inf` and post-process the output of this function.
+
+    Note that the `num_peaks` limit is applied *before* suppression of 
+    connected peaks. If you want to limit the number of peaks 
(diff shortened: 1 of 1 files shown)


skimage/feature/corner.py
```diff
@@ -921,6 +921,11 @@ def corner_peaks(image, min_distance=1, threshold_abs=None, threshold_rel=0.1,
     This differs from `skimage.feature.peak_local_max` in that it suppresses
     multiple connected peaks with the same accumulator value.
 
+    Note that the `num_peaks` limit is applied *before* suppression of
```
soupault: I think this paragraph should rather go into `Notes` section.

Backticks should be doubled (see https://github.com/scikit-image/scikit-image/issues/4117), asterisks - removed.

Otherwise, 👍 .

sciunto: I'm +1 for the note section. But please, keep simple backticks. We fixed the issue and they are correctly handled now #4188

sciunto:
look good to me!

jni:
Thank you @madebyollin!

PR #4196, opened 2019-09-30
Update optical flow functions API

rfezzani:
## Description

Here is the  implementation of my [suggestion](https://github.com/scikit-image/scikit-image/pull/4187#issuecomment-536025590) concerning the modification of optical flow API: `dtype` argument is replaced by `single_precision`.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/registration/_optical_flow.py
+++ b/skimage/registration/_optical_flow.py
@@ -137,7 +137,7 @@ def _tvl1(image0, image1, flow0, attachment, tightness, nwarp, niter,
 
 def optical_flow_tvl1(image0, image1, *, attachment=15, tightness=0.3,
                       nwarp=5, niter=10, tol=1e-4, prefilter=False,
-                      dtype='float32'):
+                      single_precision=True):
     r"""Coarse to fine optical flow estimator.
 
     The TV-L1 solver is applied at each level of the image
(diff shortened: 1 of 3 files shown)


jni:
@rfezzani I am :-1: on this, sorry that I didn't have time to chime in earlier when you first made the comment. Even though only two dtypes are supported, `dtype=` is very understandable for someone in the scientific Python ecosystem, and is additionally consistent with a bunch of other functions in scikit-image and also NumPy. In contrast, "single_precision" is a unique argument name that doesn't exist anywhere else. I'll let others chime in, but unless someone wants to advocate very hard for this, my position is to close it and stay with dtype.

jni:
Yeah I read that argument and remain unmoved. =P

More specifically:

- as mentioned above, dtype is standard, and in my opinion, properly documenting the options plus raising meaningful errors is sufficient to prevent most user confusion. The equivalent would be renaming the main argument `image_2d_or_3d` in all functions that are not nD-compatible, and I think that reads much worse.
- we don't support float16 or float128 or complex128, but maybe in the distant future we will? In which case, the argument would need to be renamed, which is annoying. This is part of the value given by consistency.

stefanv:
Yes, the dtype argument probably makes most sense.  At least, more so than inventing a new keyword argument which we'll have to standardize across the package.

PR #4197, opened 2019-09-30
Remove the dependency to ndi.zoom

rfezzani:
## Description

In this PR, the calls to `ndi.zoom` are replaced by the internal `transform.resize` function.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/segmentation/tests/test_boundaries.py
```diff
@@ -108,13 +108,14 @@ def test_mark_boundaries_subpixel():
...
     ref_result = np.array(
-        [[ 0.55,  0.63,  0.72,  0.69,  0.6 ,  0.55,  0.54],
-         [ 0.45,  0.58,  0.72,  1.  ,  1.  ,  1.  ,  0.69],
-         [ 0.42,  0.54,  0.65,  1.  ,  0.44,  1.  ,  0.89],
-         [ 0.69,  1.  ,  1.  ,  1.  ,  0.69,  1.  ,  0.83],
-         [ 0.96,  1.  ,  0.38,  1.  ,  0.79,  1.  ,  0.53],
-         [ 0.89,  1.  ,  1.  ,  1.  ,  0.38,  1.  ,  0.16],
-         [ 0.57,  0.78,  0.93,  1.  ,  0.07,  1.  ,  0.09],
-         [ 0.2 ,  0.52,  0.92,  1.  ,  1.  ,  1.  ,  0.54],
-         [ 0.02,  0.35,  0.83,  0.9 ,  0.78,  0.81,  0.87]])
```
sciunto: Can you justify why you had to change the values? (looks significant)

rfezzani: Not really, I don't know if `resize` is supposed to reproduce the results of `ndi.zoom`, I did not found any test asserting this.

rfezzani:
Thank you @sciunto for your review, I can add some tests for `resize_flow` in another PR if you don't mind...

sciunto:
> Thank you @sciunto for your review, I can add some tests for `resize_flow` in another PR if you don't mind...

That works for me.

skimage/segmentation/tests/test_boundaries.py
hmaarrfk: Is this touching up some new functionality you added to 0.16.dev0? if so, I guess you literally have a few days to change the behavior.

rfezzani: No, this PR is simply for maintenance.
Even if it touches `resize_flow`, the differences between `ndi.zoom` and `resize` outputs should be minor.  I don't think that it impacts the quality of the optical flow results since `resize_flow` only provides the initial flow at each pyramid level.

hmaarrfk: Looking at git blame, it seems that this test was from 5 years ago.

You say: *should be minor*, however the diff shows that the differences **are not** minor and can be as high as `0.1` on an absolute scale from `0.0` to `1.0` as on the line that @sciunto pointed out.

I don't think avoiding `ndi.zoom` is good enough to break people's code that rely on the public function `mark_boundaries`.

It seems that the differences are exacerbated on the perimeter, which would indicate that it is an edge problem. Maybe you can select a different boundary condition?

Honestly, what does avoiding `ndi.zoom` give you? Speed? Import time?

rfezzani: This difference of `0.1` is I guess caused by the boundary conditions on this really small array of shape (7, 7) with an interpolation of order 3.
`mark_boundaries` is made I think for visualization, so I don't think that it will impact that much users, but well :-p
I tested all the available boundary conditions and the `'symmetric' provides the closest results to the original implementation.
Removing an external dependency always gives you more control on your code base. This is I think valuable

hmaarrfk: It requires a strong force to break innertia. And I think an even stronger one to break backward incompatibility.

I don't see strong arguments for removing a dependency on ndimage (in fact, many PRs push it the other way) nor do I see a big need to break backward compatibility.

sciunto: It convinces me to keep the edits for the local imports and to investigate how we can expand our test coverage, and revert the switch from zoom to resize.

ndimage is something we assume to rely on, at least for their public functions.

jni: I'm a bit on @hmaarrfk's camp here. Sorry @rfezzani, I guess this is the 7th time more or less that we stifle you because of our concern for backwards compatibility. =P

Anyway, resize itself depends on ndimage, and it depends on ndimage strongly if the image dimension is greater than 2D. So, as @hmaarrfk points out, we have to weigh the value of changing everything to resize (minimal) against the cost of potentially breaking downstream users' code (nonzero). Yes, it's a visualisation function, but some other packages might have written tests like these that will break if we update.

rfezzani: No problem @jni, PRs are made to be reviewed and potentially rejected ;-)
I still believe that the backward compatibility requirement is a false problem, particularly for a v0 package :-P. I already told this in a previous answer, but people who care about soft stability fix the versions of their dependencies.
I work in the medical software dev field, you can trust me when I tell you that I am aware of soft stability issues...
Back to the differences observed on the test result, it is I think because of the difference of interpolation model between the two functions: pixel model VS data point model (see this [link](http://scipy.github.io/devdocs/roadmap-detailed.html#ndimage)). May be you can confirm that resize uses the pixel model?

jni: @rfezzani yes, perhaps one day we can discuss backwards compatibility over a coffee. (@NelleV is toying with the idea of organising a sprint in Paris in the not-too-distant future. ;) For now let's just assume that minimising these changes forms part of the objective function of this library. =) Certainly, having lived through several fires in skimage caused by NumPy making backwards incompatible changes, I have a lot of sympathy for users who don't pin, yet want some semblance of stability. =)

Anyway, I think you are probably quite right about the result difference. I also happen to think that that document is correct that the data-point model is better and thus that we should use `ndi.zoom` until such time as we have migrated `warp` to use the data point model. =)

jni:
@rfezzani I'm going to close this for now given the above comments. If the discussion allows, we can always reopen! Thank you again!

skimage/segmentation/tests/test_boundaries.py
rfezzani: With great pleasure @jni, you are welcome to Paris ;-)
We may also discuss interpolation models, because I am not convinced that data-point is the best model to adopt for image interpolation :smile:

PR #4201, opened 2019-10-03
Add rescale local mean

### LINKED ISSUE #2827: downscale_local_mean with non-integer factors?
I'm interested in a version of `skimage.transform.downscale_local_mean()` which handles non-evenly dividing factors without padding the image. I want each pixel in the result to be an weighted average of overlapping pixels in the source, with weights equal to the area of overlap.

To be more explicit, this transformation is linear along equal axis. For example, from size 5 to size 2, it can be written as the matrix:
```
[[0.4, 0.4, 0.2, 0.0, 0.0],
 [0.0, 0.0, 0.2, 0.4, 0.4]]
```

This is actually what I expected `skimage.transform.resize()` to do, before I noticed the docstring warning about anti-aliasing.

Would this be a welcome addition to scikit-image in some form? I wrote a version that works for my own purposes, but my guess is that it's too slow for skimage, since I use a dense matrix multiply along each axis with matrix of weights.

rfezzani:
## Description

Close #2827.

Based on @shoyer [gist](https://gist.github.com/shoyer/c0f1ddf409667650a076c058f9a17276), I implemented the `resize_local_mean` functionality.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_warps.py
+++ b/skimage/transform/_warps.py
@@ -1070,3 +1070,84 @@ def warp_polar(image, center=None, *, radius=None, output_shape=None,
                   output_shape=output_shape, **kwargs)
 
     return warped
+
+
+def _resize_weights(old_size, new_size, reflect=False):
+    """Create a weight matrix for resizing with the local mean along an axis.
+
+    Parameters
+    ----------
+    old_size: int
+        Old size.
+    new_size: int
+        New size.
+    reflect: bool
+        Whether to use reflecting boundary conditions or not.
+
+    Returns
+    -------
+    NumPy array with shape (new_size, old_size). Rows sum to 1.
+    """
+    if reflect:
+        old, new = old_size - 1, new_size - 1
+        old_breaks = np.pad(np.linspace(0.5, old-0.5, old),
+                            1, 'constant', constant_values=(0, old))
+        val = 0.5 * old / new
+        new_breaks = np.pad(np.linspace(val, old-val, new),
+                            1, 'constant', constant_values=(0, old))
+    else:
+        old_breaks = np.linspace(0, old_size, num=old_size + 1)
+        new_breaks = np.linspace(0, old_size, num=new_size + 1)
+
+    upper = np.minimum(new_breaks[1:, np.newaxis], old_breaks[np.newaxis, 1:])
+    lower = np.maximum(new_breaks[:-1, np.newaxis],
+                       old_breaks[np.newaxis, :-1])
+
+    weights = np.maximum(upper - lower, 0)
+    weights /= np.sum(weights, axis=1, keepdims=True)
+
+    return weights
+
+
+def resize_local_mean(image, output_shape, reflect_axes=None):
+    """Resize an array with the local mean / bilinear scaling.
+
+    Works for both upsampling and downsampling in a fashion equivalent
+    to block_mean and zoom, but allows for resizing by non-integer
+    multiples. Prefer block_mean and zoom when possible, as this
+    implementation is probably slower.
+
+    Parameters
+    ----------
+    image : ndarray
+        Input image.
+    output_shape : tuple or ndarray
+        Size of the generated output image `(rows, cols[, ...][, dim])`. If
+        `dim` is not provided, the number of channels is preserved. In case the
+        number of input channels does not equal the number of output channels a
+        n-dimensional interpolation is applied.
+
+    Returns
+    -------
+    Array resized to shape.
+
+    Raises:
+      ValueError: if any values in reflect_axes fall outside the interval
+        [-array.ndim, array.ndim).
+
+    """
+    reflect_axes_set = set()
+    for axis in reflect_axes:
+        if not -image.ndim <= axis < image.ndim:
+            raise ValueError('invalid axis: {}'.format(axis))
+        reflect_axes_set.add(axis % image.ndim)
+
+    output = image
+    for axis, (old_size, new_size) in enumerate(zip(image.shape,
+                                                    output_shape)):
+        reflect = axis in reflect_axes_set
+        weights = _resize_weights(old_size, new_size, reflect=reflect)
+        product = np.tensordot(output, weights, [[axis], [-1]])
+        output = np.moveaxis(product, -1, axis)
+
+    return output
--- a/skimage/transform/_warps.py
+++ b/skimage/transform/_warps.py
@@ -1072,8 +1072,9 @@ def warp_polar(image, center=None, *, radius=None, output_shape=None,
     return warped
 
 
-def _resize_weights(old_size, new_size, reflect=False):
-    """Create a weight matrix for resizing with the local mean along an axis.
+def _local_mean_weights(old_size, new_size, reflect=False):
+    """Create a weight matrix for resizing with the local mean along an
+    axis.
 
     Parameters
     ----------
@@ -1087,6 +1088,7 @@ def _resize_weights(old_size, new_size, reflect=False):
     Returns
     -------
     NumPy array with shape (new_size, old_size). Rows sum to 1.
+
     """
     if reflect:
         old, new = old_size - 1, new_size - 1
@@ -1104,18 +1106,13 @@ def _resize_weights(old_size, new_size, reflect=False):
                        old_breaks[np.newaxis, :-1])
 
     weights = np.maximum(upper - lower, 0)
-    weights /= np.sum(weights, axis=1, keepdims=True)
+    weights /= weights.sum(axis=1, keepdims=True)
 
     return weights
 
 
 def resize_local_mean(image, output_shape, reflect_axes=None):
     """Resize an array with the local mean / bilinear scaling.
-
-    Works for both upsampling and downsampling in a fashion equivalent
-    to block_mean and zoom, but allows for resizing by non-integer
-    multiples. Prefer block_mean and zoom when possible, as this
-    implementation is probably slower.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


sciunto:
I've not read the code, but I want us to keep in mind that the original code was published with an Apache license and care must be taken: https://opensource.stackexchange.com/questions/1713/adding-feature-to-an-apache-v2-package

rfezzani:
@sciunto, according to the stackexchange link that you provided, it seems enough to refer to the gist of @shoyer (in the documentation for example). Any thoughts?

shoyer:
If it helps, I can make a pull request with the contents of my gist to skimage or any other open source project under a 3 clause BSD license. I just don't have time to clean this up, and for self-contained code we're supposed to use the Apache license.

rfezzani:
Hi @shoyer, I don't know which option is the best, let the core team decide here ;-).
May be can you in the meantime review the code? Thank you

alexdesiqueira:
Hey @shoyer,
> If it helps, I can make a pull request with the contents of my gist to skimage or any other open source project under a 3 clause BSD license.

We would be very grateful if you could release the code under a BSD license. :blush: Could you help us with that?
Thank you very much!

shoyer:
OK, see https://github.com/scikit-image/scikit-image/pull/4412

rfezzani:
Thank you @shoyer ;-)

alexdesiqueira:
Thank you very much @shoyer! We really appreciate you help. :)

skimage/transform/tests/test_warps.py
```diff
@@ -632,3 +633,111 @@ def test_invalid_dimensions_polar():
...
+
+def test_resize_local_mean_nd():
+    for dim in range(1, 6):
+        shape = 2 + np.arange(dim) * 2
+        x = np.ones(shape)
+        out_shape = np.asarray(shape) * 1.5
+        resized = resize_local_mean(x, out_shape)
+        expected_shape = 1.5 * shape
+        assert_equal(resized.shape, expected_shape)
+        assert np.all(resized == 1)
```
hmaarrfk:
```suggestion
        assert_array_equal(resized, 1)
```

hmaarrfk: It provides a little more feedback when things go wrong

skimage/transform/_warps.py
```diff
@@ -16,6 +16,54 @@
...
+        If output_shape length is smaller than the image number of
+        dimensions
+
+    Notes
+    -----
+    The input image is reshaped if its number of dimensions is not
+    equal to output_shape_length.
+
+    """
+    output_shape = tuple(np.ravel(output_shape))
```
hmaarrfk: What usecase are you trying to ensure works here? why not omit this line?

hmaarrfk: i guess, if you want an `ndarray`, then `tuple(output_shape)` is good enough right?

skimage/transform/_warps.py
```diff
@@ -16,6 +16,54 @@
...
+    output_ndim = len(output_shape)
+    input_shape = image.shape
+    if output_ndim > image.ndim:
+        # append dimensions to input_shape
+        input_shape += (1, ) * (output_ndim - image.ndim)
+        image = np.reshape(image, input_shape)
+    elif output_ndim == image.ndim - 1:
```
hmaarrfk: I don't think we want to be making `assumptions` here without providing the user an explicit way out of this.

Adding the `multichannel` parameter would go a long way here.

```
elif multichannel and output_ndim == image.ndim - 1
```

skimage/transform/_warps.py
```diff
@@ -16,6 +16,54 @@
...
+        # multichannel case: append shape of last axis
+        output_shape = output_shape + (image.shape[-1], )
+    elif output_ndim < image.ndim - 1:
```
hmaarrfk: With the above statement, this logic should be 

```
elif output_ndim < image.ndim:
```

skimage/transform/_warps.py
rfezzani: It allows providing output_shape as an ndarray with shape (1, n)
```python
>>> tuple(np.ones((1, 3)))
(array([1., 1., 1.]),)
>>> tuple(np.ravel(ones((1, 3))))
(1.0, 1.0, 1.0)
```

skimage/transform/_warps.py
rfezzani: I simply wrapped all the checks/preprocessing that are already made in `resize` in this function to avoid code duplication in the new `rescale_local_mean` function...

rfezzani: @hmaarrfk, do you think then that we need to add the `multichannel` argument to `resize`?

skimage/transform/_warps.py
jni: > It allows providing output_shape as an ndarray with shape (1, n)

And this is desirable because...? 😂 

I'll note that the conversation over at the Array object data API suggests that `.shape` *must* be a 1D object:

https://data-apis.github.io/array-api/latest/API_specification/array_object.html#shape

I suggest we stick to that.

skimage/transform/_warps.py
jni: I think this is probably a good idea... BUT, anyway, I suggest pushing all this to 0.19, and I intend to make a big push on the `channel_axis` proposal in the coming months.

jni:
@rfezzani I've popped this on the 0.19 milestone, sorry for the long delay in reviewing. 😬 if needed urgently we can always backport to 0.18.x. Also, as mentioned in #5080 I'd like to start moving to more frequent releases, now that we have automatic wheel building on the main repo! 🎉

rfezzani:
OK @jni, it is fine by me :wink:

skimage/transform/_warps.py
rfezzani: Yes, I agree with you and :-1: with my old me :stuck_out_tongue_closed_eyes:. I will modify this ASAP

rfezzani: It's fixed! Thank you @hmaarrfk and @jni ;)

rfezzani:
Anyone motivated for a review? :slightly_smiling_face:

skimage/transform/_warps.py
```diff
@@ -1093,3 +1130,101 @@ def warp_polar(image, center=None, *, radius=None, output_shape=None,
...
+    image : ndarray
+        Input image.
+    output_shape : tuple or ndarray
+        Size of the generated output image `(rows, cols[, ...][, dim])`. If
+        `dim` is not provided, the number of channels is preserved. In case the
+        number of input channels does not equal the number of output channels a
+        n-dimensional interpolation is applied.
+    grid_mode : bool, optional
+        Whether to use grid data model of pixel/voxel model for
+        average weights computation.
```
grlee77, line 1195:
```suggestion
        If False, the distance from the pixel centers is zoomed. Otherwise, the
        distance including the full pixel extent is used. For example, a 1d
        signal of length 5 is considered to have length 4 when `grid_mode` is
        False, but length 5 when `grid_mode` is True. See the following
        visual illustration:

        .. code-block:: text

                | pixel 1 | pixel 2 | pixel 3 | pixel 4 | pixel 5 |
                     |<-------------------------------------->|
                                        vs.
                |<----------------------------------------------->|

        The starting point of the arrow in the diagram above corresponds to
        coordinate location 0 in each mode.
```

This is a copy of a docstring I wrote for the parameter of the same that was introduced to `scipy.ndimage.zoom` in SciPy 1.6. Hopefully it helps clarify the difference between the two modes

grlee77 (changes requested):
I made some docstring suggestions. I can add the `channel_axis` parameter to make this consistent with the new way of handling multichannel images.

grlee77:
I updated the docstrings and added a `channel_axis` kwarg. 

Alternatively, requiring `len(output_shape) == image.ndim` would allow making things much simpler with no need for the `channel_axis` argument. Let me know what you think.

skimage/transform/_warps.py
```diff
@@ -1093,3 +1131,167 @@ def warp_polar(image, center=None, *, radius=None, output_shape=None,
...
+        to channels should be specified using `channel_axis`
+    output_shape : tuple or ndarray
+        Size of the generated output image. When `channel_axis` is not None,
+        the `channel_axis` should either be omitted from `output_shape` or the
+        ``output_shape[channel_axis]`` must match
+        ``image.shape[channel_axis]``. If the length of `output_shape` exceeds
+        image.ndim, additional singleton dimensions will be appended to the
+        input ``image`` as needed.
+    grid_mode : bool, optional
+        If False, the distance from the pixel centers is zoomed. Otherwise, the
```
rfezzani, line 1198: This sentence is not clear to me. It is specific to `scipy.ndimage.zoom`.
I like the example and the illustration, but the description is confusing to me :confused:.
What about
```
Defines ``image`` pixels position: if True, pixels are assumed to be at grid intersections,
otherwise at cell centers. As a consequence, for example, a 1d ...
```

skimage/transform/_warps.py
rfezzani: I like the idea of adding support for `channel_axis` in `resize_local_mean` like you did (:pray:), but adding it here in will brake the behaviour of `resize`. This I think must be managed in a separate PR.

skimage/transform/_warps.py
grlee77: done

grlee77 (approved):

alexdesiqueira:
Thank you @rfezzani!

rfezzani:
:tada: Thank you @alexdesiqueira and @grlee77 !

rfezzani:
... and of course all the reviewers :wink:

PR #4206, opened 2019-10-04
Add a reduction operation for profile_line

PinkShnack:
## Description

WIP Work in Progress
Enhancement to profile_line function. Add a `reduce_func `parameter which allows the user to use any ufunc on the pixel values before returning. Setting `reduce_func` to None will return the entire array.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/measure/profile.py
```diff
@@ -66,7 +67,15 @@ def profile_line(image, src, dst, linewidth=1,
...
-    intensities = pixels.mean(axis=1)
+
+    if intensity_type == 'mean':
+        intensities = pixels.mean(axis=1)
+    elif intensity_type == 'max' or 'maximum':
+        intensities = pixels.max(axis=1)
+    elif intensity_type == 'min' or 'minimum':
+        intensities = pixels.min(axis=1)
+    elif intensity_type == 'sum' or 'total':
+        intensities = pixels.sum(axis=1)
```
lagru: I'd prefer to keep the number of possible values for `intensity_type` to a minimum. I'd omit `'max'` in favor of `'maximum'`, `'min'` in favor of `'minimum'` and `'sum'` in favor of `'total'`. Is there a reason you didn't add the median as a possible option?

lagru: Oh and we should raise a ValueError in case `intensity_type` contains none of those values.

skimage/measure/profile.py
```diff
@@ -3,7 +3,8 @@
 
 
 def profile_line(image, src, dst, linewidth=1,
-                 order=1, mode='constant', cval=0.0):
+                 order=1, mode='constant', cval=0.0,
+                 intensity_type='mean'):
```
lagru: I think we can make this a keyword only argument.

skimage/measure/profile.py
jni: @PinkShnack this is a great enhancement! But I think a much better implementation here is to forget about strings and allow passing any ufunc. I would call the argument (which yes, should be keyword-only) `aggregation=np.mean` or `summary_function=np.mean`, or a similar name, and then people can pass in any function that can take an array and an axis argument, and get the result:

```python
try:
    intensities = aggregation(pixels, axis=1)
except TypeError:  # function doesn't allow axis kwarg
    intensities = np.apply_along_axis(aggregation_function, pixels, axis=1)
```

Then the usage would be `profile_line(image, src, dst, aggregation=np.max)`

What do people think?

lagru: You're right, that's way better! ^^ 

I still find `merge_channels` to be a better description of the parameter though. It makes it immediately obvious that it has no effect for images with only one channel (e.g. grayscale, in which case it only serves to remove an unnecessary axis of length 1 in `pixels`).

PinkShnack: Thanks for the quick responses all!

@lagru 
**1. Greyscale.** My reason for doing this enhancement was to change how the outputted pixels were dealt when summing/averaging etc., over a certain `linewidth`. For example, if you have a `profile_line` with a `linewidth` of 5, you will get different outputs when you look at the max, mean, min, median, or summed values of those 5 values along the `profile_line`. I hadn't even thought of multi channeled images, I deal solely with greyscale!

Therefore, should we have two different parameters? One for choosing the aggregation over the linewidth, and one for choosing the aggregation over multi-channel images?
Of course, the former will have no effect over a `profile_line` with `linewidth=1`. As far as I can see, these should be different parameters. Thoughts?

2. max, min, sum are simpler alright! No reason to leave out median let's include it! Any others?

**3. Keyword only argument.** I thought it should have a default because I don't want to interrupt people's old code!

@jni 

**ufunc.** That's great, so cool, I'll start using that in my code!

lagru: Oh, I made two mistakes: I overlooked the `linewidth` argument and assumed that the mean function is used to reduce the dimension with different channels which it does not! That also makes my argument that the new keyword should relate to channels moot. As @jni suggested `aggregation` may be a good name, though now I particular like `reduce_func` which I find more explicit than the former.

> Therefore, should we have two different parameters? One for choosing the aggregation over the linewidth, and one for choosing the aggregation over multi-channel images?
Of course, the former will have no effect over a profile_line with linewidth=1. As far as I can see, these should be different parameters. Thoughts?

No I'd keep it at one parameter to control the reduction operation. As you correctly stated the old behavior is only relevant when `linewidth` is > 1. I see this mainly as a convenience argument to keep the output to one dimension. The user can always perform an additional reduction on multichannel images.

> 3. Keyword only argument. I thought it should have a default because I don't want to interrupt people's old code!

Did you perhaps misunderstand me? Being keyword-only does not preclude using a default value. With [keyword-only](https://www.python.org/dev/peps/pep-3102/) I mean parameter names that come after an `*,`, e.g.

```python
def foo(x, *, keyword_only=None):
    print(x, keyword_only)

foo(1, 2)
# "TypeError: foo() takes 1 positional argument but 2 were given"
```

rfezzani: I like `reduce_func` as argument name ;-)

skimage/measure/profile.py
```diff
@@ -27,6 +27,10 @@ def profile_line(image, src, dst, linewidth=1,
         How to compute any values falling outside of the image.
     cval : float, optional
         If `mode` is 'constant', what constant value to use outside the image.
+    reduce_func : NumPy ufunc, default np.mean
```
rfezzani: Not necessarily a Numpy ufunc, it can be any `ufunc` right?

skimage/measure/profile.py
```diff
@@ -68,14 +80,10 @@ def profile_line(image, src, dst, linewidth=1,
...
 
-    if intensity_type == 'mean':
-        intensities = pixels.mean(axis=1)
-    elif intensity_type == 'max' or 'maximum':
-        intensities = pixels.max(axis=1)
-    elif intensity_type == 'min' or 'minimum':
-        intensities = pixels.min(axis=1)
-    elif intensity_type == 'sum' or 'total':
-        intensities = pixels.sum(axis=1)
+    if reduce_func == None:
```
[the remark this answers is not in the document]
rfezzani: I vote for setting to `None` by default. BTW, the test would be `if reduce_func is None` instead of `if reduce_func == None`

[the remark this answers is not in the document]
lagru: > Should I have default as np.mean, and leave out this check?

I'd say yes, except for the fact that this will not render nicely in the HTML documentation which will show that value as `reduce_func=<function mean at 0x7fc9686c9d40>` or something similar. So perhaps we should indeed use `None` as the default value. I'd definitely prefer @jni's suggestion with the try except block.

skimage/measure/profile.py
lagru: I'd go with `reduce_func : callable, optional` and explain the rest inside the description. If we use @jni's try except suggestion we can essentially use any callable that takes an array and returns a value or array.

skimage/measure/profile.py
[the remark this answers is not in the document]
jni: Well, I kinda would like `reduce_func=None` to mean that the whole 2D array is returned... So I'm thinking that the doc problem should be a fix to sphinx rather than work around doing the wrong thing...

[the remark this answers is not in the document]
lagru: Well, we could define a class that wraps the function:

```python
class wrapped_mean:

    def __call__(self, *args, **kwargs):
        return np.mean(*args, **kwargs)

    def __repr__(self):
        return "{}.{}".format(np.mean.__module__, np.mean.__qualname__)
```
and use `reduce_func=wrapped_mean()` as a default.

... just kidding. :smile: You're probably right. But we should definitely make sure to explain the default value in the parameter description then because the signature won't be helpful there.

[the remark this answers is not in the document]
PinkShnack: Perhaps something like this?

```python
    if reduce_func is None:
        pass
    else:    
        try:
            intensities = reduce_func(pixels, axis=1)
        except TypeError:  # function doesn't allow axis kwarg
            intensities = np.apply_along_axis(reduce_func, arr=pixels, axis=1)
```

or could give a message or warning to the user instead of `pass`. Example:
`print("To reduce the array along the linewidth axis to 1D, choose a reduce_func ufunc")`

This would output the intensities array as 2D for reduce_func=None, as @jni mentioned :

> I kinda would like reduce_func=None to mean that the whole 2D array is returned...

I'd reckon it's good that the user can have this array outputted without reduction.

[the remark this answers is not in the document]
rfezzani: Am I wrong or `reduce_func` acts on linewidth "direction"? So following @jni interpretation of `reduce_func=None`, the returned "line" is 2D for gray scale images and 3D for RGB images.
If `linewidth == 1`, no need for reduction, so may be a warning can be raised only if `reduce_func is None and linewidth > 1`? What do you think?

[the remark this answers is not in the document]
lagru: @rfezzani Correct. But I don't think we need a warning. The behavior for passing `None` should be clearly documented and we can assume that if the user explicitly passes in `None` the result is expected.

Also, passing in `reduce_func=lambda arr, axis: arr` will return the same as passing in `None`. But it's probably worth it to allow `None` for the users convenience.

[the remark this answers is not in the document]
PinkShnack: You're right, my phrasing was probably a tad off :). 

Just to note, if `reduce_func=None` and `linewidth=1`, the array has the shape (N,1), which is still 2D. 
However, if `reduce_func=np.mean`, for example, and `linewidth=1`, the array has the shape (N, ). Should we make the output consistent? Or should we leave it as is, because that is what it should output...?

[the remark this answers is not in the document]
jni: The above behaviour (None -> 2D, anything else -> 1D) is the consistent behaviour, even for line width = 1. I think the following is right:

```python
    if reduce_func is None:
        intensities = pixels
    else:    
        try:
            intensities = reduce_func(pixels, axis=1)
        except TypeError:  # function doesn't allow axis kwarg
            intensities = np.apply_along_axis(reduce_func, arr=pixels, axis=1)
```

Additionally, with regards to the representation in the docs, I think @lagru your suggestion is not so terrible or funny, but rather something that could be used everywhere! I'd use a functional approach though:

```python
def pretty_print(func):
    @functools.wraps
    def inner(*args, **kwargs):
        return func(*args, **kwargs)
    inner.__str__ = lambda self: func.__module__ + '.' + func.__name__
    inner.__repr__ = inner.__str__
    return inner
```

I think this should work? Then it would be `reduce_func=pretty_print(np.mean)`, and it would show up as `numpy.mean` in the docs.

Bonus points if you can get intersphinx to link to the documentation for numpy.mean :joy: 

**update:** Ok, I tried it, and it fails spectacularly:

```python
In [10]: mean = pretty_print(np.mean)                                                                                                                                                         

In [11]: str(mean)                                                                                                                                                                            
Out[11]: "functools.partial(<function update_wrapper at 0x7f531b8dd9d8>, wrapped=<function pretty_print.<locals>.inner at 0x7f531881d1e0>, assigned=('__module__', '__name__', '__qualname__', '__doc__', '__annotations__'), updated=('__dict__',))"
```

:joy: but anyway, all this to say, there's legs to this idea, and I think it'd be generally useful, so this decorator should live in `_shared` and eventually might be graduated to e.g. numpydoc itself.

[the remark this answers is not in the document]
rfezzani: I don't really like to pass callable as default argument because the kind of problems that we are facing right now...
The behavior (None -> 2D, anything else -> 1D) being new, what about a deprecation cycle?
```python
if reduce_func is None:
    warn("profile_line offers now the ability to apply different reduction strategies "
         + "perpendicularly to profile direction. In future version default behavior will "
         + "be no reduction. To disable this warning, set reduce_func to numpy.mean",  
         FutureWarning, stacklevel=2)
    reduce_func = np.mean
try:
    intensities = reduce_func(pixels, axis=1)
except TypeError:  # function doesn't allow axis kwarg
    intensities = np.apply_along_axis(reduce_func, arr=pixels, axis=1)
```

skimage/measure/profile.py
soupault: FWIW, PyTorch uses `reduction`.

skimage/measure/profile.py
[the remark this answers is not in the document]
jni: @rfezzani I think the mean should remain the default. I don't see the documentation issue as a big problem. Having None be the default and adding a deprecation message makes it more annoying to use in the most common case. Reducing the need for warnings is definitely a goal for scikit-image.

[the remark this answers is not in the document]
lagru: I agree with @jni here. Default behavior should stay at `mean`. Accepting `None` as well is probably a good idea for less advanced users.

Also, your attempt probably didn't work because Python does method lookup on the class dictionary and not on the instance. So setting the `__repr__` attribute on the instance has no effect! However a class based wrapper should work fine. When I find the time I'll have a look into Sphinx's inspection for signatures and see if the problem should be fixed there or here.

skimage/measure/profile.py
```diff
@@ -79,11 +87,15 @@ def profile_line(image, src, dst, linewidth=1,
     else:
         pixels = ndi.map_coordinates(image, perp_lines,
                                      order=order, mode=mode, cval=cval)
+    pixels = np.flip(pixels, axis=1)
```
PinkShnack: If anyone is curious about this, The outputted array with `linewidth=1` and `reduce_func=None` gives an array where the row (axis=1) values are flipped. Here I make this consistent. Tested in` test_reduce_func_None_linewidth_3()`.

PinkShnack:
The same error seems to be causing some of the CI builds from failing: "skimage/measure/profile.py:docstring of skimage.measure.profile_line:88:Unexpected indentation."

I'm not sure what exactly the issue is here, could someone explain? 

Do I need benchmarks for this function? 

Also, a recommendation on what the PR should be named would be helpful. Any suggestions? Perhaps "add a reduction operation for profile_line"?

I'd like to finish this up if possible. :)
@jni @lagru

skimage/measure/profile.py
lagru: Great catch! That's could have let to confusing bugs down the line.

And I think your review comment would also make a great inline comment. ;) Secondly the flipping axis 1 is only necessary if `reduce_func is None` so I would move that statement into the corresponding if clause. Also maybe use `pixels[:, ::-1]` instead of `flip`? Its subjective but at least I'm more used to that syntax.

skimage/measure/profile.py
```diff
@@ -66,7 +87,15 @@ def profile_line(image, src, dst, linewidth=1,
...
-    intensities = pixels.mean(axis=1)
+    pixels = np.flip(pixels, axis=1)
+
+    if reduce_func is None:
+        intensities = pixels
+    else:
+        try:
+            intensities = reduce_func(pixels, axis=1)
+        except TypeError:  # function doesn't allow axis kwarg
+            intensities = np.apply_along_axis(reduce_func, arr=pixels, axis=1)
```
lagru: I think this line isn't covered by the tests. Could you add a test with `reduce_func=lambda x: x`?

skimage/measure/profile.py
```diff
@@ -56,6 +61,22 @@ def profile_line(image, src, dst, linewidth=1,
...
+    For different reduce_func inputs:
+    >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.mean)
+    array([ 0.66666667,  0.66666667,  0.66666667,  1.33333333])
+    >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.max)
+    array([1, 1, 1, 2])
+    >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.sum)
+    array([2, 2, 2, 4])
+
+    The full array will be returned when reduce_func=None.
+    >>> profile_line(img, (1, 2), (4, 2), linewidth=3, order=0,
```
lagru: I think adding a blank line above this one will fix the CI errors. Without it the parser doesn't recognize this as a code block and is confused by the indent below. Same applies to the paragraph above (line 66).

lagru:
> Do I need benchmarks for this function?

Benchmarks are nice to have but not required.

> Also, a recommendation on what the PR should be named would be helpful. Any suggestions? Perhaps "add a reduction operation for profile_line"?

I like it. :)

skimage/measure/profile.py
```diff
@@ -3,7 +3,8 @@
 
 
 def profile_line(image, src, dst, linewidth=1,
-                 order=1, mode='constant', cval=0.0):
+                 order=1, mode='constant', cval=0.0,
+                 reduce_func=np.mean):
```
lagru:
```suggestion
                 *, reduce_func=np.mean):
```

skimage/measure/profile.py
PinkShnack: Great I'll put in that clause!
I'm used to `np.flip` but I wonder which is more common in skimage? You call it.

By my review do you mean:

> The outputted array with linewidth=1 and reduce_func=None gives an array where the row (axis=1) values are flipped. Here I make this consistent.

PinkShnack:
> There is a merge conflict that needs resolving. Are you familiar enough with git to attempt a rebase or merge? Otherwise I'm happy to do it for you. :)

I will have a go at the rebase first, I assume it's a conflict between my fork and the skimage/master yes?

> Also I'm not sure whether I found a bug. Consider
> 
> ```python
> image = np.arange(12).reshape(4, 3)
> prof = profile_line(image, (0, 1), (3, 1), linewidth=3, reduce_func=None)
> assert_array_equal(image, prof)
> ```
> 
> As I understand it, the returned profile should be identical with the original image but we get this (value at (0, 2) is different):
> 
> ```
> AssertionError: 
> Arrays are not equal
> Mismatch: 8.33%
> Max absolute difference: 2
> Max relative difference: 0.
>  x: array([[ 0,  1,  2],
>        [ 3,  4,  5],
>        [ 6,  7,  8],
>        [ 9, 10, 11]])
>  y: array([[ 0,  1,  0],
>        [ 3,  4,  5],
>        [ 6,  7,  8],
>        [ 9, 10, 11]])
> ```
> 
> Not sure what's causing this yet...

Maybe that is due to how the function handles pixel values near the edge of the image. Default sets `mode='constant'` and `cval=0`, so I assume it is setting the "top-right" pixel value of 2 to a value of 0 because it thinks it is outside the image? As if it's a tilt in the line profile, for example. 

I'll mess around with it to see if setting `cval `to something else will reproduce the error. If it does, how would we go about fixing it, as it is probably due to the spline interpolation (`order`)?

skimage/measure/profile.py
lagru: Yeah, I think an inline comment explaining the reason for reversing the last axis would be really useful.

lagru: I'm more used to `pixels[:, ::-1]` but it doesn't really matter with a comment.

skimage/measure/profile.py
```diff
@@ -63,20 +63,28 @@ def profile_line(image, src, dst, linewidth=1,
...
     >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.mean)
     array([0.66666667, 0.66666667, 0.66666667, 1.33333333])
     >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.max)
     array([1, 1, 1, 2])
     >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.sum)
     array([2, 2, 2, 4])
 
-    The full array will be returned when reduce_func=None.
+    The full array will be returned when reduce_func=None and when reduce_func
+    acts on each pixel value individually.
```
lagru: "full array" doesn't refer to `img` here but the internal array `pixels`, correct? In that case the wording might be a bit misleading. Perhaps replace "full array" with "the unreduced array". 

Oh and as a tip: if you surround parameter names with single backticks \` the name is emphasized when rendered by Sphinx (section "4. Parameters" in [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#sections)).

skimage/measure/profile.py
lagru: > Secondly the flipping axis 1 is only necessary if reduce_func is None so I would move that statement into the corresponding if clause.

Just bringing this to your attention again.

lagru: Actually scratch that! In case `reduce_func` operates on single values flipping the last axis is necessary. So it's fine as it is. :wink:

skimage/measure/profile.py
lagru: > For the back ticks, you mean every time I refer to reduce_func in a comment, I should have it as \`reduce_func\` ?

Yes. It's suggested by [Sphinx](https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html?highlight=default_role#roles) and the recommended style by numpydoc. It's certainly not something enforced in scikit-image but I thought it worth mentioning. :)

Concerning the flipping of the last dimension: you are exactly right. Just forgot to consider functions that don't actually reduce but operate on each element individually.

skimage/measure/profile.py
```diff
@@ -56,6 +61,30 @@ def profile_line(image, src, dst, linewidth=1,
...
+    For different reduce_func inputs:
+
+    >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.mean)
+    array([0.66666667, 0.66666667, 0.66666667, 1.33333333])
+    >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.max)
+    array([1, 1, 1, 2])
+    >>> profile_line(img, (1, 0), (1, 3), linewidth=3, reduce_func=np.sum)
+    array([2, 2, 2, 4])
+
+    The unreduced array will be returned when `reduce_func` is None and when
```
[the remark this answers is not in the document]
PinkShnack: Me fail English, that's unpossible.

What's the next step now? Who/how is it merged?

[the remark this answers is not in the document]
lagru: Once another core developer comes along and agrees with these changes there's nothing stopping her or him from pushing the merge button. We recently added a ["little document"](https://scikit-image.org/docs/stable/skips/1-governance.html#decision-making-process) outlining that process if you're interested. ;)

Issue #4200, opened 2019-10-01
"corner_peaks" Function not Handling Image Borders Correctly


Tea7ime:
I am using the corner peaks function to count the number of objects in an image. I was noticing that it is returning too many peaks close to image borders. As the min_distance parameter is increased, the number of peaks found also increases (along borders). 

I don't believe this is the expected result? I would expect corner_peaks to find only one peak for each of the edge groups with 'min_distance=2'. 

![peak_corners_incorrect](https://user-images.githubusercontent.com/16522814/65978200-44cfff80-e441-11e9-8cb6-49a420041a9d.png)

Expected Result:

![peak_corners_expected](https://user-images.githubusercontent.com/16522814/65978894-77c6c300-e442-11e9-8c90-b8935d3c9d2e.png)

While troubleshooting, I'm seeing that the "skimage.feature.corner_peaks" is trying to set negative index locations on the image to false. When it tries to do this, nothing is actually changed. I believe it should be checking for negative indexes and setting them to 0. 

 For the example below, here are the problem index locations corner_peaks is trying to set to false.

Setting this location to False: [-2: 3,  -2: 3]
Setting this location to False: [-2: 3,  7: 12]
setting this location to False: [2: 7,  -2: 3]
setting this location to False: [6: 11,  -2: 3]

This is why all the peaks along the left side and top are not being reduced to a single point.

I was hoping to contribute a fix for this if you agree that this is an issue? I was able to solve the problem by checking for negatives. 

## Way to reproduce

To recreate unexpected result:

```python
import numpy as np
import matplotlib.pyplot as plt
from skimage.feature import corner_peaks

img = np.zeros((10, 10), dtype=np.float32)
img[0:2, 0] = 1.0
img[1, 1] = 0.5

img[0:2, 9] = 0.8
img[1, 8] = 0.7

img[2:4, 3] = 1.0
img[3, 4] = 0.5

img[2:4, 6] = 1.0
img[3, 5] = 0.5

img[4:6, 9] = 0.7
img[4:6, 0] = 0.7

img[6:8, 3] = 1.0
img[6, 4] = 0.5

img[6:8, 6] = 1.0
img[6, 5] = 0.5

img[8:10, 9] = .9
img[8,8] = 0.5

img[8:10, 0] = 0.6
img[8,1] = 0.5

peaks = corner_peaks(img, min_distance=2, exclude_border=False)

plt.imshow(img, cmap='gray')
[...]

PR #4207, opened 2019-10-04
corner_peaks prevent negative index locations

Tea7ime:
This is a fix for the following issue:
#4200 

Corrects how corner_peaks handles image borders. The issue results in too many peaks being returned. 

If one end of the index range is negative, array values cannot be set to false and no error is thrown. Unit test was also updated to check for this. 

Example Unit Test Output:

Check 1:
```
    corners = corner_peaks(response, exclude_border=False, min_distance=10,
                           threshold_rel=0)
    assert len(corners) == 1
```
test1

Check 2:
```
    corners = corner_peaks(response, exclude_border=False, min_distance=5,
                           threshold_rel=0)
    assert len(corners) == 2
```
test2

Check 3:
```
    corners = corner_peaks(response, exclude_border=False, min_distance=1)
    assert len(corners) == 5
```
test3

Check 4:
```
    corners = corner_peaks(response, exclude_border=False, min_distance=1,
                           indices=False)
    assert np.sum(corners) == 5
```
image

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/corner.py
+++ b/skimage/feature/corner.py
@@ -968,8 +968,8 @@ def corner_peaks(image, min_distance=1, threshold_abs=None, threshold_rel=0.1,
         coords = np.transpose(peaks.nonzero())
         for r, c in coords:
             if peaks[r, c]:
-                peaks[r - min_distance:r + min_distance + 1,
-                      c - min_distance:c + min_distance + 1] = False
+                peaks[max((r - min_distance), 0):r + min_distance + 1,
+                      max((c - min_distance), 0):c + min_distance + 1] = False
                 peaks[r, c] = True
 
     if indices is True:
--- a/skimage/feature/tests/test_corner.py
+++ b/skimage/feature/tests/test_corner.py
@@ -351,17 +351,22 @@ def test_num_peaks():
 def test_corner_peaks():
     response = np.zeros((10, 10))
     response[2:5, 2:5] = 1
+    response[8:10, 0:2] = 1
 
     corners = corner_peaks(response, exclude_border=False, min_distance=10,
                            threshold_rel=0)
     assert len(corners) == 1
 
+    corners = corner_peaks(response, exclude_border=False, min_distance=5,
+                           threshold_rel=0)
+    assert len(corners) == 2
+
     corners = corner_peaks(response, exclude_border=False, min_distance=1)
-    assert len(corners) == 4
+    assert len(corners) == 5
 
     corners = corner_peaks(response, exclude_border=False, min_distance=1,
                            indices=False)
-    assert np.sum(corners) == 4
+    assert np.sum(corners) == 5
 
 
 def test_blank_image_nans():
--- a/skimage/feature/tests/test_corner.py
+++ b/skimage/feature/tests/test_corner.py
@@ -355,14 +355,14 @@ def test_corner_peaks():
 
     corners = corner_peaks(response, exclude_border=False, min_distance=10,
                            threshold_rel=0)
-    assert len(corners) == 1
+    assert corners.shape == (1, 2)
 
     corners = corner_peaks(response, exclude_border=False, min_distance=5,
                            threshold_rel=0)
-    assert len(corners) == 2
+    assert corners.shape == (2, 2)
 
     corners = corner_peaks(response, exclude_border=False, min_distance=1)
-    assert len(corners) == 5
+    assert corners.shape == (5, 2)
 
     corners = corner_peaks(response, exclude_border=False, min_distance=1,
                            indices=False)


skimage/feature/tests/test_corner.py
```diff
@@ -351,17 +351,22 @@ def test_num_peaks():
...
                            threshold_rel=0)
     assert len(corners) == 1
 
+    corners = corner_peaks(response, exclude_border=False, min_distance=5,
+                           threshold_rel=0)
+    assert len(corners) == 2
+
     corners = corner_peaks(response, exclude_border=False, min_distance=1)
-    assert len(corners) == 4
+    assert len(corners) == 5
```
lagru: Because you're already touching this test, would you mind asserting the full shape instead of only the first dimension via the `len()` function? E.g.

```
assert corners.shape == (5, 2)
```
should be more readable.

Tea7ime: @lagru Thank you! No problem at all, that change makes sense to me. Just pushed your revision up.

skimage/feature/corner.py
```diff
@@ -968,8 +968,8 @@ def corner_peaks(image, min_distance=1, threshold_abs=None, threshold_rel=0.1,
         coords = np.transpose(peaks.nonzero())
         for r, c in coords:
             if peaks[r, c]:
```
rfezzani: I don't think that this check is necessary since `r, c` comes from `peaks.nonzero()`?

Tea7ime: I guess the problem is in how we are doing subtraction here. I've been printing out the index locations corner_peaks is trying to set. Let's say peak_local_max returns two connected peaks to corner_peaks at [0:1, 0:2], and we have min_distance=2 for example.
 
The returned values from "coords = np.transpose(peaks.nonzero())":
 [[0 0]
 [0 1]]

Now, we get this: 
`Rows = [0-2: 0+2+1],  Cols = [0-2: 0+2+1]`

And now we try to do this:
`img[-2:3, -2:3] = false`

It's not able to set those locations to false. We should end up with one peak here and instead we get two.

test4


To Reproduce:
```
import numpy as np
from skimage.feature import corner_peaks


img = np.zeros((5, 5), dtype=np.float32)
img[0:1, 0:2] = 1

corners = corner_peaks(img, exclude_border=False, min_distance=2)

plt.imshow(img, cmap='gray')
plt.plot(corners[:, 1], corners[:, 0], 'bo', markersize=2)
plt.show()
```

rfezzani: Ah, you're right! Sorry :-p

rfezzani: May we then use `scipy.spatial.distance.pdist` to find the points that are too close to each others instead of modifying the `peaks` array iteratively?

Tea7ime: @rfezzani No worries, thanks for reviewing! I like the idea. It's interesting how it actually goes about updating the array. 

Is there a performance benefit to using `scipy.spatial.distance.pdist` for this? We need to guarantee we are grabbing the top left corner within the min_distance region. So, I believe we'd still have to iterate through the entire peaks array, but this time do additional comparisons to min_distance. My thought is that we would still be modifying the array, only this time more selectively?

Tea7ime:
@lagru Thank you for reviewing!

skimage/feature/corner.py
rfezzani: Hi @Tea7ime, sorry for my late answer. I played a bit with the `pdist` function.

```python
def corner_peaks_new(image, min_distance=1, threshold_abs=None,
                     threshold_rel=0.1, exclude_border=True, indices=True,
                     num_peaks=np.inf, footprint=None, labels=None):

    peaks = peak_local_max(image, min_distance=min_distance,
                           threshold_abs=threshold_abs,
                           threshold_rel=threshold_rel,
                           exclude_border=exclude_border,
                           indices=False, num_peaks=num_peaks,
                           footprint=footprint, labels=labels)
    if min_distance > 0:
        coords = np.transpose(peaks.nonzero())
        num_peaks = coords.shape[0]
        # Compute the distance between each detected distance
        dist_mat = pdist(coords, 'chebyshev')
        rejected_peaks = set()
        start = end = 0
        step = num_peaks - 1
        for i in range(num_peaks - 2):
            end += step
            if i in rejected_peaks:
                start = end
                step -= 1
                continue
            rejected_peaks.update(i + 1 + np.nonzero(dist_mat[start:end] <
                                                     min_distance)[0])
            start = end
            step -= 1

        # Remove the peaks that are too close to each others
        rejected_peaks = list(rejected_peaks)
        if indices is True:
            return np.delete(coords, rejected_peaks, axis=0)

        peaks[coords[rejected_peaks, :]] = False

    if indices is True:
        return np.transpose(peaks.nonzero())

    return peaks
```

This implementation is slower then the actual one (but still fast), but it allows to use different distances: the actual implementation corresponds to the Chebyshev distance (it considers a square area around the peaks), but `pdist` allows to use the Euclidean distance (a circle area around the peaks).

rfezzani: I know that it is actually not the main concern of this PR. I will fill a new one instead of adding noise to this one... Sorry

hmaarrfk:
Should this get backported to 0.16? 0.14?

lagru:
I'd say so. :+1:

lagru:
@Tea7ime Thank you for this!

Issue #4135, opened 2019-09-03
Functions are not sorted in API guide


stefanv:
See, e.g., https://scikit-image.org/docs/dev/api/skimage.color.html

hmaarrfk:
how should they be sorted? by themes or alphabetical?

stefanv:
Alphabetical, I think.

soupault:
I have not thotoughly checked, but it might be that the order in Sphinx follows precisely the one in `__init__.py` - https://github.com/scikit-image/scikit-image/blob/master/skimage/color/__init__.py#L1-L67.

stefanv:
Yes, that's likely the case.

dhirensr:
should i send a PR for sorting the functions alphabetically?

soupault:
@dhirensr sure, go ahead!

dhirensr:
@soupault : should i modify the all variable in __init__.py - https://github.com/scikit-image/scikit-image/blob/master/skimage/color/__init__.py#L1-L67 or the import order?
also how do i test locally before sending PR?

jni:
@dhirensr you could try `_all_exported = [ ... ]  # the current list` and then `__all__ = sorted(_all_exported)`

To test locally, you should `cd doc`, then `make html`. You can then open the resulting page in `doc/build/html/index.html`. You will need to install everything in `requirements/docs.txt`, and possibly also pandoc and latex.

dhirensr:
@jni : I am not sure if it picks up from modules __init__.py because I removed an element from __all__ and it did not remove it from docs.
also I did make clean and make html then also It did not remove the doc.
I am also checking where does it generate from! do you have any idea? @jni @soupault

jni:
@dhirensr Hmm. I'm not actually sure. @stefanv?

dhirensr:
@jni : sorry my bad! it works fine with __init__ file
have made a PR https://github.com/scikit-image/scikit-image/pull/4208 for review
also do we change all other modules where the functions are not sorted?

mkcor:
Closing this, since functions / classes / submodules are now sorted ~alphabetically~ lexicographically.

Ultimately, the API reference should boast a new look (maybe using NumPy's or SciPy's theme?). But this would happen after 1.0 anyway.

PR #4208, opened 2019-10-05
Sort the generated API documentation alphabetically

dhirensr:
## Description
related to documentation 
https://github.com/scikit-image/scikit-image/issues/4135

## Checklist

tested it locally 

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/tools/apigen.py
+++ b/doc/tools/apigen.py
@@ -268,6 +268,9 @@ class ApiDocWriter(object):
         if not (len(functions) or len(classes) or len(submodules)) and DEBUG:
             print('WARNING: Empty -', uri)
             return ''
+        functions = sorted(functions)
+        classes = sorted(classes)
+        submodules = sorted(submodules)
 
         # Make a shorter version of the uri that omits the package name for
         # titles


hmaarrfk:
Is the goal here that they be sorted in the documentation? Or for other purposes? This seems like it hinges on boilerplate code that makes things cluttered. Can we push this kind of stuff to a linter, or to the documentation generation

dhirensr:
@hmaarrfk : the goal is to get the documentation functions sorted! According to this ticket , https://github.com/scikit-image/scikit-image/issues/4135 the feature was requested so I made the changes!
i didn't understood completely the last 2 questions! do i need to make any changes?

hmaarrfk:
I was following the discussion. But it seems like a "noisy" in terms of it scaling. I'm just trying to think out load for an alternate solution

hmaarrfk:
There seems to be a way for aphinx to do it. http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_member_order

My guess is that it was maybe turned off at some point for some reason.

emmanuelle:
Hum, actually it could make sense not to have the functions sorted alphabetically, but by decreasing importance instead... (even if the ranking could be hard to do). If we go alphabetically, I agree with @hmaarrfk that using the sphinx option is tidier.

jni:
I think for now we should go with alphabetical. I expect that people usually visit the API docs looking for a specific function, not browsing for "important" functions. After they are alphabetically sorted, we can maybe find a way to highlight important functions. A couple of suggestions to programmatically rank the functions:

- by number of uses in [this post](https://labs.quansight.org/blog/2019/05/python-package-function-usage/)
- by number of uses in our examples and/or tutorials

But, as I mentioned, I think that's for later, and sorting them alphabetically would already be an improvement.

@dhirensr sorry to have given you bad advice in #4135! Would you mind modifying this PR or submitting a new one using the sphinx configuration solution suggested by @hmaarrfk?

dhirensr:
@jni @hmaarrfk : i will look at it and send a PR in a week!

hmaarrfk:
@emanuelle. I think the idea of having hand tuned orders is why the sphinx default was changed.

dhirensr:
@jni : can you please guide me which file has the configurations? I tried to change .rst files but i think those are not the files with config

jni:
@dhirensr sorry for the slow response. I myself am quite unfamiliar with the documentation machinery, so I had to think about this. I think the file you need to edit as pointed to by @hmaarrfk is `doc/source/conf.py`.

dhirensr:
@jni : i tried inserting autodoc_member_order='alphabetical' but it doesn't changes anything and I am also a noob with sphinx so don't exactly know. and according to the docs alphabetical is anyway the default

dhirensr:
@lagru : but as @jni @hmaarrfk pointed out !scalability accross all files could be a problem!
@jni @hmaarrfk : what do you'll say on the above solution?

hmaarrfk:
why does @lagru's not scale? seems OK.

That raises the questions: is that custom sphinx generation still necessary. A question for an other day.

lagru:
We might get something equivalent by using Sphinx's built-in extension [autosummary](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html). Its ability to [customize its templates](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#customizing-templates) makes it somewhat flexible. I might be interested to give it a try when I find the time. It might be easier to maintain in the long term.

dhirensr:
@lagru : try to send a PR then maybe we close this ticket.

lagru:
@dhirensr Replacing the functionality provided by doc/tools would likely be a long term project. So I think there is still value in exploring the approach suggested in https://github.com/scikit-image/scikit-image/pull/4208#issuecomment-545816630 (provided that you are interested)! I didn't want to discourage you. :wink:

dhirensr:
@lagru @hmaarrfk @jni : please review! updated the PR as suggested by @lagru and it worked smoothly

doc/tools/apigen.py
```diff
@@ -268,6 +268,9 @@ def generate_api_doc(self, uri):
         if not (len(functions) or len(classes) or len(submodules)) and DEBUG:
             print('WARNING: Empty -', uri)
             return ''
+        functions = sorted(functions)
```
lagru: Maybe we should make this behavior toggle-able. I would add a flag as a class attribute to `ApiDocWriter`. What do others think?

skimage/color/__init__.py
```diff
@@ -68,64 +68,64 @@
...
-           'rgb_from_hpx',
-           'hpx_from_rgb',
-           'color_dict',
-           'label2rgb',
-           'deltaE_cie76',
-           'deltaE_ciede94',
-           'deltaE_ciede2000',
-           'deltaE_cmc',
-           ]
+                 'guess_spatial_dimensions',
```
lagru: It looks like you left an indent while reverting this file. Could you remove it so that this file isn't changed by this PR at all?

doc/tools/apigen.py
hmaarrfk: Seems complicated. Either we want things sorted or not. Let's leave it at that.

lagru: Works for me. :+1:

hmaarrfk:
Thanks!

jni:
Thank you @dhirensr for your persistence with this!

PR #4209, opened 2019-10-05
Refactor/move neighborhood utility functions in morphology

lagru:
## Description

Changes are extracted from #4165 hopefully making reviewing these easier. 

- This cleans up several functions `_offsets_to_raveled_neighbors`, `_fast_pad`, `_set_edge_values_inplace` and `_resolve_neighborhood` and consolidates them in the `_util` module (some where already there). These functions are used together at several places in the `morphology` module.
- Fixes a small bug that blocks #4165: `_offsets_to_raveled_neighbors` would fail in case a dimension in the structuring element was larger than in the given `image_shape`. While this use case isn't necessary for algorithms that work by padding images before constructing the raveled offsets (e.g. `flood_fill`, `local_maxima`) this precludes other functions to work with images where a dimension is < 3 (typical length of the structuring element). With the new implementation this works even if the structuring element is larger than the image itself in any dimension.

## For reviewers

  later.


skimage/morphology/_util.py
```diff
@@ -51,35 +54,182 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+    offsets = selem_indices - center
+
+    if order == 'F':
+        offsets = offsets[:, ::-1]
+        image_shape = image_shape[::-1]
+    elif not order == 'C':
+        raise ValueError("order was not 'C' or 'F'")
+
+    # Scale offsets in each dimension and sum
+    ravel_factors = image_shape[1:] + (1,)
```
rfezzani: if `image_shape` is allowed to be any sequence (as you modified it in the doc), it has to be casted to `tuple` for this line to be correct.

skimage/morphology/_util.py
```diff
@@ -51,35 +54,182 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
-    structure = structure.copy()  # Don't modify original input
-    structure[tuple(center)] = 0  # Ignore the center; it's not a neighbor
-    connection_indices = np.transpose(np.nonzero(structure))
-    offsets = (np.ravel_multi_index(connection_indices.T, image_shape,
-                                    order=order) -
-               np.ravel_multi_index(center, image_shape, order=order))
-    squared_distances = np.sum((connection_indices - center) ** 2, axis=1)
-    return offsets[np.argsort(squared_distances)]
+    selem_indices = np.array(np.nonzero(selem)).T
+    offsets = selem_indices - center
```
rfezzani: May be add a check for `center` length to be equal to `selem` number of dimensions.

skimage/morphology/_util.py
lagru: Good catch! I reverted that change. Giving the shape as anything but a tuple is non-standard anyway.

skimage/morphology/_util.py
lagru: Good idea. And a check for `image_shape` as well.

skimage/morphology/_util.py
```diff
@@ -51,35 +54,188 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+    Depending on the values of `connectivity` and `selem` this function
+    either creates a new structuring element (`selem` is None) using
+    `connectivity` or validates the given structuring element (`selem` is not
+    None).
+
+    Parameters
+    ----------
+    selem : array-like or None
+        The structuring element to validate. See same argument in
+        `local_maxima`.
```
sciunto: If  we move this function to util, this statement must be further explained.

skimage/morphology/_util.py
```diff
@@ -51,35 +54,188 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+        _, indices = np.unique(raveled_offsets, return_index=True)
+        raveled_offsets = raveled_offsets[np.sort(indices)]
+
+    # Remove "offset to center"
+    raveled_offsets = raveled_offsets[1:]
+
+    return raveled_offsets
+
+
+def _resolve_neighborhood(selem, connectivity, ndim):
```
sciunto: If I'm correct, this function has no doctest nor unitest. That would deserve to open an issue.

sciunto:
Can you add a test that failed for  #4165 please?

skimage/morphology/_util.py
lagru: All of the functions in this module are only implicitly tested through tests for the public functions. I'm hesitant to add detailed tests here because these are just convenience functions (excluding `_offsets_to_raveled_neighbors`) dealing with input and I'd rather test the public API than parts of the implementation that may change significantly down the line. E.g. `_fast_pad` may be safely removed once NumPy 1.17 is given by scikit-image's dependencies.

However `_offsets_to_raveled_neighbors` could indeed profit from a few explicit tests. I'll try to come up with a few meaningful tests...

skimage/morphology/_util.py
```diff
@@ -51,35 +54,206 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+        raise ValueError("order was not 'C' or 'F'")
+
+    # Scale offsets in each dimension and sum
+    ravel_factors = image_shape[1:] + (1,)
+    ravel_factors = np.cumprod(ravel_factors[::-1])[::-1]
+    raveled_offsets = (offsets * ravel_factors).sum(axis=1)
+
+    # Sort by distance
+    distances = np.abs(offsets).sum(axis=1)
+    raveled_offsets = raveled_offsets[np.argsort(distances)]
```
lagru: I haven't been able to figure out a way to keep the same order for neighbors of the same connectivity. Maybe that's due to `argsort` defaulting to the unstable `quicksort`? This does affect in which order algorithms iterate over neighbors with the same distance. I don't think this relevant for the algorithms relying on this function though and the test suite seems to back this up...

skimage/morphology/_util.py
lagru: Done.

skimage/morphology/_util.py
lagru: Done as well.

lagru:
@emmanuelle, I rebased this as discussed in the meeting. :)

skimage/morphology/_flood_fill.py
```diff
@@ -7,9 +7,8 @@
 import numpy as np
 from warnings import warn
 
-from .extrema import (_resolve_neighborhood, _set_edge_values_inplace,
-                      _fast_pad)
-from ._util import _offsets_to_raveled_neighbors
+from ._util import (_resolve_neighborhood, _set_edge_values_inplace,
+                    _fast_pad, _offsets_to_raveled_neighbors)
```
emmanuelle: this will only be used in your other PR, right?

skimage/morphology/_util.py
```diff
@@ -51,35 +54,206 @@ def _validate_connectivity(image_dim, connectivity, offset):
     return c_connectivity, offset
 
 
-def _offsets_to_raveled_neighbors(image_shape, structure, center, order='C'):
+def _offsets_to_raveled_neighbors(image_shape, selem, center, order='C'):
```
emmanuelle: Thanks for the change for a better consistency. We should definitely make this list of variable names in the contributing notes :-).

skimage/morphology/_util.py
emmanuelle: Do I understand correctly that the problem is that the order of indices is not always the same? (lack of reproducibility)

If yes, why not add another information to disambiguate the distance, such as a raveled index within the structuring element? You could then use the `order` parameter of `argsort`, see for example
```
>>> a = np.array([(2, 10), (2, 5), (1, 3), (1, 2), (3, 1)], dtype=[('distance', 'f8'), ('index', 'f8')])                                                
>>> np.sort(a, order=['distance', 'index'])                                                                                                             
array([(1.,  2.), (1.,  3.), (2.,  5.), (2., 10.), (3.,  1.)],
      dtype=[('distance', '<f8'), ('index', '<f8')])
>>> np.argsort(a, order=['distance', 'index'])                                                                                                          
array([3, 2, 1, 0, 4])
```

skimage/morphology/_util.py
```diff
@@ -51,35 +54,206 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+    padded_image : ndarray
+        The new image.
+
+    Notes
+    -----
+    The output of this function is equivalent to::
+
+        np.pad(image, 1, mode="constant", constant_values=value)
+
+    Up to versions < 1.17 `numpy.pad` uses concatenation to create padded
```
emmanuelle: Add this to TODO.txt? Otherwise we'll forget about it when we bump to numpy >= 1.17.

skimage/morphology/_util.py
```diff
@@ -51,35 +54,206 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+    ravel_factors = np.cumprod(ravel_factors[::-1])[::-1]
+    raveled_offsets = (offsets * ravel_factors).sum(axis=1)
+
+    # Sort by distance
+    distances = np.abs(offsets).sum(axis=1)
+    raveled_offsets = raveled_offsets[np.argsort(distances)]
+
+    # If any dimension in image_shape is smaller than selem.shape
+    # duplicates might occur, remove them
+    if any(x < y for x, y in zip(image_shape, selem.shape)):
```
emmanuelle: couldn't you just add this part to the existing function? I'm curious to know why you modified the rest of the function, since I find it harder to read than the original one. It might be only a matter of taste of course...

emmanuelle:
Thanks @lagru , I left a couple of questions for you.

skimage/morphology/_util.py
jni: Sorry, can I be the annoying objector here?

I think originally we stopped using selem because it is a very obscure term for people who are not familiar with image processing, and morphology in particular. Structure is a bit better, but imho "footprint" or "neighborhood" are the clearest. What do we think? Consistency is great, but let's think about what the best syntax is going forward, and be consistent with that.

rfezzani: I have no clear preference concerning the best argument name for this, I just want to draw attention to the existence of the `skimage.morphology.selem` module.

skimage/morphology/_util.py
lagru: Short answer: No. :wink: I probably should have done a better job highlighting the problem and what I did about it. Consider the following scenario:

I have an `image` of shape `(10, 2)` and want to compute the neighborhood offsets for a structuring element with the shape `(3, 3)` with center at `(1, 1)`. This is a valid use case we need to anticipate because we might want the algorithm be able to step from `image[i, 0]` to `image[i, 1 ]` *and* vice versa. For that we need neighbors on both sides of the center and thus a dimension length of 3.

Now the old implementation would fail because [np.ravel_multi_index](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ravel_multi_index.html) doesn't support this use case:
```python
import numpy as np
selem = np.ones((3, 3), dtype=bool)
np.ravel_multi_index(np.nonzero(selem), dims=(10, 2))
# Raises Exception: ...
#     np.ravel_multi_index(np.nonzero(selem), image.shape)
#  File "<__array_function__ internals>", line 6, in ravel_multi_index
# ValueError: invalid entry in coordinates array
np.ravel_multi_index(np.nonzero(selem), dims=(10, 3))
# Out: array([0, 1, 2, 3, 4, 5, 6, 7, 8])
```
So the previous implementation wasn't able to deal with this case because it relied on that function to compute the offsets. Unfortunately the `mode` argument is not helpful here because wrapping or clipping is not what we want. I tried adjusting the indices by the center index beforehand but again `ravel_multi_index` doesn't handle negative indices the way we want to.

But the new algorithm will lead to duplicate neighbors. I find it really hard to think about this but I'll try to explain with an example. We have the structuring element with the shape `(3, 3)` all ones and the center `(1, 1)` and a C-ordered image with the shape `(6, 2)`. After adjusting for the center we get the raveled offsets (representation with index pairs ordered by dimension):
```python
[[(-1, -1), (-1,  0), (-1,  1)],
 [( 0, -1), ( 0,  0), ( 0,  1)],
 [( 1, -1), ( 1,  0), ( 1,  1)]]
```
The second value in a pair denotes the step in the last dimension which is contiguous (-> ravel factor 1). But because samples in the first dimension don't border each other we need to address this by multiplying with the length of the previous dimension (-> ravel factor 2). For `(0, 1)` and `(1, -1)` this results `(0, 1)` and `[2, -1]` which both resolve (by summing) to the raveled index 1. Considering the array
```python
[[ 0,  1],
 [ 2,  3],
 [ 4,  5],
 [ 6,  7],
 [ 8,  9],
 [10, 11]]
```
this means that the pixel to the right of `6` is `7` and to the lower-right is `7` as well! This "memory-wise wrapping" when crossing image boundaries might be counter-intuitive to simply wrapping around in one dimension. Though, for the functions using this result it's not relevant because they always check that the image edge is not crossed while iterating the neighborhood.

It's also interesting to switch the image shape `(2, 7)` around (or use F-order) in which case we get no duplicates value wise `[-7, -6, -5, -1, 0, 1, 5,  6, 7]`. However we get the same wrapping behavior.

Okay, I hope that helped. :sweat_smile: 

Though, I wholeheartedly agree that the old implementation is simpler, easier to read and probably to maintain but I saw no way around it.

skimage/morphology/_util.py
lagru: Definitely!

skimage/morphology/_util.py
lagru: @jni I like `neighborhood`. But think this should be a separate PR. In any case I think `selem` is better than the old `structure` because of the afore mentioned consistency. So until we change this everywhere I'd leave it as `selem`. 

Not a strong opinion though. I'll change it if you find a silent consensus.

sciunto: This function is private. The choice is not critical at the end. :)

skimage/morphology/_flood_fill.py
lagru: No? Not sure what you are referring to.

`_resolve_neighborhood`, `_set_edge_values_inplace` and  `_fast_pad` are used in `_flood_fill.py` and `extrema.py`. While `_offsets_to_raveled_neighbors` are additionally used in `max_tree.py` and `watershed.py`.
#4165 will rely on `_resolve_neighborhood` and `_offsets_to_raveled_neighbors`.

skimage/morphology/_util.py
lagru: > Do I understand correctly that the problem is that the order of indices is not always the same? (lack of reproducibility)

No it's always the same and reproducable but changes between the old and new implementation. E.g. old:
```python
>>> _offsets_to_raveled_neighbors((3, 3, 3), np.ones((3, 3, 3)), (1, 1, 1))
array([ -1,   3,  -9,   1,   9,  -3,   4,   2,  12,  -2,  -4,  -6,  10,
        -8, -10, -12,   8,   6,  11, -13,   5,  -5,  -7, -11,   7,  13])
```
and new:
```python
>>> _offsets_to_raveled_neighbors((3, 3, 3), np.ones((3, 3, 3)), (1, 1, 1))
array([  3,  -9,   1,  -1,   9,  -3,   4,   2,  12,  -2,  -4,  -6,  10,
        -8, -10, -12,   8,   6,  11, -13,   5,  -5,  -7, -11,   7,  13])
```

I vaguely remember fiddling around with  the `kind=` argument to `argsort`. I think we want a stable sort that preserves the order of input arguments (which is memory-order) as long as they have the same distance. So `kind="mergesort"` and switch to `kind="stable"` once we hit NumPy 1.15?

skimage/morphology/_util.py
jni: Very interesting! Thanks for the exposé, @lagru! Stupid edge cases ruining our otherwise elegant lives... =P I'll try to do a proper review of this PR in the next couple of days... Thanks!

lagru: See also https://github.com/numpy/numpy/issues/15475.

skimage/morphology/_util.py
jni: Yes. LOL Sorry for the long delay in responding!

skimage/morphology/_util.py
```diff
@@ -51,35 +54,206 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+    """Validate or create structuring element.
+
+    Depending on the values of `connectivity` and `selem` this function
+    either creates a new structuring element (`selem` is None) using
+    `connectivity` or validates the given structuring element (`selem` is not
+    None).
+
+    Parameters
+    ----------
+    selem : ndarray, optional
```
jni: As currently implemented, this is not optional! LOL

skimage/morphology/_util.py
```diff
@@ -51,35 +54,206 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+        selem = np.asarray(selem, dtype=np.bool)
+        # Must specify neighbors for all dimensions
+        if selem.ndim != ndim:
+            raise ValueError(
+                "structuring element and image must have the same number of "
+                "dimensions"
+            )
+        # Must only specify direct neighbors
+        if any(s != 3 for s in selem.shape):
+            raise ValueError("dimension size in structuring element is not 3")
```
jni: I recently wanted to specify a structuring element as a flat disk, but was forced by scipy.ndimage to pad it with zeros on the flat axis, which I found annoying. I wonder if, for the use cases envisioned by this function, we can relax this requirement...

skimage/morphology/_util.py
```diff
@@ -51,35 +54,206 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+    >>> image
+    array([[1, 1, 1, 1, 1],
+           [1, 0, 0, 0, 1],
+           [1, 0, 0, 0, 1],
+           [1, 1, 1, 1, 1]])
+    """
+    for axis in range(image.ndim):
+        # Index first and last element in each dimension
+        sl = (slice(None),) * axis + ((0, -1),) + (...,)
+        image[sl] = value
```
jni: This function is nice! I would rename it `_set_border_values` or `_set_boundary_values`. Edge has the connotation of "edge filters" in my head. We use "border" elsewhere in the library.

skimage/morphology/_util.py
```diff
@@ -51,35 +54,206 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+    new_shape = np.array(image.shape) + 2
+    new_image = np.empty(new_shape, dtype=image.dtype, order="C")
+
+    # Copy old image into new space
+    sl = (slice(1, -1),) * image.ndim
+    new_image[sl] = image
+    # and set the edge values
+    _set_edge_values_inplace(new_image, value)
+
+    return new_image
```
jni: Very nice! =)

jni: LOL Just saw that this already existed. :joy:

skimage/morphology/tests/test_util.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+import pytest
+import numpy as np
+from numpy.testing import assert_array_equal
+
+from skimage.morphology import _util
+
+
+class TestOffsetsToRaveledNeighbors:
```
jni: Why is this a class???

skimage/morphology/_util.py
lagru: You are suggesting that this function would automatically replace `selem=np.array([[1, 1, 1]])` with `selem=np.array([[0, 0, 0], [1, 1, 1], [0, 0, 0]])`? I'm not convinced that the convenience is worth it. 

- The behavior for different values to `selem` and `connectivity` already requires extensive checks and documentation without adding the suggested behavior.
- The suggested behavior seems a little arbitrary. Even though it makes sense for your given example, another user may expect `[[1, 1, 1], [1, 1, 1], [1, 1, 1]]`.

I'd rather be explicit to avoid as much confusion as possible.

skimage/morphology/_util.py
lagru: Great suggestion, will do.

skimage/morphology/_util.py
lagru: Well I did tweak the way slices are constructed here and `_set_edge_values_inplace`. So it doesn't hurt that you looked at it again with new eyes. :smile:

skimage/morphology/_util.py
jni: I disagree. The behaviour of such a footprint is perfectly well defined in the context of convolution. This would be exactly analogous, *and*, I’m not asking for it to be expanded into anything, only to be allowed. The offsets function will work perfectly fine with that syntax.

skimage/morphology/tests/test_util.py
lagru: To be honest it's just the way got introduced using Pytest. One might argue that is allows more descriptive test names and grouping. Prepending "offsets_to_raveled_neighbors" to every test would have resulted in quite long names in this case. I'm happy to change it though.

jni: To me it’s quite un-skimage-y... =P grouping can be achieved by python modules. `test_util` -> `test_offsets_to_neighbors`

lagru: Not sure how to feel about that. I always liked that the test structure reflected to the implementation structure. That way I immediately knew where to look for the corresponding function/tests. Is there already a precedent for your suggestion?

jni: There’s definitely not a consistent 1-1 mapping currently in the library, it’s just gotten close to that because it’s the easiest way to get started. There’s precedent here at least: https://github.com/scikit-image/scikit-image/tree/master/skimage/graph/tests

Whether it’s the right thing is another matter, but it doesn’t seem like a big deal to me. You should still be able to find modules by function name. And indeed nothing is stopping you from making a nested folder!

skimage/morphology/_util.py
lagru: Ah, that is worth thinking about. Sorry for misunderstanding. Currently we expect all dimensions in `selem` to be of length 3 and assume that the center is always at `(1,) * ndim` (e.g. in `flood_fill`, `local_maxima`). In `watershed` and `max_tree` `_validate_connectivity` is used which raises if there is an ambiguous center (`selem` has dimension with even length). That might work here as well. But I'd prefer to do that in a separate PR...

jni: Ok, I suppose adding more permissive behaviour in the future is totally fine. =)

sciunto:
@lagru would you like to resolve the conflicts? @skimage-core This PR is MRG+1 for a while, we need a second review here please :)

sciunto:
@scikit-image/core This PR is MRG+1 for a while, we need a second review here please :)

skimage/morphology/_util.py
lagru: I opened #4567 as a reminder.

lagru:
@sciunto done. Thanks for bringing this to attention.

skimage/morphology/tests/test_util.py
```diff
@@ -0,0 +1,127 @@
...
+    desired = np.array([
+          3, -600,    1,   -1,  600,   -3,    4,    2,  603,   -2,   -4,
+        -597,  601, -599, -601, -603,  599,  597,  602, -604,  596, -596,
+        -598, -602,  598,  604
+    ])
+    assert_array_equal(offsets, desired)
+
+
+def test_offsets_to_raveled_neighbors_explicit_1():
+        """Check reviewed example where selem is larger in last dimension."""
```
rfezzani, line 108: Why that much indentation?

skimage/morphology/_util.py
```diff
@@ -51,35 +54,208 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+            connectivity = ndim
+        selem = ndi.generate_binary_structure(ndim, connectivity)
+    else:
+        # Validate custom structured element
+        selem = np.asarray(selem, dtype=np.bool)
+        # Must specify neighbors for all dimensions
+        if selem.ndim != ndim:
+            raise ValueError(
+                "structuring element and image must have the same number of "
+                "dimensions"
```
rfezzani, line 178:
```suggestion
                "Structuring element and image must have the same number of "
                "dimensions."
```

skimage/morphology/_util.py
```diff
@@ -51,35 +54,208 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
+        )
+
+    selem_indices = np.array(np.nonzero(selem)).T
+    offsets = selem_indices - center
+
+    if order == 'F':
+        offsets = offsets[:, ::-1]
+        image_shape = image_shape[::-1]
+    elif order != 'C':
+        raise ValueError("order was not 'C' or 'F'")
```
rfezzani, line 105:
```suggestion
        raise ValueError("order must be 'C' or 'F'.")
```

skimage/morphology/_util.py
```diff
@@ -51,35 +54,208 @@ def _validate_connectivity(image_dim, connectivity, offset):
...
     >>> _offsets_to_raveled_neighbors((4, 5), np.ones((4, 3)), (1, 1))
     array([-5, -1,  1,  5, -6, -4,  4,  6, 10,  9, 11])
+    >>> _offsets_to_raveled_neighbors((2, 3, 2), np.ones((3, 3, 3)), (1, 1, 1))
+    array([ 2, -6,  1, -1,  6, -2,  3,  8, -3, -4,  7, -5, -7, -8,  5,  4, -9,
+            9])
+    """
+    if not selem.ndim == len(image_shape) == len(center):
+        raise ValueError(
+            "number of dimensions in image_shape, selem and center "
+            "does not match"
```
rfezzani, line 95:
```suggestion
            "Number of dimensions in image_shape, selem and center "
            "does not match."
```

skimage/morphology/tests/test_util.py
lagru: Previously I had a class wrapping these tests until @jni [pointed out](https://github.com/scikit-image/scikit-image/pull/4209#discussion_r372932297) that that may be overdoing it. The leftover indent is an oversight on my part.

skimage/morphology/_util.py
lagru: This deliberately emulates what error messages from Python and NumPy itself look like: a concise statement about "what's wrong" without punctuation. So I'd prefer to keep it this way. Same for some of you other comments.

lagru:
@rfezzani thanks for the feedback. I think the discrepancy between the "connectivity" parameter in the watershed function and its usage everywhere else is confusing and we should address this. I'll probably touch the related error messages in a future PR anyway because that would allow more descriptive ones.

rfezzani:
Thank you @lagru :tada:

PR #4222, opened 2019-10-09
Fix Travis linux build by moving from Ubuntu trusty to xenial

rfezzani:
## Description

Ubuntu trusty having [reached the end of its LTS window](https://ubuntu.com/blog/ubuntu-14-04-trusty-tahr), I think that it is time to move to xenial!
This may solve the issues with Travis CI.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@
 
 # See tools/travis/notes.txt for some guidelines
 
-# If this is ever changed to xenial (Ubuntu 16.04), make sure to enable the
-# xvfb service
-dist: trusty
+dist: xenial
+services:
+  - xvfb
 language: python
 cache:
   # See https://docs.travis-ci.com/user/caching/#pip-cache
@@ -56,29 +56,17 @@ matrix:
     - os: linux
       python: 3.7
       env: QT=PyQt5 OPTIONAL_DEPS=1 BUILD_DOCS=1 DEPLOY_DOCS=1
-      dist: xenial # Required for Python 3.7
-      services:
-        - xvfb
     - os: linux
       python: 3.7
-      dist: xenial # Required for Python 3.7
       env: QT=PyQt5 OPTIONAL_DEPS=1 BUILD_DOCS=1
-      services:
-        - xvfb
     # When 3.8 gets released, and dependencies have built wheels, enable
     # this build
     # - os: linux
     #   python: 3.8
     #   env: QT=PyQt5 OPTIONAL_DEPS=1 BUILD_DOCS=1
-    #   dist: xenial # Required for Python 3.7
-    #   services:
-    #     - xvfb
     - os: linux
       python: 3.7
       env: QT=PyQt5 OPTIONAL_DEPS=1 PIP_FLAGS="--pre"
-      dist: xenial # Required for Python 3.7
-      services:
-        - xvfb
     # For smooth deployment, the osx_image here should match
     # what we set in the wheel generation travis images.
     # If not set, it will use the default version from Travis


jni:
Thanks @rfezzani!

Any chance we might want to go straight to 18.04 bionic? What's the idea with targeting very old distros? CC @stefanv

rfezzani:
Xenial being LTS, it is still widely used. I think it is better to ensure that skimage builds correctly with this distribution.

matthew-brett:
Mark - just checking what you mean - you think scikit-image should build and test on Trusty to make sure that it still works on Trusty?

At the moment, it's true that the wheel builds default to testing on Trusty - so can't you check the Trusty compatibility using the wheel-building repo - and move ahead to Xenial testing on the main repo?

hmaarrfk:
The issue occurs that the wheel repo isn't looked at every day. So if we start to depend on some newer feature of libc (unlikely, but possible), then we would only catch the issue at the time we attempt to build the final wheels. This is likely going to be of increasing importance as we move toward integrating hardware acceleration with various libraries.

I think this repo should closely mirror the deployment repo (or even, the deployment repo should be folded into this one) to avoid last minute failures.

For example, some issues that are only caught at the last minute are building with different minimum versions of numpy or 32 bit builds. Since the build matrix for the wheel bulider is more complete than this one, this already causes headaches for the release manager.

matthew-brett:
Are you already doing regular wheel builds via cron job?  Then maybe the missing piece is some good way of making sure you are seeing when it fails?

hmaarrfk:
Maybe. But I already get too many emails. It's much easier to keep in touch when I can help people's contributions at the same time.

I personally think we can also get to a tag and forget model to delivering releases, but that model doesn't seem popular with many of the core members.

stefanv:
@hmaarrfk I'm not sure it matters that much which version of libc we test on? The manylinux standard compiles against an old version of libc, and these libraries have APIs that are supported for a long time forward.  It seems like breakage is extremely unlikely, and if it does happen we will at least see it before release.

The other scenarios you raise are much more relevant: versions of NumPy, Python, etc.

Perhaps we can go light on the "NOT ok"s and "MUST"s: this PR seemed to me like a reasonable proposal, and all caps on the interwebs seldom gets a warm reception :)

hmaarrfk:
@rfezzani, @stefanv and others. I apologize for the use of caps and the overly assertive tone. It was uncalled for.

The version of libc does make a big difference and ensuring we test on a consistent platform helps the release process go smoothly. A hybrid approach might work, but as linked in the pytorch conversation, there are some in the scientific community that believe these old versions are useful.

We've seen challenges before in different versions on the osx sdk being used, and creating a split between the main repo and wheel repo for Linux may only exacerbate challenges in creating releases.

rfezzani:
Don't worry @hmaarrfk, I didn't took it personally. I simply closed this PR because it received a no-go.
I was thinking about an other option for solving the random Travis CI fails based on this [post](https://travis-ci.community/t/then-sudo-apt-get-update-failed-public-key-is-not-available-no-pubkey-6b05f25d762e3157-in-ubuntu-xenial/1728/13).

stefanv:
@hmaarrfk Getting back to this issue, I think matching versions of XCode make a lot of sense, whereas building using ancient libc libraries less so. I don't think I can remember any cases where these caused us problems.  What I'd like to see is that we avoid as much last minute breakage as we can, while letting ourselves use more modern platforms to reduce pain on Travis-CI etc.

Perhaps we can come to an agreement of which aspects of the build systems should match, and which we can be a bit more relaxed about.

jni:
@stefanv @hmaarrfk 

We need another role in the scikit-image team, in addition to Guardian of the Dimensions (me) and Custodian of the Future (still unfilled): Building Inspector. =P As mentioned elsewhere, if someone can figure out email notifications for failing builds on scikit-image-wheels master, I'd subscribe to that feed.

stefanv:
@jni 🧞https://github.com/scikit-image/scikit-image-wheels/pull/27

hmaarrfk:
Would there be an interest in folding the wheels repo into this one?

stefanv:
That could be interesting; thoughts @matthew-brett ?

PR #4223, opened 2019-10-09
DOC: TV-L1 add optional indications + DOI

sciunto:
## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


jni:
fwiw, I'm more and more leaning towards the opinion (I think shared by @JDWarner?) that we shouldn't have optional, as this can in 99.9% of cases be introspected by looking at the function signature. In the rare cases where it cannot, sometimes you need or or another argument but not both (e.g. "footprint" vs "size" vs "structure" in ndimage neighbourhood definitions), "optional" is too coarse to express this, and it should instead be explained in the argument descriptions.

Nonetheless, that's something that can be addressed package-wide sometime in the future — for now this makes the function consistent with the rest of the package. =)

PR #4230, opened 2019-10-11
MAINT: cycle_spin cleanup

grlee77:
## Description

This PR is just a minor refactoring of cycle_spin to replace `_roll_axes` with numpy's own `roll` (this is possible now that our minimum support NumPy is >= 1.12).

This is purely a maintenance PR and does not change the existing behaviour.

## Checklist

~- [ ] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)~
~- [ ] Gallery example in `./doc/examples` (new features only)~
~- [ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark~
~- [ ] Unit tests~

## For reviewers

  later.

### DIFF
--- a/skimage/restoration/_cycle_spin.py
+++ b/skimage/restoration/_cycle_spin.py
@@ -46,32 +46,6 @@ def _generate_shifts(ndim, multichannel, max_shifts, shift_steps=1):
                      s, t in zip(max_shifts, shift_steps)])
 
 
-def _roll_axes(x, rolls, axes=None):
-    """Apply np.roll along a set of axes.
-
-    Parameters
-    ----------
-    x : array-like
-        Array to roll.
-    rolls : int or sequence
-        The amount to roll along each axis in ``axes``.
-    axes : int or sequence, optional
-        The axes to roll. Default is the first ``len(rolls)`` axes of ``x``.
-
-    Returns
-    -------
-    x : array
-        Data with axes rolled.
-    """
-    if axes is None:
-        axes = np.arange(len(rolls))
-    # Replace this loop with x = np.roll(x, rolls, axes) when NumPy>=1.12
-    # becomes a requirement.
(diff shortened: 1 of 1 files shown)


skimage/restoration/_cycle_spin.py
```diff
@@ -34,44 +34,18 @@ def _generate_shifts(ndim, multichannel, max_shifts, shift_steps=1):
     elif len(shift_steps) != ndim:
         raise ValueError("max_shifts should have length ndim")
 
-    if np.any(np.asarray(shift_steps) < 1):
+    if any(s < 1 for s in shift_steps):
```
sciunto: Is it faster?

grlee77: It is faster, but we are talking differences in microseconds so it is unlikely to make a practical performance difference in applications. The calls to `func` within `cycle_spinning` are likely to take orders of magnitude longer than the call to this `_generate_shifts` function.

example:
```
shift_steps = (1, 1, 0)

%timeit np.any(np.asarray(shift_steps) < 1)
18.9 µs ± 4.22 µs per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit any(s < 1 for s in shift_steps)
581 ns ± 49.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

grlee77: actually, I will qualify that it is faster in the specific case that shift_steps is small (which will always be the case here because the length of `shift_steps` is equal to the number of data dimensions).

If there were millions of elements, numpy would be faster in general (but still slower in a case where there is a value < 1 near the start of the array)

sciunto:
thanks @grlee77 for the explanations and the cleanup!

PR #4231, opened 2019-10-11
Register rotation and scaling using warp_polar gallery example

mellertd:
## Description
This extends the `warp_polar` registration gallery example started in #4097 to the case in which two images differ in scale and rotation but do not share a center. Per the discussion in #4086 

As noted in the original issue, there is a lot going on. The example image I used is particularly challenging, and I noticed getting this to work reliably took a little more tweaking than when I have used this strategy on other images. The example feels a bit hard to follow and I could probably improve readability, so I welcome any feedback.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


doc/examples/transform/plot_register_rotation.py
```diff
@@ -88,11 +88,120 @@
...
+def window_image(image, window_diameter=0.8, window_decay=10):
+    """window_diameter is relative to length of shortest axis
+    window_decay determines steepness of window edges (higher is steeper)"""
+    window_center = np.divide(image.shape, 2)
+    radius = (window_diameter * np.min(image.shape)) / 2
+    window = np.zeros_like(image)
+    rr, cc = draw.circle(window_center[0], window_center[1], radius)
+    window[rr, cc] = 1
+    window = filters.gaussian(window, sigma=(radius / window_decay))
+    return image * window
```
jni: @mellertd thanks for this!

I don't have time for a full review right now but my first impression is that the example creates too many helper functions, that in fact would be useful functions to have. Specifically:

- perhaps we can add `feature.difference_of_gaussians` to the feature module (and use it within `blob_dog`), and
- windowing an image also seems like a common signal processing scenario. Are you sure that this doesn't already exist in `scipy.ndimage` or `scipy.signal`? If it doesn't, then I think we could add it to `skimage.filters`.

mellertd: > the example creates too many helper functions

That was my feeling too, but I was having trouble getting around that and maintaining (in my view) readability.

> perhaps we can add feature.difference_of_gaussians to the feature module (and use it within blob_dog)

This seems pretty easy to implement, but is it too easy? If you're just running a filter at two different sigmas and taking the difference, is that worth its own function? I'd be happy to put it together in any case.

> windowing an image also seems like a common signal processing scenario. Are you sure that this doesn't already exist in scipy.ndimage or scipy.signal? If it doesn't, then I think we could add it to skimage.filters

I am glad you think so, because--unless I am totally missing something--there is no n-dimensional windowing available in `scipy.signal`. You can take the outer product of two 1-dimensional windows for a 2D window, but that isn't exactly what you want because generally you won't get a window with circular symmetry. I'd have to do a little reading to get at the math that it would take to turn any arbitrary 1D window into a 2/3D window with circular/spherical symmetry, but I am sure I could figure it out if you want me to take a crack at it.

mellertd: > turn any arbitrary 1D window into a 2/3D window with circular/spherical symmetry

Actually, I just thought of a way this could be done with `transform.warp`. I shall sleep on this

mellertd: #4252

emmanuelle:
@mellertd thanks for your patience :-). Will you have time to work on this one now that #4252 has been merged?

mellertd:
@emmanuelle yes, I'll use the new window function and tidy up some other things. Any thoughts as to whether it is worthwhile to create a `difference_of_gaussians` function? My feeling is probably not, as it is usually going to be a very simple one-liner to apply, but I am happy to do it.

jni:
I think it's worth doing, especially if you support making a chain of them as done in `blob_dog`. Even though it's a one-liner, it's one that I have to think about.

mellertd:
Going to try the Difference of Gaussians function #4445

PR #4235, opened 2019-10-13
Remove copy of tifffile; install from pip. Fixes #3578

kalvdans:
Fixes https://github.com/scikit-image/scikit-image/issues/3578

- [x] Add failing test for LZW compressed TIFF files.
- [x] Add `imagecodecs-lite` as pip requirement to fix above test failure.

## For reviewers

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

PR #4236, opened 2019-10-13
Migrate to networkx 2.x

jarrodmillman:
## Description

This should complete the migration from networkx 1.x to 2.x:
https://networkx.github.io/documentation/latest/release/migration_guide_from_1.x_to_2.0.html

In 2.4, we deleted several deprecated features:
https://networkx.github.io/documentation/latest/release/release_2.4.html

According to your `requirements/default.txt` file, you currently require `networkx>=2.0`, which should be fine as the new interface was introduced in 2.0.

See #4235.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/segmentation/plot_rag.py
+++ b/doc/examples/segmentation/plot_rag.py
@@ -67,7 +67,7 @@ g.add_edge(1, 3, weight=50)
 
 # Assigning dummy labels.
 for n in g.nodes():
-    g.node[n]['labels'] = [n]
+    g.nodes[n]['labels'] = [n]
 
 gc = g.copy()
 
--- a/doc/examples/segmentation/plot_rag_merge.py
+++ b/doc/examples/segmentation/plot_rag_merge.py
@@ -36,7 +36,7 @@ def _weight_mean_color(graph, src, dst, n):
         difference of the mean color between node `dst` and `n`.
     """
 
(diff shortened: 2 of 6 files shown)


hmaarrfk:
Do we want to backport this to 0.16.1?

jarrodmillman:
I would.  You've required nx>2.0 since at least 0.15.  So these changes shouldn't create any problems.  I plan to release networkx 2.4 in about a week, so without this patch 0.16.x will fail the current version of networkx at that point.

jni:
0.16.2, 0.16.1 is already tagged (there was a snafu with the 0.16.0 tag). But yes!

hmaarrfk:
@meeseeksdev backport to v0.16.x

hmaarrfk:
I guess we shouldn't backport to 0.14.x since the primary use is 2.7 now that we will have 0.16.1 out soon.

(There are a few bugs fixed in 0.14 that have not been fixed in 0.15)

jni:
@hmaarrfk I agree, not much point adding to 0.14, but we *should* probably backport to 0.15, as I expect there will still be quite a few of those out there "in the wild".

jni:
@meesseeksdev backport to v0.15.x

Issue #4248, opened 2019-10-17
Flood_fill use "inplace" kwarg, and the library uses "in_place" elsewhere


sciunto:
## Description

https://scikit-image.org/docs/dev/api/skimage.segmentation.html#skimage.segmentation.flood_fill

To be consistent, inplace must be deprecated in favor of in_place.

robroooh:
is anyone working on this? I would like to contribute.

rfezzani:
Fixed by #4250, thx @robroooh

sciunto:
Great! Thank you @robroooh

PR #4250, opened 2019-10-17
In flood_fill deprecate inplace and use in_place instead

robroooh:
## Description

enhancement from https://github.com/scikit-image/scikit-image/issues/4248. adding deprecation warning for flood_fill param.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


hmaarrfk:
This would be included in 0.17, and removed in 0.19

sciunto:
You need to add in the function signature ", *, in_place=False", turns default for "inplace" to None, document the new option, update the former one and state that it's deprecated, handle both arguments at the top of the function to use in_place everywhere else in the function.

You also need to check for other function calls in the library, especially in the documentation :)

robroooh:
updated the pr now :smiley: Please give me feedback on what could be improved :+1:

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -93,7 +93,7 @@ def test_inplace_noncontiguous():
     # Transpose is noncontiguous
     image2 = image[::2, ::2]
 
-    flood_fill(image2, (0, 0), 5, inplace=True)
+    flood_fill(image2, (0, 0), 5, in_place=True)
```
hmaarrfk: Can you duplicate one of these tests to ensure that the old parameter still works?

That test can then be easily removed when it comes time to finally rice the behavior. There is also a way to enforce q check on the working being issued.

robroooh: That's a very nice input! Thanks :+1:

skimage/morphology/_flood_fill.py
```diff
@@ -100,10 +105,15 @@ def flood_fill(image, seed_point, new_value, *, selem=None, connectivity=None,
            [5, 5, 5, 5, 2, 2, 5],
            [5, 5, 5, 5, 5, 5, 3]])
     """
+    if inplace is not None:
+        warn('The inplace parameter will change'
```
hmaarrfk:
```suggestion
        warn('The inplace parameter will change '
```

skimage/morphology/_flood_fill.py
```diff
@@ -46,6 +46,11 @@ def flood_fill(image, seed_point, new_value, *, selem=None, connectivity=None,
         If a tolerance is provided, adjacent points with values within plus or
         minus tolerance from the seed point are filled (inclusive).
     inplace : bool, optional
```
hmaarrfk: Nit: Can we change the order of these two parameters. Have the suggestion one appear first?

skimage/morphology/_flood_fill.py
```diff
@@ -14,7 +14,7 @@
 
 
 def flood_fill(image, seed_point, new_value, *, selem=None, connectivity=None,
-               tolerance=None, inplace=False):
+               tolerance=None, inplace=None, in_place=False):
```
hmaarrfk:
```suggestion
               tolerance=None, in_place=False, inplace=None):
```

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -112,6 +112,48 @@ def test_inplace_noncontiguous():
...
+    """This test is deprecated and will be removed in
+    version 0.19.0. See #4248.
+    """
+    image = np.array([[0, 0, 0, 0, 0, 0, 0],
+                      [0, 1, 1, 0, 2, 2, 0],
+                      [0, 1, 1, 0, 2, 2, 0],
+                      [1, 0, 0, 0, 0, 0, 3],
+                      [0, 1, 1, 1, 3, 3, 4]])
+
+    flood_fill(image, (0, 0), 5, inplace=True)
```
hmaarrfk:
```suggestion
    with expected_warnings(['The inplace']):
        flood_fill(image, (0, 0), 5, inplace=True)
```
Needs a 
```python
from skimage._shared.testing import expected_warnings
```
at the top

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -112,6 +112,48 @@ def test_inplace_noncontiguous():
...
+    """This test is deprecated and will be removed in
+    version 0.19.0. See #4248.
+    """
+    image = np.array([[0, 0, 0, 0, 0, 0, 0],
+                      [0, 1, 1, 0, 2, 2, 0],
+                      [0, 1, 1, 0, 2, 2, 0],
+                      [1, 0, 0, 0, 0, 0, 3],
+                      [0, 1, 1, 1, 3, 3, 4]], dtype=np.float32)
+
+    flood_fill(image, (0, 0), 5, inplace=True)
```
hmaarrfk:
```suggestion
    with expected_warnings(['The inplace']):
        flood_fill(image, (0, 0), 5, inplace=True)
```

hmaarrfk:
I requested a few small changes that I think would go a long way to helping maintain the library.

They would mostly hide the warnings in the test suite.

https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=1733&view=logs&jobId=0c528f75-1409-556f-1999-9f155ac37a3a&taskId=1c50252d-14f6-55f6-616c-dfe500c925e4&lineStart=5675&lineEnd=5676&colStart=1&colEnd=1

Those warnings do help us identify deprecations upstream. Keeping them quiet is useful !

skimage/morphology/_flood_fill.py
lagru: Just a suggestion which may be more clearer:
```
The `inplace` parameter is depreciated and will be removed 
in version 0.19.0. Use `in_place` instead.
```
But I'm fine either.

skimage/morphology/_flood_fill.py
```diff
@@ -100,10 +105,15 @@ def flood_fill(image, seed_point, new_value, *, selem=None, connectivity=None,
            [5, 5, 5, 5, 2, 2, 5],
            [5, 5, 5, 5, 5, 5, 3]])
     """
+    if inplace is not None:
+        warn('The inplace parameter will change'
+             'to in_place in version 0.19.0.', stacklevel=2)
```
lagru: Could you add `category=FutureWarning` here?

rfezzani:
Isn't this PR related to #4154 and can't we use the same strategy as #4158 for consistency?

hmaarrfk:
You could. But I think it is Overkill and in many cases not as flexible and much more obscure. It also doesn't provide the best stack trace.

Finally, it is really easy for contributor's to understand what is going on without decorators.

I'm fine with both approaches really though.

rfezzani:
Using the decorator have some advantages over the "handmade" deprecation:
- the warning message is standardized
- less lines are introduced to the code base (if multiple such deprecation are introduced like it will be the case according to #4154 )
- the maintenance is easier (only one line to be removed at deprecation cycle end)
- less noise introduced to the documentation, since the deprecated argument is replaced by the new one instead of adding a new argument to the argument list, and only the users concerned by the deprecation are impacted.

hmaarrfk:
I am in full agreement with all the comments above. 

For me, 

> Finally, it is really easy for contributor's to understand what is going on without decorators.

might trump everything ;)

At this stage, there are a few minor fixes, and then this PR can be merged in. Cleanups seem minor, especially seeing that this code won't exist in 1 year.

rfezzani:
I also forgot to mention #4152 as issue to be fixed with the decorator.

skimage/morphology/_flood_fill.py
```diff
@@ -45,10 +45,15 @@ def flood_fill(image, seed_point, new_value, *, selem=None, connectivity=None,
         value of `image` at `seed_point` to be filled.  This is fastest.
         If a tolerance is provided, adjacent points with values within plus or
         minus tolerance from the seed point are filled (inclusive).
-    inplace : bool, optional
+    in_place : bool, optional
         If True, flood filling is applied to `image` inplace.  If False, the
```
stefanv: Our mistake from before, but note that there is no such English word as "inplace": it should be "in-place".

robroooh: thank you for the review :+1: this is fixed now.

jni:
Great stuff! Thank you @robroooh and everyone else who reviewed this!

Issue #4245, opened 2019-10-16
Output of graph.normalized_cut is not deterministic


hacktuarial:
## Description
After reading the [original paper](https://repository.upenn.edu/cgi/viewcontent.cgi?article=1101&context=cis_papers), I wasn't sure if the output was deterministic or not. Based on the example below, the output is not deterministic, even after fixing the seed. Can you provide any insight into why this is so?

## Way to reproduce

Based on https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_ncut.html

```python
from skimage import data, segmentation, color
from skimage.future import graph
from numpy.testing import assert_array_equal
import random

img = data.coffee()
labels1 = segmentation.slic(img, compactness=30, n_segments=400)
out1 = color.label2rgb(labels1, img, kind='avg')
g = graph.rag_mean_color(img, labels1, mode='similarity')

# call cut_normalized with the same inputs and seed
labels = [None] * 10
for i in range(len(labels)):
    random.seed(1234)
    labels[i] = graph.cut_normalized(labels1, g)

# outputs should be all the same
for i in range(len(labels) - 1):
    assert_array_equal(labels[i], labels[i+1])
```

## Version information
```
3.6.7 (default, May 13 2019, 16:14:45) 
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)]
Darwin-18.7.0-x86_64-i386-64bit
scikit-image version: 0.16.1
numpy version: 1.17.2
```

```python
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-3-86707092a69f> in <module>
     22 
     23 for i in range(len(labels) - 1):
---> 24     assert_array_equal(labels[i], labels[i+1])

~/.pyenv/versions/3.6.7/envs/jupyter/lib/python3.6/site-packages/numpy/testing/_private/utils.py in assert_array_equal(x, y, err_msg, verbose)
    916     __tracebackhide__ = True  # Hide traceback for py.test
    917     assert_array_compare(operator.__eq__, x, y, err_msg=err_msg,
--> 918                          verbose=verbose, header='Arrays are not equal')
    919 
    920 

~/.pyenv/versions/3.6.7/envs/jupyter/lib/python3.6/site-packages/numpy/testing/_private/utils.py in assert_array_compare(comparison, x, y, err_msg, verbose, header, precision, equal_nan, equal_inf)
[...]

PR #4251, opened 2019-10-17
Initialize starting vector for `scipy.sparse.linalg.eigsh` to ensure reproducibility in graph_cut

hacktuarial:
## Description

Addresses issue #4245, and adds unit tests to cover `inplace` argument to `cut_normalized`

## Checklist

- [ ] Gallery example in `./doc/examples` (NOT APPLICABLE)
  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/future/graph/graph_cut.py
+++ b/skimage/future/graph/graph_cut.py
@@ -268,7 +268,9 @@ def _ncut_relabel(rag, thresh, num_cuts):
         d2.data = np.reciprocal(np.sqrt(d2.data, out=d2.data), out=d2.data)
 
         # Refer Shi & Malik 2001, Equation 7, Page 891
-        vals, vectors = linalg.eigsh(d2 * (d - w) * d2, which='SM',
+        A = d2 * (d - w) * d2
+        v0 = np.random.rand(min(A.shape))
+        vals, vectors = linalg.eigsh(A, which='SM', v0=v0,
                                      k=min(100, m - 2))
 
         # Pick second smallest eigenvector.
--- a/skimage/future/graph/graph_cut.py
+++ b/skimage/future/graph/graph_cut.py
@@ -76,6 +76,7 @@ def cut_threshold(labels, rag, thresh, in_place=True):
 
 def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
                    max_edge=1.0,
+                   *,
                    random_state=None,
                    ):
     """Perform Normalized Graph cut on the Region Adjacency Graph.
--- a/skimage/future/graph/graph_cut.py
+++ b/skimage/future/graph/graph_cut.py
@@ -109,7 +109,7 @@ def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
         If RandomState instance, random_state is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`. The random state is used for the starting point
-        of `scipy.sparse.linalg.eighsh`
+        of `scipy.sparse.linalg.eigsh`.
 
     Returns
     -------
--- a/skimage/future/graph/graph_cut.py
+++ b/skimage/future/graph/graph_cut.py
@@ -6,6 +6,7 @@ except ImportError:
 import numpy as np
 from . import _ncut
 from . import _ncut_cy
+from ..._shared.utils import check_random_state
 from scipy.sparse import linalg
 
 
@@ -74,7 +75,9 @@ def cut_threshold(labels, rag, thresh, in_place=True):
 
 
 def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
-                   max_edge=1.0):
+                   max_edge=1.0,
+                   random_state=None,
+                   ):
     """Perform Normalized Graph cut on the Region Adjacency Graph.
 
     Given an image's labels and its similarity RAG, recursively perform
@@ -100,6 +103,12 @@ def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
         The maximum possible value of an edge in the RAG. This corresponds to
         an edge between identical regions. This is used to put self
         edges in the RAG.
+    random_state : int, RandomState instance or None, optional
+        If int, random_state is the seed used by the random number generator;
+        If RandomState instance, random_state is the random number generator;
+        If None, the random number generator is the RandomState instance used
+        by `np.random`. The random state is used for the starting point
+        of `scipy.sparse.linalg.eighsh`
 
     Returns
     -------
@@ -122,13 +131,14 @@ def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
            IEEE Transactions on, vol. 22, no. 8, pp. 888-905, August 2000.
 
     """
+    random_state = check_random_state(random_state)
     if not in_place:
         rag = rag.copy()
 
     for node in rag.nodes():
         rag.add_edge(node, node, weight=max_edge)
 
-    _ncut_relabel(rag, thresh, num_cuts)
+    _ncut_relabel(rag, thresh, num_cuts, random_state)
 
     map_array = np.zeros(labels.max() + 1, dtype=labels.dtype)
     # Mapping from old labels to new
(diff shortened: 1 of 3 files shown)


skimage/future/graph/rag.py
```diff
@@ -157,6 +157,21 @@ def __init__(self, label_image=None, connectivity=1, data=None, **attr):
                                   strides=((0,) * label_image.ndim)),
                 extra_arguments=(self,))
 
+    def __eq__(self, other):
```
hacktuarial: Please make changes to this - I am not very familiar with this class, but added all of the things I could think of while investigating this issue.

skimage/future/graph/tests/test_rag.py
```diff
@@ -184,6 +187,36 @@ def test_ncut_stable_subgraph():
...
+def test_random_seed():
+    """ensure cut_normalized returns the same output for the same input,
+    when setting numpy random seed
+    """
+    img = data.coffee()
+    labels1 = segmentation.slic(img, compactness=30, n_segments=400)
+    g = graph.rag_mean_color(img, labels1, mode='similarity')
+    results = [None] * 10
+    for i in range(len(results)):
+        np.random.seed(1234)
```
hacktuarial: this test failed when I used `random.seed` instead

jni:
Thanks very much @hacktuarial! I can't do a full review yet but on first glance this looks great!

skimage/future/graph/graph_cut.py
```diff
@@ -268,7 +268,9 @@ def _ncut_relabel(rag, thresh, num_cuts):
         d2.data = np.reciprocal(np.sqrt(d2.data, out=d2.data), out=d2.data)
 
         # Refer Shi & Malik 2001, Equation 7, Page 891
-        vals, vectors = linalg.eigsh(d2 * (d - w) * d2, which='SM',
+        A = d2 * (d - w) * d2
```
sciunto:
```suggestion
        A = d2 * (d - w) * d2
        # Initialize the vector to ensure reproducibility.
```

skimage/future/graph/rag.py
```diff
@@ -157,6 +157,21 @@ def __init__(self, label_image=None, connectivity=1, data=None, **attr):
...
+            return False
+        for node in self.nodes:
+            for key, this_val in self.nodes[node].items():
+                other_val = other.nodes[node][key]
+                if isinstance(this_val, np.ndarray):
+                    equal = (this_val == other_val).all()
+                else:
+                    equal = (this_val == other_val)
+                if not equal:
+                    print(f"Unequal! {node}/'{key}':{this_val} != {other_val}")
```
sciunto: I guess this line must be removed.

sciunto:
Besides, the docstring of `_ncut_relabel` is incorrect. Several parameters not set in the signature of the function are described. Would you mind fixing this as well @hacktuarial ?

skimage/future/graph/tests/test_rag.py
hacktuarial: apparently this is to be expected: https://stackoverflow.com/questions/31057197/should-i-use-random-seed-or-numpy-random-seed-to-control-random-number-gener

skimage/future/graph/rag.py
hacktuarial: good catch! leftover from my debugging

skimage/future/graph/graph_cut.py
```diff
@@ -268,7 +263,10 @@ def _ncut_relabel(rag, thresh, num_cuts):
...
+        # Initialize the vector to ensure reproducibility.
+        v0 = np.random.rand(min(A.shape))
```
hmaarrfk: How does this ensure reproducibility? We are initializing it with a random number?

Maybe provide the user to override this?

hmaarrfk: Can you specify exactly which dimension should be used? Is min really the only way to know which one should be used?

hacktuarial: `linalg.eigsh` calls out to FORTRAN to compute the eigenvalues of `A`. I haven't read the FORTRAN code, but I'm guessing what happens is some kind of random initialization like this. But since it's done in FORTRAN, the Python end-user has no control over it. Using `numpy` to set the initial guess for the eigenvector lets the user ensure reproducibility by setting `np.random.seed`, as in the unit test.

hacktuarial: And you're right, `A` has to be square anyway, so I'll just take the first element instead of `min`

hacktuarial: Also, I don't think it makes sense to pass `v0` from `cut_normalized` to `_ncut_relabel`, since `_ncut_relabel` recursively calls itself. If we did that, then every recursive call would start with the same initial vector, which I imagine could behave strangely in some edge cases. As currently proposed, we use different random initializations, but in a predictable way through `np.random.seed`

hmaarrfk: Maybe then you can pass a [random number] generator? 
https://docs.scipy.org/doc/numpy/reference/random/index.html

If it is None, then it can go and fetch the default one from numpy as you've done.

hacktuarial: OK, see a3fdb30. I followed the usage in `feature.haar.draw_haar_like_feature`.

jni:
Wow, I'd totally forgotten about `check_random_state`. Great find @hacktuarial and thanks @hmaarrfk and @sciunto for the reviews. Again I'm writing this in a rush, but for now everything looks good and the only thing I'm not sure about is the `__eq__` operator. I think that needs to be thought through a bit more thoroughly (it is very tempting at this point to go back and change "needs to be" to "ought to be" =P). For one, we are not checking for edge equality, which we should. But there may be other issues — I need to think about it some more.

hacktuarial:
@jni what do you think about splitting this into 2 separate PRs? one for the random seed stuff, and another for testing the `in_place` argument?

jni:
@hacktuarial I'm game for that. =)

skimage/future/graph/graph_cut.py
```diff
@@ -74,7 +75,9 @@ def cut_threshold(labels, rag, thresh, in_place=True):
 
 
 def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
-                   max_edge=1.0):
+                   max_edge=1.0,
+                   random_state=None,
```
hmaarrfk:
```suggestion
                   *,
                   random_state=None,
```

skimage/future/graph/rag.py
hmaarrfk: Maybe just move this function to the test methods. It seems that would avoid the introduction of a new public method. You can add a comment that you think it might be useful to move it back to the class in the future

hacktuarial: My preference is for separate PRs, this method needs to be modified to consider edges, as @jni pointed out.

skimage/future/graph/graph_cut.py
```diff
@@ -100,6 +104,12 @@ def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
         The maximum possible value of an edge in the RAG. This corresponds to
         an edge between identical regions. This is used to put self
         edges in the RAG.
+    random_state : int, RandomState instance or None, optional
+        If int, random_state is the seed used by the random number generator;
+        If RandomState instance, random_state is the random number generator;
+        If None, the random number generator is the RandomState instance used
+        by `np.random`. The random state is used for the starting point
+        of `scipy.sparse.linalg.eighsh`
```
sciunto:
```suggestion
        of `scipy.sparse.linalg.eigsh`.
```

skimage/future/graph/graph_cut.py
```diff
@@ -245,18 +256,15 @@ def _ncut_relabel(rag, thresh, num_cuts):
...
     thresh : float
         The threshold. A subgraph won't be further subdivided if the
         value of the N-cut exceeds `thresh`.
     num_cuts : int
         The number or N-cuts to perform before determining the optimal one.
-    map_array : array
-        The array which maps old labels to new ones. This is modified inside
-        the function.
+    random_state: RandomState instance
+        Provides initial values for eigenvalue solver
```
sciunto:
```suggestion
        Provides initial values for eigenvalue solver.
```

PR #4252, opened 2019-10-18
Function to create nD windows

mellertd:
## Description

This function uses `transform.warp` to create 2D windows from those returned by `scipy.signal.get_window`. This is very incomplete, but I want some general feedback before I go on. My approach here feels a little hacky, but it works and is pretty fast. I am also unsure how to handle the docstring--could I point to the `scipy` docs or should I recreate their API docs here? If the latter, is it okay to straight copy, and how to best reference/give credit?

@jni , this would cover one of the needs you noted in your comments on #4231.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


jni:
@mellertd is there any reason why this couldn't be nD rather than 2D?

mellertd:
@jni the tldr is that no, there is no reason this couldn't be nD, but I am probably not the one to pull it off.

Longer explanation: this approach using `transform.warp` isn't suitable for n-dimensions because of how `transform.warp` works. Namely, I can't simply supply a mapping function, I would have to specify the mapping coordinates directly. This is okay for 3D and is the approach I would/will take. This is still doable for nD, and I came up with a solution, but it's a bit more complicated (involving building coordinate sets programmatically with `np.meshgrid`).

The larger problem is one of memory management. The overall idea here is to produce  hyperspherically symmetric windows. A 2D window of your typical 2D image size (e.g., 512x512, 1024x1024) is trivial. Most 3D images are going to have anisotropy anyway, but it isn't uncommon to see an MRI that is, e.g., 512 on every axis. But now memory starts to become a concern (and creating the window much less snappy). A 4D array that is 512 on every axis, now that is pretty big and I don't even know how I would approach that. Add another dimension, forget about it.

But backing up a bit, I am not sure whether it makes sense to apply this type of windowing to an anisotropic image. In the simplest case, one might use a window that is circularly symmetric in the 2D plane and use broadcasting to multiply each pixel with a new 1D window of the appropriate shape to extend into the 3rd dimension. You wouldn't have hyperspherical symmetry, but it would be good enough for most applications, probably.

On a more practical level, I can't imagine any common use case where windowing above 3 dimensions would be useful for image analysis. But I know such a claim would not compel the Protector of the Dimensions :)

mellertd:
@jni after all my moaning, I got this to work for nd.

I could use some advice for how to deal with the fact that you can get yourself into trouble very quickly by accidentally creating a too-large array, though. What is the best way to handle? Have a safety parameter (like, raise an exception if someone tries to create an array over a certain size and have a parameter to override the exception?)

By the way, here is some quick and dirty timing:

```
# the original function I created
In [29]: %timeit get_window2('hann', 512)                                                                                                                                                          
40.7 ms ± 308 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

# nd version for 2d, possible faster?!
In [31]: %timeit get_windownd('hann', 512, 2)                                                                                                                                                      
36.8 ms ± 2.4 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

mellertd:
And a 3D Hann window. Pretty slow, but it's a huge array. God help you if you try to add another dimension of that size.

```
In [32]: %timeit get_windownd('hann', 512, 3)                                                                                                                                                      
17.3 s ± 1.9 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

skimage/filters/__init__.py
```diff
@@ -59,4 +60,5 @@
            'threshold_multiotsu',
            'apply_hysteresis_threshold',
            'rank',
-           'unsharp_mask']
+           'unsharp_mask',
+           'get_windownd']
```
jni: Can we just call it `get_window`? Long term, it should be the assumption in scikit-image that functions are nd. ;)

skimage/filters/window.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    window_size = output_coords.shape[1]
+    center = (window_size / 2) - 0.5
+    coords = np.zeros_like(output_coords, dtype=np.double)
+    coords[0, ...] = np.sqrt(((output_coords - center) ** 2).sum(axis=0))
+    return coords
+
+
+def get_windownd(window, size, ndim=2):
+    """Docstring here
```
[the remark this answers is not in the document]
mellertd: Yes, this and the other docstring suggestions are in progress! (WIP and all that :) )

skimage/filters/__init__.py
mellertd: I was worried about possible confusion with `scipy.signal.get_window`, but if this is not a concern, I agree that would be preferred

mellertd:
@jni took care of your suggested changes. Any suggestions to continue to improve are appreciate. I will work on tests next.

mellertd:
Alrighty @jni, I have taken this pretty far and even have a gallery example now. Let me know if you still want more explanation in the docstring, whether I still need to clean up the code to make it more understandable, whether I should nuke this from orbit and take a different approach, etc.

emmanuelle:
Very nice addition, thank you @mellertd. I have one question about the API: should we use `get_window` or just `window`? When we use a prefix, it comes with the obligation of consistency and in the past we've often transitioned to simpler names (ie `median_filter` is now `median` etc.).

skimage/filters/window.py
```diff
@@ -0,0 +1,90 @@
...
+    nd_window : ndarray
+        A window of of `ndim` dimensions with a length of `size`
+        along each axis. `dtype` is `np.double`.
+
+    Notes
+    -----
+    This function is based on `scipy.signal.get_window` and thus can access
+    all of the window types available to that function
+    (e.g., `"hann"`, `"boxcar"`). Note that certain window types require
+    parameters that have to supplied with the window name as a tuple
```
emmanuelle: have to be supplied

skimage/filters/window.py
```diff
@@ -0,0 +1,90 @@
...
+from ..transform import warp
+from .._shared.utils import safe_as_int
+from scipy.signal import get_window as get_window1d
+
+
+def get_window(window, size, ndim=2, **kwargs):
+    """Return an n-dimensional window of a given size and dimensionality.
+
+    Parameters
+    ----------
```
emmanuelle: Here I would copy as many information from the scipy docstring as needed (that was one of your questions). In particular put all the names of the available windows. Let's acknowledge that we're all lazy :-).

skimage/filters/tests/test_get_window.py
```diff
@@ -0,0 +1,34 @@
...
+from skimage.filters import get_window
+from skimage._shared.testing import parametrize, raises
+
+
+@parametrize("window",
+             [16,
+              'hann',
+              ('tukey', 0.8)])
+@parametrize("size", [5, 8, 15])
+@parametrize("ndim", [2, 3, 4])
```
emmanuelle: how much time does it take to execute the test with `ndim=4` and `size=15`? The test suite is quite long, so we want to make sure a new test executes in a time way below the s (ideally below 0.1 s I would say).

doc/examples/filters/plot_get_window.py
```diff
@@ -0,0 +1,50 @@
...
+discontinuities distort the output of the FFT, resulting in energy from "real"
+frequency components leaking into wider frequencies.
+
+The effects of spectral leakage can be reduced by multiplying the signal with
+a window function. Windowing smoothly reduces the amplitude of the signal as it
+reaches the edges, removing the effect of the artificial discontinuity that
+results from the FFT.
+
+In this example, we see that the FFT of a typical image can show strong
+spectral leakage along the x and y axes. The application of a two-dimensional
```
emmanuelle: along the x and y axes (see the vertical and horizontal lines in the figure).

jni:
Sorry for my prolonged absence! My inbox has been overflowing and this slipped through the cracks (despite @perlman's prods, @mellertd! ;). I'll aim to review properly before the Pacific meeting, but you might want to address @emmanuelle's comments in the meantime. =)

skimage/filters/window.py
mellertd: I will include the available windows; I think that is the only information that is really missing at the moment. One concern: if a window function is added/removed in `scipy.signal`, it won't necessarily show up here. Unless I add a test to look for the availability of all docstring-listed windows (should I do that?).

skimage/filters/tests/test_get_window.py
mellertd: The tests I had in there were way overkill. Pruned way back

skimage/filters/window.py
```diff
@@ -0,0 +1,112 @@
...
+        raise ValueError("Number of dimensions must be greater than zero")
+
+    w = get_window(window, size, fftbins=False)
+    wcenter = (size / 2) - 0.5
+    w = np.reshape(w, (-1,) + (1,) * (ndim-1))
+
+    # Create coords for warping following `ndimage.map_coordinates` convention.
+    L = [np.arange(size, dtype=np.double) for i in range(ndim)]
+    coords = np.stack((np.meshgrid(*L)))
+    center = (size / 2) - 0.5
```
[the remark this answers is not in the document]
mellertd: good catch on the centers thing! Conceptually the input and output window centers are different, but they are mathematically the same obviously 🤦‍♂ 

I am sure about the 0.5 in that it is the only way to pass the tests and produce the right values! I am not sure whether this is philosophically the right approach, although whatever the correct approach is, it needs to account for windows with even/odd number sizes. Thoughts?

skimage/filters/window.py
```diff
@@ -0,0 +1,112 @@
...
+    >>> from skimage.filters import get_window
+    >>> w = get_window('hann', 512)
+
+    Return a Kaiser window with beta parameter of 16 and shape (256, 256, 256):
+
+    >>> w = get_window(16, 256, ndim=3)
+
+    Return a Tukey window with an alpha parameter of 0.8 and shape (100, 100):
+
+    >>> w = get_window(('tukey', 0.8), 100)
```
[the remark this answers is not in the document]
mellertd: fixed, thanks for the catch!

skimage/filters/window.py
```diff
@@ -0,0 +1,112 @@
...
+    wcenter = (size / 2) - 0.5
+    w = np.reshape(w, (-1,) + (1,) * (ndim-1))
+
+    # Create coords for warping following `ndimage.map_coordinates` convention.
+    L = [np.arange(size, dtype=np.double) for i in range(ndim)]
+    coords = np.stack((np.meshgrid(*L)))
+    center = (size / 2) - 0.5
+    coords[0, ...] = np.sqrt(((coords - center) ** 2).sum(axis=0)) + wcenter
+    coords[1:, ...] = 0
+    return warp(w, coords, **kwargs)
```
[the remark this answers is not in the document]
mellertd: added clarification in docstring

skimage/filters/_window.py
```diff
@@ -0,0 +1,122 @@
...
+        `scipy.signal.get_window` is allowed here. See notes below for a
+        current list, or the SciPy documentation for the version of SciPy
+        on your machine.
+    size : int
+        The size of the window along each axis (all axes will be equal length).
+    ndim : int, optional (default: 2)
+        The number of dimensions of the window.
+    warp_kwargs : dict
+        Keyword arguments passed to `transform.warp` (e.g.,
+        `warp_kwargs={'order':3}` to change interpolation method).
```
jni: These need to be double backticks:

```
``warp_kwargs={'order': 3}``
```

One of those annoying differences between ReStructured Text and markdown.

Same for `transform.warp`, or you can use single quotes with a fully-qualified name, ie `skimage.transform.warp`. (Single backticks will be converted to links to the appropriate documentation.)

skimage/filters/_window.py
```diff
@@ -0,0 +1,122 @@
...
+    parameters that have to be supplied with the window name as a tuple
+    (e.g., `("tukey", 0.8)`). If only a float is supplied, it is interpreted
+    as the beta parameter of the Kaiser window.
+
+    See https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.windows.get_window.html
+    for more details.
+
+    Note that this function can generate very large arrays that can consume
+    a large amount of available memory. For example, if `size=512` and
+    `ndim=4` are given as parameters to `filters.get_window`, it will attempt
```
jni: double backticks + `filters.window`

skimage/filters/_window.py
```diff
@@ -0,0 +1,122 @@
...
+    Euclidean distance from the center of the intended nD window to each
+    position in the array. Then it uses that distance to sample, with
+    interpolation, from a 1D window returned from `scipy.signal.get_window`.
+    The method of interpolation can be changed with the `order` keyword
+    argument passed to `transform.warp`.
+
+    Some coordinates in the output window will be outside of the original
+    signal. These will be filled in according to the default behavior of
+    `transform.warp` (i.e. zeros). This can be changed with the `mode`
+    keyword argument passed to `transform.warp`.
```
jni: imho we should set that mode explicitly. If zeros are needed (I think they are?), then relying on default behaviour is risky.

mellertd:
Is there a preference for merging vs. rebasing?

jni:
These days we squash commits upon merging, so it doesn't matter. Only in cases where we explicitly want to preserve the commit history do we prefer rebase.

jni:
I’ve taken the liberty of updating the title of this PR to better reflect its awesomeness. =D

grlee77:
This implementation looks good, but I would like it if the separate `size` and `ndim` arguments with a single `shape` argument. This can allow for the creation of windows with anisotropic size (such images/volumes are pretty common in practice). 

I played around with it a little and came up with an implementation that requires only fairly minor modification to the coordinate generation code. I am happy to push a commit here adding this feature if others agree.

Specifically here is an example using the proposed modification:
```Python
from skimage.filters import window
plt.figure()
plt.imshow(window('hann', shape=(256, 128)))
```
window_anisotropic

The key change needed is just to scale L based on the relative size of each axis:
```Python
    max_size = functools.reduce(max, shape)
    w = get_window(window_type, max_size, fftbins=False)
    w = np.reshape(w, (-1,) + (1,) * (ndim-1))

    # Create coords for warping following `ndimage.map_coordinates` convention.
    L = [np.arange(s, dtype=np.double) * (max_size / s) for s in shape]
```

grlee77:
Another potential future enhancement would be to add a boolean `separable` argument that if True would just apply 1D windows along each axis in a separable fashion. That would be computationally simpler than coordinate warping, but would obviously no longer provide the same sort of radially symmetric windows.

lagru:
@grlee77 I really like your suggestion and be in favor of a `shape` parameter. It would fit so well with the behavior we are used to from NumPy's array creation routines!

mellertd:
@grlee77 The API change makes sense and has an appeal, but I am not sure that such windows make much sense with respect to the math (but I am by no means an expert in this domain).

Specifically, if you are windowing only to minimize spectral leakage due to image edges and don't care about anisotropy, the separable form (i.e. applying 1D windows) is probably sufficient. If you need the radial form, distorting the function to match anisotropy of your image is going to kill the advantage of the radial form to begin with (consistent frequency response independent of orientation). I could be wrong here, but in that case I would certainly appreciate an explanation from someone with more signal processing expertise than me! :)

I can tell you that 2D windows that don't have circular symmetry would not work well for the use case that inspired this function in the first place: solving scaling and rotation differences between two 2D images.

grlee77:
> I can tell you that 2D windows that don't have circular symmetry would not work well for the use case that inspired this function in the first place: solving scaling and rotation differences between two 2D images.

Is the window applied in the image domain in this application? What do you do then for anisotropic images. Do you keep a "square" window and zero-pad? 

I was thinking more along the lines of using this as an apodization window to reduce [Gibbs ringing artifacts](http://mriquestions.com/gibbs-artifact.html) which I think is what you meant by "spectral leakage"? At least for MRI, it is common to collect Fourier domain data with anisotropic shape, but I think a separable filter approach is typically sufficient.

I don't want to hold up the PR too much, but if there is going to be an API change we should make it now before there is a release so we don't have to go through deprecation cycles later.

skimage/filters/_window.py
```diff
@@ -0,0 +1,120 @@
...
+
+    w = get_window(window_type, size, fftbins=False)
+    w = np.reshape(w, (-1,) + (1,) * (ndim-1))
+
+    # Create coords for warping following `ndimage.map_coordinates` convention.
+    L = [np.arange(size, dtype=np.double) for i in range(ndim)]
+    coords = np.stack((np.meshgrid(*L)))
+    center = (size / 2) - 0.5
+    coords[0, ...] = np.sqrt(((coords - center) ** 2).sum(axis=0)) + center
+    coords[1:, ...] = 0
```
grlee77: The following is more efficient and has lower memory use at cost of being more verbose. The key component is using the `sparse` argument to meshgrid which gives much smaller broadcastable coordinate arrays.


```Python
    center = (size / 2) - 0.5
    dist = 0
    for g in np.meshgrid(*L, sparse=True):
        g -= center
        dist = dist + g * g
    dist = np.sqrt(dist)
    coords = np.zeros((ndim,) + dist.shape, dtype=np.double)
    coords[0] = dist + center
```

It may also be sufficient to just use `np.float32` for the `coords` array.

mellertd:
> Is the window applied in the image domain in this application? What do you do then for anisotropic images. Do you keep a "square" window and zero-pad?

For true anisotropy (e.g., confocal stacks), I am not processing the whole stack at once. It is a registration task. I am solving first for xy differences, then subsequently dealing with any translations or rotations in xz/yz. Scaling differences are largely recovered when doing the xy bit.

For the 2d registrations, the window is applied in the image domain prior to the FFT. If the aspect ratio is not 1:1, I just crop down to 1:1 rather than zero pad. This is okay because the two images I want to register are going to have most of their meaningful information roughly centered in the frame.

I need the frequency response to be consistent across all orientations because the next step is to log-polar transform the frequency spectrum and then use phase correlation to capture theta/r differences (rotation and scaling, respectively).

Anyway, this is all a very specific situation. If you think the 'stretched' radial windows would be superior to separable windows for things like Gibbs artifacts, then I think the API change would be a great addition!

grlee77:
> Anyway, this is all a very specific situation. If you think the 'stretched' radial windows would be superior to separable windows for things like Gibbs artifacts, then I think the API change would be a great addition!

Hi @jni & @emmanuelle. As previous reviewers here, what are your thoughts on the API choice? The TLDR version is which of the following two would be preferable:

1.) isotropic windows where the user specifies a scalar `size` and `ndim`
2.) (optionally) anisotropic windows where the user speciifies a `shape` tuple (removing the `size` and `ndim` arguments)

mellertd:
You know, one other idea would be to make 'separable' windows the default and have a parameter for 'radial' windows that would enforce equal-length axes for the window. Unless, again, there is an advantage to the stretched radial windows over the separable windows for the general case. I have neither the knowledge nor the data to know whether that is the case.

mellertd:
So I went ahead and did some testing with synthetic data, and my feeling is that:

1. Separable (i.e., outer product) windows are kind of lousy
2. The stretched windows per @grlee77's comment are pretty close to cropping + symmetrical windows, at least for 2D windows
3. Cropping + symmetrical windows probably perform marginally better than stretched windows, but not enough for me to argue against the API change @grlee77 suggested.

So I would be totally happy with the API change and performance improvements -- if @grlee77 wants to contribute the change and the associated changes to the tests that would be appreciated!

If anyone is interested in checking my work behind this opinion, let me know and I'll tidy up the code and make it public. But I basically just created some synthetic images from a few sine waves at different angles and visually inspected the peaks in the abs() of the resulting ffts

grlee77:
> So I would be totally happy with the API change and performance improvements -- if @grlee77 wants to contribute the change and the associated changes to the tests that would be appreciated!

Sounds good. I can probably get to it later today.

emmanuelle:
@mellertd I'd be curious to know why separable windows are lousy? Is it a performance (speed / memory) issue, or more about the result?

mellertd:
@emmanuelle speaking of the result, with a caveat being that I didn't approach this quantitatively.
Examples with synthetic signal (3 different frequencies along x and y and rotated by some angle, all summed)


Result with no window:
nowin

Separable hann window:
separablewin

"Stretched" radial hann window:
stretchedwin

Cropped image, circularly symmetric hann window:
croppedwin

mellertd:
Note that those images are cropped to the peaks (roughly), which is why they might look a little weird

emmanuelle:
Hum interesting, thank you @mellertd. I wonder why the spectral leakage is so important for the separable windows...

mellertd:
It surprised me how much was there in the separable windows. I suspect it has something to do with the windows being still a bit "rectangle" shaped, but I also see similar artifacts with gaussian windows, so I am not entirely sure what's going on.

grlee77:
see mellertd/scikit-image#1 for an implementation of the (nonseparable) anisotropic window variant

mellertd:
Hmm, there seems to be a problem with master building with xcode. I think the changes in this PR are fine

mellertd:
> Hmm, there seems to be a problem with master building with xcode. I think the changes in this PR are fine

Ah, just saw #4369

jni:
@emmanuelle, I'll let you push the green button. =) @mellertd last thing, could you share your code for the comparison plots above? I think they would make for a useful exposition! (Doesn't have to be in this PR, but putting it in a gist and sharing a link here would help us to improve our docs on this feature in the future!)

mellertd:
@jni sure thing! I need to rework it for the new API, but that should actually make it much tidier (I was working in a notebook so it was in ugly shape anyway). I'll put something presentable together as a gist. Will try to get to it on Thursday

mellertd:
As promised, @jni : https://gist.github.com/mellertd/6ef1b8178ef2b61941568859e0a57a1e

One thing that should be noted is that in the example images I showed above, those had log transformed intensities, which does visually exaggerate the amount of spectral leak. This may have been why it looked so bad to @emmanuelle. So I included in the gist some masking of the origin of the FFT so that it can be plotted a little more "fairly". In those plots, the separable windows don't look so bad, although they still clearly don't perform nearly as well as those created by `window`.

emmanuelle:
Thank you for the nice addition and for your patience @mellertd ! Merging.

PR #4263, opened 2019-10-21
Fix peak finding for empty images when indices=True

ttung:
## Description

#3984 adds an optimization where a flat image immediately returns.  However, it returns an array where the shape is incorrectly set to always be 2D when indices=True.  A test is added to catch this scenario, and a fix is introduced where we set the output array's shape based on the input array's shape.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/tests/test_peak.py
+++ b/skimage/feature/tests/test_peak.py
@@ -234,8 +234,7 @@ class TestPeakLocalMax():
                                      footprint=np.ones((3, 3), bool),
                                      min_distance=1, threshold_rel=0,
                                      indices=True, exclude_border=False)
-        assert result.shape[0] == 0
-        assert result.shape[1] == image.ndim
+        assert result.shape == (0, image.ndim)
 
     def test_one_point(self):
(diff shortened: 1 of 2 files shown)


skimage/feature/tests/test_peak.py
```diff
@@ -228,6 +228,15 @@ def test_empty(self):
                                      indices=False, exclude_border=False)
         assert np.all(~ result)
 
+    def test_empty_non2d_indices(self):
+        image = np.zeros((10, 10, 10))
+        result = peak.peak_local_max(image,
+                                     footprint=np.ones((3, 3), bool),
+                                     min_distance=1, threshold_rel=0,
+                                     indices=True, exclude_border=False)
+        assert result.shape[0] == 0
```
rfezzani: May be use a single check:
```suggestion
        assert result.shape == (0, image.ndim)
```

ttung: applied.

PR #4265, opened 2019-10-22
Warn when rescaling with NaN in exposure.intensity_range

lagru:
## Description

I propose to use nanmin/nanmax by default for dtypes in `DTYPE_RANGE` that support NaN (complex type is not included) in `exposure.ìntensity_range`.

Consider this:

```python
>>> from skimage import exposure
>>> image = np.arange(12, dtype=float).reshape(3, 4)
>>> image[1, 1] = np.nan
>>> image
array([[ 0.,  1.,  2.,  3.],
       [ 4., nan,  6.,  7.],
       [ 8.,  9., 10., 11.]])
>>> exposure.rescale_intensity(image)
array([[nan, nan, nan, nan],
       [nan, nan, nan, nan],
       [nan, nan, nan, nan]])
```
This happens because `intensity_range` (called to find image range) simply uses `np.max` and `np.min`. If we pass in the desired range ourselves we get something more useful:
```python
>>> image_range = (np.nanmin(image), np.nanmax(image))
>>> exposure.rescale_intensity(image, image_range)
array([[0.        , 0.09090909, 0.18181818, 0.27272727],
       [0.36363636,        nan, 0.54545455, 0.63636364],
       [0.72727273, 0.81818182, 0.90909091, 1.        ]])
```
I doubt that most users actually expect the behavior where NaN broadcasts to the whole rescaled image. And not every user might immediately see the workaround. This just came up as a [question on Zulip](https://skimage.zulipchat.com/#narrow/stream/176554-Technical-problems/topic/exposure.2Erescale.20intensity.20with.20NaN.20values/near/178780205).

I'll add tests later in case this is wanted. 

## Checklist

- [x] <strike>Gallery example in `./doc/examples` (new features only)</strike>
  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


hmaarrfk:
I don't think we should make the whole library nansafe by default. It is a huge performance decrease

```
import numpy as np
a = np.zeros((2048, 2048), dtype='float32')
# Make sure to have the whole array in RAM
a[...] = 1
np.min(a)
Out[4]: 1.0

%timeit np.min(a)
1 ms ± 49.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit np.nanmin(a)
2.73 ms ± 78.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit np.min(a)
946 µs ± 31 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit np.nanmin(a)
2.99 ms ± 140 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

numpy has a diferent `min` function for a reason.

lagru:
That's fair and given that there is a reasonable workaround I'm inclined to agree for this case. How would you feel about raising a constructive warning detailing the workaround in case `imin` or `imax` is NaN?

hmaarrfk:
That could work, but there might very well be a need for nan-safe functions.

Some of our functions take in masks, which can mask away NaN from statistics. 

An other strategy is to add a parameter`nansafe=True/False` or something like that. As long as that argument isn't checked in a tight loop, that is fine.

Mostly, I want to make sure we retain the ability to remove the training wheels in favor of performance.

jni:
Thanks @hmaarrfk for raising that point. I would have missed it. 👍 for a `nansafe=` or equivalent keyword argument wherever it makes sense in the library. I don't love nansafe as a name though — it's a little bit technical. I don't have a better alternative but just want to prompt people to propose alternatives. Example: `ignore_na`.

sciunto:
I guess you meant ignore_nan? Otherwise, +1.

lagru:
Honestly I think adding a nan-safe toggle is not a full solution in this case. Users calling `rescale_intensity` might only notice much further down the line that NaN was broadcasted. An early warning in `rescale_intensity` would address this without a performance impact.

Adding a new keyword to two functions that is only effective if `in_range` / `range_values="image"` seems overkill to me and I like to avoid parameters whose effect depends on other parameters if possible. Although nan-safe toggles it might be a useful pattern elsewhere.

Problem with choosing a name is that the meaning of NaN-safe depends entirely on the function. One might ignore it while another one might treat it as a feature or another value. And although @jni doesn't like the name, `nan_safe` might be the most general applicable and the detailed meaning could be explained in the docstring.

skimage/exposure/exposure.py
```diff
@@ -346,6 +346,14 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     imin, imax = intensity_range(image, in_range)
```
rfezzani: The check for nans may be implemented only for `imin` an `imax` using
```python
np.isnan([imin, imax]).any()
```
If true, the user should be warned that the input image contains nans and that this causes the return array to be full of nans.  You may then directly return `np.full_like(image, np.nan)`

rfezzani: :-) I am sorry if my comment was not clear enough..
I mean that you don't need to check if `omax` and `omin` are nans and you can early stop the algorithm by returning `np.full_like(image, np.nan)` to simply prevents from doing unnecessary computations...

lagru: No problem. Actually we do need to check `omin` and `omax` as well. E.g. if `out_range="image"` they can be NaN as well. 

I'd like to avoid extra code paths if possible. And I consider cases were NaN is broadcasted to the full image as undesired behavior most of the time. Unnecessary computations don't matter then.

rfezzani: I only considered the case where `in_range` is `image`, here is my mistake, sorry.

skimage/exposure/exposure.py
```diff
@@ -346,6 +346,14 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
...
     omin, omax = intensity_range(image, out_range, clip_negative=(imin >= 0))
 
+    if np.isnan(imin + imax + omin + omax):
```
hmaarrfk: This seems like a hack. But otherwise ok

lagru: Is it a bad hack? :sweat_smile: I find it clear enough with the warning message directly below. It's faster than other options but seems insignificant when timing the whole function.

hmaarrfk: Just a hack. Can't comment if good or bad. But maybe requires a comment to those new to scientific computing.

rfezzani: I personally prefer `np.isnan([imin, imax, omin, omax]).any()`, but it's fine :-)

hmaarrfk: Thanks for the comment up there. I'm OK with this now.

lagru:
@hmaarrfk thanks for the changes and the lesson on how to handle the temporary deprecation warning! The CI still finds some [unexpected warnings](https://travis-ci.org/scikit-image/scikit-image/jobs/609978468#L2842). I'll look into it when I find the time.

emmanuelle:
@lagru did you have the chance to take a look at the CI failure? It seems to be genuinely associated with the changes. Are tests passing on your machine?

lagru:
@emmanuelle I think all warnings related to passing NaNs around NumPy functions should be addressed now. Those changed quite a bit between NumPy versions. I think the CI failure is due to a network timeout. I don't think I have the rights to trigger a rerun?

emmanuelle:
:+1:, just waiting for the CI buillds to complete.

jni:
Thanks everyone for the discussion here, I think we've ended up in a really good place!

lagru:
Agreed. :)

PR #4273, opened 2019-10-29
Add toggle for prompts in docstring examples next to copybutton

lagru:
## Description

As discussed in the [last meeting](https://github.com/scikit-image/meeting-notes/blob/master/2019/2019-10-28.markdown) sphinx_copybutton has the disadvantage that prefixes such as `>>> ` and output are copied as well. 

I propose to replace this with a simple javascript copied from [python/python-docs-theme](https://github.com/python/python-docs-theme/blob/b862f1159388ed3134818e56b065e18b835d652c/python_docs_theme/static/copybutton.js). The replacement allows to hide these and allows the user to select content by himself. This should make copying and running examples inside your own console a bit less tedious.

Not sure if I accommodated the license here in the right way. Do we need to include tools that are not included in the distribution archives in the license as well (e.g. as with tifffile)?

E.g. [skimage.exposure.histogram](https://scikit-image.org/docs/stable/api/skimage.exposure.html#histogram):

image
after clicking the button
image

## Checklist

## For reviewers

  later.

cc @jni, @alexdesiqueira

### DIFF
--- a/doc/source/themes/scikit-image/static/css/custom.css
+++ b/doc/source/themes/scikit-image/static/css/custom.css
@@ -297,6 +297,30 @@ p.sphx-glr-signature a.reference.external {
   display: none !important;
 }
 
+
+span.togglebutton {
+    cursor: pointer;
+    position: absolute;
+    top: 6px;
+    right: 25px;
+    font-family: "monospace";
+    line-height: 100%;
+    border-style: solid;
+    border-width: 1px;
+    border-radius: 2px;
+    padding-left: 0.2em;
+    padding-right: 0.2em;
+    margin: 0.5px;
+    color: #777777;  /* Matches sphinx_copybutton */
+    opacity: 0.3;
+    transition: opacity 0.5s;
+}
+
+.highlight:hover .togglebutton {
+    opacity: 1;
+}
+
+
 a.copybtn {
     background: transparent no-repeat none;
     border: none;
--- a/doc/source/themes/scikit-image/layout.html
+++ b/doc/source/themes/scikit-image/layout.html
@@ -17,6 +17,7 @@
 {%- macro script() %}
     <script src="https://code.jquery.com/jquery-latest.js"></script>
     <script src="{{ pathto('_static/', 1) }}js/bootstrap.min.js"></script>
+    <script src="{{ pathto('_static/', 1) }}js/togglebutton.js"></script>
     <script type="text/javascript">
       var DOCUMENTATION_OPTIONS = {
         URL_ROOT:    '{{ url_root }}',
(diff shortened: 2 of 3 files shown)


hmaarrfk:
I think you need to include the PSF license in the JS file somewhere maybe?

lagru:
@hmaarrfk Good idea, can do!

hmaarrfk:
This file is only used to a generate an other file correct? Therefore the license won't be shipped to every user on every page correct?

lagru:
It's included in every generated HTML page with `<script type="text/javascript" src="_static/copybutton.js"></script>`. That means it's loaded on every page. :/ Not sure what to do about that. `app.add_js_file` only allows to include a script file globally.

hmaarrfk:
Whatever, let's let the content delivery network guys worry about shipping the license text to the user.

lagru:
Sorry? :sweat_smile: What do you mean by that? 

[Test failure](https://travis-ci.org/scikit-image/scikit-image/jobs/604497522#L2563) looks unrelated.

hmaarrfk:
Typically people try to avoid the extra kilobyte of data from the license text. But that is a problem for the guys hosting the servers. Not us.

jni:
@lagru is there no way to have both buttons together? Maybe @choldgraf has ideas/opinions here. ;)

Other than that, looks good!

lagru:
@jni I think I figured it out:

image
image

I tried to get both buttons as consistent as possible but I'm a real beginner with JS and CSS. I couldn't get the tooltip to align in the same way that it does for the copybutton.

lagru:
@hmaarrfk Concerning the license: I opted to clearly state the license and a link to it in the scripts header. This seems to be standard practice when looking at other scripts loaded on our page.

doc/source/themes/scikit-image/static/css/custom.css
```diff
@@ -297,6 +297,30 @@ p.sphx-glr-signature a.reference.external {
...
 }
 
+
+span.togglebutton {
+    cursor: pointer;
+    position: absolute;
+    top: 6px;
+    right: 25px;
+    font-family: "monospace";
+    line-height: 100%;
```
lagru: I'm not sure whether the absolute positioning, padding, line-height and such will translate gracefully to other browsers and machines. Any CSS experts that have more insight?

hmaarrfk:
>  This seems to be standard practice when looking at other scripts loaded on our page.

It can be standard practice, but it doesn't mean that it is exactly correct.

According to point #2:
> provided, however, that PSF's License Agreement and PSF's notice of copyright

So you've included the notice of copyright, but not the PSF License Agreement itself.

I imagine much of the internet is tainted by this, people using part of other's code all the time without following 100% of the license agreement.

alexdesiqueira:
Thanlks @lagru! That'll be a great addition to our site :)

jni:
@meeseeksdev backport to v0.16.x

PR #2043, opened 2016-04-14
Add line_nd to draw module

jni:
Add a function drawing a line in n-dimensional space.

There is one failure edge case to the function, which is numpy's rounding behaviour to the nearest even number:

``` python
In [1]: import numpy as np
In [2]: arr = np.arange(0.5, 8, 1)
In [3]: arr
Out[3]: array([ 0.5,  1.5,  2.5,  3.5,  4.5,  5.5,  6.5,  7.5])
In [4]: np.round(arr)
Out[4]: array([ 0.,  2.,  2.,  4.,  4.,  6.,  6.,  8.])
```

This would result in broken lines. This can be detected, though. The question is what to do when detected. My preference would be to replace round with floor (ie rounding towards 0) when this happens:

``` python
In [5]: np.all(arr % 1 == 0.5)
Out[5]: True
In [6]: np.floor(arr)
Out[6]: array([ 0.,  1.,  2.,  3.,  4.,  5.,  6.,  7.])
```

Should I just add that in?
## Checklist

[It's fine to submit PRs which are a work in progress! But before they are merged, all PRs should provide:]
- [x] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Gallery example in `./doc/examples` (new features only)
- [ ] Unit tests

PR #4277, opened 2019-11-03
WIP: Use a n-dimensional Bresenham algorithm in draw_nd

marcotama:
## Description

This is a possible enhancement of the `draw_nd` function that uses the Bresenham algorithm (extended to n dimensions).
I suppose its strongest advantage is that it is unaffected by floating point error.

The current proposed version is vectorized in the number of dimensions, while a Python loop iterates over the various points. I suppose the original implementation is faster in most use cases although I have not run any tests.

The current algorithm is a WIP because it works exclusively on integers (as did the original Bresenham. Ideas for how to handle floats are welcome (I am currently truncating them).

My idea I have is to use the decimal value to initialize the errors, but to do that I think I need to find the least common denominator of said decimals after converting them to fractions, which has two problems:
    a) it only works with fractional values (ok, on a computer no really real numbers exist, but then the LCD could have a huge denominator resulting in huge initial errors (think plotting from `(0,0)` to `(cos(30°), sin(30°))`, and
    b) computing the LCD sounds like extra work one might not want to do (but I guess we have the `integer` flag for that?).

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/draw/_draw_nd.pyx
+++ b/skimage/draw/_draw_nd.pyx
@@ -1,19 +1,22 @@
-import numpy as np
-cimport numpy as np
 cimport cython
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-def _line_nd_cy(Py_ssize_t[::1] cur, Py_ssize_t[::1] stop, Py_ssize_t[::1] step,
-                *, Py_ssize_t[::1] delta, Py_ssize_t x_dim,
-                bint endpoint, Py_ssize_t[::1] error, Py_ssize_t[:, ::1] coords):
+def _line_nd_cy(
+        Py_ssize_t[::1] cur,
+        Py_ssize_t[::1] step,
+        Py_ssize_t[::1] delta,
+        *,
+        Py_ssize_t x_dim,
+        Py_ssize_t[::1] error,
+        Py_ssize_t[:, ::1] coords
+):
 
-    ndim = error.shape[0]
-    n_points = coords.shape[1]
-
+    cdef int n_dim = error.shape[0]
+    cdef int n_points = coords.shape[1]
     cdef int i_pt
     for i_pt in range(n_points):
-        for i_dim in range(ndim):
+        for i_dim in range(n_dim):
             coords[i_dim, i_pt] = cur[i_dim]
             if error[i_dim] > 0:
                 coords[i_dim, i_pt] += step[i_dim]
--- a/skimage/draw/_draw_nd.pyx
+++ b/skimage/draw/_draw_nd.pyx
@@ -1,0 +1,48 @@
+import numpy as np
+cimport numpy as np
+cimport cython
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def my_line_nd_cython(long[::1] start, long[::1] stop, bint endpoint=False, bint integer=True):
+    # TODO currently ignoring the decimal part of the number
+    if stop.shape[0] != start.shape[0]:
+        raise ValueError("Different lengths")
+    cdef int n = stop.shape[0]
+    cdef int[::1] delta = np.empty(n, dtype=np.intc)
+    cdef int[::1] steps = np.empty(n, dtype=np.intc)
+    cdef int n_points = 0
+    cdef int x_dim = -1
+    cdef int i_dim
+    for i_dim in range(n):
+        if stop[i_dim] >= start[i_dim]:
+            delta[i_dim] = stop[i_dim] - start[i_dim]
+            steps[i_dim] = +1
+        else:
+            delta[i_dim] = start[i_dim] - stop[i_dim]
+            steps[i_dim] = -1
+        
+        if abs(delta[i_dim]) > n_points:
+            n_points = delta[i_dim]
+            x_dim = i_dim
+    
+    if endpoint:
+        n_points += 1
+
+    cdef int[::1] cum_error = np.empty(n, dtype=np.intc)
+    cdef int[:, ::1] coords = np.empty([n_points, n], dtype=np.intc)
+    
+    for i_dim in range(n):
+        cum_error[i_dim] = 4 * delta[i_dim] - delta[x_dim]
+        coords[0, i_dim] = start[i_dim]
+    
+    cdef int i_pt
+    for i_pt in range(1, n_points):
+        for i_dim in range(n):
+            coords[i_pt, i_dim] = coords[i_pt-1, i_dim]
+            if cum_error[i_dim] > 0:
+                coords[i_pt, i_dim] += steps[i_dim]
+                cum_error[i_dim] -= 2 * delta[x_dim]
+            cum_error[i_dim] += 2 * delta[i_dim]
+        
+    return np.asarray(coords)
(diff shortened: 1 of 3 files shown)


jni:
Hey @marcotama! This is great! @stefanv was not super happy that we didn't get this right in #2043. There's a bit of cleanup to be done before merging, but in principle we totally want this in!

I'm sorry that I don't have time to do a full review right now (I'm not actually familiar with the Bresenham algorithm), but the way forward would be to replace the original `draw_line_nd` function with this one, *once* the inner loop is sped up by either vectorising or switching to Cython.

In other news, I notice you're in Melbourne??? We should meet up and finish this up in person! =D You can email me at juan.nunez-iglesias@monash.edu

marcotama:
I implemented a Cython version of the same algorithm, but I am not an expert, and I am pretty sure it can be optimized further. It's obviously faster than the original one, but a fair comparison would have Cython version of both.

Here are the results of a simple test; as expected, the `linspace` based algorithm outperforms  Bresenham on few dimensions, in the Python form.
comparison

@jni I sent you an email :)

jni:
@marcotama awesome! Can you produce the same plots for just 3D/4D and line length of 16, 32, ..., 16384? I think that is a much more common use case than 100-D data. :joy:

marcotama:
Obviously I wasn't thinking fourth-dimensionally! Silly me!

Here you go!

2dim
3dim
4dim

marcotama:
(ignore the lines, they are linear interpolations)

jni:
@marcotama I pushed our changes from today to your branch.

@scikit-image/core this is still WIP and not ready for review... Some of our adaptations of the algorithm to floating point values are not working correctly.

marcotama:
@jni Can you have a look at the current status? I believed I fixed the math, and the results are now almost identical to the existing implementation. (Cleanup remains a todo)

I say "almost" because there are differences due to a couple of reasons:
- I removed the banker's rounding (i.e. rounding to the nearest even number) for the following reason: banker's rounding was first introduced (afaik) to maintain (on average) the energy in "the system". By sometimes rounding `~.5` up and sometimes rounding it down you get that effect. However, in a graphic library I put forward the case that that should not be the policy. If that were the case, a line that passes at some `~.5` and is shifted up 1px at a time in an animation will show jitter if banker's rounding is used. This should not be the case.
- Different algorithms have different biases for cases when the line passes exactly between two pixels (regardless whether the bias is explicit, like when using rounding, or implicit, due to e.g. >/>= in the code on negative/positive numbers).

Obviously this needs to be double and triple checked ;)

marcotama:
I have written a little notebook to generate some plots to compare the two algorithms. Here are the plots of the test cases in the code (I made the 3D to 2D).

(0, 0) to (2, 2)
(0, 0) to (3, -5)
(0, 0) to (4, 2)
(0, 0) to (5, -3)
(1, 0) to (5, -1 5)
(1, 1) to (3, 3)
(1, 1) to (3, 5)
(1, 1) to (5, -2 4)
(1, 1) to (5, -2 5)
(1, 1) to (5, -2 6)
(1, 1) to (5, 2 5)
(1, 1) to (5, 3)
(1 1, 1 4) to (5 3, 2 2)

marcotama:
The code I used is:
```python
from skimage.draw.draw_nd import line_nd, my_line_nd
from matplotlib.collections import PatchCollection
from matplotlib.patches import Rectangle

tests = [
    ((0, 0), (2, 2)), # m=1 (x0 is even)
    ((1, 1), (3, 3)), # m=1 (x0 is odd)
    ((0, 0), (4, 2)), # m=2 (x0 is even)
    ((1, 1), (5, 3)), # m=2 (x0 is odd)
    ((1, 1), (3, 5)), # x is the second coordinate
    ((1, 1), (5, 2.5)), # y1 is decimal
    ((1.1, 1.4), (5.3, 2.2)), # all decimals (this produces a bug - a repeated point in the original implementation, due to the call to np.ceil (maybe a -0.5 is due in there))
    ((0, 0), (5, -3)), # m<0
    ((0, 0), (3, -5)), # m<0, x is the second coordinate
    ((1, 1), (5, 2.5)), # 3 dimensions
    ((1, 1), (5, -2.5)), # 3 dimensions, m>0 and m<0
    ((1, 0), (5, -1.5)), # 3 dimensions, m>0 and m<0, shifting causes different shape in the old implementation (compared to previous test)
    ((1, 1), (5, -2.4)), # just need to tease out why the results are what they are in this and the next example
    ((1, 1), (5, -2.6))
]

def plot_pixels(ax, pixels, line):
    facecolor='k'
    edgecolor='k'
    alpha=0.5
    x0, y0, x1, y1 = line
        
    ax.grid()
    ax.axes.set_aspect('equal', 'datalim')
    squares = []
    xs, ys = pixels
    for x, y in zip(xs, ys):
        square = Rectangle((x-.5, y-.5), 1, 1)
        squares.append(square)
    pc = PatchCollection(squares, facecolor=facecolor, alpha=alpha, edgecolor=edgecolor)
    ax.add_collection(pc)
    if abs(x1-x0) > abs(y1-y0):
        ax.set_xlim(min(xs) - 1, max(xs) + 1)
    else:
        ax.set_ylim(min(ys) - 1, max(ys) + 1)
    
    ax.plot([x0, x1], [y0, y1], color='k')
    
    

for start, stop in tests:
    fig, (ax1, ax2) = plt.subplots(nrows=1, ncols=2, figsize=(10,5))
    fig.suptitle(f"{start} to {stop}")
    line = start[0], start[1], stop[0], stop[1]
    
    ax1.set_title("Original")
    pixels1 = line_nd(start, stop, endpoint=True)
    plot_pixels(ax1, pixels1, line)
    
    ax2.set_title("New")
    pixels2 = my_line_nd(start, stop, endpoint=True)
    plot_pixels(ax2, pixels2, line)
    
    fig.savefig(f"{start} to {stop}.png")
```

jni:
@marcotama can you explain

```
    ((1, 1), (5, 2.5)), # 3 dimensions
```
? That is clearly two dimensions. =D

Thanks for the code and images, though, this is super thorough and easy to introspect! I love it! Gonna have a proper look in the next few days.

jni:
@marcotama looking at the examples again, this kinda hits the nail on the head for me:

> I aesthetically prefer the new algorithm, but the old algorithm seems to have less error

Is this caused by the choice of rounding or something else? It makes me tic a little. =P

>  These two show that the two algorithms have a different bias for rounding. There is not such thing as "correct".

See above.

I'd like to understand why the error is higher in the new version, as I don't really see why that should be the case. But, other than that, I love these diagnostics and am convinced that this is a better algorithm for this purpose. Thanks for the work so far @marcotama! It's great! I don't have time to review the code thoroughly right now but I hope to be able to do so later this week. 

I will say that eventually we want your code to completely replace my existing code, but happy to keep it there while we keep diagnosing this.

stefanv:
It would be fun to do the animation that you hinted at above; I wonder if it will show interesting artifacts.

It seems like the old algorithm has a tendency to be more "symmetric", which intuitively feels the way it should be; any intuition why?  I'm also surprised that the error is different between the two—what could cause that?

matthew-brett:

<details>

## Rationale

</details>

PR #4297, opened 2019-11-14
Prevent radon from upcasting float32 arrays to double

rfezzani:
## Description

If a `float32` array is provided to `radon` function, it will output a double precision array:

```python
>>> from skimage._shared.utils import convert_to_float
>>> from skimage.data import shepp_logan_phantom
>>> from skimage.transform import radon
>>> img = convert_to_float(shepp_logan_phantom(), False)
>>> radon(img).dtype
dtype('float64')
>>> radon(img.astype(np.float32)).dtype
dtype('float64')
```

This PR prevents this behavior and adds a test asserting the fix.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/transform/radon_transform.py
```diff
@@ -110,7 +110,8 @@ def radon(image, theta=None, circle=True, *, preserve_range=None):
     if padded_image.shape[0] != padded_image.shape[1]:
         raise ValueError('padded_image must be a square')
     center = padded_image.shape[0] // 2
-    radon_image = np.zeros((padded_image.shape[0], len(theta)))
+    radon_image = np.zeros((padded_image.shape[0], len(theta)),
+                           dtype=image.dtype)
 
     for i, angle in enumerate(np.deg2rad(theta)):
```
hmaarrfk: does theta need to be declared as single as well in this case?

rfezzani: Not necessarily: in `warp`,  only `padded_image` dtype acts on output dtype, but for consistency, the rotation matrix `R` can be cast to dtype. ;-)

PR #4298, opened 2019-11-14
Manage iradon input and output data type

rfezzani:
## Description

`iradon` actually accepts any input data type, even if the documentation stipulates that `radon_image` should be a float array.

This PR proposes to make use of `convert_to_float` to preprocess the input image.
The `preserve_range` parameter is added to the signature of `iradon` and care is taken to not upcast `float32` arrays to `double`.
~~The proposed behavior for `iradon` in this PR is to add a `dtype` parameter set to `None` by default and corresponding to the desired output data type:
By default, if input data type is not float, a warning is raised and input is cast to double, otherwise `dtype` is set to input data type.
If `dtype` is set to non floating point, a `ValueError` is raised.~~
Tests asserting correct output datatype is also added.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/transform/radon_transform.py
```diff
@@ -188,22 +188,22 @@ def _get_fourier_filter(size, filter_name):
     return fourier_filter[:, np.newaxis]
 
 
-def iradon(radon_image, theta=None, output_size=None,
-           filter="ramp", interpolation="linear", circle=True):
+def iradon(radon_image, theta=None, output_size=None, filter="ramp",
+           interpolation="linear", circle=True, preserve_range=False):
```
hmaarrfk: A test for the new parameter would go a long way. Otherwise, maybe just remove it as part of this PR

rfezzani: Thinking about a test for this parameter, I am skeptical about if we should modify users inputs values in this function...
I would rather just care about output dtype with respect with input dtype. What do you think?

rfezzani: I modified this PR in order to not modify users input data range. I removed the use of `convert_to_float` and introduced instead the `dtype` parameter. See the new description of the PR for details.

skimage/transform/radon_transform.py
```diff
@@ -262,6 +266,20 @@ def iradon(radon_image, theta=None, output_size=None,
...
+        else:
+            warn("Only floating point data type are valid for inverse radon "
+                 "transform. Input data is cast to float. To disable this "
+                 "warning, please cast image_radon to float.")
+            dtype = float
+    elif np.dtype(dtype).char not in 'efdg':
+        raise ValueError("Only floating point data type are valid for inverse "
+                         "radon transform.")
+
+    radon_image = radon_image.astype(dtype, copy=False)
```
hmaarrfk: I think instead you should have a `preserve_range` argument and call `img_as_float(preserve_range)` or osmething like that. By default, preserve range would be `True` with a deprecation cycle bringing back to None.

rfezzani: Is the deprecation cycle necessary? Wouldn't be `preserve_range` be directly set to `None` by default?

skimage/transform/radon_transform.py
```diff
@@ -262,6 +267,12 @@ def iradon(radon_image, theta=None, output_size=None,
     if filter not in filter_types:
         raise ValueError("Unknown filter: %s" % filter)
 
+    if preserve_range is None:
+        preserve_range = True
```
hmaarrfk: Did the old behavior act as True? If so, I think we are trying to deprecate that

rfezzani: In fact, previously `iradon` was accepting any input array data type and was casting it implicitly to `double`.
I remember now that I already saw in some discussion that `preserve_range` should be `False` by default.

emmanuelle:
Hi @hmaarrfk @rfezzani I think that in the future we'd rather try to have `preserve_range=True` by default because it's the less confusing option for users. @rfezzani was there a specific discussion about this?

rfezzani:
Hi @emmanuelle, sorry but I can't find the referred discussion :-(

hmaarrfk:
You would like uint8 inputs to have values between 0 and 255 float? Could we at least coordinate with matplotlib to make that the standard? That said, could we also coordinate with Vispy? 

That is a major change and sure to cause some roughness.

rfezzani:
I don't think that this should be the case, this PR is only about rationalising input and output dtypes. The `preserve_range` gives control to the user over the behavior of iradon.

jni:
> You would like uint8 inputs to have values between 0 and 255 float?

Yes. Over and over again I see that experienced users are annoyed by our automatic range conversion, and inexperienced ones are bitten by it ("why is my image black" is a frequently-asked question on SO and elsewhere).

> Could we at least coordinate with matplotlib to make that the standard?

We should. CC @tacaswell @story645 😬 What do you think about no longer assuming range in [0, 1] for float images? (Also! Looking forward to seeing you in 10 days! 😁)

> That said, could we also coordinate with Vispy?

I think VisPy does not assume anything at this point and always rescales to 0-1 internally. I *was* hoping to change that there too.

> That is a major change and sure to cause some roughness.

Agreed. But I think the question that @emmanuelle was getting at is where we want to go. Then we can decide how to get there. In this specific instance, `preserve_range` was already implicitly True, and the PR proposes to deprecate that behaviour towards False. I think that is stepping in the wrong direction.

The long term vision here for me is roughly:

- stop rescaling values (progressively, with deprecations, or like a bandaid in 1.0)
- educate users on how to rescale values where necessary
- provide a whole bunch of convenient rescaling functions and context managers.

rfezzani:
All right, so no deprecation should be applied here, right? I can directly set `preserve_range` to `True` since it was the implicit behavior....

rfezzani:
I removed the deprecation warning and set `preserve_range` to `True` by default.

rfezzani:

tacaswell:
In Matplotlib we do not infer anything about how to treat images from the type (but do look at the _shape_ to detect 2D we-need-to-colormap data from 3D RGB(A) data that we should just put on the screen (an the RGB(A) case I think we do look at the type but would have to check that)...). 

 We have had a number of bug reports which amounted to "why _arn't_ my images black" in cases where we were rescaling dark uint images to the data range (not the full theoreticical range).  I'm also not really sold on inferring the "full range" from the datatype in general, I have worked with too many weird bit-depth cameras.

rfezzani:
>  I'm also not really sold on inferring the "full range" from the datatype in general, I have worked with too many weird bit-depth cameras.

I agree with you @tacaswell, and that's why I like to use the `convert_to_float` function and its `preserve_range` argument instead of `img_as_float` that rescales input data according to data range.

jni (approved):
This LGTM, apologies to @rfezzani for letting this sit for so long!

rfezzani:
No worry @jni :wink:, thank you!

Issue #3111, opened 2018-05-28
Fused type required for floating point Cython operations


stefanv:
Since we're now (soon ;) letting through `float32` arrays in `img_as_float`, we need to add a fused dtype to all Cython code that consumes doubles to also accept single precision floats.

grlee77:
Agreed, although some cases may need to use double precision internally for accuracy.  An example is the change from float32 to float64 when using the fast integral-image representation for non-local means.  A good image demonstrating the problem with float32 is here:  https://github.com/scikit-image/scikit-image/pull/2878#issuecomment-346344773

jni:
@grlee77 for those functions we should be using `img_as_float64` instead of `img_as_float`.

rfezzani:
A big :+1: for this!
The `np_floats` fused type is already defined in `skimage._shared.fused_numerics`. Generalizing its use in Cython operations may in fact save memory use and computation time.
It also prevent some hidden data casting from `float32` to `double`, like for example in `_slic_cython` where the input image is silently converted to `double`...

rfezzani:
Oups, I just realized that this issue is more than one year old!!! :laughing:

stefanv:
I think we still need someone to do a proper review of all changes required; but since this issue is so vague I'm going to close it.

rfezzani:
I started this review yesterday when I noticed the silent cast to `double` in `_slic_cython`. That's how I found this discussion :-)
I will start by reporting the `_slic_cython` issue and refer to this one. I will try to investigate more on this.

PR #4300, opened 2019-11-15
Manage iradon_sart input and output data type

rfezzani:
## Description

In the same spirit of #4298, the input and output data type of `iradon_sart` is managed.

In this PR, a `dtype` parameter is added do `iradon_sart` and set to `None` by default and corresponds to the desired output data type:
By default, if input data type is not float, a warning is raised and input is cast to double, otherwise dtype is set to input data type.
If dtype is set to non floating point, a ValueError is raised.
The fused type `np_floats` is used to add single precision support in `_radon_transform` Cython module (see #3111).
Tests asserting correct output datatype is also added.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_radon_transform.pyx
+++ b/skimage/transform/_radon_transform.pyx
@@ -7,10 +7,11 @@ import numpy as np
 cimport numpy as cnp
 cimport cython
 from libc.math cimport cos, sin, floor, ceil, sqrt, abs, M_PI
+from .._shared.fused_numerics cimport np_floats
 
 
-cpdef bilinear_ray_sum(cnp.double_t[:, :] image, cnp.double_t theta,
-                       cnp.double_t ray_position):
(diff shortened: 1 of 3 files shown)


skimage/transform/_radon_transform.pyx
```diff
@@ -172,10 +173,10 @@ cpdef bilinear_ray_update(cnp.double_t[:, :] image,
 
 
 @cython.boundscheck(True)
-def sart_projection_update(cnp.double_t[:, :] image not None,
-                           cnp.double_t theta,
-                           cnp.double_t[:] projection not None,
-                           cnp.double_t projection_shift=0.):
```
hmaarrfk: Any clue why bounds check is enabled here?

rfezzani: Not really, I was also wondering why... I will investigate a little bit on this... Thank you anyway @hmaarrfk for your review.

PR #4307, opened 2019-11-18
Manage moments_hu output dtype

rfezzani:
## Description

`moments_hu` casts input image to double before processing.

This PR proposes to add a `preserve_range` parameter and to use `np_floats` fused type to add support for `float32` arrays (see #3111).
Tests for output data type are also added.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


jni:
I'm not sure that it's a good idea to *always* use the input image dtype. I might be missing something, but wouldn't the 8 bit inputs be very prone to overflow?

rfezzani:
Hi @jni, and thank you for your answer... That's exactly the kind of questions that came to my mind when I opened the issue #4308.
Should we then consider that any measure is floating point, and add a `dtype` parameter specifying if the computation is made using single or double precision like it is made in `registration.optical_flow_tvl1`?

emmanuelle:
I think moments computations should be done as floats (they are defined as part of integral calculus, which is by essence on real numbers). It might be different for other parts of the `measure` submodule. I'd be fine with a `dtype` parameter for moments functions.

rfezzani:
Thank you @emmanuelle for your comment, do you think that this approach (`dtype` argument) can be generalized to all the `meaure` module? May be can we elaborate on this in #4308?

jni:
Yes, I do think `dtype=` would be a useful generalisation in measure and elsewhere.

rfezzani:
I added the `dtype` parameter to `moments_hu`. By default, output data type will the same as input image if its dtype is floating point, otherwise double precision will be used.
The user may also specify which floating point precision he wants to use.

rfezzani:
Duplicate of #5200.

Issue #4261, opened 2019-10-19
sdist review (Oct 2019)


hmaarrfk:
This post here is to review the status of the sdist as part of scikit-image in light of recent developments in how python is now packaged in late 2019.

A few recent developments have been:
1. The removal of pyproject.toml https://github.com/scikit-image/scikit-image/pull/3914 
    * This was done because pyproject.toml would create a fresh environment, which could result building scikit-image with a newer version of numpy than the user had installed in their current environment. This would result in strange and confusing bugs.
2. The removal of the cython md5 hashing https://github.com/scikit-image/scikit-image/pull/3254
    * Unfortunately this only hashed the pyx (implementations) and not the pxd (declartions). This could cause the generation of files with different headers which could cause segfaults when developing cython code.

In the latest 0.16.0 release, we had a hiccup due to an old version of cython being used to generate the c files released in the sdist. @stefanv mentionned that we should consider not shipping c files with the sdist.

Numpy actually recommends very aggressively new versions of cython for their downstream packagers to use. We can maybe consider moving up our pinning to bleeding edge???

The major change in the python community has been the ease at which users on the major platforms can install scikit-image. Wheels are provided for all our core dependencies, and this greatly reduces the complexity of installing (and building from source) scikit-image.

To summarize:
* [ ] Should we move to bleeding edge cython?
* [ ] Should we remove pregenerated c files from the sdist?

stefanv:
I'm +1 on both. We should consider that sdist install will now fail until the right version of Cython is available, and print a helpful error message, similar to what we do for NumPy.

jni:
@hmaarrfk I feel like you left out

- [ ] should we re-add pyproject.toml with correct build dependencies (probably involves pinning numpy at build time but not at runtime?)

Note that the old Cython version was not due to an old version on the builder's machine, but rather, to not rebuilding the Cython files when using sdist. imho sdist should always result in a clean build of everything. It might be possible to hook in a `make clean` to the sdist command?

Cadair:
I am having a lot of issues with scikit-image and Python 3.8. This is because for scikit-image to compile from source numpy needs to be installed *before* running `pip install scikit-image`.

[...]

PR #4312, opened 2019-11-21
[WIP] Use PEP 517 Build Isolation

Cadair:
## Description

This is following on from a discussion in #4261 

This works for me currently when installing from a checkout in a completely clean venv. What I am unclear on at this point is how the C files are generated for the sdist, I am proposing that we remove that behaviour and have people generate the C files during build (with a pinned known version of Cython).

## For reviewers

  later.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools", "setuptools_scm", "wheel", "numpy==1.14.5", "cython==0.29.0"]
+requires = ["setuptools", "wheel", "numpy==1.14.5", "cython==0.29.0"]
 build-backend = 'setuptools.build_meta'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "setuptools_scm", "wheel", "numpy==1.14.5", "cython==0.29.0"]
+build-backend = 'setuptools.build_meta'
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,9 @@ For Python 2.7, please install the 0.14.x Long Term Support using:
     sys.stderr.write(error + "\n")
     sys.exit(1)
 
+# Add the current directory to the path to allow import of skimage
+sys.path.append(os.path.abspath("."))
+
 import builtins
 
 # This is a bit (!) hackish: we are setting a global variable so that the main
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "wheel", "numpy==1.14.5", "cython==0.29.0"]
```
Cadair: I have pinned numpy 1.14.5 as it's the first numpy version with py37 wheels on pypi. I have no idea what a good Cython pin is, there is no reason why we couldn't go very new here if we needed to.

setup.py
```diff
@@ -163,39 +166,16 @@ def configuration(parent_package='', top_path=None):
...
-        else:
-            print('To install scikit-image from source, you will need numpy.\n' +
-                  'Install numpy with pip:\n' +
-                  'pip install numpy\n'
-                  'Or use your operating system package manager. For more\n' +
-                  'details, see https://scikit-image.org/docs/stable/install.html')
-            sys.exit(1)
+    from numpy.distutils.core import setup
+    extra = {'configuration': configuration}
+    # Do not try and upgrade larger dependencies
```
Cadair: This feels very weird to me, but I have left it alone.

pyproject.toml
hmaarrfk: This works for python 3.8?

grlee77: I think the strategy is generally to choose the minimum NumPy that is both compatible with scikit-image and for which wheels exist. That is why SciPy has all of the following cases:

https://github.com/scipy/scipy/blob/6644a54ff862e4d90d7f9bc43ece9b0b1f681861/pyproject.toml#L5-L13

grlee77: Also, I think Cython >= 0.29.13 is the minimum compatible with Python 3.8.

Cadair: yes:

```
$ python --version
Python 3.8.0
$ pip freeze
pkg-resources==0.0.0
$ pip install .
```
numpy is now availible as a wheel for 3.8 but Cython 0.29.0 isn't, so I assume it has compiled it from source.

Cadair: I know that this build isolation used to rely on wheels existing, but it seems from this experiment that that is no longer the case. I wonder if I can find a pip reference for that somewhere.

PR #4318, opened 2019-11-24
WIP: force single axis shift from xy translation shift

PinkShnack:
## Description

This change to `register_translation` is a possible solution for [this stack overflow question](https://stackoverflow.com/questions/58388392/measuring-shift-between-two-images-along-one-direction-only), and situations where the user knows the image shift should be constrained to one axis, but where `register_translation` is "incorrectly" returning an xy shift. Essentially, it takes the xy shift translation, and forces it to just one axis with some triangles. It's limited to two dimensions right now.

I'm not sure if this "fits" into this function, or whether it's useful. Perhaps it could be its own function, or maybe people should do the maths themselves, any advice is greatly appreciated.

Also, maybe `axis` isn't the best param name here. `limit_to_axis` might be better.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -106,7 +106,7 @@ def _compute_error(cross_correlation_max, src_amp, target_amp):
 
 
 def register_translation(src_image, target_image, upsample_factor=1,
-                         space="real", return_error=True):
+                         space="real", return_error=True, axis=None):
     """
     Efficient subpixel image translation registration by cross-correlation.
 
@@ -134,6 +134,9 @@ def register_translation(src_image, target_image, upsample_factor=1,
     return_error : bool, optional
         Returns error and phase difference if on,
         otherwise only shifts are returned
+    axis : int, optional
+        The `axis` to which the translation shift is limited. Useful if the
+        user knows the translation shift should only be along one axis.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


jni:
@PinkShnack Thanks for implementing this! I'm not sure about it though. My feeling is that this solution only makes sense when the input has the specific form in the SO question: repetitive lines with a particular orientation (which is what you are calculating). My proposal instead actually *constrains* the translation to the specified direction, then finds the optimum along that axis. I think that (a) is more general, (b) is n-dimensional directly, and (c) can be constrained along multiple axes, rather than just one of two.

PinkShnack:
Indeed I thought as much. It's too limited and specific. Oh well. I has also tried a simple fftn.ifftn(...axes=(0,)) parameter, but of course it just calculates the x, y, or z shift. 
If I find the time I'll have to try they way you'd outlined. Would be really cool. Shall we leave this PR up, for just that?

emmanuelle:
Hi @PinkShnack did you have the time to take a look at implementing @jni's suggestion?

PinkShnack:
> Hi @PinkShnack did you have the time to take a look at implementing @jni's suggestion?

Hey @emmanuelle, I have not unfortunately. Would you like to have a go at it? 
I can also close this PR if it's a bit stagnant.

PinkShnack:
I'm going to go ahead and close this PR, as I won't be implementing the idea.

@emmanuelle Did you ever get around to having a go at this?

PR #4319, opened 2019-11-24
Add check to avoid import craching

asaf-kali:
Some `.ini` files are automatically generated by windows, and downloaded when using Google Drive's Backup and Sync app. 
When scanning for plugins in each `.ini` file, if 'providers' is not found in the file (the `meta-data` dict parsed from the file), the import craches.

### DIFF
--- a/skimage/io/manage_plugins.py
+++ b/skimage/io/manage_plugins.py
@@ -294,7 +294,7 @@ def _load(plugin):
     else:
         modname = plugin_module_name[plugin]
         plugin_module = __import__('skimage.io._plugins.' + modname,
-            fromlist=[modname])
+                                   fromlist=[modname])
 
     provides = plugin_provides[plugin]
     for p in provides:
--- a/skimage/io/manage_plugins.py
+++ b/skimage/io/manage_plugins.py
@@ -15,10 +15,9 @@ can be multiple states for a given plugin:
         loaded explicitly by the user.
 
 """
-import sys
-
-from configparser import ConfigParser
 import os.path
+import warnings
+from configparser import ConfigParser
 from glob import glob
 
 from .collection import imread_collection_wrapper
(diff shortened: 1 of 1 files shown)


jni:
... Hmm, it looks like this change breaks our plugin imports! See this error in our tests:

https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=1930&view=logs&jobId=0c528f75-1409-556f-1999-9f155ac37a3a&taskId=1c50252d-14f6-55f6-616c-dfe500c925e4&lineStart=31&lineEnd=45&colStart=1&colEnd=1

The errors need to be fixed before we can merge this.

skimage/io/manage_plugins.py
```diff
@@ -114,8 +112,9 @@ def _scan_plugins():
 
     for filename in config_files:
         name, meta_data = _parse_config_file(filename)
+        if 'providers' not in meta_data:
```
jni: Oh, I see the error: it's not 'providers', it's 'provides' (see line below).

```suggestion
        if 'provides' not in meta_data:
```

asaf-kali: Great. I'll take a look to see that it actually fixes the issues soon.
Thanks @jni!

jni:
@scikit-image/core please squash these commits on merge!

skimage/io/manage_plugins.py
```diff
@@ -114,8 +111,10 @@ def _scan_plugins():
 
     for filename in config_files:
         name, meta_data = _parse_config_file(filename)
+        if 'provides' not in meta_data:
+            warnings.warn(f'file {filename} not recognized as a scikit-image io plugin, skipping.')
```
hmaarrfk: should stack level be 3 or 4?

jni: It's unclear to me, but I also don't think it's important. In this case, the fix is not in the user code, but in the directory described in the error message.

hmaarrfk: but this would be the user's plugin right?

jni: Yes and no. If you look at the original bug report, the issue was Windows putting .ini files in random folders, including the plugins folder...

hmaarrfk:
Thanks

PR #4321, opened 2019-11-25
Add function alphanumeric_key in __all__ of io.collection.py

### LINKED ISSUE #4309: io documentation is incomplete
## Description

`skimage.io.collection.alphanumeric_key` is a useful function (my second addition to skimage! :smile:), but it is currently not documented (https://scikit-image.org/docs/dev/api/skimage.io.html), it is importable from `skimage.io.collection` instead of `skimage.io` (is this how we want it?), and cannot be found by `skimage.lookfor`.

## Way to reproduce
```python
import skimage
skimage.lookfor('alphanumeric_key')
from skimage.io.collection import alphanumeric_key  # works
```

sciunto:
## Description

Fixes #4309

Test:
```
In [3]: skimage.lookfor('alphanu' 
   ...: )                                                                                                                                                                                                                                    
Search results for 'alphanu'
----------------------------
skimage.io.alphanumeric_key
    Convert string to list of strings and ints that gives intuitive sorting.
skimage.io.ImageCollection
    Load and manage a collection of image files.

```


## Checklist

<!-- It's fine to submit PRs which are a work in progress! -->
<!-- But before they are merged, all PRs should provide: -->
  existing benchmark
- [ ] Unit tests

<!-- For detailed information on these and other aspects see -->
<!-- the scikit-image contribution guidelines. -->
<!-- https://scikit-image.org/docs/dev/contribute.html -->

## For reviewers

<!-- Don't remove the checklist below. -->
  later.

### DIFF
--- a/skimage/io/collection.py
+++ b/skimage/io/collection.py
@@ -14,7 +14,7 @@ from ..external.tifffile import TiffFile
 
 
 __all__ = ['MultiImage', 'ImageCollection', 'concatenate_images',
-           'imread_collection_wrapper']
+           'imread_collection_wrapper', 'alphanumeric_key']
 
 
 def concatenate_images(ic):


soupault:
Frankly speaking, I am not sure if this is a good addition to the public API, since the function is rather simple and cover quite specific naming case. Perhaps, we could have more core members to share their opinions.

jni:
well, we wanted to use it in at least one other place, namely napari. I also don't think it warrants its own package or anything (hello left-pad), and it is complex enough that I don't want to reimplement it everywhere... But I agree that it's kinda weird to provide it. I just don't think there is a better option.

hmaarrfk:
https://pypi.org/project/natsort/ ???

jni:
@hmaarrfk that's pretty nice, thanks!

sciunto:
Should I close this? Should we remove this function?

jni:
I would not remove it for sure — no chance we want that dependency. The only question is whether we want it to be part of the public API. Of that I’m not sure. I suggest we wait for more devs to weigh in.

sciunto:
I would be in favor to add the link to natsort in the dosctring as an indication.

stefanv:
Err, woops, just left a comment on the wrong issue.  Let's try again :)

@jni I'd suggest renaming this function to better explain what it does, as well as moving it from io to util.

lagru:
How about `as_alphanumeric_key` or `as_natural_sort_key`?

emmanuelle:
This one is a bit tricky, maybe we should take the opportunity of a dev meeting to discuss the best name.

emmanuelle:
As discussed in the Atlantic meeting Jan 31st, let's keep this function private and add an example with `natsort` in the documentation.

jni:
Cool, sounds good to me.

PR #4322, opened 2019-11-25
Fix denoise_nl_means data type management

rfezzani:
## Description

Based on @grlee77 [comment](https://github.com/scikit-image/scikit-image/issues/3009#issuecomment-472553515), this PR proposes to fix `denoise_nl_means` data type management:

- the `preserve_range` parameter is added to function signature,
- use `np_floats` fused type (see #3111).

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/restoration/non_local_means.py
```diff
@@ -132,16 +139,34 @@ def denoise_nl_means(image, patch_size=7, patch_distance=11, h=0.1,
...
+             'change to preserve_range=False. To avoid this warning, '
+             'explicitly specify the preserve_range parameter.',
+             stacklevel=2)
+        preserve_range = True
+
+    image = convert_to_float(image, preserve_range)
+    h = image.dtype.type(h)
+    sigma = image.dtype.type(sigma)
+
+    ctype = 'float32_t' if image.dtype == np.float32 else 'float64_t'
```
grlee77: This needs to use `float64_t` if `fast_mode` is True to avoid the numerical precision issues fixed previously in #2878. If the input was float32 for fast_mode we could still convert back to float32 after the Cython call to _nl_means_denoising..., but should use float64 internally.

rfezzani: Hi @grlee77, thank you very much for  pointing me out this very interesting issue, I will definitely give a look to the provided reference there!
In this PR, the input image is provided to `convert_to_float`, and any array with a dtype different then `float32` will become `float64`.
The use of `float32` arrays is generally intentional, and I think that users doing so are aware of this kind of problem :-)
Using this strategy gives the user full control on what he wants to do: if he cares about the precision issues, he only has to cast the input array to double by himself...

grlee77: > The use of float32 arrays is generally intentional, and I think that users doing so are aware of this kind of problem :-)

I doubt most users will be aware of the potential issue in this particular case. It is not just an issue of precision down at the level of 1e-7 or so, but potentially highly degraded output. A compromise could be to just raise a warning on float32 input with fast_mode True.

That PR discussion is a bit long, but there is a [nice visual example in this comment](https://github.com/scikit-image/scikit-image/pull/2878#issuecomment-346344773). Basically there are cumulative sums being used that become increasingly inaccurate for larger image sizes.

rfezzani: > I doubt most users will be aware of the potential issue in this particular case.

You're right... I just had a look at *fast_exp.h* and at the reference paper, the implemented `fast_exp` is double precision specific, and it is not really clear to me what cython is doing for single precision inputs...

I will investigate and see if there is any single precision alternative...

rfezzani: According to your [answer](https://github.com/scikit-image/scikit-image/pull/2878#issuecomment-346176122) 
>  I see that you have changed both variants to use double precision, but I think only the fast, integral image approach will have a precision that decreases with increasing image size.

Isn't only arrays with large integral values that suffers for the approximation error?
Using `convert_to_float` this kind of arrays is converted to `float64`...
Now the open question is what about `float32` with large values? Should some tests asserting `fast_exp` behavior with float32 arrays be added?

grlee77: Yes, the arrays that needs to always remain double precision is `integral` as used in `_integral_image_2d` and `_integral_image_3d`. This is where the large size of the summations accumulates a lot of error for large images/volumes. 

It looks like this is indeed still the case here so +1 from me

skimage/_shared/fast_exp.h
```diff
@@ -38,3 +41,64 @@ __inline double fast_exp (double y)
 
     return _eco.d;
 }
+
+/* Simple cast to float */
+__inline float _fast_expf(float y)
```
grlee77: I don't see that this version is still being used anywhere.  Isn't `fast_expf` now used instead?

skimage/restoration/tests/test_denoise.py
```diff
@@ -289,77 +289,92 @@ def test_denoise_bilateral_nan():
...
 
+        denoised_f32 = restoration.denoise_nl_means(img_f32, 7, 5, 0.2,
+                                                    fast_mode=fast_mode,
+                                                    multichannel=True,
+                                                    sigma=s)
+        # make sure noise is reduced
+        assert_(img.std() > denoised_f32.std())
+
+        # Sheck single precision result
+        assert np.allclose(denoised_f32, denoised, atol=1e-2)
```
grlee77: probably worth checking that the dtype matches our expectation

grlee77: actually I see you already added a separate test case for this below, so disregard this suggestion

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -80,16 +26,16 @@ cdef inline double patch_distance_2dmultichannel(IMGDTYPE [:, :, :] p1,
         Second patch, 2D image with last dimension corresponding to channels.
     w : 2-D array_like
         Array of weights for the different pixels of the patches.
-    s : int
+    s : Py_ssize_t
         Linear size of the patches.
-    var : double
+    var : np_floats
```
grlee77: I see that you moved the `2 * var` outside of this function for efficiency. That seems fine, but I would update the docstring here and possibly the variable name to indicate that this is not actually the variance of the noise anymore.
Maybe use a name something like `var_diff`, indicating that `var_diff == 2 * var` (i.e. var_diff is the variance of the subtraction of two random variables, each with variance `var`)

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -580,85 +520,90 @@ def _fast_nl_means_denoising_2d(image, int s=7, int d=13, double h=0.1,
...
+    cdef Py_ssize_t row_start, row_end, row_shift, col_shift
+    cdef Py_ssize_t offset = s / 2
+    cdef Py_ssize_t pad_size = offset + d + 1
+
+    cdef np_floats [:, :, ::1] padded = np.ascontiguousarray(
+        np.pad(image, ((pad_size, pad_size), (pad_size, pad_size), (0, 0)),
+               mode='reflect'))
+    cdef np_floats [:, :, ::1] result = np.zeros_like(padded)
+    cdef np_floats [:, ::1] weights = np.zeros_like(padded[..., 0], order='C')
+    cdef np_floats [:, ::1] integral = np.zeros_like(padded[..., 0], order='C')
```
grlee77: `integral` was previously initialized using `empty_like` rather than `zeros_like` which has slightly less overhead. Is there a reason this now needs to be zeros?

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -693,30 +638,28 @@ def _fast_nl_means_denoising_3d(image, int s=5, int d=7, double h=0.1,
...
-    cdef IMGDTYPE [:, :, ::1] result = np.zeros_like(padded)
-    cdef IMGDTYPE [:, :, ::1] weights = np.zeros_like(padded)
-    cdef IMGDTYPE [:, :, ::1] integral = np.empty_like(padded)
-    cdef int n_pln, n_row, n_col, t_pln, t_row, t_col, \
+    cdef Py_ssize_t pad_size = offset + d + 1
+    cdef np_floats [:, :, ::1] padded = np.ascontiguousarray(
+        np.pad(image, pad_size, mode='reflect'))
+    cdef np_floats [:, :, ::1] result = np.zeros_like(padded)
+    cdef np_floats [:, :, ::1] weights = np.zeros_like(padded)
+    cdef np_floats [:, :, ::1] integral = np.zeros_like(padded)
```
grlee77: same comment as above for `empty_like` vs `zeros_like` for `integral`

skimage/restoration/_nl_means_denoising.pyx
```diff
@@ -725,42 +668,49 @@ def _fast_nl_means_denoising_3d(image, int s=5, int d=7, double h=0.1,
...
-                    if t_col == 0 and (t_pln is not 0 or t_row is not 0):
-                        alpha = 0.5
-                    else:
-                        alpha = 1.0
+                    col_dist_min = offset
+                    col_dist_max = n_col - offset - t_col
 
                     # Compute integral image of the squared difference between
                     # padded and the same image shifted by (t_pln, t_row, t_col)
-                    integral[:, :] = 0
```
grlee77: Not sure why this only had [:, :] and not [:, :, :], but I guess either would be equivalent. Seems okay to remove it, but see comment above regarding doing a manual check of the demo output.

skimage/restoration/_nl_means_denoising.pyx
grlee77: On followup, I see that you have removed the resetting `integral[:, :] = 0` below. 

On first glance this seems to be okay, but probably worth double checking. I'm not sure our test cases are super strict. Can you manually verify that the PSNR values and output of [this demo](https://scikit-image.org/docs/dev/auto_examples/filters/plot_nonlocal_means.html#non-local-means-denoising-for-preserving-textures) still look consistent?

skimage/_shared/fast_exp.h
rfezzani: Hi grlee77, thank you very much for your review, I am currently doing some tests, and I am observing that the `fast_expf` implementation suffers a lot from the stair case effect described in the reference paper and I am hesitating to simply using this version `_fast_expf` with a simple cast from single to double precision which is the only overhead compared to the double precision implementation and still faster then the `expf` function.
I will post some figures to illustrate this, like @blochl did in #2878.

skimage/restoration/_nl_means_denoising.pyx
rfezzani: Yes, that's it, `integral` was set to zero inside the loop instead of being directly initialized to zero.

rfezzani:
Here are the results for the demo example @grlee77, every thing looks fine ;-)

## Double precision result

nl_means

```python
>>> print(f"PSNR (noisy) = {psnr_noisy:0.2f}")
PSNR (noisy) = 22.16
>>> print(f"PSNR (slow) = {psnr:0.2f}")
PSNR (slow) = 29.38
>>> print(f"PSNR (slow, using sigma) = {psnr2:0.2f}")
PSNR (slow, using sigma) = 29.77
>>> print(f"PSNR (fast) = {psnr_fast:0.2f}")
PSNR (fast) = 28.43
>>> print(f"PSNR (fast, using sigma) = {psnr2_fast:0.2f}")
PSNR (fast, using sigma) = 29.26
```
## Single precision result
nl_means_f32

```python
>>> print(f"PSNR (noisy) = {psnr_noisy:0.2f}")
PSNR (noisy) = 22.16
>>> print(f"PSNR (slow) = {psnr:0.2f}")
PSNR (slow) = 29.38
>>> print(f"PSNR (slow, using sigma) = {psnr2:0.2f}")
PSNR (slow, using sigma) = 29.77
>>> print(f"PSNR (fast) = {psnr_fast:0.2f}")
PSNR (fast) = 28.44
>>> print(f"PSNR (fast, using sigma) = {psnr2_fast:0.2f}")
PSNR (fast, using sigma) = 29.26
```

skimage/_shared/fast_exp.h
rfezzani: Here is the figure
fast_exp
As you can see, not really good :-(
May be @blochl, if you still have the data shown in your previous [PR](https://github.com/scikit-image/scikit-image/pull/2878#issuecomment-346344773), can you have a try with this implementation and see if you observe the same artifacts? Thank you.

skimage/restoration/_nl_means_denoising.pyx
grlee77: Great. Thanks for checking

skimage/_shared/fast_exp.h
grlee77: The inaccuracy shouldn't matter much for this application, but I would be fine with using the typecast version instead. I doubt it makes a substantial difference in run time either way.

skimage/_shared/fast_exp.h
```diff
@@ -3,15 +3,18 @@
  * Reference [2]: https://doi.org/10.1162/089976600300015033
  * Additional improvements by Leonid Bloch. */
 
+#include <stdint.h>
+
 /* use just EXP_A = 1512775 for integer version, to avoid FP calculations */
-#define EXP_A (1512775.3951951856938)  /* 2^20*ln2 */
+#define EXP_A (1512775.3951951856938)  /* 2^20÷ln2 */
```
blochl: Thanks for the fix! But can you please separate it to a different commit? It has nothing to do with introducing a single precision version. Also, maybe `/` instead of `÷` - non ASCII characters can cause problems in some scenarios. Nothing against using them when needed, but here a simple `/` will suffice.

skimage/_shared/fast_exp.h
```diff
@@ -3,15 +3,18 @@
...
 /* For min. RMS error */
 #define EXP_BC 1072632447              /* 1023*2^20 - 60801 */
 /* For min. max. relative error */
 /* #define EXP_BC 1072647449 */        /* 1023*2^20 - 45799 */
 /* For min. mean relative error */
 /* #define EXP_BC 1072625005 */        /* 1023*2^20 - 68243 */
 
+
```
blochl: Why an extra line? In two places below as well... In any case, it has nothing to do with introducing a single precision version. :)

skimage/_shared/fast_exp.h
```diff
@@ -38,3 +41,36 @@ __inline double fast_exp (double y)
...
     return _eco.d;
 }
+
+
+#define EXPF_A (184.6649652337873)  /* 2^7÷ln2 */
+/* For min. RMS error */
+#define EXPF_BC 16249              /* 127*2^7 - 7 */
+
+
+__inline float fast_expf (float y)
```
blochl: Is the 32-bit version really needed? When you work with small volumes/images the memory conservation is negligible, and with large ones - double precision is needed.

skimage/_shared/fast_exp.h
```diff
@@ -62,12 +90,12 @@ __inline float fast_expf (float y)
...
         case 1:
             /* Little endian */
-            _eco.n.j = (int)(EXPF_A*(y)) + EXPF_BC;
+            _eco.n.j = (int16_t)(EXPF_A*(y)) + EXPF_BC;
             _eco.n.i = 0;
             break;
         case 0:
             /* Big endian */
-            _eco.n.i = (int)(EXPF_A*(y)) + EXPF_BC;
+            _eco.n.i = (int16_t)(EXPF_A*(y)) + EXPF_BC;
```
blochl: :+1:

skimage/_shared/fast_exp.h
```diff
@@ -62,12 +90,12 @@ __inline float fast_expf (float y)
     switch(_eco.t[0]) {
         case 1:
             /* Little endian */
-            _eco.n.j = (int)(EXPF_A*(y)) + EXPF_BC;
+            _eco.n.j = (int16_t)(EXPF_A*(y)) + EXPF_BC;
```
blochl: :+1: 
I'd use `int32_t` explicitly in the original function as well (in a separate commit, obviously).

skimage/_shared/fast_exp.h
blochl: > May be @blochl, if you still have the data shown in your previous PR

Sorry, I don't have it anymore.
___
...And I still don't get why `{_,}fast_expf` are needed (reasoning in my review comment).

skimage/_shared/fast_exp.h
rfezzani: `/` will of course suffices :-)

skimage/_shared/fast_exp.h
rfezzani: One of my PEP8 habits... Not relevant here, you are right.

skimage/_shared/fast_exp.h
rfezzani: I introduces the 32-bit version because I still don't know what Cython is doing when fast_exp is called with a single precision float. I wanted to make it explicit.

blochl: I'm not a maintainer of this project, so I can only suggest, but to me it looks like `fast_exp` just shouldn't be called on single precision floats. Because of the reasoning above - with small datasets usage of double precision does not hurt much, and with large datasets - single precision is not enough.

rfezzani: From my point of view, the use of single precision arrays is an explicit user choice. Enforcing the use of double precision in this case is not the right option IMHO.

blochl: Okay, that is a valid point, but in any case, it feels like there should be no two single precision functions. I'd just pick one that works well, and keep it. But again, I'm not a maintainer, so please regard this as a mere opinion.

rfezzani: You are absolutely right, and if you have a look at [one](https://github.com/scikit-image/scikit-image/pull/4322#discussion_r353033407) of my previous comments, you can see that this is kind of exploratory and that the question of which version to keep is still open!

rfezzani:
I propose to add a note about the possible artifacts apparition due to round-off errors accumulation when single precision is used on large volumetric data. Is it OK for you?

grlee77:
> I propose to add a note about the possible artifacts apparition due to round-off errors accumulation when single precision is used on large volumetric data. Is it OK for you?

I would be okay with it, but it seems that a more conservative and backwards compatible solution is to just always leave the `integral` arrays as double rather than allowing fused types there. This wouldn't prevent using float32 for the input & output arrays.

skimage/_shared/fast_exp.h
emmanuelle: As @grlee77, I would not worry too much here about these (small) artifacts. The weights used in the non-local means algorithm are quite ad hoc anyway.

rfezzani:
@grlee77, I fixed the `integral` array type to `double`, and removed the note about round off error accumulation when using single precision. There is no more backward compatibility issue. Do you see any other issue with this PR?

rfezzani:
Thank you @grlee77 !

skimage/_shared/fast_exp.pxd
```diff
@@ -0,0 +1,17 @@
...
+cdef extern from "fast_exp.h":
+    double fast_exp(double y) nogil
+    float fast_expf(float y) nogil
+
+
+cdef inline np_floats _fast_exp(np_floats x) nogil:
+    if np_floats is cnp.float32_t:
+        return fast_expf(x)
+    else:
+        return fast_exp(x)
```
jni: @rfezzani could you add some benchmarks on float32 and float64 volumes and check before and after this PR, and report back? I would like to be reassured that this extra branch (?) isn't actually resulting in greater runtime for float64...

grlee77:
@jni:
 I just tried the basic benchmark pasted below (involving 256 x 256 x 32 volumetric data) and this PR is a bit faster than master.

|  Branch  | Precision | Duration
| -------- | --------- | --------
| master  | double | 4.27 s 
| master  | single | 4.39 s
| ----------- |  ----------| ----------
| this PR | double | 4.04 s
| this PR | single | 3.59 s

```Python
from time import time
import numpy as np
import skimage
from skimage.restoration import denoise_nl_means
import skimage.data

nz = 32
vol = np.stack([skimage.data.camera()[::2, ::2], ] * nz, axis=-1).astype(float) / 255
sigma = .05
voln = vol + sigma * np.random.randn(*vol.shape)
volnf = voln.astype(np.float32)

tstart = time()
denoise_nl_means(voln, patch_size=3, patch_distance=2, sigma=sigma, h=0.7*sigma, fast_mode=True)
dur_double = time() - tstart
print(dur_double)

tstart = time()
denoise_nl_means(volnf, patch_size=3, patch_distance=2, sigma=sigma, h=0.7*sigma, fast_mode=True)
dur_single = time() - tstart
print(dur_single)
```

rfezzani:
Thank you @grlee77 for tackling this, I complitely forgot about it (sorry @jni)

jni:
@grlee77 that's great news!

@rfezzani it would be great to add the above to our benchmarks in this PR. Have you seen @hmaarrfk's benchmarks writing guide? https://scikit-image.org/docs/dev/contribute.html#benchmarks

jni:
Woot! :tada:

rfezzani:
Thank you @jni

PR #4324, opened 2019-11-27
Fix border removal in peak_local_max with footprints

ttung:
## Description

The current code will erroneously remove the last row of an image when the footprint is size 1 along any dimension.  This is because `remove` will be 1, and (-1 // 2) is -1, and we slice out (-1, None).

In this PR, we update the `remove` to actually reflect the number of pixels we want to trim on each side of the image, and if it is 0, then we continue the loop.

Added tests to put a peak in each pixel in a 3x3 image, and verified that the peaks are correctly detected in all locations.  Prior to the fix, it would fail on any of the last row and last column.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -48,10 +48,14 @@ def _exclude_border(mask, footprint, exclude_border):
     """
     # zero out the image borders
     for i in range(mask.ndim):
-        remove = (footprint.shape[i] if footprint is not None
-                  else 2 * exclude_border)
-        mask[(slice(None),) * i + (slice(None, remove // 2),)] = False
-        mask[(slice(None),) * i + (slice(-remove // 2, None),)] = False
+        remove = (footprint.shape[i] // 2 if footprint is not None
+                  else exclude_border)
+        # if we want to remove 0 pixels, we need to bail.  otherwise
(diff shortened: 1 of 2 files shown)


skimage/feature/peak.py
```diff
@@ -48,10 +48,14 @@ def _exclude_border(mask, footprint, exclude_border):
     """
     # zero out the image borders
     for i in range(mask.ndim):
-        remove = (footprint.shape[i] if footprint is not None
-                  else 2 * exclude_border)
-        mask[(slice(None),) * i + (slice(None, remove // 2),)] = False
-        mask[(slice(None),) * i + (slice(-remove // 2, None),)] = False
+        remove = (footprint.shape[i] // 2 if footprint is not None
+                  else exclude_border)
```
jni: Should this be `int(exclude_border)`, or is that handled earlier?

ttung:
`test_blob.py` was broken to begin with.  when a footprint is passed into `peak_local_max`, `exclude_border` is only checked for truthiness.  A numerical value is ignored.  I'm going to make a new PR that makes `exclude_border` more sane as well as anisotropic.

ttung:
Closed in favor of #4325

PR #4325, opened 2019-11-28
Make exclude_border independent and anisotropic

ttung:
## Description

Currently, `exclude_border` is only tested for truthiness when a footprint is defined.  In reality, it should be independent and anisotropic.  This PR makes it both.

This adds tests to verify that `exclude_border` behaves as expected.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -1,6 +1,5 @@
 import numpy as np
 import scipy.ndimage as ndi
-from ..segmentation import relabel_sequential
 from .. import measure
 from ..filters import rank_order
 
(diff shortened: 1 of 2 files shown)


ttung:
The blobs test takes an image, makes many versions of it with varying degrees of gaussian blur.  It calls `peak_local_max` finder with exclude_border=5, expecting to remove 5 px from each dimension (which would be wrong even if `peak_local_max` interpreted exclude_border=5 correctly), but in fact was just removing pixels based on the footprint.

Prior to any change, it was removing 1 pixel across each of the dimensions closest to the origin, and 2 pixels across each of the dimensions farthest from the origin.  This happened to clip out the peak.

After #4324, it was removing 1 pixel across each of the dimensions closest to the origin, and 1 pixel across each of the dimensions farthest from the origin.  This happened to not clip out the peak, and reported a spot.

With this PR, it removes 5 pixels from each of the dimensions closest to the origin, and 5 pixels across each of the dimensions farthest from the origin.  This happened to clip out the peak.

The actual correct behavior requires modifying the blobs code, which I will do.

ttung:
See https://github.com/scikit-image/scikit-image/pull/4327 :)

skimage/feature/peak.py
```diff
@@ -85,9 +84,13 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
         Minimum intensity of peaks, calculated as `max(image) * threshold_rel`.
-    exclude_border : int or bool, optional
-        If nonzero int, `exclude_border` excludes peaks from
-        within `exclude_border`-pixels of the border of the image.
+    exclude_border : tuple of ints or bool, optional
+        If tuple of ints, the length of the tuple must match the input array's
+        dimensionality.  Each element of the tuple will exclude peaks from
+        within `exclude_border`-pixels of the border of the image along that
+        dimension.
+        If nonzero int, `exclude_border` excludes peaks from within
```
emmanuelle: Could you please put the "nonzero int" case first (before "tuple of ints") since I would expect it to be more common?

ttung: Updated.

ttung:
I added some test cases for illegal values (negative integers, and tuples containing negative integers).

emmanuelle:
Thanks a lot @ttung, merging!

ttung:
Any chance https://github.com/scikit-image/scikit-image/pull/4327 can also get reviewed? :)

PR #4327, opened 2019-11-29
Fix blob_log/blob_dog and their corresponding tests

ttung:
## Description

1. With the change in https://github.com/scikit-image/scikit-image/pull/4325, it's now possible to specify the excluded region on a per-dimension basis.  We can therefore explicitly exclude the borders for the image dimensions as needed.
2. test_blob_log_excl_border was calling blob_dog and not blob_log.  Whoops copy pasta.
3. exclude_border is set to 6.  If the circle is centered at (5, 5), one needs to chop off 6 pixels to remove the circle's center.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -185,6 +185,29 @@ def _prune_blobs(blobs_array, overlap, *, sigma_dim=1):
     return np.array([b for b in blobs_array if b[-1] > 0])
 
 
+def _format_exclude_border(img_ndim, exclude_border):
+    """Format an ``exclude_border`` argument as a tuple of ints for calling
+    ``peak_local_max``.
+    """
+    if isinstance(exclude_border, tuple):
+        if len(exclude_border) != img_ndim:
+            raise ValueError(
+                "`exclude_border` should have the same length as the "
+                "dimensionality of the image.")
+        for exclude in exclude_border:
+            if not isinstance(exclude, int):
+                raise ValueError(
+                    "exclude border, when expressed as a tuple, must only "
+                    "contain ints.")
+        return exclude_border
+    elif isinstance(exclude_border, int):
+        return (exclude_border,) * img_ndim + (0,)
+    elif exclude_border is True:
+        raise ValueError("exclude_border cannot be True")
+    elif exclude_border is False:
+        return (0,) * (img_ndim + 1)
+
+
 def blob_dog(image, min_sigma=1, max_sigma=50, sigma_ratio=1.6, threshold=2.0,
              overlap=.5, *, exclude_border=False):
     r"""Finds blobs in the given grayscale image.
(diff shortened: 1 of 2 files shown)


ttung:
This should be ready for review now.

skimage/feature/blob.py
```diff
@@ -185,6 +185,29 @@ def _prune_blobs(blobs_array, overlap, *, sigma_dim=1):
...
+                    "exclude border, when expressed as a tuple, must only "
+                    "contain ints.")
+        return exclude_border
+    elif isinstance(exclude_border, int):
+        return (exclude_border,) * img_ndim + (0,)
+    elif exclude_border is True:
+        raise ValueError("exclude_border cannot be True")
+    elif exclude_border is False:
+        return (0,) * (img_ndim + 1)
+
```
emmanuelle: I think a final `else` should be added to catch other cases (for example float values) and raise a `ValueError`.

skimage/feature/tests/test_blob.py
```diff
@@ -246,11 +246,11 @@ def test_blob_log_exclude_border():
     b = blobs[0]
     assert b[0] == b[1] == 5, "blob should be 5 px from x and y borders"
 
-    blobs = blob_dog(
+    blobs = blob_log(
         img,
         min_sigma=1.5,
         max_sigma=5,
-        exclude_border=5
+        exclude_border=6,
```
emmanuelle: humm why was the test passing before then?

emmanuelle:
Hi @ttung thank you for your PR and sorry it took so long to start a review. I left a couple of comments which are hopefully easy to address, hope we can merge this one soon!

skimage/feature/blob.py
ttung: Added.

skimage/feature/tests/test_blob.py
ttung: blob_* methods takes the input image of dimensionality N and applies a gaussian filter with 7 different sigmas, and makes an image of dimensionality N+1.  Then it applies peak_local_max across that composite image.  however, since `exclude_border` was previously _not_ anistropic and was actually ignored if a footprint was provided, it ended up zeroing out a bunch of the gaussian images.

example:

- input image is (500,500)
- blob_* takes the image and applies gaussian filter with varying sigmas, and produces a (7, 500, 500) image.
- previous behavior of exclude_border + footprint erroneously zeroes [0, :, :], [1, :, :], [5, :, :], and [6, :, :].

This resulted in `blob_log` finding no spots, not as a result of removing the border, but as a result of erroneously zeroing out data.

jni: @ttung brilliant explanation! 😀

skimage/feature/blob.py
```diff
@@ -185,6 +185,32 @@ def _prune_blobs(blobs_array, overlap, *, sigma_dim=1):
...
+        return exclude_border
+    elif isinstance(exclude_border, int):
+        return (exclude_border,) * img_ndim + (0,)
+    elif exclude_border is True:
+        raise ValueError("exclude_border cannot be True")
+    elif exclude_border is False:
+        return (0,) * (img_ndim + 1)
+    else:
+        raise ValueError(
+            "Unsupported value ({}) for exclude_border".format(exclude_border))
```
jni: You don't need to fix it for this PR, but note that we are now targeting Py 3.6 and above, so you can replace this with:

```suggestion
            f"Unsupported value ({exclude_border}) for exclude_border"
        )
```

emmanuelle:
Great thanks! I'll merge tomorrow unless you want first to change the string formatting, which is indeed more readable but your version is perfectly valid.

ttung:
I updated with the f-strings.

emmanuelle:
Brilliant! Thanks, merging :-).

PR #4329, opened 2019-11-30
Use gaussian filter output array if provided

### LINKED ISSUE #4014: `skimage.filters.gaussian` doesn't use `output` parameter at all

samir-nasibli:
## Description

Closes #4014
skimage.filters.gaussian didn't use the value of the output parameter. It returned only the image of float64 dtype, even if explicitly specify the output dtype. Fix it and add test.

* fix `gaussian` filter output bug
* add `test_preserve_output`
* add `test_output_error`
* update docstrings for `gaussian`
* add a note into TODO.txt

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


samir-nasibli:
Earlier the `gaussian` filters behavior was not quite expected
``` python
In [1]: from skimage.filters import gaussian
In [2]: import numpy  as np
In [3]: image = np.arange(25, dtype=np.uint8).reshape((5,5))
In [4]: image
Out[4]:
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14],
       [15, 16, 17, 18, 19],
       [20, 21, 22, 23, 24]], dtype=uint8)
In [5]: gaussian(image)
Out[5]:
array([[0.00855964, 0.01130214, 0.01497617, 0.0186502 , 0.0213927 ],
       [0.02227215, 0.02501465, 0.02868868, 0.03236271, 0.03510521],
       [0.04064229, 0.04338479, 0.04705882, 0.05073285, 0.05347535],
       [0.05901244, 0.06175494, 0.06542897, 0.069103  , 0.0718455 ],
       [0.07272495, 0.07546745, 0.07914148, 0.08281551, 0.08555801]])
In [6]: output = np.zeros_like(image)
In [7]: filtered_gaussian = gaussian(image, output=output)
In [8]: filtered_gaussian
Out[8]:
array([[0.00855964, 0.01130214, 0.01497617, 0.0186502 , 0.0213927 ],
       [0.02227215, 0.02501465, 0.02868868, 0.03236271, 0.03510521],
       [0.04064229, 0.04338479, 0.04705882, 0.05073285, 0.05347535],
       [0.05901244, 0.06175494, 0.06542897, 0.069103  , 0.0718455 ],
       [0.07272495, 0.07546745, 0.07914148, 0.08281551, 0.08555801]])
In [9]: id(filtered_gaussian)==id(output)
Out[9]: False
```

Even if you explicitly specified/allocated output dtype/array, it returned the image only float dtype and restored it in other memory.
Now `gaussian` also processes the output parameter and has the expected behavior:
``` python
In [1]: from skimage.filters import gaussian
In [2]: import numpy  as np
In [3]: image = np.arange(25, dtype=np.uint8).reshape((5,5))
In [4]: image
Out[4]:
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14],
       [15, 16, 17, 18, 19],
       [20, 21, 22, 23, 24]], dtype=uint8)
In [5]: gaussian(image)
Out[5]:
array([[ 1,  2,  3,  3,  4],
       [ 5,  6,  7,  7,  8],
       [10, 11, 12, 12, 13],
       [14, 15, 16, 16, 17],
       [18, 19, 20, 20, 21]], dtype=uint8)
In [6]: output = np.zeros_like(image)
In [7]: filtered_gaussian = gaussian(image, output=output)
In [8]: filtered_gaussian
Out[8]:
array([[ 1,  2,  3,  3,  4],
       [ 5,  6,  7,  7,  8],
       [10, 11, 12, 12, 13],
       [14, 15, 16, 16, 17],
       [18, 19, 20, 20, 21]], dtype=uint8)
In [9]: id(filtered_gaussian)==id(output)
Out[9]: True
```

samir-nasibli:
@hmaarrfkam, am I right that, the function should not convert everything to the floats, since scipy doesn't do so by default?

skimage/feature/_canny.py
```diff
@@ -171,7 +171,7 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
         mask = np.ones(image.shape, dtype=bool)
 
     def fsmooth(x):
-        return img_as_float(gaussian(x, sigma, mode='constant'))
+        return img_as_float(gaussian(x, sigma, output=float, mode='constant'))
```
rfezzani: This is not needed since `convert_to_float` is called internally in `gaussian`

skimage/filters/_gaussian.py
```diff
@@ -55,7 +55,7 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
     -----
     This function is a wrapper around :func:`scipy.ndi.gaussian_filter`.
 
-    Integer arrays are converted to float.
+    Integer arrays are converted to float, if output specified as float.
```
rfezzani: I personally, would not allow this case, since the documentation clearly stipulates that `output` is an array, but well...

skimage/filters/_gaussian.py
```diff
@@ -112,9 +112,17 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
...
-                               truncate=truncate)
+    if output is not None:
+        if isinstance(output, np.ndarray):
+            # convert image to float if output is float numpy ndarray
+            if issubclass(output.dtype.type, np.floating):
+                image = convert_to_float(image, preserve_range)
+        elif issubclass(type(output), type):
+            # convert image to float if output is float dtype
+            if issubclass(output, (np.floating, float)):
+                image = convert_to_float(image, preserve_range)
```
rfezzani: I again, personally, prefer to raise an error if output is not an array.

skimage/filters/_gaussian.py
```diff
@@ -112,9 +112,17 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
             sigma = [sigma] * (image.ndim - 1)
         if len(sigma) != image.ndim:
             sigma = np.concatenate((np.asarray(sigma), [0]))
-    image = convert_to_float(image, preserve_range)
-    return ndi.gaussian_filter(image, sigma, mode=mode, cval=cval,
-                               truncate=truncate)
+    if output is not None:
+        if isinstance(output, np.ndarray):
+            # convert image to float if output is float numpy ndarray
+            if issubclass(output.dtype.type, np.floating):
```
rfezzani: if output `dtype` is already float, `convert_to_float` will not touch it and return it as is...
Moreover, not converting non floating point arrays introduces a backward incompatibility, and core team will not be happy with it...

samir-nasibli:
Hi @rfezzani, thank you for your great review. You can see my other comment [here](https://github.com/scikit-image/scikit-image/issues/4014#issuecomment-566308804).

samir-nasibli:
I have update my PR according your logic from this [comment](https://github.com/scikit-image/scikit-image/issues/4014#issuecomment-562569651) and your review.
* added `output` check on numpy array
* added compatibility check for `output` type and `convert_to_float` output type according to input image `dtype`
* revert unnecessary changes in PR based on old logic

skimage/filters/_gaussian.py
```diff
@@ -112,9 +112,17 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
             sigma = [sigma] * (image.ndim - 1)
         if len(sigma) != image.ndim:
             sigma = np.concatenate((np.asarray(sigma), [0]))
-    image = convert_to_float(image, preserve_range)
-    return ndi.gaussian_filter(image, sigma, mode=mode, cval=cval,
-                               truncate=truncate)
+    if output is not None:
+        if not isinstance(output, np.ndarray):
+            raise ValueError("Output is not an numpy array")
+        else:
```
rfezzani: I think that you can remove this `else` block: what ever is image dtype, you can call `convert_to_float` an then check if output dtype corresponds to the converted image dtype and warn the user if it is not the case.

samir-nasibli:
@rfezzani update my PR. I am little bit not sure about warn message.

skimage/filters/_gaussian.py
```diff
@@ -112,9 +112,17 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
...
+            raise ValueError("Output is not an numpy array")
     image = convert_to_float(image, preserve_range)
-    return ndi.gaussian_filter(image, sigma, mode=mode, cval=cval,
-                               truncate=truncate)
+    if output:
+        if issubclass(image.dtype.type, output.dtype.type):
+            msg = ("Converted image dtype doesn't correspond output dtype, "
+                   "possible loss of data.")
+            warn(RuntimeWarning(msg))
+    return ndi.gaussian_filter(image, sigma, output=output, mode=mode,
```
rfezzani: I would suggest something like this

```suggestion
    image = convert_to_float(image, preserve_range)
    if output is None:
        output = np.empty_like(image)
    elif not isinstance(output, np.ndarray):
        raise ValueError("provided output argument is not a numpy array")
    elif image.dtype != output.dtype:
        warn("Invalid output datatype. A new array is assigned.")
        output = np.empty_like(image)
    ndi.gaussian_filter(image, sigma, output=output, mode=mode,
                                   cval=cval, truncate=truncate)
    return output
```
In fact, with your implementation, if output is not None, ndi.gaussian_filter returns None

rfezzani: My bad! ndi.gaussian_filter returns output even if it is  not None, sorry :-)

samir-nasibli: @rfezzani `ndi.gaussian_filter` works great, even if `output` is not specified : ) `ndi.gaussian_filter` uses funcs, that just call `scipy`'s own `_ni_support._get_output`. That's why no need for `output = np.empty_like(image)`.

samir-nasibli: If the `skimage.filters.gaussian`'s return is only floating data type, then your suggestion is ok, in this case it is needed to add in doc strings this requirement for output's dtype.  
But in case if we are saving `output`'s data type, then it looks like a bug.

rfezzani: Since `skimage.filters.gaussian` uses the `preserve_range` argument as input, which is then provided to the `convert_t_float` function, I suppose that it returns only floating data types. And you are right, it must be in fact clearly  documented.

samir-nasibli: Thank you for clarifying!
I was thinking that `gaussian` return's data type is the same as `output`'s, and actually if it's provided, then filtering result should be restored in `output`,  in other case image is converting to float and passing to `scipy.ndimage.gaussian_filter`. 
Existing `preserver_range` parameter does not give a clear understanding, that return type is floating. As the same as existing `cval` doesn't mean that `gaussian` uses only `constant` border mode, it will be ignored if `mode` is not `constant` (and this is documentated) :)

skimage/filters/_gaussian.py
```diff
@@ -113,8 +117,14 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
         if len(sigma) != image.ndim:
             sigma = np.concatenate((np.asarray(sigma), [0]))
     image = convert_to_float(image, preserve_range)
-    return ndi.gaussian_filter(image, sigma, mode=mode, cval=cval,
-                               truncate=truncate)
+    if not isinstance(output, np.ndarray):
```
rfezzani:
```suggestion
    if output is not None and not isinstance(output, np.ndarray):
```
;-)

skimage/filters/_gaussian.py
```diff
@@ -113,8 +117,14 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
...
+        raise ValueError("Provided output argument is not a numpy array")
+    elif image.dtype != output.dtype:
```
rfezzani: You also have to check if output is not None here before checking its data type!

skimage/filters/_gaussian.py
samir-nasibli: My fault, of course it is needed :)

skimage/filters/_gaussian.py
samir-nasibli: Added check  if `output` is None for all conditions.

skimage/filters/_gaussian.py
```diff
@@ -113,8 +117,15 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
...
-    return ndi.gaussian_filter(image, sigma, mode=mode, cval=cval,
-                               truncate=truncate)
+    if output is not None:
+        if not isinstance(output, np.ndarray):
+            raise ValueError("Provided output argument is not a numpy array")
+        elif image.dtype != output.dtype:
+            msg = ("Invalid output datatype. A new array is assigned.")
+            output = np.empty_like(image)
+            warn(RuntimeWarning(msg))
+    return ndi.gaussian_filter(image, sigma, output=output, mode=mode,
```
rfezzani: Looks like with minimum scipy version 0.19 supported by skimage, `ndi.gaussian_filter` doesn't return anything if `output` is not `None`! that's what is causing CI failure... I wasn't completely mistaken in my previous suggestion :-D

skimage/filters/_gaussian.py
```diff
@@ -121,8 +121,8 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
         if not isinstance(output, np.ndarray):
             raise ValueError("Provided output argument is not a numpy array")
         elif image.dtype != output.dtype:
-            msg = ("Invalid output datatype. A new array is assigned.")
-            output = np.empty_like(image)
+            msg = ("Invalid output datatype. A new array wiil be assigned.")
+            output = None
```
rfezzani: This will not solve the CI failures with scipy 0.19: In this case if output is None you can directly return ndi.gaussian_filter(image, **kwargs) as it is done now but when output is set to an array, you have to return output.

rfezzani: An other option would be to set output to an empty array if it is None, call ndi.gaussian_filter and then return output.

samir-nasibli: This doesn't solves CI failures with scipy 0.19, it just transfers the initialization of `output` to the `ndi.gaussian_filter`.

rfezzani: This is true, but the CI still fails :-)

samir-nasibli: It seems like for different versions of `scipy` `ndi.gaussian_filter` has different behavior, when `output` is `None`.

samir-nasibli: I have made update, if will not work, I will return `output`.

rfezzani: Unfortunately, CI with fail when scipy version is 0.19... You are right, `ndi.gaussian_filter` behavior changed with the more recent versions of scipy.

samir-nasibli: Have changed, now it returns output. But I think this may cause the other tests to crash.

rfezzani: The CI failure is not caused by your changes. IMHO every thing is fine ;-) Good job!

samir-nasibli: Thank you! I was confused with `gaussian skimage.filters.gaussian` input requirements and `ndi.ndimage.gaussian_filter` different versions behavoiur. So it took more time than I expected to fix this bug.
Anyway bug is fixed, docstrings and tests are added :)

samir-nasibli:
@rfezzani Thank you! I was confused with `gaussian skimage.filters.gaussian` input requirements and `ndi.ndimage.gaussian_filter` different versions behavoiur. So it took more time than I expected to fix this bug.
Anyway bug is fixed, docstrings and tests are added :)

skimage/filters/_gaussian.py
samir-nasibli: Thank you :)

skimage/filters/_gaussian.py
```diff
@@ -113,8 +117,17 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
         if len(sigma) != image.ndim:
             sigma = np.concatenate((np.asarray(sigma), [0]))
     image = convert_to_float(image, preserve_range)
-    return ndi.gaussian_filter(image, sigma, mode=mode, cval=cval,
-                               truncate=truncate)
+    if output is None:
+        output = np.empty_like(image)
+    elif not isinstance(output, np.ndarray):
```
hmaarrfk: I don't think this is quite right.

I kinda like making the assumption of `array-like` interfaces. 

Output is only used when users want to save on memory allocation and deallocation. they should know it is a numpy array, and not a list.

I would do
https://docs.scipy.org/doc/numpy/reference/generated/numpy.issubdtype.html

```
output is not None and np.issubdtype(output.dtype, np.floating)
```
and let scipy deal with cross floating point precision arithmetic.

skimage/filters/_gaussian.py
```diff
@@ -113,8 +117,17 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
...
+        raise ValueError("Provided output argument is not a numpy array")
+    elif image.dtype != output.dtype:
+        msg = ("Invalid output datatype. A new array is assigned.")
+        output = np.empty_like(image)
+        warn(RuntimeWarning(msg))
```
hmaarrfk: This is pretty strange and likely to go uncaught by a user. When somebody passes in the output array, they are doing so intentionally. The coding style:

```
output = func(...., output=output)
```
is the general case where this would not be a bug, but I think this is pretty terrible coding style.


IMO, users should either use:

```
output = func(...)
```


```
func(..., output=output)
```

Other libraries may wish to use the "bad syntax" to support users of either camp, but we shouldn't encourage end users to do so.

TLDR, if you don't like the data type provided by the user, raise an error.

IMO, just let scipy deal with that.

skimage/filters/tests/test_gaussian.py
```diff
@@ -82,3 +89,14 @@ def test_guess_spatial_dimensions():
     testing.assert_equal(_guess_spatial_dimensions(im4), 3)
     with testing.raises(ValueError):
         _guess_spatial_dimensions(im5)
+
+
+@pytest.mark.parametrize(
+    "dtype", [np.float32, np.float64]
+)
+def test_preserve_output(image, dtype):
```
hmaarrfk: I would get rid of the fixture. It isn't providing anything more than a convenience, (no speed increase), and hides the fact that it is a number between 0 and 10.

samir-nasibli: Very good point, hard to disagree.

skimage/filters/_gaussian.py
samir-nasibli: Thank you for this comment. I absolutely agree with this point. This code style may lead to uncaught bugs, but I just I follow scipy logic.

samir-nasibli: I am also agree with raising type error, when data type provided by user is not correspond with image's one. Unlike you other reviewer instead on warning. I followed the recommendation.

skimage/filters/_gaussian.py
samir-nasibli: If user know that, the `output` is numpy array and this information is well documented, then the suggestion is pretty good.

hmaarrfk:
@samir-nasibli sorry about going back forth on the recommendation. I tried to see what the user wrote, but didn't catch anything in my scan.

skimage/filters/_gaussian.py
hmaarrfk: The reason i suggest against this is that many people are experimentng with numpy-like arrays.

GPU arrays, distributed arrays, they are becoming fashionable, and pretty useful!

skimage/filters/tests/test_gaussian.py
```diff
@@ -82,3 +84,14 @@ def test_guess_spatial_dimensions():
     testing.assert_equal(_guess_spatial_dimensions(im4), 3)
     with testing.raises(ValueError):
```
hmaarrfk: You should include a test to ensure the error you want is raised.

hmaarrfk:
@jni do you want to look over the proposed behavior?

skimage/filters/tests/test_gaussian.py
samir-nasibli: `test_output_error` is added for it.

samir-nasibli:
> @samir-nasibli sorry about going back forth on the recommendation. I tried to see what the user wrote, but didn't catch anything in my scan.

@hmaarrfk due to misunderstanding between members, contributors and me, misunderstanding of requirements this PR has grown over 30 commits. If this is necessary for your convenience, then it will not be difficult for me to close this PR and transfer all the changes to the new PR, making within 2 commits there.

skimage/filters/_gaussian.py
samir-nasibli: You are right, it is very useful comment. I think in this case definitely no need to limit the output type.

jni:
@samir-nasibli don't worry, if needed we will squash the history ourselves. The only thing that matters is where the PR is at merge time. =)

skimage/filters/_gaussian.py
```diff
@@ -57,6 +57,10 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
 
     Integer arrays are converted to float.
 
+    The ``output`` should be floating point data type, since gaussian converts
```
hmaarrfk:
```suggestion
    The ``output`` should be floating point data type since gaussian converts
```

skimage/filters/_gaussian.py
```diff
@@ -57,6 +57,10 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
...
+    to float provided ``image``. Otherwise, another array will be assigned
```
hmaarrfk:
```suggestion
    to float provided ``image``. If ``output`` is not provided, another array will be allocated
    and returned as the result.
```

skimage/filters/_gaussian.py
```diff
@@ -113,8 +117,13 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
         if len(sigma) != image.ndim:
             sigma = np.concatenate((np.asarray(sigma), [0]))
     image = convert_to_float(image, preserve_range)
-    return ndi.gaussian_filter(image, sigma, mode=mode, cval=cval,
-                               truncate=truncate)
+    if output is None:
+        output = np.empty_like(image)
```
hmaarrfk: What if you simply omitted this and allowed scipy to allocate it?

skimage/filters/_gaussian.py
```diff
@@ -113,8 +117,13 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
...
+    elif not np.issubdtype(output.dtype, np.floating):
+        raise ValueError("Provided output data type is not float")
+    ndi.gaussian_filter(image, sigma, output=output, mode=mode, cval=cval,
```
hmaarrfk: `output = ` ??? This would allow scipy to allocate things right?

rfezzani: The problem is that in scipy version 0.19, which is still supported by skimage, if `output` is not `None`, `ndi.gaussian_filter` returns None :-(

hmaarrfk: I see. 

Yeah. Let's leave this PR as is and maybe bump up scipy minimum version.

samir-nasibli: @hmaarrfk different versions of `scipy.ndimage.gaussian_filter` has different behavior, when output is None.

samir-nasibli:
@hmaarrfk due to last [changes](https://github.com/scikit-image/scikit-image/pull/4372) it may be appropriate to replace my latest additions with this:
``` python
    image = convert_to_float(image, preserve_range)
    if output is not None and not np.issubdtype(output.dtype, np.floating):
        raise ValueError("Provided output data type is not float")
    return ndi.gaussian_filter(image, sigma, output=output, mode=mode,
                               cval=cval, truncate=truncate)
```
Removing allocating `output` and allowing scipy to do this. But in this case we are getting to our earlier [discussion](https://github.com/scikit-image/scikit-image/pull/4329#discussion_r360561344).

Indeed, maybe it would be better to leave it as is.

hmaarrfk:
In my opinion, this contribution is ready as is.

If you want to be perfect, then I would strive to avoid allocations, but yeah, it works as is.

samir-nasibli:
so, I have removed allocations and allow `scipy` to do this. I am striving for the best :)

samir-nasibli:
So, we have the same failures, because of having different behavior on `scipy` version >= 1.0.x. for `scipy.ndimage.gaussian_filter`. The problem can be solved if the minimum `scipy` version >=1.1.x

samir-nasibli:
@hmaarrfk If there is no way to bump up `scipy` minimum version, than I can easily revert my last commit.

hmaarrfk:
1.1 is a harder argument since that would still be 6 months out

hmaarrfk:
I guess the real way to be better is to add a note in the TODO.txt that this needs to be changed when scipy upgrades to 1.1

jni:
Thank you @samir-nasibli for this contribution, and @hmaarrfk and @rfezzani for your reviews! :tada:

PR #4330, opened 2019-11-30
Move cython pinning forward

hmaarrfk:
## Description
Cython 0.29.13 is necessary for Python 3.8 support

If this pinning is good enough for scipy, it is good enough for us!
https://github.com/scipy/scipy/blob/master/pyproject.toml#L5

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


emmanuelle:
Just a naive question BTW: why do we need to bump the dep here, wouldn't pip install automatically a cython version compatible with Python 3.8 if the version of Python used is 3.8?

hmaarrfk:
that is if we stop shipping C files. That seems like  harder decision ;)

hmaarrfk:
and harder to ensure that the new build process goes well.

jni:
@emmanuelle imho we should aim for the generated C code to match in all Python versions. I think it's probably a good idea to be on the bleeding edge with Cython, as they are making performance improvements all the time, and it also frees us to use more modern constructs like their [pure Python mode](https://cython.readthedocs.io/en/latest/src/tutorial/pure.html).

I'll let this sit for a bit longer in case @stefanv or others want to comment. Thanks @hmaarrfk!

stefanv:
This is the simplest solution for now.  In the longer term, I agree that we should stop shipping C files in the source bundle.

emmanuelle:
Thanks everyone for the explanations, merging!

PR #4337, opened 2019-12-03
Fix CI failures caused by incompatible imagecodec minimum requirement 

rfezzani:
## Description

NumPy version is bumped to 1.14.6 to fix the following CI failures:

>ERROR: imagecodecs-lite 2019.12.2 has requirement numpy>=1.14.6, but you'll have numpy 1.14.1 which is incompatible.
> ERROR: tifffile 2019.7.26.2 has requirement numpy>=1.14.6, but you'll have numpy 1.14.1 which is incompatible.

### DIFF
--- a/requirements/build.txt
+++ b/requirements/build.txt
@@ -1,3 +1,3 @@
 Cython>=0.29.13
 wheel
-numpy>=1.14.1
+numpy>=1.14.6
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,4 +1,4 @@
-numpy>=1.14.1
+numpy>=1.14.6
 scipy>=0.19.0
 matplotlib>=2.0.0,!=3.0.0
 networkx>=2.0


rfezzani:
It looks like there is a small bug in @cgohlke `tiffile` package lines 645 and 659 causing the travis-CI failure:
```python
try:
    import imagecodecs
except ImportError:  # here
    import zlib

    try:
        import imagecodecs_lite as imagecodecs
    except ImportError:  # and here
        imagecodecs = None
```
`ModuleNotFoundError` should be added to the exceptions.

cgohlke:
> ModuleNotFoundError should be added to the exceptions.

I don't think so. ModuleNotFoundError is derived from ImportError.

jni:
I noticed this kind of error today installing packages for a workshop, and I had never seen it before. Is this a bug in a newly released pip? There is no contradiction, pip just needs to install NumPy >= 1.14.6 and all requirements will be satisfied. What am I missing?

rfezzani:
You are right @cgohlke, sorry for the spam  :pray:

rfezzani:
Looks like today's last version of `imagecodecs-lite` fixes the CI problems: see [this](https://travis-ci.org/scikit-image/scikit-image/jobs/620543189#L2120) compared to [this](https://travis-ci.org/scikit-image/scikit-image/jobs/620102235#L2311).
I think that this PR can be closed!

mrakitin:
We at @NSLS-II are observing the same issue, reported via https://github.com/blink1073/tifffile/issues/60.

emmanuelle:
ok, it seems that the PR can be closed, if we have second thoughts we can always re-open. Thanks anyway @rfezzani !

PR #4338, opened 2019-12-04
Add explicit use of 32-bit int in fast_exp

rfezzani:
## Description

Implements @blochl [advice](https://github.com/scikit-image/scikit-image/pull/4322#discussion_r353567366).

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/fast_exp.h
+++ b/skimage/_shared/fast_exp.h
@@ -4,7 +4,7 @@
  * Additional improvements by Leonid Bloch. */
 
 /* use just EXP_A = 1512775 for integer version, to avoid FP calculations */
-#define EXP_A (1512775.3951951856938)  /* 2^20*ln2 */
+#define EXP_A (1512775.3951951856938)  /* 2^20/ln2 */
 /* For min. RMS error */
 #define EXP_BC 1072632447              /* 1023*2^20 - 60801 */
 /* For min. max. relative error */
--- a/skimage/_shared/fast_exp.h
+++ b/skimage/_shared/fast_exp.h
@@ -3,6 +3,8 @@
  * Reference [2]: https://doi.org/10.1162/089976600300015033
  * Additional improvements by Leonid Bloch. */
 
+#include <stdint.h>
+
 /* use just EXP_A = 1512775 for integer version, to avoid FP calculations */
 #define EXP_A (1512775.3951951856938)  /* 2^20/ln2 */
 /* For min. RMS error */
@@ -17,7 +19,7 @@ __inline double fast_exp (double y)
     union
     {
         double d;
-        struct { int i, j; } n;
+        struct { int32_t i, j; } n;
         char t[8];
     } _eco;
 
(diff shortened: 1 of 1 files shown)


skimage/_shared/fast_exp.h
```diff
@@ -3,8 +3,10 @@
  * Reference [2]: https://doi.org/10.1162/089976600300015033
  * Additional improvements by Leonid Bloch. */
 
+#include <stdint.h>
+
 /* use just EXP_A = 1512775 for integer version, to avoid FP calculations */
-#define EXP_A (1512775.3951951856938)  /* 2^20*ln2 */
+#define EXP_A (1512775.3951951856938)  /* 2^20/ln2 */
```
blochl: Since this fix is unrelated, please put it to a separate commit. It will make things easier if the functional part of this commit will ever need to be reverted, and in general making the commits as specific as possible is a good practice.

blochl:
@rfezzani, thanks, but what you have done now is introduce two completely redundant commits to the code: the original commit, then the revert of the original commit, and then the two correct commits. This is a clutter for the Git tree. Please read about [editing commits](https://stackoverflow.com/questions/1186535/how-to-modify-a-specified-commit) if you feel like it.

What you can do now, is `git rebase -i 7a3e0d7902^` (note the `^` in the end) and in the editor that opens automatically, delete the two lines which correspond to the first commit and its revert - these should be the first two lines, save and close. Then verify that they are indeed gone in your branch (`git log --graph --oneline --decorate --all`, or `git log --graph --oneline --decorate`) and then **force** push your branch for modifying this PR.

hmaarrfk:
how was this working before?

can we create a test for this if we enable aarch64 on travis?>?

rfezzani:
Hi @hmaarrfk, in #4322, I added a `fast_exp.pyx` file for which I planed to add tests. Do you prefer this to be done here?

blochl:
> how was this working before?
> 
> can we create a test for this if we enable aarch64 on travis?>?

int is 32-bit by default on most platforms, on aarch64 as well.

hmaarrfk:
hmm, ok you are right! i guess i forgot that..... so confusing

skimage/_shared/fast_exp.h
hmaarrfk: We squash many PRs. @rfezzani if you want us to merge each commit, you will have to rebase and clean the history a little as suggested by blochl

rfezzani: I don't really care if the PR is squashed. I applied blochl suggestion to not be lazy :-)

blochl: Thanks!!
Not sure if squashing the entire PR is a good idea for later references/fixes/feature history, but hey - not my call. :)

hmaarrfk:
I guess you can do what you want in terms of incorporating tests. IMO, this is OK to merge as is.

jni:
I'm curious about why this is needed. In general, my preference is to let the compiler/OS choose the "best" int type for the platform, unless there is good reason to explicitly choose one type over the other. Sorry, I haven't been following the full discussion from the beginning, but I tried following a few links to the previous discussion and didn't see any comment that properly explained why int32 should be the appropriate data type here...?

@blochl I admire your zeal for clean commit histories, but we are generally quite lax at skimage with it, and prefer to squash (thinking that a PR itself should be *reasonably* self-contained) rather than force contributors to jump through many git hoops. Many newcomers might have found your git instructions quite intimidating!

Thanks everyone for all the work so far on this issue, it is impressive!

rfezzani:
Hi @jni, 32-bit integer are required for the method to work(see the reference [here](https://schraudolph.org/pubs/Schraudolph99.pdf)), it's part of the trick ;-)

hmaarrfk:
Yeah. I read the reference a while back. The trick is actually really useful for 8 bit integers.

jni:
Very good, thank you for the pointers! =)

PR #4340, opened 2019-12-04
add binder links to dev example pages

emmanuelle:
Here is a proof-of-concept PR showing how to add binder links to examples of the gallery.

A few questions / comments 
- when we merge something on master, only the dev folder in `docs/gh-pages` is updated, is this correct? I need to find the scripts which generate the documentation. 
- if for now we have binder links only in the `dev` version, then in the new `requirements.txt` file (which is for binder) we should use scikit-image master. I know it's possible to pip install from github but I don't know if you can do the same in a requirements.txt, will have to check.

If you want to check the changes, do `make html` inside the `doc` directory. You will see how a `notebooks` and a `binder` directories are generated.

### DIFF
--- a/doc/binder/requirements.txt
+++ b/doc/binder/requirements.txt
@@ -1,0 +1,6 @@
+# For binder
+numpy
+scikit-learn
+pandas
+seaborn
+matplotlib
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -68,6 +68,17 @@ sphinx_gallery_conf = {
         '../examples/applications',
         '../examples/developers',
     ]),
+    'binder': {
+     # Required keys
+     'org': 'scikit-image',
+     'repo': 'docs',
+     'branch': 'gh-pages',  # Can be any branch, tag, or commit hash. Use a branch that hosts your docs.
+     'binderhub_url': 'https://mybinder.org',  # Any URL of a binderhub deployment. Must be full URL (e.g. https://mybinder.org).
+     'dependencies': '../requirements.txt',
+     # Optional keys
+     'filepath_prefix': 'dev/', # A prefix to prepend to any filepaths in Binder links.
+     'use_jupyter_lab': False  # Whether Binder links should start Jupyter Lab instead of the Jupyter Notebook interface.
+     }
 }
 
 # Determine if the matplotlib has a recent enough version of the
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -74,7 +74,7 @@ sphinx_gallery_conf = {
         'repo': 'docs',
         'branch': 'gh-pages',  # Can be any branch, tag, or commit hash
         'binderhub_url': 'https://mybinder.org',  # Any URL of a binderhub.
-        'dependencies': '../requirements.txt',
+        'dependencies': '../binder/requirements.txt',
         # Optional keys
         'filepath_prefix': 'dev/', # A prefix to prepend to filepaths in links.
         'use_jupyter_lab': False
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -69,15 +69,15 @@ sphinx_gallery_conf = {
         '../examples/developers',
     ]),
     'binder': {
-     # Required keys
-     'org': 'scikit-image',
-     'repo': 'docs',
-     'branch': 'gh-pages',  # Can be any branch, tag, or commit hash. Use a branch that hosts your docs.
-     'binderhub_url': 'https://mybinder.org',  # Any URL of a binderhub deployment. Must be full URL (e.g. https://mybinder.org).
-     'dependencies': '../requirements.txt',
-     # Optional keys
-     'filepath_prefix': 'dev/', # A prefix to prepend to any filepaths in Binder links.
-     'use_jupyter_lab': False  # Whether Binder links should start Jupyter Lab instead of the Jupyter Notebook interface.
+        # Required keys
+        'org': 'scikit-image',
+        'repo': 'docs',
+        'branch': 'gh-pages',  # Can be any branch, tag, or commit hash
+        'binderhub_url': 'https://mybinder.org',  # Any URL of a binderhub.
+        'dependencies': '../requirements.txt',
+        # Optional keys
+        'filepath_prefix': 'dev/', # A prefix to prepend to filepaths in links.
+        'use_jupyter_lab': False
      }
 }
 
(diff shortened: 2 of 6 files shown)


jni:
@emmanuelle is this really it??? That's awesome!

I'm pretty sure you can write `git+https://github.com/scikit-image/scikit-image.git` as a line in the requirements file and it works.

Having said that, the binder links should point to the right version from the right version of the docs, at least going forward, so probably that line needs to be automatically generated when the docs are generated?

Other than that I'm +:100: on this!

emmanuelle:
Yeah, sphinx-gallery really makes it easy :-). I don't see how to do what you suggest (using the version of scikit-image corresponding to the version of the docs), because binder will launch a single container for the whole `docs` repository. One possible way would be to add a cell at the beginning of the notebooks which would pip install the right version of scikit-image, but this takes some time so I would not favour it. So I think we have to stick with one version only, we'll have to decide whether it's master or the latest release (hopefully frequent releases will make the difference not very important ;-)).

stefanv:
I spent some time looking through the Netlify docs, and I think they can help us here.

1. We need to split out each version of the docs into a separate branch for MyBinder to work correctly.  Each branch gets its own `.binder/requirements.txt`.
2. Netlify is used to deploy each docs branch, and to put a redirect in place from our main website (they can handle redirects, which GitHub doesn't).

For anyone who wants to investigate, here are the relevant Netlify docs links:

- https://docs.netlify.com/site-deploys/overview/#branch-deploy-controls
- https://docs.netlify.com/configure-builds/file-based-configuration/#deploy-contexts
- https://docs.netlify.com/configure-builds/file-based-configuration/#sample-file

emmanuelle:
Thank you @stefanv for investigating!

emmanuelle:
However, is the addition of binder links really worth changing the way we deploy the docs? If we were using master or the last release, it would cover the needs of most users. We don't have statistics for the website any more (something else I need to work on) but I would expect that the traffic of pages other than stable or dev is very very low...

stefanv:
I think it complicates things too much, so I agree—not worth it here.

emmanuelle:
I put the installation of scikit-image master in a postBuild script (binder looks for such scripts) since I could not make it work only with a `requirements.txt`: scikit-image needs numpy to build so the installation was failing (I'm not aware of a way to tell pip: install these packages first, and then these ones etc.)

stefanv:
`pip` doesn't respect the order of packages provided in `requirements.txt`?  Otherwise, simply mentioning numpy first should be enough.

stefanv:
Oh, wait, build: why are we building? We should probably just use the daily wheel.

stefanv:
Yeah, the more I think about it the more I think we should only add this for the latest version released to pypi.

emmanuelle:
Ha ha ha I didn't even know we had daily wheels! How are they built (I mean which scripts should I read to understand how this is done)?

doc/binder/requirements.txt
```diff
@@ -0,0 +1,6 @@
+# For binder
```
soupault: Should we link to one or multiple files from https://github.com/scikit-image/scikit-image/tree/master/requirements to reduce the maintenance demand here? Similarly to https://github.com/scikit-image/scikit-image/blob/master/requirements.txt

emmanuelle: It's a good idea I'll move the file to the `requirements` directory (I think we need a separate file though because we don't need all the doc-building dependencies). Thanks!

emmanuelle:
@stefanv could you please tell me more about the daily wheel? (where does it live and how to pip install from it?)

stefanv:
In retrospect, I don't think we can point to those nightly wheels.  Best to just add this feature for the last stable release.

Note that all binder files can also go into a `.binder` directory.

> Configuration files may be placed in the root of your repository or in a binder/ folder in the root of your repository (i.e. myproject/binder/). If a binder/ folder is used, Binder will only read configuration files from that location (i.e. myproject/binder/requirements.txt) and will ignore those in the repository’s root (myproject/environment.yml and myproject/requirements.txt)

emmanuelle:
So can we merge this one :-) ?

doc/source/conf.py
```diff
@@ -68,6 +68,17 @@
         '../examples/applications',
         '../examples/developers',
     ]),
+    'binder': {
+        # Required keys
+        'org': 'scikit-image',
+        'repo': 'docs',
+        'branch': 'gh-pages',  # Can be any branch, tag, or commit hash
+        'binderhub_url': 'https://mybinder.org',  # Any URL of a binderhub.
+        'dependencies': '../../requirements/binder/requirements.txt',
```
jni: why not `requirements/binder.txt`?

doc/source/conf.py
```diff
@@ -68,6 +68,17 @@
...
+        # Optional keys
+        'filepath_prefix': 'dev/', # A prefix to prepend to filepaths in links.
```
jni: I thought we were going for `stable`?

jni: (though if you managed to make dev work, all the better!)

doc/source/conf.py
emmanuelle: because this file will be copied directly by sphinx-gallery without renaming it and binder will need it to be called `requirements.txt`.

doc/source/conf.py
emmanuelle: because for me `dev` is the only part that new commits can modify at the moment: I consider `stable` and other `0.x` directories to be frozen as they have been generated at release time, so I would not like to modify them?

requirements/binder/requirements.txt
```diff
@@ -0,0 +1,6 @@
+# For binder
```
sciunto:
```suggestion
# For sphinx-gallery.binder
```

tools/travis/deploy_docs.sh
```diff
@@ -19,7 +19,10 @@ then
 
     git rm -r dev
     cp -r ../doc/build/html dev
+    mv dev/binder .
+    cp ../binder/postBuild binder/
```
sciunto: Can we change the paths with a name containing sphinx-gallery. I'm afraid that soon, i'll wonder what's binder.

sciunto: My remark also applies to "requirements/binder"

emmanuelle:
Another idea could be to use thebelab https://thebelab.readthedocs.io/en/latest/

emmanuelle:
TODO
- create new repo to host generated notebooks
- in this repo, have one branch per version of the docs

sciunto:
Superseeded, so I'm closing this one.

PR #4342, opened 2019-12-06
Refactor canny edge detector

### LINKED ISSUE #4282: use_quantiles description in canny is incorrect
## Description
The description of `use_quantiles` is "If True then treat low_threshold and high_threshold as quantiles of the edge magnitude image, rather than absolute edge magnitude values. **If True then the thresholds must be in the range [0, 1].**"

However, he thresholds must be in the range [0, maximum of the image's dtype ]

## Way to reproduce
```python
from skimage import data
from skimage.feature import canny
image = data.camera()
edge_canny = canny(image, low_threshold=50, high_threshold=150, use_quantiles=True)
# No error is produced
```

## Cause
This happens because the thresholds are divided by maximum of the image's dtype (lines 163, 168 in [_canny.py](https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/_canny.py))

## Probable solution
The probable solution is not to divide the thresholds if `use_quantiles` is True or just simpler change the description.

## Version information
```python
# your output here
3.7.3 (default, Sep  5 2019, 17:14:41) 
[Clang 11.0.0 (clang-1100.0.33.8)]
Darwin-19.0.0-x86_64-i386-64bit
scikit-image version: 0.16.2
numpy version: 1.17.3
```

rfezzani:
## Description

In this PR, the `skimage.feature._canny` module is refactored:

- ~~`low_threshold` and `high_threshold` default values are now respectively `0.1` and `0.2` instead of `None`.~~
- ~~setting `low_threshold` and `high_threshold` to `None` is deprecated~~
- ~~the `preserve_range` argument is added to manage the issue #4282: `low_threshold` and `high_threshold` are only divided by `dtype_max` if `preserve_range` is False.~~
- a `ValueError` is raised if `low_threshold > high_threshold`
- `smooth_with_function_and_mask` is deprecated
- intermediary functions `_preprocess`, `_get_local_maxima` and `_set_local_maxima` are defined.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/feature/_canny.py
```diff
@@ -155,126 +330,52 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
...
+        if high_threshold < low_threshold:
+            raise ValueError(
+                "low_threshold should be lower then high_threshold")
+        if high_threshold > 1.0 or low_threshold > 1.0:
+            raise ValueError("Quantile thresholds must not be > 1.0")
+        if high_threshold < 0.0 or low_threshold < 0.0:
+            raise ValueError("Quantile thresholds must not be < 0.0")
 
-    def fsmooth(x):
-        return img_as_float(gaussian(x, sigma, mode='constant'))
```
rfezzani: In the previous version, `preserve_range` was implicitly set to `False` as the default argument of `gaussian`

skimage/feature/_canny.py
```diff
@@ -155,126 +330,52 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
...
-    #--------- Find local maxima --------------
-    #
-    # Assign each point to have a normal of 0-45 degrees, 45-90 degrees,
-    # 90-135 degrees and 135-180 degrees.
-    #
-    local_maxima = np.zeros(image.shape, bool)
-    #----- 0 to 45 degrees ------
-    pts_plus = (isobel >= 0) & (jsobel >= 0) & (abs_isobel >= abs_jsobel)
-    pts_minus = (isobel <= 0) & (jsobel <= 0) & (abs_isobel >= abs_jsobel)
-    pts = pts_plus | pts_minus
```
rfezzani: These boolean expressions are simplified in `_get_local_maxima`

skimage/feature/_canny.py
```diff
@@ -155,126 +330,52 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
...
-    # top right of pts.
-    c1 = magnitude[1:, :][pts[:-1, :]]
-    c2 = magnitude[1:, 1:][pts[:-1, :-1]]
-    m = magnitude[pts]
-    w = abs_jsobel[pts] / abs_isobel[pts]
-    c_plus = c2 * w + c1 * (1 - w) <= m
-    c1 = magnitude[:-1, :][pts[1:, :]]
-    c2 = magnitude[:-1, :-1][pts[1:, 1:]]
-    c_minus = c2 * w + c1 * (1 - w) <= m
-    local_maxima[pts] = c_plus & c_minus
```
rfezzani: This interpolation step is performed in `_set_local_maxima`

skimage/feature/_canny.py
```diff
@@ -14,44 +14,175 @@
...
+    sigma : scalar or sequence of scalars, optional
+        Standard deviation for Gaussian kernel. The standard
+        deviations of the Gaussian filter are given for each axis as a
+        sequence, or as a single number, in which case it is equal for
+        all axes.
+    mode : {'reflect', 'constant', 'nearest', 'mirror', 'wrap'}, optional
+        The ``mode`` parameter determines how the array borders are
+        handled, where ``cval`` is the value when mode is equal to
+        'constant'. Default is 'nearest'.
+    preserve_range : bool, optional
```
sciunto, line 44: This option does not exist any more.

skimage/feature/_canny.py
```diff
@@ -14,44 +14,175 @@
...
-        A function that does image smoothing.
+        Image to be smoothed.
     mask : array
         Mask with 1's for significant pixels, 0's for masked pixels.
+    sigma : scalar or sequence of scalars, optional
+        Standard deviation for Gaussian kernel. The standard
+        deviations of the Gaussian filter are given for each axis as a
+        sequence, or as a single number, in which case it is equal for
+        all axes.
+    mode : {'reflect', 'constant', 'nearest', 'mirror', 'wrap'}, optional
```
sciunto, line 40: Looks like this function is used only below, with mode='constant'. Is it useful to keep this option? I think nobody will touch it.
Otherwise, optional must be removed.

skimage/feature/_canny.py
```diff
@@ -14,44 +14,175 @@
...
     Parameters
     ----------
     image : array
-        Image you want to smooth.
-    function : callable
-        A function that does image smoothing.
+        Image to be smoothed.
     mask : array
         Mask with 1's for significant pixels, 0's for masked pixels.
+    sigma : scalar or sequence of scalars, optional
```
sciunto, line 35:
```suggestion
    sigma : scalar or sequence of scalars
```

skimage/feature/_canny.py
```diff
@@ -14,44 +14,175 @@
...
 from ..filters import gaussian
-from .. import dtype_limits, img_as_float
+from .. import dtype_limits
 from .._shared.utils import check_nD
 
 
-def smooth_with_function_and_mask(image, function, mask):
-    """Smooth an image with a linear function, ignoring masked pixels.
+def _preprocess(image, mask, sigma, mode):
+    """Preprocess the image and mask before applying canny edge detection.
```
sciunto, line 24:
```suggestion
    """Generate a smoothed image and an eroded mask.
```

skimage/feature/_canny.py
```diff
@@ -14,44 +14,175 @@
...
+    # out the image edges for us.
+    s = ndi.generate_binary_structure(2, 2)
+    eroded_mask = ndi.binary_erosion(mask, s, border_value=0)
+
+    return smoothed_image, eroded_mask
+
+
+def _set_local_maxima(magnitude, pts, w_num, w_denum, row_slices,
+                      col_slices, out):
+    # Get the magnitudes shifted left to make a matrix of the points to the
```
sciunto, line 88: Would it be possible to turn this in a docstring please?

skimage/feature/_canny.py
```diff
@@ -14,44 +14,175 @@
...
+    c2 = magnitude[r_3, c_3][pts[r_2, c_2]]
+    c_minus = c2 * w + c1 * (1 - w) <= m
+    out[pts] = c_plus & c_minus
+
+    return out
+
+
+def _get_local_maxima(isobel, jsobel, magnitude, eroded_mask):
+    #
+    # Find the normal to the edge at each point using the arctangent of the
```
sciunto, line 108: This comment seems to apply to the first block of code only, which confused me. I think a doctring would help also.

skimage/feature/_canny.py
rfezzani: You are right, it is actually not used, but it may be should! Please see my [comment](https://github.com/scikit-image/scikit-image/pull/4342/files#r354774473).

rfezzani:
Thank you @sciunto for your review and comments ;)

skimage/feature/_canny.py
rfezzani: Please don't take my last comment into account, I will fix that ;) I for sure need one coffee more this morning =p

skimage/feature/_canny.py
rfezzani: In the previous implementation, `mode` is set to 'constant', but I think that the default recommended value 'nearest' used in `gaussian` must be used here also.
But this may be addressed in another PR ;)

sciunto: :coffee:

sciunto: +1 to check that in another PR.

skimage/feature/_canny.py
```diff
@@ -69,9 +201,10 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
     mask : array, dtype=bool, optional
         Mask to limit the application of Canny to a certain area.
     use_quantiles : bool, optional
-        If True then treat low_threshold and high_threshold as quantiles of the
-        edge magnitude image, rather than absolute edge magnitude values. If True
-        then the thresholds must be in the range [0, 1].
+        If True then treat low_threshold and high_threshold as
+        quantiles of the edge magnitude image, rather than absolute
+        edge magnitude values. If True then the thresholds ar also
```
sciunto, line 206: remove "ar also"?

skimage/feature/_canny.py
```diff
@@ -161,137 +296,51 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
         low_threshold = 0.1
     elif use_quantiles:
         if not(0.0 <= low_threshold <= 1.0):
-            raise ValueError("Quantile thresholds must be between 0 and 1.")
+            raise ValueError("quantile thresholds must be between 0 and 1.")
```
sciunto, line 299: Just my curiosity, why this change?

skimage/feature/_canny.py
```diff
@@ -14,44 +14,176 @@
...
-    and apply the function. The resulting values will be lower by the
-    bleed-over fraction, so you can recalibrate by dividing by the function
-    on the mask to recover the effect of smoothing from just the significant
-    pixels.
     """
-    bleed_over = function(mask.astype(float))
-    masked_image = np.zeros(image.shape, image.dtype)
+
+    if mask is None:
+        mask = np.ones(image.shape, dtype=bool)
```
sciunto, line 55: I think I prefer to see these two lines in the main function (that defines a default behavior)

skimage/feature/_canny.py
```diff
@@ -161,137 +296,51 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
...
 
-    if mask is None:
-        mask = np.ones(image.shape, dtype=bool)
+    if high_threshold < low_threshold:
+        raise ValueError("low_threshold should be lower then high_threshold")
 
-    def fsmooth(x):
-        return img_as_float(gaussian(x, sigma, mode='constant'))
+    # Image filtering
+    smoothed, eroded_mask = _preprocess(image, mask, sigma, 'constant')
```
sciunto, line 315: Is there a plan to put the mode in the API? If not, why this is part of the _preprocess function API if this mode is expected for a canny filter?

skimage/feature/_canny.py
```diff
@@ -69,9 +201,10 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
     mask : array, dtype=bool, optional
         Mask to limit the application of Canny to a certain area.
     use_quantiles : bool, optional
-        If True then treat low_threshold and high_threshold as quantiles of the
-        edge magnitude image, rather than absolute edge magnitude values. If True
-        then the thresholds must be in the range [0, 1].
+        If True then treat low_threshold and high_threshold as
```
sciunto, line 204: Minor: it is normally better to put backticks around True False and variables, but I never remember if simple or double backticks must be used.

skimage/feature/_canny.py
```diff
@@ -14,44 +14,176 @@
...
         Mask with 1's for significant pixels, 0's for masked pixels.
+    sigma : scalar or sequence of scalars
+        Standard deviation for Gaussian kernel. The standard
+        deviations of the Gaussian filter are given for each axis as a
+        sequence, or as a single number, in which case it is equal for
+        all axes.
+    mode : str, {'reflect', 'constant', 'nearest', 'mirror', 'wrap'}
+        The ``mode`` parameter determines how the array borders are
+        handled, where ``cval`` is the value when mode is equal to
+        'constant'. Default is 'nearest'.
```
sciunto, line 43:
```suggestion
        'constant'.
```

sciunto (commented):
I made few additional comments @rfezzani although it is a nice refactoring.

skimage/feature/_canny.py
rfezzani: I can't remember why I did this :confused: Will revert :pray:

skimage/feature/_canny.py
rfezzani: I would love to see it in the API since it was implicit in the previous implementation (see https://github.com/scikit-image/scikit-image/pull/4342#discussion_r354774473)

skimage/feature/_canny.py
rfezzani: We don't really need `mask` in the main function! Defining it there will keep it in memory until the end of the function call :confused:

skimage/feature/_canny.py
rfezzani: @sciunto, should we expose `mode` in `canny` API?

grlee77:
Hi @rfezzani, I added the benchmark commit from #5162 here. Let me know if you would prefer I open a pull request vs. your repo instead of pushing here directly.

rfezzani:
Excellent! No worry @grlee77, thank you :wink:

grlee77:
I ran the benchmark on bbc4460, which is just this PR immediately after addition of the benchmark itself. The last two performance-related commits were ported from #5162.  Unlike in the other PR, I ran these benchmarks with `asv run` mode instead of `asv dev` mode, so it should be a bit more accurate.

time_canny (ms) | commit
-----------------------|-------
1040   | v0.18.1
994     | bbc4460
770     | 4c60efe
752     | 2aea7db 

Rather than hard-coding `mode` internally for the unmasked case, I added a keyword-only argument for it to the API as @rfezzani had mentioned wanting to do above. There is one test case (`test_use_quantiles`) that needs this to be specified as mode='nearest' ('reflect' and 'mirror' also pass, but mode='constant' does not).

grlee77 (approved):
The style changes here look good to me.

@rfezzani, please review the three commits I had added. For the benchmark, I think there was an error in setting up the test image previously because it contained only noise! I switched to using the `astronaut` image as a more realistic case.

I can confirm that the changes here give the same result as for the image comparison in the first commit of #5162. It may be worth changing the default mode='constant' to mode='nearest' or mode='reflect' to keep results the most similar to before in the unmasked case.

skimage/feature/_canny.py
```diff
@@ -205,6 +212,10 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None,
         quantiles of the edge magnitude image, rather than absolute
         edge magnitude values. If ``True`` then the thresholds must be
         in the range [0, 1].
+    mode : str, {'reflect', 'constant', 'nearest', 'mirror', 'wrap'}
+        The ``mode`` parameter determines how the array borders are
+        handled during Gaussian filtering, where ``cval`` is the value when
```
rfezzani, line 217: Should we add the `cval` argument also? It must then be passed to the `_preprocess` function.

rfezzani:
@sciunto, do you still see requested changes? Thank you for your time ;)

skimage/feature/_canny.py
grlee77: I went ahead and added it

grlee77:
@sciunto: I think this is ready, but it needs you to mark the requested changes as resolved to proceed. One outstanding issue to be left to a separate PR is changing the default mode to 'nearest'.

PR #4343, opened 2019-12-06
Fix Frangi artefacts around the image

### LINKED ISSUE #4249: Frangi artefacts around the image
[img_raw.zip](https://github.com/scikit-image/scikit-image/files/3738350/img_raw.zip)

## Description
frangi problem

Hi,
I upgraded Scikit from 0.14.2 to 0.16.1 and now I experience some problems. As illustrated in the picture, with the new Frangi I get a a bright square all across the image which I did not get using the old version (and also not present in the raw image)
I used Frangi in 2D and I still use the beta1 and beta2 parameters (instead of beta and gamma). Switching over to the new parameters does not solve it

                              'scale_range'= [5,6]
                              'scale_step' = 1
                              'beta1'=1         
                              'beta2'=0.0005
                              'black_ridges':True
I don't pass on the alpha parameter (because I work in 2D), but I do notice that this alpha is still used in calculations (maybe ok, I don't know)

I will have to revert to 14.01 for now, but if this could be fixed that would be great

thanks !
Wim

## Way to reproduce
from tifffile import TiffFile,imsave
from skimage.filters import frangi
from skimage import img_as_uint

with TiffFile("img_raw.tif") as tif:
    img_raw= tif.asarray()
frangi_img = frangi(img_raw,scale_range=[5,6],
                    scale_step=1,
                    beta1 = 1, 
                    beta2 = 0.0005,
                    black_ridges=True)

imsave("frangi.tif", img_as_uint(frangi_img))

## Version information
```python 3.7.3 

Running the previous version scikit : 
Linux-4.4.0-18362-Microsoft-x86_64-with-debian-buster-sid
scikit-image version: 0.14.2
numpy version: 1.16.2

running  the new version scikit
Linux-4.4.0-18362-Microsoft-x86_64-with-debian-buster-sid
scikit-image version: 0.16.1
numpy version: 1.16.2
frangi_issue_recreate.py:18: UserWarning: Use keyword parameter `beta` instead of `beta1` which will be removed in version 0.17.
  black_ridges=True)
frangi_issue_recreate.py:18: UserWarning: Use keyword parameter `gamma` instead of `beta2` which will be removed in version 0.17.
  black_ridges=True)
frangi_issue_recreate.py:18: UserWarning: Use keyword parameter `sigmas` instead of `scale_range` and `scale_range` which will be removed in version 0.17.
  black_ridges=True)

### PR DESCRIPTION
**Pull request by rfezzani**
## Description

Closes #4249.

This PR fixes this issue by adding adding border conditions to `compute_hessian_eigenvalues` and setting it to `'reflect'` in `frangi`. Here is the output of @wimthiels's example:

frangi

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/edges/plot_ridge_filter.py
+++ b/doc/examples/edges/plot_ridge_filter.py
@@ -65,7 +65,7 @@ for i, black_ridges in enumerate([1, 0]):
     for j, func in enumerate([identity, meijering, sato, frangi, hessian]):
         kwargs['black_ridges'] = black_ridges
         result = func(image, **kwargs)
-        if func in (meijering, frangi):
+        if func in (meijering):
             # Crop by 4 pixels for rendering purpose.
             result = result[4:-4, 4:-4]
         axes[i, j].imshow(result, cmap=cmap, aspect='auto')
--- a/doc/examples/edges/plot_ridge_filter.py
+++ b/doc/examples/edges/plot_ridge_filter.py
@@ -65,7 +65,7 @@ for i, black_ridges in enumerate([1, 0]):
     for j, func in enumerate([identity, meijering, sato, frangi, hessian]):
         kwargs['black_ridges'] = black_ridges
         result = func(image, **kwargs)
-        if func in (meijering):
+        if func is meijering:
             # Crop by 4 pixels for rendering purpose.
             result = result[4:-4, 4:-4]
         axes[i, j].imshow(result, cmap=cmap, aspect='auto')
--- a/skimage/filters/ridges.py
+++ b/skimage/filters/ridges.py
@@ -428,6 +428,10 @@ def frangi(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
                                                         sorting='abs',
                                                         mode='reflect')
 
+        # Compute sensitivity to deviation from a plate-like
+        # structure see equations (11) and (15) in reference [1]_
+        r_a = np.inf if ndim == 2 else _divide_nonzero(*lambdas) ** 2
+
         # Compute sensitivity to deviation from a blob-like structure,
         # see equations (10) and (15) in reference [1]_,
         # np.abs(lambda2) in 2D, np.sqrt(np.abs(lambda2 * lambda3)) in 3D
@@ -440,13 +444,9 @@ def frangi(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
 
         # Compute output image for given (sigma) scale and store results in
         # (n+1)D matrices, see equations (13) and (15) in reference [1]_
-        filtered_array[i] = (np.exp(-r_b / beta_sq)
+        filtered_array[i] = ((1 - np.exp(-r_a / alpha_sq))
+                             * np.exp(-r_b / beta_sq)
                              * (1 - np.exp(-r_g / gamma_sq)))
-        if ndim == 3:
-            # Compute sensitivity to deviation from a plate-like
-            # structure see equations (11) and (15) in reference [1]_
-            r_a = _divide_nonzero(*lambdas) ** 2
-            filtered_array[i] *= (1 - np.exp(-r_a / alpha_sq))
 
         lambdas_array[i] = np.max(lambdas, axis=0)
 
--- a/skimage/filters/ridges.py
+++ b/skimage/filters/ridges.py
@@ -15,6 +15,7 @@ import numpy as np
 
 from ..util import img_as_float, invert
 from .._shared.utils import check_nD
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by emmanuelle** [comment]
Thank you @rfezzani ! Could you please add a test checking the correct behaviour on the boundary? It would be a test that would fail on master and pass with your fix.

File: skimage/filters/ridges.py [hunk 1]
```diff
@@ -438,9 +440,14 @@ def frangi(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
 
         # Compute output image for given (sigma) scale and store results in
         # (n+1)D matrices, see equations (13) and (15) in reference [1]_
-        filtered_array[i] = ((1 - np.exp(-r_a / alpha_sq))
-                             * np.exp(-r_b / beta_sq)
+        filtered_array[i] = (np.exp(-r_b / beta_sq)
                              * (1 - np.exp(-r_g / gamma_sq)))
+        if ndim == 3:
```
emmanuelle: Why did you add this part? Was the behaviour not correct for 3D images? If this is the case, could you please also add a test here?

rfezzani: The 3D image case is managed [here](https://github.com/scikit-image/scikit-image/pull/4343/files#diff-9a205ce8d1dfde393db115dfeb32740bL427) in the previous implementation. It is working perfectly well, I just rewrite things to be sure that I correctly understood the behavior :slightly_smiling_face: I will revert it, sorry :pray:

skimage/filters/ridges.py
```diff
@@ -105,20 +112,18 @@ def compute_hessian_eigenvalues(image, sigma, sorting='none'):
...
-    # Import has to be here due to circular import error
-    from ..feature import hessian_matrix, hessian_matrix_eigvals
-
     # Convert image to float
     image = img_as_float(image)
 
     # Make nD hessian
-    hessian_elements = hessian_matrix(image, sigma=sigma, order='rc')
+    hessian_elements = hessian_matrix(image, sigma=sigma, order='rc',
+                                      mode=mode)
```
shenker: Shouldn't you pass cval in here too, in case `mode='constant'`?

skimage/filters/ridges.py
```diff
@@ -413,18 +418,15 @@ def frangi(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
...
 
         # Calculate (abs sorted) eigenvalues
         lambda1, *lambdas = compute_hessian_eigenvalues(image, sigma,
-                                                        sorting='abs')
-
-        # Compute sensitivity to deviation from a plate-like structure
-        # see equations (11) and (15) in reference [1]_
-        r_a = np.inf if ndim == 2 else _divide_nonzero(*lambdas) ** 2
+                                                        sorting='abs',
+                                                        mode='reflect')
```
shenker: Is there a reason mode is hardcoded to be `'reflect'`? I'd vote to have `mode` and `cval` be kwargs in `frangi` and pass them both through to `compute_hessian_eigenvalues`.

skimage/filters/ridges.py
rfezzani: Fair enough! ;-) Sorry for the delay...

skimage/filters/ridges.py
rfezzani: I agree, good catch ;-)

rfezzani:
Thank you @shenker for you review! I like your suggestion of adding border management to ridge filters, we have to come with new PR for this ;-)

jni:
@rfezzani Great PR! imho this is ready except for the need for a test! 😃

rfezzani:
Thx @jni, the test is coming soon ;-)

rfezzani:
Here comes the test! I hope you like it ;-)

doc/examples/edges/plot_ridge_filter.py
```diff
@@ -65,7 +65,7 @@ def identity(image, **kwargs):
     for j, func in enumerate([identity, meijering, sato, frangi, hessian]):
         kwargs['black_ridges'] = black_ridges
         result = func(image, **kwargs)
-        if func in (meijering, frangi):
+        if func in (meijering):
```
emmanuelle: you need to put the comma back `(meijering,)` otherwise it's not iterable. But do we need a loop if there is only one element?

emmanuelle: see the failing doc build in the CI

emmanuelle:
For me this is good to go. @jni you said the same but did not formally approve, can I let you merge this one?

PR #4344, opened 2019-12-07
added matomo code

emmanuelle:
This PR is to track again the analytics of the website using matomo (used to be piwik). This is the cloud version for the 21-day trial that I activated today, which will hopefully let us the time to set up an on-premise server or see if we're willing to pay. But I wanted to get the ball rolling, in particular for the CZI project it would be good to know the analytics at the beginning of the grant.

### DIFF
--- a/doc/source/themes/scikit-image/layout.html
+++ b/doc/source/themes/scikit-image/layout.html
@@ -109,18 +109,20 @@
 </html>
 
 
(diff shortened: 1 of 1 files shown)


emmanuelle:
oops I pushed the branch to the scikit-image repo, I will delete the branch and open a new PR originating from my fork instead.

PR #4345, opened 2019-12-07
added matomo code

emmanuelle:
This PR is to track again the analytics of the website using matomo (used to be piwik). This is the cloud version for the 21-day trial that I activated today, which will hopefully let us the time to set up an on-premise server or see if we're willing to pay. But I wanted to get the ball rolling, in particular for the CZI project it would be good to know the analytics at the beginning of the grant.

### DIFF
--- a/doc/source/themes/scikit-image/layout.html
+++ b/doc/source/themes/scikit-image/layout.html
@@ -109,18 +109,20 @@
 </html>
 
 
-<!-- Piwik -->
+<!-- Matomo -->
 <script type="text/javascript">
-  var _paq = _paq || [];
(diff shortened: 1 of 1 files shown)


emmanuelle:
Are you proud of us @sciunto :p ? Every time that with @jni we're on the brink of surrendering to the dark side and set up GA, we think we'd rather not disappoint you ;-).

emmanuelle:
Also, matomo emailed me they've reset our 21-day trial to start when they received the first analytics (just after the PR was merged). Quite elegant, I like that!

sciunto:
Definitely @emmanuelle  :)

PR #4348, opened 2019-12-09
Include example use of function regionprops_table in the docs

### LINKED ISSUE #4335: regionsprops_table is not mentioned in the doc
We should use it in an existing example, or add a new example.

mkcor:
## Description

Closes #4335.

First, I thought about adding another region. It's only a toy example, but it feels too odd to write a for loop to iterate on a length-1 list.

Then, I appended a call to `regionprops_table` and displayed the result.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


doc/examples/segmentation/plot_regionprops.py
```diff
@@ -3,15 +3,23 @@
 Measure region properties
 =========================
 
-This example shows how to measure properties of labelled image regions.
+This example shows how to measure properties of labelled image regions. It is
+a toy example, where the image regions are two ellipses. The outputs are the
+following:
```
soupault: I believe the outputs should rather be described near the function calls. It is not neccessary to add `:py:func:`regionprops``, etc, since the hyperlinks will be automatically added to the corresponding code.

doc/examples/segmentation/plot_regionprops.py
```diff
@@ -3,15 +3,23 @@
 Measure region properties
 =========================
 
-This example shows how to measure properties of labelled image regions.
+This example shows how to measure properties of labelled image regions. It is
+a toy example, where the image regions are two ellipses. The outputs are the
```
soupault: An example is an example :). I would modify like this:
`It is a toy example, ...` -> `We analyze an image with two ellipses.`

doc/examples/segmentation/plot_regionprops.py
```diff
@@ -22,6 +30,9 @@
 
 image = rotate(image, angle=15, order=0)
 
+rrr, ccc = ellipse(100, 100, 60, 50)
```
soupault:
```suggestion
rr, cc = ellipse(100, 100, 60, 50)
```

doc/examples/segmentation/plot_regionprops.py
```diff
@@ -47,3 +58,12 @@
 
 ax.axis((0, 600, 600, 0))
 plt.show()
+
+props = regionprops_table(label_img, properties=('centroid',
```
soupault: We could perhaps say here that `regionprops_table` actually computes the properties, while `regionprops` only does lazy execution.

requirements/docs.txt
```diff
@@ -10,3 +10,4 @@ matplotlib>=3.0.1
 dask[array]>=0.15.0
 # cloudpickle is necessary to provide the 'processes' scheduler for dask
 cloudpickle>=0.2.1
+pandas
```
soupault: Some reasonable lower bound for the version would be nice. As a reference, I suggest to use NEP29 (see https://github.com/scikit-image/scikit-image/issues/4371).

doc/examples/segmentation/plot_regionprops.py
mkcor: > I believe the outputs should rather be described near the function calls.

Ok.

> It is not neccessary to add :py:func:regionprops``, etc, since the hyperlinks will be automatically added to the corresponding code.

Really? We can see that
https://github.com/scikit-image/scikit-image/blob/d7694bacb43f5246662ead4b1a7a3bd3d7a196f5/doc/examples/features_detection/plot_template.py#L13
does not result in any hyperlink:
https://scikit-image.org/docs/stable/auto_examples/features_detection/plot_template.html
whereas
https://github.com/scikit-image/scikit-image/blob/d7694bacb43f5246662ead4b1a7a3bd3d7a196f5/doc/examples/segmentation/plot_compact_watershed.py#L14
does:
https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_compact_watershed.html
So I should actually write py:func:`skimage.measure.regionprops` and :py:func:`skimage.measure.regionprops_table`...

lagru: I'm guessing that @soupault is talking about hyperlinks in the code blocks. You don't need to explicitly reference :py:func:regionprops in the text because it get's automatically linked by sphinx-gallery in the code sections while it is used.

In my opinion it doesn't hurt to have extra hyperlinks here but @soupault 's suggestion to move the descriptions closer to the outputs is a good one.

doc/examples/segmentation/plot_regionprops.py
mkcor: Agreed :)
Included in fd4bcfa -- I manually created an "Apply suggestions from code review" commit with you @soupault as a co-author, hope that's all right!

doc/examples/segmentation/plot_regionprops.py
mkcor: Ok. Need to change the next line as well (hence committing manually 3094978).

doc/examples/segmentation/plot_regionprops.py
mkcor: I see. I moved the descriptions next to the function calls (fd4bcfa22bbd262b8dee867064a82f4044e88c8b), which results in a high density of hyperlinks...

doc/examples/segmentation/plot_regionprops.py
```diff
@@ -22,9 +24,17 @@
 
 image = rotate(image, angle=15, order=0)
 
+rr, cc = ellipse(100, 100, 60, 50)
+image[rr, cc] = 1
```
soupault: I have just realized that there is an issue with coordinates in ploting - https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_regionprops.html. See how the axes are transposed with respect to the ellipse.

doc/examples/segmentation/plot_regionprops.py
soupault: Thanks @lagru , that was exactly my thinking. I am fine with the current version anyway 👍

mkcor: Great. Thank you both :)

doc/examples/segmentation/plot_regionprops.py
mkcor: Why not; done in 38ddd671d81db0eb69f28351094ed4ea995511dc.

doc/examples/segmentation/plot_regionprops.py
mkcor: Good catch! Fixing the major vs minor axes now.

mkcor: Done in cdf1678503c5dd55172fa322d3458727eb0758a6. Thanks again :)

PR #4350, opened 2019-12-09
Make hough_circle_peaks respect min_xdistance, min_ydistance

xiaoyu-wu:
## Description

Closes issue #2728 
For circles with different radius but close in distance, only keeping the one with highest peak intensity.

Using the [example code](https://gist.github.com/xiaoyu-wu/7116c5d133f01718e601365532fe03a6) in the issue, now only one circle is drawn per coin.
hough_circle_peaks

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/hough_transform.py
+++ b/skimage/transform/hough_transform.py
@@ -362,6 +362,10 @@ def hough_circle_peaks(hspaces, radii, min_xdistance=1, min_ydistance=1,
             cx.append(x)
             cy.append(y)
             r.append(ra)
+    accum = np.array(accum)
+    cx = np.array(cx)
+    cy = np.array(cy)
+    r = np.array(r)
 
     if total_num_peaks != np.inf:
         tnp = total_num_peaks
--- a/skimage/transform/hough_transform.py
+++ b/skimage/transform/hough_transform.py
@@ -358,7 +358,8 @@ def hough_circle_peaks(hspaces, radii, min_xdistance=1, min_ydistance=1,
 
     # Skip searching for neighboring circles
     # if default min_xdistance and min_ydistance are used
-    if min_xdistance == 1 and min_ydistance == 1:
+    # or if no peak was detected
+    if (min_xdistance == 1 and min_ydistance == 1) or len(accum_sorted) == 0:
         return (accum_sorted[:tnp],
                 cx_sorted[:tnp],
                 cy_sorted[:tnp],
@@ -366,41 +367,56 @@ def hough_circle_peaks(hspaces, radii, min_xdistance=1, min_ydistance=1,
 
     # For circles with centers too close, only keep the one with
     # the highest peak
-    removed = np.zeros(len(accum_sorted), dtype=bool)
-    accum_kept = []
-    cx_kept = []
-    cy_kept = []
-    r_kept = []
+    should_keep = label_distant_points(
+        cx_sorted, cy_sorted, min_xdistance, min_ydistance, tnp
+    )
+    return (accum_sorted[should_keep],
+            cx_sorted[should_keep],
+            cy_sorted[should_keep],
+            r_sorted[should_keep])
+
+
+def label_distant_points(xs, ys, min_xdistance, min_ydistance, max_points):
+    """Keep points that are separated by certain distance in each dimension.
+
+    Parameters
+    ----------
+    xs : array
+        X coordinates of points.
+    ys : array
+        Y coordinates of points.
+    min_xdistance : int
+        Minimum distance separating points in the x dimension.
+    min_ydistance : int
+        Minimum distance separating points in the y dimension.
+    max_points : int
+        Max number of distant points to keep.
+
+    Returns
+    -------
+    should_keep : array of bool
+        A mask array for distant points to keep.
+    """
+    is_neighbor = np.zeros(len(xs), dtype=bool)
+    coordinates = np.stack([xs, ys], axis=1)
+    # Use a KDTree to search for neighboring points effectively
+    kd_tree = cKDTree(coordinates)
     i = 0
-
-    # Use a KDTree to search for neighboring circles effectively
-    scaled_cx_sorted = cx_sorted / float(min_xdistance)
-    scaled_cy_sorted = cy_sorted / float(min_ydistance)
-    scaled_coordinates = np.stack([scaled_cx_sorted, scaled_cy_sorted], axis=1)
-    if scaled_coordinates.size == 0:
-        # Return empty arrays early
-        # as cKDTree throws an error given zero-sized data
-        return (np.array(accum_kept), np.array(cx_kept),
-                np.array(cy_kept), np.array(r_kept))
-    kd_tree = cKDTree(scaled_coordinates)
-    while i < tnp:
-        if not removed[i]:
-            accum_kept.append(accum_sorted[i])
-            cx_kept.append(cx_sorted[i])
-            cy_kept.append(cy_sorted[i])
-            r_kept.append(r_sorted[i])
+    while i < max_points:
+        if not is_neighbor[i]:
             # Find a short list of candidates to remove
-            # by searching within a radius of sqrt(2) in scaled space
+            # by searching within a circle
             neighbors_i = kd_tree.query_ball_point(
-                (scaled_cx_sorted[i], scaled_cy_sorted[i]), np.sqrt(2)
+                (xs[i], ys[i]),
+                np.hypot(min_xdistance, min_ydistance)
             )
-            # Check distance in both dimensions and mark as removed if close
+            # Check distance in both dimensions and mark if close
             for ni in neighbors_i:
-                x_close = abs(scaled_cx_sorted[ni] - scaled_cx_sorted[i]) <= 1
-                y_close = abs(scaled_cy_sorted[ni] - scaled_cy_sorted[i]) <= 1
-                if x_close or y_close:
-                    removed[ni] = True
+                x_close = abs(xs[ni] - xs[i]) <= min_xdistance
+                y_close = abs(ys[ni] - ys[i]) <= min_ydistance
+                if (x_close or y_close) and (ni > i):
+                    is_neighbor[ni] = True
         i += 1
-
-    return (np.array(accum_kept), np.array(cx_kept),
-            np.array(cy_kept), np.array(r_kept))
+    should_keep = ~is_neighbor
+    should_keep[max_points:] = False
+    return should_keep
--- a/skimage/transform/hough_transform.py
+++ b/skimage/transform/hough_transform.py
@@ -343,9 +343,25 @@ def hough_circle_peaks(hspaces, radii, min_xdistance=1, min_ydistance=1,
     else:
         s = np.argsort(accum)
 
+    # For circles with centers too close, only keep the one with
+    # the highest peak
+    accum_sorted, cx_sorted, cy_sorted, r_sorted = \
+        accum[s][::-1], cx[s][::-1], cy[s][::-1], r[s][::-1]
+    accum = []
+    cx = []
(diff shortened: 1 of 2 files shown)


sciunto:
It's a good suggestion.

After reading the code, not tested it, I think there is a flaw. Peak's heights depend on the perimeter. So this algorithm will prefer larger circles. A dimensionless peak height should be better.

xiaoyu-wu:
> It's a good suggestion.
> 
> After reading the code, not tested it, I think there is a flaw. Peak's heights depend on the perimeter. So this algorithm will prefer larger circles. A dimensionless peak height should be better.

Thanks for pointing out. Whether larger circles are preferred is already controlled by the kwarg `normalized`. If `normalized` is set False, then originally the returned tuple are in fact sorted with larger circles first. Enforcing a dimensionless peak height will break `test_hough_circle_peaks_total_peak`. Maybe I can add more to the docstring and be explicit about the preference toggled by `normalized`?

sciunto:
I'm not sure what would be the meaning of normalized=False regarding the removed circles. (I admit that the interpretation is also unclear for the sorting feature, but at least, it is less harmful). Any comment @scikit-image/core ?

skimage/transform/hough_transform.py
```diff
@@ -343,9 +344,31 @@ def hough_circle_peaks(hspaces, radii, min_xdistance=1, min_ydistance=1,
...
 
+    # For circles with centers too close, only keep the one with
+    # the highest peak
+    accum_sorted, cx_sorted, cy_sorted, r_sorted = \
+        accum[s][::-1], cx[s][::-1], cy[s][::-1], r[s][::-1]
+    accum = []
+    cx = []
+    cy = []
+    r = []
+    for ac, x, y, ra in zip(accum_sorted, cx_sorted, cy_sorted, r_sorted):
```
emmanuelle: could we execute this piece of code only if `min/max_xdistance>1`? I guess tt can take quite a long time for a large number of circles with the two loops (I haven't tried). Or could cythonize this part. I wonder how long it takes says for a 2000x2000 image with 1000 random circles?

emmanuelle: Another option could be to use a KDTree here for the x and y coordinates (one might have to add some jitter, I don't know from the top of my head if duplicate coordinates are allowed for KDTree), and use the `query_ball_point` method (https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.spatial.KDTree.query_ball_point.html#scipy.spatial.KDTree.query_ball_point) to keep a complexity of O(N) (with N the number of circles).

emmanuelle:
@sciunto, regarding normalization, I think the code is ok, what is needed is more explanations in the docstring (probably in a Notes section). Indeed, there are two ways of normalizing:
- when computing the hough transform with `hough_circle`. Since the `threshold` of `hough_circle_peaks` is the same for all radii, you will have a different result whether you had normalized the transform or not.
- and then you have the `normalize` argument for `hough_circle_peaks`, which allows you to give more or less importance to large circles when comparing circles with different radii. 

So either you normalize in the hough_transform and then there is no bias between small and large circles (with the caveat that you can detect a lot of small spurious circles), or you always have the option to normalize at the `hough_circle_peaks` stage. 

Would you agree with the above @sciunto ? I might be missing something. In any case this will require a solid paragraph in the docstring. 

And thank you @xiaoyu-wu for taking care of this!

skimage/transform/hough_transform.py
lagru: Just to quickly chime in: in case this data structure is a good fit and you decide to use it I'd recommend [cKDTree](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html) which is the faster version written in C. I've had great experiences (speed ups) with it so far! :)

lagru: I think cKDTree supports duplicate coordinates and it will return duplicates as if queried.

sciunto:
@emmanuelle I agree, I just do not find it super-safe.

skimage/transform/hough_transform.py
xiaoyu-wu: > could we execute this piece of code only if `min/max_xdistance>1`? I guess it can take quite a long time for a large number of circles with the two loops (I haven't tried). Or could cythonize this part. 

it is definitely helpful to skip picking when default values are used. and i agree that the two nested loops can led to O(N^2) complexity (N = number of peaks) and that this part of code should be optimized.

xiaoyu-wu: > Another option could be to use a KDTree here for the x and y coordinates ...

I have tried using a cKDTree as suggested by @lagru and I believe the code now has better performance.
 
One subtile difference is that `query_ball_point` might still keep circles which centers are close in one dimension but far enough than specified distance in 2d plane. For example, two circles A, B with center (0, 0), (10, 10) should be considered close if min distance is 10 in both x and y directions. However, B will not be returned if we search A's neighbors within a distance of 10, i.e. sqrt(2) > 1.

emmanuelle:
@sciunto I agree but I still think that the proposed behaviour is better than the existing one. So let's add explanations and see how it goes...

skimage/transform/hough_transform.py
emmanuelle: Oh you're right. We really made our life more complicated with having two parameters for the distances along `x` and `y`, but ok why not. One could still use the cKDTree with the largest distance times sqrt(2), and then test if the KDE neighbours should be eliminated or not.

jni: > with the largest distance times sqrt(2)

or np.hypot(dx, dy)? =)

xiaoyu-wu: In my latest change, I scaled in both x and y dimension so we don't get an unnecessarily long list of neighbors.

xiaoyu-wu:
Can someone help with the failing CI? It says "The job exceeded the maximum time limit for jobs, and has been terminated." I have no clue why. Doesn't seem to be related with my changes.

jni:
Sometimes the builds hang. I've restarted it. Thank you!

skimage/transform/hough_transform.py
```diff
@@ -315,6 +317,12 @@ def hough_circle_peaks(hspaces, radii, min_xdistance=1, min_ydistance=1,
     >>> img[x, y] = 1
     >>> hspaces = transform.hough_circle(img, radius)
     >>> accum, cx, cy, rad = hough_circle_peaks(hspaces, [radius,])
+
+    Notes
+    -----
+    Circles with bigger radius have higher peaks in Hough space. If larger
+    circles are preferred over smaller ones, `normalize` should be False.
+    Otherwise, circles will be returned in the order of decreasing perfectness.
```
emmanuelle: I'm not a huge fan of "perfectness" here. How about "voting number" (this is the terminology used in https://en.wikipedia.org/wiki/Circle_Hough_Transform).

skimage/transform/hough_transform.py
```diff
@@ -342,10 +350,56 @@ def hough_circle_peaks(hspaces, radii, min_xdistance=1, min_ydistance=1,
...
+        return (np.array(accum_kept), np.array(cx_kept),
+                np.array(cy_kept), np.array(r_kept))
+    kd_tree = cKDTree(scaled_coordinates)
+    while i < tnp:
+        if not removed[i]:
+            accum_kept.append(accum_sorted[i])
+            cx_kept.append(cx_sorted[i])
+            cy_kept.append(cy_sorted[i])
+            r_kept.append(r_sorted[i])
+            # Find a shrot list of candidates to remove
```
emmanuelle: short

skimage/transform/hough_transform.py
xiaoyu-wu: agreed and changed.

skimage/transform/hough_transform.py
xiaoyu-wu: thanks for catching that.

xiaoyu-wu:
@jni Can you please restart CI?

skimage/transform/hough_transform.py
```diff
@@ -342,10 +351,56 @@ def hough_circle_peaks(hspaces, radii, min_xdistance=1, min_ydistance=1,
...
+            neighbors_i = kd_tree.query_ball_point(
+                (scaled_cx_sorted[i], scaled_cy_sorted[i]), np.sqrt(2)
+            )
+            # Check distance in both dimensions and mark as removed if close
+            for ni in neighbors_i:
+                x_close = abs(scaled_cx_sorted[ni] - scaled_cx_sorted[i]) <= 1
+                y_close = abs(scaled_cy_sorted[ni] - scaled_cy_sorted[i]) <= 1
+                if x_close or y_close:
+                    removed[ni] = True
+        i += 1
```
jni: For this section, I would:

- use a for-loop instead of a while-loop. `for i in range(tnp):`
- not scale, but rather use `np.hypot(min_xdistance, min_ydistance)` for the ball tree radius, and compare the distance to `min_xdistance` and `min_ydistance` instead of 1. This removes some additional variables, and makes reasoning about the code more local.
- not create lists, but rather, use `removed` at the very end to subset the arrays with boolean masks when returning.

@xiaoyu-wu would you be happy to make these modifications? If you don't have time, I would be happy to take over for you if needed.

jni: I would also make this whole chunk a separate function that we call out to, to keep the hough_peaks code itself shorter and more readable.

xiaoyu-wu:
No problem. I can address your comments in this PR.

skimage/transform/hough_transform.py
xiaoyu-wu: > use a for-loop instead of a while-loop. `for i in range(tnp):`

Here the while loop is used to stop the iteration early once the number of distant circles reach max numbers to keep. The number of iteration is unknown, which needs a while loop.

xiaoyu-wu: > not scale, but rather use np.hypot(min_xdistance, min_ydistance) for the ball tree radius, and compare the distance to min_xdistance and min_ydistance instead of 1. This removes some additional variables, and makes reasoning about the code more local.

I agree that the code is easier to read this way. There is one potential pitfall when `min_xdistance` and `min_ydistance` differ largely. Each time the list returned by `query_ball_point` can include too many unnecessary candidates. But I understand it might be a very rare case.

> not create lists, but rather, use removed at the very end to subset the arrays with boolean masks when returning.

Agreed and changed as suggested.

jni: @xiaoyu-wu I don't understand: `i` starts at 0, is incremented by 1 each time, and is checked against a maximum value. There are no other increments of i and there is no `break` anywhere? Right now you have:

```python
i = 0
while i < max_points:
    # bunch of code that does not modify i
    i += 1
```

This is exactly equivalent to:

```python
for i in range(max_points):
    # bunch of code that does not modify i
```

... Or am I missing something?

xiaoyu-wu: @jni you are right. it was a bug. It could return circles less than specified `total_num_peaks`. It wasn't caught in test because `total_num_peaks` was tested with value 1. I fixed the bug (and now using a for loop) and added another regression test.

jni: 🙏 !

xiaoyu-wu: @jni  With the latest change, I think I have addressed all your comments and the PR is ready for a final round of review. (CI needs to be restarted.)

jni:
Thank you for your contribution and for your patience iterating with us, @xiaoyu-wu! 🎉

xiaoyu-wu:
@jni @emmanuelle @sciunto @lagru Thank you all for your feedback. I learned a lot in this PR.

PR #4351, opened 2019-12-10
Deprecate CONTRIBUTORS.txt and replace by git shortlog command

stefanv:
The CONTRIBUTORS.txt file was not being consistently updated, so there's
not much sense in keeping it around as a mechanism of credit.


stefanv:
Thanks for the quick review and merge, @jni and @soupault. Much appreciated!

PR #4356, opened 2019-12-11
Add example to highlight regionprops_table

mkcor:
## Description

This PR addresses #4335 and follows up with conversation at #4346.

I hardly modified @jni's proposal. I included a slightly verbose description to contextualize and motivate the example. In the visualization, I used `jitter=0.2` instead of the default `jitter=True` to scatter the dots a tiny bit more.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/segmentation/plot_regionprops_table.py
+++ b/doc/examples/segmentation/plot_regionprops_table.py
@@ -26,7 +26,7 @@ fractions = np.linspace(0.05, 0.5, 10)
 
 #####################################################################
 # 2D images
-# ========
+# =========
 
 images = [data.binary_blobs(volume_fraction=f) for f in fractions]
 
@@ -68,14 +68,13 @@ Python library dedicated to statistical data visualization) with argument
 sns_fig, sns_ax = plt.subplots()
 sns.stripplot(x='volume fraction', y='area', data=areas, jitter=True,
               ax=sns_ax)
-x_format = sns_ax.xaxis.get_major_formatter()
-x_format.seq = ['{:0.2f}'.format(float(s)) for s in x_format.seq]
-sns_ax.xaxis.set_major_formatter(x_format)
+# Fix floating point rendering
+sns_ax.set_xticklabels([f'{frac:.2f}' for frac in fractions])
 plt.show()
 
 #####################################################################
 # 3D images
-# ========
+# =========
 # Doing the same analysis in 3D, we find a much more dramatic behaviour: blobs
 # coalesce into a single, giant piece as the volume fraction crosses ~0.25.
 
@@ -98,8 +97,7 @@ blob_volumes = pd.concat(tables, axis=0)
 d3_fig, d3_ax = plt.subplots()
 sns.stripplot(x='volume fraction', y='area', data=blob_volumes, jitter=True,
               ax=d3_ax)
-x_format = d3_ax.xaxis.get_major_formatter()
-x_format.seq = ['{:0.2f}'.format(float(s)) for s in x_format.seq]
-d3_ax.xaxis.set_major_formatter(x_format)
 d3_ax.set_ylabel('blob size (3D)')
+# Fix floating point rendering
+d3_ax.set_xticklabels([f'{frac:.2f}' for frac in fractions])
 plt.show()
--- a/doc/examples/segmentation/plot_regionprops_table.py
+++ b/doc/examples/segmentation/plot_regionprops_table.py
@@ -1,10 +1,10 @@
 """
-=========================
-Tabulate areas of regions
-=========================
+===================================================
+Explore and visualize region properties with pandas
+===================================================
 
 This toy example shows how to compute the area of every labelled region in a
-series of 3D images. The blob-like regions are generated synthetically. As the
+series of 10 images. The blob-like regions are generated synthetically. As the
 fraction of image pixels covered by the blobs increases, the number of blobs
 (regions) decreases, and the size (area) of a single region can get larger and
 larger. The area values are available in a pandas-compatible format, which
@@ -16,15 +16,13 @@ Beyond area, many other region properties are available.
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
-import seaborn as sns
 
 from skimage import data, measure
 
 
 fractions = np.linspace(0.05, 0.5, 10)
 
-images = [data.binary_blobs(length=128, n_dim=3, volume_fraction=f)
-          for f in fractions]
+images = [data.binary_blobs(volume_fraction=f) for f in fractions]
 
 labeled_images = [measure.label(image) for image in images]
 
@@ -39,8 +37,5 @@ for fraction, table in zip(fractions, tables):
 
 areas = pd.concat(tables, axis=0)
 
-ax = sns.stripplot(data=areas, x='volume fraction', y='area', jitter=0.2)
-x_format = ax.xaxis.get_major_formatter()
-x_format.seq = ["{:0.2f}".format(float(s)) for s in x_format.seq]
-ax.xaxis.set_major_formatter(x_format)
+areas.plot(x='volume fraction', y='area', kind='scatter')
 plt.show()
--- a/doc/examples/segmentation/plot_regionprops_table.py
+++ b/doc/examples/segmentation/plot_regionprops_table.py
@@ -50,3 +50,21 @@ ax1.imshow(images[0])
 ax2 = fig.add_subplot(222)
 ax2.imshow(images[-1])
 plt.show()
+
+"""
+In the scatterplot, many points seem to be overlapping at low area values.
+To get a better sense of the distribution, we may want to add some 'jitter'
+to the visualization. To this end, we import `seaborn`, the Python library
+dedicated to statistical data visualization, and use `stripplot` with
+argument `jitter=True`.
+"""
+
+import seaborn as sns
+
+sns_fig, sns_ax = plt.subplots()
+sns.stripplot(x='volume fraction', y='area', data=areas, jitter=True,
+              ax=sns_ax)
+x_format = sns_ax.xaxis.get_major_formatter()
+x_format.seq = ['{:0.2f}'.format(float(s)) for s in x_format.seq]
+sns_ax.xaxis.set_major_formatter(x_format)
+plt.show()
--- a/doc/examples/segmentation/plot_regionprops_table.py
+++ b/doc/examples/segmentation/plot_regionprops_table.py
@@ -1,0 +1,46 @@
+"""
+=========================
+Tabulate areas of regions
+=========================
+
+This toy example shows how to compute the area of every labelled region in a
+series of 3D images. The blob-like regions are generated synthetically. As the
+fraction of image pixels covered by the blobs increases, the number of blobs
+(regions) decreases, and the size (area) of a single region can get larger and
+larger. The area values are available in a pandas-compatible format, which
+makes for convenient data analysis and visualization.
+
+Beyond area, many other region properties are available.
+
+"""
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import seaborn as sns
+
+from skimage import data, measure
+
+
+fractions = np.linspace(0.05, 0.5, 10)
+
+images = [data.binary_blobs(length=128, n_dim=3, volume_fraction=f)
+          for f in fractions]
+
+labeled_images = [measure.label(image) for image in images]
+
+properties = ['label', 'area']
+
+tables = [measure.regionprops_table(image, properties=properties)
+          for image in labeled_images]
+tables = [pd.DataFrame(table) for table in tables]
+
+for fraction, table in zip(fractions, tables):
+    table['volume fraction'] = fraction
+
+areas = pd.concat(tables, axis=0)
+
+ax = sns.stripplot(data=areas, x='volume fraction', y='area', jitter=0.2)
+x_format = ax.xaxis.get_major_formatter()
+x_format.seq = ["{:0.2f}".format(float(s)) for s in x_format.seq]
+ax.xaxis.set_major_formatter(x_format)
+plt.show()
--- a/doc/examples/segmentation/plot_regionprops_table.py
+++ b/doc/examples/segmentation/plot_regionprops_table.py
@@ -3,12 +3,13 @@
 Explore and visualize region properties with pandas
 ===================================================
 
-This toy example shows how to compute the area of every labelled region in a
-series of 10 images. The blob-like regions are generated synthetically. As the
-fraction of image pixels covered by the blobs increases, the number of blobs
-(regions) decreases, and the size (area) of a single region can get larger and
-larger. The area values are available in a pandas-compatible format, which
-makes for convenient data analysis and visualization.
+This toy example shows how to compute the size of every labelled region in a
+series of 10 images. We use 2D images and then 3D images. The blob-like
+regions are generated synthetically. As the volume fraction (i.e., number of
+pixels or voxels covered by the blobs) increases, the number of blobs
+(regions) decreases, and the size (area or volume) of a single region can get
+larger and larger. The area (size) values are available in a pandas-compatible
+format, which makes for convenient data analysis and visualization.
 
 Beyond area, many other region properties are available.
 
@@ -16,11 +17,17 @@ Beyond area, many other region properties are available.
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
+import seaborn as sns
 
 from skimage import data, measure
 
 
 fractions = np.linspace(0.05, 0.5, 10)
+
+"""
+2D images
+---------
+"""
 
 images = [data.binary_blobs(volume_fraction=f) for f in fractions]
 
@@ -54,12 +61,10 @@ plt.show()
 """
 In the scatterplot, many points seem to be overlapping at low area values.
 To get a better sense of the distribution, we may want to add some 'jitter'
-to the visualization. To this end, we import `seaborn`, the Python library
-dedicated to statistical data visualization, and use `stripplot` with
-argument `jitter=True`.
+to the visualization. To this end, we use `stripplot` (from `seaborn`, the
+Python library dedicated to statistical data visualization) with argument
+`jitter=True`.
 """
-
-import seaborn as sns
 
 sns_fig, sns_ax = plt.subplots()
 sns.stripplot(x='volume fraction', y='area', data=areas, jitter=True,
@@ -67,4 +72,34 @@ sns.stripplot(x='volume fraction', y='area', data=areas, jitter=True,
 x_format = sns_ax.xaxis.get_major_formatter()
 x_format.seq = ['{:0.2f}'.format(float(s)) for s in x_format.seq]
 sns_ax.xaxis.set_major_formatter(x_format)
+plt.show()
+
+"""
+3D images
+---------
+"""
+
+images = [data.binary_blobs(length=128, n_dim=3, volume_fraction=f)
+          for f in fractions]
+
+labeled_images = [measure.label(image) for image in images]
+
+properties = ['label', 'area']
+
+tables = [measure.regionprops_table(image, properties=properties)
+          for image in labeled_images]
+tables = [pd.DataFrame(table) for table in tables]
+
+for fraction, table in zip(fractions, tables):
+    table['volume fraction'] = fraction
+
+blob_volumes = pd.concat(tables, axis=0)
+
+d3_fig, d3_ax = plt.subplots()
+sns.stripplot(x='volume fraction', y='area', data=blob_volumes, jitter=True,
+              ax=d3_ax)
+x_format = d3_ax.xaxis.get_major_formatter()
+x_format.seq = ['{:0.2f}'.format(float(s)) for s in x_format.seq]
+d3_ax.xaxis.set_major_formatter(x_format)
+d3_ax.set_ylabel('blob size (3D)')
 plt.show()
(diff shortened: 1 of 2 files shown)


doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,46 @@
+"""
+=========================
+Tabulate areas of regions
```
emmanuelle: Maybe a more generic title such as "Explore and visualize region properties with pandas"?

emmanuelle:
Great example, thank you very much! I just left a comment about the title. I also wonder (and this is also a question for @jni ) why the example uses 3d images and not 2d ones. As much as I love 3d, here we are hitting the problem that area means volume here in 3d, and it might be confusing. Should we use 2d instead?? Also, I played a bit with the different seaborn functions (`sns.violinplot`, `sns.boxplot`) and maybe the example can be even cooler with one of the fancy statistical charts. What do you think @mkcor @jni?

mkcor:
> Should we use 2d instead??

I thought about it, and suspected that @jni chose 3d images to please you!! Personally, I'm happy to get myself and future readers of the docs more acquainted with 3d images... But, minimally, I would need to replace "area" with "hyper-area" or "(hyper-)area" in the narrative parts.

Another advantage of using 2d images would be the weaker growth of "area(volume fraction)"... making for a finer visualization. I prefer keeping seaborn's jitter plot (as opposed to resorting to standard / rigorous statistical plots), because it's perfect at *giving a feeling* for the behaviour at play. In my understanding, the point of these toy examples (MWE) is neither to communicate quantitative results nor to introduce not-so-related tools... But maybe everyone in the scikit-image community knows what a violin plot is, I wouldn't know.

doc/examples/segmentation/plot_regionprops_table.py
mkcor: I'm okay generalizing to (all possible) region properties. With "tabulate" I meant to express "pandas" ;)
The only think about "Explore and visualize ... with pandas" is that I would expect the example not to depend on seaborn, but to use the matplotlib API via pandas:
```py
areas.plot(x='volume fraction', y='area', kind='scatter')
```
which results in a less fancy plot, but saves three lines of code, and looks more readable / powerful.

mkcor:
@emmanuelle The more I think about it, the more I think 2d images would be better-suited in this particular example.

mkcor:
Commit bdf2153 makes the example more to-the-point (i.e., showing the strengths of `_table`) and less distracting.

emmanuelle:
Thank you @mkcor. I agree that 2d is clearer and that we don't need seaborn here, just the plotting backend of pandas is fine. So :+1: from me. @jni can you review as well? Or someone else from @skimage/core?

doc/examples/segmentation/plot_regionprops_table.py
jni: I agree with self-contained, and maybe readable, but certainly not more powerful! =)

jni:
Hi @emmanuelle and @mkcor! Sorry for my silence, I had a busy weekend.

My example was 3D only because that is where I observed the effect that, beyond a threshold, you get a "giant" component (I noticed this while writing the napari blog post), which I thought was pretty interesting! It is *also* interesting that you don't get quite the same effect in 2D!

Looking at the example now, my feeling about 3D is that, if you're not going to show the image, you might as well use 3D, as the effect of a single giant component is more interesting than having some smattering of big segments, to me at least. Otherwise, if you prefer to keep 2D, I would show e.g. the label2rgb images of the lowest and highest volume fraction images. Having the images also has the advantage of showing the `ax=` parameter to pandas `.plot`, which is a nice advanced-ish usage of the matplotlib's object-oriented API. It is still [matplotlib appreciation day](https://twitter.com/ChelleGentemann/status/1206343689720803335), after all, in the US and Europe! ;)

Finally, I have a slight preference towards going back to seaborn for the plot. It makes a much snazzier plot, that is also useful (you get a better sense of how many points there are in each bucket). But I'm not set on that.

So, to summarise:

- Either use 3D, or use 2D but add images
- Use the `ax` keyword argument to create the plot.
- (optional) use seaborn for a snazzier plot

emmanuelle:
@jni percolation theory is cool :-).

doc/examples/segmentation/plot_regionprops_table.py
mkcor: @jni Sure, seaborn is much more powerful in absolute terms! What I had in mind was that we get to showcase the power of pandas' plotting backend *in addition to* (hence, more) its 'base' capabilities of manipulating tables.
In any case, I think we want to either reflect the example title ("Explore and visualize region properties with pandas") or change the title to reflect the example.

mkcor:
@jni Thank you for your thorough review! I see a correlation between (3D vs 2D with images) and (seaborn vs pandas' plotting backend): in the 3D case, since the highest values completely crush the lowest values, the jitter rendering is more critical to get a sense of these low-value buckets.

jni:
@mkcor imho the jitter is valuable in both cases: we are trying to display a distribution at a particular value, and an unjittered plot is a bad method to visualise a distribution.

@emmanuelle I was unaware that this theory had a name. Perhaps there is a paper to link to in the 3D example? =)

Another option is to extend the example to include *both* 2D and 3D, and show the best of all worlds. I think it can be pretty nice to have more extended examples that show you different ways of doing things...

mkcor:
If you double the length of a segment, the size of a corresponding square (2d) is multiplied by 4, while the size of a corresponding cube (3d) is multiplied by 8. The polynomial growth increases with the dimension. I'm happy to include both 2d and 3d in the example then.

mkcor:
> showing the `ax=` parameter to pandas `.plot`, which is a nice advanced-ish usage of the matplotlib's object-oriented API

@jni What do you think about:
subplots
from b6f302c?

Now adding seaborn and 3D.

mkcor:
> Hello @mkcor! Thanks for updating this PR. We checked the lines you've touched for PEP 8 issues, and found:

Wow! I'm impressed. I thought it made more sense in our (notebook-style) narrative to import `seaborn` only at some point... But I'll comply.

mkcor:
Not sure what the CI fail is about... 🤔
@jni please review again 🙏

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,105 @@
...
+Python library dedicated to statistical data visualization) with argument
+`jitter=True`.
+"""
+
+sns_fig, sns_ax = plt.subplots()
+sns.stripplot(x='volume fraction', y='area', data=areas, jitter=True,
+              ax=sns_ax)
+x_format = sns_ax.xaxis.get_major_formatter()
+x_format.seq = ['{:0.2f}'.format(float(s)) for s in x_format.seq]
+sns_ax.xaxis.set_major_formatter(x_format)
```
jni: I'd replace these three lines with a single matplotlib call, together with an explanatory comment:

```python
# small fix for floating point rendering
sns_ax.set_xticklabels([f'{fract:.2f}' for fract in fractions])
```

jni: And same below of course

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,105 @@
...
+plt.show()
+
+"""
+3D images
+---------
+"""
```
jni: I think this syntax no longer works in sphinx-gallery, and this needs to be a comment (see other examples), but @emmanuelle knows best... Personally I like this better, if it works!

Finally, I would add a paragraph to 3D: "Doing the same analysis in 3D, we find a completely different behavior, with blobs coalescing into a single, giant component as the volume fraction crosses ~0.25."

Otherwise, I really like how you've incorporated both pandas and seaborn plotting into the example! 🎉

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,105 @@
...
+
+# Divide the figure into a 2x1 grid
+fig, (im, ax) = plt.subplots(2, 1)
+# Turn off axes for top subplot
+im.axis('off')
+# Plot area vs volume fraction
+areas.plot(x='volume fraction', y='area', kind='scatter', ax=ax)
+# Show image with lowest volume fraction
+ax1 = fig.add_subplot(221)
+ax1.imshow(images[0])
```
jni: Since we are going with percolation theory, how about making black holes on a white background?

```suggestion
ax1.imshow(images[0], cmap='gray_r')
ax1.set_axis_off()
ax1.suptitle(f'fraction {fractions[0]}')
```

Same below.

mkcor: Love it! What did you mean by `suptitle` though (`AttributeError: 'AxesSubplot' object has no attribute 'suptitle'`)? I'll replace it with `set_title` unless you notify me otherwise.

doc/examples/segmentation/plot_regionprops_table.py
mkcor: > I think this syntax no longer works in sphinx-gallery, and this needs to be a comment (see other examples), but @emmanuelle knows best... Personally I like this better, if it works!

Ok, I have found Markdown in between code at https://scikit-image.org/docs/dev/auto_examples/developers/plot_max_tree.html; will use this comment-like syntax then.

mkcor: Done in c747775.

> "Doing the same analysis in 3D, we find a completely different behavior

It's not obvious to me (judging only from this example) that it would be _qualitatively_ (and, hence, completely) different... Some kind of phase transition would need to be evidenced for me to call it a "completely different behaviour." So I've described it as "much more dramatic," to stay of the safe side.

doc/examples/segmentation/plot_regionprops_table.py
mkcor: Done in a61e812.

doc/examples/segmentation/plot_regionprops_table.py
mkcor: Done in 654d47f.

mkcor:
@jni Thank you! I think I've gone through all your comments (and I thought of using https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors in the last two commits). ;)

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,107 @@
...
+ax2.set_title(f'fraction {fractions[-1]}')
+plt.show()
+
+"""
+In the scatterplot, many points seem to be overlapping at low area values.
+To get a better sense of the distribution, we may want to add some 'jitter'
+to the visualization. To this end, we use `stripplot` (from `seaborn`, the
+Python library dedicated to statistical data visualization) with argument
+`jitter=True`.
+"""
```
jni: This also needs to be a comment.

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,107 @@
...
+              ax=sns_ax)
+# Fix floating point rendering
+sns_ax.set_xticklabels([f'{frac:.2f}' for frac in fractions])
+plt.show()
+
+#####################################################################
+# 3D images
+# =========
+# Doing the same analysis in 3D, we find a much more dramatic behaviour: blobs
+# coalesce into a single, giant piece as the volume fraction crosses ~0.25.
```
jni: I still would call it a phase transition, but your wording is totally fine by me. Perhaps @emmanuelle has the definitive answer? Is it normally called a phase transition in this context?

jni:
@mkcor btw the multiple authors use thing is amazing, I think this is the first explicit use of that feature in the history of the project! (cherry-pick does something like this automatically.)

The macOS Travis failures are not related to this PR.

doc/examples/segmentation/plot_regionprops_table.py
mkcor: I do believe that there is a phase transition here. But to assert it, we would need (at the very least) to do some modelling (e.g., regression), in order to measure (estimate) either the sharp slope change or some power law exponent. So it's not that writing "phase transition" or "completely different behaviour" would be inaccurate -- it may very well be accurate. But for the example to make sense (and be self-contained), if we were to write that, we would need to either derive it, or point to a reference (and not just a general reference, I believe, but one that smoothly connects this toy example to a derivation of some critical volume fraction...).

doc/examples/segmentation/plot_regionprops_table.py
mkcor: Oops! Fixed in d1b9933.

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,106 @@
+"""
+===================================================
+Explore and visualize region properties with pandas
+===================================================
+
+This toy example shows how to compute the size of every labelled region in a
+series of 10 images. We use 2D images and then 3D images. The blob-like
+regions are generated synthetically. As the volume fraction (i.e., number of
```
soupault:
```suggestion
regions are generated synthetically. As the volume fraction (i.e., ratio of
```

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,106 @@
...
+pixels or voxels covered by the blobs) increases, the number of blobs
+(regions) decreases, and the size (area or volume) of a single region can get
```
soupault:
```suggestion
(regions) decreases, and the size (area or volume) of a single region gets
```

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,106 @@
...
+larger and larger. The area (size) values are available in a pandas-compatible
+format, which makes for convenient data analysis and visualization.
```
soupault:
```suggestion
format, which makes subsequent data analysis and visualization convenient.
```

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,106 @@
...
+
+Beyond area, many other region properties are available.
```
soupault:
```suggestion
Besides area, many other region properties are available.
```

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,106 @@
...
+areas = pd.concat(tables, axis=0)
+
+# Divide the figure into a 2x1 grid
+fig, (im, ax) = plt.subplots(2, 1)
+# Turn off axes for top subplot
+im.axis('off')
+# Plot area vs volume fraction
+areas.plot(x='volume fraction', y='area', kind='scatter', ax=ax)
+# Show image with lowest volume fraction
+ax1 = fig.add_subplot(221)
```
soupault: It is a bit confusing that you are using two different ways of allocating the subplots: `fig, (im, ax) = plt.subplots(2, 1)` and `ax1 = fig.add_subplot(221)`. `plt.subplots` could be used solely, see https://github.com/scikit-image/scikit-image/blob/master/doc/examples/numpy_operations/plot_view_as_blocks.py#L46-L52 for the example.

doc/examples/segmentation/plot_regionprops_table.py
```diff
@@ -0,0 +1,106 @@
...
+tables = [measure.regionprops_table(image, properties=properties)
+          for image in labeled_images]
+tables = [pd.DataFrame(table) for table in tables]
+
+for fraction, table in zip(fractions, tables):
+    table['volume fraction'] = fraction
+
+blob_volumes = pd.concat(tables, axis=0)
+
+d3_fig, d3_ax = plt.subplots()
```
soupault: Consider using the same variable names: `fig, ax = plt.subplots()`.

requirements/docs.txt
```diff
@@ -10,3 +10,5 @@ matplotlib>=3.0.1
 dask[array]>=0.15.0
 # cloudpickle is necessary to provide the 'processes' scheduler for dask
 cloudpickle>=0.2.1
+pandas
+seaborn
```
soupault: Please, specify some minimal version here, similar to another PR of yours.

soupault:
I've made some stylistic suggestions, otherwise :+1: . Thank you!

doc/examples/segmentation/plot_regionprops_table.py
mkcor: I would actually keep the probabilistic touch.

doc/examples/segmentation/plot_regionprops_table.py
mkcor: I thought it was better to distinguish between all the `fig` and `ax` objects but, now that you point it out, I realize that possible styling additions (by calling methods of these objects) would take place within the same code cell (say, running in a Jupyter notebook) so there should be no confusion...

soupault: As it goes so far, the examples are made of one or multiple sections (≈ cells), each of which is logically ended up with `plt.plot()` call. So I'd say that your hypothetic scenario is unlikely.

I personally believe that clarity is the metric we should "optimize" the examples for. :)

mkcor: I agree with you. Thanks.

mkcor: Fixed in ae3e04c775acc973b8c59d42b557bda358b6181a.

doc/examples/segmentation/plot_regionprops_table.py
mkcor: I don't find the suggested change better than the original...

requirements/docs.txt
mkcor: Fixed in 7f9a303c5d6a61f8611de4c3d67caa59c7d0e573. Too bad the NEP 29 script didn't work out of the box for `pandas`, because it has old versions such as `'0.2b1'` and `'2beta'`!

doc/examples/segmentation/plot_regionprops_table.py
mkcor: I didn't know of the `ravel` method, thanks. If I start with a 2x2 grid of subplots, I don't know how to 'merge' the bottom subplots to get this rendering:
add_subplot
Using
```py
fig, axes = plt.subplots(2, 2)
ax = axes.ravel()
# Plot area vs volume fraction
areas.plot(x='volume fraction', y='area', kind='scatter', ax=ax[2])
# Show image with lowest volume fraction
ax[0].axis('off')
ax[0] = fig.add_subplot(221)
ax[0].imshow(images[0], cmap='gray_r')
ax[0].set_axis_off()
ax[0].set_title(f'fraction {fractions[0]}')
# Show image with highest volume fraction
ax[1].axis('off')
ax[1] = fig.add_subplot(222)
ax[1].imshow(images[-1], cmap='gray_r')
ax[1].set_axis_off()
ax[1].set_title(f'fraction {fractions[-1]}')
plt.show()
```
I get
ravel
... as expected. Morally, I would like to express (in quotes) `areas.plot(x='volume fraction', y='area', kind='scatter', "ax=ax[2] + ax[3]")`: Do you know how to achieve this?

doc/examples/segmentation/plot_regionprops_table.py
jni: I don't think we need to prove that it's a phase transition. We are not publishing a scientific paper, just illustrating a few scientific ideas. People who need the proof can go read the scientific literature.

doc/examples/segmentation/plot_regionprops_table.py
jni: @mkcor I think the canonical way is to use gridspec: https://matplotlib.org/3.1.1/tutorials/intermediate/gridspec.html

I don't know of a way to use `plt.subplots` and then join two plots. I don't think that exists. Having said that, I dislike the `add_subplot(221)` syntax and prefer to use `add_subplot(2, 2, 1)` if necessary. ;)

soupault: I think GridSpec is the way to go. It is unlikely that we can merge subplots after `.subplots` call 🙂 .

doc/examples/segmentation/plot_regionprops_table.py
soupault: If it is a correct English, I am more than happy to keep your version :).

mkcor: Here's a reference for you: https://www.collinsdictionary.com/dictionary/english/make-for ;)

doc/examples/segmentation/plot_regionprops_table.py
soupault: Here is an amazing recent post in matplotlib blog, btw - https://matplotlib.org/matplotblog/posts/an-inquiry-into-matplotlib-figures/ .

mkcor: Thank you, @soupault and @jni! Done in e6ed72a38ed22abf27cbe3e6dbee7fed3be9741b. I confess that I followed https://jakevdp.github.io/PythonDataScienceHandbook/04.08-multiple-subplots.html, not just because I :heart: @jakevdp, but also to save an extra import.

doc/examples/segmentation/plot_regionprops_table.py
emmanuelle: You can add a sentence here saying "This corresponds to the percolation threshold in statistical physics and graph theory" and link to https://en.wikipedia.org/wiki/Percolation_threshold. No need to say more, most people probably don't care :-).

mkcor: Love it! Done in 73acfbe51a56be0a3e1d687c5090815923890768.

Issue #2360, opened 2016-10-30
Numerical inaccuracies with the random_walker


hugues-talbot:
## Description

The random walker segmentation code is very useful but behaves sometimes very strangely with respect to numerical results.

See following code (sorry this is slow):

result:

Should be between 0 and 1: [0.000000 ; 1.103701]
## Way to reproduce
```
import numpy as np
from skimage.filters import gaussian
from skimage.segmentation import random_walker

np.random.seed(42)
blob3d = np.random.randn(101,101,101) # 3D image of random noise
blob3ds = gaussian(blob3d,sigma=10.) # makes a Gaussian random field
p05=np.percentile(blob3ds,5.) # Normalize the GRF between 0 and 1
p95=np.percentile(blob3ds,95.)
blob3ds = (blob3ds-p05)/(p95-p05)
blob3ds = np.clip(blob3ds,0,1)

markers3d = np.zeros((101,101,101)) 
markers3d[45:55,45:55,45:55] = 1  # central box
markers3d[:,:,0]=markers3d[:,0,:]=markers3d[0,:,:] = 2 # border of the box
markers3d[100,:,:]=markers3d[:,100,:] = markers3d[:,:,100] = 2 # border of the box cont.

grow3d = random_walker(blob3ds, markers3d, beta=5e4, mode='cg_mg', tol=1e-5, return_full_prob=True)

print("Should be between 0 and 1: [%f ; %f]" % (np.min(grow3d[1]), np.max(grow3d[1])))
```

jni:
@JDWarner any ideas about this? Pretty funny to see a probability of 1.1. =D

jmetz:
This code doesn't work for me... I get 
```
Traceback (most recent call last):
  File "issue_2360.py", line 18, in <module>
    grow3d = random_walker(blob3ds, markers3d, beta=5e4, mode='mg_cg', tol=1e-5, return_full_prob=True)
  File ".../skimage/segmentation/random_walker_segmentation.py", line 459, in random_walker
    for Xline in X])
UnboundLocalError: local variable 'X' referenced before assignment
```

This should probably be added as a bug also?!?

jni:
@jmetz LOL #2361 #2362 

The correct mode is `'cg_mg'`.

jmetz:
Thanks for that @jni - I got it to work and can confirm a similar result using the `"cg"` method also:

```
Method cg
Should be between 0 and 1: [0.000000 ; 1.256141]
[...]

PR #4359, opened 2019-12-13
Add Jacobi precondtioned conjugate gradient mode to random_walker

rfezzani:
## Description

With the global refactoring of the `random_walker` function, this PR proposes:

- the 'cg_j' mode (conjugate gradient with Jacobi precondtioner) is added,
- `labels` and `data` shapes validation,
- conjugate gradient convergence validation,
- default mode modification in favor of 'cg_j',
- warnings messages update,
- warnings `stacklevel` fixing.
- `return_full_prob` argument documentation is updated to warn user about the behavior observed in #2360

The new 'cg_j' mode offers a good option without the need for installing the extra dependencies pyamg nor UMFPACK. 

#### 2D case

Example from the [gallery](https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_random_walker_segmentation.html#sphx-glr-auto-examples-segmentation-plot-random-walker-segmentation-py):

##### Settings

```python
In [1]: from skimage.segmentation import random_walker 
   ...: from skimage.data import binary_blobs 
   ...: from skimage.exposure import rescale_intensity 
   ...: import skimage 
   ...:  
   ...: # Generate noisy synthetic data 
   ...: data = skimage.img_as_float(binary_blobs(length=1024, seed=1)) 
   ...: sigma = 0.35 
   ...: data += np.random.normal(loc=0, scale=sigma, size=data.shape) 
   ...: data = rescale_intensity(data, in_range=(-sigma, 1 + sigma), 
   ...:                          out_range=(-1, 1)) 
   ...:  
   ...: # The range of the binary image spans over (-1, 1). 
   ...: # We choose the hottest and the coldest pixels as markers. 
   ...: markers = np.zeros(data.shape, dtype=np.uint) 
   ...: markers[data < -0.95] = 1 
   ...: markers[data > 0.95] = 2
```

##### Version 0.16.2

```python
In [3]: %timeit random_walker(data, markers, beta=10, mode='cg') 
 /home/rfez/.conda/envs/skimage-plain/bin/ipython:1: UserWarning: "cg" mode will be used, but it may be slower than "bf" because SciPy was built without UMFPACK. Consider rebuilding SciPy with UMFPACK; this will greatly accelerate the conjugate gradient ("cg") solver. You may also install pyamg and run the random_walker function in "cg_mg" mode (see docstring).
  #!/home/rfez/.conda/envs/skimage-plain/bin/python
1.46 s ± 30.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)                            

In [4]: %timeit random_walker(data, markers, beta=10, mode='cg_mg')                                       
2.14 s ± 7.36 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: %timeit random_walker(data, markers, beta=10, mode='bf')                                        
9.78 s ± 700 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

##### Proposed implementation

```python
In [2]: %timeit random_walker(data, markers, beta=10, mode='cg')                                         
/home/rfez/Projects/Misc/skimage/skimage/segmentation/random_walker_segmentation.py:486: UserWarning: 'cg' mode may be slow because UMFPACK is not availabel. Consider building Scipy with UMFPACK or use a preconditioned version of CG ('cg_j' or 'cg_mg' modes).
  X = _solve_linear_system(lap_sparse, B, tol, mode)
1.55 s ± 72.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit random_walker(data, markers, beta=10, mode='cg_j')
1.48 s ± 12.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %timeit random_walker(data, markers, beta=10, mode='cg_mg')                                      
2.18 s ± 23.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: %timeit random_walker(data, markers, beta=10, mode='bf')                                         
9.5 s ± 314 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

##### Settings

Example from #2360:

```python
In [6]: import numpy as np 
   ...: from skimage.filters import gaussian 
   ...: from skimage.segmentation import random_walker 
   ...:  
   ...: np.random.seed(42) 
   ...: blob3d = np.random.randn(101,101,101) # 3D image of random noise 
   ...: blob3ds = gaussian(blob3d,sigma=10.) # makes a Gaussian random field 
   ...: p05=np.percentile(blob3ds,5.) # Normalize the GRF between 0 and 1 
   ...: p95=np.percentile(blob3ds,95.) 
   ...: blob3ds = (blob3ds-p05)/(p95-p05) 
   ...: blob3ds = np.clip(blob3ds,0,1) 
   ...:  
   ...: markers3d = np.zeros((101,101,101)) 
   ...: markers3d[45:55,45:55,45:55] = 1 # central box 
   ...: markers3d[:,:,0]=markers3d[:,0,:]=markers3d[0,:,:] = 2 # border of the box 
   ...: markers3d[100,:,:]=markers3d[:,100,:] = markers3d[:,:,100] = 2 # border of the box cont.   
```

##### Version 0.16.2

```python
In [3]: %timeit random_walker(data, markers, mode='cg') 
 /home/rfez/.conda/envs/skimage-plain/bin/ipython:1: UserWarning: "cg" mode will be used, but it may be slower than "bf" because SciPy was built without UMFPACK. Consider rebuilding SciPy with UMFPACK; this will greatly accelerate the conjugate gradient ("cg") solver. You may also install pyamg and run the random_walker function in "cg_mg" mode (see docstring).
  #!/home/rfez/.conda/envs/skimage-plain/bin/python
28.7 s ± 454 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)            

In [5]: %timeit random_walker(data, markers, mode='cg_mg')                                   
8.23 s ± 21.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Brute force computation is not affordable without UMFPACK.

##### Proposed version

```python      
In [7]: %timeit random_walker(blob3d, markers3d, mode='cg')                                              
/home/rfez/Projects/Misc/skimage/skimage/segmentation/random_walker_segmentation.py:486: UserWarning: 'cg' mode may be slow because UMFPACK is not availabel. Consider building Scipy with UMFPACK or use a preconditioned version of CG ('cg_j' or 'cg_mg' modes).
28.1 s ± 1.05 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit random_walker(blob3d, markers3d, mode='cg_j')                                            
16.5 s ± 158 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: %timeit random_walker(blob3d, markers3d, mode='cg_mg')                                           
8.34 s ± 9.92 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/segmentation/random_walker_segmentation.py
```diff
@@ -272,16 +299,18 @@ def random_walker(data, labels, beta=130, mode='bf', tol=1.e-3, copy=True,
           using the Conjugate Gradient method from scipy.sparse.linalg. This is
           less memory-consuming than the brute force method for large images,
           but it is quite slow.
+        - 'cg_j' (conjugate gradient with Jacobi preconditionner): the
+          Jacobi preconditionner is applyed during the Conjugate
+          gradient method iterations. This may accelerate the
+          convergeance of the 'cg' method.
```
emmanuelle: convergence

emmanuelle:
Thank you @rfezzani ! I'll review the PR since I'm quite familiar with the random walker. What is the memory footprint of this new mode compared to the existing ones, for large images (such as the ones of your benchmarks)?

emmanuelle:
Also, in the benchmarks you wrote (thank you for these! It really helps.), it appears that scipy was build without umfpack. If this is confirmed, what is the advantage of `cg_j` when its execution time is of the order of (the slow) `cg`?

jni:
@emmanuelle since SciPy's wheels on PyPI are built without UMFPACK (just confirmed this with the newly-released 1.4.0), I think speedups in those cases are welcome. "On the order of" is a bit of an exaggeration imho since this is 2x faster for large 3D volumes!

emmanuelle:
@jni you're right I had concentrated on the 2d case, a factor of 2 is indeed good! Now that we're going to have four modes, it could be interesting to have helper functions to help a user benchmark the different modes. Something for our common performance  goal for CZI?

jni:
> it could be interesting to have helper functions to help a user benchmark the different modes. Something for our common performance goal for CZI?

That's very interesting! One could imagine something general, like:

```python
from skimage.speed import benchmark_parameter

param_values = ['cg', 'cg_j', 'bf', 'cg_mg']
times = benchmark_parameter(segmentation.random_walker, 'mode', param_values,
                            args=[mydata, mymarkers], kwargs=dict(beta=10))
```

With a bunch of examples, this could be a very useful function.

skimage/segmentation/random_walker_segmentation.py
rfezzani: Thx, fixed!

rfezzani:
> What is the memory footprint of this new mode compared to the existing ones, for large images (such as the ones of your benchmarks)?

Hi @emmanuelle, the Jacobi preconditioner (also known as the diagonal preconditioner) is the inverse of the diagonal of the matrix, so it only requires `num_unlabelled_pixels` more floats of memory usage.

I have a couple of suggestions, but I don't know if they are acceptable:
- what about deprecating 'cg' mode? Jacobi preconditioner accelerates the convergence of CG with a reasonable overhead without the need for any extra package installation. Removing 'cg' also allows to remove the warning and the documentations about 'umfpack'.
- another suggestion also would be to allow Cholesky decomposition for "brute force" mode if `scikit-sparse` is available.

Finally, If you agree with the use of 'cg_j' as default mode, a deprecation cycle has to be started.

skimage/segmentation/random_walker_segmentation.py
```diff
@@ -290,21 +319,25 @@ def random_walker(data, labels, beta=130, mode='bf', tol=1.e-3, copy=True,
         If True, input data is parsed as multichannel data (see 'data' above
         for proper input format in this case).
     return_full_prob : bool, optional
-        If True, the probability that a pixel belongs to each of the labels
-        will be returned, instead of only the most likely label.
+        If True, the probability that a pixel belongs to each of the
+        labels will be returned, instead of only the most likely
+        label. The returned probability values may be inaccurate
```
emmanuelle: hum, this sentence is worrying but not very informative... I think we should raise a warning instead as was suggested in #2360.

skimage/segmentation/random_walker_segmentation.py
```diff
@@ -77,163 +74,195 @@ def _make_graph_edges_3d(n_x, n_y, n_z):
...
+        [np.diff(data[..., 0], axis=ax).ravel() / spacing[ax]
+         for ax in [2, 1, 0] if data.shape[ax] > 1], axis=0) ** 2
+    for channel in range(1, data.shape[-1]):
+        gradients += np.concatenate(
+            [np.diff(data[..., channel], axis=ax).ravel() / spacing[ax]
+             for ax in [2, 1, 0] if data.shape[ax] > 1], axis=0) ** 2
+
     # All channels considered together in this standard deviation
-    beta /= 10 * data.std()
+    scale_factor = -beta / (10 * data.std())
```
rfezzani: I don't really understand/like this hard coded multiplication by 10, but it is in the actual implementation...

emmanuelle: Yeah it's quite bad :-(. (it's likely I was the one to introduce this :-()

skimage/segmentation/random_walker_segmentation.py
rfezzani: I agree, but we have to set a tolerance above which the warning is raised, because it is not rare that the returned "probability" value is > 1!
The problem is in the use of the word "probability" since the method doesn't give any guaranty that the result is in the range (0, 1).
I see in the [ref](http://vision.cse.psu.edu/people/chenpingY/paper/grady2006random.pdf) page 6 that the author uses the term "probabilities/potentials". Can't we use this wording instead?

skimage/segmentation/random_walker_segmentation.py
```diff
@@ -77,163 +74,195 @@ def _make_graph_edges_3d(n_x, n_y, n_z):
...
+        maxiter = None
+        if mode == 'cg':
+            if UmfpackContext is None:
+                warn('"cg" mode may be slow because UMFPACK is not availabel. '
+                     'Consider building Scipy with UMFPACK or use a '
+                     'preconditioned version of CG ("cg_j" or "cg_mg" modes).',
+                     stacklevel=2)
+            M = None
+        elif mode == 'cg_j':
+            M = sparse.diags(1.0 / lap_sparse.diagonal())
```
emmanuelle: Beautifully simple :-). Do you confirm that there shouldn't be any zeros on the diagonal because this happens only for isolated pixels, which are taken care of before?

rfezzani: Yes, exactly: the diagonal is minus the sum of the extra-diagonal elements of the matrix for each row. These extra-diagonal elements are all negative/non null, the diagonal elements can then not be null.

skimage/segmentation/random_walker_segmentation.py
emmanuelle: How about we just test whether the returned probability is >1 (possibly with some small tolerance like 1.e-4)? It does not have to be in this PR, it's just that it'd prefer to have a warning rather than a sentence which will mostly scare people or be unnoticed.

emmanuelle:
This is beautiful work, thank you @rfezzani for all the improvements. (quite a lot of code shuffling and refactoring so it was easier for me to review the new files than the diff). Just a small hesitation about the docstring for `return_full_prob`, but I'll let you decide here.

emmanuelle:
Now we need a second reviewer @scikit-image/core :-)

skimage/segmentation/random_walker_segmentation.py
rfezzani: I agree that the sentence that I introduced can be discouraging...
I am not against raising a warning, I am just saying that if we replace the argument name`return_full_prob` by ` return_full_potentials`, we alleviate the need for ensuring that the returned values are in the range (0, 1)...
May be someone else vote can ease our decision :-)

rfezzani:
Thank you very much @emmanuelle for your review, let @scikit-image/core reviewers help us now  decide about the `return_full_prob` issue ;-) Thanks again.

skimage/segmentation/random_walker_segmentation.py
jni: I suggest we remove this comment for now (it is a separate issue from this PR), and create a PR specifically for it. `return_full_potentials` is not a bad choice, but it's unclear that it's worth a deprecation cycle. It's good to keep PRs as small and self-contained as possible!

emmanuelle:
This PR is waiting for a second review / approval :-) @skimage/core

jni:
@JDWarner any chance you could have a look at this sometime soon? ;) From a quick go-over it looks good to me, but I am unfamiliar with the method so I would like to defer to more experienced core members such as yourself. =)

emmanuelle:
@jni the method itself is not modified, it's just a lot of refactoring + a new mode to invert a linear system.

skimage/segmentation/random_walker_segmentation.py
```diff
@@ -30,7 +30,7 @@ def new_del(self):
             pass
     umfpack.UmfpackContext.__del__ = new_del
     UmfpackContext = umfpack.UmfpackContext()
-except:
+except Exception:
```
alexdesiqueira: I didn't understand what could trigger this exception. Could we have a more specific `Exception` here? :blush:

rfezzani: You are right, I specified the `ImportError` raised if umfpack is not installed ;-)

emmanuelle:
@alexdesiqueira are you able either to approve the PR or to request other changes?

alexdesiqueira:
@emmanuelle sorry about that, I didn't know I was the one blocking this :sweat:
Thank you for your work @rfezzani!

emmanuelle:
no worries @alexdesiqueira I was just eagerly waiting for a second approval to merge this PR :-). It's just a bit annoying that the CI is not completely green and I don't think to have the credentials on Azure to restart the build... @scikit-image/core is there a list of whitelisted persons for the Azure CI or should it just read Github permissions?

rfezzani:
Thank you @emmanuelle and @alexdesiqueira ;-)

emmanuelle:
@rfezzani don't thank us yet I would like to sort out this CI thing first :-)

emmanuelle:
OK, merging since the Azure thing is taken care of in #4411 . Thanks @rfezzani !

soupault:
@emmanuelle I have added you to the team in Azure Pipelines. Could you check if you have the permissions now?

jni:
@soupault could you add everyone in @scikit-image/core?

emmanuelle:
@soupault thanks! I can see that I'm part of the scikit-image team on Azure but I don't know what I should check about permissions. For example, I still do not find where to restart jobs but maybe this is not an Azure feature?

soupault:
@emmanuelle if you can see "Edit" and "Run pipeline" in the top right here - https://dev.azure.com/scikit-image/scikit-image/_build?definitionId=1&_a=summary, that is a good sign. To restart, you should, for example, press "Rerun failed jobs" here - https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=2332&view=results, or use "Run new" to start a customized build.

soupault:
@jni I have invited all the active team members (was not able to get the emails of others 😅).

jni:
Thank you! :pray:

jni:
I don't see it. Could you use my institutional address, juan.nunez-iglesias@monash.edu ? That appears to be the one Azure has picked up from GitHub...

soupault:
Done!

PR #4363, opened 2019-12-15
Fix float32 promotion in cubic interpolation

Mostafa-Alaa:
## Description

This PR fixes a problem where floats get promoted to doubles in cubic interpolation, yielding in performance degradation.

Disassembling generated bicubic_interpolation for float32 in _warps_cy.so I noticed that a lot of conversions are done from float32 to float64 and all operations are done in float64 then converted back to float32, that's mainly due to mixing operations with floating-point literals that are of type float64.

``` python
from math import radians
from timeit import timeit

import numpy as np
from skimage.data import astronaut
from skimage.transform import AffineTransform, warp


astro = astronaut()
astrof = astro.astype(np.float32)
astrod = astro.astype(np.float64)


def transform(image, angle, scale):
    center = np.array(image.shape[:2]) / 2. - 0.5
    tform1 = AffineTransform(translation=center)
    tform2 = AffineTransform(rotation=radians(angle), scale=(scale, scale))
    tform3 = AffineTransform(translation=-center)
    tform = tform3 + tform2 + tform1
    warped = warp(image, tform, order=3, preserve_range=True)
    return warped


timeit_globals = {'transform': transform, 'astrof': astrof, 'astrod': astrod}


if __name__ == '__main__':
    print(timeit('transform(astrof, 20, 1.2)', number=1000, globals=timeit_globals))
    print(timeit('transform(astrod, 20, 1.2)', number=1000, globals=timeit_globals))
```

Running this quick benchmark shows the following results:
```
    Old float32 = 72.4
    New float32 = 66.9
    float64 = 74.5
```

The fix offers a 10.2% improvement instead of originally 2.8% when switching from float64 to float32.

Any comments, changes are welcome.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/interpolation.pxd
+++ b/skimage/_shared/interpolation.pxd
@@ -203,11 +203,17 @@ cdef inline np_floats cubic_interpolation(np_floats x, np_real_numeric[4] f) nog
         Interpolated value to be used in bicubic_interpolation.
 
     """
-    return (\
-        f[1] + 0.5 * x * \
-            (f[2] - f[0] + x * \
-                (2.0 * f[0] - 5.0 * f[1] + 4.0 * f[2] - f[3] + x * \
-                    (3.0 * (f[1] - f[2]) + f[3] - f[0]))))
+    return (
+        f[1] + <np_floats>0.5 * x * (
+            f[2] - f[0] + x * (
+                <np_floats>2.0 * f[0] -
+                <np_floats>5.0 * f[1] +
+                <np_floats>4.0 * f[2] - f[3] + x * (
+                    <np_floats>3.0 * (f[1] - f[2]) + f[3] - f[0]
+                )
+            )
+        )
+    )
 
 
 cdef inline void bicubic_interpolation(np_real_numeric* image,
--- a/skimage/_shared/interpolation.pxd
+++ b/skimage/_shared/interpolation.pxd
@@ -205,6 +205,9 @@ cdef inline np_floats cubic_interpolation(np_floats x, np_real_numeric[4] f) nog
         Interpolated value to be used in bicubic_interpolation.
 
     """
+
+    # Explicitly cast a floating point literal to the other operand's type
+    # to prevent promoting operands unnecessarily to double precision
     return (
         f[1] + <np_floats>0.5 * x * (
             f[2] - f[0] + x * (
--- a/skimage/_shared/interpolation.pxd
+++ b/skimage/_shared/interpolation.pxd
@@ -23,7 +23,9 @@ cimport numpy as np
 from .fused_numerics cimport np_real_numeric, np_floats
 
 cdef inline Py_ssize_t round(np_floats r) nogil:
-    return <Py_ssize_t>((r + 0.5) if (r > 0.0) else (r - 0.5))
+    return <Py_ssize_t>(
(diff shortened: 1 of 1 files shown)


hmaarrfk:
Ginius!!

emmanuelle:
I also saw a slight improvement for float 64, I don't know if this is expected? When I run the benchmarks on my computer, the result of the timeit can fluctuate quite widely (more then 10%) but overall there is a net gain, even if it's just 10% (don't take it badly @Mostafa-Alaa we're happy to gain 10% but it's always a trade-off between code readability and speed improvement. I think here the code is a little more verbose but it's ok).

emmanuelle:
Could you please just add comments so that we know why the casting is done here? Then I'll merge.

Mostafa-Alaa:
@emmanuelle Thank you for your interest, I think the float64 improvement is just fluctuations in the results, it should have the same behaviour as before it would only benefit float32 ( and maybe float16 in the future ).

I completely agree with the readability vs speed trade-off but since those functions are critical and called maybe millions of times that the difference starts to add up and make an impact.

As for the comments, sure I'll add them right away.

hmaarrfk:
> Cython should fix

Does C treat 0.5 * float32 as a float32?

emmanuelle:
Thank you @Mostafa-Alaa ! Merging.

jni:
@hmaarrfk 

> Does C treat 0.5 * float32 as a float32? 

Good question! My C is super rusty (programming language pun totally not intended nor sensical in this context =P), but it seems C happily casts between the two types willy-nilly:

```c
#include <stdio.h>
#include <math.h>

int main(void) {
   float x = pow(2.0, 24) + 1;
   double y = 0.5 * x;
   float z = 0.5 * x;
   printf("x = %f\n", x);
   printf("y = %f\n", y);
   printf("z = %f\n", z);
}
```

This just works. I confirmed that x is single precision because it is an even number, and y and z are equal. This compiles without any complaints with `gcc -Wall -O3 float.c -o float`. 

@Mostafa-Alaa your sleuthing is even more impressive now that I see the C compiler provides essentially no information that this is happening! (Unless my minimal example above is totally flawed, which is entirely possible...) It's also unclear *why* in the code changed in this PR there should be a float64 cast, since the return value is float32 and every non-literal in that expression is float32.

Mostafa-Alaa:
> I feel like this is something that Cython should fix?
> @Mostafa-Alaa do you think you could open an issue in https://github.com/cython/cython for this?

I definitely think that Cython should fix that, in C you could add a suffix to the literal to statically determine its precision like (0.5f for float32), but it's not available in Cython, but I think it would be harder to add for fused types.

But worth giving a shot anyway !

Mostafa-Alaa:
@hmaarrfk @jni 
> 
> > Does C treat 0.5 * float32 as a float32?
> 
> but it seems C happily casts between the two types willy-nilly:

> It's also unclear _why_ in the code changed in this PR there should be a float64 cast, since the return value is float32 and every non-literal in that expression is float32.

Actually according to the C standard there should be no problem in your code, unfortunately that's how C works if you have one operand of type float64 (which is the default for the literals) and a 100 of float32 the compiler might promote the 100 operands for the sake of better accuracy, but in our case the user is already willing to sacrifice some precision for some speed, maybe Cython could have a  chance with that !

PR #4368, opened 2019-12-19
clarify register_translation example description

Evidlo:
## Description

The `register_translation()` example [here](https://scikit-image.org/docs/dev/auto_examples/transform/plot_register_translation.html#sphx-glr-auto-examples-transform-plot-register-translation-py) claims to use phase correlation when it is actually using cross correlation implemented via FFTs.  

Phase correlation yields different results and is not simply an alternative implementation of cross correlation. (see [here](https://dsp.stackexchange.com/questions/31919/phase-correlation-vs-normalized-cross-correlation) for a comparison).

The relevant lines in `register_translation.py` are [here](https://github.com/scikit-image/scikit-image/blob/b1ed36a33be0a8da4be985aa89e24b37e5a34c76/doc/examples/transform/plot_register_translation.py#L51-L52).  Note the absence of elementwise normalization.

## For reviewers

  later.

### DIFF
--- a/doc/examples/transform/plot_register_translation.py
+++ b/doc/examples/transform/plot_register_translation.py
@@ -1,9 +1,9 @@
 """
 =====================================
-Cross-Correlation (Phase Correlation)
+Image Registration
 =====================================
 
-In this example, we use phase correlation to identify the relative shift
+In this example, we use cross-correlation to identify the relative shift
 between two similar-sized images.
 
 The ``register_translation`` function uses cross-correlation in Fourier space,


jni:
@Evidlo thanks for that! On first reading your fix seems correct to me, but I'd like to take the time to understand it all properly before merging. I'll try to get to it in the coming days, but maybe someone else in the team can have a look in the meantime. (Though as you can imagine it is a busy time of year, so please bear with us!)

Evidlo:
Also a relevant excerpt from the paper

>The three algorithms can be straightforwardly extended to include variations of the cross-correlation method, such as phase correlation, and to include window functions to account for edge effects [13].

emmanuelle:
Thank you @Evidlo for identifying this. It seems indeed that we are not using one of the classical methods (normalized cross-correlation or phase correlation). Not normalizing the cross-correlation can give more weight to high-intensity regions of the image, which is suboptimal.

emmanuelle:
@scikit-image/core we should reach a decision for this one (for me it's a no-brainer).

alexdesiqueira:
Thank you for your work @Evidlo, nice catch!

Issue #4371, opened 2019-12-25
Drop scipy<1?


hmaarrfk:
## Description

Following NEP29, I think we should drop scipy < 1.0.

I used my [NEP29 calculator](https://gist.github.com/hmaarrfk/0e2a01d57f07f55eb99d177da92691b6) find that if we released today, the oldest version of scipy we should support is about 1.0.1.

This would help at least 1 PR have cleaner code IMO.

```
[('1.4.1', datetime.datetime(2019, 12, 19, 15, 49, 20, 447964)),
 ('1.4.0', datetime.datetime(2019, 12, 16, 23, 0, 54, 761686)),
 ('1.3.3', datetime.datetime(2019, 11, 23, 21, 16, 19, 515201)),
 ('1.3.2', datetime.datetime(2019, 11, 9, 21, 30, 52, 395905)),
 ('1.3.1', datetime.datetime(2019, 8, 9, 3, 33, 23, 839598)),
 ('1.2.2', datetime.datetime(2019, 6, 7, 1, 19, 15, 77293)),
 ('1.3.0', datetime.datetime(2019, 5, 17, 18, 21, 36, 442178)),
 ('1.2.1', datetime.datetime(2019, 2, 9, 5, 27, 19, 728046)),
 ('1.2.0', datetime.datetime(2018, 12, 18, 3, 30, 14, 284589)),
 ('1.1.0', datetime.datetime(2018, 5, 5, 18, 35, 20, 885004)),
 ('1.0.1', datetime.datetime(2018, 3, 24, 19, 39, 46, 16659))]
```

https://github.com/scikit-image/scikit-image/pull/4329

jni:
@hmaarrfk 👍 from me.

Issue #4369, opened 2019-12-24
macOS builds failing


jni:
## Description

Could be related to the recent NumPy 1.18 release, or just an update to Travis compilers, not sure. Here's an example failure from #4356:

https://travis-ci.org/scikit-image/scikit-image/jobs/628839752#L2728-L2730

```
  g++: skimage/feature/_cascade.cpp
2729  error: invalid argument '-std=c99' not allowed with 'C++'
2730  error: Command "g++ -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -arch x86_64 -g -std=c99 -I/Users/travis/venv/lib/python3.6/site-packages/numpy/core/include -I/Users/travis/venv/lib/python3.6/site-packages/numpy/core/include -I/Library/Frameworks/Python.framework/Versions/3.6/include/python3.6m -c skimage/feature/_cascade.cpp -o build/temp.macosx-10.9-x86_64-3.6/skimage/feature/_cascade.o -MMD -MF build/temp.macosx-10.9-x86_64-3.6/skimage/feature/_cascade.o.d" failed with exit status 1
2731
```

hmaarrfk:
I don't understand where that compilation flag is being added. It really isn't obvious where the fix should go. In our scripts, we never specify c99

hmaarrfk:
It seems related to this new feature
https://github.com/numpy/numpy/blob/4f972297277e09f41f315fd75510cb313926fe67/doc/source/release/1.18.0-notes.rst#stdc99-added-if-compiler-is-named-gcc

hmaarrfk:
https://github.com/numpy/numpy/blob/master/numpy/distutils/ccompiler.py#L538

hmaarrfk:
@mattip any clue how to get around this? I guess we were using gcc on OSX, and the build flags weren't really happy.

mattip:
The check specifically checks for `gcc`, how is it being changed to `g++`?

mattip:
Should be fixed in numpy/numpy#15194

PR #4372, opened 2019-12-26
Bump scipy minimum version to 1.0.1

hmaarrfk:
## Description
Fixes https://github.com/scikit-image/scikit-image/issues/4371

This supports 2 years of scipy from today.
Rather conservative, but gets us to the 1.0 mark, which I'm really excited about.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,5 +1,5 @@
 numpy>=1.14.1
-scipy>=0.19.0
+scipy>=1.0.1
 matplotlib>=2.0.0,!=3.0.0
 networkx>=2.0
 pillow>=4.3.0


hmaarrfk:
Yeah, i went through the other dependencies, and some of them would be a nuisance to keep updating, especially since we need to "predict" when the new release would be. 

A large challenge exists when projects go a LONG time between releases (PyWaelets)

jni:
@hmaarrfk no need to predict anything. We only update things from master on the date that the policy "expires", and releases happen when they happen. We should *not* pre-emptively remove things based on when we expect the next release.

jni:
Failures are due to #4369, so I'm merging! Thanks @hmaarrfk, this is indeed exciting!

PR #4374, opened 2019-12-27
sphinx gallery configuration (inspect_global_variables)

emmanuelle:
Sphinx-gallery 0.5 has introduced a new feature: it introduces links also for class instances (see the note block in https://sphinx-gallery.github.io/stable/configuration.html#references-to-examples). This results in very busy gallery pages with a lot of links, since we have intersphinx enabled for numpy and matplotlib so most lines are bright orange :-) (see for example https://scikit-image.org/docs/dev/auto_examples/color_exposure/plot_rgb_to_hsv.html).

This PR reverts this behaviour (I think it'd be better, but let's discuss about it).

Relevant links:
- the issue I opened on sphinx-gallery https://github.com/sphinx-gallery/sphinx-gallery/issues/580
- scikit-learn's corresponding issue and PR https://github.com/scikit-learn/scikit-learn/issues/15812 and https://github.com/scikit-learn/scikit-learn/pull/15844

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -56,6 +56,7 @@ sphinx_gallery_conf = {
     'gallery_dirs': 'auto_examples',
     'backreferences_dir': 'api',
     'reference_url': {'skimage': None},
+    'inspect_global_variables'  : False,
     'subsection_order': ExplicitOrder([
         '../examples/data',
         '../examples/numpy_operations',


jni:
Oh really? I really like this! Linking to the Python 3.7 docs for "float" is a bit much, but the matplotlib API links I find super useful! Not to mention e.g. sklearn, pandas, and seaborn APIs when we start using those more in examples.

emmanuelle:
So we already had links for matplotlib functions and classes but now this change in sphinx-gallery default behaviour also create links for instances -- do you confirm that you like this behaviour? We could remove Python (standard library) from our intersphinx links so that we do not have links for numerical types. But for example for numpy arrays since all images will be arrays we will have a lot of links. 

Compare for example https://scikit-image.org/docs/dev/auto_examples/color_exposure/plot_rgb_to_hsv.html and https://scikit-image.org/docs/stable/auto_examples/color_exposure/plot_rgb_to_hsv.html (dev and stable). 

The good news is that the sphinx-gallery is currently discussing options for more fine-tuning.

jni:
Yeah I'm in two minds about this! It's actually nice to be able to click on the `ax...` calls and get taken to the matplotlib.Axes documentation. Currently one would have to go to the plt.subplots call, then follow the matplotlib documentation through there. I see your point though and could go either way — let's see what others think!

emmanuelle:
As a possible trade-off, we could remove Python from our intersphinx links, I don't think we use it a lot and then it will remove the links for all the numerical variables, the tuples, etc.

emmanuelle:
See https://github.com/sphinx-gallery/sphinx-gallery/pull/581, we should not hurry to merge this one.

soupault:
I agree with Juan that having documentation links for instances looks convenient, at least, for matplotlib and pandas. At the same time, visual highlighting of the corresponding snippets really pollutes the examples. From my perspective, having the old behaviour (https://scikit-image.org/docs/stable/auto_examples/color_exposure/plot_rgb_to_hsv.html) plus the links for instances (visually as a plain text) would be the best choice.

>we could remove Python from our intersphinx links

This will also remove the links for `range`, `open`, `OrderedDict`, etc, won't it?

emmanuelle:
> This will also remove the links for range, open, OrderedDict, etc, won't it?

Yes, it would. But maybe we can decide that the goal of scikit-image gallery examples is not to teach the Python language so much?

lagru:
I'd actually prefer if there was no hyperlink formatting whatsoever. 

Proposing a different solution: we could patch the CSS to suppress hyperlink coloring for code blocks. In that case the default syntax coloring would be preserved while still providing convenient back links. I'd guess a simple `color: unset` with the right selector at the right place would do the trick. It would be similar to what sphinx uses when linking back from included source code (if that option is enabled).

soupault:
@lagru see the linked issue - https://github.com/sphinx-gallery/sphinx-gallery/pull/581. :)

Actually, I've just realized that currently all the imports are highlighted (incl. e.g. matplotlib `plt.`). Maybe a better way would be to keep the distinct colors only for skimage imports and, potentially, instances...

emmanuelle:
closing this PR since https://github.com/sphinx-gallery/sphinx-gallery/pull/581 has been merged into sphinx-gallery. Always better to take the upstream solution ;-).

PR #1780, opened 2015-11-10
maximum Feret diameter in regionprops

stevendbrown:
Added maximum Feret diameter of a region's convex hull. There is an alternative implementation in which only lines passing through the center of mass are considered, but my understanding is that the Feret diameter has only been evaluated in the context of convex objects. Here's hoping that someone here knows more than me on this subject!

PR #4379, opened 2019-12-30
Feret diameter

emmanuelle:
Supersedes #1780 by @stevendbrown (fixed the merging conflicts). Please see the discussion in #1780, an extensive review already took place.


emmanuelle:
Compared to the original PR I made the property 3d.

stevendbrown:
@emmanuelle This is awesome! Thank you!

emmanuelle:
Thank you @stevendbrown :-). The hard work was yours, I just used the 3d method for computing surfaces. Please comment whether the 3d implementation looks correct (I ran a few tests and it looked ok to me).

skimage/measure/_regionprops.py
```diff
@@ -227,6 +233,20 @@ def euler_number(self):
     def extent(self):
         return self.area / self.image.size
 
+    @property
+    def feret_diameter(self):
+        from ..morphology.convex_hull import convex_hull_image
+        label_image = self._label_image
+        label = self.label
+        identity_convex_hull = convex_hull_image(label_image == label)
```
jni: @emmanuelle this is much more quickly implemented with the `.image` property, which contains only the object in question, already in bool form. (It uses `props.slice` to evaluate only in the region area.)

skimage/measure/tests/test_regionprops.py
```diff
@@ -71,6 +71,22 @@ def test_ndim():
...
+    comparator_result = 17.029386365926403
+    test_result = regionprops(SAMPLE)[0].feret_diameter
+    assert_almost_equal(comparator_result, test_result)
+    # square, test that Feret diameter is sqrt(2) * square side
+    img = np.zeros((20, 20), dtype=np.uint8)
+    img[2:-2, 2:-2] = 1
+    feret_diameter = regionprops(img)[0].feret_diameter
+    assert np.abs(feret_diameter - 16 * np.sqrt(2)) < 1
+    # 3d example
+    img_3d = np.dstack((img,) * 3)
```
jni: @emmanuelle can you make this a separate test?

skimage/measure/_regionprops.py
```diff
@@ -227,6 +233,19 @@ def euler_number(self):
     def extent(self):
         return self.area / self.image.size
 
+    @property
+    def feret_diameter(self):
+        from ..morphology.convex_hull import convex_hull_image
+        identity_convex_hull = np.pad(self.convex_image,
+                                      2, mode='constant', constant_values=0)
```
jni: @emmanuelle is this padding really necessary??? This points to a bug in `find_contours`, no? Or...?

I also wonder whether we should skip convex hull computation in 3D, where it is quite expensive. It might actually be faster to use the coordinates from marching cubes directly.

emmanuelle: I don't know if it's a bug, but yes `find_contours(np.ones((5, 5)), 0.5)` gives an error (I guess because it's not possible to decide what is outside). I add to add this padding otherwise there can be errors for an image full of ones.

For the 3d case, do you mean retrieving coordinates of the surface from marching cubes, and then compute the convex hull of these points?

stevendbrown: A region can be a rectangle which results in a `self.convex_image` which is all ones, correct? To prevent problems with `find_contours` in that case, the padding is definitely necessary, I think.

skimage/measure/tests/test_regionprops.py
stevendbrown: @emmanuelle The test here looks like it'll stack up and leave the 1 region without boundaries on the third axis. Could you instead do the following?

```
img_3d = np.zeros((20, 20, 3), dtype=np.uint8)
img_3d[2:-2, 2:-2, 1] = 1
```

skimage/measure/_regionprops.py
jni: @emmanuelle 

> find_contours(np.ones((5, 5)), 0.5) gives an error (I guess because it's not possible to decide what is outside)

I feel like the contour of the image should be the contour then! (ie the implicit padding should be 0 in this case). I'll raise an issue, hopefully we can fix this is the future and get rid of this padding.

> For the 3d case, do you mean retrieving coordinates of the surface from marching cubes, and then compute the convex hull of these points?

I didn't even think that computing the hull would be necessary, as the number of points returned by marching cubes might itself be reasonably small, but yes! =) Ideally, compare the speed of the three methods (convex hull -> marching cubes -> distances, marching cubes -> distances, marching cubes -> convex hull -> distances) for some dummy data (e.g. the binary blobs) and see whether there are any clear winners...

skimage/measure/tests/test_regionprops.py
stevendbrown: Sorry, strike this comment...I've just realized that the 3D case won't be subject to the same preparatory transformations as the 2D case! 😳

skimage/measure/_regionprops.py
emmanuelle: I agree with the implicit padding for marching cubes.

For the convex hull, the Feret diameter is defined on the convex hull.

rfezzani:
The circular import pointed out by @stevendbrown https://github.com/scikit-image/scikit-image/pull/1780#discussion_r44736830 is fixed in #4400

skimage/measure/_regionprops.py
jni: @emmanuelle 

> For the convex hull, the Feret diameter is defined on the convex hull.

The feret diameter is defined for any object, the convex hull just makes it faster to compute since the largest diameter must join two vertices of the hull. But the question is that whether "hull + feret on hull vertices" is faster than "feret on all vertices". I guess the former is faster for most data given existing implementations.

jni:
@scikit-image/core this is ready to merge as far as I'm concerned — my performance questions are theoretical and not rooted in measurements. I suggest we get it right first (this PR) and address performance concerns as needed in future PRs.

skimage/measure/_regionprops.py
```diff
@@ -227,6 +233,19 @@ def euler_number(self):
     def extent(self):
         return self.area / self.image.size
 
+    @property
+    def feret_diameter(self):
+        from ..morphology.convex_hull import convex_hull_image
```
rfezzani, line 238: Is this import used anywhere here?
```suggestion
```

jni:
@rfezzani I've committed your suggestion, so this should be ready to go...!u

sciunto:
I changed the title to that it makes sense when we will make a new release ;)

PR #4005, opened 2019-07-11
[WIP] Hausdorff Distance (new)

clementkng:
## Description
Supersedes gh-742. 

This is my attempt to fix the feedback on @josteinbf 's Hausdorff distance PR and add docs, examples, and extra tests. I've changed the methods to a single `hausdorff_distance` method to be more clear.

Existing discussion at gh-738. WIP b/c I still need to add benchmarks and gallery examples.

Pinging the original people that still had feedback open on the original PR: @stefanv @tonysyu @ahojnnes @emmanuelle @jni @alexandrejaguar 

## Checklist

- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [x] Gallery example in `./doc/examples` (new features only)
- [x] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
- [x] Unit tests
- [x] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

## Gallery Example Output
image

## For reviewers

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

PR #742, opened 2013-09-29
Hausdorff distance

josteinbf:
This PR implements the Hausdorff distance. Docs, examples and tests are incomplete, but the functionality seems to be ok.

Some discussion has already taken place here: #738.

Ping @jni @JDWarner

Issue #738, opened 2013-09-26
Lack of symmetry in find_boundaries?


josteinbf:
I'm trying to use the find_boundaries function, and to figure out what output to expect, I tested it on an array with only a single pixel being ``True`` and the rest being ``False``:

```python
import numpy as np
from skimage.segmentation import find_boundaries

a = np.zeros((5, 5), dtype='bool')
a[2, 2] = True

boundary = find_boundaries(a)
print(boundary)
```
The output was the following:
```python
[[False False False False False]
 [False False False False False]
 [False False  True  True False]
 [False False  True False False]
 [False False False False False]]
```

I was surprised by this output, as it seems to me that it is less symmetric than it ought to be. I was expecting either just a single pixel being ``True``, or perhaps a 2x2 block being ``True``. I decided to test the symmetry a bit further by finding the boundary of the same array after flipping it:

```python
boundary_inverted = find_boundaries(a[::-1, ::-1])[::-1, ::-1]
print(boundary_inverted)
print(np.all(boundary == boundary_inverted))
```
The output was
```python
[[False False False False False]
 [False False  True False False]
 [False  True  True False False]
 [False False False False False]
 [False False False False False]]
False
```

I'm not sure if this is a bug, but it may well be. The docs could certainly be improved, though, as I cannot understand why the result is what it is.

Can anyone please help me clarify what to expect here?

PR #4382, opened 2019-12-31
Hausdorff Distance (updated)

clementkng:
## Description
Supersedes gh-742 and gh-4005

This is my attempt to fix the feedback on @josteinbf 's Hausdorff distance PR and add docs, examples, and extra tests. I've changed the methods to a single `hausdorff_distance` method to be more clear.

Existing discussion at gh-738 and gh-4005.

Hausdorff distance can be used to determine the degree of resemblance two objects have between each other when superimposed on top of each other, for example in this [paper](https://people.eecs.berkeley.edu/~malik/cs294/Huttenlocher93.pdf).

## Gallery Example Output
image

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/metrics/_set_metrics.pyx
```diff
@@ -0,0 +1,52 @@
...
+#cython: nonecheck=False
+#cython: wraparound=False
+
+cimport numpy as cnp
+from libc.math cimport sqrt
+from libc.float cimport DBL_MAX
+
+import numpy as np
+
+def hausdorff_distance_onesided(cnp.float64_t[:, ::1] points_sup,
```
emmanuelle: One of my pet comments here: in terms of speed, how would this Cython implementation compare to using a CKDTree, in particular https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.query.html#scipy.spatial.cKDTree.query (the `query` method for one nearest-neighbor, which directly returns the distance)?

My guestimate is that for a small number of points the Cython will be faster, but that for a large number of points the KDTree might win. Are you interested in testing this @clementkng ? If you don't have the time I might give it a try. 

I guess it also depends on the application for which we expect this function to be used. Will it be rather for very sparse images or for more busy ones?

clementkng: Hi @emmanuelle, I can try to compare the cKDTree vs cython, but I'm not super familiar w/ how we'd test this. I'm guessing this is similar to the benchmarking process for the initial implementation of Hausdorff distance?

I expect the function to be used for reasonably busy images, as one of the applications is to find and recognize image shapes within a larger image.

emmanuelle: I would just take a realistic example and see how the two methods compare, and also rescale the image of interest with several factors so that you can see if the fastest method depends on the number of non-zero pixels. Automatic benchmarks with asv are useful but here it's just to decide which method is better.

clementkng: @emmanuelle to ensure reproducibility, should I push the code I use to compare the methods into the benchmark_metrics.py as well?

emmanuelle: not at this stage because if one method is a clear winner we will not bother with the other. What I would do is write a python script calling the function on images of different sizes, and run this script in two different branches (each branch with a different implementation), and paste the script and its result here. But really whatever workflow works for you is fine :-).

lagru: A workflow I have employed successfully in the past uses notebooks to compare two implementations side by side using the "%timeit" magic command. E.g. see this [old gist](https://gist.github.com/lagru/4cd7ab94a101bbea15af216a3d44d1ca) were I imported the new implementation from the PR branch and declared the old implementation inside the notebook to compare.

soupault: I would love to see a comparison of your implementation to https://github.com/deepmind/surface-distance/blob/master/surface_distance/metrics.py#L209 in terms of speed on a pair of 3D images (e.g. 400x400x400).

clementkng: @soupault I'm not sure I've seen the surface-distance library used in an Jupyter notebook. I'm having a little trouble including it in the comparisons, is there a code sample somewhere of how to import the package and use it if possible?

lagru: Are you asking how to install this library? You should be able to import `compute_robust_hausdorff from surface_distance` after following the install instructions in the [readme](https://github.com/deepmind/surface-distance#installation).

clementkng: So something like `from surface_distance import compute_robust_hausdorff` should work once I've installed the package right? I believe I'm having some issues since surface distance is not [Python3 compatible](https://github.com/deepmind/surface-distance/pull/1), but I need Python3 to run/build scikit-image.

clementkng: Here is a [link](https://gist.github.com/clementkng/cf0321e2b62daed21189867187e36751) to a first draft of benchmarks against cKD tree.

lagru: @clementkng From the issue you linked it looks like only that one import must be updated to be compatible with Python 3? You might succeed by applying that change to your local copy. 

Thanks for the notebook! I've left a suggestion to make the comparison more meaningful.

emmanuelle:
Thanks a lot @clementkng for this thorough benchmarks, it's a lot of work. So it seems that for a 256x256 image the cKDTree method is faster, this is the intuition that I had (cKDTree should be faster when images are large). Most images will be larger than 256^2, so I'd vote for the cKDTree, Thoughts from @lagru, @soupault or others? @clementkng would you still have the energy to make this change :-) ?

jni:
@emmanuelle I agree that we should use the cKDTree version by default. Also, did you see my comment: it might be interesting to vendor the Cython cKDTree code so that we don't pay the function call overhead there. That could potentially accelerate the function even much more!

lagru:
> it might be interesting to vendor the Cython cKDTree code

Definitely worth looking into! With this PR and #4165 we'd have at least two algorithms that might benefit. Although that should definitely happen in a separate PR and involve some actual benchmarks that demonstrate a significant speed-up (I strongly suspect that there will be one). Beforehand we should compare scipy's cKDTree with [sklearn's KDTree implementation](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/neighbors/_kd_tree.pyx) and others if there are any.

emmanuelle:
@jni I saw you comment and commented too ;-). We might get away with a faster solution by calling `tree.query` on a whole array, then no need to vendor anything.

jni:
@emmanuelle very nice catch! The new numbers are eye-popping! 🙌

emmanuelle:
ok great. And thank you @clementkng for sticking with us and our experiment ideas ;-)/

emmanuelle:
@clementkng any chance we can get this one into 0.17? (the release should be in one or two weeks)

emmanuelle:
@clementkng can we help you with something to finish this PR :-) ?

clementkng:
> @clementkng can we help you with something to finish this PR :-) ?
Hi @emmanuelle and @rfezzani, I realized I accidentally assumed the PR was over when the benchmarks proved that the scipy.cKDTree significantly outperformed this implementation, so I'm not sure what finishing this PR would mean. Sorry for the late response and misunderstanding! Based on the benchmark, should I start using scipy.cKDTree in my implementation? What's the best way to proceed? 😅

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+    if a_points.shape[0] == 0 or b_points.shape[0] == 0:
+        if a_points.shape[0] == b_points.shape[0]:
+            # Both sets are empty and thus the distance is zero
+            return 0.
+        else:
+            # Exactly one set is empty; the distance is infinite
+            return np.inf
+
+    a_points = np.require(a_points, np.float64, ['C'])
+    b_points = np.require(b_points, np.float64, ['C'])
```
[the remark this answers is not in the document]
clementkng: This line converts `a_points` and `b_points` from int64 to float64

skimage/metrics/_set_metrics.pyx
```diff
@@ -0,0 +1,52 @@
...
+
+    Returns
+    -------
+    distance : float
+        The Hausdorff one-sided distance between sets ``points_sup`` and
+        ``points_inf``, using Euclidean distance to calculate the distance
+        between points in ``points_sup`` and ``points_inf``.
+
+    """
+    assert points_sup.shape[1] == points_inf.shape[1]
```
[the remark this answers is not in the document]
clementkng: Done

skimage/metrics/tests/test_set_metrics.py
```diff
@@ -0,0 +1,141 @@
...
+    coords_a = np.zeros(shape, dtype=np.bool)
+    coords_b = np.zeros(shape, dtype=np.bool)
+    coords_a[points_a] = True
+    coords_b[points_b] = True
+    distance = np.sqrt(sum((ca - cb) ** 2
+                           for ca, cb in zip(points_a, points_b)))
+    assert_almost_equal(hausdorff_distance(coords_a, coords_b), distance)
+
+
+class TestHausdorffRegionSingle:
```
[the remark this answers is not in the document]
clementkng: Done

clementkng:
So I believe I'll need to rebase to resolve the conflict that is preventing tests from running, but last time I did that I wasn't able to resolve the merge conflict and as a result had to make a new branch and PR (which is this one). Would anyone be willing to help me rebase?

jni:
Hi @clementkng! We usually squash and merge these days when there is a long and tortuous commit history, so you should be good to `git merge master` while on this branch. If you *do* want to rebase, then you can do:

```
git checkout hausdorff-distance-new
git rebase master
git status
# fix any merge conflicts
# use git add to mark resolved files
git rebase --continue
# repeat above until git status is clean
git push origin --force-with-lease
```

skimage/measure/__init__.py
```diff
@@ -48,5 +48,5 @@
            'compare_mse',
            'compare_nrmse',
            'compare_psnr',
-           'shannon_entropy',
-]
+           'shannon_entropy'
+           ]
```
jni, line 52: You might not need a rebase if you revert this change!

skimage/metrics/setup.py
```diff
@@ -0,0 +1,31 @@
...
+                         include_dirs=[get_numpy_include_dirs(), '../shared'])
+
+    return config
+
+
+if __name__ == '__main__':
+    from numpy.distutils.core import setup
+    setup(maintainer='scikit-image Developers',
+          maintainer_email='scikit-image@python.org',
+          description='Graph-based Image-processing Algorithms',
```
jni, line 27: Description should be changed, e.g. "Metrics to compare images"

jni:
@clementkng I had to go over the code and review the history to remind myself of the current status.

> I realized I accidentally assumed the PR was over when the benchmarks proved that the scipy.cKDTree significantly outperformed this implementation, so I'm not sure what finishing this PR would mean. Sorry for the late response and misunderstanding! Based on the benchmark, should I start using scipy.cKDTree in my implementation? What's the best way to proceed?

Yes, I think the idea is to replace `hausdorff_distance_onesided` in this PR with the cKDTree implementation. Then it'll be ready to merge!

skimage/metrics/setup.py
clementkng: Done

clementkng:
Thanks for the feedback @jni 😄

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+
+from ._set_metrics import hausdorff_distance_onesided
+
+
+def hausdorff_distance(a, b):
```
jni, line 6: I think we can probably do better with these parameter names. For images we use `image0, image1`, or `image_true, image_test` so here I think we should use `points0, points1` (my preference) or `points_true, points_test`, or even `coords0, coords1`. Any thoughts on this @clementkng @scikit-image/core ?

jni: ... Actually, I just saw that these are images, not actual points! So the docstring must be updated (suggestion below) and I would call these `image0` and `image1`.

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+    a, b : ndarray, dtype=bool
+        Arrays where ``True`` represents a point that is included in a
+        set of points. Both arrays must have the same shape.
+
+    Returns
+    -------
+    distance : float
+        The Hausdorff distance between sets ``a`` and ``b``, using
+        Euclidian distance to calculate the distance between points in ``a``
+        and ``b``.
```
jni, line 24:
```suggestion
        The Hausdorff distance between coordinates of nonzero pixels in
        ``image0`` and ``image1``, using the Euclidian distance.
```

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+def hausdorff_distance(a, b):
+    """
+    Calculate the Hausdorff distance [1]_ between two sets of points.
+
+    The Hausdorff distance is the maximum distance between any point on
+    ``a`` and its nearest point on ``b``, and vice-versa.
+
+    Parameters
+    ----------
+    a, b : ndarray, dtype=bool
```
jni, line 15: I'd remove the dtype specification, since this works totally fine for integer arrays. It could be used to compare two segmentations, for example.

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+
+from ._set_metrics import hausdorff_distance_onesided
+
+
+def hausdorff_distance(a, b):
+    """
+    Calculate the Hausdorff distance [1]_ between two sets of points.
+
+    The Hausdorff distance is the maximum distance between any point on
```
jni, line 10:
```suggestion
    """
    Calculate the Hausdorff distance between nonzero elements of given images.

    The Hausdorff distance [1]_ is the maximum distance between any point on
```

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+    >>> coords_a[points_a] = True
+    >>> coords_b[points_b] = True
+    >>> hausdorff_distance(coords_a, coords_b)
+    3.0
+
+    """
+    if a.dtype != np.bool or b.dtype != np.bool:
+        raise ValueError('Arrays must have dtype = \'bool\'')
+    if a.shape != b.shape:
+        raise ValueError('Array shapes must be identical')
```
jni, line 46: Personally, I would remove these checks.

jni: (certainly the first one!)

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,63 @@
...
+    b_points = np.transpose(np.nonzero(b))
+
+    # Handle empty sets properly
+    if a_points.shape[0] == 0 or b_points.shape[0] == 0:
+        if a_points.shape[0] == b_points.shape[0]:
+            # Both sets are empty and thus the distance is zero
+            return 0.
+        else:
+            # Exactly one set is empty; the distance is infinite
+            return np.inf
```
jni, line 58: Here I would use `len` instead of `.shape[0]`, as that is more readable as "if there are zero points".

skimage/metrics/set_metrics.py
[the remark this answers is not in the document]
jni: Personally I would solve that by adding `.astype(np.float64)` to the creation calls...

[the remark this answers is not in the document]
clementkng: I still need a C-contiguous array for the Cython, so I'll use `np.ascontiguousarray(a_points, dtype=np.float64)`. How does that sound?

skimage/metrics/set_metrics.py
clementkng: Done

skimage/metrics/set_metrics.py
clementkng: Done

skimage/metrics/set_metrics.py
clementkng: Done

skimage/metrics/set_metrics.py
clementkng: Done

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+
+from ._set_metrics import hausdorff_distance_onesided
+
+
+def hausdorff_distance(image0, image1):
+    """
+    Calculate the Hausdorff distance between nonzero elements of given images.
```
jni, line 8: The docstring should be on the same line as the quotes (many UIs rely on this to render the first line as per PEP257).

```suggestion
    """Calculate the Hausdorff distance between nonzero elements of given images.
```

clementkng: Done

jni:
@emmanuelle do you have time to review and 🤞 merge today?

jni:
ping @scikit-image/core anyone have time to review/potentially merge this PR?

clementkng:
@rfezzani Why would `hausdorff_distance_onesided` need to be a hidden function?

skimage/metrics/setup.py
```diff
@@ -0,0 +1,31 @@
...
+
+import os
+base_path = os.path.abspath(os.path.dirname(__file__))
+
+
+def configuration(parent_package='', top_path=None):
+    from numpy.distutils.misc_util import Configuration, get_numpy_include_dirs
+
+    config = Configuration('metrics', parent_package, top_path)
+    config.add_data_dir('tests')
```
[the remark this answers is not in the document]
clementkng: Removed

skimage/metrics/set_metrics.py
```diff
@@ -0,0 +1,56 @@
...
+    b_points = np.transpose(np.nonzero(image1))
+
+    # Handle empty sets properly
+    if len(a_points) == 0 or len(b_points) == 0:
+        if len(a_points) == len(b_points):
+            # Both sets are empty and thus the distance is zero
+            return 0.
+        else:
+            # Exactly one set is empty; the distance is infinite
+            return np.inf
```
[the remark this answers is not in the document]
clementkng: Done

skimage/setup.py
```diff
@@ -18,6 +18,7 @@ def configuration(parent_package='', top_path=None):
     config.add_subpackage('graph')
     config.add_subpackage('io')
     config.add_subpackage('measure')
+    config.add_subpackage('metrics')
```
rfezzani, line 21: @jni, is this necessary? Because I just noticed that the `registration` subpackage is not added to config :/ ...

rfezzani:
> @rfezzani Why would `hausdorff_distance_onesided` need to be a hidden function?

Because it was just used internally by the `hausdorff_distance` function, but not defining the `hausdorff_distance_onesided` function as you did is a good option too ;)

skimage/setup.py
jni: This is needed when there is Cython to compile, otherwise it is not needed, but doesn't hurt anything.

jni:
🎉 Thank you for your undeterred patience, @clementkng! It's so awesome to have this in!

rfezzani:
:tada: thank you @clementkng!

clementkng:
@jni @rfezzani  thank you for your reviews! I'm glad I was able to revive this PR after being gone for so long.

PR #4392, opened 2020-01-10
setup: read long description from README

stefanv:
Our description on PyPi is woefully short, but they do allow rendering Markdown.  So, this change will update the page at https://pypi.org/project/scikit-image/ on the next upload.


stefanv:
Failure seems unrelated.

emmanuelle:
Thank you, merging!

PR #4396, opened 2020-01-15
Add border conditions to ridge filters

### LINKED ISSUE #4394: meijering filter artefacts around the image
## Description

`maijering` filter suffers from the same artifacts as `frangi` described in #4249 .

## Way to reproduce

```python
import pylab as plt
from skimage import color, data
from skimage.filters import meijering
img = color.rgb2gray(data.retina())[300:500, 700:900]
out = meijering(img, sigmas=[1])
plt.imshow(out)
```

meijerin

Managing border conditions for the ridges filters as suggested by @shenker https://github.com/scikit-image/scikit-image/pull/4343#pullrequestreview-337754448 may solve the problem.

## Version information
```
[GCC 7.3.0]
Linux-4.4.0-171-generic-x86_64-with-debian-stretch-sid
scikit-image version: 0.17.dev0
numpy version: 1.17.4
```

rfezzani:
## Description

Closes #4394.

- Border condition handling is added to the `meijering`, `sato` and `hessian` filters,
- border artifact in `meijering` are removed using the 'reflect' mode,
- implicit 'constant' mode used in `sato` and `hessian` are deprecated in favor of 'reflect' mode.

This PR comes from a suggestion of @shenker https://github.com/scikit-image/scikit-image/pull/4343#pullrequestreview-337754448.

## For reviewers

  later.

### DIFF
--- a/skimage/filters/ridges.py
+++ b/skimage/filters/ridges.py
@@ -478,7 +478,7 @@ def frangi(image, sigmas=range(1, 10, 2), scale_range=None,
 
 def hessian(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
             beta1=None, beta2=None, alpha=0.5, beta=0.5, gamma=15,
-            black_ridges=True, mode='reflect', cval=0):
+            black_ridges=True, mode='constant', cval=0):
     """Filter an image with the Hybrid Hessian filter.
 
     This filter can be used to detect continuous edges, e.g. vessels,
--- a/skimage/filters/ridges.py
+++ b/skimage/filters/ridges.py
@@ -478,7 +478,7 @@ def frangi(image, sigmas=range(1, 10, 2), scale_range=None,
 
 def hessian(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
             beta1=None, beta2=None, alpha=0.5, beta=0.5, gamma=15,
-            black_ridges=True):
+            black_ridges=True, mode='reflect', cval=0):
     """Filter an image with the Hybrid Hessian filter.
 
     This filter can be used to detect continuous edges, e.g. vessels,
@@ -538,7 +538,7 @@ def hessian(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
     filtered = frangi(image, sigmas=sigmas, scale_range=scale_range,
                       scale_step=scale_step, beta1=beta1, beta2=beta2,
                       alpha=alpha, beta=beta, gamma=gamma,
-                      black_ridges=black_ridges)
+                      black_ridges=black_ridges, mode=mode, cval=cval)
 
     filtered[filtered <= 0] = 1
     return filtered
--- a/skimage/filters/ridges.py
+++ b/skimage/filters/ridges.py
@@ -140,7 +140,7 @@ def compute_hessian_eigenvalues(image, sigma, sorting='none',
 
 
 def meijering(image, sigmas=range(1, 10, 2), alpha=None,
-              black_ridges=True):
+              black_ridges=True, mode='reflect', cval=0):
     """
     Filter an image with the Meijering neuriteness filter.
 
@@ -163,6 +163,11 @@ def meijering(image, sigmas=range(1, 10, 2), alpha=None,
     black_ridges : boolean, optional
         When True (the default), the filter detects black ridges; when
         False, it detects white ridges.
+    mode : {'constant', 'reflect', 'wrap', 'nearest', 'mirror'}, optional
+        How to handle values outside the image borders.
+    cval : float, optional
+        Used in conjunction with mode 'constant', the value outside
+        the image boundaries.
 
     Returns
     -------
@@ -208,7 +213,8 @@ def meijering(image, sigmas=range(1, 10, 2), alpha=None,
     for i, sigma in enumerate(sigmas):
 
         # Calculate (sorted) eigenvalues
-        eigenvalues = compute_hessian_eigenvalues(image, sigma, sorting='abs')
+        eigenvalues = compute_hessian_eigenvalues(image, sigma, sorting='abs',
+                                                  mode=mode, cval=cval)
 
         if ndim > 1:
 
--- a/skimage/filters/ridges.py
+++ b/skimage/filters/ridges.py
@@ -243,7 +243,7 @@ def meijering(image, sigmas=range(1, 10, 2), alpha=None,
 
 
 def sato(image, sigmas=range(1, 10, 2), black_ridges=True,
-         mode='reflect', cval=0):
+         mode='constant', cval=0):
     """
     Filter an image with the Sato tubeness filter.
 
--- a/skimage/filters/ridges.py
+++ b/skimage/filters/ridges.py
@@ -509,6 +509,11 @@ def hessian(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
     black_ridges : boolean, optional
         When True (the default), the filter detects black ridges; when
         False, it detects white ridges.
+    mode : {'constant', 'reflect', 'wrap', 'nearest', 'mirror'}, optional
+        How to handle values outside the image borders.
+    cval : float, optional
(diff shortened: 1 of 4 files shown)


skimage/filters/ridges.py
```diff
@@ -286,6 +298,13 @@ def sato(image, sigmas=range(1, 10, 2), black_ridges=True):
     if np.any(sigmas < 0.0):
         raise ValueError('Sigma values less than zero are not valid')
 
+    if mode is None:
+        warn("'constant' borders handling mode is deprecated and will be "
```
lagru: First sentence here (and similar warning below) sounds a little off but that might just be the non-native speaker in me talking. Either use `'constant' mode is deprecated...` or `'constant' border handling is deprecated ...`? :wink:

jni: I agree that the wording is a bit awkward, but honestly, the constant mode is so terrible for edge filters, that I would call our previous behaviour a bug, and here just leave `mode='reflect'` as the default, then add a note in the release notes API changes that the border mode in certain filters has been updated to remove artifacts, and to use `mode='constant'` to recover the old behaviour.

What do others think here? @scikit-image/core

jni: At any rate, it's not 'constant' border mode that deprecated, since we didn't have a border mode before. The message should say something like "previously, this function implicitly used 'constant' as the border mode when dealing with the edge of the array. The new behavior is 'reflect'. To recover the old behavior, use mode='constant'". Something like that.

rfezzani: I agree, I just added the warning here because `sato` and `hessian` were not suffering from the border artifacts like `frangi` and `meijering` did, so I did not considered that it was a bug. A simple warning like @jni is suggesting looks good to me.

stefanv: +!

jni: Just fwiw, my suggestion here was not to warn, but rather to put this note in the release notes. @stefanv, too much?

stefanv: It feels like that's a bit hidden; I presume few enough people read that, and those who do forget what it said after a few days.

jni: The workflow I was thinking was that essentially everyone would just be fine with this change, but people whose tests would fail would go and look at the release notes and find the culprit, which I've done a few times. But, I'm happy to leave it as a warning for now.

skimage/filters/ridges.py
```diff
@@ -208,7 +213,8 @@ def meijering(image, sigmas=range(1, 10, 2), alpha=None,
     for i, sigma in enumerate(sigmas):
 
         # Calculate (sorted) eigenvalues
-        eigenvalues = compute_hessian_eigenvalues(image, sigma, sorting='abs')
+        eigenvalues = compute_hessian_eigenvalues(image, sigma, sorting='abs',
+                                                  mode=mode, cval=cval)
```
jni: In this case, there will be no warning about the change of behaviour to meijering. Do we want to add this to the release notes under "API changes"?

skimage/filters/ridges.py
```diff
@@ -286,6 +298,14 @@ def sato(image, sigmas=range(1, 10, 2), black_ridges=True):
     if np.any(sigmas < 0.0):
         raise ValueError('Sigma values less than zero are not valid')
 
+    if mode is None:
+        warn("Previously, sato implicitly used 'constant' as the "
+             "border mode when dealing with the edge of the array. The new "
+             "behavior is 'reflect'. To recover the old behavior, use "
+             "mode='constant'. To avoid this warning, please explicitely "
+             "set the mode.", category=FutureWarning, stacklevel=2)
+        mode = 'reflect'
```
jni: Are there any tests where this warning is raised? See `skimage._shared._testing.expected_warnings`.

skimage/filters/ridges.py
rfezzani: Here I considered that it was a bug fix since `meijering` was suffering from artifacts on the image borders, that why I did not raised any warning.
What about adding the API changes for all the ridge filters in the release note?

skimage/filters/ridges.py
rfezzani: I took care of the warning raise by specifying the mode in the tests ;-)

jni: Yes, but you need a test where you check that the warning is correctly raised when no mode is provided. We once got some code in where there was a RuntimeError inside the warning code, because no test covered that code! I think `warn` had not been imported iirc. 😬

rfezzani: Done ;)

rfezzani:
Thank you @jni for your review.

skimage/filters/ridges.py
```diff
@@ -286,6 +298,14 @@ def sato(image, sigmas=range(1, 10, 2), black_ridges=True):
     if np.any(sigmas < 0.0):
         raise ValueError('Sigma values less than zero are not valid')
```
alexdesiqueira: Hey @rfezzani,
since you're still here, could you change this `ValueError` to something more friendly? I would go for `'Sigma values should be greater than zero.'`

skimage/filters/ridges.py
```diff
@@ -522,10 +547,17 @@ def hessian(image, sigmas=range(1, 10, 2), scale_range=None, scale_step=None,
     .. [2] Kroon, D. J.: Hessian based Frangi vesselness filter.
     """
 
+    if mode is None:
+        warn("'constant' borders handling mode is deprecated and will be "
+             "replaced by 'reflect' mode in version 0.19. To avoid this "
+             "warning, please explicitely set the mode argument.",
```
alexdesiqueira: `... please explicitly...`

skimage/filters/ridges.py
alexdesiqueira: >  the constant mode is so terrible for edge filters, that I would call our previous behaviour a bug, and here just leave mode='reflect' as the default, then add a note in the release notes API changes that the border mode in certain filters has been updated to remove artifacts, and to use mode='constant' to recover the old behaviour.

Sorry for not commenting earlier y'all, but I think this warning is too verbose... I'd go for a warning on the API change – since this is kinda of a big deal, shouldn't it be a note at the docstrings instead? I think this is something that would annoy more than inform, as it is.

skimage/filters/ridges.py
rfezzani: Shouldn't `sigma` be strictly positive?

rfezzani: No, I checked, `sigma = 0` is OK, sorry :pray:

skimage/filters/ridges.py
jni: @alexdesiqueira I disagree! We have actually made a conscious effort to move our warnings and errors towards more verbosity. The idea is to contain all the information you need to fix the warning within the warning itself, rather than having to hunt around for a fix. I think that is the least annoying path, and I believe @stefanv agrees.

rfezzani:
Thank you @alexdesiqueira for your review, I added the `_sheck_sigmas` function to modify the `ValueError` message when any sigma value is negative as you suggested in one unique location.

skimage/filters/ridges.py
alexdesiqueira: Could you change the message for `'Sigma values should be equal to or greater than zero.'`, then? I don't like the _tone_ of that message, actually. :blush:

rfezzani: I already modified the message as you suggested in the new `_check_sigmas` function :-)
What about `Sigma values should be positive.`?

alexdesiqueira: Positive means that the zero is not included :blush:
Since zero can be used, something more accurate would be non-negative, in this case. However, what non-negative means is not that clear – the user could think that positive and non-negative mean the same thing, but they don't.
Hence, `'Sigma values should be equal to or greater than zero.'` :smile:

rfezzani: No objection :-).

rfezzani:
:tada: Thx @jni!

Issue #1608, opened 2015-07-14
Consider flattening skimage namespace


jni:
Lately I've been finding the API rather cumbersome. Is canny in features or filters? Is watershed in segmentation or morphology? Where is relabel_sequential?

I suggest importing everything to the root skimage namespace, coupled perhaps with an aliasing convention, e.g. `import skimage as ski` or `import skimage as im`. (Yes, `im` is very common and I'm not advocating for it... But you gotta admit `edges = im.sobel(image)` would be pretty nifty. =)

vighneshbirodkar:
I'm against it. The inconvenience you speak of is very minor and would take only a minute of the programmer's time to overcome. 
On the other hand 
Consider these two snippets

```python
from skiamge import filters
filters.sobel (img)
```

```python
import skimage as ski
ski.sobel (img)
```

The person reading the first snippet, would know about the `filters` package and would find other filtering methods inside it. He might try a few of those to get better results.

The person reading the second piece has to spend more time hunting for filters.

JDWarner:
My main concern is an increasingly crowded namespace, which will only become less intuitive to explore. User's thought process: *Is loading an image load? I'm looking under "L" - nothing - perhaps read - no, nothing in "R" - maybe open - still no. Here it is, weird, it's prefixed with "im"...*

However, if the import time was reasonable, I would be fine with importing all packages by default but leaving them at the various levels. Then someone introspecting `skimage.` **<TAB>** has a better sense of what is available. My sense was that the import time was an issue, though.

Hopefully @stefanv can chime in with the original design goals.

blink1073:
We could add an indexed search function at the top level that will greedily import and check docstrings for a search string, or optionally just search function names if desired.

jni:
I like @blink1073's idea. My point is that the search is very inefficient when it requires switching back to a browser (different workspace, for me), waiting for search results, scrolling to the right part of the API doc... If I could quickly search within IPython I would be happy.

Two other options for consideration:

- `skimage.directory.foo` would contain a string with the full package path to the `foo` function. Then it's just a matter of typing `skimage.directory.rela<tab><enter>` to find out that `relabel_sequential` is in `segmentation` (for some reason) (yes I'm aware I put it there. =P).
- `skimage.all.func` could be an *optional* namespace containing all of skimage.

ahojnnes:
skimage.all / skimage.directory sounds like a good idea to me. Maybe with
the option to output the actual package path.

[...]

PR #4400, opened 2020-01-16
[WIP] Fix the circular import between measure, transform and segmentation modules

rfezzani:
## Description

This PR fixes the circular import between `skimage.measure._regionprops` and `skimage.mormphology.convex_hull` forcing the ugly [import](https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/_regionprops.py#L195) in the `RegionProperties.convex_image` property.

This circular import is also pointed out by @stevendbrown https://github.com/scikit-image/scikit-image/pull/1780#discussion_r44736830.

## For reviewers

  later.


skimage/transform/pyramids.py
```diff
@@ -1,7 +1,7 @@
 import math
 import numpy as np
 from scipy import ndimage as ndi
-from ..transform import resize
+from ._warps import resize
```
jni: This appears to be breaking some CI builds... https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=2282&view=logs&j=bc76316b-1dff-5b05-0ff1-0b85ce9c0f8b&t=7992ab4a-b77a-5ac4-43d7-f576e552c3e1

jni:
This is related to #3397, and I'm not sure whether @stefanv likes the solution proposed here, namely, using fully qualified imports. At any rate, I also don't know whether it solves the problem, because importing the a module in a package automatically imports everything in the `__init__.py`, even if you use a qualified module name within that package...

stefanv:
> This is related to #3397, and I'm not sure whether @stefanv likes the solution proposed here, namely, using fully qualified imports. At any rate, I also don't know whether it solves the problem, because importing the a module in a package automatically imports everything in the `__init__.py`, even if you use a qualified module name within that package...

Yes, I suspect it'd have to be something like

```
from skimage.morphology.watershed import watershed
```

Even then not sure it'd guarantee avoiding the circular import.

rfezzani:
This PR moved much further then I expected!
It is in fact related to #3397, and I found that the `skimage._shared` submodule is a good place for shared functions :-)

## Modules/functions move

- `transform._warps_cy` -> `_shared._warps_cy`
- `transform._warps.warp` -> `_shared._warp.warp`
- `transform._warps.resize` -> `_shared._warp.resize`
- `transform._geometric` -> `_shared._geometric`
- `morphology._flood_fill` -> `segmentation._flood_fill`
- `morphology._flood_fill_cy` -> `segmentation._flood_fill_cy`
- `morphology.watershed` -> `segmentation.watershed`
- `morphology.heap_watershed.pxi` -> `segmentation.heap_watershed.pxi`
- `morphology.heap_general.pxi` -> `segmentation.heap_general.pxi`
- `morphology._convex_hull` -> `_shared._convex_hull`
- `morphology.util` -> `_shared.morpho_util`
- `measure._label` -> `_shared._label`
- `measure._ccomp` -> `_shared._ccomp`
- `measure._pnpoly` -> `_shared._pnpoly`
- `measure.pnpoly` -> `_shared.pnpoly`

The corresponding tests also moved to the new modules locations.

## Deprecations

Some functions are actually duplicated in different modules, I suggest to deprecate some to keep only one location for each function:

- `measure.label` in favor of `morphology.label`
- `morphology.watershed` in favor of `segmentation.watershed`
- `morphology.flood` in favor of `segmentation.flood`
- `morphology.flood_fill` in favor of `segmentation.flood_fill`

This PR being huge, it is may be a good idea to split it multiple ones, I let you decide here ;-)

sciunto:
Nice work but I think we must discuss these changes. I guess there was some reasons to put labels in measure and watershed in morphology. I think @jni knows.

jni:
> I think @jni knows.

He does! 😂 Well, for watershed. As far as I know, labels has always been in measure.

Anyway, the historical location for watershed was morphology, because it is considered a classical morphological algorithm. For reasons apparent in this PR, people including myself would look for it in segmentation instead. Indeed, a bunch of functions could easily fit in more than one of our modules. We actually debated a flat namespace for skimage at https://github.com/scikit-image/scikit-image/issues/1608, and one of the outcomes of that debate was to create aliases for such “difficult to classify” functions. 

So, I’m not sure we want to back away from that solution...!

> we must discuss these changes.

I agree. This PR is pretty massive and it’s unclear to be that it is the best solution (though it might be!)

rfezzani:
> I agree. This PR is pretty massive and it’s unclear to be that it is the best solution (though it might be!)

I also agree, this  PR is much like a POC or a demo allowing to discuss based on concrete changes :smile:

The deprecation introduced in this PR are not about where functions must belong, but I think that a function must not be included in two different modules!

stefanv:
Yes, I may be backtracking on that proposed solution myself.  I get that it can be confusing, but then rather move it to the most logical place, and put a deprecated link to it from the other module.  Simply permanently aliasing it from both will lead to confusing code examples out there.

jni:
@stefanv it's nice to be able to revisit decisions. =)

Well, I propose then that watershed goes to segmentation, since that is the *purpose* of watershed... That it came from morphology is an implementation detail, I think. =)

stefanv:
Yes, I agree with that choice.

rfezzani:
As you advised, `watershed` is moved from `morphology` to `segmentation` in #4443.
What about the other deprecation proposed in this PR?

jni:
@rfezzani so I think `_shared` is going to become too crowded if we put anything that is used by more than one module in there. I was thinking about this today with #4443, and I came up with the following idea:

- For every top-level module (segmentation, filters, etc), create a private module, _segmentation, _filters, etc.
- Put *everything* in there, but no `__init__.py`. To import from those packages, you need to import the full module path (e.g. `from .._filters.edge import sobel`).
- The top-level modules stay there, but they import all the relevant names from their corresponding private modules
- internally, we *always* use the fully qualified, private module path
- (Bonus) in `skimage/__init__.py`, add a module `__getattr__` ([PEP-562](https://www.python.org/dev/peps/pep-0562/)) to implement lazy importing, so that `import skimage` is instant, but still gives you the use of `skimage.filters.sobel` using lazy importing.

Cross referencing to #3397 because, pending comments reminding me of some essential information that I've overlooked, I think this might actually be the right solution for our issues!

sciunto:
I like your proposition @jni Are we sure that no code is shared across submodules? Otherwise, wouldn't it be relevant to keep _shared for that?

jni:
Oh yeah we should keep `_shared`, I just don't want it to be an enormous, amorphous pile. When things make sense in one location, put them in that location, when they are of general utility, they can go in `_shared`/`_util`.

rfezzani:
What about organizing the `_shared` module?
 I mean duplicating the structure of the to level module in `_shared`: (`_shared/segmentation`, `_shared/morphology`...) and importing the effectively shared functions from there...

rfezzani:
I like the lazy importing  idea ;-)

rfezzani:
@jni, I may be misunderstood your proposition, but I don't see how your solution fixes the circular imports between the created private modules...

rfezzani:
BTW, may initial question
> What about the other deprecation proposed in this PR?

was about: 

>Some functions are actually duplicated in different modules, I suggest to deprecate some to keep only one location for each function:
>
> - `measure.label` in favor of `morphology.label`
> - `morphology.flood` in favor of `segmentation.flood`
> - `morphology.flood_fill` in favor of `segmentation.flood_fill`

Are these deprecation OK for you?

jni:
@rfezzani 

> What about organizing the _shared module?

This introduces an extra level of nesting, which I dislike: `from .._filters.edge import sobel` vs `from .._shared.filters.edge import sobel`. In addition, I don't think `_shared` captures what I want here: *all* modules and functions would live there, and the packages would just import from that. So maybe I would change it to `.._skimage.filters.edge`, but I'm not sure...

> I don't see how your solution fixes the circular imports between the created private modules...

The problem is that, right now, if you only want the `sobel` filter, you *have* to import *all* filters, because when you write `from ..filters.edge import sobel`, `filters/__init__.py` gets run, so you're importing *all* things there, which in turn imports a heck of a lot of functions from a heck of a lot of other modules.

By eliminating the `__init__.py` imports in the `_module`s, we can break apart files until nothing unnecessary is imported. In short, removing the `__init__.py` gives us the freedom to break these cycles *without* having to move everything useful to `_shared`.

> measure.label in favor of morphology.label
> morphology.flood in favor of segmentation.flood
> morphology.flood_fill in favor of segmentation.flood_fill

I'm not sure about these. I think `measure.label` makes sense to me. `flood_fill` does not have to be about segmentation. It can be used for painting, for example. Anyway, others can weigh in, but my instinct is: it is not high-enough value to move them. Some functions have ambiguous location, and you have to pick a side, but if it's not extremely clear that moving is the right choice, we should probably err on the side of not touching things.

rfezzani:
Brilliant :tada:
Thank you @jni for these details.
Saying that, I feel that the `_shared` and `util` modules become redundant, isn't it?

> The top-level modules stay there, but they import all the relevant names from their corresponding private modules

If the top level modules are empty directories with just a `__init__.py` file, why not just make them simple files: `segmentation/__init__.py` -> `segmentation.py`... and so on?

rfezzani:
> I'm not sure about these. I think measure.label makes sense to me. flood_fill does not have to be about segmentation. It can be used for painting, for example. Anyway, others can weigh in, but my instinct is: it is not high-enough value to move them. Some functions have ambiguous location, and you have to pick a side, but if it's not extremely clear that moving is the right choice, we should probably err on the side of not touching things.

I am not really concerned about where these functions must be, but I am more concerned by the fact that they are duplicated in both modules. Each function, should be in only one location, we just have to decide where ;-)

rfezzani:
@jni, I gave a try to the strategy that you propose, it unfortunatly not as simple as it looks :-(
some modules such as `color.colorconv` and `exposure.exposure` have a clear circular import :cry:.

hmaarrfk:
lazy importing isn't as great as it touts to be.

The main problem with python is that it delays: "the import delay time", and "hides import problems".

Actually calling functions needs to be done to test, instead of just calling "import scikit-image" and assuming that it works.

I guess that is the challenge with most interpreted languages as opposed to compiled

hmaarrfk:
For building some of my applications, importing all the python modules is so slow i had to create a splash screen kinda like spyder's.

If you lazy import all, the logic of that splash screen will actually have to fetch all the functions I need, one by one, as opposed to calling `import`

rfezzani:
Sorry, I closed this PR by mistake, I don't know how to reopen it :cry:

hmaarrfk:
I don't think you can until you restore your branch. 

GitHub might have some magical settings if you go into your own settings.

jni:
@hmaarrfk, but surely by delaying imports, you can eliminate the need for a splash screen altogether?

Regarding delaying errors, I mean, sure, but testing only needs to be done once per install. Then forever more you get nicer syntax, nicer behaviour (e.g. currently if you `import skimage`, *some* of the submodules are imported because of our crazy import structure, but not others. Search for `skimage.measure` in issues for where that can come crashing down.), and faster start times. Paying a cost slowly over time is very often nicer than paying it all at once, especially on zero interest. =)

Anyway, that's a side issue relative to improving the import structure, which can be done without lazy importing.

@rfezzani can you elaborate on exactly the cycle you are having difficulty breaking? It can always be broken by importing within a function, even though it's somewhat icky to do that. =)

rfezzani:
@jni, `exposure.exposure` imports `color.colorconv.rgb2gray` from `color.colorconv` while `exposure.exposure.rescale_intensity` is imported in `color.colorconv.combine_stains`.
This loop is already broken by importing within the function, but that's exactly what we want to avoid.

PR #4401, opened 2020-01-16
Documentation fixes for transform (rescale, warp_polar)

mellertd:
## Description

This is a very small change to the documentation in `transform`. In addition to a few typos, it is fixing an error in the docstring for `transform.rescale`:

```
anti_aliasing_sigma{float, tuple of floats}, optional
Standard deviation for Gaussian filtering to avoid aliasing artifacts. By default, this value is chosen as (1 - s) / 2 where s is the down-scaling factor."
```
`(1 - s) / 2` is incorrect. This should read `(s - 1) / 2` as per the docs for `transform.resize` on which this function is based.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


jni:
🙏

PR #4409, opened 2020-01-24
Fix subtract_mean underflow correction

### LINKED ISSUE #4365: filters.rank.subtract_mean underflows on uint16, otherwise gives strange output
`skimage.filters.rank.subtract_mean` does not seem to behave in the way I'd expect, nor the way it is documented. Witness:

1. Undocumented shift by 127
```python
In [3]: a = np.array([[10, 10, 10]], dtype=np.uint16)                                                                                                                                                                                                                                                                                                                       

In [4]: skimage.filters.rank.subtract_mean(a, np.ones((1, 3)))                                                                                                                                                                                                                                                                                                              
Out[4]: array([[127, 127, 127]], dtype=uint16)
```

I'd expect zero values here, but I get 127. Given operation on unsigned integers, it's perhaps reasonable that we have some way to account for negative values resulting from mean subtraction, but this seems to be undocumented and certainly not what I'd expect for `uint16` input.

2. Undocumented division by two
```python
In [5]: a = np.array([[10, 10, 40]], dtype=np.uint16)                                                                                                                                                                                                                                                                                                                       

In [6]: skimage.filters.rank.subtract_mean(a, np.ones((1, 3)))                                                                                                                                                                                                                                                                                                              
Out[6]: array([[127, 122, 134]], dtype=uint16)
```
If we add ten to the mean at the center value, leaving it unchanged, the result should change by ten. But it changes by five.

3. Underflow
```python
In [7]: a = np.array([[384, 0, 384]], dtype=np.uint16)                                                                                                                                                                                                                                                                                                                      

In [8]: skimage.filters.rank.subtract_mean(a, np.ones((1, 3)))                                                                                                                                                                                                                                                                                                              
Out[8]: array([[  223, 65535,   223]], dtype=uint16)
```
That's not what I'd expect here; this function is documented as accepting `uint16` input, but it causes underflow in surprising cases. (Note that values less than 384 don't cause underflow here.)

It's easy to see that this results from the implementation at https://github.com/scikit-image/scikit-image/blob/d0c78d7809fd21116ac386f81ea964b8350b9813/skimage/filters/rank/generic_cy.pyx#L169.

Version info:
```
3.6.8 (default, Oct  9 2019, 14:04:01) 
[GCC 5.4.0 20160609]
Linux-4.16.18-rt12-x86_64-with-Ubuntu-16.04-xenial
scikit-image version: 0.16.2
numpy version: 1.17.4
```

rfezzani:
## Description

Fixes #4365.

This PR implements the solution proposed here https://github.com/scikit-image/scikit-image/issues/4365#issuecomment-578029063 to fix undocumented behavior of `subtract_mean`.

In addition, memory allocation is optimized when mask is `None` for the `skimage.filters.rank` functions.

## For reviewers

  later.


skimage/filters/rank/generic.py
```diff
@@ -567,6 +560,14 @@ def subtract_mean(image, selem, out=None, mask=None, shift_x=False,
     out : 2-D array (same dtype as input image)
         Output image.
 
+    Notes
+    -----
+    Subtracting the mean value amy introduce underflow. To compensate
```
emmanuelle:
```suggestion
    Subtracting the mean value may introduce underflow. To compensate
```

skimage/filters/rank/generic.py
```diff
@@ -544,7 +537,7 @@ def geometric_mean(image, selem, out=None, mask=None,
 
 def subtract_mean(image, selem, out=None, mask=None, shift_x=False,
                   shift_y=False):
-    """Return image subtracted from its local mean.
+    """Return image subtracted from its local mean.)
```
rfezzani:
```suggestion
    """Return image subtracted from its local mean.
```

emmanuelle:
Merging, thanks!

PR #4410, opened 2020-01-24
Speedup warp for 3D images

Mostafa-Alaa:
## Description

This PR speeds up warping for 3D images (3D or multichannel 2D) up-to 5x by writing `_warp_fast_batch` which handles a batch of 2D images directly.

The original approach had some performance problems like:
- Overhead slicing and repacking the image
- Force converting image slices to contiguous arrays which in some cases might not be needed
- Recomputing the transform at each iteration which only depends on the coordinates in the 2D plane

And also an API problem where the input had to have its channel axis placed last, overcoming that would require transposing the image before and after the computation

The proposed code works on images with format [plane, row, column] or [row, column, channel]
by specifying the `channel_axis` parameter and controls whether the iteration over the channel axis should be the innermost loop or the outermost loop which basically impacts the caching.

I've ran some benchmarks and here are the results
CF32
CF64
CL32
CL64
Each figure has 3 rows (for interpolation order) and 5 columns (for image size nxn) of subplots which compares the number of channels vs speedup

A bunch of TODOs:
- Settle on the logic for choosing channel loop placement
- Documentation
- Benchmark code

Reviews and change request are most welcome

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_warps_cy.pyx
+++ b/skimage/transform/_warps_cy.pyx
@@ -143,6 +143,13 @@ def _warp_fast(
     would be [0, 1, 2, 1, 0, 1, 2].
 
     """
+    cdef np.intp_t[2] shape
+    if output_shape is None:
+        shape[0] = np.PyArray_DIM(image, 0)
+        shape[1] = np.PyArray_DIM(image, 1)
+    else:
+        shape[0] = output_shape[0]
+        shape[1] = output_shape[1]
 
     cdef np_floats[:, ::1] img = np.PyArray_GETCONTIGUOUS(image)
     cdef np_floats[:, ::1] M = np.PyArray_GETCONTIGUOUS(H)
@@ -180,14 +187,6 @@ def _warp_fast(
     else:
         raise ValueError("Unsupported interpolation order", order)
 
-    cdef np.intp_t[2] shape
-    if output_shape is None:
-        shape[0] = img.shape[0]
-        shape[1] = img.shape[1]
-    else:
-        shape[0] = output_shape[0]
-        shape[1] = output_shape[1]
-
     cdef np.ndarray out = np.PyArray_ZEROS(2, shape, dtype, 0)
     cdef np_floats[:, ::1] vout = out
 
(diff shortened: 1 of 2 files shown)


skimage/transform/_warps_cy.pyx
```diff
@@ -2,8 +2,10 @@
 #cython: boundscheck=False
 #cython: nonecheck=False
 #cython: wraparound=False
-import numpy as np
-cimport numpy as cnp
+cimport numpy as np
+np.import_array()
```
hmaarrfk: what does this do?

hmaarrfk: I don't think this necessary as `cimport` should take care of it. I think..... otherwise, we have real issue in the library ^_^

Mostafa-Alaa: Well then it seems we do have a library issue, check [this](https://numpy.org/devdocs/reference/c-api/array.html#importing-the-api).

skimage/transform/_warps_cy.pyx
```diff
@@ -119,37 +143,126 @@ def _warp_fast(np_floats[:, :] image, np_floats[:, :] H, output_shape=None,
...
+    if output_shape is None:
+        shape[0] = np.PyArray_DIM(image, 0)
+        shape[1] = np.PyArray_DIM(image, 1)
+    else:
+        shape[0] = output_shape[0]
+        shape[1] = output_shape[1]
 
-    cdef np_floats[:, ::1] img = np.ascontiguousarray(image)
-    cdef np_floats[:, ::1] M = np.ascontiguousarray(H)
+    cdef np_floats[:, ::1] img = np.PyArray_GETCONTIGUOUS(image)
```
hmaarrfk: Why are you using the capi?

Why not simply leave this as is, I don't think you are avoiding copies here

https://github.com/numpy/numpy/blob/v1.17.0/numpy/core/_asarray.py#L179

skimage/transform/_warps_cy.pyx
hmaarrfk: Ah, we seldom use the capi directly, you seem to be using it.

hmaarrfk:
Is using the CAPI meaningful? We typically use memoryvirws and the python API.

Could you share a benchmark with u with code and the benchmark in the appropriate folder.

Mostafa-Alaa:
> Could you share a benchmark with u with code and the benchmark in the appropriate folder.

Sure, [here](https://gist.github.com/Mostafa-Alaa/98c322226374bc4357396e6ca0f1798a) is how I'm benchmarking the code.

You just need to update `warp` in `_warp.py` to accept `channel_axis` and `channel_innermost_loop`, I didn't want to include them for the time being as they are subject to review and change.

Mostafa-Alaa:
@hmaarrfk I just checked multiple Cython files across the library and noticed that the Numpy's Python API is the one being used, while the C API used for simple stuff like data type checks.

I see that as a problem since Cython is basically C so it would seem more natural to use the C API rather than the Python one, it's like you go from C to Python to C so why not just use C code allover.

I've actually benchmarked the use of C vs Python API in `_warp_fast` without running the actual warp code (the code where the GIL is released and most of the time takes place) and got about 10x speedup.

Now I think opening an issue and having a discussion about the use of C API would be better than in this PR, so what do you think should I revert to Python API for now ?

hmaarrfk:
I can see how we are using c datatypes from numpy, but most of the code is using cython memoryviews isn't it?

Mostafa-Alaa:
Doing the computations on the memoryviews is definitely the right call, the problem here is with the setup code (like making sure the numpy array is contiguous or creating an array), using Numpy's C API avoids some unnecessary overhead.

hmaarrfk:
Is that seriously a large cost? It seems to me that that is the cost paid for by using numpy and python. 

I would be happy to work with you to isolate your two changes: your computation order changes. The usage of the numpy c api.

I'm still not totally convinced that the usage of the Capi is the right call, but interested in being shown otherwise.

Mostafa-Alaa:
> Is that seriously a large cost? It seems to me that that is the cost paid for by using numpy and python.

The cost is really subjective, for a cheap computation (small image size, nearest neighbour interpolation) reducing the overhead will help, but when the computation is expensive (large image size, bicubic interpolation) it doesn't really matter as most of the time share is spent on the computation itself.

> I'm still not totally convinced that the usage of the Capi is the right call, but interested in being shown otherwise.

I know it might be inconvenient to use the C API as you're probably more used to using the Python API (who isn't), but the C API is definitely the right call here as Cython generates more code like fetching the numpy module, fetching and invoking the function, more checks, while when using the C API it just calls the function.

> I would be happy to work with you to isolate your two changes: your computation order changes. The usage of the numpy c api.

That's great, I think separating the two changes would be better.

matthew-brett:

<details>

## Rationale

</details>

PR #4416, opened 2020-01-27
Fix profile_line interpolation errors

### LINKED ISSUE #4279: Interpolation error in measure.profile_line
## Description

#4206, once merged, exposes interesting behavior for [measure.profile_line](https://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.profile_line). Consider

```python
image = np.arange(12).reshape(4, 3)
# reduce_func is a new parameter proposed in #4206
prof = profile_line(image, (0, 1), (3, 1), linewidth=3, reduce_func=None)
assert_array_equal(image, prof)
```
As I understand it, the returned profile should be identical with the original image but we get this (value at (0, 2) is different):
```
AssertionError: 
Arrays are not equal
Mismatch: 8.33%
Max absolute difference: 2
Max relative difference: 0.
 x: array([[ 0,  1,  2],
       [ 3,  4,  5],
       [ 6,  7,  8],
       [ 9, 10, 11]])
 y: array([[ 0,  1,  0],
       [ 3,  4,  5],
       [ 6,  7,  8],
       [ 9, 10, 11]])
```
This might originate in [scipy.ndimage.map_coordinates](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.map_coordinates.html#scipy.ndimage.map_coordinates). Looks like a potential bug to me but might be an edge case for the underlying interpolation that can be ignored/is not reasonable to fix. 
This behavior is already relevant even before merging #4206 because the faulty value will change the mean value of the profile line.

https://github.com/scikit-image/scikit-image/pull/4206#issuecomment-544920806 also highlights additional strange behavior.

## Version information

Current master, exposed once #4206 is merged.

rfezzani:
## Description

Fixes #4279 based on my comment https://github.com/scikit-image/scikit-image/issues/4279#issuecomment-578996196

## For reviewers

  later.


skimage/measure/tests/test_profile.py
```diff
@@ -110,21 +110,21 @@ def test_pythagorean_triangle_transpose_left_down_linewidth():
 def test_reduce_func_mean():
     prof = profile_line(pyth_image, (0, 1), (3, 1), linewidth=3, order=0,
                         reduce_func=np.mean)
-    expected_prof = np.array([0, 0.8, 1, 0.2])
+    expected_prof = pyth_image[:4, :3].mean(-1)
```
emmanuelle: not sure I understand why this should work...

skimage/measure/profile.py
```diff
@@ -14,10 +14,11 @@ def profile_line(image, src, dst, linewidth=1,
         (3D array, where the final axis contains the channel
         information).
     src : 2-tuple of numeric scalar (float or int)
-        The start point of the scan line.
+        The positive coordinates of the start point of the scan line.
```
emmanuelle: Is it worth adding this? An error message will be raised anyway otherwise. I'm all for clarity, but also short docstrings are easier to read.

rfezzani: I will relax this requirement according to @stefanv https://github.com/scikit-image/scikit-image/issues/4279#issuecomment-580461567: negative values will be considered to be inside the the image (a warning will be raised anyway). Only out of range coordinates will cause an error to be raised.

skimage/measure/tests/test_profile.py
rfezzani: That's exactly what the `reduce_func` argument is supposed to do (see #4206).

emmanuelle: this I got, but the `pyth_image` is special so that its non-zero pixels correspond exactly to the scan line, do I understand correctly? This last point is not completely clear to me.

emmanuelle:
Could you please add a test with negative coordinates?

skimage/measure/tests/test_profile.py
rfezzani: No, it is just because `src` is set to `(0, 1)` and `dst` to `(3, 1)` with a `linewidth=3`:
`src` and `dst` are indices positioned on the same column (`src[1] = dst[1] = 1`). So the profile line will be computed using 3 (`linewidth`) columns vertically centered around the second column, from the row 0 (`src[0]`) to the row 3 (`dst[0]`).
That's why the expected result is computed using `pyth_image[:4, :3] = pyth_image[src[0]:dst[0]+1, src[1]-1:dst[1]+2]`.
Sorry if this was not clear.

rfezzani:
Yes, the test for negative coordinates and the deprecation warning for out of bound coordinates are on my today todo list ;-)

skimage/measure/tests/test_profile.py
emmanuelle: oh ok, no sorry, it's just me being slow :-)

rfezzani:
Thanks again @emmanuelle ;-)

emmanuelle:
@skimage/core who want to review (and possibly press the green button)?

rfezzani:
@jni I reverted all my previous changes and deprecated the  `constant` mode in favor of `reflect`.

skimage/measure/profile.py
```diff
@@ -85,7 +87,16 @@ def profile_line(image, src, dst, linewidth=1,
            [1.        , 1.        , 0.        ],
            [1.        , 1.        , 0.        ],
            [1.41421356, 1.41421356, 0.        ]])
+
     """
+
+    if mode is None:
+        warn("Default out of bound interpolation 'constant' mode is "
```
jni:
```suggestion
        warn("Default out of bounds interpolation mode 'constant' is "
```

skimage/measure/profile.py
```diff
@@ -85,7 +87,16 @@ def profile_line(image, src, dst, linewidth=1,
...
+             "Deprecated. In version 0.19 it will be set to 'reflect'."
```
jni:
```suggestion
             "deprecated. In version 0.19 it will be set to 'reflect'. "
```

skimage/measure/profile.py
```diff
@@ -85,7 +87,16 @@ def profile_line(image, src, dst, linewidth=1,
...
+             "To avoid this warning please set explicitely the mode argument.",
```
jni:
```suggestion
             "To avoid this warning, set `mode=` explicitly.",
```

rfezzani:
Thank you @jni, and sorry @emmanuelle for requesting a re-review from you :pray:

skimage/measure/profile.py
```diff
@@ -4,21 +4,22 @@
...
+    image : ndarray, shape (M, N[, C])
         The image, either grayscale (2D array) or multichannel
         (3D array, where the final axis contains the channel
         information).
-    src : 2-tuple of numeric scalar (float or int)
-        The start point of the scan line.
-    dst : 2-tuple of numeric scalar (float or int)
-        The end point of the scan line. The destination point is *included*
-        in the profile, in contrast to standard numpy indexing.
+    src : array_like, shape (2, )
```
stefanv: Just double checking: is this function 2D-only for now?

rfezzani: Yes it is :-(

rfezzani:
Does anyone want to push the button? :sweat_smile:

rfezzani:
Gentle up :slightly_smiling_face:

jni:
:tada: thank you @rfezzani!

rfezzani:
Thank you @jni ;)

Issue #3200, opened 2018-06-14
Should rgb2gray accept gray images?


hmaarrfk:
https://github.com/scikit-image/scikit-image/blob/4686b7da06593ff002986de6a378cba44f768148/skimage/color/colorconv.py#L800

See issue: #3185

it is pretty difficult to correctly guess if an ND image is grey. I think ND color conversion would be super useful for skimage, so that is why I'm asking for input for how to deal with this.

I could do something like:

```python
if rgb.shape[-1] not in [3, 4]:
```
or 
```python
if rgb.shape[-1] > 4:
```
but that would break things for tiny little images ;)

maybe this kind of automagic behaviour is better suited for an other wrapper function if we want to keep it?

hmaarrfk:
As a followup question, should `gray2rgb` attempt to guess if an image is RGB?

Also, the logic is a little wonky. it uses squeeze, though, not entirely correctly.

https://github.com/scikit-image/scikit-image/blob/4686b7da06593ff002986de6a378cba44f768148/skimage/color/colorconv.py#L843

hmaarrfk:
Could I add a `multichannel` parameter to turn off all this automagic logic?

jni:
@hmaarrfk yes, our new mission statement will include something to the effect of "we don't like magic", but there is some legacy code that violates that, and guessing RGB is one of these. So, we are in favour of removing this magic, but we have to be careful with backwards compatibility and not breaking people's code.

The concern for tiny axes is very real and iirc someone *was* bitten by this very problem a while back.

Something like a keyword argument would be useful, but we have to be careful with it: 

- `multichannel=None` -> do some magic
- `multichannel=True` -> always assume RGB (raise ValueError if shape incorrect)
- `multichannel=False` -> ???? 

So, it's a bit wonky for this function. But something along these lines is desirable, yes.

hmaarrfk:
I kinda get stuck thinking that these functions should not take a `guess_channels` (instead of `multichannel`) parameter later in the project life cycle.

  - `guess_channels=True` old behaviour where it tries to guess in certain places.
  - `guess_channels=False` new behaviour, where if it works with `numpy`, then it is probably the right thing?

  - `guess_channels=None` is always default but changes behaviour depending on the version number

  - 0.15: `None` behaves like `True` to retain the old behaviour. Raise a warning if `None` or `True` are set. 
  - 0.16: `None` behaves like `False`. 
[...]

PR #4417, opened 2020-01-28
Fix rgb2gray docstring to show that 2-D arrays are allowed as inputs

parekhnish:
## Description

This is a docstring fix, refers to discussion in #3200, specifically [this reply](https://github.com/scikit-image/scikit-image/issues/3200#issuecomment-578013692).

`rgb2gray()` allows 2-D arrays as inputs, but the docstring stated it would raise a `ValueError`. This PR does the following:
- Changes the docstring to show that 2-D inputs are allowed.
- Explains what happens to 2-D inputs (essentially, nothing changes except converting it to row-major order, if needed).

## For reviewers

  later.

### DIFF
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -753,7 +753,10 @@ def rgb2gray(rgb):
     rgb : array_like
         The image in RGB format, in a 3-D or 4-D array of shape
         ``(.., ..,[ ..,] 3)``, or in RGBA format with shape
-        ``(.., ..,[ ..,] 4)``.
+        ``(.., ..,[ ..,] 4)``
+        or
+        The image in grayscale format, in a 2-D array of shape
+        `(M, N)``
 
     Returns
     -------
@@ -764,7 +767,7 @@ def rgb2gray(rgb):
     Raises
     ------
     ValueError
-        If `rgb2gray` is not a 3-D or 4-D arrays of shape
+        If `rgb` is a 1-D array, or not a 3-D or 4-D array of shape
         ``(.., ..,[ ..,] 3)`` or ``(.., ..,[ ..,] 4)``.
 
     References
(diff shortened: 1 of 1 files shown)


skimage/color/colorconv.py
```diff
@@ -753,7 +753,10 @@ def rgb2gray(rgb):
     rgb : array_like
         The image in RGB format, in a 3-D or 4-D array of shape
         ``(.., ..,[ ..,] 3)``, or in RGBA format with shape
-        ``(.., ..,[ ..,] 4)``.
+        ``(.., ..,[ ..,] 4)``
```
rfezzani: We have to remove the mention to arrays of shape ``(.., ..,[ ..,] 4)`` since `_prepare_colorarray` doesn't accept RGBA images.

skimage/color/colorconv.py
```diff
@@ -764,7 +767,7 @@ def rgb2gray(rgb):
     Raises
     ------
     ValueError
-        If `rgb2gray` is not a 3-D or 4-D arrays of shape
+        If `rgb` is a 1-D array, or not a 3-D or 4-D array of shape
         ``(.., ..,[ ..,] 3)`` or ``(.., ..,[ ..,] 4)``.
```
rfezzani: Same as above, RGBA images are not supported...

skimage/color/colorconv.py
```diff
@@ -753,7 +753,10 @@ def rgb2gray(rgb):
     rgb : array_like
         The image in RGB format, in a 3-D or 4-D array of shape
         ``(.., ..,[ ..,] 3)``, or in RGBA format with shape
-        ``(.., ..,[ ..,] 4)``.
+        ``(.., ..,[ ..,] 4)``
+        or
```
rfezzani: This can simply be added in the `Notes` section, what do you think?

skimage/color/colorconv.py
```diff
@@ -780,6 +783,9 @@ def rgb2gray(rgb):
 
     If there is an alpha channel present, it is ignored.
 
+    If the input array is 2-D, it is converted to a row-major (i.e. C-style)
```
hmaarrfk: I would be fine adding al lthe comments in the `Notes` section stating that calling this function with `2D` images, while supported for now, is heavily discouraged.

soupault:
@parekhnish thanks for your effort! The solution has been discussed in https://github.com/scikit-image/scikit-image/issues/3200 and implemented in https://github.com/scikit-image/scikit-image/pull/4438.

PR #4418, opened 2020-01-28
Add nD support to colorconv module

### LINKED ISSUE #3185: ND multichannel arrays in color conversion

### LINKED ISSUE #3200: Should rgb2gray accept gray images?

rfezzani:
## Description

This PR is an attempt to Fix #3185.  If accepted, it also closes #4417.

- nD support is added to `colorconv` module,
- gray level image conversion with `rgb2gray` is deprecated,
- RGBA image conversion with `rgb2gray` is deprecated.

Comments are welcome.

Reequired PRs:
- [x] #4424 
- [x] #4438

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/color/colorconv.py
```diff
@@ -788,10 +791,21 @@ def rgb2gray(rgb):
...
+             'a ValueError will be raised if input image last dimension '
+             'length is not 3.', FutureWarning, stacklevel=2)
         return np.ascontiguousarray(rgb)
 
-    rgb = _prepare_colorarray(rgb[..., :3])
+    if rgb.shape[-1] == 4:
+        warn('RGBA image conversion is now deprecated. In version 0.19, '
+             'a ValueError will be raised if input image last dimension '
+             'length is not 3.', FutureWarning, stacklevel=2)
+        rgb = rgb[..., :3]
```
rfezzani: this line can be replaced by
```python
    return rgba2gray(rgb)
```
using #4419 .

rfezzani:
@hmaarrfk, since you already tried to address this issue in #3199, may can you have a first review on this one if you have time? Thank you ;-)

hmaarrfk:
The hardest things about this PR is making rgb2gray work when the image is input as grayscale. Has that behavior been deprecated yet?

rfezzani:
I propose in this PR to deprecate grayscale and RGBA images conversion using `rgb2gray`.
Grayscale images conversion using `gray2rgb` is also deprecated.

hmaarrfk:
> I propose in this PR to deprecate grayscale and RGBA images conversion using rgb2gray.

Unfortunately, combining it with a 600 + line change isn't the way to focus the discussion on the deprecation. 

Lets focus our attention on the grayscale deprecation. I think you had an other PR for that.

rfezzani:
The problem is that the deprecation is necessary for a complete support of nD arrays in the colorconv module.
In fact, all the guessing about input dimensions is made because we wanted to add edge cases support to gray2rgb (RGB input and RGBA input and output) and rgb2gray (gray level and RGBA inputs).
Without this guessing, adding the nD support becomes obvious...

As you mentioned, RGBA support is deprecated in gray2rgb in #4442. We can in fact think about opening an other PR to deprecate the 3 other edge cases support, but does it really ease the review of this one?

skimage/color/colorconv.py
```diff
@@ -155,34 +162,34 @@ def _prepare_rgba_array(arr):
     """
     arr = np.asanyarray(arr)
 
-    if arr.ndim not in [3, 4] or arr.shape[-1] != 4:
-        msg = ("the input array must have a shape == (.., ..,[ ..,] 4)), "
+    if arr.ndim < 2 or arr.shape[-1] != 4:
```
hmaarrfk: this should probably be combined with the above, with the `n_channels` parameter added to switch between `arr.shape[-1] != n_channels`

skimage/color/colorconv.py
```diff
@@ -789,10 +808,22 @@ def rgb2gray(rgb):
     >>> img_gray = rgb2gray(img)
     """
 
-    if rgb.ndim == 2:
+    rgb = np.asanyarray(rgb)
+
+    if rgb.ndim == 2 and rgb.shape[-1] != 3:
```
hmaarrfk: I would really like to see a PR with **only** these changes.

skimage/color/colorconv.py
```diff
@@ -789,10 +808,22 @@ def rgb2gray(rgb):
     >>> img_gray = rgb2gray(img)
     """
 
-    if rgb.ndim == 2:
+    rgb = np.asanyarray(rgb)
```
hmaarrfk: does anybody really supply `non-arrays` as images? 

I know it is convenient to type lists of lists, but this `np.asanyarray` is quite costly to call all the time, and for such functions, I don't know if it really helps the user experience.

skimage/color/tests/test_colorconv.py
```diff
@@ -295,10 +296,12 @@ def test_rgb2gray_contiguous(self):
 
     def test_rgb2gray_alpha(self):
         x = np.random.rand(10, 10, 4)
-        assert rgb2gray(x).ndim == 2
+        with expected_warnings(['RGBA image conversion']):
```
hmaarrfk: this is likely one of the most important lines of this PR.

skimage/color/tests/test_colorconv.py
```diff
@@ -295,10 +296,12 @@ def test_rgb2gray_contiguous(self):
...
+            assert rgb2gray(x).ndim == 2
 
     def test_rgb2gray_on_gray(self):
-        rgb2gray(np.random.rand(5, 5))
+        with expected_warnings(['Gray scale image conversion']):
```
hmaarrfk: this one too.

hmaarrfk:
> The problem is that the deprecation is necessary for a complete support of nD arrays in the colorconv module.

I would support a PR that focused on the deprecation first.
Then in a few days, following up with this nD colorconv one.

The issue with combining the two is that it doesn't allow us to focus on the user experience of the `ensure_gray` behavior of the current `rgb2gray` function.

skimage/color/colorconv.py
rfezzani: I was even more hesitating to remove this function since it is only used in rgba2rgb

skimage/color/colorconv.py
rfezzani: It is just made to allow calling `rgb.ndim` and `rgb.shape` in the next line

skimage/color/colorconv.py
rfezzani: I will open such a PR ;-)

skimage/color/colorconv.py
```diff
@@ -789,10 +808,22 @@ def rgb2gray(rgb):
...
+
+    if rgb.ndim == 2 and rgb.shape[-1] != 3:
+        warn('Gray scale image conversion is now deprecated. In version 0.19, '
+             'a ValueError will be raised if input image last dimension '
+             'length is not 3.', FutureWarning, stacklevel=2)
         return np.ascontiguousarray(rgb)
 
-    rgb = _prepare_colorarray(rgb[..., :3])
+    if rgb.shape[-1] == 4:
```
rfezzani: @hmaarrfk, should I open a separate PR for this deprecation also?

skimage/color/colorconv.py
```diff
@@ -842,6 +868,9 @@ def gray2rgb(image, alpha=None):
             is_rgb = True
 
     if is_rgb:
+        warn('RGB and RGBA images conversion is now deprecated. In version '
```
rfezzani: @hmaarrfk, and for this one too?

skimage/color/colorconv.py
rfezzani: Please see #4438

skimage/color/colorconv.py
hmaarrfk: Yeah. You could include those.

Just magic behavior change is hard to come to a convenus to. So the smaller the change, the easier it is to process

skimage/color/colorconv.py
rfezzani: Please see #4439

skimage/color/colorconv.py
jni: If we say in the documentation that these should be arrays (which we do I think?), then it is ok to just assume that `ndim` and `shape` are both attributes of the input.

rfezzani: `image` is documented as `array_like`.

hmaarrfk:
Do we have a checklist with the PR that you pulled out of this one?

rfezzani:
Nop, but only #4438 remains not merged ;-)

hmaarrfk:
Cool. Thanks so much for doing this. 

I have a feeling that now that we don't support so many edge cases, we can start to vectorize more operations and start to think about even cooler ways to speed things up.

rfezzani:
Oops, @hmaarrfk, I also forgot about #4424...
Yes we are really close now ;-)

hmaarrfk:
i added a checklist at the top.

skimage/color/colorconv.py
```diff
@@ -132,54 +140,39 @@ def convert_colorspace(arr, fromspace, tospace):
...
-    out : ndarray
-        The image in RGB format, in a 3-D array of shape ``(.., .., 3)``.
+    out : (.., 3) ndarray
+        The image in RGB format. Same dimensions as input.
 
     Raises
     ------
     ValueError
-        If `rgba` is not a 3-D array of shape ``(.., .., 4)``.
+        If `rgba` is not at least 2-D with shape (.., 4).
```
emmanuelle, line 175: Should we add a note to say that the output is float (and <1)? It was not obvious to me, like with a default background I could have expected that just the last channel was dropped (but ok for this one could also do just slicing). @rfezzani happy to open an issue for this instead because this has nothing to do with your PR.

skimage/color/tests/test_colorconv.py
```diff
@@ -729,3 +730,27 @@ def test_gray2rgba_alpha():
     with pytest.raises(ValueError) as err:
         rgba = gray2rgba(img, alpha)
     assert expected_err_msg == str(err.value)
+
+
+@pytest.mark.parametrize("func", [rgb2hsv, hsv2rgb,
```
emmanuelle, line 735: how many functions have a different shape for input and output? I can see `rgb2gray` for sure but maybe others. Splitting in two test functions, one for functions preserving the shape and another for the other functions, would allow us to have a more coercive test for functions preserving the shape.

emmanuelle:
Thanks @rfezzani for a PR which must have been a lot of work! I only had two minor comments. It'd be great if another person could review this one @skimage/core.

emmanuelle:
The conflict is a tiny one so it shouldn't deter anyone from reviewing ;-).

skimage/color/tests/test_colorconv.py
rfezzani: I think `rgb2gray` is the only one!

rfezzani: `gray2rgb`, `gray2rgba`, `rgb2gray` and `rgba2rgb` are the four functions that have a different input and output shape. Should I write specific tests for these functions? Because only `rgba2rgb` can't fit this test because input shape last dimension must be 4 instead of 3...

emmanuelle: Ideally - but I know I'm nitpicking here - I would keep the tests which you have written for `gray2rgb`, `gray2rgba`, and `rgb2gray`, write a specific test for `rgba2rgb` and for all functions which preserve shape, test that the output shape is exactly the input shape.

rfezzani: I implemented your suggestion, thank you @emmanuelle

skimage/color/colorconv.py
```diff
@@ -1048,7 +1043,9 @@ def lab2xyz(lab, illuminant="D65", observer="2"):
              stacklevel=2)
         z[invalid] = 0
 
-    out = np.dstack([x, y, z])
+    out = np.concatenate([x[..., np.newaxis],
+                          y[..., np.newaxis],
+                          z[..., np.newaxis]], axis=-1)
```
jni, line 1048:
```suggestion
    out = np.stack([x, y, z], axis=-1)
```

skimage/color/colorconv.py
```diff
@@ -132,54 +140,39 @@ def convert_colorspace(arr, fromspace, tospace):
...
     """Check the shape of the array and convert it to
     floating point representation.
     """
     arr = np.asanyarray(arr)
 
-    if arr.ndim not in [3, 4] or arr.shape[-1] != 3:
-        msg = ("the input array must be have a shape == (.., ..,[ ..,] 3)), " +
-               "got (" + (", ".join(map(str, arr.shape))) + ")")
-        raise ValueError(msg)
+    if arr.ndim < 2 or arr.shape[-1] != 3:
```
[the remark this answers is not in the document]
rfezzani: I added 1D array support, thank you for the suggestion @jni.

skimage/color/colorconv.py
```diff
@@ -192,20 +185,30 @@ def rgba2rgb(rgba, background=(1, 1, 1)):
...
+    if len(background) != 3:
+        raise ValueError('background must be an iterable containing 3 RGB '
+                         'values. Got {0} items'.format(len(background)))
+    if np.any(background < 0) or np.any(background > 1):
+        raise ValueError('background RGB values must be floats between '
+                         '0 and 1.')
+
+    background = background[np.newaxis, ...]
+    alpha = arr[..., -1, np.newaxis]
+    channels = arr[np.newaxis, ..., :-1]
```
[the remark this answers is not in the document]
rfezzani: Broadcasting will work between `background` and `channels`, but unfortunately not between `alpha` and `background`:
```python
>>> from numpy.random import rand
>>> import numpy as np
>>> 
>>> background = rand(3)
>>> alpha = rand(5, 6)
>>> channels = rand(5, 6, 3)
>>> out = np.clip((1 - alpha) * background + alpha * channels,
...                a_min=0, a_max=1)
ValueError: operands could not be broadcast together with shapes (5,6) (3,) 
```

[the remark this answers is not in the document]
jni: I meant removing it only in channels, not alpha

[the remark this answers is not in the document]
rfezzani: OK, I see, sorry @jni... It works in fact, but the `np.squeeze` is still needed when the input is 1D (`rgba.shape == (4, )`), because in this case `out.shape ==  (1, 3)`:
```python
    background = background[np.newaxis, ...]
    alpha = arr[..., -1, np.newaxis]
    channels = arr[..., :-1]

    out = np.clip((1 - alpha) * background + alpha * channels,
                  a_min=0, a_max=1)

    if arr.ndim == 1:
        # To ensure that output ndim is the same input ndim
        out = np.squeeze(out, axis=0)
```
Do you think this modification is necessary?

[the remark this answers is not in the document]
jni: Let's forget about this! Thanks for the clarification! =)

jni:
@scikit-image/core this is ready to merge imho!

rfezzani:

emmanuelle:
It's in the box @rfezzani !

rfezzani:
:champagne: Thank you @emmanuelle

Issue #4294, opened 2019-11-13
Proposal: deprecate `multichannel=` in favour of `channel_axis=`


jni:
This might need a SKIP. =P But won't if we get unanimous agreement!

The name `multichannel` is a bit clunky. Worse than that, it does not quite provide sufficient information: we hardcode the channel axis as -1, which might actually be incorrect. Allowing for arbitrary channel axes allows (1) giving the data in acquisition axis order, which often does not have the channels as the last axis, and (2) allowing channels to be the leading axis, which in many cases is more efficient, as the processing happens independently on each channel (e.g. gaussian filtering), and this is more cache friendly if each channel is contiguous.

Over at napari, we've settled on `channel_axis=None` being equivalent to `multichannel=False`, and `channel_axis=<int>` to point to an axis to interpret as channel rather than spatial data. I quite like it, and seeing an issue like #4293, where we might want to either specify *no channels* or a *different channel axis* makes me think that we can refine scikit-image to use this also.

Looking forward to hearing your thoughts!

grlee77:
I also happened to use exactly this `channel_axis` naming in some local code dealing with multichannel MRI data, as some software assumed channels last while others assume channels are first.

I am +1 on moving toward using `channel_axis`, but how do you anticipate handling existing code that assumes channels are the last dimension? The simplest solution is probably to manually call `swapaxes` on function input and then undo the swap on output, but that comes with some performance penalty for cases where the channel axis is not already last.

jni:
> how do you anticipate handling existing code that assumes channels are the last dimension? The simplest solution is probably to manually call swapaxes on function input and then undo the swap on output

Yes, that would be my proposal. The API matters much, much more than performance. We can deal with performance later, when it becomes a measured problem. This would *not* be a performance regression for any existing code. We can do plenty of approximate tweaks, like, if `channel_axis` is 0, and the image size is greater than 4MB (or your L1 cache, or whatever), then do a copy of the axis swapped-data before processing. But we should not clutter the code with these until someone complains.

lagru:
Not sure about the feasibility of the optimizations but I really like API a `channel_axis` argument would enable. It kinda reflects NumPy's well established `axis` argument. +1

hmaarrfk:
My biggest fear is that the release cycle is too slow to make all this maintenance work worthwhile.

[...]

PR #4419, opened 2020-01-28
Add rgba2gray

rfezzani:
## Description

The `rgba2gray` function is added. It helps to fix/deprecate the RGBA image handling in `rgb2gray`.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/color/colorconv.py
```diff
@@ -745,6 +745,38 @@ def rgbcie2rgb(rgbcie):
     return _convert(rgb_from_rgbcie, rgbcie)
 
 
+def rgba2gray(rgba, background=(1, 1, 1)):
```
soupault:
```suggestion
def rgba2gray(rgba, *, background=(1, 1, 1)):
```

soupault:
I think that rgba2gray worth adding to the library, but, perhaps, there is a different opinion among the core devs.

rfezzani:
`rgba2gray` alleviate the strange behavior of `rgb2gray` that accepts RGBA images and simply ignores the alpha channels:
https://github.com/scikit-image/scikit-image/blob/master/skimage/color/colorconv.py#L795

jni:
@rfezzani but could we not fix `rgb2gray` with some keyword arguments instead? To me that is a nicer solution...

rfezzani:
This PR combined with #4424 and #4418 is an attempt to solve #3185, ie add nD support to `colorconv` module. In fact, by defining `rgba2gray` and `gray2rgba`, we alleviate all the necessary guessing in `rgb2gray` and `gray2rgb` and it becomes easy to add nD support using `Ellipsis`...

rfezzani:
@jni, to complete my previous answer, do your proposition looks like:

- the argument `background` set by default to `None` is added.
- if `background is not None and image.shape[:-1] == 4`, consider the input image as RGBA otherwise it is RGB.
- raise an error if `background is not None and image.shape[:-1] != 4`.

From the KISS principle point of view (one function -> one behavior) I am :-1:...

jni:
@rfezzani simplicity scores very highly with me, but increasing the API surface area also increases complexity, albeit along a different axis. And in this case, because of the similarity of the function names, it makes the API harder to use from the standpoint of autocompletion.

I'm also wary of adding new color manipulation functions right now, before dealing with #4294. How do you feel about putting this on hold and revisiting it once we have changed that?

rfezzani:
In my mind, RGB != RGBA, that's why it sounds not natural for me to have a function called `rgb2something` accepting RGBA images.

If the solution of adding a keyword to deal with alpha channel in `rgb2gray` is acceptable for you, there is no reason for not implementing it for all the `rgb2something` functions.

An other option would be to not implement the `rgba2gray` function (close this PR), but deprecate the RGBA support in `rgb2gray` and advise to preprocess RGBA images with `rgba2rgb` before any manipulation.

jni:
> An other option would be to not implement the rgba2gray function (close this PR), but deprecate the RGBA support in rgb2gray and advise to preprocess RGBA images with rgba2rgb before any manipulation.

I don't mind this idea. Let's wait to see what others think!

rfezzani:
> And in this case, because of the similarity of the function names, it makes the API harder to use from the standpoint of autocompletion.

I'm not sure it could be worse: Capture du 2020-01-31 10-27-13
:rofl:

alexdesiqueira:
Hey everyone,
I'm kindly against implementing somehting like that. Since this function is "only" calling `rgb2gray(rgba2rgb(rgba, background))`, I think a note on `rgb2gray` and an illustrated example should suffice. What do you think?

rfezzani:
The main goal of this PR is the deprecation of the RGBA support from `rgb2gray` by proposing an alternative! This deprecation considerably ease adding the nD support for the `colorconv` module like  in #4418.
@alexdesiqueira's solution is perfectly fine for me.

rfezzani:
Please see  this [solution](https://github.com/scikit-image/scikit-image/pull/4418/files#diff-9f4dedd98e7cca606dec9f0b792c1d94R820) without `rgba2gray` function definition.

rfezzani:
May I close this PR?

rfezzani:
Closing :-)

PR #4423, opened 2020-01-29
Deprecate rank.tophat and rank.bottomhat

sciunto:
## Description

`rank.tophat()` and `rank.bottomhat()` are two filters for which the behavior does not correspond to the standard nomenclature. We found a little interest in these filters. These filters can be found in `morphology`.

See https://github.com/scikit-image/scikit-image/pull/3614 for the explanations.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/filters/rank/__init__.py
```diff
@@ -12,7 +12,6 @@
 
 __all__ = ['autolevel',
            'autolevel_percentile',
-           'bottomhat',
```
hmaarrfk: should these be removed now?

sciunto: I removed them to facilitate the tests (it uses this variable)I can handle it differently if necessary

hmaarrfk: __all__ helps with autocomplete and introspection. So I would prefer to keep it until we really don't want users to use it.

hmaarrfk:
There seem to be a few warnings that are issued without the user directly calling tophat

```
filters/rank/generic.py::skimage.filters.rank.generic.bottomhat

  /home/travis/venv/lib/python3.7/site-packages/skimage/filters/rank/generic.py:1: FutureWarning: rank.bottomhat is deprecated. This filter is named incorrectly, which can be confusing. As we believe that the usefulness is narrow, this function will be removed in 0.19.

    r"""

filters/rank/generic.py::skimage.filters.rank.generic.tophat

  /home/travis/venv/lib/python3.7/site-packages/skimage/filters/rank/generic.py:1: FutureWarning: rank.tophat is deprecated. This filter is named incorrectly, which can be confusing. As we believe that the usefulness is narrow, this function will be removed in 0.19.

    r"""


```

skimage/filters/rank/__init__.py
emmanuelle: Hum, I agree here, if it's in the API (albeit deprecated, it should stay in `__all__`).

emmanuelle:
I will approve once the functions make it back to `__all__` :-)

sciunto:
@hmaarrfk Good catch, but I did not find the reason for this... :/ Do you have any idea?

sciunto:
The ``__all__`` thing is fixed.

emmanuelle:
@hmaarrfk where did you see the spurious warnings? I thought our CI failed on uncaught warnings?

hmaarrfk:
Emmanuelle, our CI does not fail on uncaught warnings. That was decided to be too harsh on PRs. But generally, I look through Travis logs.

emmanuelle:
oh ok, thanks @hmaarrfk

emmanuelle:
The warning is weird... I believe it comes from this import https://github.com/scikit-image/scikit-image/blob/master/skimage/filters/tests/test_median.py#L7 (trying to guess from the position at which the warning occurs, but not 100% sure), but I'm still puzzled that it would trigger the warning since the tophat etc. functions are not called.

emmanuelle:
Anyway this is annoying but it should not block the PR I hope, it's important to get this one on the release wagon since it involves a deprecation.

sciunto:
@emmanuelle It's up to you :)

emmanuelle:
Thanks @sciunto !

PR #4424, opened 2020-01-29
Add gray2rgba and deprecate RGBA support in gray2rgb

rfezzani:
## Description

This PR adds the `gray2rgba` function to `colorconv` module.
This is part of the #4418 effort to add nD support to `colorconv` since it helps deprecating RGBA images conversion using `gray2rgb`.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/color/colorconv.py
```diff
@@ -804,6 +804,46 @@ def rgb2grey(rgb):
...
+        RGBA image. A new dimension of length 4 is added to input
+        image shape.
+
+    Raises
+    ------
+    ValueError
+
+    """
+
+    arr = dtype.img_as_float(image)
```
hmaarrfk: why are you so interested in making the image a float? I don't think there is any loss of information if you leave this as the input dtype.

rfezzani: I don't really know how to manage alpha channel if `image.dtype` is float and `image` values are not between 0 and 1 :-S.

rfezzani: I just realized that I have to check if `0 <= alpha <= 1` :-)

hmaarrfk: I guess I would simply not worry about that.

Junk in means junk out for me.

At the very least, if dtype is uint8, which many images are, then it is very likely OK to keep it as that.

rfezzani: I am not in favor of this "junk in junk out" behavior when it can be avoided.
I understand the uint8 case, but what about the int case? The user will end up with an alpha channel filled with 9223372036854775807!
The majority of the functions of this module converts the input image to float, so I guess that this conversion will not be that surprising for the user.
But well, if a majority votes for no conversion...

jni: @rfezzani `int` images are a big problem for us in general. I recommend you read #3009 for some background if you haven't already. The converse problem is that converting to float will probably result in very tiny values if the image is higher depth than 16-bit.

But I very much would like to preserve the type for uint8 images at least, which is a common use case. Probably 16-bit as well...

hmaarrfk: according to our data types document, setting the alpha channel to 2^32-1 is the right thing to do:

The thing is that data validation isn't cheap, especially with image data. If you want to validate at every step, you are essentially guaranteeing that nobody that cares about performance will use these functions.

For batch image processing, everybody will start to look elsewhere if calls to `min` and `max` are performed multiple times in the library without recourse to remove them.

rfezzani: I know that 2^32-1 is the correct value, I am just saying that in this case, "junk in" results in "huge :hankey: "!
:smile: 
I am not in favor of validation, but conversion is fine since it is used elsewhere in multiple functions in `skimage`.
By the way, I removed the call to `img_as_float`...

skimage/color/colorconv.py
```diff
@@ -804,6 +804,65 @@ def rgb2grey(rgb):
...
+    Returns
+    -------
+    rgba : ndarray
+        RGBA image. A new dimension of length 4 is added to input
+        image shape.
+
+    Raises
+    ------
+    ValueError
+        If ``alpha`` is a scala and is not in ``image`` data type
```
hmaarrfk:
```suggestion
        If ``alpha`` is a scalar and is not in ``image`` data type
```

skimage/color/colorconv.py
```diff
@@ -804,6 +804,65 @@ def rgb2grey(rgb):
...
+    if alpha is None:
+        alpha = alpha_max
+    elif np.isscalar(alpha):
+        if alpha_min > alpha or alpha > alpha_max:
+            raise ValueError("alpha is not in image data type limits.")
+        alpha_cast = arr.dtype.type(alpha)
+        if alpha_cast != alpha:
+            warn("alpha is cast to {}".format(arr.dtype.name))
+        alpha = alpha_cast
+    else:
```
hmaarrfk: I would simply remove all these checks.

skimage/color/colorconv.py
```diff
@@ -804,6 +804,65 @@ def rgb2grey(rgb):
...
+def gray2rgba(image, alpha=None):
+    """Create a RGBA representation of a gray-level image.
+
+    Parameters
+    ----------
+    image : array_like
+        Input image.
+    alpha : array_like, optional
+        Alpha channel of the output image. It may be a scalar or an
+        array with the same shape as ``image``. If not specified it is
```
hmaarrfk:
```suggestion
        array that can be broadcast to ``image``. If not specified it is
```

skimage/color/colorconv.py
```diff
@@ -804,6 +804,65 @@ def rgb2grey(rgb):
...
+
+    """
+
+    arr = np.asarray(image)
+
+    alpha_min, alpha_max = dtype_limits(arr, clip_negative=False)
+
+    if alpha is None:
+        alpha = alpha_max
+    elif np.isscalar(alpha):
```
hmaarrfk: Again, I would simply remove all these checks.

Checking so often will cause headaches. The core functionality, broadcasting and adding a channel at the end remains the same, even if the user provides `alpha=2.0` for floats.

skimage/color/colorconv.py
```diff
@@ -804,6 +804,65 @@ def rgb2grey(rgb):
...
+        if alpha_min > alpha or alpha > alpha_max:
+            raise ValueError("alpha is not in image data type limits.")
+        alpha_cast = arr.dtype.type(alpha)
+        if alpha_cast != alpha:
```
hmaarrfk: The problem here is that you can't check for equality in floating point numbers.

```
In [13]: a = 0.3                                                                

In [14]: np.float32(a) == np.float64(a)                                         
Out[14]: False

In [15]: np.float32(a) == np.float64(np.float32(a))                             
Out[15]: True

In [16]: np.float64(np.float32(a)) == np.float64(a)                             
Out[16]: False
```

skimage/color/colorconv.py
```diff
@@ -804,6 +804,65 @@ def rgb2grey(rgb):
...
+    """
+
+    arr = np.asarray(image)
+
+    alpha_min, alpha_max = dtype_limits(arr, clip_negative=False)
+
+    if alpha is None:
+        alpha = alpha_max
+    elif np.isscalar(alpha):
+        if alpha_min > alpha or alpha > alpha_max:
```
hmaarrfk: I don't think this check is necessary for floating point numbers.

For scalars, maybe.

skimage/color/colorconv.py
hmaarrfk: Numpy will check if they are broadcastable for you.

skimage/color/colorconv.py
rfezzani: You are right, the right way to do that is to use `np.can_cast`

skimage/color/colorconv.py
rfezzani: For floating point numbers, `dtype_limits` are [-1, 1]...
Even if we don't want to restrict images range, we have to be careful about the meaning of an alpha channel. Even if the "junk in -> junk out" credo is reasonable, the user must at least be warned that he provided what we consider junk ;-)

skimage/color/colorconv.py
```diff
@@ -804,6 +804,60 @@ def rgb2grey(rgb):
...
+            warn("alpha = {} is not in image data type limits."
+                 .format(alpha))
+
+    if not np.can_cast(alpha, image.dtype):
+        warn("alpha can't be safely cast to image dtype {}"
+             .format(arr.dtype.name))
+
+    rgba = np.empty(arr.shape + (4, ), dtype=arr.dtype)
+    rgba[..., :3] = arr[..., np.newaxis]
+    rgba[..., 3] = alpha
```
hmaarrfk: Do we want to convert alpha to the appropriate units here? Using the convert function

skimage/color/colorconv.py
```diff
@@ -804,6 +804,60 @@ def rgb2grey(rgb):
...
+    ValueError
+        If ``alpha`` is an array and have a different shape then
+        ``image``.
+    ValueError
+        If ``alpha`` is an array and have a different data type then
+        ``image``.
+
+    """
+
+    arr = np.asarray(image)
```
hmaarrfk: I think this is the job of a static type checker, or a linter. No need to always cast this to an array here.

rfezzani: All right, but `image` has to be documented as `array`, or `ndarray` then... This also prevent for example from passing only one value or a small list of values...
The only thing preventing from using `asarray` is if we want to use `dask`, but does this function really need to be accelerated?
I think that the small overhead of `asarray` worth it...

skimage/color/colorconv.py
rfezzani: i sticked to what you advised before: "junk in -> junk out"...
I am not against the conversion here, I was even more strict in the first release of this PR since it was raising an error if alpha is an array of different dtype then image....

hmaarrfk: I'll have to check the truth table for np.can_cast

rfezzani:
Anyone can review this PR? It is prerequisite for #4418

skimage/color/colorconv.py
hmaarrfk: This is the only thing I would like to bring to the attention of the 2nd reviewer.

skimage/color/colorconv.py
```diff
@@ -811,6 +811,47 @@ def rgb2grey(rgb):
...
+
+    alpha_min, alpha_max = dtype_limits(arr, clip_negative=False)
+
+    if alpha is None:
+        alpha = alpha_max
+    elif np.isscalar(alpha):
+        if alpha_min > alpha or alpha > alpha_max:
+            warn("alpha = {} is not in image data type limits."
+                 .format(alpha))
```
jni:
```suggestion
            warn(f'alpha={alpha} is not in image dtype '
                 f'{arr.dtype.name} limits [{alpha_min}, {alpha_max}]',
                 stacklevel=2)
```

skimage/color/colorconv.py
jni: So, my take is that we shouldn't raise for out of range (e.g. `alpha=2.5` in a float image), but we *should* raise on under/overflow, such as `alpha=4096` for a uint8 image. What do you think about this suggestion @rfezzani?

skimage/color/colorconv.py
```diff
@@ -811,6 +811,47 @@ def rgb2grey(rgb):
...
+    """Create a RGBA representation of a gray-level image.
+
+    Parameters
+    ----------
+    image : array_like
+        Input image.
+    alpha : array_like, optional
+        Alpha channel of the output image. It may be a scalar or an
+        array that can be broadcast to ``image``. If not specified it is
+        set to the maximum limit correspondin to ``image`` data type.
```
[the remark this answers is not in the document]
stefanv: Also not quite accurate: this would be 1 for floating point images, I suppose?

skimage/color/colorconv.py
stefanv: I agree that we don't really want to do this unless necessary.  That said, this problem is so pervasive across all of scikit-image that, for now, it's probably safer to make the asarray call (we don't generally support dask arrays, e.g.).

skimage/color/colorconv.py
stefanv: I guess, now that we've decided to loosen up on floating point limits in general, this is a problematic situation.  I don't think there is a "correct" solution now, but limiting the alpha value specified may go against what the user wants.

skimage/color/colorconv.py
stefanv: I agree with @jni.

Also, I think no explicit conversion needs to be done by us.

skimage/color/colorconv.py
```diff
@@ -811,6 +811,47 @@ def rgb2grey(rgb):
...
+        if alpha_min > alpha or alpha > alpha_max:
+            warn("alpha = {} is not in image data type limits."
+                 .format(alpha))
+
+    if not np.can_cast(alpha, image.dtype):
+        warn("alpha can't be safely cast to image dtype {}"
+             .format(arr.dtype.name))
+
+    rgba = np.empty(arr.shape + (4, ), dtype=arr.dtype)
+    rgba[..., :3] = arr[..., np.newaxis]
```
stefanv: What is going on here?  Where is the necessary call to `gray2rgb`?

skimage/color/tests/test_colorconv.py
```diff
@@ -645,20 +646,99 @@ def test_gray2rgb_alpha():
...
+    img_f64 = np.random.random((5, 5))
+    img_f32 = img_f64.astype('float32')
+    img_u8 = img_as_ubyte(img_f64)
+    img_int = img_u8.astype(int)
+
+    for img in [img_f64, img_f32, img_u8, img_int]:
+        assert gray2rgba(img).dtype == img.dtype
+
+
+def test_gray2rgba_alpha():
```
stefanv: There's no check for the image *content* here, only alpha, which is why it passes incorrectly.

stefanv:
Disregard comments you received on rgb2gray; my brain was having a restful moment.

skimage/color/colorconv.py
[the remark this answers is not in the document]
jni: @stefanv the line is supposed to be referring to our internal dtype limits, as specified in our docs and in the `dtype_limits` function. Could you suggest a rephrase that would make that clearer?

skimage/color/colorconv.py
rfezzani: @jni and @stefanv, this is already checked [here](https://github.com/scikit-image/scikit-image/pull/4424/files#diff-9f4dedd98e7cca606dec9f0b792c1d94R845) ;-)

skimage/color/colorconv.py
rfezzani: Do you mean that this check is no more necessary?

skimage/color/tests/test_colorconv.py
rfezzani: @stefanv, any numerical array like is valid input for `gray2rgba`, so I don't see any non obvious test for this purpose...

stefanv:
```
assert_equal(output[..., :3] = gray2rgb(input))
```

rfezzani: The requested test are implemented now @stefanv. Thank you for your suggestion.

rfezzani:
Anymore requested changes for this PR? Thank you

stefanv:
Thanks, @rfezzani!

rfezzani:
:tada: Thank you @stefanv

Issue #4406, opened 2020-01-22
API: disk vs circle vs circle_perimeter


sciunto:
## Description

In [morphology](https://scikit-image.org/docs/stable/api/skimage.morphology.html#disk) disk means a full circle
In [Hough transform](https://scikit-image.org/docs/stable/api/skimage.transform.html#hough-circle) circle means the edge of a circle
In [draw.circle](https://scikit-image.org/docs/stable/api/skimage.draw.html#circle) means a full circle
In [draw.circle_perimeter](https://scikit-image.org/docs/stable/api/skimage.draw.html#circle-perimeter) and [draw.circle_perimeter_aa](https://scikit-image.org/docs/stable/api/skimage.draw.html#circle-perimeter_aa) circle perimeter means the edge of a circle.
In [measure circleModel](https://scikit-image.org/docs/stable/api/skimage.measure.html#circlemodel), circle means edge of a circle
In [measure circle_level_set ](https://scikit-image.org/docs/stable/api/skimage.segmentation.html#circle-level-set) circle is a full circle.

I admit this can be disturbing.

I recommend to rename
* [x] draw.circle -> draw.disk (#4428)
* [ ] draw.circle_perimeter -> draw.circle
* [ ] draw.circle_perimeter_aa -> draw.circle_aa
* [x] skimage.segmentation.circle_level_set -> skimage.segmentation.disk_level_set (#4421) 

In my mind, disk = circle filled up with pixels, circle = mathematical line.

What do you think @scikit-image/core

jni:
@sciunto I'm :+1: on this proposal! Great catch(es)!

hmaarrfk:
How will you handle the deprecation cycle for the new circle? Will you do it in a single shot or two?

alexdesiqueira:
Great catches indeed @sciunto, one more step for our consistency!

sciunto:
@hmaarrfk Good question. The number/type of arguments for the new disk and the new circle functions are the same. Therefore, it's hard to detect users who have an outdated code, embarrassing for a single shot.

I would suggest:
* Step 1: warn that draw.circle is deprecated. Implement draw.disk. (For 2 versions).
* Step 2: move the code in draw.circle_perimeter in draw.circle. Deprecated the old function. (For 2 versions).

I think this will give the best user experience.

hmaarrfk:
[...]

PR #4428, opened 2020-01-30
MAINT: rename draw.circle() to draw.disk()

sciunto:
## Description

`draw.circle` draws a filled circle aka a disk. By renaming the function, we make the nomenclature mathematically correct, as well as more consistent with the rest of the library.

See #4406 for details.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/draw/draw.py
+++ b/skimage/draw/draw.py
@@ -1,3 +1,4 @@
+import warnings
 import numpy as np
 
 from .._shared._geometry import polygon_clip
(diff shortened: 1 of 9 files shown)


skimage/draw/draw.py
```diff
@@ -148,28 +149,66 @@ def circle(r, c, radius, shape=None):
...
+    Returns
+    -------
+    rr, cc : ndarray of int
+        Pixel coordinates of disk.
+        May be used to directly index into an array, e.g.
+        ``img[rr, cc] = 1``.
+
+    Warns
+    -----
+    Deprecated:
```
hmaarrfk: do you want to mangle the docstring instead?

```
if disk.__doc__ is not None:
    circle.__doc__ = disk.__doc__ + """ Your doc string here"""
```

skimage/draw/draw.py
```diff
@@ -148,28 +149,66 @@ def circle(r, c, radius, shape=None):
...
+    return disk(r, c, radius, shape=shape)
+
+
+def disk(r, c, radius, *, shape=None):
+    """Generate coordinates of pixels within circle.
+
+    Parameters
+    ----------
+    r, c : double
+        Center coordinate of disk.
```
jni: @sciunto since we are making a new function here, how do you feel about starting with the "nD style" API?

```python
def disk(center, radius, *, shape=None):
```

Radius could be an int now but a vector in future versions where we might want to draw disks in an nD space...

On the other hand, maybe it's better to upgrade the whole module in one go, in which case then it's better to use a different PR for that. I'll leave the decision to you!

sciunto: Excellent remark! I think I can change in this PR the signature. Imo, one deprecation is better than two, even if in the mean time it is more heterogeneous. The burden caused by updating existing code is more important.

I'll change the signature asap.

sciunto: @jni can you check https://github.com/scikit-image/scikit-image/pull/4428/commits/070bf872e67d2a1856f46cf0ea60d8b442c229e9 (can you confirm that the order of the center coordinates is correct?

skimage/draw/draw.py
```diff
@@ -148,28 +149,66 @@ def circle(r, c, radius, shape=None):
...
+                  FutureWarning, stacklevel=2)
+    return disk((r, c), radius, shape=shape)
+
+
+def disk(center, radius, *, shape=None):
+    """Generate coordinates of pixels within circle.
+
+    Parameters
+    ----------
+    center : tuple
```
rfezzani: `array_like`?

jni:
@sciunto yep, the ordering is correct. Thanks for these changes! :tada:

jni:
@sciunto note also that I have expanded the text of the PR description just a little bit, as @NelleV found it frustrating to have to do extra clicks to get the most basic context when creating the release notes.

sciunto:
thanks @jni

hmaarrfk:
hmm, this seems to have caused warnings to appear in draw_random_shape.

hmaarrfk:
do we have somewhere where we can store: "definitions" for scikit-image? like

> a circle is just the perimeter of a disk?

sciunto:
Do you mean [this](https://scikit-image.org/docs/stable/glossary.html)?

hmaarrfk:
yeah, i feel like when decisions like this get made, they should come with corresponding documentation in the gloassary

hmaarrfk:
we can push for taht in the future. but disk/circle seems like a good one too

sciunto:
+1

Issue #3955, opened 2019-06-03
support nD arrays in skimage.filters.rank


d-v-b:
## Description

Currently the rank filters are all 2D-only; it would be great if these filters worked for nD-arrays. I don't know the underlying code well at all (if I did, this would be a PR...); is there anything blocking these filters working in higher dimensions, besides the fact that nobody has written the code?

stefanv:
No, if anything I think the rank filters benefit you even more in 3D. We'd 
love a PR.


I also noticed that the current implementation could do with some better 
internal documentation. Eg, the "pop" variable should clearly be delineated 
as "population", not "queue pop".

d-v-b:
Having looked at the code a bit more, the [core moving-histogram code](https://github.com/scikit-image/scikit-image/blob/master/skimage/filters/rank/core_cy.pyx#L45) is _very_ specific to the number of dimensions -- in the current code, there are several instances where 1 variable is defined per dimension; converting this to N dimensions seems tricky. Can you recommend any good examples of cythonized nD-filtering that I could use as a reference?

grlee77:
nD support is still missing, but 3D support was added by #4435. As mentioned above, refactoring to a general nD case would be a decent amount of work. I'm not sure if there are sufficient number of use cases beyond 2D and 3D to justify this?

We will likely have an nD version of this for GPUs in [cuCIM](https://github.com/rapidsai/cucim) at some point. That is mainly because, for cuCIM, I would start by adapting kernels from out `cupyx.scipy.ndimage` rank filters code which is already nD.

PR #4435, opened 2020-02-03
3D Rank Filters

ryanlu41:
## Description
Added 3D support for filters.rank.equalize and filters.rank.otsu. Created a 3D traversal algorithm as well as frameworks for increased 3D support and testing with the module. This enhancement helps with issue #3955 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


jni:
Thanks @ryanlu41 (and all others who authored commits here!) for tackling this! 3D rank filters have been on my wish list for a very long time. I don't have time to do a deep review right now. The only thing I would say right now is that we might want to wait for #3945 to get merged, and then add the dataset to that, since skimage wheels are already pretty big.

Anyway, thanks again, I look forward to merging this! :grinning:

ryanlu41:
Changed to WIP as we try to get the entire module into 3d

emmanuelle:
Thank you very much for your work. Please ping us here when you want a review :-).

xialuoke4062:
> Thank you very much for your work. Please ping us here when you want a review :-).

Thank you for following up. The core algorithm was passing all tests up to [aaeb2ff](https://github.com/scikit-image/scikit-image/pull/4435/commits/aaeb2ffb8646336b0c7f6ca6b3f49eadf7296804), but then the Travis kept having minor problems that did not exist previously. I'll continue trying but those might be easy fixes that you are potentially familiar with?

emmanuelle:
I think tests are failing because you also need to update `skimage/data/_registry.py` to register the new data file. @skimage/core should we open a `skimage/skimage-data` repo to store new datasets?

xialuoke4062:
> I think tests are failing because you also need to update `skimage/data/_registry.py` to register the new data file. @skimage/core should we open a `skimage/skimage-data` repo to store new datasets?

Thanks for the prompt reply. I've tried updating the _registry.py file in [0dff5c5](https://github.com/scikit-image/scikit-image/pull/4435/commits/0dff5c5c4ab32a4d5b4d8191372833f268d0b130) but it seems like I cannot generate the correct hash with my local repo (without adding new data to master first)

stefanv:
> data

I'd be fine with that.

stefanv:
> I think tests are failing because you also need to update `skimage/data/_registry.py` to register the new data file. @scikit-image/core should we open a `skimage/skimage-data` repo to store new datasets?

I'd be fine with that.

jni:
I see now why @emmanuelle pinged me in #4538. I think creating a repo for data now is a fine stopgap. It could be on GitHub, but in case we want it on gitlab for the added bandwith I went ahead created it:

https://gitlab.com/scikit-image/data

xialuoke4062:
> It could be on GitHub, but in case we want it on gitlab for the added bandwith I went ahead created it:
> 
> https://gitlab.com/scikit-image/data

Do I need to create another image_fetcher for this gitlab url? Currently there is only one cells.tif item in registry_urls; maybe it could be moved to gitlab?

jni:
@xialuoke4062 

> Do I need to create another image_fetcher for this gitlab url?

No, rather, that `brain.tiff` should be placed in that repo (and removed from this PR), then you would add the hash to the `registry` dictionary and the URL in gitlab to the `registry_urls` dictionary. The (current) image_fetcher will take care of the rest!

The nice thing is that we can do this now, and whatever we decide later (github, github pages, stick with gitlab), we can update the URL and keep everything else the same.

stefanv:
> @xialuoke4062
> 
> > Do I need to create another image_fetcher for this gitlab url?
> 
> No, rather, that `brain.tiff` should be placed in that repo (and removed from this PR), then you would add the hash to the `registry` dictionary and the URL in gitlab to the `registry_urls` dictionary. The (current) image_fetcher will take care of the rest!
> 
> The nice thing is that we can do this now, and whatever we decide later (github, github pages, stick with gitlab), we can update the URL and keep everything else the same.

Also note that this PR now *has* to be squashed before merge.

xialuoke4062:
> The nice thing is that we can do this now, and whatever we decide later (github, github pages, stick with gitlab), we can update the URL and keep everything else the same.

The hash of downloaded "data/rank_filters_tests_3d.npz" file from gitlab keeps changing; perhaps it's because it's still on the patch-1 branch instead of master?

xialuoke4062:
> The nice thing is that we can do this now, and whatever we decide later (github, github pages, stick with gitlab), we can update the URL and keep everything else the same.

The hash of "data/rank_filters_tests_3d.npz" downloaded from website seems to be unstable, potentially due to [this](https://github.com/numpy/numpy/issues/9439)? I'm curious to find out how the other npz files hash were created previously? Many thanks!!

stefanv:
> > The nice thing is that we can do this now, and whatever we decide later (github, github pages, stick with gitlab), we can update the URL and keep everything else the same.
> 
> The hash of my new "data/rank_filters_tests_3d.npz" downloaded from website seems to be indeterministic, potentially due to [this](https://github.com/numpy/numpy/issues/9439)? I'm curious to find out how the other npz files hash were created previously? Many thanks!!

Once the file is uploaded to gitlab, the "raw" link to it (i.e., not to the file page, but to the actual data) should return a constant hash.

xialuoke4062:
I believe all tests are passing except the one failing on master. Thank you for your time and review! @emmanuelle @stefanv @jni

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+                    if t_n[p, r, c]:
+                        se_n_r[num_se_n] = r - centre_r
+                        se_n_c[num_se_n] = c - centre_c
+                        se_n_p[num_se_n] = p - centre_p
+                        num_se_n += 1
+                    if t_s[p, r, c]:
+                        se_s_r[num_se_s] = r - centre_r
+                        se_s_c[num_se_s] = c - centre_c
+                        se_s_p[num_se_s] = p - centre_p
+                        num_se_s += 1
```
jni, line 444: This could be moved to its own function.

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+
+        se_n_r = <Py_ssize_t*>malloc(se_size)
+        se_n_c = <Py_ssize_t*>malloc(se_size)
+        se_n_p = <Py_ssize_t*>malloc(se_size)
+
+        se_s_r = <Py_ssize_t*>malloc(se_size)
+        se_s_c = <Py_ssize_t*>malloc(se_size)
+        se_s_p = <Py_ssize_t*>malloc(se_size)
+
+        histo = <Py_ssize_t*>malloc(n_bins * sizeof(Py_ssize_t))
```
jni, line 394: As above: this would be nicer as a numpy array created outside the loop, which then allows constructs like `histo[:] = 0`, not to mention `histo = np.zeros(n_bins, dtype=np.intp)`. Though, tell me if that causes a big hit on performance. I think most of the performance penalties can be removed with `@cython.boundscheck(False)` and `@cython.wraparound(False)`

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+                        num_se_n += 1
+                    if t_s[p, r, c]:
+                        se_s_r[num_se_s] = r - centre_r
+                        se_s_c[num_se_s] = c - centre_c
+                        se_s_p[num_se_s] = p - centre_p
+                        num_se_s += 1
+
+        for p in range(planes):
+            for i in range(n_bins):
+                histo[i] = 0
```
jni, line 448: if `histo[:] = 0` (with histo being a numpy array) is too slow, I would still make this its own private, potentially inline function:

```python
_zero_out_array(histo, size=n_bins)
```

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+            pop = 0
+
+            for r in range(srows):
+                for c in range(scols):
+                    for j in range(splanes):
+                        rr = r - centre_r
+                        cc = c - centre_c
+                        pp = j - centre_p + p
+
+                        # change this to accept 3D
```
jni, line 458: What does this comment mean? Is it outdated?

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+                        num_se_s += 1
+
+        for p in range(planes):
+            for i in range(n_bins):
+                histo[i] = 0
+            pop = 0
+
+            for r in range(srows):
+                for c in range(scols):
+                    for j in range(splanes):
```
jni, line 453: As above, I think this could be an inlined function, which would help readability. For example:

```python
for p in range(planes):
    _zero_out_histogram(histo, size=n_bins)
    _build_initial_histogram_from_neighborhood(image, selem, histo)
    # ... mainloop
```

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+            kernel(&out[p, r, c, 0], odepth, histo, pop, image[p, r, c],
+                   n_bins, mid_bin, p0, p1, s0, s1)
+
+        # main loop
+
+            r = 0
+            for even_row in range(0, rows, 2):
+
+                # ---> west to east
+                for c in range(1, cols):
```
jni, line 476: I don't understand why things start at 1 here?

jni: I see that you have applied the kernel at r=0, c=0, before you even start, above, but that only applies to the top-left pixel? What about the pixels at [p, 2, 0], [p, 4, 0], ...?

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+
+                for j in range(num_se_n):
+                    rr = r + se_n_r[j] - 1
+                    cc = c + se_n_c[j]
+                    pp = p + se_n_p[j]
+                    if is_in_mask_3D(rows, cols, planes, rr, cc, pp,
+                                     mask_data):
+                        histogram_decrement(histo, &pop, image[pp, rr, cc])
+
+                kernel(&out[p, r, c, 0], odepth, histo, pop, image[p, r, c],
```
jni, line 562: what is the `0` for in `out[p, r, c, 0]`?

skimage/filters/rank/generic.py
```diff
@@ -146,6 +150,57 @@ def _preprocess_input(image, selem=None, out=None, mask=None, out_dtype=None,
     return image, selem, out, mask, n_bins
 
 
+def _handle_input_3D(image, selem, out, mask, out_dtype=None, pixel_size=1):
```
jni, line 153: Can you add a docstring here that explains all the checks that happen in this function?

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+                                     mask_data):
+                        histogram_increment(histo, &pop, image[pp, rr, cc])
+
+                for j in range(num_se_n):
+                    rr = r + se_n_r[j] - 1
+                    cc = c + se_n_c[j]
+                    pp = p + se_n_p[j]
+                    if is_in_mask_3D(rows, cols, planes, rr, cc, pp,
+                                     mask_data):
+                        histogram_decrement(histo, &pop, image[pp, rr, cc])
```
jni, line 515: This pattern is repeated lots of times, where only `num_se_*` and `se_*_{p,r,c}` change, and the rest of the processing is the same. So, it could be a function! `_update_histogram(num_se_s, num_se_n, se_s, se_n, ...)`... And, if you store all the data in a bigger array where e/w/n/s are encoded as indices along an array axis, it can be even simpler, `_update_histogram(num_se, se, axis_inc=1, axis_dec=2)` (for example). In fact, since you are always decrementing/incrementing along the same direction, if you store the data in the position (e/n/w/s), you can always assume axis_inc=axis_dec + 2.

Does this all make sense? I think it would make the inner loop a lot easier to read...

skimage/filters/rank/generic.py
```diff
@@ -146,6 +150,57 @@ def _preprocess_input(image, selem=None, out=None, mask=None, out_dtype=None,
...
+
+    check_nD(image, [2, 3])
```
jni, line 155: why 2? shouldn't it be always 3D in this function?

skimage/filters/rank/generic.py
```diff
@@ -1088,28 +1161,29 @@ def entropy(image, selem, out=None, mask=None, shift_x=False,
...
-    shift_x, shift_y : int, optional
+    shift_x, shift_y, shift_z : int
         Offset added to the structuring element center point. Shift is bounded
         to the structuring element sizes (center must be inside the given
         structuring element).
 
     Returns
     -------
-    out : 2-D array (same dtype as input image)
+    out : (N, M[,P]) ndarray (same dtype as input image)
```
jni, line 1186: We typically write `([P,] M, N)`.

skimage/filters/rank/tests/test_rank.py
```diff
@@ -351,11 +401,26 @@ def test_compare_8bit_vs_16bit(self, method):
         image16 = image8.astype(np.uint16)
         assert_equal(image8, image16)
 
+        np.random.seed(0)
+        volume8 = np.random.randint(128, high=256,
+                                    size=(10, 10, 10), dtype=np.uint8)
+        volume16 = volume8.astype(np.uint16)
+        assert_equal(volume8, volume16)
```
jni, line 408: this assertion isn't necessary

skimage/filters/rank/tests/test_rank.py
```diff
@@ -340,6 +380,16 @@ def test_compare_8bit_unsigned_vs_signed(self):
...
             assert_equal(out_u, out_s)
 
+        methods_3d = ['equalize', 'otsu']
+
+        for method in methods_3d:
+            func = getattr(rank, method)
+            out_u = func(volume_u, ball(3))
+            with expected_warnings(["Possible precision loss"]):
+                out_s = func(volume_s, ball(3))
+            assert_equal(out_u, out_s)
```
jni, line 390: Since both the setup and the test for 3D are separate, you might as well put this in a separate test method, e.g. `def test_compare_8bit_unsigned_vs_signed_3d(self):`

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
...
+
+fig, axes = plt.subplots(nrows=2, ncols=2, figsize=(12, 12),
+                         sharex=True, sharey=True)
+ax = axes.ravel()
+
+fig.colorbar(ax[0].imshow(p8[3], cmap=plt.cm.gray), ax=ax[0])
+ax[0].set_title('Original')
+
+fig.colorbar(ax[1].imshow(t_loc_otsu[3], cmap=plt.cm.gray), ax=ax[1])
+ax[1].set_title('Local Otsu ($r=%d$)' % radius)
```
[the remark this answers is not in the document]
xialuoke4062: This and many following suggestions are on templates copied from the 2D version code; should we change the 2D version accordingly?

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
...
+
+ax[2].imshow(p8[3] >= t_loc_otsu[3], cmap=plt.cm.gray)
+ax[2].set_title('Original >= local Otsu' % t_glob_otsu)
```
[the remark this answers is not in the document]
xialuoke4062: copied from 2D

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
...
+
+ax[3].imshow(glob_otsu[3], cmap=plt.cm.gray)
+ax[3].set_title('Global Otsu ($t=%d$)' % t_glob_otsu)
```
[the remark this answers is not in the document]
xialuoke4062: 2d?

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
...
+
+for a in ax:
+    a.axis('off')
+
+plt.tight_layout()
```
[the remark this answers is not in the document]
xialuoke4062: 2D?

doc/examples/color_exposure/plot_local_equalize.py
```diff
@@ -93,6 +96,82 @@ def plot_img_and_hist(image, axes, bins=256):
...
+
+# Display results
+fig = plt.figure(figsize=(8, 5))
+axes = np.zeros((2, 3), dtype=np.object)
+axes[0, 0] = plt.subplot(2, 3, 1)
+axes[0, 1] = plt.subplot(2, 3, 2, sharex=axes[0, 0], sharey=axes[0, 0])
+axes[0, 2] = plt.subplot(2, 3, 3, sharex=axes[0, 0], sharey=axes[0, 0])
+axes[1, 0] = plt.subplot(2, 3, 4)
+axes[1, 1] = plt.subplot(2, 3, 5)
+axes[1, 2] = plt.subplot(2, 3, 6)
```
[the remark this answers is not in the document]
xialuoke4062: 2D?

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
...
+def normalize_pixels(i):
+    i = 256*(i/np.amax(i))
+    return i.astype(np.uint8)
+
+p8 = data.brain()
+
+p8 = normalize_pixels(p8)
+
+radius = 50
+selem = ball(radius)
```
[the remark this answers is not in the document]
xialuoke4062: This and many following suggestions are on reoccurring templates copied from the 2D version code; should we change the 2D version accordingly?

skimage/data/__init__.py
```diff
@@ -930,3 +931,25 @@ def lfw_subset():
...
+    Notes
+    -----
+    The 3D volume consists of 10 layers from the larger volume.
+
+    References
+    ----------
+    .. [1] https://graphics.stanford.edu/data/voldata/
+
+    """
+    return _load("data/brain.tiff")
```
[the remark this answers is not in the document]
xialuoke4062: I think the other functions all call _load only because _load calls _fetch internally. If we are going to change this one, should we change the rest of the functions?

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+
+    # check that structuring element center is inside the element bounding box
+    assert centre_r >= 0
+    assert centre_c >= 0
+    assert centre_p >= 0
+    assert centre_r < srows
+    assert centre_c < scols
+    assert centre_p < splanes
+
+    cdef Py_ssize_t mid_bin = n_bins / 2
```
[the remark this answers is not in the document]
xialuoke4062: 2D change?

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+    cdef unsigned char[:, :, :] t_e = (np.diff(t, axis=2) < 0).view(np.uint8)
+
+    t = np.dstack((np.zeros((selem.shape[0], selem.shape[1], 1)), selem))
+    cdef unsigned char[:, :, :] t_w = (np.diff(t, axis=2) > 0).view(np.uint8)
+
+    t = np.hstack((selem, np.zeros((selem.shape[0], 1, selem.shape[2]))))
+    cdef unsigned char[:, :, :] t_s = (np.diff(t, axis=1) < 0).view(np.uint8)
+
+    t = np.hstack((np.zeros((selem.shape[0], 1, selem.shape[2])), selem))
+    cdef unsigned char[:, :, :] t_n = (np.diff(t, axis=1) > 0).view(np.uint8)
```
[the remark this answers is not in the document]
xialuoke4062: I can look into this but maybe @ryanlu41 could give a better answer?

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+        se_w_c = <Py_ssize_t*>malloc(se_size)
+        se_w_p = <Py_ssize_t*>malloc(se_size)
+
+        se_n_r = <Py_ssize_t*>malloc(se_size)
+        se_n_c = <Py_ssize_t*>malloc(se_size)
+        se_n_p = <Py_ssize_t*>malloc(se_size)
+
+        se_s_r = <Py_ssize_t*>malloc(se_size)
+        se_s_c = <Py_ssize_t*>malloc(se_size)
+        se_s_p = <Py_ssize_t*>malloc(se_size)
```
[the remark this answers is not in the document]
xialuoke4062: > It's also unclear to me why these need to be malloc'ed within a gil-free block, rather than stored in a numpy array created outside of the nogil block? I also wonder if it would not be easier to store these elements as an array of shape `(4, 3, se_size)` rather than as separate variables. What do you think about this idea?

I believe we copied this code template from the 2D version above. I can look into this, and should we change both version?

skimage/filters/rank/core_cy.pyx
xialuoke4062: Same as the above. They are copied from 2D and I can look into them, perhaps changing both versions as well?

skimage/filters/rank/core_cy.pyx
xialuoke4062: Sure. Change for both 2D and 3D?

skimage/filters/rank/core_cy.pyx
xialuoke4062: Sure. Change for both 2D and 3D?

skimage/filters/rank/core_cy.pyx
xialuoke4062: Sure. Change for both 2D and 3D?

skimage/filters/rank/core_cy.pyx
xialuoke4062: I believe so. @ryanlu41 ?

skimage/filters/rank/core_cy.pyx
xialuoke4062: I believe this is also copied from the 2D version. But I will look into this.

skimage/filters/rank/core_cy.pyx
xialuoke4062: Sure, I can look into this. I guess I should also change the 2D version?

skimage/filters/rank/core_cy.pyx
xialuoke4062: from 2D, will look into it, before that @ryanlu41 ?

skimage/filters/rank/generic.py
xialuoke4062: Sure. @ryanlu41 ?

skimage/filters/rank/generic.py
xialuoke4062: I believe it should be only 3. @ryanlu41 ?

skimage/filters/rank/tests/test_rank.py
xialuoke4062: I'm not sure about the difference between the image assertion versus the volume one. @ryanlu41 ?

xialuoke4062:
> I've left quite a lot of comments about code style/structure, but I believe they should be easy to address. I hope they seem reasonable.

Thank you so much for taking time writing these detailed comments! As you could tell from the replies, many code templates were copied from and modified based on the 2D version; if we were to change them, should we also refactor the 2D codes?
 
> Additionally, I am confused about why you only ported otsu and equalize. As I understand it, now that you have made the contribution to the core, it should be easy to support all the kernels from the current rank filters?

Yes, it should be easy. I was trapped by the travis and hesitant to move on as the PR was successful when it was first submitted. I will add the rest of rank filters as soon as possible. 

> Finally, it would be interesting to test expected results by benchmarking against `ndi.generic_filter`, for example.

Definitely. Should the results go to benchmarks/benchmark_rank.py?

Thank you again for your time!

doc/examples/applications/plot_rank_filters.py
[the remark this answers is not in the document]
jni: @xialuoke4062 apologies for the confusion here and elsewhere. The 2D code was written a long time ago, and our coding practices and standards have evolved since then. Therefore, when adding new code, we might be more careful than before and we might prefer newer practices over consistency with the old code.

In any case, **no**, we don't expect you to fix the old code. Indeed it's preferable if you don't, to keep this already big PR from growing even bigger. If the inconsistency bugs you after this is merged, you could submit another, separate PR improving the 2D code. =) But for now, we just want to make sure that any new code is as good as possible!

Thank you!!!

doc/examples/applications/plot_rank_filters.py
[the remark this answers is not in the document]
jni: As above: don't fix existing code, but let's make sure that any new code is the best we can do!

doc/examples/applications/plot_rank_filters.py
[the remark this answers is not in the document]
jni: No need, but this one is so obvious that you might as well fix it. =)

doc/examples/applications/plot_rank_filters.py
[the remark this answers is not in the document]
jni: No, thank you. =)

doc/examples/applications/plot_rank_filters.py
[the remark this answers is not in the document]
jni: Not necessary in this PR.

doc/examples/color_exposure/plot_local_equalize.py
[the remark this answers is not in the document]
jni: No, thank you, let's just make sure new code is good.

skimage/data/__init__.py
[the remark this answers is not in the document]
jni: Apologies, I didn't see that `_load` calls `_fetch`, I was looking just above at `lfw_subset()`, which calls both. The whole `pooch` thing only started recently, so we are still figuring it out! 😅

skimage/filters/rank/core_cy.pyx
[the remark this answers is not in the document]
jni: No, just new code.

skimage/filters/rank/core_cy.pyx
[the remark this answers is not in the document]
jni: From my interpretation, which I figured out mostly after I wrote this, each plane gets its own rolling histogram, but this histogram is built from scratch for each plane. That is, if you had an image shaped (1000, 5, 5), this algorithm would perform quite poorly. However, most images have reasonably-sized planes, so you amortise the cost of building a rolling histogram after each plane, rather than having a rolling histogram across all axes. So, I think we can leave this for now, but @ryanlu41 let us know if you disagree or if my interpretation is totally incorrect.

Actually, if it *is* correct, it is worth adding a comment at this point that explains this.

skimage/filters/rank/core_cy.pyx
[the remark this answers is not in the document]
jni: As with the other comments, let us focus on improving the 3D version, so that new code added to the library is always better than what came before. The 2D version can be cleaned up in a future pull request if we think it important. Thank you!

skimage/filters/rank/core_cy.pyx
jni: Just 3D.

skimage/filters/rank/core_cy.pyx
jni: 3D only for now (in this PR).

skimage/filters/rank/core_cy.pyx
jni: 3D only.

skimage/filters/rank/core_cy.pyx
jni: 3D.

skimage/filters/rank/core_cy.pyx
jni: I think I figured it out: the 0th pixel gets processed when moving N to S in the last iteration of the loop. Is that right @ryanlu41?

skimage/filters/rank/generic.py
jni: Thank you, and again, apologies that my review is holding you to a higher standard than existing code, which might be surprising! The library has evolved a lot since this code was written, including having many more users and contributors!

skimage/filters/rank/core_cy.pyx
jni: Let's do the 3D version only for now, to keep the code changes as small as possible.

jni: The nice thing about all this, btw, is that I finally understand the 2D version, from having to review your PR. I never took the time to look at the 2D code very closely. =D

doc/examples/applications/plot_rank_filters.py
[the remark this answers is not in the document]
xialuoke4062: > We're debating what the best way to call these things going forward, as `selem` is a specialised term that is not very meaningful to beginners. I would suggest `neighborhood = ball(radius)`.

I realize that seem is almost used every place as a variable name. Do you only want this changed in this website demo file?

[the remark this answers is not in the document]
jni: Yes, only here. You are not responsible for our past mistakes! 😅

skimage/filters/rank/generic.py
ryanlu41: yes, i think that might be left over from when I tried to do 2D and 3D in one function

skimage/filters/rank/core_cy.pyx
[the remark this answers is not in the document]
ryanlu41: Yes, each plane gets its own rolling histogram. I would be happy to optimize it further, I just wanted to get a working implementation for now

skimage/filters/rank/core_cy.pyx
ryanlu41: yes i believe so, apologies

skimage/filters/rank/core_cy.pyx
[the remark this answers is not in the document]
jni: Let’s get a working implementation in first, and worry about optimising later.

skimage/filters/rank/tests/test_rank.py
```diff
@@ -1,11 +1,13 @@
+import os
```
rfezzani, line 1: It looks like `os` is imported but not used.

skimage/filters/rank/tests/test_rank.py
```diff
@@ -1,11 +1,13 @@
...
 import numpy as np
 from skimage._shared.testing import (assert_equal, assert_array_equal,
                                      assert_allclose)
 from skimage._shared import testing
 
+import skimage
```
rfezzani, line 7: I don't see it used neither :-/

emmanuelle:
@ryanlu41 @xialuoke4062 please tell us if the reviewers comments are clear enough :-).

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
...
+
+# t_loc_otsu is an image
+t_loc_otsu = otsu(p8, selem)
```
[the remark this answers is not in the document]
xialuoke4062: I changed `otsu` to `rank.otsu` since rank was at one point imported by the contrast enhancement demo code. 
I cannot really remove the rank filters import because it is also used by the old 2d otsu demo. In fact almost all demo (except contrast enhancement) uses something like `from skimage.filters.rank import xxx`

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
...
+######################################################################
+# The example compares the local threshold with the global threshold in 3D
+
+def normalize_pixels(i):
+    i = 256*(i/np.amax(i))
+    return i.astype(np.uint8)
+
+p8 = data.brain()
+
+p8 = normalize_pixels(p8)
```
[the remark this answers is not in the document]
xialuoke4062: fixed

doc/examples/applications/plot_rank_filters.py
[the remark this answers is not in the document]
xialuoke4062: fixed

doc/examples/applications/plot_rank_filters.py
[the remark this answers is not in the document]
xialuoke4062: I changed `otsu` to `rank.otsu` because it is at one point imported by the contrast enhancement demo code. 
I should not really remove the rank filter import because the old 2d otsu demo also uses it. In fact, almost all filter demos use the format `from skimage.filters.rank import xxx`

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
...
+
+# t_glob_otsu is a scalar
+t_glob_otsu = threshold_otsu(p8)
+glob_otsu = p8 >= t_glob_otsu
+
+fig, axes = plt.subplots(nrows=2, ncols=2, figsize=(12, 12),
+                         sharex=True, sharey=True)
+ax = axes.ravel()
+
+fig.colorbar(ax[0].imshow(p8[3], cmap=plt.cm.gray), ax=ax[0])
```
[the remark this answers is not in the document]
xialuoke4062: fixed

doc/examples/color_exposure/plot_local_equalize.py
```diff
@@ -93,6 +96,82 @@ def plot_img_and_hist(image, axes, bins=256):
...
+
+
+# Load an example image
+img = img_as_ubyte(data.brain())
+
+# Global equalize
+img_rescale = exposure.equalize_hist(img)
+
+# Equalization
+selem = ball(20)
```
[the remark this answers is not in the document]
xialuoke4062: fixed

doc/examples/color_exposure/plot_local_equalize.py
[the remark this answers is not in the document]
xialuoke4062: @jni I apologize for not understanding this change entirely. I modified this block of code to:
`# Display results`
`fig, axes = plt.subplots(2, 3, figsize=(8, 5))`
`axes[0, 1] = plt.subplot(2, 3, 2, sharex=axes[0, 0], sharey=axes[0, 0])`
`axes[0, 2] = plt.subplot(2, 3, 3, sharex=axes[0, 0], sharey=axes[0, 0])`
It works fine in the demo - is it what you wanted?

doc/release/release_dev.rst
```diff
@@ -13,13 +13,44 @@ https://scikit-image.org
...
 introduced NumPy deprecation policy -- NEP 29
-(https://github.com/numpy/numpy/blob/master/doc/neps/nep-0029-deprecation_policy.rst). 
+(https://github.com/numpy/numpy/blob/master/doc/neps/nep-0029-deprecation_policy.rst).
 This release of scikit-image officially supports Python 3.6, 3.7, and
 3.8.
 
 New Features
 ------------
-
+- Added majority rank filter - ``filters.rank.majority``.
```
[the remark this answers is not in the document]
xialuoke4062: I removed this. It was probably pre-v0.16 when we forked it.

skimage/data/__init__.py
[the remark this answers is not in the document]
xialuoke4062: closed

skimage/filters/rank/tests/test_rank.py
xialuoke4062: fixed

skimage/filters/rank/tests/test_rank.py
xialuoke4062: fixed

skimage/filters/rank/tests/test_rank.py
xialuoke4062: @jni I removed the assertion but not sure why; it is the same assertion that 2d uses.

skimage/filters/rank/tests/test_rank.py
xialuoke4062: fixed, did the same to test_compare_ubyte_vs_float_3d.

skimage/filters/rank/generic.py
xialuoke4062: fixed

skimage/filters/rank/generic.py
xialuoke4062: fixed

skimage/filters/rank/generic.py
xialuoke4062: fixed

skimage/filters/rank/core_cy.pyx
```diff
@@ -44,6 +44,21 @@ cdef inline char is_in_mask(Py_ssize_t rows, Py_ssize_t cols,
             return 0
 
 
+cdef inline char is_in_mask_3D(Py_ssize_t rows, Py_ssize_t cols,
+                               Py_ssize_t planes, Py_ssize_t r,
+                               Py_ssize_t c, Py_ssize_t p,
+                               char* mask) nogil:
```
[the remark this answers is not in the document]
xialuoke4062: fixed

skimage/filters/rank/core_cy.pyx
```diff
@@ -272,3 +287,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+    cdef Py_ssize_t srows = selem.shape[1]
+    cdef Py_ssize_t scols = selem.shape[2]
+    cdef Py_ssize_t odepth = out.shape[3]
+
+    cdef Py_ssize_t centre_p = <Py_ssize_t>(selem.shape[0] / 2) + shift_x
+    cdef Py_ssize_t centre_r = <Py_ssize_t>(selem.shape[1] / 2) + shift_y
+    cdef Py_ssize_t centre_c = <Py_ssize_t>(selem.shape[2] / 2) + shift_z
+
+    # check that structuring element center is inside the element bounding box
+    assert centre_r >= 0
```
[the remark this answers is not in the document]
xialuoke4062: fixed

skimage/filters/rank/core_cy.pyx
```diff
@@ -271,3 +286,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+    assert centre_c >= 0
+    assert centre_p >= 0
+    assert centre_r < srows
+    assert centre_c < scols
+    assert centre_p < splanes
```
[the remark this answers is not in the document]
xialuoke4062: fixed

skimage/filters/rank/core_cy.pyx
xialuoke4062: fixed

skimage/filters/rank/core_cy.pyx
```diff
@@ -272,3 +287,293 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+                                                    image[pp, rr, cc])
+
+            r = 0
+            c = 0
+            kernel(&out[p, r, c, 0], odepth, histo, pop, image[p, r, c],
+                   n_bins, mid_bin, p0, p1, s0, s1)
+
+        # main loop
+
+            r = 0
```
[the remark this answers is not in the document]
xialuoke4062: fixed

skimage/filters/rank/core_cy.pyx
xialuoke4062: fixed

xialuoke4062:
> @ryanlu41 @xialuoke4062 please tell us if the reviewers comments are clear enough :-).

Apologize for the delay. Now most parts of the review are fixed, but I am still working on the core _cy.pyx refactoring. I have to admit that I am not super familiar with cython.

skimage/filters/rank/core_cy.pyx
xialuoke4062: @jni I am having some difficulties making `histo = np.zeros(n_bins, dtype=np.intp)` and `histo[:] = 0` to work - I encountered errors like `Cannot assign type 'long' to 'Py_ssize_t *'`

jni: @xialuoke4062 could you push that code so that I can test locally? It can be to a different branch if you want to keep this one clean, or directly to this one — it's easy enough to revert a commit!

skimage/filters/rank/core_cy.pyx
```diff
@@ -272,3 +284,239 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+
+    t = np.hstack((selem, np.zeros((selem.shape[0], 1, selem.shape[2]))))
+    cdef unsigned char[:, :, :] t_s = (np.diff(t, axis=1) < 0).view(np.uint8)
+
+    t = np.hstack((np.zeros((selem.shape[0], 1, selem.shape[2])), selem))
+    cdef unsigned char[:, :, :] t_n = (np.diff(t, axis=1) > 0).view(np.uint8)
+
+    # the current local histogram distribution
+    # cdef Py_ssize_t* histo
+    cdef Py_ssize_t* histo
```
rfezzani, line 350:
```suggestion
    cdef Py_ssize_t [::1] histo = np.zeros(n_bins, dtype=np.intp)
```

skimage/filters/rank/core_cy.pyx
```diff
@@ -272,3 +284,239 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+    se_s_r = np.zeros(se_size, dtype=np.intp)
+    se_s_c = np.zeros(se_size, dtype=np.intp)
+
+    # number of element in each attack border
+    cdef Py_ssize_t num_se_n, num_se_s, num_se_e, num_se_w, num_se_u, num_se_d
+
+    histo = np.zeros(n_bins, dtype=np.intp)
+    # for i in range(n_bins):
+    #     histo[i] = 0
+    histo[:] = 0
```
rfezzani, line 378:
```suggestion
```
No need for initialization anymore after creation with `np.zeros` ;)

skimage/filters/rank/core_cy.pyx
```diff
@@ -272,3 +284,239 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
...
+    se_w_r = np.zeros(se_size, dtype=np.intp)
+    se_w_c = np.zeros(se_size, dtype=np.intp)
+
+    se_n_p = np.zeros(se_size, dtype=np.intp)
+    se_n_r = np.zeros(se_size, dtype=np.intp)
+    se_n_c = np.zeros(se_size, dtype=np.intp)
+
+    se_s_p = np.zeros(se_size, dtype=np.intp)
+    se_s_r = np.zeros(se_size, dtype=np.intp)
+    se_s_c = np.zeros(se_size, dtype=np.intp)
```
rfezzani, line 370:
```suggestion
    cdef Py_ssize_t se_size = srows * scols * splanes
    cdef Py_ssize_t [::1] se_e_r = np.empty(se_size, dtype=np.intp)
    cdef Py_ssize_t [::1] se_e_c = np.empty(se_size, dtype=np.intp)
    cdef Py_ssize_t [::1] se_e_p = np.empty(se_size, dtype=np.intp)

    cdef Py_ssize_t [::1] se_w_p = np.empty(se_size, dtype=np.intp)
    cdef Py_ssize_t [::1] se_w_r = np.empty(se_size, dtype=np.intp)
    cdef Py_ssize_t [::1] se_w_c = np.empty(se_size, dtype=np.intp)

    cdef Py_ssize_t [::1] se_n_p = np.empty(se_size, dtype=np.intp)
    cdef Py_ssize_t [::1] se_n_r = np.empty(se_size, dtype=np.intp)
    cdef Py_ssize_t [::1] se_n_c = np.empty(se_size, dtype=np.intp)

    cdef Py_ssize_t [::1] se_s_p = np.empty(se_size, dtype=np.intp)
    cdef Py_ssize_t [::1] se_s_r = np.empty(se_size, dtype=np.intp)
    cdef Py_ssize_t [::1] se_s_c = np.empty(se_size, dtype=np.intp)
```
You can find more doc about numpy usage in cython [here](https://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html#numpy-tutorial) ;)

xialuoke4062:
> Thank you @xialuoke4062, I left some suggestions trying to fix CI failures.

Thank you for your help! I have to look back into this in a few days.

jni:
@ryanlu41 @xialuoke4062 please let me know if you would still like to complete this — if not, we can take over.

xialuoke4062:
> @ryanlu41 @xialuoke4062 please let me know if you would still like to complete this — if not, we can take over.

Sorry for the delay - we will try to finish it in the next week or two.

jni:
Absolutely no worries, I was just checking in! Thank you for your contribution here! 😊

xialuoke4062:
> Absolutely no worries, I was just checking in! Thank you for your contribution here! 😊

Apologize for the delay. I think this refactoring commit is the last step for this PR - hopefully, it matches the requirements. 
Thank you again for your patience!

jni:
@xialuoke4062 thanks! I merged on gitlab. I think we'll probably want to change the structure going forward, but that can come later. In the meantime, could you update this PR with an empty commit to trigger new checks? You can use `git commit -a --allow-empty -m "empty commit to trigger CI`.

emmanuelle:
Looks like the execution of the gallery example doc/examples/color_exposure/plot_local_equalize.py segfaults; can you reproduce the segfault locally?

xialuoke4062:
In the last commit, the 3D demo is commented out and no brain.tiff data is downloaded in the script, but the error lasted. 
In the meantime, our forked repo only produced the same failure as the scikit-image:master. When I pull from the master, that failure already exists. So I think I should pull again and try after the master's failure is fixed. 
Thank you again for your time and patience!

doc/examples/color_exposure/plot_local_equalize.py
```diff
@@ -67,7 +70,7 @@ def plot_img_and_hist(image, axes, bins=256):
 img_rescale = exposure.equalize_hist(img)
 
 # Equalization
-selem = disk(30)
+selem = disk(1)
```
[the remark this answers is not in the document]
xialuoke4062: This was me trying to debug the travis failure on this file that cannot be reproduced locally.

doc/examples/color_exposure/plot_local_equalize.py
```diff
@@ -96,3 +99,74 @@ def plot_img_and_hist(image, axes, bins=256):
 # prevent overlap of y-axis labels
 fig.tight_layout()
 plt.show()
```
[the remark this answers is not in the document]
xialuoke4062: I'm not sure - it was left unchanged from the older version

doc/examples/color_exposure/plot_local_equalize.py
```diff
@@ -96,3 +99,74 @@ def plot_img_and_hist(image, axes, bins=256):
...
+
+
+# ######################################################################
```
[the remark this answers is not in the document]
xialuoke4062: It was for debugging purpose as well.

doc/release/release_dev.rst
```diff
@@ -11,10 +11,45 @@ For more information, examples, and documentation, please visit our website:
...
+- Added 3D implementation of rank sum filter - ``filters.rank.sum``.
+# - Added 3D implementation of rank sum_bilateral filter - ``filters.rank.sum_bilateral``.
+# - Added 3D implementation of rank sum_percentile filter - ``filters.rank.sum_percentile``.
+- Added 3D implementation of rank threshold filter - ``filters.rank.threshold``.
+# - Added 3D implementation of rank threshold_percentile filter - ``filters.rank.threshold_percentile``.
+- Added 3D implementation of rank noise_filter filter - ``filters.rank.noise_filter``.
+- Added 3D implementation of rank entropy filter - ``filters.rank.entropy``.
+- Added 3D implementation of rank otsu filter - ``filters.rank.otsu``.
+# - Added 3D implementation of rank percentile filter - ``filters.rank.percentile``.
+# - Added 3D implementation of rank windowed_histogram filter - ``filters.rank.windowed_histogram``.
```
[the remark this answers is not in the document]
xialuoke4062: The commented out filters were not implemented though

doc/release/release_dev.rst
```diff
@@ -11,10 +11,45 @@ For more information, examples, and documentation, please visit our website:
 
 https://scikit-image.org
 
+Starting from release 0.16, scikit-image follows the spirit of the recently
+introduced NumPy deprecation policy -- NEP 29
+(https://github.com/numpy/numpy/blob/master/doc/neps/nep-0029-deprecation_policy.rst).
+This release of scikit-image officially supports Python 3.6, 3.7, and
+3.8.
```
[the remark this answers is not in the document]
xialuoke4062: removed

skimage/filters/rank/core_cy.pyx
```diff
@@ -83,7 +212,7 @@ cdef void _core(void kernel(dtype_t_out*, Py_ssize_t, Py_ssize_t*, double,
         mask_data = &mask[0, 0]
 
     # define local variable types
-    cdef Py_ssize_t r, c, rr, cc, s, value, local_max, i, even_row
+    cdef Py_ssize_t r, c, rr, cc, s, j, value, local_max, i, even_row
```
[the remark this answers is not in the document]
xialuoke4062: removed

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
 
 plt.tight_layout()
 
+######################################################################
+# The example compares the local threshold with the global threshold in 3D
+
+from skimage import exposure
+
+brain = data.brain()
+
```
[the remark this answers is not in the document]
xialuoke4062: removed

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +403,49 @@
 
 plt.tight_layout()
 
+######################################################################
+# The example compares the local threshold with the global threshold in 3D
+
+from skimage import exposure
```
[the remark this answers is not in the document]
xialuoke4062: done

skimage/filters/rank/__init__.py
```diff
@@ -41,5 +42,23 @@
            'entropy',
            'otsu',
            'percentile',
-           'windowed_histogram',
-           'majority']
+           'windowed_histogram']
+
+__3Dfilters__ = ['autolevel',
```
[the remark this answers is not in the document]
xialuoke4062: I commented this out - I'm not sure how to keep the 3D ones separated for tests/test_rank.py: the 2D and 3D tests are different

xialuoke4062:
Documentation Testing error remains - any suggestion about what it might be?

doc/release/release_dev.rst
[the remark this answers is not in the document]
jni: Nevertheless, this is very unusual for our features announcement. I would summarise all these as "Added 3D support for many rank filters in ``skimage.filters.rank``.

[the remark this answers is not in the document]
xialuoke4062: Okay I will change it; thank you!

doc/examples/applications/plot_rank_filters.py
```diff
@@ -401,6 +404,45 @@
 
 plt.tight_layout()
 
+######################################################################
```
emmanuelle, line 407:
```suggestion
#########################################################################
```

emmanuelle:
@xialuoke4062 I wanted to take a look at the CI error but the build is too old and it's been removed. Could you please push an empty commit to trigger the CI again? (maybe you can merge master too)

xialuoke4062:
> @xialuoke4062 I wanted to take a look at the CI error but the build is too old and it's been removed. Could you please push an empty commit to trigger the CI again? (maybe you can merge master too)

@emmanuelle thanks for the reply - I have merged the master here!

jni:
@xialuoke4062 looking at the [failures](https://travis-ci.org/github/scikit-image/scikit-image/builds/715958724), there appears to be a segfault when MINIMUM_REQUIREMENTS=1. I suspect that is caused by the Cython version (0.29.13 in those builds, rather than 0.29.21 currently released). Could you bump up the required version to 0.29.21 in the `requirements/build.txt` file to see whether that fixes the error?

Sorry that this has taken so long but I believe we are very close, and I'd love to merge it for 0.18! =)

jni:
In fact, from my current reading of the code, this is ready! 🎉

xialuoke4062:
> @xialuoke4062 looking at the [failures](https://travis-ci.org/github/scikit-image/scikit-image/builds/715958724), there appears to be a segfault when MINIMUM_REQUIREMENTS=1. I suspect that is caused by the Cython version (0.29.13 in those builds, rather than 0.29.21 currently released). Could you bump up the required version to 0.29.21 in the `requirements/build.txt` file to see whether that fixes the error?
> 
> Sorry that this has taken so long but I believe we are very close, and I'd love to merge it for 0.18! =)

@jni Thank you so much for looking into this. At this point, I am curious to find out if we still want this buggy Otsu demo, since the only two demos we added are for Otsu and Equalizer (I did not add any for the rest 16 implemented 3D rank filters). And if we do want the demos, do you have any recommendations for the photos to be used? (Currently the brain.tiff we use is 256x256x10 which might be causing the problem?)

jni:
Well, there's no reason the demos should be buggy, and maybe we should figure out the bugs instead! 😅 It looks like bumping Cython didn't work, so the big question is what else could cause this. NumPy is the only real candidate I can think of, so you could update that dependency to 1.16. Unfortunately that is the newest we can do until January, according to [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html), so if it doesn't fix the problem, then we have to figure out where exactly the segfault is coming from — which isn't a terrible idea anyway, because usually it means that there is a hidden logic error, which new compilers simply mask for you.

I wonder if @stefanv or anyone else in @scikit-image/core knows how we can run our sphinx build in debug mode so that the segfault is caught rather than crashing Python? For pytest, there is `pytest-faulthandler`, which we should probably add to our test dependencies independently of this PR...

grlee77:
@ryanlu41, please check a small PR I made to your repo (NeuroDataDesign/scikit-image#7). This fixes the 3D gallery example for me and will hopefully also fix the segfault issue.

grlee77:
It is probably worth updating the test cases to include a volume that has a different size along each axis to help avoid bugs of this type.

benchmarks/benchmark_rank.py
```diff
@@ -2,17 +2,24 @@
 from skimage import img_as_ubyte
 from skimage.filters import rank
 from skimage.filters.rank import __all__ as all_rank_filters
-from skimage.morphology import grey, disk
+from skimage.filters.rank import __3Dfilters__ as all_3d_rank_filters
```
grlee77, line 5:
```suggestion
from skimage.filters.rank import __3Dfilters as all_3d_rank_filters
```

match the name in `skimage/filters/rank/__init__.py`

benchmarks/benchmark_rank.py
```diff
@@ -2,17 +2,24 @@
...
+from skimage.morphology import grey, disk, ball
 
 
 class RankSuite(object):
 
-    param_names = ["filter", "shape"]
-    params = [sorted(all_rank_filters), [(32, 32), (256, 256)]]
+    param_names = ["filter", "shape", "filter3d", "shape3d"]
+    params = [sorted(all_rank_filters), [(32, 32), (256, 256)],
+              sorted(all_3d_rank_filters), [(32, 32, 32), (256, 256, 256)]]
```
grlee77, line 13: I would just split the 2D and 3D cases to separate classes. Otherwise, the way this is now there all 2D cases get repeated for every 3D case and vice versa
.
In other words set of `params` results in `len(all_rank_filters) * 2 * len(all_3d_rank_filters) * 2` test cases instead of  the intended `len(all_rank_filters) * 2 + len(all_3d_rank_filters) * 2` cases

e.g. the output looks like...
```
             ------------------------------------------------------------- --------------
                          filter               shape          filter3d       (32, 32, 32) 
              ============================= ============ ================== ==============
                        autolevel             (32, 32)       autolevel         33.0±0ms   
                        autolevel             (32, 32)    enhance_contrast     21.2±0ms   
                        autolevel             (32, 32)        entropy          41.7±0ms   
                        autolevel             (32, 32)        equalize         32.3±0ms   
                        autolevel             (32, 32)     geometric_mean      37.4±0ms   
                        autolevel             (32, 32)        gradient         34.4±0ms   
                        autolevel             (32, 32)        majority         26.4±0ms   
                        autolevel             (32, 32)        maximum          32.7±0ms   
                        autolevel             (32, 32)          mean           35.2±0ms   
                        autolevel             (32, 32)         median          22.5±0ms   
                        autolevel             (32, 32)        minimum          32.1±0ms   
                        autolevel             (32, 32)         modal           36.6±0ms   
                        autolevel             (32, 32)      noise_filter       35.6±0ms   
                        autolevel             (32, 32)          otsu           42.2±0ms   
                        autolevel             (32, 32)          pop            30.6±0ms   
                        autolevel             (32, 32)     subtract_mean       36.8±0ms   
                        autolevel             (32, 32)          sum            23.5±0ms   
                        autolevel             (32, 32)       threshold         37.7±0ms   
                        autolevel            (256, 256)      autolevel         21.2±0ms   
                        autolevel            (256, 256)   enhance_contrast     32.4±0ms   
                        autolevel            (256, 256)       entropy          44.2±0ms   
                        autolevel            (256, 256)       equalize         30.6±0ms   
                        autolevel            (256, 256)    geometric_mean      40.5±0ms   
                        autolevel            (256, 256)       gradient         29.8±0ms   
                        autolevel            (256, 256)       majority         38.9±0ms   
                        autolevel            (256, 256)       maximum          20.8±0ms   
                        autolevel            (256, 256)         mean           35.1±0ms   
                        autolevel            (256, 256)        median          34.7±0ms   
                        autolevel            (256, 256)       minimum          29.1±0ms   
                        autolevel            (256, 256)        modal           39.5±0ms   
                        autolevel            (256, 256)     noise_filter       23.6±0ms   
                        autolevel            (256, 256)         otsu           42.5±0ms   
                        autolevel            (256, 256)         pop            31.1±0ms   
                        autolevel            (256, 256)    subtract_mean       35.3±0ms   
                        autolevel            (256, 256)         sum            36.8±0ms   
                        autolevel            (256, 256)      threshold         35.3±0ms   
```

grlee77: also (256, 256, 256) takes pretty long to run (~10 s for a single run for each filter), so perhaps reduce to (128, 128, 128) to make the overall run time more reasonable.

jni:
Damn, nice catch @grlee77! 🎉 🚀

grlee77:
> Damn, nice catch @grlee77! 

The clue was that in looking at the output in the plot of the brain in the demo, the image was non-zero only in a strip about the size of the "planes" dimensions so I figured the axes were probably swapped somewhere!

jni:
@ryanlu41 @xialuoke4062 last requested change, could you change the test cases to have different plane/row/column dimensions, as suggested by @grlee77? After that imho this will be ready to go in!

doc/examples/color_exposure/plot_local_equalize.py
```diff
@@ -168,3 +168,4 @@ def plot_img_and_hist(image, axes, bins=256):
 
 # prevent overlap of y-axis labels
 fig.tight_layout()
+plt.show()
```
xialuoke4062, line 171: @grlee77 I just recalled that @rfezzani suggested removing this line of code in the previous code review, so I removed it from both 2D and 3D demos in this file - I am curious about the difference of having it versus not? 

On a side note, this equalization rank filter demo exists in both color_exposure/plot_local_equalize.py and applications/plot_rank_filters.py; is this (and potentially other rank filter demos) repetition created on purpose? @jni  (we only added the equalization demo in this file)

xialuoke4062:
> > Damn, nice catch @grlee77!
> 
> The clue was that in looking at the output in the plot of the brain in the demo, the image was non-zero only in a strip about the size of the "planes" dimensions so I figured the axes were probably swapped somewhere!

Thank you so much for spotting the error! The image did look weird but I couldn't figure out why.

doc/examples/color_exposure/plot_local_equalize.py
grlee77: I think the request was just to remove the extra `plt.show()` from the middle of the file (between the 2D and 3D demo). I think there should still be a final `plt.show()` at the end of the file. This makes it so that the figures don't just pop up and then immediately disappear if the example is run from the command line.

xialuoke4062: Thank you!

emmanuelle:
I think this PR is ready now? @grlee77 @jni

grlee77 (approved):
Thanks for adding the anisotropic test cases. Let's get this merged

PR #4437, opened 2020-02-07
Fix use_quantiles behavior in canny

### LINKED ISSUE #4282: use_quantiles description in canny is incorrect

rfezzani:
## Description

Fixes #4282.

A test asserting the bug fix is added.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/tests/test_canny.py
+++ b/skimage/feature/tests/test_canny.py
@@ -105,6 +105,11 @@ class TestCanny(unittest.TestCase):
         self.assertRaises(ValueError, F.canny, image, use_quantiles=True,
                           low_threshold=0.5, high_threshold=-100)
 
+        # Example from issue #4282
+        image = data.camera()
+        self.assertRaises(ValueError, F.canny, image, use_quantiles=True,
+                          low_threshold=50, high_threshold=150)
+
     def test_dtype(self):
         """Check that the same output is produced regardless of image dtype."""
         image_uint8 = data.camera()
--- a/skimage/feature/_canny.py
+++ b/skimage/feature/_canny.py
@@ -169,7 +169,7 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
         high_threshold = 0.2
     elif use_quantiles:
         if not(0.0 <= high_threshold <= 1.0):
-            raise ValueError("Quantile thresholds must not be > 1.0")
+            raise ValueError("Quantile thresholds must be between 0 and 1.")
     else:
         high_threshold = high_threshold / dtype_max
 
--- a/skimage/feature/_canny.py
+++ b/skimage/feature/_canny.py
@@ -161,7 +161,7 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
         low_threshold = 0.1
     elif use_quantiles:
         if not(0.0 <= low_threshold <= 1.0):
-            raise ValueError("Quantile thresholds must not be > 1.0")
+            raise ValueError("Quantile thresholds must be between 0 and 1.")
     else:
         low_threshold = low_threshold / dtype_max
 
--- a/skimage/feature/_canny.py
+++ b/skimage/feature/_canny.py
@@ -159,11 +159,17 @@ def canny(image, sigma=1., low_threshold=None, high_threshold=None, mask=None,
 
     if low_threshold is None:
         low_threshold = 0.1
+    elif use_quantiles:
+        if not(0.0 <= low_threshold <= 1.0):
+            raise ValueError("Quantile thresholds must not be > 1.0")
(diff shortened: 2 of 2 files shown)


rfezzani:
To ease the integration of this bug fix, I decoupled it from `canny` refactoring in #4342.

skimage/feature/tests/test_canny.py
```diff
@@ -105,6 +105,11 @@ def test_invalid_use_quantiles(self):
         self.assertRaises(ValueError, F.canny, image, use_quantiles=True,
                           low_threshold=0.5, high_threshold=-100)
 
+        # Example from issue #4282
+        image = data.camera()
+        self.assertRaises(ValueError, F.canny, image, use_quantiles=True,
```
sciunto: Can you explain why this is not catched by lines 266-271 in _canny.py? Can you justify why these lines (266-271) are still needed? Although I can reproduce that this test fails with master, just by reading the code, I do not understand it well

sciunto: What I understand is that the lines 163 and 168 must be done after the checks and that the checks came too late.

sciunto: Is this correct? Then, you can clean up the function.

rfezzani: Yes that's it ;-)

rfezzani: I don't get what you mean by "clean up the function."

sciunto: Great. As soon as you remove the unnecessary checks, I approve. ;)

rfezzani: I think that there is a misunderstanding here, which unnecessary checks?

sciunto: I would remove this ```
        if high_threshold > 1.0 or low_threshold > 1.0:
            raise ValueError("Quantile thresholds must not be > 1.0")
            raise ValueError("Quantile thresholds must not be > 1.0")
        if high_threshold < 0.0 or low_threshold < 0.0:
        if high_threshold < 0.0 or low_threshold < 0.0:
            raise ValueError("Quantile thresholds must not be < 0.0")
            raise ValueError("Quantile thresholds must not be < 0.0")
```
and handle everything at the top of the function.

sciunto: To me, it's not easy to read to have checks fairly late in a function. The sooner we detect something wrong, the better it is. And as far as I understand, there is no need to delay. High and low thresholds are not modified after the line 168.

rfezzani: The problem is that previously the  check for quantiles came after the division of the input low and high thresholds by `dtype_max`. That's why the error was not raised, and that's what is reported in #4437 .

rfezzani: >  The sooner we detect something wrong, the better it is
I agree, but I don't see how it can be checked sooner since low and high thresholds can be `None`...

sciunto: Am I wrong if I do this? (I didn't check) https://gist.github.com/sciunto/079fd7d466ca50bd8f6e133f882b035d

rfezzani: When `use_quantile` is True, the low and high threshold are not supposed to be normalized, they are simply required to be between 0 and 1.
When it is False, the thresholds are divided by `dtype_max` because the input image is internally cast to float using `img_as_float`, and it is not required to be between 0 and 1, so the second check that you introduced is wrong...

rfezzani:
Thank you @sciunto.

rfezzani:
Thank you @jni!

PR #4438, opened 2020-02-09
Deprecate gray scale image support for rgb2gray

### LINKED ISSUE #3200: Should rgb2gray accept gray images?

rfezzani:
## Description

Fixes #3200 

Following @hmaarrfk [comment](https://github.com/scikit-image/scikit-image/pull/4418/files#r376811009), this PR is a prerequisite to #4418.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -793,7 +793,7 @@ def rgb2gray(rgb):
         warn('The behavior of rgb2gray will change in scikit-image 0.19. '
              'Currently, rgb2gray allows 2D grayscale image to be passed '
              'as inputs and leaves them unmodified as outputs. '
-             'In version 0.19, this will no longer be these images will '
+             'Starting from version 0.19, 2D arrays will '
              'be treated as 1D images with 3 channels.',
              FutureWarning, stacklevel=2)
         return np.ascontiguousarray(rgb)
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -790,7 +790,11 @@ def rgb2gray(rgb):
     """
 
     if rgb.ndim == 2:
-        warn('Gray scale image conversion is now deprecated.',
+        warn('The behavior of rgb2gray will change in scikit-image 0.19. '
+             'Currently, rgb2gray allows 2D grayscale image to be passed '
+             'as inputs and leaves them unmodified as outputs. '
+             'In version 0.19, this will no longer be these images will '
+             'be treated as 1D images with 3 channels.',
              FutureWarning, stacklevel=2)
         return np.ascontiguousarray(rgb)
 
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -790,6 +790,8 @@ def rgb2gray(rgb):
     """
 
     if rgb.ndim == 2:
+        warn('Gray scale image conversion is now deprecated.',
+             FutureWarning, stacklevel=2)
         return np.ascontiguousarray(rgb)
 
     rgb = _prepare_colorarray(rgb[..., :3])
--- a/TODO.txt
+++ b/TODO.txt
@@ -51,7 +51,8 @@ Version 0.19
 * Remove the warnings in ``skimage/filters/ridges.py`` from sato and hessian
   functions.
 * In ``skimage/color/colorconv.py``, remove  `rgb2grey` and `grey2rgb`.
-
+* In ``skimage/color/colorconv.py``, remove gray scale image support
+  deprecation from rgb2gray.
 
 Other
 -----
--- a/skimage/color/tests/test_colorconv.py
+++ b/skimage/color/tests/test_colorconv.py
@@ -298,7 +298,8 @@ class TestColorconv(TestCase):
         assert rgb2gray(x).ndim == 2
 
     def test_rgb2gray_on_gray(self):
-        rgb2gray(np.random.rand(5, 5))
+        with expected_warnings(['Gray scale image conversion']):
+            rgb2gray(np.random.rand(5, 5))
 
     def test_rgb2gray_dtype(self):
         img = np.random.rand(10, 10, 3).astype('float64')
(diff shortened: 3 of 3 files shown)


skimage/color/colorconv.py
```diff
@@ -790,6 +790,8 @@ def rgb2gray(rgb):
     """
 
     if rgb.ndim == 2:
+        warn('Gray scale image conversion is now deprecated.',
```
hmaarrfk: Honest question, do we want to raise an error in the future? 

I guess for a line scanner a 1D image with RGB channels make sense.

Maybe we want to specify how we will be interpreting this in the future?

hmaarrfk: I think this warning messages needs to be fleshed out more.

```
The behavior of rgb2gray will change in scikit-image 0.19. Currently, rgb2gray allows 2D grayscale image to be passed as inputs and leaves them unmodified as outputs. In version 0.19, this will no longer be these images will be treated as 1D images with 3 channels.
```

rfezzani: I supposed that the new behavior would be described when implemented, in #4418 :-)

rfezzani: I think that an error should be raised if the last dimension is not 3, what ever is the number of dimensions

skimage/color/colorconv.py
```diff
@@ -790,6 +790,12 @@ def rgb2gray(rgb):
     """
 
     if rgb.ndim == 2:
+        warn('The behavior of rgb2gray will change in scikit-image 0.19. '
```
hmaarrfk: I think we need a multichannel argument to allow the user to use the new behavior today

skimage/color/colorconv.py
```diff
@@ -790,6 +790,12 @@ def rgb2gray(rgb):
...
+             'Currently, rgb2gray allows 2D grayscale image to be passed '
+             'as inputs and leaves them unmodified as outputs. '
+             'In version 0.19, this will no longer be these images will '
```
soupault:
```suggestion
             'Starting from version 0.19, 2D arrays will '
```

skimage/color/colorconv.py
jni: @hmaarrfk I thought about that, but then that suggests either that we will remove the multichannel argument in the future (another breaking change :man_facepalming:) or that it will do nothing in the future (not great either), or that we will continue to support multichannel forever (which as far as I know is false at this point). :thinking:

hmaarrfk: This was always the blocker for this deprecation. 

I don't think this new behavior is too critical, so we can avoid the introduction of the new parameter.

rfezzani: Yes, the warning message is explicit enough to not make the user confused: old users are not disturbed since the old behavior is kept, and users needing the new behavior need to wait version 0.19.
To avoid this warning, they only have to check to not provide 2D arrays as input...
Same argument is valid for #4440

jni: @rfezzani I got confused between the two PRs. In this one, it's fine, as the 2D use case isn't that compelling. In the other one, people who want to convert 3D images to RGB where the last dimension has shape 3 have no way to use the library in a future-proof way.

rfezzani: >  In the other one, people who want to convert 3D images to RGB where the last dimension has shape 3 have no way to use the library in a future-proof way.

We may advise to add a dummy dimension and to remove it after conversion
`rgb2gray(rgb[np.newaxis, ...])[0, ...]`

rfezzani: Again confused :smile:! `rgb2gray(rgb[np.newaxis, ...])[0, ...]` is valid in this PR, and `gray2rgb(image[np.newaxis, ...])[0, ...]` also if we remove this [line](https://github.com/scikit-image/scikit-image/pull/4440/files#diff-9f4dedd98e7cca606dec9f0b792c1d94R835).

rfezzani:
Any other required change @hmaarrfk ?

skimage/color/colorconv.py
jni: Ah! Advising reshapes before and after is a great idea! :tada: Could you add the correct one in each case? =)

`gray2rgb(image.ravel()).reshape(image.shape + (3,))` will work for nD gray2rgb.

rfezzani: What about adding these advises when the 3D support is effectively allowed (in #4418)? :wink:

rfezzani:
Thank you @hmaarrfk ;-)

hmaarrfk:
Yeah. I have a PR open for this nD conversion as well, but didn't have a partner that was as motivated as me at the time. Hopefully somebody else can put the nail on this one.

What scares me is that by providing conveniences for advances users, we are forgetting conveniences for novice users.

This "ensure grayscale" or "ensure RGB" was quite convenient.

rfezzani:
I personally believe that the old behavior was confusing: how **rgb**2gray can accept grayscale as input :smile:

rfezzani:
Anyone can review this PR? It is prerequisite for #4418... Thx

rfezzani:
This PR is the only remaining prerequisite for #4418, a second review please ;-)

emmanuelle:
ok, let's merge this one. The `ensure gray` or `ensure rgb` was nice indeed but it just adds a test on the number of dimensions (we should assume that users know the geometry of their data, or they can always test the value of the last dimension).

rfezzani:
Thank you @emmanuelle :tada:

PR #4440, opened 2020-02-10
Deprecate non gray scale image conversion in gray2rgb

rfezzani:
## Description

Following @hmaarrfk recommandations, this PR is a prerequisite to #4418.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -842,9 +842,12 @@ def gray2rgb(image, alpha=None):
             is_rgb = True
 
     if is_rgb:
-        warn('Non gray scale images conversion is now deprecated. In version '
-             '0.19, input image  will be considered as grey scale, whatever '
-             'is its shape.', FutureWarning, stacklevel=2)
+        warn('Pass-through of possibly RGB images in gray2rgb is deprecated. '
+             'In version 0.19, input arrays will always be considered '
+             'grayscale, even if the last dimension has length 3 or 4. '
+             'To prevent this warning and ensure compatibility with future '
+             'versions, detect RGB images outside of this function.',
+             FutureWarning, stacklevel=2)
         if alpha is False:
             image = image[..., :3]
 
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -842,6 +842,9 @@ def gray2rgb(image, alpha=None):
             is_rgb = True
 
     if is_rgb:
+        warn('Non gray scale images conversion is now deprecated. In version '
+             '0.19, input image  will be considered as grey scale, whatever '
+             'is its shape.', FutureWarning, stacklevel=2)
         if alpha is False:
             image = image[..., :3]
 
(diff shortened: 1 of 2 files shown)


skimage/color/colorconv.py
```diff
@@ -842,6 +842,9 @@ def gray2rgb(image, alpha=None):
             is_rgb = True
 
     if is_rgb:
+        warn('Non gray scale images conversion is now deprecated. In version '
+             '0.19, input image  will be considered as grey scale, whatever '
+             'is its shape.', FutureWarning, stacklevel=2)
```
jni: gray scale/grey scale -> grayscale, and remove "whatever is its shape". ie:

"Pass-through of possibly RGB images in `gray2rgb` is deprecated. In version 0.19, input arrays will always be considered grayscale, even if the last dimension has length 3 or 4. To prevent this warning and ensure compatibility with future versions, detect RGB images outside of this function."

We also need a way to silence this warning when this function is used correctly with a grayscale array with last axis shape (3,)...

rfezzani: Do you mean that we need to add an argument (`nd` for example) to be set to True if the user wants effectively convert nD grayscale image to RGB?

skimage/color/colorconv.py
```diff
@@ -842,6 +842,12 @@ def gray2rgb(image, alpha=None):
             is_rgb = True
 
     if is_rgb:
+        warn('Pass-through of possibly RGB images in gray2rgb is deprecated. '
```
hmaarrfk: I think the appropriate parameter to add is one called multichannel

rfezzani: `multichannel` is in fact well suited for this, but we all agree that in version 0.19 this parameter will be deprecated since nD will be the default behavior of gray2rgb?

hmaarrfk: Yeah..... Maybe the best way is to ask users to import the function from a "future" module???

Definitely annoying

hmaarrfk: Not sure the best way to go about this one

jni: So, one thing I've been thinking about is to add a note to the next two release notes, 0.17 and 0.18, that users should pin to <1.0, because 1.0 will contain breaking changes, and then work to get 1.0 out after that, with a bunch of breaking changes. This would solve this situation and a bunch of others... But probably I should start a SKIP for this. =)

hmaarrfk: I'm ok with this as is.

Users that avoid this warning will have code that is correct all the way through

skimage/color/colorconv.py
jni: Yes, exactly. But `nd` isn't a great argument name, and anyway, in the future we would want to remove it. So I don't really know what to do here. =\

emmanuelle:
Thanks @rfezzani, merging!

rfezzani:
Thank you @emmanuelle

PR #4444, opened 2020-02-14
WIP: Move RAGs from future to segmentation

jni:
## Description

This is a rare opportunity to really clean up APIs and so on. I still have some work to do, but would like to put this out here for people to start having a look. One question I have is whether we want rag functions to live in their own `graph` submodule, or to be present in the top level `segmentation` namespace directly.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/segmentation/graph/rag.py
```diff
@@ -455,13 +455,11 @@ def rag_boundary(labels, edge_map, connectivity=2):
     data_coo = sparse.coo_matrix((data, (labels_small, labels_large)))
     graph_matrix = data_coo.tocsr()
     graph_matrix.data /= count_matrix.data
-
     rag = RAG()
     rag.add_weighted_edges_from(_edge_generator_from_csr(graph_matrix),
                                 weight='weight')
-    rag.add_weighted_edges_from(_edge_generator_from_csr(count_matrix),
```
emmanuelle, line 440: here, calling `add_edges_from` was overwriting the `weight` key of edges. I'm not sure the code should be modified here or in the RAG  `add_edges_from` but this fixes one of the two test failures. @jni

skimage/segmentation/graph/graph_merge.py
```diff
@@ -118,20 +118,17 @@ def merge_hierarchical(labels, rag, thresh, rag_copy, in_place_merge,
...
-                _rename_node(rag, n2, next_id)
-                src, dst = n1, next_id
-            else:
-                src, dst = n1, n2
+            src, dst = n1, n2
 
             merge_func(rag, src, dst)
-            new_id = rag.merge_nodes(src, dst, weight_func)
+            new_id = rag.merge_nodes(
+                src, dst, weight_func, in_place=in_place_merge
```
emmanuelle, line 125: I'm not sure this function works as intended for `in_place=False` (I had to change one test to use `in_place=True` so that it worked).

skimage/segmentation/graph/tests/test_rag.py
```diff
@@ -121,7 +140,7 @@ def _pre_merge_mean_color(graph, src, dst):
 
 
 def merge_hierarchical_mean_color(labels, rag, thresh, rag_copy=True,
-                                  in_place_merge=False):
+                                  in_place_merge=True):
```
emmanuelle, line 143: this is one change I had to make so that `test_rag_hierarchical` passes.

skimage/segmentation/graph/graph_merge.py
emmanuelle: in the failing example the new node was empty

emmanuelle:
ok there are still some problems for building and testing the doc.

jni:
@emmanuelle thank you so much for helping me get unstuck here! I think all of the related docs are fixed now! :tada: 

Now we need to think hard about the API before merging. My intuition is:
- we should not have `rag_mean_boundary` and `rag_mean_color` functions, but rather use the RAG object with different constructors arguments. Preferably, some kind of functions that specify the cached properties and how they get merged.
- we should "build in" regionprops to the RAGs. Maybe this can come later but we should consider how to do it when thinking about the API.
- do we call it RAG or RegionAdjacencyGraph?
- any other thoughts?

After 0.18, there are many speedups I implemented in my Gala library, see e.g. https://github.com/janelia-flyem/gala/pull/82, after these classes were added, that could be ported here. I noticed building the graph for `data.eagle()` is super slow.

doc/examples/segmentation/plot_rag_boundary.py
```diff
@@ -24,8 +24,10 @@
...
 g = graph.rag_boundary(labels, edges)
+
+fig, ax = plt.subplots()
 lc = graph.show_rag(labels, g, edges_rgb, img_cmap=None, edge_cmap='viridis',
-                    edge_width=1.2)
+                    edge_width=1.2, ax=ax)
 
-plt.colorbar(lc, fraction=0.03)
+fig.colorbar(lc, fraction=0.03)
 io.show()
```
rfezzani, line 33:
```suggestion
plt.show()
```
Isn't it confusing here?

grlee77:
@jni: If this is going to take a while to sort out, do you want to remove from the 0.19 milestone?

PR #4445, opened 2020-02-15
Difference of Gaussians function

mellertd:
## Description

Function for applying a band pass filter using the Difference of Gaussians method. Discussed in #4231. I am trying to keep this pretty simple ... just catching any weird input with the sigmas and passing along to `filters.gaussian`. However, unlike `filters.gaussian`, I am currently not trying to guess number of dimensions, multichannel vs not, etc. User must specify whether image is multichannel or not (default is False).

Starting this PR as a WIP for early feedback.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


mellertd:
Changed the title from WIP, as this is pretty much ready for review. I don't have a gallery example, but I am not sure whether one is needed.

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,117 @@ def _guess_spatial_dimensions(image):
         return 3
     else:
         raise ValueError("Expected 2D, 3D, or 4D array, got %iD." % image.ndim)
+
+
+def difference_of_gaussians(image, sigma1, sigma2=None, *,
```
jni: I think here I would prefer `low_sigma` and `high_sigma`, or even `min_sigma` and `max_sigma`, even if they are conceptually a bit different from `blob_dog`.

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,117 @@ def _guess_spatial_dimensions(image):
...
+        Whether the last axis of the image is to be interpreted as multiple
+        channels. If True, each channel is filtered separately (channels are
+        not mixed together).
+    truncate : float, optional (default is 4.0)
+        Truncate the filter at this many standard deviations.
+
+    Returns
+    -------
+    filtered_image : ndarray
+        the filtered array
```
jni:
```suggestion
        The filtered array.
```

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,117 @@ def _guess_spatial_dimensions(image):
...
+    -----
+    This function will subtract an array filtered with a guassian kernel
+    with sigmas given by ``sigma2`` from an array filtered with a gaussian
+    kernel with sigmas provided by ``sigma1``. The values for ``sigma2`` must
+    always be greater than or equal to the corresponding values in ``sigma1``,
+    or a ``ValueError`` will be raised.
+
+    When ``sigma2`` is none, the values for ``sigma2`` will be calculated
+    as 1.6x the corresponding values in ``sigma1``. This approximates the
+    inverted Laplacian of Guassian, commonly used in edge and blob detection.
```
jni:
```suggestion
    inverted Laplacian of Gaussian, commonly used in edge and blob detection.
```

Maybe add a comment about where the value 1.6 comes from?

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,117 @@ def _guess_spatial_dimensions(image):
         return 3
     else:
         raise ValueError("Expected 2D, 3D, or 4D array, got %iD." % image.ndim)
+
+
+def difference_of_gaussians(image, sigma1, sigma2=None, *,
+                            mode='nearest', cval=0, multichannel=False,
+                            truncate=4.0):
+    """Multi-dimensional band-pass filter using the Difference of Gaussians
+    method.
```
jni: This should fit in one line, with additional comments separated by a blank line. Also, "multidimensional band-pass filter" is very jargony. I would suggest something like:

```
Find features between `low_sigma` and `high_sigma` in size.
```

You can expand on band-pass filters all you like in a subsequent paragraph. =)

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,117 @@ def _guess_spatial_dimensions(image):
...
+
+    Except for sigma values, all parameters are used for both filters.
+
+    Examples
+    --------
+    Apply a simple Difference of Gaussians filter to a color image:
+
+    >>> from skimage.data import astronaut
+    >>> from skimage.filters import difference_of_gaussians
+    >>> image = astronaut()
```
jni: I feel like this line can be skipped — just use it inline in the call to `dog`.

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,117 @@ def _guess_spatial_dimensions(image):
...
+    Apply a Difference of Gaussians filter to a grayscale image using different
+    sigma values for each axis:
+
+    >>> from skimage.data import camera
+    >>> image = camera()
+    >>> filtered_image = difference_of_gaussians(image, (2,5), (3,20))
+
+    """
+    image = img_as_float(image)
+    sigma1 = np.array(sigma1, dtype='float', ndmin=1)
```
jni: fascinating! Never knew about ndmin. =D

skimage/filters/_gaussian.py
mellertd: I am worried about "low" and "high" being confusing, e.g., are we talking about the high frequency cutoff (adjusted with the smaller sigma value) or the sigma value which is larger in magnitude?  "min" and "max" would cause less confusion, but it's weird to me to use the terms min and max for two values (although technically correct I suppose). What about `small_sigma` and `large_sigma`?

skimage/filters/_gaussian.py
mellertd: yes I taught myself that for this PR, when I learned that giving `np.array` a scalar without forcing a dimension on it returns, unsurprisingly, a scalar. Since a numpy scalar has no length it ruined the numpy-based approach I had for handling my sigma inputs. There was another way I had of doing this using list comprehensions and checking for iterables, but that ended up looking way more complicated (to me) than what I went with.

skimage/filters/_gaussian.py
mellertd: The honest answer is the `blob_dog` documentation and Wikipedia, although it appears Marr and Hildreth (1980) is the original source.

Although after further reading and poking at the math a bit, it seems that *all* DoGs approximate the LoG. The 1.6 is apparently picked based on a tradeoff between sensitivity and bandwidth.

In any case, I'll fix the docs to be more precise -- should I reference Marr and Hildreth? If I do so, I should probably get my hands on a copy of the paper and give it a read.

skimage/filters/_gaussian.py
mellertd: I ended up going with `lesser_sigma` and `greater_sigma`. If you don't like that I can go with min/max.

skimage/filters/_gaussian.py
mellertd: I went ahead and referenced Marr and Hildreth and provided a brief explanation

mellertd:
addressed @jni 's comments and will add a gallery example soon

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,132 @@ def _guess_spatial_dimensions(image):
         return 3
     else:
         raise ValueError("Expected 2D, 3D, or 4D array, got %iD." % image.ndim)
+
+
+def difference_of_gaussians(image, lesser_sigma, greater_sigma=None, *,
```
jni: I still think `high` and `low` are better choices here. One debate that we are [having now](https://github.com/scikit-image/meeting-notes/blob/master/2020/2020-02-17.markdown) is how best to make all APIs across scikit-image consistent, including parameters names. I think high and low or max and min are much more common in the API currently. Anyway, `sigma` is a common notation in image domain, rather than frequency domain, where people typically use theta. So, I think the risk of confusing high and low with the frequency band is very low.

mellertd:
I went ahead and changed it to low and high anyway. No sense in making a big deal out of it -- I'll probably be the only one that's ever confused anyway :)

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,132 @@ def _guess_spatial_dimensions(image):
         return 3
     else:
         raise ValueError("Expected 2D, 3D, or 4D array, got %iD." % image.ndim)
+
+
+def difference_of_gaussians(image, low_sigma, high_sigma=None, *,
+                            mode='nearest', cval=0, multichannel=False,
```
alexdesiqueira: Isn't it better to have `multichannel` as the last argument here?

jni: they are keyword-only arguments, so this doesn't matter at all, we can reorder them any time we like. That's the point! =D

alexdesiqueira: But... but we like order! ... don't we? :)

mellertd: I actually just followed the ordering in `gaussian`, except adding the `*` as @jni noted.

alexdesiqueira: Don't worry @mellertd, the order is different in my head :)

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,132 @@ def _guess_spatial_dimensions(image):
...
+        Truncate the filter at this many standard deviations.
+
+    Returns
+    -------
+    filtered_image : ndarray
+        the filtered array.
+
+    Notes
+    -----
+    This function will subtract an array filtered with a guassian kernel
```
stefanv:
```suggestion
    This function will subtract an array filtered with a Gaussian kernel
```

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,132 @@ def _guess_spatial_dimensions(image):
...
+    with sigmas given by ``high_sigma`` from an array filtered with a
+    gaussian kernel with sigmas provided by ``low_sigma``. The values for
```
stefanv:
```suggestion
    Gaussian kernel with sigmas provided by ``low_sigma``. The values for
```

mellertd:
Hello @stefanv thanks for the feedback. I didn't want to change the implementation of blob_dog because it takes a fairly different approach for how the sigmas and differences are calculated.

I committed your other comments.

mellertd:
Hello everyone! Just checking whether any more work is needed on this. I am traveling and have some plane time :)

emmanuelle:
Hi @mellertd thanks, taking a look right now :-). I ran the gallery example and I wonder about the origin of the spectral leakage around the horizontal and vertical directions. Maybe a windowing function could help? You contributed nD windowing functions if I remember well ;-).

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,136 @@ def _guess_spatial_dimensions(image):
...
+                            mode='nearest', cval=0, multichannel=False,
+                            truncate=4.0):
+    """Find features between ``low_sigma`` and ``high_sigma`` in size.
+
+    This function uses the Difference of Gaussians method for applying
+    band-pass filters to multi-dimensional arrays. The input array is
+    blurred with two Gaussian kernels of differing sigmas to produce two
+    intermediate, filtered images. The more-blurred image is then subtracted
+    from the less-blurred image. The final output image will therefore have
+    had high frequency components attenuated by the smaller-sigma Gaussian, and
```
emmanuelle:
```suggestion
    had high-frequency components attenuated by the smaller-sigma Gaussian, and
```

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,136 @@ def _guess_spatial_dimensions(image):
...
+        Standard deviation(s) for the Gaussian kernel with the smaller sigmas
+        across all axes. The standard deviations are given for each axis as a
+        sequence, or as a single number, in which case the single number is
+        used as the standard deviation value for all axes.
+    high_sigma : scalar or sequence of scalars, optional (default is None)
+        Standard deviation(s) for the Gaussian kernel with the larger sigmas
+        across all axes. The standard deviations are given for each axis as a
+        sequence, or as a single number, in which case the single number is
+        used as the standard deviation value for all axes. If None is given
+        (default), sigmas for all axes are calculated as 1.6 * low_sigma.
```
emmanuelle: why 1.6? Is it arbitrary, does it come from the literature?

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,136 @@ def _guess_spatial_dimensions(image):
...
+    filtered_image : ndarray
+        the filtered array.
+
+    See also
+    --------
+    skimage.feature.blog_dog
+
+    Notes
+    -----
+    This function will subtract an array filtered with a Guassian kernel
```
emmanuelle:
```suggestion
    This function will subtract an array filtered with a Gaussian kernel
```

skimage/filters/_gaussian.py
emmanuelle: ah ok it's explained below :-).

skimage/filters/_gaussian.py
```diff
@@ -155,3 +155,136 @@ def _guess_spatial_dimensions(image):
...
+    with sigmas given by ``high_sigma`` from an array filtered with a
+    Gaussian kernel with sigmas provided by ``low_sigma``. The values for
+    ``high_sigma`` must always be greater than or equal to the corresponding
+    values in ``low_sigma``, or a ``ValueError`` will be raised.
+
+    When ``high_sigma`` is none, the values for ``high_sigma`` will be
+    calculated as 1.6x the corresponding values in ``low_sigma``. This ratio
+    was originally proposed by Marr and Hildreth (1980) [1]_ and is commonly
+    used when approximating the inverted Laplacian of Guassian, which is used
```
emmanuelle:
```suggestion
    used when approximating the inverted Laplacian of Gaussian, which is used
```

emmanuelle:
@mellertd just a question about the example, and suggestions about typo, then I'll approve the PR :-).

skimage/filters/_gaussian.py
mellertd: It seems my fingers really wanted to misspell that word

mellertd:
indeed, thanks for the suggestion, @emmanuelle !

jni:
:tada:

PR #4457, opened 2020-02-17
Address FutureWarning from numpy in subdtype check in reginoprops

hmaarrfk:
## Description

Avoids
```python
In [1]: import numpy as np                                                      

In [2]: np.issubdtype(np.float, np.bool)                                        
/home/mark/miniconda3/envs/mcam_dev/bin/ipython:1: FutureWarning: Conversion of the second argument of issubdtype from `bool` to `np.generic` is deprecated. In future, it will be treated as `np.bool_ == np.dtype(bool).type`
```
## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_regionprops.py
+++ b/skimage/measure/_regionprops.py
@@ -853,7 +853,7 @@ def regionprops(label_image, intensity_image=None, cache=True,
         raise TypeError('Only 2-D and 3-D images supported.')
 
     if not np.issubdtype(label_image.dtype, np.integer):
-        if np.issubdtype(label_image.dtype, np.bool):
+        if np.issubdtype(label_image.dtype, np.bool_):
             raise TypeError(
                     'Non-integer image types are ambiguous: '
                     'use skimage.measure.label to label the connected'


skimage/measure/_regionprops.py
```diff
@@ -853,7 +853,7 @@ def regionprops(label_image, intensity_image=None, cache=True,
         raise TypeError('Only 2-D and 3-D images supported.')
 
     if not np.issubdtype(label_image.dtype, np.integer):
-        if np.issubdtype(label_image.dtype, np.bool):
+        if np.issubdtype(label_image.dtype, np.bool_):
```
grlee77:
```suggestion
        if label_image.dtype.kind == 'b':
```

grlee77: In CuPy we typically check the kind attribute for this sort of thing because it is very fast, but I understand if others would find it less readable. I doubt this line is critical to performance, so whichever is more maintainable is probably best.

hmaarrfk: Good tip. 

I think it is too much microoptimization for scikit image though

PR #4464, opened 2020-02-18
Deprecators

hmaarrfk:
## Description

@jni, @alexdesiqueira this is the demo that I was talking about regarding how the decorator can be used to basically make the change "once".

In version 1.0, you are left with "zombie" decorators that would basically do nothing.

https://github.com/hmaarrfk/wabisabi/blob/master/wabisabi/kwonly_change.py#L89
I forced push to my old PR, now i can't reopen it.
https://github.com/scikit-image/scikit-image/pull/3346

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/_deprecators.py
+++ b/skimage/_shared/_deprecators.py
@@ -1,0 +1,12 @@
+from functools import partial
+from ..version import __version__
+from deprecation_factory import kwonly_change
+from deprecation_factory import default_parameter_change
+
+kwonly_change = partial(kwonly_change,
+                        library_name='scikit-image',
+                        current_library_version=__version__)
+
+default_parameter_change = partial(default_parameter_change,
+                                   library_name='scikit-image',
+                                   current_library_version=__version__)
--- a/skimage/_shared/_deprecators.py
+++ b/skimage/_shared/_deprecators.py
@@ -1,7 +1,6 @@
 from functools import partial
 from ..version import __version__
-from deprecation_factory import kwonly_change
-from deprecation_factory import default_parameter_change
+from wabisabi import kwonly_change, default_parameter_change
 
 kwonly_change = partial(kwonly_change,
                         library_name='scikit-image',
--- a/skimage/transform/hough_transform.py
+++ b/skimage/transform/hough_transform.py
@@ -4,9 +4,10 @@ from ._hough_transform import (_hough_circle,
                                _hough_ellipse,
                                _hough_line,
                                _probabilistic_hough_line as _prob_hough_line)
+from .._shared._deprecators import kwonly_change
 
-
-def hough_line_peaks(hspace, angles, dists, min_distance=9, min_angle=10,
+@kwonly_change('1.0')
+def hough_line_peaks(hspace, angles, dists, *, min_distance=9, min_angle=10,
                      threshold=None, num_peaks=np.inf):
     """Return peaks in a straight line Hough transform.
 
@@ -68,7 +69,8 @@ def hough_line_peaks(hspace, angles, dists, min_distance=9, min_angle=10,
         return (h, np.array([]), np.array([]))
 
 
-def hough_circle(image, radius, normalize=True, full_output=False):
+@kwonly_change('1.0')
+def hough_circle(image, radius, *, normalize=True, full_output=False):
     """Perform a circular Hough transform.
 
     Parameters
@@ -112,7 +114,8 @@ def hough_circle(image, radius, normalize=True, full_output=False):
                          normalize=normalize, full_output=full_output)
 
 
-def hough_ellipse(image, threshold=4, accuracy=1, min_size=4, max_size=None):
+@kwonly_change('1.0')
+def hough_ellipse(image, *, threshold=4, accuracy=1, min_size=4, max_size=None):
     """Perform an elliptical Hough transform.
 
     Parameters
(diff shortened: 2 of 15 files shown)


soupault:
This is a pretty cool thing for long-term maintenance 👍.

rfezzani:
Very cool stuff. What about #4158 then? Close, postpone....

alexdesiqueira:
A small complaining on the tests, @hmaarrfk:
```
/home/travis/venv/lib/python3.7/site-packages/skimage/transform/radon_transform.py:docstring of skimage.transform.iradon:57:Inline literal start-string without end-string.
```

hmaarrfk:
So if people are in favor of this, I suggest we either:

A. Version wabisabi (well what we need from it)
B. Depend on it

and start doing this kind of deprecation 1 file at a time, taking care of internal warnings at every step of the way.

This is a "one time necessary change" with the change acting as a no-op if the version specified has passed.

alexdesiqueira:
I'm +1 here, but I have two questions to help others decide @hmaarrfk:

> A. Version wabisabi (well what we need from it)

How difficult (and heavy) would that be?

> B. Depend on it

How heavy would that be?

hmaarrfk:
The only dependency is numpydoc for documentation mangling.

That said, it is optional, and for many people it will be installed automatically when they install Spyder.

Versioning it would require copying in 3 files or so, maybe more for tests, but the license is already included in the files, and I can give my blessing to provide full copyright to scikit image if it helps.

rfezzani:
> This is a "one time necessary change" with the change acting as a no-op if the version specified has passed.

Even if the decorators acts punctually, I don't think that it is a good idea to keep ghost/useless lines in the code base... Even if it doesn't hurt to keep these lines, it would be a good idea to plan some cleanings at release time...

hmaarrfk:
Yes of course, but I don't think it can be considered "technical Dept" just an annoyance in the codebase that would require minimal review to be pushed through

skimage/version.py
```diff
@@ -0,0 +1,2 @@
+# Avoid circular imports
```
rfezzani: :cry: We really need to dive in `skimage` import strategy to fix it...

skimage/_shared/_deprecators.py
```diff
@@ -0,0 +1,11 @@
+from functools import partial
+from ..version import __version__
+from wabisabi import kwonly_change, default_parameter_change
+
+kwonly_change = partial(kwonly_change,
+                        library_name='scikit-image',
+                        current_library_version=__version__)
```
rfezzani: shouldn't the version number hard coded for each deprecation?

skimage/_shared/_deprecators.py
```diff
@@ -0,0 +1,11 @@
...
+
+default_parameter_change = partial(default_parameter_change,
+                                   library_name='scikit-image',
+                                   current_library_version=__version__)
```
rfezzani: Same question here?

skimage/_shared/_deprecators.py
hmaarrfk: This corresponds 0.17, not 1.0 today

rfezzani: I know, but I though that it will be used for docstring mangling... I mean to document when the changes have been introduced...

skimage/transform/hough_transform.py
```diff
@@ -4,9 +4,10 @@
                                _hough_ellipse,
                                _hough_line,
                                _probabilistic_hough_line as _prob_hough_line)
+from .._shared._deprecators import kwonly_change
 
-
-def hough_line_peaks(hspace, angles, dists, min_distance=9, min_angle=10,
+@kwonly_change('1.0')
```
jni: Ironically, I think this should be a keyword-only argument. 😂 

```suggestion
@kwonly_change(until_version='1.0')
```

skimage/transform/radon_transform.py
```diff
@@ -191,7 +193,10 @@ def _get_fourier_filter(size, filter_name):
 
 @deprecate_kwarg(kwarg_mapping={'filter': 'filter_name'},
                  removed_version="0.19")
-def iradon(radon_image, theta=None, output_size=None,
+@kwonly_change('1.0', previous_arg_order=['radon_image', 'theta',
+                                          'output_size', 'filter_name',
+                                          'interpolation', 'circle'])
```
jni: Whoa. This is a super-cool feature.

hmaarrfk:
Rescale behavior is a little tougher. We can try to write a spec for itto see how the input output response would look like.

I would rather not make breaking changes there, but I don't think I use the kwonly deprecator in my production code so maybe it is ok to break?

You can suggest a PR there if you like

jni:
> I would rather not make breaking changes there

Aw. But it would be so delightfully meta to use wabisabi on wabisabi itself! :joy: Anyway, I'll have a look. My main proposals right now are:

- `kwonly_change` -> `update_keyword_only[_args]`
- `version` -> `*, until_version`

The function definition itself can be simplified by using `toolz.curry` and having `func` be the only positional argument.

hmaarrfk:
do you really want to depend on toolz?

`until_version` makes sense

do you want all the functions to start with `update_`??

jni:
> do you really want to depend on toolz?

Yep, pretty lightweight, pure python, and extremely stable. Dask depends on it so it's not even like it's unmaintained.

> do you want all the functions to start with update_??

Maybe? I like my code to read as close to English as possible. And it's equivalent to having them all end with `_change`. ;)

grlee77:
A `deprecate_kwarg` utility was merged in #4158 and that same file also has decorators for `remove_arg` and `change_default_value` . 

Is there additional functionality here that we still want to consider or can this PR be closed?

hmaarrfk:
No. I don't think so. I was following your progress and it seems you added the important features.

PR #4465, opened 2020-02-18
Fix edgecase bugs in segmentation.relabel_sequential

uschmidt83:
## Description

- Fix edgecase bugs in `segmentation.relabel_sequential` (see modified tests).

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -121,21 +121,21 @@ def relabel_sequential(label_field, offset=1):
         raise ValueError("Offset must be strictly positive.")
     if np.min(label_field) < 0:
         raise ValueError("Cannot relabel array that contains negative values.")
-    m = label_field.max()
+    max_label = label_field.max()
     if not np.issubdtype(label_field.dtype, np.integer):
-        new_type = np.min_scalar_type(int(m))
+        new_type = np.min_scalar_type(int(max_label))
         label_field = label_field.astype(new_type)
-        m = m.astype(new_type)  # Ensures m is an integer
+        max_label = max_label.astype(new_type)  # Ensures max_label is an integer
     labels = np.unique(label_field)
     labels0 = labels[labels != 0]
-    required_type = np.min_scalar_type(offset + len(labels0))
+    new_max_label = offset - 1 + len(labels0)
+    new_labels0 = np.arange(offset, int(new_max_label + 1))
+    required_type = np.min_scalar_type(new_max_label)
     if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
         label_field = label_field.astype(required_type)
-    new_m = offset - 1 + len(labels0)
-    new_labels0 = np.arange(offset, int(new_m + 1))
-    forward_map = np.zeros(int(m + 1), dtype=label_field.dtype)
+    forward_map = np.zeros(int(max_label + 1), dtype=label_field.dtype)
     forward_map[labels0] = new_labels0
-    inverse_map = np.zeros(int(new_m + 1), dtype=label_field.dtype)
+    inverse_map = np.zeros(int(new_max_label + 1), dtype=label_field.dtype)
     inverse_map[offset:] = labels0
     relabeled = forward_map[label_field]
     return relabeled, forward_map, inverse_map
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -121,21 +121,21 @@ def relabel_sequential(label_field, offset=1):
         raise ValueError("Offset must be strictly positive.")
     if np.min(label_field) < 0:
         raise ValueError("Cannot relabel array that contains negative values.")
-    max_label = label_field.max()
+    max_label = int(label_field.max()) # Ensure max_label is an integer
     if not np.issubdtype(label_field.dtype, np.integer):
-        new_type = np.min_scalar_type(int(max_label))
+        new_type = np.min_scalar_type(max_label)
         label_field = label_field.astype(new_type)
-        max_label = max_label.astype(new_type)  # Ensures max_label is an integer
     labels = np.unique(label_field)
     labels0 = labels[labels != 0]
     new_max_label = offset - 1 + len(labels0)
-    new_labels0 = np.arange(offset, int(new_max_label + 1))
+    new_labels0 = np.arange(offset, new_max_label + 1)
+    output_type = label_field.dtype
     required_type = np.min_scalar_type(new_max_label)
     if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
-        label_field = label_field.astype(required_type)
-    forward_map = np.zeros(int(max_label + 1), dtype=label_field.dtype)
+        output_type = required_type
+    forward_map = np.zeros(max_label + 1, dtype=output_type)
     forward_map[labels0] = new_labels0
-    inverse_map = np.zeros(int(new_max_label + 1), dtype=label_field.dtype)
+    inverse_map = np.zeros(new_max_label + 1, dtype=output_type)
     inverse_map[offset:] = labels0
     relabeled = forward_map[label_field]
     return relabeled, forward_map, inverse_map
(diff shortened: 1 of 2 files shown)


skimage/segmentation/_join.py
```diff
@@ -132,7 +132,9 @@ def relabel_sequential(label_field, offset=1):
     if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
         label_field = label_field.astype(required_type)
     new_labels0 = np.arange(offset, offset + len(labels0))
-    if np.all(labels0 == new_labels0):
+    if offset == 1 and np.all(labels0 == new_labels0):
+        if not (labels == 0).any():
+            labels = np.concatenate(([0], labels))
         return label_field, labels, labels
```
uschmidt83: Maybe it would be best to not handle this special case separately and simply remove these lines of code.

skimage/segmentation/tests/test_join.py
```diff
@@ -30,9 +30,15 @@ def test_join_segmentations():
         join_segmentations(s1, s3)
 
 
+def _check_maps(ar, ar_relab, fw, inv):
+    assert_array_equal(fw[ar], ar_relab)
+    assert_array_equal(inv[ar_relab], ar)
```
jni: Ooh I like this. =)

skimage/segmentation/_join.py
jni: @uschmidt83 I'm actually confused about what's happening in this edge case. The test example has a lot going on so it's still not clear to me. Could you elaborate on the problem you're fixing and why it goes wrong in the current code?

uschmidt83: Hi @jni, these lines intended to check if `label_field` already contains sequential labels, hence there would be nothing to do. Unfortunately, the logic only works if `offset == 1` and `label_field` contains the 0 label. Hence, I added these things.

uschmidt83: This check was probably intended to avoid some computation, but it is unnecessary.
I just pushed a commit that (IMO) simplifies the function and makes it easier to understand and maintain.

skimage/segmentation/_join.py
```diff
@@ -131,14 +131,11 @@ def relabel_sequential(label_field, offset=1):
     required_type = np.min_scalar_type(offset + len(labels0))
     if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
         label_field = label_field.astype(required_type)
-    new_labels0 = np.arange(offset, offset + len(labels0))
-    if np.all(labels0 == new_labels0):
-        return label_field, labels, labels
+    new_m = offset - 1 + len(labels0)
```
jni: Is this short for `new_maximum` or `new_max_label`? If so, I wouldn't mind this being the new name rather than `new_m`.

uschmidt83: > Is this short for `new_maximum` or `new_max_label`?

Yes. After relabeling, `relabeled.max() == new_m`.

> If so, I wouldn't mind this being the new name rather than `new_m`.

I personally don't have an opinion on how to call this variable. I just chose the name because `m` is already used:

https://github.com/scikit-image/scikit-image/blob/7d70b5e7dd1f5e33326cb50b339aa0b5c13073c4/skimage/segmentation/_join.py#L124

jni: @uschmidt83 [who would write such code???](https://github.com/scikit-image/scikit-image/commit/dc377970f520f8f04a2b3d879983c3a1171518e7#diff-09660f97c6b4c3dc5d143d3be12e8aaaR85) :joy: :joy: :joy: 

Well, our preference has evolved to more self-explanatory variable names, so I would support a renaming of `m` to `max_label` and `new_m` to `new_max_label`. Feel free to leave that one to us, though, as indeed it is an orthogonal change to your contribution. Thank you for this fix!

uschmidt83: I just changed the variable names as requested.

skimage/segmentation/_join.py
```diff
@@ -121,24 +121,21 @@ def relabel_sequential(label_field, offset=1):
...
+        new_type = np.min_scalar_type(int(max_label))
         label_field = label_field.astype(new_type)
-        m = m.astype(new_type)  # Ensures m is an integer
+        max_label = max_label.astype(new_type)  # Ensures max_label is an integer
     labels = np.unique(label_field)
     labels0 = labels[labels != 0]
-    required_type = np.min_scalar_type(offset + len(labels0))
+    new_max_label = offset - 1 + len(labels0)
+    new_labels0 = np.arange(offset, int(new_max_label + 1))
+    required_type = np.min_scalar_type(new_max_label)
```
uschmidt83: I also changed another subtlety regarding `required_type`.
(It is amazing how much complexity can be in such a small function.)

jni: >  (It is amazing how much complexity can be in such a small function.)

I know! Thank you again for your persistence here! It's a much better function thanks to your contributions!

skimage/segmentation/_join.py
```diff
@@ -121,24 +121,21 @@ def relabel_sequential(label_field, offset=1):
...
-        new_type = np.min_scalar_type(int(m))
+        new_type = np.min_scalar_type(int(max_label))
         label_field = label_field.astype(new_type)
-        m = m.astype(new_type)  # Ensures m is an integer
+        max_label = max_label.astype(new_type)  # Ensures max_label is an integer
     labels = np.unique(label_field)
     labels0 = labels[labels != 0]
-    required_type = np.min_scalar_type(offset + len(labels0))
+    new_max_label = offset - 1 + len(labels0)
+    new_labels0 = np.arange(offset, int(new_max_label + 1))
```
rfezzani: No need for explicitely casting to int
```suggestion
    new_labels0 = np.arange(offset, new_max_label + 1)
```

skimage/segmentation/tests/test_join.py
rfezzani: :+1:

skimage/segmentation/_join.py
```diff
@@ -121,24 +121,21 @@ def relabel_sequential(label_field, offset=1):
...
+    required_type = np.min_scalar_type(new_max_label)
     if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
         label_field = label_field.astype(required_type)
-    new_labels0 = np.arange(offset, offset + len(labels0))
-    if np.all(labels0 == new_labels0):
-        return label_field, labels, labels
-    forward_map = np.zeros(int(m + 1), dtype=label_field.dtype)
+    forward_map = np.zeros(int(max_label + 1), dtype=label_field.dtype)
```
rfezzani: I think that you can directly use `new_type` here and avoid the `if` above
```suggestion
    forward_map = np.zeros(int(max_label + 1), dtype=new_type)
```

skimage/segmentation/_join.py
```diff
@@ -121,24 +121,21 @@ def relabel_sequential(label_field, offset=1):
...
     forward_map[labels0] = new_labels0
-    if not (labels == 0).any():
-        labels = np.concatenate(([0], labels))
-    inverse_map = np.zeros(offset - 1 + len(labels), dtype=label_field.dtype)
-    inverse_map[(offset - 1):] = labels
+    inverse_map = np.zeros(int(new_max_label + 1), dtype=label_field.dtype)
```
rfezzani: Here again
```suggestion
    inverse_map = np.zeros(int(new_max_label + 1), dtype=new_type)
```

skimage/segmentation/_join.py
uschmidt83: Hi @rfezzani, you are right about that. I was wrongly assuming that `int(v)` would not change the data type of `v` if it was "integer-based" (e.g. `np.uint16`).

I went ahead and changed the code some more, as seen in the latest commit.

skimage/segmentation/_join.py
uschmidt83: > I think that you can directly use `required_type` here and avoid the `if` above

In my last pull request (#3740), I changed the function to retain the data type of the input `label_field`. Hence, I would keep that. But you are right that it is not necessary to convert (i.e. make a copy of) `label_field` to the new data type. I changed that in the latest commit.

PR #4468, opened 2020-02-19
Added implementation of SSIM max/min algorithm from paper

henry-eigen:
## Description

Enhancement

Added ssim_optimize.py to transform. This file contains two functions, maximize_ssim and minimize_ssim, which respectively search for the image with the highest and lowest ssim to a reference image on a specified L-2 ball from the original image.

This is an implementation of the algorithm described in "Image quality assessment: From error visibility to structural similarity" (Wang, Z., Bovik, A. C., Sheikh, H. R. et al.) 

https://ece.uwaterloo.ca/~z70wang/publications/ssim.pdf

An example of the need for this algorithm is shown in "On the Suitability of Lp-norms for
Creating and Preventing Adversarial Examples" (Sharif, M., Bauer, L., Reiter, M. K.)

https://arxiv.org/pdf/1802.09653.pdf

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/ssim_optimize.py
+++ b/skimage/transform/ssim_optimize.py
@@ -1,10 +1,8 @@
 import numpy as np
 from numpy.linalg import norm
 
-from skimage.metrics._structural_similarity import (
-    structural_similarity
-    as ssim
-)
+from ..metrics._structural_similarity import structural_similarity as ssim
+
 
 def E(X, Y):
     return (Y - X) / norm(Y - X)
@@ -42,7 +40,7 @@ def optimize_func(img, adv_x, op, _lambda, iters, sigma):
     Compute the image with the highest or lowest SSIM to a reference image
     on the L-2 ball of the reference image. Follows a two step iterative
     gradient descent procedure.
-    
+
     For actual use, the minimize_ssim or maximize_ssim wrapper functions
     should be used.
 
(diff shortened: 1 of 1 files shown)


grlee77:
I think this is interesting and it is neat to see a concrete application of the gradient of the SSIM. I don't think we have any gallery examples making use of that currently.

If I understand this PR correctly:

1.) The proposed functions start by generating a noisy image at a fixed mean-squared error to the reference (the user can specify sigma to control the MSE). The optimization then produces the optimal image with an equal MSE to the reference as the noisy image, where "optimal" can be defined as having either the best or worst SSIM as compared to the reference.

2.) The proposed application is in augmenting training data for deep learning applications in computer-vision / image processing.

Does it make sense to also allow a user-specified "noisy" image (i.e. the initial `adv_x` that you currently create in `optimize_func`). In that way, couldn't the same optimization function also find application in image restoration by maximizing SSIM relative to a reference image?

henry-eigen:
Thanks for taking a look!  I am not sure what benefit the ability to specify a starting image provides. Since the reference image is needed anyways, I am not sure why the restoration process is necessary. The brief tests I ran suggest that the adv_x will converge to roughly the same image regardless of starting point, so the result would likely be the same. If I misunderstood your suggestion though, and you think adding that functionality would be helpful, I'd be more than happy to.

henry-eigen:
@grlee77  I refactored a few of the functions and added the ability to specify the starting image as you suggested.

PR #1909, opened 2016-01-25
Color difference: DeltaE cmc 0

arokem:
Addresses #1908?

PR #4469, opened 2020-02-20
Fix deltaE cmc close colors bug

### LINKED ISSUE #1908: deltaE_cmc can return NaN for valid input values
Simple to reproduce:

```
import numpy as np
from skimage import color

c = np.array([0., 1.59607713, 0.87755709])
print(color.deltaE_cmc(c, c))
```

rfezzani:
## Description

Fixes #1908 and closes #1909.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/delta_e.py
+++ b/skimage/color/delta_e.py
@@ -20,7 +20,7 @@ https://en.wikipedia.org/wiki/Color_difference
 
 import numpy as np
 
-from ..color.colorconv import lab2lch, _cart2polar_2pi
+from .colorconv import lab2lch, _cart2polar_2pi
 
 
 def deltaE_cie76(lab1, lab2):
@@ -116,7 +116,7 @@ def deltaE_ciede94(lab1, lab2, kH=1, kC=1, kL=1, k1=0.045, k2=0.015):
     dE2 = (dL / (kL * SL)) ** 2
     dE2 += (dC / (kC * SC)) ** 2
     dE2 += dH2 / (kH * SH) ** 2
-    return np.sqrt(dE2)
+    return np.sqrt(np.max(dE2, 0))
 
 
 def deltaE_ciede2000(lab1, lab2, kL=1, kC=1, kH=1):
@@ -238,7 +238,7 @@ def deltaE_ciede2000(lab1, lab2, kL=1, kC=1, kH=1):
     dE2 += C_term ** 2
     dE2 += H_term ** 2
     dE2 += R_term
-    ans = np.sqrt(dE2)
+    ans = np.sqrt(max(0, dE2))
     if unroll:
         ans = ans[0]
     return ans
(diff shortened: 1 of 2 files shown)


rfezzani:
Thank you @jni

hmaarrfk:
So i agree that this might address things in deltaE returning NaN, but it seems strange to pin this function to 0 in certain cases.


The challenge seems to be when computing `dE2` in, some pixels might actually return negative values.

I think you might as well use "clip" or an other formula that is guaranteed to be stable.

hmaarrfk:
Yeah, I think you want to use `np.max(0, 2 * term)` to ensure that you always return a value above 0 for small roundoff errors.

rfezzani:
You are right @hmaarrfk, I fixed it, but called `max` inside all the `sqrt` calls.

skimage/color/delta_e.py
```diff
@@ -116,7 +116,7 @@ def deltaE_ciede94(lab1, lab2, kH=1, kC=1, kL=1, k1=0.045, k2=0.015):
     dE2 = (dL / (kL * SL)) ** 2
     dE2 += (dC / (kC * SC)) ** 2
     dE2 += dH2 / (kH * SH) ** 2
-    return np.sqrt(dE2)
+    return np.sqrt(np.maximum(dE2, 0))
```
hmaarrfk:
```suggestion
    return np.sqrt(np.maximum(dE2, 0, maximum=dE2))
```

skimage/color/delta_e.py
```diff
@@ -238,7 +238,7 @@ def deltaE_ciede2000(lab1, lab2, kL=1, kC=1, kH=1):
     dE2 += C_term ** 2
     dE2 += H_term ** 2
     dE2 += R_term
-    ans = np.sqrt(dE2)
+    ans = np.sqrt(np.maximum(0, dE2))
```
hmaarrfk:
```suggestion
    return np.sqrt(np.maximum(dE2, 0, maximum=dE2))
```

skimage/color/delta_e.py
```diff
@@ -304,7 +304,8 @@ def deltaE_cmc(lab1, lab2, kL=1, kC=1):
     dE2 = (dL / (kL * SL)) ** 2
     dE2 += (dC / (kC * SC)) ** 2
     dE2 += dH2 / (SH ** 2)
-    return np.sqrt(dE2)
+
+    return np.sqrt(np.maximum(0, dE2))
```
hmaarrfk:
```suggestion
    return np.sqrt(np.maximum(dE2, 0, out=dE2))
```

rfezzani:
I tried your suggestion @hmaarrfk, but it raises a `TypeError: return arrays must be of ArrayType` when a single color is provided :-(

rfezzani:
Thank you @hmaarrfk, Are the added comments good enough?

emmanuelle:
Merging, thanks @rfezzani !

rfezzani:
Thank you @emmanuelle :tada:

PR #4470, opened 2020-02-21
Fix bool array warping

### LINKED ISSUE #4292: Rescaling boolean arrays yields plausible result or array of zeros depending on the scale factor
## Description
Hello!
I am using `skimage.transform.rescale` to down-scale boolean arrays that have a single connected component.
Depending on the scale factor selected, the output yields a result that seems correct or an array full of zeros.

If I convert the boolean array to float, the result is different than when I pass the array with values as bool or int. 
Setting 'anti-aliasing' to False yields the same result for int, bool or float values in the array.

Furthermore, the expression given in the `rescale` function documentation to compute the "anti_aliasing_sigma" is different than the actual expression used inside `skimage.transform.resize` (which is called in `skimage.transform.rescale`):
- `skimage.transform.rescale`: "By default, this value is chosen as (1 - s) / 2 where s is the down-scaling factor."
- `skimage.transform.resize`: "By default, this value is chosen as (s - 1) / 2 where s is the down-scaling factor, where s > 1."

## Way to reproduce
```python
import numpy as np
from skimage.transform import rescale

input_matrix_shape = (1000, 1200, 400)
high_resolution_matrix = np.zeros(input_matrix_shape)
high_resolution_matrix[200:800, 300:900, 100:300] = 1 # Fill matrix with some ones
high_resolution_matrix = high_resolution_matrix.astype(np.bool_)
scale_factor_list = [0.625, 0.305]

for current_scale_factor in scale_factor_list:
    rescaled_matrix = rescale(high_resolution_matrix, current_scale_factor, order = 0, anti_aliasing=True, multichannel=False, preserve_range=True)
    print('Scale factor: {}'.format(current_scale_factor))
    print('\tResult shape: {}'.format(rescaled_matrix.shape))
    print('\tTotal non-zero values: {}'.format(np.sum(rescaled_matrix > 0)))
    print()
```

## Version information
```python
```

```python
3.7.5 (default, Oct 17 2019, 12:21:00) 
[GCC 8.3.1 20190223 (Red Hat 8.3.1-2)]
Linux-5.2.17-100.fc29.x86_64-x86_64-with-fedora-29-Twenty_Nine
scikit-image version: 0.16.2
numpy version: 1.17.3
```

rfezzani:
## Description

Fixes #4292.

This PR proposes
- to set default spline interpolation's order to 0 if input array dtype is bool and 1 otherwise. The function `skimage._shared.utils._set_order` is created for this purpose and is used wherever the `order` argument is used (`skimage.transform._warps` module and `skimage.measure.profile.profile_line` function).
- to set default anti_aliasing parameter to False if input array dtype is bool and True otherwise.

~~No deprecation cycle is started since it is a bug fix, do you think it should?~~

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


rfezzani:
Gentle up for this PR :slightly_smiling_face:

skimage/_shared/utils.py
```diff
@@ -286,3 +286,41 @@ def convert_to_float(image, preserve_range):
     else:
         image = img_as_float(image)
     return image
+
+
+def _set_order(image_dtype, order):
```
emmanuelle, line 291: `_set_interpolation_order` would be more informative

emmanuelle: `_validate_interpolation_order` or `_coerce_interpolation_order` would be even better.

skimage/transform/_warps.py
```diff
@@ -60,9 +61,10 @@ def resize(image, output_shape, order=1, mode='reflect', cval=0, clip=True,
         image is converted according to the conventions of `img_as_float`.
         Also see https://scikit-image.org/docs/dev/user_guide/data_types.html
     anti_aliasing : bool, optional
-        Whether to apply a Gaussian filter to smooth the image prior to
-        down-scaling. It is crucial to filter when down-sampling the image to
-        avoid aliasing artifacts.
+        Whether to apply a Gaussian filter to smooth the image prior
+        to down-scaling. It is crucial to filter when down-sampling
+        the image to avoid aliasing artifacts. If input image data
+        type is bool, no anti-aliasing is applyed.
```
emmanuelle, line 67:
```suggestion
        type is bool, no anti-aliasing is applied.
```

skimage/transform/_warps.py
```diff
@@ -231,9 +246,10 @@ def rescale(image, scale, order=1, mode='reflect', cval=0, clip=True,
         Whether the last axis of the image is to be interpreted as multiple
         channels or another spatial dimension.
     anti_aliasing : bool, optional
-        Whether to apply a Gaussian filter to smooth the image prior to
-        down-scaling. It is crucial to filter when down-sampling the image to
-        avoid aliasing artifacts.
+        Whether to apply a Gaussian filter to smooth the image prior
+        to down-scaling. It is crucial to filter when down-sampling
+        the image to avoid aliasing artifacts. If input image data
+        type is bool, no anti-aliasing is applyed.
```
emmanuelle, line 252:
```suggestion
        type is bool, no anti-aliasing is applied.
```

emmanuelle:
Thanks for the PR @rfezzani. So we're doing something different for bool and uint8 arrays with this change (was it already the case?). I don't mind this but should this be said in an even louder way in the docstrings?

rfezzani:
Thank you @emmanuelle for your review!
The bugs and issues related to bool array transformation comes from the fact that interpolation and linear combination is not defined for boolean values. So this PR simply proposes to avoid anti-aliasing and interpolation with `bool` arrays.
The behavior of `uint8` arrays doesn't change here (linear interpolation and anti-aliasing by default)

rfezzani:
:tada: Thank you @soupault!

emmanuelle:
Thank you @rfezzani merging!

hmaarrfk:
Thanks!

I think thisis a really nice way to address this issue!

PR #4471, opened 2020-02-21
Fix bool array profile_line

### LINKED ISSUE #2258: profile_line gives wrong output for a binary mask
## Description

As of now (skimage 0.12.3), profile_line sometimes returns wrong output for binary masks (np.array(dtype=bool)). See below for an example. If the image is cast to an uint8 first, it returns the expected result.

A possible patch might be even just specifying the allowed dtypes in the docstring.
## Way to reproduce

``` python
from numpy import meshgrid, pi, cos, sin, all, uint8
from skimage.measure import profile_line

shape = (200, 200)
center_x, center_y = (140, 150)
radius = 20
x, y = meshgrid(range(shape[1]), range(shape[0]))
mask = (y - center_y)**2 + (x - center_x)**2 < radius**2
src = (center_y, center_x)
phi = 4*pi/9.
dy = 31 * cos(phi)
dx = 31 * sin(phi)
dst = (center_y + dy, center_x + dx)

profile_uint8 = profile_line(mask.astype(uint8), src, dst)
print profile_uint8
assert all(profile_uint8[:radius] == 1)

profile_bool = profile_line(mask, src, dst)
print profile_bool
assert all(profile_bool[:radius] == 1) # Fails

assert all(profile_bool == profile_uint8) # Fails
```

The output is (there is a zero in the middle of the profile_line output):

```
[ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.
  1.  1.  1.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.]
[ 1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  1.  0. (ZERO HERE?!)  1.  1.
  1.  1.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.  0.]

AssertionError                            Traceback (most recent call last)
<ipython-input-24-1a94903a5586> in <module>()
     19 profile_bool = profile_line(mask, src, dst)
     20 print profile_bool
---> 21 assert all(profile_bool[:radius] == 1) # Fails
     22 
     23 assert all(profile_bool == profile_uint8) # Fails

AssertionError: 
```

rfezzani:
## Description

Fixes #2258.

This PR proposes to set default spline interpolation's order to 0 if input array dtype is bool and 1 otherwise.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/profile.py
+++ b/skimage/measure/profile.py
@@ -94,7 +94,9 @@ def profile_line(image, src, dst, linewidth=1,
     if image.dtype == bool and order != 0:
         warn("Input image dtype is bool. Interpolation is not defined "
              "with bool data type. Please set order to 0 or explicitely "
-             "cast input image to another data type.", stacklevel=2)
+             "cast input image to another data type. Starting from version "
+             "0.19 a ValueError will be raised instead of this warning.",
+             FutureWarning, stacklevel=2)
 
     perp_lines = _line_profile_coordinates(src, dst, linewidth=linewidth)
     if image.ndim == 3:
--- a/skimage/measure/profile.py
+++ b/skimage/measure/profile.py
@@ -22,9 +22,7 @@ def profile_line(image, src, dst, linewidth=1,
     linewidth : int, optional
         Width of the scan, perpendicular to the line
     order : int in {0, 1, 2, 3, 4, 5}, optional
-        The order of the spline interpolation, default is 1. The order has to
         The order of the spline interpolation, default is 0 if
-        be in the range 0-5. See `skimage.transform.warp` for detail.
         image.dtype is bool and 1 otherwise. The order has to be in
         the range 0-5. See `skimage.transform.warp` for detail.
     mode : {'constant', 'nearest', 'reflect', 'mirror', 'wrap'}, optional
@@ -90,31 +88,24 @@ def profile_line(image, src, dst, linewidth=1,
            [1.        , 1.        , 0.        ],
            [1.41421356, 1.41421356, 0.        ]])
     """
-    image_is_bool = image.dtype == bool
-
     if order is None:
-        if image_is_bool:
-            order = 0
-        else:
-            order = 1
+        order = 0 if image.dtype == bool else 1
 
-    if image_is_bool and order != 0:
+    if image.dtype == bool and order != 0:
         warn("Input image dtype is bool. Interpolation is not defined "
              "with bool data type. Please set order to 0 or explicitely "
-             "cast input image to an other data type.")
-
-    prefilter = order > 1
+             "cast input image to another data type.", stacklevel=2)
 
     perp_lines = _line_profile_coordinates(src, dst, linewidth=linewidth)
     if image.ndim == 3:
         pixels = [ndi.map_coordinates(image[..., i], perp_lines,
-                                      prefilter=prefilter,
+                                      prefilter=order > 1,
                                       order=order, mode=mode,
                                       cval=cval) for i in
                   range(image.shape[2])]
         pixels = np.transpose(np.asarray(pixels), (1, 2, 0))
     else:
-        pixels = ndi.map_coordinates(image, perp_lines, prefilter=prefilter,
+        pixels = ndi.map_coordinates(image, perp_lines, prefilter=order > 1,
                                      order=order, mode=mode, cval=cval)
     # The outputted array with reduce_func=None gives an array where the
     # row values (axis=1) are flipped. Here, we make this consistent.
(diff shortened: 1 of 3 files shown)


skimage/measure/profile.py
```diff
@@ -86,14 +90,31 @@ def profile_line(image, src, dst, linewidth=1,
...
+    if order is None:
+        if image_is_bool:
+            order = 0
+        else:
+            order = 1
+
+    if image_is_bool and order != 0:
+        warn("Input image dtype is bool. Interpolation is not defined "
+             "with bool data type. Please set order to 0 or explicitely "
+             "cast input image to an other data type.")
```
jni: stacklevel=2?

also: an other -> another

Finally, I wonder whether we should simply raise here, since the output is liable to be erroneous here...

skimage/measure/profile.py
```diff
@@ -86,14 +90,31 @@ def profile_line(image, src, dst, linewidth=1,
...
+
+    prefilter = order > 1
+
     perp_lines = _line_profile_coordinates(src, dst, linewidth=linewidth)
     if image.ndim == 3:
         pixels = [ndi.map_coordinates(image[..., i], perp_lines,
-                                      order=order, mode=mode, cval=cval)
-                  for i in range(image.shape[2])]
+                                      prefilter=prefilter,
```
jni: Again, I don't think it's necessary to define `prefilter`, just write `prefilter=(order > 1),` in the function call.

skimage/measure/profile.py
rfezzani: I am not against raising but it is a little bit aggressive. May be can we then start a deprecation cycle specifying that starting from version 0.19 an error will be raised in this case.?

rfezzani:
Thank you @jni for your review. I was wondering we can not use for consistency the [`_set_order`](https://github.com/scikit-image/scikit-image/pull/4470/files#diff-c0f33f5600401be736e79b65164b4c81R688) defined in #4470?

skimage/measure/profile.py
```diff
@@ -21,8 +22,11 @@ def profile_line(image, src, dst, linewidth=1,
     linewidth : int, optional
         Width of the scan, perpendicular to the line
     order : int in {0, 1, 2, 3, 4, 5}, optional
-        The order of the spline interpolation to compute image values at
-        non-integer coordinates. 0 means nearest-neighbor interpolation.
+        The order of the spline interpolation, default is 1. The order has to
+        The order of the spline interpolation, default is 0 if
+        be in the range 0-5. See `skimage.transform.warp` for detail.
+        image.dtype is bool and 1 otherwise. The order has to be in
+        the range 0-5. See `skimage.transform.warp` for detail.
```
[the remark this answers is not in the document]
jni: Erm, `scipy.ndimage.map_coordinates`, surely?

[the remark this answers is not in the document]
rfezzani: I wanted to make the documentation consistent with `skimage.transform.resize` ans `skimage.transform.rescale`.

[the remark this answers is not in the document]
jni: Personally I think this should point to the underlying function, but your point is taken. Let’s wait for the second review.

skimage/measure/profile.py
```diff
@@ -86,14 +88,24 @@ def profile_line(image, src, dst, linewidth=1,
            [1.        , 1.        , 0.        ],
            [1.41421356, 1.41421356, 0.        ]])
     """
+    if order is None:
+        order = 0 if image.dtype == bool else 1
+
+    if image.dtype == bool and order != 0:
+        warn("Input image dtype is bool. Interpolation is not defined "
```
hmaarrfk: Why not just raise? I'm affraid that setting this precedent means that we will have to support this kind of operation almost indefinitely greatly increasing the scope of the API

rfezzani: > I am not against raising but it is a little bit aggressive. May be can we then start a deprecation cycle specifying that starting from version 0.19 an error will be raised in this case.?

rfezzani:
I think that we all agree to say that order > 0 for bool array must not be supported. The question is when to start raising an error, now or in version 0.19...
Starting a deprecation cycle gives some time to affected users to fix their code without breaking it...

hmaarrfk:
All good when passes on Travis.

rfezzani:
Thank you @hmaarrfk :tada:

PR #4472, opened 2020-02-21
Fix values link in governance

stefanv:
This tidies up the docs a bit:

- Fixes a link to our values from the governance doc
- Better structures the documentation index
- Removes mentions of SciKit capitalization, which can be confusing when trying to figure out the package name


doc/source/index.rst
```diff
@@ -76,17 +76,14 @@ Sections
 
    * - `SKIPs <skips/index.html>`_
 
-     - 
-
-       SciKit-Image Proposals, documents describing major changes to the
+       **s**\ ci\ **k**\ it-\ **i**\ mage **p**\ roposals, documents describing major changes to the
```
jni: Well, this formatting is hideous, but I'll allow it. :joy:

doc/source/skips/1-governance.rst
```diff
@@ -107,7 +107,7 @@ van der Walt, Juan Nunez-Iglesias, Emmanuelle Gouillart, Josh Warner, and
 Zachary Pincus. The SC membership is revisited every January. SC members who do
 not actively engage with the SC duties are expected to resign. New members are
 added by nomination by a core developer. Nominees should have demonstrated
-long-term, continued commitment to the project and its :ref:`values`. A
+long-term, continued commitment to the project and its :ref:`values <values>`. A
```
jni: This seems like it should be unnecessary? Do you know why it's needed?

PR #4480, opened 2020-02-27
reduce runtime of non-local means tests

grlee77:
## Description

This PR applies a couple of small fixes to the tests for `denoise_nl_means`:

1.) I set `patch_distance` to a reduced value in `test_denoise_nl_means_3d_dtype` and `test_denoise_nl_means_2d_dtype`. For the 3D case this reduces runtime from around 10 seconds to < 1 second. The patch distance is not important for this test as it is just verifying the dtype of the output.

2.) Two other tests involving 2D data seemed to have been setting `multichannel=True` when it seems the value for this case should be `multichannel=False`.

## Checklist

## For reviewers

  later.

### DIFF
--- a/skimage/restoration/tests/test_denoise.py
+++ b/skimage/restoration/tests/test_denoise.py
@@ -318,14 +318,14 @@ def test_denoise_nl_means_2d(fast_mode):
     for s in [sigma, 0]:
         denoised = restoration.denoise_nl_means(img, 7, 5, 0.2,
                                                 fast_mode=fast_mode,
-                                                multichannel=True,
+                                                multichannel=False,
                                                 sigma=s)
         # make sure noise is reduced
         assert_(img.std() > denoised.std())
 
         denoised_f32 = restoration.denoise_nl_means(img_f32, 7, 5, 0.2,
                                                     fast_mode=fast_mode,
-                                                    multichannel=True,
+                                                    multichannel=False,
                                                     sigma=s)
         # make sure noise is reduced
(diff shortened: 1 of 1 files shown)


grlee77:
It is a separate issue, that the default `patch_distance` of 11 is pretty huge for 3D. That results in a search window extent (2*11 + 1)**3 = 12,167 pixels!

skimage/restoration/tests/test_denoise.py
```diff
@@ -318,14 +318,14 @@ def test_denoise_nl_means_2d(fast_mode):
     for s in [sigma, 0]:
         denoised = restoration.denoise_nl_means(img, 7, 5, 0.2,
                                                 fast_mode=fast_mode,
-                                                multichannel=True,
+                                                multichannel=False,
```
rfezzani: Good catch! :wink:

skimage/restoration/tests/test_denoise.py
```diff
@@ -448,13 +448,13 @@ def test_denoise_nl_means_3d_dtype(fast_mode):
 
     with expected_warnings(['Image dtype is not float']):
         assert restoration.denoise_nl_means(
-            img, fast_mode=fast_mode).dtype == 'float64'
+            img, patch_distance=2, fast_mode=fast_mode).dtype == 'float64'
```
rfezzani: Excellent remark :+1:

soupault:
Thank you @grlee77 !

PR #4483, opened 2020-02-29
Add pooch as our data manager

jni:
## Description

This PR adds pooch as a new dependency and adds the cells.tif dataset from our tutorials as the single external dataset for now. In the future it enables scikit-image to provide much more diverse data e.g. in tutorials.

Supersedes #3945. 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


hmaarrfk:
I think this needs a rebase... thats where i got stuck.

emmanuelle:
@hmaarrfk CI is failing for something introduced in this PR (which was not in your original one). @jni you can also take it from #4484 where I merged master (since we squashing I guess it's ok to merge instead of rebasing?).

skimage/data/setup.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
+from ._registry import legacy_datasets
```
emmanuelle: this line makes CI fail, please remove the relative import

jni: @emmanuelle do you have any idea why this fails, though?

hmaarrfk: I think setup is run prior to install

emmanuelle: ah ok interesting, thanks @hmaarrfk. But removing the relative import also fails the CI... Should this be `from skimage.data._registry import legacy_datasets`?

PR #4485, opened 2020-03-03
[WIP] Add binder specific requirements config

betatim:
This PR is a proposal for how to setup things to convert all the examples from the documentation gallery to notebooks that users can try on mybinder.org

The idea is to have a specific `.binder/requirements.txt` which installs the packages needed to run the examples. I tried to reuse the files in `requirements/` but ran into an issue with duplicate dependency specification. Maybe someone who knows a bit more about how things are organised.

You can launch it via https://mybinder.org/v2/gh/betatim/scikit-image/binder-examples

This came out of https://github.com/jupyter/repo2docker/issues/853

## For reviewers

  later.

### DIFF
--- a/.binder/postBuild
+++ b/.binder/postBuild
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+set -e
+
+# Taken from https://github.com/scikit-learn/scikit-learn/blob/72b3041ed57e42817e4c5c9853b3a2597cab3654/.binder/postBuild
+# under BSD3 license, copyright the scikit-learn contributors
+
+python -m pip install .
+
+# This script is called in a binder context. When this script is called, we are
+# inside a git checkout of the scikit-image/scikit-image repo. This script
+# generates notebooks from the scikit-image python examples.
+
+if [[ ! -f /.dockerenv ]]; then
+    echo "This script was written for repo2docker and is supposed to run inside a docker container."
+    echo "Exiting because this script can delete data if run outside of a docker container."
+    exit 1
+fi
+
+# Copy content we need from the scikit-image repo
+TMP_CONTENT_DIR=/tmp/scikit-image
+mkdir -p $TMP_CONTENT_DIR
+cp -r doc/examples .binder $TMP_CONTENT_DIR
+# delete everything in current directory including dot files and dot folders
+# to create a "clean" experience for readers
+find . -delete
+
+# Generate notebooks and remove other files from examples folder
+GENERATED_NOTEBOOKS_DIR=notebooks
+cp -r $TMP_CONTENT_DIR/examples $GENERATED_NOTEBOOKS_DIR
+
+find $GENERATED_NOTEBOOKS_DIR -name '*.py' -exec sphx_glr_python_to_jupyter.py '{}' +
+NON_NOTEBOOKS=$(find $GENERATED_NOTEBOOKS_DIR -type f | grep -v '\.ipynb')
+rm -f $NON_NOTEBOOKS
+
+# Put the .binder folder back (may be useful for debugging purposes)
+mv $TMP_CONTENT_DIR/.binder .
+# Final clean up
+rm -rf $TMP_CONTENT_DIR
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,0 +1,11 @@
+-r ../requirements/default.txt
+# Explicitly adding Cython instead of using `requirements/build.txt`
+# because we'd get a double numpy dependency otherwise
+Cython>=0.29.13
+sphinx-gallery
+scikit-learn
+dask[array]>=0.15.0
+# cloudpickle is necessary to provide the 'processes' scheduler for dask
+cloudpickle>=0.2.1
+pandas>=0.23.0
+seaborn>=0.7.1


emmanuelle:
Hi @betatim great to see you on the scikit-image repo :-). Thanks a lot for your contribution! Did you see that we had another pull request #4340 also aiming at adding binder links? I had tried to use the sphinx-gallery machinery for this. One thing we were a bit hesitant is which version of scikit-image should run on binder (master or which release). On https://scikit-image.org/ we have several versions of the docs (several releases and master), ideally launching binder from a specific example and doc version should use the same version of scikit-image. I'll take a closer look at your PR to see how it compares to the open one, thanks again!

emmanuelle:
oops ok I should have followed the link, this was discussed with @stefanv already. Let me read the PR first!

.binder/postBuild
```diff
@@ -0,0 +1,39 @@
...
+cp -r doc/examples .binder $TMP_CONTENT_DIR
+# delete everything in current directory including dot files and dot folders
+# to create a "clean" experience for readers
+find . -delete
+
+# Generate notebooks and remove other files from examples folder
+GENERATED_NOTEBOOKS_DIR=notebooks
+cp -r $TMP_CONTENT_DIR/examples $GENERATED_NOTEBOOKS_DIR
+
+find $GENERATED_NOTEBOOKS_DIR -name '*.py' -exec sphx_glr_python_to_jupyter.py '{}' +
```
emmanuelle: `sphx_glr_python_to_jupyter.py` is somewhere in the path thanks to `sphinx-gallery`, right?

emmanuelle:
ok, so the idea is that we would give binder the url of the different branches (0.17.x, etc.) and then the notebooks would be built within this branch? Some of our examples take some time to run, can you confirm that the container is cached (hence the notebooks as well) unless there has been a new commit since the last time the container was used?

.binder/postBuild
betatim: Nods.

stefanv:
@emmanuelle I think this totally solves our problem, albeit in a slightly counter-intuitive way :)  Let's figure out how to use our requirements.txt files, then we're good to go.

@betatim Thanks very much for the POC!

emmanuelle:
@stefanv , yes, I think we just have to modify the `conf.py` to point to the master branch and at release time to point to the release branch. @betatim should we take over from your POC for these adjustments?

stefanv:
I think @betatim provided this as a starting point for us, so go ahead and take the PR.

betatim:
I'd be happy if you want to take over or revive your original PR. I'd also be happy to keep working on it (as an excuse to understand sphinx-gallery a bit better).

betatim:
This is the snippet used on scikit-learn to configure different branches for different versions of the docs build:

https://github.com/scikit-learn/scikit-learn/blob/8910b14fa8630c4d9ef477a2d339fdf68eecd4b9/doc/conf.py#L302-L308

The `binder_branch` variable is defined around L257. This is the bit that creates links to mybinder.org with different branch names for different versions.

> Some of our examples take some time to run, can you confirm that the container is cached (hence the notebooks as well) unless there has been a new commit since the last time the container was used?

Short answer: yes. Longer answer: when you visit a link like https://mybinder.org/v2/gh/binder-examples/requirements/master we first resolve the `master` (which also could be a SHA, tag or branch) to a SHA. Then we check our docker image cache for an image named `munged-and-escaped-version-of-binder-examples-requirements:SHA`. If that exists we launch it. If we can't find it we build it. This means if the SHA a branch resolves to changes we build a new image.

There is one more (frequent) reason why we might not find a image in our cache even though nothing has changed: we operate four clusters that don't share a cache. This means you can get sent to "the wrong" cluster. We try very hard to always send the same repo to the same cluster but sometimes the cluster is full or down or something else is up. In those cases we send users to a different cluster where we need to rebuild the image. (Sometimes, less than once every 6months(??) we might also remove "rarely used" images from our cache but we don't do that very often.)

---

I don't think that `sphx_glr_python_to_jupyter.py` tries to execute the examples while converting them to notebooks. At least on my laptop it took "no time at all" to do the conversion, which I think means no execution happened.

---

Next steps I'd take:
* integrate sphinx gallery config and understand how to create the correct value for`binder_branch` 
* figure out a nice way to deal with what dependencies need to be installed and how to manage those (especially across releases)
* check which path sphinx gallery expects notebooks to be at and adjust `postBuild` to move the notebooks there. In sklearn it seems to be `notebooks/auto_examples/<example-name.ipynb>` , this is what `?urlpath=lab/tree/notebooks/auto_examples/plot_anomaly_comparison.ipynb` in the [launch URLs](https://mybinder.org/v2/gh/scikit-learn/scikit-learn/0.22.X?urlpath=lab/tree/notebooks/auto_examples/plot_anomaly_comparison.ipynb) tells you. The `lab` bit is so that users get JupyterLab instead of classic notebook (I prefer classic notebook but that is by the by)
* understand how to test that sphinx-gallery generates the right links. Probably by running a docs build locally. Unsure how to test for a branch that isn't `master` though. Maybe I have to learn how to backport stuff??
* understand what happens if the docs for an old release are rebuilt. Will they end up containing links to mybinder.org? If yes do we need to backport config stuff? Should they even contain links?

(sorry for the very long post)

rfezzani:
Closing, since #4543 is merged. Thank you again @betatim ;-)

PR #4487, opened 2020-03-03
reduced the duration of the longest tests

emmanuelle:
Two tests were taking much more time that the others, this PR makes them shorter. On my machine this has changed the total duration of the test suite from 3 to 2 minutes.


emmanuelle:
Oh wow, and CI is green on top ;-)

PR #4493, opened 2020-03-05
Random ellipses

### LINKED ISSUE #4491: Possibility to have ellipses in random shapes
## Description

It could be interesting to have the possibility to draw random ellipses when using the [`random_shapes` function](https://scikit-image.org/docs/dev/auto_examples/edges/plot_random_shapes.html).
Random ellipses are often used as a toy example for different research papers, e.g.:
- https://arxiv.org/abs/2001.01258
- https://arxiv.org/abs/1707.06474

I can definitely try and submit a PR for this.

zaccharieramzi:
## Description
This PR introduces the possibility to draw ellipses when using `random_shapes`. It closes #4491 .

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/draw/_random_shapes.py
+++ b/skimage/draw/_random_shapes.py
@@ -325,7 +325,7 @@ def random_shapes(image_shape,
         Number of channels in the generated image. If 1, generate monochrome
         images, else color images with multiple channels. Ignored if
         ``multichannel`` is set to False.
-    shape : {rectangle, circle, triangle, None} str, optional
+    shape : {rectangle, circle, triangle, ellipse, None} str, optional
         The name of the shape to generate or `None` to pick random ones.
     intensity_range : {tuple of tuples of uint8, tuple of uint8}, optional
         The range of values to sample pixel values from. For grayscale images
--- a/skimage/draw/_random_shapes.py
+++ b/skimage/draw/_random_shapes.py
@@ -225,11 +225,11 @@ def _generate_ellipse_mask(point, image, shape, random):
         rotation=rotation,
     )
     max_radius = math.ceil(max(r_radius, c_radius))
-    # NOTE: again taking max radius is very conservative, we could look into
-    # computing the exact bounding box, using e.g.
-    # https://gist.github.com/smidm/b398312a13f60c24449a2c7533877dc0
-    label = ('ellipse', ((point[0] - max_radius + 1, point[0] + max_radius),
-                         (point[1] - max_radius + 1, point[1] + max_radius)))
+    min_x = np.min(ellipse[0])
+    max_x = np.max(ellipse[0]) + 1
+    min_y = np.min(ellipse[1])
+    max_y = np.max(ellipse[1]) + 1
+    label = ('ellipse', ((min_x, max_x), (min_y, max_y)))
 
     return ellipse, label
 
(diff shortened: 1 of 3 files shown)


zaccharieramzi:
The travis failure is only in Python 3.8 on `TestRank.test_rank_filter`, so I don't think it's from me.

skimage/draw/_random_shapes.py
```diff
@@ -163,11 +165,82 @@ def _generate_triangle_mask(point, image, shape, random):
...
+        r_radius,
+        c_radius,
+        shape=image[:2],
+        rotation=rotation,
+    )
+    max_radius = math.ceil(max(r_radius, c_radius))
+    # NOTE: again taking max radius is very conservative, we could look into
+    # computing the exact bounding box, using e.g.
+    # https://gist.github.com/smidm/b398312a13f60c24449a2c7533877dc0
+    label = ('ellipse', ((point[0] - max_radius + 1, point[0] + max_radius),
```
emmanuelle: Here I think users expect a tight bounding box indeed. You can use the mix and max values of the indices in `ellipse` I guess?

skimage/draw/_random_shapes.py
```diff
@@ -163,11 +165,82 @@ def _generate_triangle_mask(point, image, shape, random):
...
+    bottom = image[0] - point[0]
+    available_radius = min(left, right, top, bottom, max_radius)
+    if available_radius < min_radius:
+        raise ArithmeticError('cannot fit shape to image')
+    # NOTE: very conservative because we could take into account the fact that
+    # we have 2 different radii, but this is a good first approximation.
+    # Also, we can afford to have a uniform sampling because the ellipse will
+    # be rotated.
+    r_radius = random.uniform(min_radius, available_radius + 1)
+    c_radius = random.uniform(min_radius, available_radius + 1)
```
emmanuelle: I was wondering if here we want uniform sampling on the two radii or on one radius and on the aspect ratio (say min over max radii) but from a quick trial it does not seem to make a big difference, so happy with this sampling choice.

emmanuelle:
Thank you for your PR! It looks good to me, I just think the bounding box should be tighter.

Could you please also
- add `ellipse` in the docstring of random_shapes here https://github.com/scikit-image/scikit-image/pull/4493/files#diff-4aa1f84dc766e3ff3d84b6348c8a0de0L255
- run the gallery example with random shapes to check that it does add ellipses (you can post a screenshot here), and in this example fix the seed because sometimes the rendered images do not look great, for example https://scikit-image.org/docs/dev/auto_examples/edges/plot_random_shapes.html

zaccharieramzi:
random_shapes_contrib
Does this look good enough?

emmanuelle:
Thanks for your reactivity, I'll approve while you push the seed in example commit.

zaccharieramzi:
Oh I thought there was a seed, I didn't pay attention.
random_shapes_contrib
This is what I got with `random_seed=0`. There is btw a mismatch in the `random_shapes` doc between the name of the actual argument and the one reported in the docs. Should I take advantage of this PR to change it?

emmanuelle:
Yes please for the `seed/random_seed`!

emmanuelle:
I'm merging this this PR had had two :+1: since one week.

PR #4501, opened 2020-03-10
Fix peak_local_max coordinates ordering

rfezzani:
## Description

When `indices` is True, `peak_local_max` returns the coordinates of the peaks. These coordinates are only sorted if the number of peaks is greater then `num_peaks`.

This PR proposes to always sort the coordinates according to the peaks values.

## For reviewers

  later.


skimage/transform/tests/test_warps.py
```diff
@@ -613,8 +613,9 @@ def test_log_warp_polar():
         image[rr, cc] = val
     warped = warp_polar(image, radius=200, scaling='log')
     profile = warped.mean(axis=0)
-    peaks = peak_local_max(profile)
-    gaps = peaks[:-1]-peaks[1:]
+    peaks_coord = peak_local_max(profile)
+    peaks_coord.sort(axis=0)
```
jni: What's happening here? You're independently sorting the row and column coordinates of the peaks? :thinking:

rfezzani: No, the coordinates are 1D! The coordinates being returned sorted according to peaks values, there is no guaranty that they are in ascending order (it is in fact not the case). The `gaps` variable computation in the next line makes then no sens.

sciunto:
Thanks @rfezzani

Issue #3494, opened 2018-10-22
Return cross corelation images for all our matching algorithms


sciunto:
## Description

See this comment: https://github.com/scikit-image/scikit-image/pull/3334#issuecomment-428586974

The idea is to provide cross correlation (CC) images for our different algorithms. Currently, 'match_template' does not.
Related idea, put this CC functions in a dedicated file.

Assigned to myself.

sciunto:
xref #2444

PR #4502, opened 2020-03-10
Move register_translation from skimage.feature to skimage.registration

rfezzani:
## Description

This PR proposes to deprecate `skimage.feature.register_transtlation` in favor of `skimage.registration.register_transtlation`.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/feature/__init__.py
```diff
@@ -27,6 +28,17 @@
 from .haar import (haar_like_feature, haar_like_feature_coord,
                    draw_haar_like_feature)
 
+from ..registration.register_translation import (
+    register_translation as _register_translation)
```
jni: This makes my head spin. Can you rename the file something like `_fourier_space_translation` or somesuch? In general, we have tried to avoid files (modules) having the exact same name as functions because then it is ambiguous when you write `registration.register_translation` whether you mean the function or the module.

skimage/registration/__init__.py
```diff
@@ -1 +1,8 @@
 from ._optical_flow import optical_flow_tvl1
+from .register_translation import register_translation
```
jni: See above.

skimage/registration/__init__.py
```diff
@@ -1 +1,8 @@
...
+
+
+__all__ = [
+    'optical_flow_tvl1',
+    'register_translation'
+    ]
```
jni: We need to think about the API here. `registration.register_translation` no longer makes sense, it would be kind of like `filters.gaussian_filter` rather than `filters.gaussian`.

Some alternatives: `registration.fourier_cross_correlation`, `registration.translation`, `registration.maximum_cross_correlation`, `registration.phase_shift`...

skimage/feature/__init__.py
jni: (Note: I realise that the original file doesn't follow this convention. We haven't been great at enforcing this policy but I think it's worth enforcing going forward.)

skimage/registration/__init__.py
rfezzani: I personally vote for `phase_shift`

jni: @scikit-image/core anyone else want to weigh in here?

grlee77: Perhaps `phase_correlation`? The original publication describing it is apparently:
```
Kuglin, C., Hines, D., 1975. The phase correlation image alignment method. In: Proc.Internat. Conf. on Cybernetics and Society, IEEE, pp. 163–165
```
and that is also the title of it's [Wikipedia article](https://en.wikipedia.org/wiki/Phase_correlation).

OpenCV has [phaseCorrelate](https://docs.opencv.org/4.2.0/d7/df3/group__imgproc__motion.html#ga552420a2ace9ef3fb053cd630fdb4952) to calculate the shift.

sciunto: There is also the function match_template that would be good to move and rename as well. For this one, it uses a zero normalized cross correlation (https://en.wikipedia.org/wiki/Cross-correlation#Terminology_in_image_processing)

I suggest to think for these two functions at ones, and I would prefer phase_cross_correlation and zero_normalized_cross_correlation.

rfezzani: @sciunto, even if `template_matching` uses zncc internally, I don't think that the function name should be modified: template matching operation is well identified. Concerning `phase_cross_correlation` suggestion, why not, but if `phase_correlation` is well accepted in the community, why not keep it?

rfezzani: Shouldn't the reference provided by @grlee77 be added to the documentation?

sciunto: template_matching can be used to do other things than matching a template, for instance, you can use it to find shifts between images. Actually, the day I wanted to perform zncc, it took me some time to figure out that template_matching was actually what I wanted to use. That's why I advocate for changing this name.

rfezzani: OK @sciunto, that's an interesting point!
The `*_cross_correlation` functions must then coherent behavior. What if the `*_cross_correlation` functions return a `response` images as `match_template` does and make the `register_translation` return a unique `shift` based on any cross correlation method?
The [`_compute_error`](https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/register_translation.py#L90) function may then be exposed and the [`return_error`](https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/register_translation.py#L134) argument becomes obsolete!

sciunto: That sounds like a good idea. Xref: #3494

rfezzani: Do we all agree with `phase_cross_correlation` or do we choose `phase_correlation` instead?

@sciunto, I propose to address #3494 in an other PR ;-)

sciunto: yes, sounds good.

rfezzani:
I renamed `register_translation` to `phase_cross_correlation`, according to our previous discussion, any objection?

doc/examples/registration/plot_register_translation.py
```diff
@@ -1,14 +1,13 @@
-"""
-=====================================
+"""=====================================
```
jni, line 1: revert this change?

doc/examples/registration/plot_register_rotation.py
```diff
@@ -83,7 +84,8 @@
 plt.show()
 
 # setting `upsample_factor` can increase precision
-tparams = register_translation(image_polar, rescaled_polar, upsample_factor=20)
+tparams = phase_cross_correlation(image_polar, rescaled_polar,
+                                  upsample_factor=20)
 shifts, error, phasediff = tparams
```
jni, line 89:
```suggestion
shifts, error, phasediff = phase_cross_correlation(
    image_polar, rescaled_polar, upsample_factor=20
)
```

jni: I might also do, `from skimage import registration` ... `registration.phase_cross_correlation(...)`

skimage/feature/__init__.py
```diff
@@ -28,6 +29,16 @@
                    draw_haar_like_feature)
 
 
+@deprecated(alt_func='skimage.registration.register_translation',
+            removed_version='0.19')
+def register_translation(src_image, target_image, upsample_factor=1,
+                         space="real", return_error=True):
+    from ..registration._phase_cross_correlation import (
+        phase_cross_correlation as func)
```
jni, line 37: Why as func???

sciunto: Because func is short, func is universal, func is the solution to any issue, because the secret goal of this library is to provide a unique function func, that will perform a unified image processing.

rfezzani: > Because func is short,

That's the main reason! it eases the formatting in the next line.
Since it is a local variable and that this code bloc will be removed in v0.19, I supposed that it is fine...

rfezzani:
:tada: The CI is finally all green!

jni:
@scikit-image/core anyone wanna push the green button? A reminder to squash the commits. =)

sciunto:
I'm on it.

skimage/feature/tests/test_masked_register_translation.py
```diff
@@ -230,14 +232,13 @@ def test_cross_correlate_masked_autocorrelation_trivial_masks():
...
-    arr2 = camera()
 
     # Random masks with 75% of pixels being valid
     m1 = np.random.choice([True, False], arr1.shape, p=[3 / 4, 1 / 4])
-    m2 = np.random.choice([True, False], arr2.shape, p=[3 / 4, 1 / 4])
+    m2 = np.random.choice([True, False], arr1.shape, p=[3 / 4, 1 / 4])
 
-    xcorr = cross_correlate_masked(arr1, arr1, m1, m1, axes=(0, 1),
-                 mode='same', overlap_ratio=0).real
+    xcorr = cross_correlate_masked(arr1, arr1, m1, m2, axes=(0, 1),
```
sciunto, line 240: Whaoo, that was a bug.

skimage/registration/_phase_cross_correlation.py
```diff
@@ -118,31 +117,33 @@ def register_translation(src_image, target_image, upsample_factor=1,
...
         Reference image.
-    target_image : array
-        Image to register.  Must be same dimensionality as ``src_image``.
+    moving_image : array
+        Image to register. Must be same dimensionality as
+        ``reference_image``.
     upsample_factor : int, optional
         Upsampling factor. Images will be registered to within
-        ``1 / upsample_factor`` of a pixel. For example
+        ``1 /upsample_factor`` of a pixel. For example
```
sciunto, line 127: Is this change intended?

sciunto, line 127:
```suggestion
        ``1 / upsample_factor`` of a pixel. For example
```

rfezzani: Nop, sorry :pray:

sciunto:
Perfect! Thank you @rfezzani

PR #4503, opened 2020-03-10
Move masked_register_translation from feature to registration

rfezzani:
## Description

This PR proposes to deprecate `skimage.feature.masked_register_translation` in favor of `skimage.registration.masked_register_translation`.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


jni:
@rfezzani do you think we could revisit this API as simply passing `src_mask` and `dst_mask` to `register_translation`? These could be keyword only arguments. And we should also take this opportunity to rename the images `reference_image` and `moving_image`. What do you think?

rfezzani:
@jni, I implemented your proposed API changes.

doc/examples/registration/plot_masked_register_translation.py
```diff
@@ -1,27 +1,30 @@
-"""
-===================================
+"""===================================
```
jni:
```suggestion
"""
===================================
```

skimage/feature/__init__.py
```diff
@@ -27,6 +28,17 @@
 from .haar import (haar_like_feature, haar_like_feature_coord,
                    draw_haar_like_feature)
 
+from ..registration.masked_register_translation import (
+    masked_register_translation as _masked_register_translation)
```
jni: @rfezzani as in the other PR, the file should probably be renamed.

skimage/registration/masked_register_translation.py
```diff
@@ -15,31 +15,26 @@
...
-        target_image,
-        src_mask,
-        target_mask=None,
-        overlap_ratio=3 / 10):
-    """
-    Masked image translation registration by masked normalized
+def masked_register_translation(reference_image, moving_image, reference_mask,
+                                moving_mask=None, overlap_ratio=0.3):
+    """Masked image translation registration by masked normalized
     cross-correlation.
```
jni:
```suggestion
def masked_register_translation(reference_image, moving_image, *, reference_mask,
                                moving_mask=None, overlap_ratio=0.3):
    """Image translation registration by masked normalized cross-correlation.
```

skimage/registration/masked_register_translation.py
```diff
@@ -62,33 +58,37 @@ def masked_register_translation(
     .. [2] D. Padfield. "Masked FFT registration". In Proc. Computer Vision and
            Pattern Recognition, pp. 2918-2925 (2010).
            :DOI:`10.1109/CVPR.2010.5540032`
+
     """
-    if target_mask is None:
-        target_mask = np.array(src_mask, dtype=np.bool, copy=True)
+    if moving_mask is None:
+        moving_mask = np.array(reference_mask, dtype=np.bool, copy=True)
```
jni:
```suggestion
        moving_mask = reference_mask.astype(np.bool, copy=True)

```

jni: Come to think of it, is the mask modified? Maybe the copy is unnecessary here?

skimage/registration/masked_register_translation.py
rfezzani: Should `reference_mask` be a required named argument?

skimage/registration/masked_register_translation.py
rfezzani: I think it is not...

skimage/registration/masked_register_translation.py
rfezzani: @jni, `reference_mask` is now a required named argument.

rfezzani:
What about merging `register_translation` and `masked_register_translation` in the unique `registration.phase_cross_correlaion` function once #4502 is merged?

rfezzani:
> @rfezzani do you think we could revisit this API as simply passing src_mask and dst_mask to register_translation? These could be keyword only arguments. And we should also take this opportunity to rename the images reference_image and moving_image. What do you think?

@jni, I am just realizing that you already suggested  what I am proposing in my previous post :slightly_smiling_face: So yes, I am +1

jni:
@scikit-image/core I think this is ready to merge! The Travis CI failure was unrelated.

skimage/registration/_masked_register_translation.py
```diff
@@ -15,31 +15,27 @@
...
 
-def masked_register_translation(
-        src_image,
-        target_image,
-        src_mask,
-        target_mask=None,
-        overlap_ratio=3 / 10):
-    """
-    Masked image translation registration by masked normalized
+def masked_register_translation(reference_image, moving_image,
```
emmanuelle, line 18: Do we keep this function in the public API? If not it should be renamed `_masked_register_translation`.

rfezzani: Since it is in the `registration._masked_register_translation` module and it is not exposed in `registration.__init__.py`, I don't consider that it is in the public API. Is it?

emmanuelle: No it's not (I was checking that I was not missing something). Then your stylistic convention is to prepend non-exposed functions with an underscore, so `_masked_register_translation`.

rfezzani: @emmanuelle, I renamed `masked_register_translation` to `_masked_phase_cross_correlation` for consistency with the new name.

rfezzani: I see in `skimage.feature._canny.py` the `smooth_with_function_and_mask` that suffers from the same drawback =p.
But this is fortunately fixed in #4342, any review is welcome ;)

jni: There’s actually lots of internal modules where the functions don’t have a leading underscore... I don’t think we have adhered to this policy very strictly at all.

emmanuelle:
Thank you @rfezzani, merging! I love PRs which reduce the public API ;-).

rfezzani:
Thank you @emmanuelle!

hmaarrfk:
I think we need to move the pooch reference files for the datasets as well.

Issue #4500, opened 2020-03-10
register_translation order of operands matters


carterbox:
## Description

I expected that `register_translation(A, B)` to equal `-register_translation(B, A)`, but this is not always the case. I believe this is because "the first" maximum of the cross-correlation function is returned (`np.argmax`).

https://github.com/scikit-image/scikit-image/blob/88af67f1d2222dac21edd4548bfa50bdf76d0b4f/skimage/feature/register_translation.py#L210-L211

It would be preferable if the smallest shift was returned instead of whichever one happens to correspond to the correlation maximum with the smallest index. This will probably add some computational cost, but it would solve the following problems:

1. As noted in the opening paragraph and in the example below, switchin which image is the source and which is the target changes the result, and this doesn't meet my expectations.

2. When there are multiple shifts with the same correlation, not sorting shifts by shift distance could cause the images to drift away from a good initial guess near the center of the images, towards a solution where there is less evidence for correlation because the images overlap less.

I willing to work on this and open a pull if my arguments make sense.

## Way to reproduce
```python
# Define two arrays to be registered which have
# two possible alignments with the same error.
N = 32
A = np.zeros((N, N))
A[N//2, N//2] = 1
B = np.zeros((N, N))
B[N//2, N//2 - 1] = 1
B[N//2, N//2 + 3] = 1

# Draw the arrays
plt.figure()
plt.subplot(1, 2, 1)
plt.title('A')
plt.imshow(A)
plt.subplot(1, 2, 2)
plt.title('B')
plt.imshow(B)

print(register_translation(src_image=A, target_image=B))
print(register_translation(src_image=B, target_image=A))
# (array([0., 1.]), 0.7071067811865474, -2.0816681711721685e-17)
# (array([0., 3.]), 0.7071067811865474, -5.551115123125783e-17)

```

## Version information
```python
3.7.6 | packaged by conda-forge | (default, Jan  7 2020, 22:33:48) 
[GCC 7.3.0]
Linux-3.10.0-1062.9.1.el7.x86_64-x86_64-with-redhat-7.7-Maipo
scikit-image version: 0.16.2
numpy version: 1.18.1

```

jni:
Hi @carterbox! I'm totally on board with your proposal, provided of course that the errors are identical, which is the case here. That still leaves the question of what to do in the case of equidistant shifts, but, let's tackle one problem at a time. =)

hmaarrfk:
[...]

PR #4508, opened 2020-03-12
API: Add cross_correlation regularization to register_translation

carterbox:
## Description
`register_translation(A, B) != -register_translation(B, A)`, but it should be!

Shifts are chosen by using the "the first" maximum of the cross_correlation function, but this means that the shifts are always biased towards one of the corners of the image. Instead when multiple
shifts have the same cross_correlation, we should choose the smallest shift.

We can favor smaller shifts by adding a regularizer to the cross_correlation which favors higher overlap between the two images.

Closes scikit-image/scikit-image#4500

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -128,7 +128,7 @@ def _area_overlap(A):
         if dim == 0:
             w = _triangle(shape)
         else:
-            w = w[..., None] @ _triangle(shape)[None, ]
+            w = w[..., np.newaxis] @ _triangle(shape)[np.newaxis, ...]
     return w
 
 
--- a/skimage/feature/register_translation.py
+++ b/skimage/feature/register_translation.py
@@ -111,7 +111,8 @@ def _triangle(N):
     return 1 - abs(x - 0.5)
 
 def _area_overlap(A):
-    """Return overlapping area of A with itself.
+    """
+    Return overlapping area of A with itself.
 
     Create overlap arrays for higher dimensions using matrix multiplication.
 
@@ -132,7 +133,7 @@ def _area_overlap(A):
 
 
 def register_translation(src_image, target_image, upsample_factor=1,
-                         space="real", return_error=True):
+                         space="real", return_error=True, reg_weight=1e-12):
     """
     Efficient subpixel image translation registration by cross-correlation.
 
@@ -141,6 +142,10 @@ def register_translation(src_image, target_image, upsample_factor=1,
     It obtains an initial estimate of the cross-correlation peak by an FFT and
     then refines the shift estimation by upsampling the DFT only in a small
     neighborhood of that estimate by means of a matrix-multiply DFT.
+
+    When there are multiple cross-correlation peaks, ties are broken by a
+    regularizer which favors smaller shifts. This regularization may be
+    disabled by setting ``reg_weight`` to zero.
 
     Parameters
     ----------
@@ -152,14 +157,19 @@ def register_translation(src_image, target_image, upsample_factor=1,
         Upsampling factor. Images will be registered to within
         ``1 / upsample_factor`` of a pixel. For example
         ``upsample_factor == 20`` means the images will be registered
-        within 1/20th of a pixel.  Default is 1 (no upsampling)
+        within 1/20th of a pixel.  Default is 1 (no upsampling).
     space : string, one of "real" or "fourier", optional
         Defines how the algorithm interprets input data.  "real" means data
         will be FFT'd to compute the correlation, while "fourier" data will
         bypass FFT of input data.  Case insensitive.
     return_error : bool, optional
-        Returns error and phase difference if on,
-        otherwise only shifts are returned
+        Returns error and phase difference if when True, otherwise only shifts
+        are returned.
+    reg_weight : float, optional
+        Determines the strength of shift regularization.
+        .. versionadded:: 0.17
+           ``reg_weight`` was introduced to break ties between
+           cross-correlation peaks.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


skimage/feature/register_translation.py
```diff
@@ -105,8 +105,36 @@ def _compute_error(cross_correlation_max, src_amp, target_amp):
...
+    >>> _area_overlap(np.empty((3, 5)))
+    array([[0.4       , 0.53333333, 0.66666667, 0.53333333, 0.4       ],
+           [0.6       , 0.8       , 1.        , 0.8       , 0.6       ],
+           [0.4       , 0.53333333, 0.66666667, 0.53333333, 0.4       ]])
+    """
+    for dim, shape in enumerate(A.shape):
+        if dim == 0:
+            w = _triangle(shape)
+        else:
+            w = w[..., None] @ _triangle(shape)[None, ]
```
jni:
```suggestion
            w = w[..., np.newaxis] @ _triangle(shape)[np.newaxis, ...]
```

skimage/feature/register_translation.py
```diff
@@ -126,14 +158,21 @@ def register_translation(src_image, target_image, upsample_factor=1,
...
-        within 1/20th of a pixel.  Default is 1 (no upsampling)
+        within 1/20th of a pixel.  Default is 1 (no upsampling).
     space : string, one of "real" or "fourier", optional
         Defines how the algorithm interprets input data.  "real" means data
         will be FFT'd to compute the correlation, while "fourier" data will
         bypass FFT of input data.  Case insensitive.
     return_error : bool, optional
-        Returns error and phase difference if on,
-        otherwise only shifts are returned
+        Returns error and phase difference if when True, otherwise only shifts
```
jni:
```suggestion
        Returns error and phase difference if True, otherwise only shifts
```

skimage/feature/register_translation.py
```diff
@@ -198,9 +245,9 @@ def register_translation(src_image, target_image, upsample_factor=1,
         # Center of output array at dftshift + 1
         dftshift = np.fix(upsampled_region_size / 2.0)
         upsample_factor = np.array(upsample_factor, dtype=np.float64)
-        normalization = (src_freq.size * upsample_factor ** 2)
+        normalization = (src_freq.size * upsample_factor**2)
         # Matrix multiply DFT around the current shift estimate
-        sample_region_offset = dftshift - shifts*upsample_factor
+        sample_region_offset = dftshift - shifts * upsample_factor
```
jni: Could you please undo all these operator spacing changes? Although they are probably for the better, they make it harder to trace the authorship of particular lines, for little gain in this case.

jni:
Also, sorry that this took so long! My open review tab unfortunately got buried for a while. 😬

carterbox:
@jni, Do you want a doctest or test functions in tests/?

jni:
@carterbox I think this would be difficult to make concise enough for a doctest, so I would go for a full test in `tests/`. But, if you think you can make it concise enough, feel free to make it a doctest!

carterbox:
@jni, the doctest I created fails on Travis CI, but it passes on my computer:

```
151     Define two arrays to be registered which have two possible alignments
152     with the same error.
153     >>> A = np.zeros((8, 8))
154     >>> A[4, 4] = 1
155     >>> B = np.zeros((8, 8))
156     >>> B[3, 3], B[5, 7] = 1, 1
157     >>> register_translation(src_image=A, target_image=B)
Expected:
    (array([1., 1.]), 0.7071067811865476, 0.0)
Got:
    (array([1., 1.]), 0.7071067811865476, 1.6653345369377348e-16)
```
Should I get rid of the doctest? I guess this problem exists because we are using a python native test to compare the values of NumPy arrays.

jni:
Damn! LOL. Yeah, probably removing the doctest is the best option. Alternatively, you could just grab the first value out of the function, which would then match.

carterbox:
@jni CI is failing, but it's not my fault! There something wrong with pillow?

jni:
Yep yep, internet is down with Pillow 7.0. LOL Though I see they just quickly pushed another release — let me see if that fixes it.

sciunto:
xref: #4547 4548

skimage/feature/register_translation.py
```diff
@@ -105,8 +105,36 @@ def _compute_error(cross_correlation_max, src_amp, target_amp):
...
+        if dim == 0:
+            w = _triangle(shape)
+        else:
+            w = w[..., np.newaxis] @ _triangle(shape)[np.newaxis, ...]
+    return w
+
+
 def register_translation(src_image, target_image, upsample_factor=1,
-                         space="real", return_error=True):
+                         space="real", return_error=True, reg_weight=1e-12):
```
sciunto, line 137: I may suggest to set reg_weight to zero by default though...

carterbox: On second thought, I disagree because I consider this pull request to be fixing bug not adding a feature (#4500). Please call a vote of the maintainers if necessary.

sciunto: Actually, I didn't read the issue as a bug, more as a potential improvement :)
How to you justify the default value? Does this prevent the "bug" at 100% with this default value? I guess not, so leaving to the choice of the user with a default on disable sounds a good compromise to me.

carterbox: Is your objection because you don't consider this a bug fix or because you need proof that the default value is 100% effective?

In the former case, we disagree (which is fine); edits from maintainers is allowed. In the later case, I will need conduct a study of whether `1e-12` is the most effective value or if the regularizer should be changed.

Issue #4442, opened 2020-02-11
Deprecate 32 and 64 bits integer types support


rfezzani:
## Description

I experienced some confusion with behavior definition with images with `int` dtype https://github.com/scikit-image/scikit-image/pull/4424#discussion_r372821261.

@jni pointed out that "`int` images are a big problem for us in general" (see #3009 for example).

I wonder if we can not simply deprecate the support of `int` images with higher depth than 16-bit:

- Does `int` images with higher depth than 16-bit represent a real use case?
- If any use case exist, can't we advise users to cast the images to float and use the `preserve_range` argument?

hmaarrfk:
Fixed point math is quite useful. I wouldn't throw it away because of convenience.

What does deprecating a datatype mean in python? Do we have to check at every entry point that we don't have int32? Seems tiresome.

What about checking for int24? Or other strange datatypes.

rfezzani:
I am not saying that fixed point is not useful :-) I am just saying that in the context of image processing, I don't see much applications with more then 16  bits int, but I am pretty sure that I am mistaken here :smile: 

Yes, this is made by checking `np.issubdtype(img.dtype, np.integer) and img.dtype.itemsize < 3`, mainly in the `skimage.util.dtype` module... It will hopefully impact a great majority of the package.

We may also apply this check in some specific modules/ functions where it makes sens, I think about `skimage.color`, but there is probably more...

We may then specify in the user guide  data types [section](https://scikit-image.org/docs/stable/user_guide/data_types.html?highlight=data%20type) that we don't recommend the use of 32 and 64 bits integers.

rfezzani:
I see in the user guide data types [section](https://scikit-image.org/docs/stable/user_guide/data_types.html?highlight=data%20type) that _With a few exceptions, 64-bit (u)int images are not supported._ already :smile:

jni:
I think this discussion needs to take place in the context of #3009. I generally advocate for deprecating implicit rescaling in the library altogether. In that case, it doesn't make sense to deprecate those dtypes.

It *does* make sense to deprecate rescaling of int32 range to [0, 1]. I've never encountered a case where that was a good idea.

PR #4510, opened 2020-03-13
Add cast to minimum dtype before imsave

### LINKED ISSUE #4509: saving a np.uint64 to png removes (low contrast) image content
## Description

I was trying to save an image to a png file. It appeared to me that `imsave` removed the content of this image. I am by means no expert with libpng, but that data is lost completely appears like a bug to me. Note for the following, when starting from `np.float64`, the problem does not appear. 

I had to dig down this route as I used some code that did `np.sum([np.uint8-array, np.uint8-array])` which gave an `np.ndarray` with `np.uint64` in the first place.

## Way to reproduce

```python
import numpy as np
from skimage.io import imsave, imread
img = np.zeros((32,32),dtype=np.uint64
img[8:24,8:24] = 1
imsave("test.png",img)
#this issue the usual warning 
#__main__:1: UserWarning: test.png is a low contrast image
#Lossy conversion from uint64 to uint8. Losing 56 bits of resolution. Convert image to uint8 prior to #saving to suppress this warning.
rimg = imread('test.png')
print(rimg.sum()) #Yikes, this prints 0
```

expected

## Version information
```python
[GCC 7.3.0]
Linux-5.5.8-200.fc31.x86_64-x86_64-with-fedora-31-Thirty_One
scikit-image version: 0.16.2


numpy version: 1.17.2
```

rfezzani:
## Description

Fixes #4509.

This PR proposes to cast the image  to the dtype with the smallest size and smallest scalar kind which can hold its value before saving. This may prevent the `#Lossy conversion from uint64 to uint8. Losing 56 bits of resolution`  `imageio` warning to be raised (See #4509).

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/testing.py
+++ b/skimage/_shared/testing.py
@@ -141,7 +141,7 @@ def color_check(plugin, fmt='png'):
         testing.assert_allclose(r4, img4)
     else:
         r4 = roundtrip(img4, plugin, fmt)
-        testing.assert_allclose(r4, img_as_ubyte(img4))
+        testing.assert_allclose(r4, img_as_ubyte(img4.astype(np.uint16)))
 
     img5 = img_as_uint(img)
     r5 = roundtrip(img5, plugin, fmt)
@@ -176,7 +176,7 @@ def mono_check(plugin, fmt='png'):
         testing.assert_allclose(r4, img4)
     else:
         r4 = roundtrip(img4, plugin, fmt)
-        testing.assert_allclose(r4, img_as_uint(img4))
+        testing.assert_allclose(r4, img4)
 
     img5 = img_as_uint(img)
     r5 = roundtrip(img5, plugin, fmt)
--- a/skimage/io/_io.py
+++ b/skimage/io/_io.py
@@ -130,10 +130,14 @@ def imsave(fname, arr, plugin=None, check_contrast=True, **plugin_args):
         warn('%s is a low contrast image' % fname)
     if arr.dtype == bool:
         warn('%s is a boolean image: setting True to 1 and False to 0' % fname)
-    min_dtype = np.result_type(np.min_scalar_type(arr.max()),
-                               np.min_scalar_type(arr.min()))
-    return call_plugin('imsave', fname, arr.astype(min_dtype),
-                       plugin=plugin, **plugin_args)
+    arr = np.asarray(arr)
+    min_dtype = arr.dtype
+    if np.issubdtype(min_dtype, np.integer):
+        min_dtype = np.result_type(np.min_scalar_type(arr.max()),
+                                   np.min_scalar_type(arr.min()))
+    print("------------", arr.dtype, min_dtype, arr.min(), arr.max())
+    return call_plugin('imsave', fname, arr.astype(min_dtype), plugin=plugin,
+                       **plugin_args)
 
 
 def imshow(arr, plugin=None, **plugin_args):
--- a/skimage/io/_io.py
+++ b/skimage/io/_io.py
@@ -130,7 +130,10 @@ def imsave(fname, arr, plugin=None, check_contrast=True, **plugin_args):
         warn('%s is a low contrast image' % fname)
     if arr.dtype == bool:
         warn('%s is a boolean image: setting True to 1 and False to 0' % fname)
-    return call_plugin('imsave', fname, arr, plugin=plugin, **plugin_args)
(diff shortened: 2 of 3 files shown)


hmaarrfk:
I'm really against this kind of magic casting.

Personally, as many people using scikit-image are interested in scientific image analysis, 
I would almost be more happy if the saving functions just returned an error instead of downcasting.

instead of simply saying "no", I would really like to see this discussion in the broader discussion of "data types in scikit-image".

This kind of "feature" is tremendously hard to roll back, and only serves the delay teaching users about data types. An inevitable task for anybody working with images.

emmanuelle:
I quite err on the "no magics" side, and also think we should make this decision within the broader context of 1.0 (no range change behind the scene for example). That said, a utility function downcasting to the smallest dtype compatible with the data range would be very useful for users.

emmanuelle:
ah ok there is already `np.min_scalar_type`, didn't know about that one :-). Even if it's a 1.5 liner, the utility function might still be useful? Or a documentation example?

jni:
I favor the documentation example, perhaps in that same data types page. I 100% agree with @hmaarrfk that we should not automagically downcast.

rfezzani:
I am sorry, but I don't see any magic in here... The proposed option doesn't involve any data range modification, it conversely prevent data range modification when integer data are saved using `imsave` and then reloaded with `imread` (please see our test suite [here](https://github.com/scikit-image/scikit-image/blob/master/skimage/_shared/testing.py#L179)).

jni:
@rfezzani it's absolutely magic, especially until we implement our new data range policy (ie skimage 1.0). A dim uint16 picture with values only in 0-255 should remain a dim uint16 picture, not a full-range uint8 picture, unless the user explicitly says otherwise. (imho!)

rfezzani:
Well @jni, I understand your point, and I agree, but I still think that it's not magic: there is no trick here, it's an implementation choice ;-).
All this discussion (#4509 included), is again linked to #3009 and #4442...
Closing...

jni:
@rfezzani True, nothing in Python is actually magic. =) My off-the-cuff definition of magic: automatic behaviour that attempts to *guess* the desires of the user. Magic is particularly bad if the user can't turn it off, as would be the case here.

psteinb:
@emmanuelle @rfezzani for the time being, I am most in favor of adding an example for documentation. I am happy to contribute this. As I am new to the project, I'd appreciate some pointers on where to add this example. I'll submit a PR then.

jni:
Hi @psteinb and thanks for offering to help! I see two possible places:

(1) in our [examples gallery](https://scikit-image.org/docs/dev/auto_examples/). You could add an example under "data" titled "input and output", in which you can demonstrate this issue. To do this, add a new `.py` file to the [`doc/examples/data/`](https://github.com/scikit-image/scikit-image/tree/master/doc/examples/data) directory. You can look at the other files there for syntax examples, but for reference, they use [sphinx-gallery](https://sphinx-gallery.github.io/stable/syntax.html) syntax. OR,

(2) as a section in our [image data types and what they mean](https://scikit-image.org/docs/dev/user_guide/data_types.html) page. To do this, you would edit the file [`doc/source/user_guide/data_types.rst`](https://github.com/scikit-image/scikit-image/blob/master/doc/source/user_guide/data_types.rst). This uses the sphinx [ReStructured Text syntax](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html).

Or, we could add both. =) Since data type ranges are a major source of confusion for new users, it wouldn't hurt to have information about this in multiple places.

We have more general contributing guidelines [here](https://scikit-image.org/docs/dev/contribute.html)

Thanks again!

psteinb:
see https://github.com/scikit-image/scikit-image/pull/4517

alexdesiqueira:
>  I understand your point, and I agree, but I still think that it's not magic: there is no trick here, it's an implementation choice ;-)
> (...)
> My off-the-cuff definition of magic: automatic behaviour that attempts to guess the desires of the user. 

I know this is closed and all, but I think such a change on our "paradigms" would require a [SKIP](https://scikit-image.org/docs/dev/skips/0-skip-process.html) and a bit of discussion.

rfezzani:
@alexdesiqueira, please open such a SKIP ;-)

alexdesiqueira:
@rfezzani actually I agree with the current behavior we have: not making that kind of changes "under the hood". I think we should mess with the user data as minimum as possible :smile:

PR #4512, opened 2020-03-15
Fix bool array save with imageio plugin

### LINKED ISSUE #4511: ValueError raised when bool array saving with imageio plugin
## Description

A `ValueError` is raised when attempting to save a bool array using the 'imageio' plugin

## Way to reproduce

```python
>>> from skimage import io
>>> a = np.ones((5, 5), bool)
>>> io.imsave('/tmp/test.png', a, 'imageio')
__main__:1: UserWarning: /tmp/test.png is a low contrast image
__main__:1: UserWarning: /tmp/test.png is a boolean image: setting True to 1 and False to 0
  File ".conda/envs/skimage-debug/lib/python3.6/site-packages/skimage/io/_io.py", line 131, in imsave
    return call_plugin('imsave', fname, arr, plugin=plugin, **plugin_args)
  File "conda/envs/skimage-debug/lib/python3.6/site-packages/skimage/io/manage_plugins.py", line 210, in call_plugin
    return func(*args, **kwargs)
  File "conda/envs/skimage-debug/lib/python3.6/site-packages/imageio/core/functions.py", line 293, in imwrite
    raise ValueError("Image is not numeric, but {}.".format(imt.__name__))
ValueError: Image is not numeric, but ndarray.
```
## Version information

```python
3.6.10 |Anaconda, Inc.| (default, Jan  7 2020, 21:14:29) 
[GCC 7.3.0]
Linux-4.4.0-174-generic-x86_64-with-debian-stretch-sid
scikit-image version: 0.16.2
numpy version: 1.18.1
```

rfezzani:
## Description

Fixes #4511.

## For reviewers

  later.


jni:
@rfezzani I'm confused why so many lines of code are touched in this PR, when really the only thing it's proposing a `.astype(uint8)` call, right? What has changed in all the tests? I find those changes hard to follow...

rfezzani:
I agree, I initially wanted to use `testing.mono_check` and `testing.color_check` for testing the fix, and this introduced all these changes :cry:
I noticed that these test functions are only used for the `PIL` plugin... It is strange to have them defined in the `testing` module without being used for all the plugins...

hmaarrfk:
i really think this issue should be brought up with the backend.

if we want the user to have a good experience, we should either avoid the backends we know are problematic (by checking their version numbers), or raise an error if we can't.

Ultimately, imageio is a newer project than scikit-image and I think it is really valuable to focus on the "saving + long term compatibility" aspect of this problem in that project. 

pngs and bitmaps both support 1 bit per pixel formats, which can help really save on storage space if that is the user's goal.

Personally, i don't think bool images should be mapped to `0` and `1` `uint8`. This is almost never the intended consequence because want interpolation to work "as it does on a grayscale image`. Interpolating between two discrete values is impossible, and requires application specific knowledge.

In my own "preservation library" https://github.com/hmaarrfk/redpil/blob/master/redpil/bmp.py#L52 I decided to map bool to min and max, 0 and 255. This seems to be running against the current thoughts of scikit-image. Therefore, before adding adhoc support here, I would really want to know:
1. How do we want to support different dtypes.
2. Do we want to support bool
3. If true: How do we support bool?

For now, for a new user, I would much rather respond:
> Hey your code isn't working as you expect because bool is actually a pretty badly defined data type in terms of image processing, we suggest you don't use it.

rather than

> Hey, here is the subset of functions that do support bool type, you can try to use them, and maybe you will get the results you want.

hmaarrfk:
sorry for making this a "controversitial" PR, maybe lets keep the discussion in the issue, rather than thinking about your exact fix.

rfezzani:
Thank you @hmaarrfk for your detailed argumentation, I absolutely agree with all what you are saying!
Raising issues and PR makes the project evolve and controversy forces us to take decisions ;-)
As I already said, I would push for deprecating all the `skimage.io` module.

hmaarrfk:
> As I already said, I would push for deprecating all the skimage.io module.

I too, so lets just avoid any unrelated PRs? that has been my strategy at least ;)

rfezzani:
> I too, so lets just avoid any unrelated PRs? that has been my strategy at least ;)

Opening issues shows that there are problems and forces us to take decisions ;-)

JDWarner:
Just to chime in here, I agree with @hmaarrfk that a Boolean type being mapped to 0 and 1 for uint8 is jarring and *every time I have encountered this behavior it was irritating and against my expectations*.  This isssue is virtually never encountered until image saving, because NumPy type rules are lenient enough to do what the user wanted.

If the destination type supports 1-bit-per-pixel encoding that should be used.  In all other cases, IFF the conversion needs to happen, what I wanted was all values to be 0 or 255.  I end up doing this conversion myself via some motif like 

``` python
img = bool_img.astype(np.uint8) * 255  # Why did I need to do this
# Now I can save img
```

after realizing the apparently black images I just saved were using 0-1 out of 255 possible values.

sciunto:
I'm just +1 with @JDWarner (although I would replace 255 by the max of the considered dtype, I'm nitpicky...)

rfezzani:
I agree with you @JDWarner, it is usually surprising to obtain an almost all black image after saving a mask :-). I tried in this PR to stick to what the original warning message was claiming: ` warn('%s is a boolean image: setting True to 1 and False to 0' % fname)`.
One option would be to convert the image to `float32` instead of `uint8`

rfezzani:
>  I'm confused why so many lines of code are touched in this PR
@jni, I reverted the too much modifications introduced in this PR ;-)

rfezzani:
When I see the [CI failures](https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=2878&view=logs&j=75347757-894e-5c54-3c11-df095f4d729a&t=a5971a41-540c-5744-112c-796ac46524ed&l=5761), I remember now why I converted to `uint8` in the initial PR: it is the expected behavior with the 'pil' plugin..

sciunto:
I might be confused but I think @JDWarner said to convert to int, but true false values must be 0 255, and not 0 1.

rfezzani:
You are right @sciunto , converting to `float32` doesn't fully do the trick:
```python
>>> import numpy as np
>>> img = np.eye(3, dtype=bool)
>>> io.imsave(fname, img)
Lossy conversion from float32 to uint8. Range [0, 1]. Convert image to uint8 prior to saving to suppress this warning.
>>> img2 = io.imread(fname)
>>> img2
array([[255,   0,   0],
       [  0, 255,   0],
       [  0,   0, 255]], dtype=uint8)
```

rfezzani:
Thank you @sciunto :tada:

emmanuelle:
Thanks for the PR @rfezzani and thanks to the many reviewers. I agree that saving a bool array as a black and white array is what should be done (it is always possible to save with numpy format `npy` if the array is not to be visualized). Merging since the PR has already been approved.

PR #4515, opened 2020-03-17
example: register rotation and scaling with no shared center

mellertd:
## Description

This is basically following on #4231 

This example shows how rotation and scaling differences between two images can be determined by registering the log-polar-transformed FFT magnitudes of both images. This approach has many references in the literature, but I am not sure there is one standard approach, and there is a little bit of complexity:

1) images are band-pass filtered (I am not aware of a single best filter, so this would be a parameter that needs to adjusted)
2) images are windowed (not aware of single best window, so another potential parameter to be adjusted)
3) abs of FFTs are taken for both images and log-polar transformed
4) x-y translation is recovered by something like cross-correlation
5) translation parameters are converted back into rotation and scaling

When this works, it's quite fast, but I am not sure how it would compare with other methods. It can work even when the images don't necessarily have nice features to grab onto for feature-matching based methods.

I am not an expert in this domain, but have effectively used this method in real workflows.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


doc/examples/transform/plot_register_rotation.py
```diff
@@ -96,3 +96,117 @@
...
+shape = image_fs.shape
+radius = shape[0] // 8  # only take lower frequencies
+warped_image_fs = warp_polar(image_fs, radius=radius, output_shape=shape,
+                             scaling='log', order=0)
+warped_rts_fs = warp_polar(rts_fs, radius=radius, output_shape=shape,
+                           scaling='log', order=0)
+
+warped_image_fs = warped_image_fs[:shape[0] // 2, :]  # only use half of FFT
+warped_rts_fs = warped_rts_fs[:shape[0] // 2, :]
+tparams = register_translation(warped_image_fs, warped_rts_fs,
```
emmanuelle, line 180: Could you please use here `registration.phase_cross_correlation` since the name changed?

emmanuelle:
Hi @mellertd thank you for the great example and sorry we did not get back to you earlier. I believe the conflicts are due to the change of name `register_translation` to `phase_cross_correlation`. Also, could you please add one or two sentences about why this strategy works (probably for the whole tutorial not only the part you added)? If I understand well, one applies transformations to the images so that their differences amounts to a translation in a given space, since a translation is well detected by phase cross-correlation in Fourier space. So a rotation amounts to a translation along the polar axis, a scaling is a translation in the log of the radius, and a translation is just a phase shift in Fourier. If this is correct -- probably you can improve these explanations - I think it is worth adding them.

mellertd:
How does this look (assuming it passes CI checks)? A lot more verbose, but hopefully more clear.

emmanuelle:
It's great, thank you so much for the pedagogy effort!

doc/examples/registration/plot_register_rotation.py
```diff
@@ -1,22 +1,36 @@
...
-(``registration.phase_cross_correlation``)
-to recover rotation and scaling differences between two images that share a
-center point.
+==========================================================
+Using Polar and Log-Polar Transformations for Registration
+==========================================================
+
+Phase correlation (``registration.phase_cross_correlation``) is an efficient
+method for determining translation offset between pairs of similar images.
+However this approach relies on a near absense of rotation/scaling differences
```
jni, line 8:
```suggestion
However this approach relies on a near absence of rotation/scaling differences
```

emmanuelle:
Merging, thanks @mellertd !

mellertd:
argh, I found two more typos as I was fixing the first! Should I open a new small PR? @emmanuelle @jni

emmanuelle:
go ahead and for typos in docs PRs, just one :+1: is enough :-). By the way I'm very curious to know more about the use cases in which you have applied these techniques.

mellertd:
#4623 

I was registering widefield fluorescence images for a group. These were taken on an inverted scope in a petri dish that was treated, imaged, taken off the scope, treated, etc. for several cycles. So differences in rotation and translation were inevitable (scaling not such a problem). I was struggling a lot with feature matching, because the treatments involved some restaining and in general the images didn't have great features. This worked well and was pretty fast.

Happy to go into even more detail, but this is probably not the best forum.

stefanv:
Wow, I've been using log polar images for a long time to do registration, but I've never seen the FFT step before.  This is quite a fortuitous coming together of all the attributes of the Fourier transform!

jni:
@mellertd if you have sample images that could be shared and uploaded to somewhere like figshare, this could be added to our gallery, now that we have support for remote datasets with pooch. We don't have a good guide for how to use pooch (https://github.com/scikit-image/scikit-image/issues/4522), but you can see a small example here:

https://github.com/scikit-image/scikit-image/pull/4435/files#diff-1f4aa96c33f197d27ea7adb25ce03702R136

jni:
Needless to say, this would be so cool. =D

stefanv:
> Wow, I've been using log polar images for a long time to do registration, but I've never seen the FFT step before. This is quite a fortuitous coming together of all the attributes of the Fourier transform!

I keep thinking about this and I am, after all, not sure I understand that last step.  Would you mind explaining why it works?  With the FFT, you expect low frequency components to always reside near the center.  So, why would aligning spectra give you these attributes you seek?

mellertd:
@stefanv if you find thinking about the math more intuitive, you might check out the early work using a variant of this technique:

 B.S. Reddy and B.N. Chatterji. An FFT-based technique for translation, rotation and scale- invariant image registration. IEEE Trans. Image Processing, 5(8):1266–1271, 1996.

Also a decent explanation and many references in:

Tzimiropoulos, Georgios, and Tania Stathaki. "Robust FFT-based scale-invariant image registration." In 4th SEAS DTC Technical Conference. 2009.

Here is my crack at an explanation, as someone who is not so mathematically inclined:

If you have two images with largely the same content shifted by a modest translation, their fourier magnitudes will look nearly identical. If you rotate the image, you simply rotate the fourier magnitudes. So a polar transform of the fourier magnitude will turn the rotation into a translation. This just leaves the question of scaling (I think this is all clear to you, just trying to circumscribe the point in question for my own thinking).

The easiest way for me to wrap my head around the scaling issue is to think about the simplest case of a size wave of period p. The FFT of that wave would produce a peak at frequency f. If I double the period (2p), I halve the frequency (f/2). Of course doubling the period amounts to a scale change of 2x, which in log2 space would be a positive shift of one unit. Thus in frequency space, the scale change of 2 becomes a scale change of 1/2. Which is a negative shift of one unit in log2. You may notice a subtlety in the example where I am using `shift_scale = 1 / (np.exp(shiftc / klog))` to calculate scale in the first scaling example, and  `shift_scale = np.exp(shiftc / klog)` in the one in question. I could have also used `shift_scale = 1 / (np.exp(-shiftc / klog))`.

So as you increase the scale of your image, you decrease the scale of the frequency by an inverse amount. But this is still a scale change in the frequency domain and the math is rather straightforward.

The major issue that pops up is that, because you don't know the scale difference ahead of time, and you need to do some bandpass filtering of the images, you end up attenuating different features across the images prior to the registration. Note that when you look at the log-polar transformed magnitude spectra, the don't really look "alike". Yet if you look closely there are some common patterns in those images, and they do end up aligning pretty well by phase correlation.

Another issue is that very large changes in scale will dramatically alter the magnitude spectra (especially since a big change in scale will usually be accompanied by some cropping and loss of information content). The literature advises staying within 1.8-2x scale change. This is fine for most biological imaging applications (my domain).

Phew. I hope this was at least a little helpful! Not my area of expertise so it's a bit hard to explain

mellertd:
@jni I'll look into sharing those data. It's been a while...I'll also need to update the code. The first time around, I was using opencv, because skimage didn't have a polar warping function! :)

zpincus:
@mellertd Both the math and non-math explanation you have above are really useful. Thanks for that! I was wondering too :)

I wonder if iIt might be worthwhile adding some of that discussion to the example itself (or perhaps just linking to your comment above if it's hard to formalize enough for it to truly belong in the example?). Especially the practical notes about the range in which scale changes can be handled, and such.

stefanv:
Yes, that was actually very helpful!  I got the 1/f scaling but missed that the log transform takes care of that.  I agree with @zpincus that this would be a good addition to the example.  And, the practical ranges should probably go in the notes section of the docstring.

Now, I'm wondering what can be done to address those filtering issues.

mellertd:
I think the practical range only applies to the registration application...I am not sure which docstring the note would best fit in. Unless you mean to put the note in the gallery example itself.

Although I think the original issue was to make the whole thing a new function -- @jni suggested to start with a gallery example, and then several contributions later, here we are!

jni:
> I am not sure which docstring the note would best fit in. Unless you mean to put the note in the gallery example itself.

I think that's what they meant!

PR #4519, opened 2020-03-17
Fix exposure_adapthist return when clip_limit == 1

### LINKED ISSUE #4516: exposure.equalize_adapthist changes output datatype
* **[ORIGINAL TITLE] Setting clip-limit to 1 does not return original image.**

## Description
In the _clahe function, it says setting the clip limit to one will return the original image. That is not true because before it is passed through to that function, clip limit is multiplied by "nbins".

rfezzani:
## Description

Fixes #4516 

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -51,7 +51,7 @@ def equalize_adapthist(image, kernel_size=None,
     Returns
     -------
     out : (M, N[, C]) ndarray
-        Equalized image.
+        Equalized image with float64 dtype.
 
     See Also
     --------
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -70,6 +70,10 @@ def equalize_adapthist(image, kernel_size=None,
     .. [1] http://tog.acm.org/resources/GraphicsGems/
     .. [2] https://en.wikipedia.org/wiki/CLAHE#CLAHE
     """
+
+    if clip_limit == 1.0:
+        return image  # is OK, immediately returns original image.
+
     image = img_as_uint(image)
     image = rescale_intensity(image, out_range=(0, NR_OF_GRAY - 1))
 
@@ -82,7 +86,7 @@ def equalize_adapthist(image, kernel_size=None,
 
     kernel_size = [int(k) for k in kernel_size]
 
-    image = _clahe(image, kernel_size, clip_limit * nbins, nbins)
+    image = _clahe(image, kernel_size, clip_limit, nbins)
     image = img_as_float(image)
     return rescale_intensity(image)
 
@@ -112,12 +116,6 @@ def _clahe(image, kernel_size, clip_limit, nbins):
     minimum and maximum value as the input image. A clip limit smaller than 1
     results in standard (non-contrast limited) AHE.
     """
-
-    if clip_limit == 1.0:
-        return image  # is OK, immediately returns original image.
-
-    clip_limit /= nbins
-
     row_step = int(image.shape[0] // np.ceil(image.shape[0] / kernel_size[0]))
     col_step = int(image.shape[1] // np.ceil(image.shape[1] / kernel_size[1]))
 
--- a/skimage/exposure/tests/test_exposure.py
+++ b/skimage/exposure/tests/test_exposure.py
@@ -374,6 +374,12 @@ def test_adapthist_borders():
     assert norm_brightness_err(adapted[-width, :], img[-width, :]) > 1e-3
 
 
+def test_adapthist_clip_limit():
+    img = data.moon()
+    img_clahe = exposure.equalize_adapthist(img, clip_limit=1)
+    assert_array_equal(img, img_clahe)
+
+
 def peak_snr(img1, img2):
     """Peak signal to noise ratio of two images
 
--- a/skimage/exposure/tests/test_exposure.py
+++ b/skimage/exposure/tests/test_exposure.py
@@ -375,9 +375,16 @@ def test_adapthist_borders():
 
 
 def test_adapthist_clip_limit():
-    img = data.moon()
-    img_clahe = exposure.equalize_adapthist(img, clip_limit=1)
-    assert_array_equal(img, img_clahe)
+    img_u = data.moon()
+    img_f = util.img_as_float(img_u)
+
+    # uint8 input
+    img_clahe = exposure.equalize_adapthist(img_u, clip_limit=1)
+    assert_array_equal(img_f, img_clahe)
+
+    # float64 input
+    img_clahe = exposure.equalize_adapthist(img_f, clip_limit=1)
+    assert_array_equal(img_f, img_clahe)
 
 
 def peak_snr(img1, img2):


alexdesiqueira:
Hey @rfezzani,
thanks for that, but I think this is not the optimal way to proceed here. IMHO, we should have a solution that returns the same data as the input datatype, aligned to [our mission](https://scikit-image.org/docs/dev/values.html):

> **caring for users’ data.** We (...) don’t modify input arrays unless explicitly directed to do so.

jni:
@alexdesiqueira but that ship has sailed when it comes to the dtype. The full fix needs to wait for 1.0 I think.

alexdesiqueira:
@jni got it. So we're good to go with @rfezzani solution, then?

jni:
I don't actually understand why the clip_limit was multiplied by nbins before, or why removing it is the right solution... But otherwise, yes!

rfezzani:
I agree @jni, the multiplication of `clip_limit` by `nbins` is strange... I introduced the division by the same `nbins` [here](https://github.com/scikit-image/scikit-image/commit/4729a8e10457e6becf7c431e9ce204161721144c#diff-ab83bf08dd2a5277d4375e9746f0778f).
Following @alexdesiqueira [experiment](https://github.com/scikit-image/scikit-image/issues/4516#issuecomment-600314170), `_clahe` doesn't exit when checking for [`clip_limit == 1`](https://github.com/scikit-image/scikit-image/blob/master/skimage/exposure/_adapthist.py#L116) since it is equal to 256 (because of its multiplication by `nbins`). But the result of `exposure_adapthist` converted back to `uint8` is effectively equal to the input image... This shows that we effectively need to check for `clip_limit == 1` and not `(clip_limit * nbins) == 1`.

alexdesiqueira:
> Following @alexdesiqueira experiment, _clahe doesn't exit when checking for clip_limit == 1 since it is equal to 256 (because of its multiplication by nbins).

@rfezzani actually `_clahe` _does_ exit, [in this return](https://github.com/scikit-image/scikit-image/blob/786adad0da9b399c9c4689bf43b44b95011c13ac/skimage/exposure/_adapthist.py#L117):
```
    if clip_limit == 1.0:
        return image  # is OK, immediately returns original image.
```
With this code,
* when `clip_limit == 1.0` and we enter an `uint8` image, we will receive back an `uint8` image; 
* when `clip_limit != 1.0` and we enter an `uint8` image, we will receive back a `float` image.
I don't think that's "desirable" :) could you think on a solution for that, @rfezzani?

rfezzani:
Sorry to contradict you @alexdesiqueira, but `clip_limit` is multiplied by `nbins` [here](https://github.com/scikit-image/scikit-image/blob/786adad0da9b399c9c4689bf43b44b95011c13ac/skimage/exposure/_adapthist.py#L85) when `_clahe` is called...

rfezzani:
>  - when clip_limit == 1.0 and we enter an uint8 image, we will receive back an uint8 image;
>  - when clip_limit != 1.0 and we enter an uint8 image, we will receive back a float image.
    I don't think that's "desirable" :) could you think on a solution for that, @rfezzani?

Here I am facing a mystery... Before return, the output of `_clahe` is converted to float using `img_as_float` then intensities are rescaled using `rescale_intensity`, so the output should always be float!!

rfezzani:
> - when clip_limit == 1.0 and we enter an uint8 image, we will receive back an uint8 image;
> - when clip_limit != 1.0 and we enter an uint8 image, we will receive back a float image.
    I don't think that's "desirable" :) could you think on a solution for that, @rfezzani?

Actually, this is not the current behavior of `skimage`:

```python
>>> from skimage import data, exposure, util
>>> image = data.moon()  # dtype('uint8')>>> image_clahe = exposure.equalize_adapthist(image, clip_limit=1)  # dtype('float64')
>>> image_clahe.dtype
dtype('float64')
>>> image.dtype
dtype('uint8')
```
but the behavior with the proposed fix of this PR... For a coherent behavior, maybe a call to `img_as_float` before return is possible, but is it really what we want to do?

alexdesiqueira:
> Sorry to contradict you @alexdesiqueira, but `clip_limit` is multiplied by `nbins` here when `_clahe` is called...

You're not contradicting me @rfezzani :) we're saying different things, only.

> Here I am facing a mystery... Before return, the output of `_clahe` is converted to float using `img_as_float` then intensities are rescaled using `rescale_intensity`, so the output should always be float!!

Actually `_clahe` does not ask to convert to float, [`equalize_adapthist` does](https://github.com/scikit-image/scikit-image/blob/786adad0da9b399c9c4689bf43b44b95011c13ac/skimage/exposure/_adapthist.py#L85):

```
image = _clahe(image, kernel_size, clip_limit * nbins, nbins)
image = img_as_float(image)
```

However, you got a point. I assumed wrongly that the output of `_clahe` would be the actual output, and your solution works :) I stand corrected.
Thank you for your work @rfezzani!

rfezzani:
> Actually _clahe does not ask to convert to float, equalize_adapthist does

That's what I meant :-p
Thank you @alexdesiqueira :tada:

rfezzani:
@sciunto, this is not the case if `clip_limit == 1` and input image is not float (please see my  https://github.com/scikit-image/scikit-image/pull/4519#issuecomment-600813717)

sciunto:
Of course...

sciunto:
I would suggest to always return the same dtype, so to convert in this case. Otherwise, in a pipeline, it could generate issues due to edge effects. Any other opinion?

alexdesiqueira:
> I would suggest to always return the same dtype, so to convert in this case. 

That was the discussion I was having before @sciunto @rfezzani :) however, the function always return `float`. Please check my last comment.
I think @sciunto has a point here. Could you please specify that the function will return a `float`, @rfezzani? Thank you!

skimage/exposure/_adapthist.py
```diff
@@ -70,6 +70,10 @@ def equalize_adapthist(image, kernel_size=None,
     .. [1] http://tog.acm.org/resources/GraphicsGems/
     .. [2] https://en.wikipedia.org/wiki/CLAHE#CLAHE
     """
+
+    if clip_limit == 1.0:
+        return image  # is OK, immediately returns original image.
```
sciunto:
```suggestion
        return img_as_float(image)  # convert to float for concistancy
```

Just to be clear of I would prefer, but we can debate :)

alexdesiqueira: @sciunto `_clahe` always returns an `uint`. The output of `equalize_adapthist`, that receives `_clahe`, [converts to `float` already](https://github.com/scikit-image/scikit-image/blob/203cf03f8c523ab00ed13c58b2fa77eb7f51ee11/skimage/exposure/_adapthist.py#L89)... this would be redundant, I guess.

sciunto: but if clip_limit==1, you bypass _clahe, if I'm correct. So we need it here as well for the case of on input with an integer dtype.

alexdesiqueira: ... nevermind, @sciunto. You are right, I saw it before and I "un-saw" it somehow :sweat_smile: @rfezzani could you fix it, please? @sciunto could write in a few words what I couldn't explain before — and even find in the code at some point :)

rfezzani: I already proposed that in a previous comment:
> For a coherent behavior, maybe a call to img_as_float before return is possible, but is it really what we want to do?

alexdesiqueira: > I already proposed that in a previous comment:

Yes, and I'm sorry. As I said,

> I saw it before and I "un-saw" it somehow :sweat_smile:

alexdesiqueira:
Great @rfezzani, now we are only missing the info that it returns a float. Could you put that in the Notes section, please? Something like that:
```
Notes
--------
The datatype of the output array will be `float`.
(...)
```
Thank you!

sciunto:
Perfect!

alexdesiqueira:
I actually prefer the way I put (on `Notes`), but since you got it @sciunto, that's fine by me :)

PR #4524, opened 2020-03-23
Fix similarity transform scale

### LINKED ISSUE #4523: SimilarityTransform.scale implementation sometimes returns incorrect scale factor
## Description

The [current implementation](https://github.com/scikit-image/scikit-image/blob/e8dd5e0204e912931437d8da0f4f0ea353fe1950/skimage/transform/_geometric.py#L1169) used to obtain the scale factor relies on dividing out the cos(rot) or sin(rot) from the (0,0) or (1,0) element of the affine matrix, respectively, depending on if cos(rot) is close to 0. The initial check has the potential to fail, resulting in the wrong scale factor being returned.

A simpler, more stable implementation would be to use the determinant of the matrix to obtain the scale factor:

```python
# det = scale**(# of dimensions), therefore scale = det**(1/2)
np.sqrt(np.linalg.det(self.params))
```

## Way to reproduce

The following code defines an affine transformation with a rotation of 90 degrees and a scale factor of 1 (you can tell because element 1,0 is 1). 

```python
>>> import numpy as np; import skimage.transform
>>> m = np.array([[0,-1,1.33226763e-15],[1,2.22044605e-16,-1.33226763e-15],[0,0,1.00000000e+00]])
>>> skimage.transform.SimilarityTransform(m).scale
0.0
```

## Version information
```python
Linux-3.10.0-514.21.1.el7.x86_64-x86_64-with-redhat-7.3-Maipo
scikit-image version: 0.16.2
numpy version: 1.17.3
```

84danie:
## Description

Fixes #4523 

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -1167,12 +1167,8 @@ class SimilarityTransform(EuclideanTransform):
 
     @property
     def scale(self):
-        if abs(math.cos(self.rotation)) < np.spacing(1):
-            # sin(self.rotation) == 1
-            scale = self.params[1, 0]
-        else:
-            scale = self.params[0, 0] / math.cos(self.rotation)
-        return scale
+        # det = scale**(# of dimensions), therefore scale = det**(1/2)
+        return np.sqrt(np.linalg.det(self.params))
 
 
 class PolynomialTransform(GeometricTransform):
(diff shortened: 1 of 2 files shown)


sciunto:
Many thanks @84danie !

stefanv:
@84danie If we wanted to rewrite that test in terms of parameters, to make it clearer what is happening, do you remember which angles you used to generate those values?

84danie:
@stefanv The actual transformation was estimated by the [`astroalign`](https://astroalign.readthedocs.io/en/latest/api.html#astroalign.find_transform) library, which is why the values aren't exact - the expected values I put in the unit test are indeed the "true" values of the affine transformation. I tried to find a "cleaner" example, but this particular bug seemed to rely on the fact that the numbers _weren't_ exact.

PR #4535, opened 2020-03-27
New: Added new feature called augment_img for image augmentation

YashawantParab:
## Description

Added new feature for image augmentation.

augment_img tries to collaborate some of the image augmentation techniques into a single file.

Augmentation techniques such as:
    [1] Anti clockwise rotation
    [2] RGB to gray conversion
    [3] Clockwise rotation
    [4] Horizontal flip
    [5] Vertical flip
    [6] Warp shift
    [7] Add noise
    [8] Blur image
    [9] Logarithmic correction
    [10] Sigmoid correction
    [11] Rescale exposure intensity
    [12] Adjust gamma exposure 

This function help to generate images with different techniques used for image processing.


jni:
Thanks for opening this PR, @YashawantParab!

In general, I think image augmentation is in-scope for scikit-image, although the rest of the community would have to chime in.

I would also prefer an API that is:
- based on images, not image paths
- functional, rather than based on a class
- allows the user more control over the kinds of augmentation that they want. Someone might want flips along one axis but not another, and no rotations, for example.
- composable with other parts of scikit-image. In this case what am suggesting is a generator-based interface.
- allows multidimensional input. Specifically, I wouldn't use `v_flip` or things like that, rather than `np.flip(axis=...)`.

An example below:

```python
import os
from glob import glob
from skimage.io import imread, imsave
from skimage.transforms import augment

image_filenames = glob('path/to/input/images/*.png')
images = map(imread, image_filenames)

output_image_sets = (augment(image, modes=['flip', 'noise']) for image in images)

for input_filename, output_set in zip(image_filenames, output_image_sets):
    output_file_prefix = os.path.splitext(os.path.basename(input_filename))[0]
    for i, augmented in enumerate(output_set):
        output_file = os.path.join('path/to/output/images', output_file_prefix + f'_{i}.png')
        imsave(output_file, agumented)
```

This is more in line with our existing APIs. What do you think?

@scikit-image/core any more comments? Does data augmentation fit within things we want to provide in skimage?

sciunto:
I do not think that the gain of adding this feature is particularly significant, as I believe that the users who may call this function will be able to write it themselves (for machine learning?). Also, there is no unique combination and process, and so far, I do not see the point for providing a particular one.

emmanuelle:
Hi @YashawantParab thank you for your interest in scikit-image :-). Since this represents a significant addition to the API, could you please another time submit an issue first so that we can discuss whether this is in the scope of scikit-image or not? That said, we should make it clearer in the contributing notes that significant contributions should be discussed in issues first.

I am also hesitant about adding image augmentation to scikit-image, since good Python libraries already exist for this, such as imgaug (https://imgaug.readthedocs.io/en/latest/). Image augmentation would probably need to be a whole submodule like `skimage.aug`. Not saying we don't want it but it's a lot of work IMO. A first step would be to see what's been done in `imgaug` (or other libraries) and see if we could have an added value compared to it, for example a syntax more in line either with scikit-image or deep learning packages.

jni:
> see if we could have an added value compared to it

one possible added value is 3D and nD support (not sure whether imgaug has this, but this would need to be checked).

stefanv:
Also see https://keras.io/preprocessing/image/#imagedatagenerator-class

soupault:
My strong feeling is that augmentations are out-of-scope for skimage. There are multiple mature libraries already available implementing the augmentations in an efficient way:
- `albumentations` - https://github.com/albumentations-team/albumentations
- `solt` - https://github.com/MIPT-Oulu/solt
- `torchvision` - https://pytorch.org/docs/stable/torchvision/index.html
- `torchio` - https://github.com/fepegar/torchio
- `augmentor` - https://github.com/mdbloice/Augmentor
- `imgaug` - https://imgaug.readthedocs.io/en/latest/
- etc

If the final decision will still be to add this feature, I strongly suggest to focus on implementing the crucial aspects of augmentations, which are:
1. multiprocessing / utilization of all available CPU cores (because of this, many libraries from the list above use OpenCV under the hood)
2. just-in-time optimization of the composite function (for example, `solt` performs fusion of multiple subsequent geometric operations into one)
3. performance (i.e. operations/sec, see https://github.com/MIPT-Oulu/solt#benchmark)
4. serialization of the transforms for experiment tracking and reproducibility (killer-feature of `solt`)
5. full support for multichannel images (this is quite rare due to ambiguity of the colorspace selection)
6. full support for 3D images (this is very rare and, I believe, requires a powerful backend, such as GPU - torchvision, torchio are on this path)
7. extensive testing (`albumentations` and `solt` have massive test coverage)
8. support for different data types (besides intensity images), such as masks, keypoints (I think, `solt` is currently the only library that does that).

Finally, even if using image augmentation is still quite common, I think better ways of regularization for DNNs are emerging (https://arxiv.org/abs/1710.10686, https://arxiv.org/abs/1707.03631, http://openaccess.thecvf.com/content_cvpr_2018/papers/Poursaeed_Generative_Adversarial_Perturbations_CVPR_2018_paper.pdf).

Disclaimer: `solt` is developed in my department by @lext.

stefanv:
I agree with @soupault, but would like to see a chapter on this topic in the user guide.

jni:
> full support for 3D images (this is very rare and, I believe, requires a powerful backend, such as GPU - torchvision, torchio are on this path)

@soupault that's an opportunity then, and I disagree about powerful backend being needed: augmentation is very unlikely to be the rate-limiting step in 3D deep learning. Even if you spend (checks) 60s adding Gaussian noise to a (512, 512, 512) image, you're likely to spend 1h or more doing gradient descent on it, so the initial cost is almost trivial.

Additionally, do note that we have plans to start incorporating OpenCL or similar into our processing, so things like this might be accelerated "automagically" in the future. The API and the value-added components are far more important than performance.

Having said all this, I sense that most people are not interested in maintaining an image augmentation module in skimage, but a user guide entry showing how to augment nD images using scikit-image might be welcome. As such, I am going to close this PR. @YashawantParab thank you for your efforts! I would suggest that you aim to distribute your work as a standalone Python package. Please see this excellent guide by the Python Packaging Authority (PyPA):

https://packaging.python.org/tutorials/packaging-projects/

Thanks again.

PR #4543, opened 2020-03-31
Documentation: add binder to gallery examples

emmanuelle:
Supersedes #4485 and #4340, in order to add binder links to gallery examples. The strategy is the one proposed by @betatim in #4485: add a `postBuild` script for binder that will build the notebooks using the scikit-image version from the branch from which binder was launched. This solves elegantly version problems. I checked from another branch on my fork that it works correctly (if you want to check, pull https://github.com/emmanuelle/scikit-image/tree/binder-from-tim, build the doc locally and follow a binder link).

If this PR is merged binder links will work for the dev examples.

At release time, one should update the `doc/source/conf.py` as explained in the updated release notes to generate suitable links for the stable examples.

Apologies to @betatim, I should have started from your commits to give you better credit. Whoever merges this PR, please mention Tim in the commit message.

### DIFF
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,0 +1,25 @@
+Cython>=0.29.13
+wheel
+# numpy 1.18.0 breaks builds on MacOSX
+# https://github.com/numpy/numpy/pull/15194
+numpy>=1.15.1,!=1.18.0
+scikit-learn
+matplotlib>=3.0.1
+dask[array]>=0.15.0
+# cloudpickle is necessary to provide the 'processes' scheduler for dask
+cloudpickle>=0.2.1
+pandas>=0.23.0
+seaborn>=0.7.1
+scipy>=1.0.1
+networkx>=2.0
+pillow>=4.3.0
+imageio>=2.3.0
+tifffile>=2019.7.26
+PyWavelets>=0.5.2
+pooch>=0.5.2
+SimpleITK
+astropy>=1.2.0
+qtpy
+pyamg
+
+
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -21,5 +21,4 @@ SimpleITK
 astropy>=1.2.0
 qtpy
 pyamg
-
-
+sphinx-gallery
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,24 +1,8 @@
-Cython>=0.29.13
-wheel
-# numpy 1.18.0 breaks builds on MacOSX
-# https://github.com/numpy/numpy/pull/15194
-numpy>=1.15.1,!=1.18.0
+-r ../requirements/default.txt
+-r ../requirements/build.txt
 scikit-learn
-matplotlib>=3.0.1
 dask[array]>=0.15.0
-# cloudpickle is necessary to provide the 'processes' scheduler for dask
 cloudpickle>=0.2.1
 pandas>=0.23.0
 seaborn>=0.7.1
-scipy>=1.0.1
-networkx>=2.0
-pillow>=4.3.0
-imageio>=2.3.0
-tifffile>=2019.7.26
-PyWavelets>=0.5.2
-pooch>=0.5.2
-SimpleITK
-astropy>=1.2.0
-qtpy
-pyamg
 sphinx-gallery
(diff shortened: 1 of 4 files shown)


emmanuelle:
thanks @jni! Kudos should go to @betatim mostly

betatim:
Thanks for picking this up and working on it!

.binder/requirements.txt
```diff
@@ -0,0 +1,24 @@
+Cython>=0.29.13
```
soupault, line 1: Could we revisit my earlier suggestion for the requirements file - https://github.com/scikit-image/scikit-image/pull/4340#discussion_r357905639 ?

emmanuelle: So I partially intregrated your suggestion :-), in the sense that I used two existing files but for the doc-specific requirements (scikit-learn etc.) I did not want to use all the requirements of `docs.txt` since we don't need sphinx copy-button etc. Another possibility would be to separate docs.txt in two files, one  `docs_build.txt` for the build itself and one `docs_examples.txt` for running the examples only.

emmanuelle:
> Thanks for picking this up and working on it!
> 
> Did you try to do something like https://github.com/scikit-learn/scikit-learn/blob/8910b14fa8630c4d9ef477a2d339fdf68eecd4b9/doc/conf.py#L257-L261 to get the branch "automagically" instead of having to change it manually when releasing?

Great idea! I integrated this suggestion and changed the comment in the release notes to just verify that names are correct (we can remove this comment after one or two release but let's double-check the first time!).

emmanuelle:
CI failure will hopefully be fixed by #4551

doc/source/conf.py
```diff
@@ -157,6 +131,57 @@
...
+        '../examples/data',
+        '../examples/numpy_operations',
+        '../examples/color_exposure',
+        '../examples/edges',
+        '../examples/transform',
+        '../examples/filters',
+        '../examples/features_detection',
+        '../examples/segmentation',
+        '../examples/applications',
+        '../examples/developers',
```
rfezzani, line 170: Not specifically related to this issue, but I see that `'../examples/registration'` is not included to the examples gallery, otherwise very cool stuff. Thank you.

emmanuelle: good catch @rfezzani I've opened #4553 to track this (I didn't just fix it here because it would be good also to move some examples from transform to registration)

emmanuelle:
so two :+1: please merge so that we can check if the dev doc if binder links work :-)

rfezzani:
I hope I correctly gave credits to @betatim ;-)

betatim:
There is now a badge at the bottom of the examples and the link looks reasonable too. Waiting for the image to build :)

Screenshot 2020-04-04 at 14 23 10

Unfortunately the build fails with this:
```
Step 42/51 : RUN ${KERNEL_PYTHON_PREFIX}/bin/pip install --no-cache-dir -r ".binder/requirements.txt"
 ---> Running in 6d744265fa86
WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip.
Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue.
To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.
ERROR: Double requirement given: numpy!=1.18.0,>=1.15.1 (from -r .binder/../requirements/build.txt (line 5)) (already in numpy>=1.15.1 (from -r .binder/../requirements/default.txt (line 1)), name='numpy')
Removing intermediate container 6d744265fa86
The command '/bin/sh -c ${KERNEL_PYTHON_PREFIX}/bin/pip install --no-cache-dir -r ".binder/requirements.txt"' returned a non-zero code: 1
```
I think the "pip wrapper script" warning is irrelevant for this (sth repo2docker needs to fix). The error makes me think that the sum of all requirements files that get included mention `numpy` twice. The constraints seem compatible but apparently pip doesn't like you mentioning something twice.

Should we make a new issue or does someone have an idea how to rearrange the requirements files (-> directly open a PR)?

emmanuelle:
I opened #4555

.binder/requirements.txt
soupault: I would just use the same docs requirements, I do not see overshooting to be a major issue. The current version is also fine. Thank you, Emma!

Issue #4540, opened 2020-03-30
Broken plot in userguide


sciunto:
## Description
In this section.
https://scikit-image.org/docs/stable/user_guide/numpy_images.html#color-images

emmanuelle:
This is weird because we are using matplotlib's plot directive, which is correctly imported (I checked it in #4554). Note that we are only using the plot directive for this example so if we completely change this example maybe you can remove this dependency.

dsandeep0138:
@emmanuelle Can you please review the PR? Plot directive is working now with the changes. Below is the screenshot of the page after the fix.


![scikit-image](https://user-images.githubusercontent.com/3874111/85373702-42df2c80-b4f9-11ea-8f58-5968f78742f5.PNG)

PR #4554, opened 2020-04-04
Debug plot directive

emmanuelle:
Follow-up on #4541 and #4540, trying to understand how the plot directive is imported.


emmanuelle:
So here I'm trying to understand why the plot directive is not executed properly in the example of #4540. Note that we use this plot directive only once in the whole library.

emmanuelle:
So it seems that the plot directive is imported correctly from matplotlib. Don't understand why our example is not displayed properly. Closing for now.

PR #4555, opened 2020-04-04
Fix binder requirements

emmanuelle:
The binder deployment currently fails because of duplicated requirements for numpy.

(Sorry about that, I had first tested the deployment using a local branch, and then I addressed a review comments but did not check again...)

### DIFF
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,5 +1,7 @@
 -r ../requirements/default.txt
-# numpy is present both in default.txt and build.txt, with pinning instructions which are slightly different, therefore pip gives up and errors
+# numpy is present both in default.txt and build.txt
+# with pinning instructions which are slightly different, 
+# therefore pip gives up and errors
 Cython>=0.29.13
 wheel
 scikit-learn
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,4 +1,5 @@
 -r ../requirements/default.txt
+# numpy is present both in default.txt and build.txt, with pinning instructions which are slightly different, therefore pip gives up and errors
 Cython>=0.29.13
 wheel
 scikit-learn
--- a/.binder/requirements.txt
+++ b/.binder/requirements.txt
@@ -1,5 +1,6 @@
 -r ../requirements/default.txt
--r ../requirements/build.txt
+Cython>=0.29.13
+wheel
 scikit-learn
 dask[array]>=0.15.0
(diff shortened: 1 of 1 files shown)


betatim:
This PR (https://mybinder.org/v2/gh/emmanuelle/scikit-image/fix-binder-requirements) builds for me and I ran the "color_exposure/plot_histogram_matching.ipyn" example which  worked.

So a 👍  from me.

hmaarrfk:
can you please add a comment as to why this is done. 

I have also hit the error you did, but it is rather obscur.

emmanuelle:
@hmaarrfk sure, the problem is that numpy is present both in `default.txt` and `build.txt`, with pinning instructions which are slightly different, therefore pip gives up and errors

.binder/requirements.txt
```diff
@@ -1,5 +1,6 @@
 -r ../requirements/default.txt
--r ../requirements/build.txt
+Cython>=0.29.13
```
hmaarrfk, line 2:
```suggestion
# numpy is present both in default.txt and build.txt, with pinning instructions which are slightly different, therefore pip gives up and errors
Cython>=0.29.13
```

hmaarrfk:
Shoudl we just make the requirements the same? I dont' really mind skipping 1.18.0 all together for the builds.

emmanuelle:
@hmaarrfk I wouldn't mind but I did not want to make changes to other files in a PR which is just a quick bug fix, since our binder links in the dev gallery currently link to a broken deployment... I'm happy to open an issue about syncing numpy versions and changing the binder requirements if you merge this one :-).

emmanuelle:
@rfezzani sorry I don't see your comment...

.binder/requirements.txt
```diff
@@ -1,5 +1,7 @@
 -r ../requirements/default.txt
--r ../requirements/build.txt
+# numpy is present both in default.txt and build.txt, with pinning instructions which are slightly different, therefore pip gives up and errors
```
rfezzani, line 2: Please wrap this line to 80 characters length, if you wish :-)

rfezzani:
Oups, I did not push the send button :sweat_smile:

.binder/requirements.txt
emmanuelle, line 2:
```suggestion
# numpy is present both in default.txt and build.txt
# with pinning instructions which are slightly different, 
# therefore pip gives up and errors
```

jni:
🎉

PR #4560, opened 2020-04-07
Do not enforce pil plugin in skimage.data

sciunto:
## Description

I noticed that the imread call in our data module currently enforces the pil plugin, this was the case even before the integration of pooch by @hmaarrfk 

I propose to leave it to None to let the best plugin to be chosen, imageio in most cases.
 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


hmaarrfk:
I think it was introoduced when we made the switch from PIL -> Imageio to avoid any bugs in the transition.

emmanuelle:
Thanks @sciunto, merging.

Issue #3407, opened 2018-09-13
potential updates related to PyWavelets


grlee77:
## Description
This issue is just to document a few potential changes related to the PyWavelets 1.0 release that I wanted to document here while they are on my mind. These are potential changes that may be worth making at a future point once the minimum requirement is bumped up to 1.0 (probably best to wait a while still as it was only recently released). 

**1. new thresholding mode**
In addtion to 'soft' and 'hard' thresholding, PyWavelets 1.0 also accepts  'garotte' for non-negative Garotte thresholding which is intermediate between soft and hard thresholding. This can be used without any changes to the scikit-image `denoise_wavelet` code, so we could optionally update the denoise_wavlet docstring to add that to the set of modes listed for the `mode` parameter.

**2. code simplification**
PyWavelets 1.0 also allows minor simplification of current code in `restoration/_denoising.py`:

from:
```Python
    if wavelet_levels is None:
        # Determine the maximum number of possible levels for image
        dlen = wavelet.dec_len
        wavelet_levels = np.min(
            [pywt.dwt_max_level(s, dlen) for s in image.shape])
```
to:
```Python
    if wavelet_levels is None:
        # Determine the maximum number of possible levels for image
        wavelet_levels = pywt.dwtn_max_level(image.shape, wavelet)
```

**3. potential refactor**
Finally, we might get a minor performance boost in the multichannel case by passing the axes to transform rather than looping over the channels axis such as (would require minor changes within `_wavelet_threshold` itself). This should also work in PyWavelets 0.5.x:

current code:
```Python
out = np.empty_like(image)
for c in range(image.shape[-1]):
    out[..., c] = _wavelet_threshold(image[..., c],
                                     wavelet=wavelet,
                                     method=method,
                                     sigma=sigma[c], mode=mode,
                                     wavelet_levels=wavelet_levels)
```

new code:
```Python
dwt_axes = tuple(range(image.ndim - 1))
out = _wavelet_threshold(image,
                         wavelet=wavelet,
                         method=method,
                         sigma=sigma, mode=mode,
                         wavelet_levels=wavelet_levels,
                         dwt_axes=dwt_axes)
```

grlee77:
Closing as mostly done in #4568 and I no longer think the remaining refactor for item 3. above is desirable. The current implementation makes it easier to deal with varying noise across channels, while the refactored version would always use the same value for all channels unless further refactoring was done.

PR #4568, opened 2020-04-08
Bump Pywavelets to 1.0.0

sciunto:
## Description

We can discuss either we merge this PR now or after 0.17 is out. This PR corresponds to one of the tasks in the TODO list.
Pywavelets 1.0.0 was released in Aug 2018. https://github.com/PyWavelets/pywt/releases

It also applies the suggestion 2. in #3407

### DIFF
--- a/skimage/restoration/tests/test_denoise.py
+++ b/skimage/restoration/tests/test_denoise.py
@@ -15,11 +15,6 @@ from skimage._shared._warnings import expected_warnings
 from distutils.version import LooseVersion as Version
 
 
-if (Version(pywt.__version__) <= '0.5.2'):
-    PYWAVELET_ND_INDEXING_WARNING = 'Using a non-tuple sequence'
-else:
-    PYWAVELET_ND_INDEXING_WARNING = None
-
 try:
     import dask
 except ImportError:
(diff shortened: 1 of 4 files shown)


sciunto:
The failure is unrelated.

JDWarner:
Should we consider going all the way to 1.1 (basically pywt current release)?

1.1 is where they dropped Python 2, so thematically it seems like it would match our intent to do the same.  The changes from pywt 1.0 to 1.1 seem essentially uniformly positive.  The backwards incompatible changes are both actually bugfixes from what I can tell.

emmanuelle:
@JDWarner what would we gain from bumping to 1.1 instead of 1.0? If the warnings disappear, I'm happy with the lowest version enabling this.

rfezzani:
Minimum version set to 1.0 is fine for me.

JDWarner:
Moving to 1.1 directly keeps us from indirectly creating a breaking change later when we do.

One of the changes in 1.1 is a fix where previously the `cwt` results replicated the results in Matlab's 2012a version.  Except Matlab was wrong.  The correct results per the literature reference paper are the complex conjugate of what Matlab 2012a and prior (as well as pywt < 1.1) returned.  Matlab has since fixed this.

The `cwt` wavelet in 1.1 also now has a FFT mode which dramatically improves performance.

emmanuelle:
OK I see. Thanks for the update @JDWarner. If there are important bug fixes then it's different. Anyway it's not like asking people to upgrade their numpy or jupyter, it's a less frequent dependency so we won't mess up too much with people environments.

sciunto:
> Anyway it's not like asking people to upgrade their numpy or jupyter, it's a less frequent dependency so we won't mess up too much with people environments.

@emmanuelle but pywt is in the default requirements. What makes the difference with numpy?

> Moving to 1.1 directly keeps us from indirectly creating a breaking change later when we do.

Does it mean that if I increase the min version, I have to do extra work on this PR? (I'm not sure to understand what you mean...)

emmanuelle:
@sciunto the difference is that `pywt` is used by less projects than numpy (but checking on Github's "used by", pywt is still used by 16.4k projects... for comparison scikit-image is 35k so it's not a small number). Therefore the probability that people need to use an old version of pywt for a different project is quite small.

I think that @JDWarner is just suggesting to change to 1.1.1. At the moment we're not using `pywt.cwt` (from a quick grep, correct me if I'm wrong) so we're not breaking anything, but if we'd start using cwt then it would be a breaking change to switch to 1.0 to 1.1. Is this what you meant @JDWarner ?

sciunto:
Thank you for the clarifications @emmanuelle I let @JDWarner confirms.

@scikit-image/core Do we have a consensus/majority to set the min version to 1.1.1 for pywavelets?

grlee77:
I am +1 on bumping to 1.1. Functionality should be the same as what we have now, but it does allow some minor refactoring/cleanup as partially done in this PR. I would be happy to make a PR with the other suggested change from #3407 afterwards.

rfezzani:
I am +1 for 1.1

skimage/restoration/tests/test_denoise.py
```diff
@@ -898,12 +868,10 @@ def test_cycle_spinning_num_workers():
     func_kw = dict(sigma=sigma, multichannel=True, rescale_sigma=True)
 
     # same results are expected whether using 1 worker or multiple workers
-    with expected_warnings([PYWAVELET_ND_INDEXING_WARNING]):
-        dn_cc1 = restoration.cycle_spin(noisy, denoise_func, max_shifts=1,
+    dn_cc1 = restoration.cycle_spin(noisy, denoise_func, max_shifts=1,
                                         func_kw=func_kw, multichannel=False,
```
stefanv, line 872: Just check indentation here.

rfezzani:
:tada: Thank you @sciunto!

PR #4571, opened 2020-04-09
Remove binary specification from match_descriptors docstring

### LINKED ISSUE #4570: match_descriptors docstring describes descriptors too narrowly
The docstring for `match_descriptors` currently reads:

```
Parameters
----------
descriptors1 : (M, P) array
    Binary descriptors of size P about M keypoints in the first image.
descriptors2 : (N, P) array
    Binary descriptors of size P about N keypoints in the second image.
metric : {'euclidean', 'cityblock', 'minkowski', 'hamming', ...} , optional
    The metric to compute the distance between two descriptors. See
    `scipy.spatial.distance.cdist` for all possible types. The hamming
    distance should be used for binary descriptors. By default the L2-norm
    is used for all descriptors of dtype float or double and the Hamming
    distance is used for binary descriptors automatically.
...
```
however, the method supports float descriptors as well. The docstring description for `descriptors1` and `descriptors2` should drop the `Binary` adjective.

notmatthancock:
Closes #4570

### DIFF
--- a/skimage/feature/match.py
+++ b/skimage/feature/match.py
@@ -13,9 +13,9 @@ def match_descriptors(descriptors1, descriptors2, metric=None, p=2,
     Parameters
     ----------
     descriptors1 : (M, P) array
-        Binary descriptors of size P about M keypoints in the first image.
+        Descriptors of size P about M keypoints in the first image.
     descriptors2 : (N, P) array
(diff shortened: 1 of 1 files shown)


soupault:
@notmatthancock good catch! Thanks for the contribution!

PR #4579, opened 2020-04-13
update task in TODO.txt

sciunto:
## Description

It seems that this task has been done already. The mentioned file is not in the repository anymore.


emmanuelle:
the file has been moved to `morphology` but I think the fused type still exists?

sciunto:
You're right @emmanuelle I didn't catch it. I fixed the task.

PR #4585, opened 2020-04-16
Update rescale_intensity to prevent under/overflow and produce proper output dtype

### LINKED ISSUE #4583: funky behavior of `rescale_intensity`
## Description
Looks like `rescale_intensity` does some weird stuff when the `in_range` argument has `np.int8` dtype. Unfortunately, this naturally arises when `in_range` is a result computed from the data to be rescaled. E.g., if I have an array of `int8` that I want to rescale, the min and max of that array will also be `int8`, which apparently induces weird output from `rescale_intensity`.   

## Way to reproduce
```python
from skimage.exposure import rescale_intensity
import numpy as np

out_range = 'uint8'
arr = np.array([-128, 0, 127]).astype('int16')

print("arr:", arr)

clims_a = (-128, 127)
a = rescale_intensity(arr, in_range=clims_a, out_range=out_range)
print("a:  ", a)

clims_b = (np.int8(-128), np.int8(127))
b = rescale_intensity(arr, in_range=clims_b, out_range=out_range)
print("b:  ", b)
```

## Version information
```python
3.8.1 (default, Jan  8 2020, 22:29:32) 
[GCC 7.3.0]
Linux-3.10.0-957.10.1.el7.x86_64-x86_64-with-glibc2.10
scikit-image version: 0.16.2
numpy version: 1.18.1
```

## Output
```python
arr: [-128    0  127]
a:   [  0 128 255]
b:   [     0 -32640    511]
```

jni:
## Description

Fixes #4583 

- All computations are now in float, preventing under/overflow
- If `out_range` is a dtype, then the output will be in that dtype (currently it is the input dtype, which is insane).
- If `out_range` is a range of numbers, the output will be of type float and the caller will be responsible for converting to the correct dtype. **This is a breaking change in some circumstances.** See our use of `rescale_intensity` within `equalize_adapthist` which I had to fix.

## Checklist

  existing benchmark~~
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -75,7 +75,9 @@ def equalize_adapthist(image, kernel_size=None,
         return img_as_float(image)  # convert to float for consistency
 
     image = img_as_uint(image)
-    image = rescale_intensity(image, out_range=(0, NR_OF_GRAY - 1))
+    image = np.round(
+        rescale_intensity(image, out_range=(0, NR_OF_GRAY - 1))
+    ).astype(np.uint16)
 
     if kernel_size is None:
         kernel_size = (image.shape[0] // 8, image.shape[1] // 8)
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -65,6 +65,10 @@ def equalize_adapthist(image, kernel_size=None,
        - The image is converted back to RGB space and returned
     * For RGBA images, the original alpha channel is removed.
 
+    .. versionchanged:: 0.17
+        The values returned by this function are slightly shifted upwards
+        because of an internal change in rounding behavior.
+
     References
     ----------
     .. [1] http://tog.acm.org/resources/GraphicsGems/
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -271,8 +271,8 @@ def _output_dtype(dtype_or_range):
 
     The dtype is determined according to the following rules:
     - if ``dtype_or_range`` is a dtype, that is the output dtype.
-    - if ``dtype_or_range`` is a dtype string, that is the dtype used, unless it
-      is not a NumPy data type (e.g. 'uint12' for 12-bit unsigned integers),
+    - if ``dtype_or_range`` is a dtype string, that is the dtype used, unless
+      it is not a NumPy data type (e.g. 'uint12' for 12-bit unsigned integers),
       in which case the data type that can contain it will be used
       (e.g. uint16 in this case).
     - if ``dtype_or_range`` is a pair of values, the output data type will be
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -270,12 +270,12 @@ def _output_dtype(dtype_or_range):
     """Determine the output dtype for rescale_intensity.
 
     The dtype is determined according to the following rules:
-    - if ``out_range`` is a dtype, that is the output dtype.
-    - if ``out_range`` is a dtype string, that is the dtype used, unless it
+    - if ``dtype_or_range`` is a dtype, that is the output dtype.
+    - if ``dtype_or_range`` is a dtype string, that is the dtype used, unless it
       is not a NumPy data type (e.g. 'uint12' for 12-bit unsigned integers),
       in which case the data type that can contain it will be used
       (e.g. uint16 in this case).
-    - if ``out_range`` is a pair of values, the output data type will be
+    - if ``dtype_or_range`` is a pair of values, the output data type will be
       float.
 
     Parameters
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -303,6 +303,11 @@ def _output_dtype(dtype_or_range):
         except TypeError:  # uint10, uint12, uint14
             # otherwise, return uint16
             return np.uint16
+    else:
+        raise ValueError(
+            'Incorrect value for out_range, should be a valid image data '
+            f'type or a pair of values, got {dtype_or_range}.'
+        )
 
 
 def rescale_intensity(image, in_range='image', out_range='dtype'):
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -385,9 +385,8 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     array([  0,  63, 127], dtype=int8)
 
     """
-    input_dtype = image.dtype.type
     if out_range in ['dtype', 'image']:
-        out_dtype = _output_dtype(input_dtype)
+        out_dtype = _output_dtype(image.dtype.type)
     else:
         out_dtype = _output_dtype(out_range)
 
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -341,6 +341,12 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
         Image array after rescaling its intensity. This image is the same dtype
         as the input image.
 
+    Notes
+    -----
+    .. versionchanged:: 0.17
+        The dtype of the output array has changed to match the output dtype, or
+        float if the output range is specified by a pair of floats.
+
     See Also
     --------
     equalize_hist
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -374,10 +374,15 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     array([  0,  63, 127], dtype=int8)
 
     """
-    dtype = image.dtype.type
+    input_dtype = image.dtype.type
+    if out_range in ['dtype', 'image']:
+        out_dtype = _output_dtype(input_dtype)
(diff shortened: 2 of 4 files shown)


skimage/exposure/exposure.py
```diff
@@ -266,6 +266,39 @@ def intensity_range(image, range_values='image', clip_negative=False):
...
+        return out_range
+    if out_range in DTYPE_RANGE:
+        # string key in DTYPE_RANGE dictionary
+        try:
+            # if it's a canonical numpy dtype, convert
+            return np.dtype(out_range).type
+        except TypeError:  # uint10, uint12, uint14
+            # otherwise, return uint16
+            return np.uint16
+
```
emmanuelle, line 300: add an else here to raise a `TypeError` for example in the case of typos (`flat` instead of `float`). `output_dtype` should be None if none of the ifs is True, and `np.array([1, 2, 3], dtype=None)` does not error...

emmanuelle:
One possibility we did not consider would be to accept for `output_range` a tuple with 3 elements, the last one being a dtype, for example `(0, 200, np.uint8)`. Not sure we want this, but this could avoid an `astype` call when wanting a specific dtype and not floats.

skimage/exposure/exposure.py
```diff
@@ -266,6 +266,39 @@ def intensity_range(image, range_values='image', clip_negative=False):
     return i_min, i_max
 
 
+def _output_dtype(out_range):
```
lagru, line 269: I feel like we could choose a more general name that covers `input_dtype` as well. Maybe `dtype_or_range`? Might be a little easier to read.

skimage/exposure/exposure.py
```diff
@@ -341,10 +374,15 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     array([  0,  63, 127], dtype=int8)
 
     """
-    dtype = image.dtype.type
+    input_dtype = image.dtype.type
```
lagru, line 377: This is only used once now, maybe move it into the if-clause or remove the assignment altogether?

skimage/exposure/tests/test_exposure.py
```diff
@@ -237,10 +237,16 @@ def test_rescale_in_range_clip():
...
+
+    .. versionchanged:: 0.17
+        This function used to return dtype matching the input dtype. It now
+        matches the output.
+    """
     image = np.array([-10, 0, 10], dtype=np.int8)
     out = exposure.rescale_intensity(image, out_range=(0, 127))
-    assert out.dtype == np.int8
-    assert_array_almost_equal(out, [0, 63, 127])
+    assert out.dtype == np.float_
```
lagru, line 248:
```suggestion
    assert out.dtype == np.float64
```
I find this more explicit or is there a reason you prefer `float_`? Same in the test below...

skimage/exposure/exposure.py
lagru: I understand and support the first part of your suggestion. 

But why should `output_dtype` be `None`. The way I understand the new behavior is 
- `out_range` is an actual range -> output dtype is `float64`
- `out_range` is already a dtype (corresponding to the input dtype) -> use same type for output

`None` would just result in the returned output being of dtype `float64` wouldn't it?

emmanuelle: @lagru probably what I wrote was not clear, please tell me if it's clearer now. The function should never return None, it should return a valid value or error out.

lagru: Ah, then we are in agreement. :+1:

skimage/exposure/exposure.py
jni: @lagru 

1. the function name should relate to its return value (always a dtype), not its input parameter! So in this case I think you would want `_dtype_from_dtype_or_range`, which is quite a mouthful!
2. The input dtype is never in question, because it comes from the input array, which is given. So I don't think it's super useful to generalise this for now.

skimage/exposure/tests/test_exposure.py
jni: I prefer `float_` because it is the default float value for that platform. In some platforms, `np.zeros(a_shape)` returns float32, not float64. I think we should return this default type, and that is what `float_` evaluates to.

skimage/exposure/exposure.py
jni: LOL **ignore me**, I just saw that you were referring to the argument. :man_facepalming: I shouldn't comment on PRs before I've finished my morning coffee! :coffee: :sweat_smile:

jni:
@emmanuelle @lagru Thank you for the reviews! I've addressed all of your comments.

@emmanuelle 

> One possibility we did not consider would be to accept for output_range a tuple with 3 elements, the last one being a dtype, for example `(0, 200, np.uint8)`

I don't like that solution very much, it seems clunky and it will complicate the code quite a lot. I *would* consider adding `dtype=` as an additional keyword argument, but then the desired behaviour would be unclear if e.g. I pass `out_range='uint16', dtype=np.uint32`. I would say that the `dtype=` argument should override all previous options, but that's just one of many possible alternatives.

At any rate, I think we can discuss this separately and address it in a separate PR, since this is already a big improvement on existing behaviour. None of the above proposals would allow existing code to continue working when the output dtype was incorrectly being inferred from the input dtype (as in my fix for `equalize_adapthist`), so I don't think they are critical to get in now.

skimage/exposure/tests/test_exposure.py
lagru: Ah that makes sense. :+1:  Strange that my search pointed me to an old NumPy doc version with the statement "[Shorthand for float64](https://docs.scipy.org/doc/numpy-1.13.0/user/basics.types.html#array-types-and-conversions-between-types)" first.

skimage/exposure/exposure.py
```diff
@@ -266,6 +266,50 @@ def intensity_range(image, range_values='image', clip_negative=False):
...
+def _output_dtype(dtype_or_range):
+    """Determine the output dtype for rescale_intensity.
+
+    The dtype is determined according to the following rules:
+    - if ``out_range`` is a dtype, that is the output dtype.
+    - if ``out_range`` is a dtype string, that is the dtype used, unless it
+      is not a NumPy data type (e.g. 'uint12' for 12-bit unsigned integers),
+      in which case the data type that can contain it will be used
+      (e.g. uint16 in this case).
+    - if ``out_range`` is a pair of values, the output data type will be
```
lagru, line 278:
```suggestion
    - if ``dtype_or_range`` is a dtype, that is the output dtype.
    - if ``dtype_or_range`` is a dtype string, that is the dtype used, unless it
      is not a NumPy data type (e.g. 'uint12' for 12-bit unsigned integers),
      in which case the data type that can contain it will be used
      (e.g. uint16 in this case).
    - if ``dtype_or_range`` is a pair of values, the output data type will be
```

jni:
Thanks for keeping me honest, @lagru! 😂

skimage/exposure/_adapthist.py
```diff
@@ -75,7 +75,9 @@ def equalize_adapthist(image, kernel_size=None,
         return img_as_float(image)  # convert to float for consistency
 
     image = img_as_uint(image)
-    image = rescale_intensity(image, out_range=(0, NR_OF_GRAY - 1))
+    image = np.round(
+        rescale_intensity(image, out_range=(0, NR_OF_GRAY - 1))
+    ).astype(np.uint16)
```
lagru, line 80: Do we need to worry that the rounding here is a breaking change? See the changed unit test below..

m-albert:
Thanks for the fix! After this is merged I'd raise another issue (I think independent of this one) on this function with a proposed fix. Cheers

jni:
@m-albert why do you need to wait until this is merged?

skimage/exposure/_adapthist.py
jni: @lagru Ah, yes, I'd meant to discuss that. imho rounding is the correct behaviour and I would consider this a bug fix, but we should definitely discuss it! @scikit-image/core any thoughts on this?

m-albert:
@jni Somehow I was assuming it would be cleaner to address it once this is merged, but actually I can just open the issue, sorry for the mystery. See https://github.com/scikit-image/scikit-image/issues/4596

skimage/exposure/_adapthist.py
rfezzani: I agree with @jni that the the rounding is the correct behavior. The best that we can do is may be to add a note in the documentation about this breaking change.

alexdesiqueira: > imho rounding is the correct behaviour and I would consider this a bug fix

+1.

jni: I've added a note to the release notes and a versionchanged marker to both functions.

jni:
@scikit-image/core I've addressed the concerns about changed behaviours by adding more explicit notes in the release notes, as well as `.. versionchanged::` markers in the docstrings. Pending CI I think this is ready to be merged!

emmanuelle:
Thank you @jni, merging!

PR #4587, opened 2020-04-16
DO NOT MERGE Fix empty image after skeletonize_3D

### LINKED ISSUE #3757: skeletonize_3d produces a zero array in certain conditions
## Description
skeletonize_3d produces a zero array at certain sizes/thickness of the structure in the volume.

## Way to reproduce
In this code I create a 3D volume and draw a simple straight line with to different sizes.
The code here below produces a coherent skeleton:
```python
import numpy as np
from skimage.morphology import skeletonize_3d

image = np.zeros((100,100,100), np.uint8)
image[50:55,50:54, :] = 1

skeleton = skeletonize_3d(image)

print(skeleton.sum())
# output >> 96
```

this produces a zero array. The only parameter changed is the thickness of the line in the first dimension.
```python
image = np.zeros((100,100,100), np.uint8)
image[50:54,50:54, :] = 1

skeleton = skeletonize_3d(image)

print(skeleton.sum())
# output >> 0
```

## Version information
```python
```

```python
3.5.6 |Anaconda custom (64-bit)| (default, Aug 26 2018, 21:41:56) 
[GCC 7.3.0]
Linux-4.15.0-22-generic-x86_64-with-debian-buster-sid
scikit-image version: 0.14.2
numpy version: 1.15.2
```

sciunto:
## Description

This is an attempt to fix #3757 based on the patch written by DMST1990 https://github.com/scikit-image/scikit-image/issues/3757#issuecomment-613863424

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


sciunto:
imagej has not patched yet.

In short: we copied imagej. There is also a matlab version. The matlab version was fixed by a contributor. The contributor does not have the skills to implement it in java. I got this explanation two days ago and I learnt that matlab was patched at this occasion. It seems imagej was not aware as well (according to their issue tracker).

sciunto:
I have to modify the comment for this test at least.

sciunto:
Done. And I simplified the logic as well.

sciunto:
I keep testing this. It turns out that a 2D square is not skeletonized with skeletonize_3d, in contrast to the original version. Just an observation for now...

sciunto:
Checked with napari in 3D, the patch is not satisfying... :(

PR #4589, opened 2020-04-16
[MRG] Fix Negative eigenvalue in inertia_tensor_eigvals due to floating point precision

### LINKED ISSUE #4588: Negative eigenvalue in inertia_tensor_eigvals due to floating point precision
## Description
skimage.measure.inertia_tensor_eigvals can return a very slightly negative eigenvalue even though it is given a value of mu that should yield a positive semidefinite matrix with only non-negative eigenvalues. Subsequently taking the square root of a negative eigenvalue, which would occur when computing the length of an ellipse axis, produces an error.

## Way to reproduce
```python
import numpy as np
from skimage.measure import inertia_tensor_eigvals
image = np.array([[0,0],[0,0]])
mu = np.array([[3, 0, 98], [0, 14, 0], [2, 0, 98]])
eigvals = inertia_tensor_eigvals(image=image, mu=mu, T=None)
print(eigvals)
```

## Version information
```python
3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
Linux-5.3.0-46-generic-x86_64-with-Ubuntu-18.04-bionic
scikit-image version: 0.16.2
numpy version: 1.18.2
```

## Expected Results
```python
[33.33333333333333, 0.]
```

## Actual Results
```python
[33.33333333333333, -2.220446049250313e-16]
```

Leengit:
## Description

This closes #4588 "Negative eigenvalue in inertia_tensor_eigvals due to floating point precision"

## Checklist
  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_moments.py
+++ b/skimage/measure/_moments.py
@@ -445,4 +445,11 @@ def inertia_tensor_eigvals(image, mu=None, T=None):
     if T is None:
         T = inertia_tensor(image, mu)
     eigvals = np.linalg.eigvalsh(T)
+    """
+    Floating point precision problems could make a positive
+    semidefinite matrix have an eigenvalue that is very slightly
+    negative.  This can cause problems down the line, so set values
+    very near zero to zero.
+    """
+    eigvals = np.where(abs(eigvals) < 1e-12, 0, eigvals)
     return sorted(eigvals, reverse=True)
--- a/skimage/measure/_moments.py
+++ b/skimage/measure/_moments.py
@@ -445,11 +445,9 @@ def inertia_tensor_eigvals(image, mu=None, T=None):
     if T is None:
         T = inertia_tensor(image, mu)
     eigvals = np.linalg.eigvalsh(T)
-    """
-    Floating point precision problems could make a positive
-    semidefinite matrix have an eigenvalue that is very slightly
-    negative.  This can cause problems down the line, so set values
-    very near zero to zero.
-    """
-    eigvals = np.where(abs(eigvals) < 1e-12, 0, eigvals)
+    # Floating point precision problems could make a positive
+    # semidefinite matrix have an eigenvalue that is very slightly
+    # negative. This can cause problems down the line, so set values
+    # very near zero to zero.
+    eigvals = np.clip(eigvals, 0, None, out=eigvals)
     return sorted(eigvals, reverse=True)
--- a/skimage/measure/tests/test_moments.py
+++ b/skimage/measure/tests/test_moments.py
@@ -179,9 +179,9 @@ def test_inertia_tensor_eigvals():
     # Floating point precision problems could make a positive
     # semidefinite matrix have an eigenvalue that is very slightly
     # negative.  Check that we have caught and fixed this problem.
-    image = np.array([[1,0,0,0,0,0,0,0,0,0,0,0,0,0,0],
-                      [0,0,0,0,0,0,0,1,0,0,0,0,0,0,0],
-                      [0,0,0,0,0,0,0,0,0,0,0,0,0,0,1]])
+    image = np.array([[1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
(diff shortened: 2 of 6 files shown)


skimage/measure/_moments.py
```diff
@@ -445,4 +445,11 @@ def inertia_tensor_eigvals(image, mu=None, T=None):
     if T is None:
         T = inertia_tensor(image, mu)
     eigvals = np.linalg.eigvalsh(T)
+    """
+    Floating point precision problems could make a positive
+    semidefinite matrix have an eigenvalue that is very slightly
+    negative.  This can cause problems down the line, so set values
+    very near zero to zero.
+    """
```
jni, line 453:
```suggestion
    # Floating point precision problems could make a positive
    # semidefinite matrix have an eigenvalue that is very slightly
    # negative. This can cause problems down the line, so set values
    # very near zero to zero.
```

skimage/measure/_moments.py
```diff
@@ -445,4 +445,11 @@ def inertia_tensor_eigvals(image, mu=None, T=None):
...
+    eigvals = np.where(abs(eigvals) < 1e-12, 0, eigvals)
```
jni, line 454: Rather than clamping, would it not be better to clip negative eigenvalues altogether? ie `eigvals = np.clip(eigvals, 0, None)`. I don't see why `+1e-13` should be set to 0?

jni:
```suggestion
    eigvals = np.clip(eigvals, 0, None, out=eigvals)
```

Leengit: In theory, someone could supply a mu or T argument that was deliberately not originated from a positive semidefinite matrix.  In that case, significantly negative values in the response would be justified.  Whether someone is actually abusing the code in that way, I don't  know.

Leengit: Does usng the out=eigvals argument make the assignment to eigvals redundant?

skimage/measure/tests/test_moments.py
```diff
@@ -173,3 +173,15 @@ def test_inertia_tensor_3d():
...
+
+
+def test_inertia_tensor_eigvals():
+    # Floating point precision problems could make a positive
+    # semidefinite matrix have an eigenvalue that is very slightly
+    # negative.  Check that we have caught and fixed this problem.
+    image = np.array([[1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
+                      [0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0],
+                      [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]])
+    # mu = np.array([[3, 0, 98], [0, 14, 0], [2, 0, 98]])
```
emmanuelle, line 185:
```suggestion
   ```

emmanuelle: please removed commented line

jni:
Thank you for the fix @Leengit! @emmanuelle as you said, I didn't consider the comment a blocker for merging (it could even be seen as a clarifying comment), and would rather have this in ASAP!

It does seem like this might be a NumPy issue, but at the same time, I think the philosophy there is to leave these precision issues to the calling libraries. Indeed, one might argue that the fix should come from OpenBLAS or whatever underlying linear algebra library NumPy is using!

Leengit:
@emmanuelle @jni I would say that this is not a bug in the underlying np.linalg.eigsh function.  That function is intended to accept any Hermitian or real symmetric matrix, even those that are not positive definite or positive semidefinite.  That is, np.linalg.eigsh legitimately returns negative eigenvalues for some inputs.  In our test case, the input we supply to np.linalg.eigsh is intended to be positive semidefinite, but we haven't told that to np.linalg.eigsh, so I am thinking that np.linalg.eigsh can be forgiven for not knowing that and acting accordingly.

PR #4590, opened 2020-04-17
Rename convert to _convert, as it is a private function

sciunto:
## Description
I do not understand why `convert`, which is not a private function, is not exposed in the API.

I'm opened to discussion with this PR.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_transform_warp.py
+++ b/benchmarks/benchmark_transform_warp.py
@@ -1,6 +1,6 @@
 import numpy as np
 from skimage.transform import SimilarityTransform, warp
-from skimage.util.dtype import convert
+from skimage.util.dtype import _convert as convert
 import warnings
 import functools
 import inspect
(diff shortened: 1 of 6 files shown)


emmanuelle:
It would be logical to expose this function indeed. That said, its behaviour will probably change quite a lot when we drop the range rescaling in version 1.0, so I'm tempted to wait until then so that it's one less function to deprecate...

jni:
I agree with @emmanuelle, I would rather leave this semi-private for now. I would be more inclined to merge a PR renaming it to `_convert`.

sciunto:
Hopefully,  it passes the tests.

hmaarrfk:
It isn't fun to have these decisions be made without deprecations.

I would like to make a PR for a deprecation cycle to be entered.

convert is very useful when trying to convert between dtypes programatically.

emmanuelle:
@hmaarrfk the only reason a deprecation cycle was not used is because the function was not in the public API...

Issue #4597, opened 2020-04-20
`equalize_adapthist` border artifacts still present


m-albert:
## Description

The border artefacts described here https://github.com/scikit-image/scikit-image/issues/4160 are still present after fix https://github.com/scikit-image/scikit-image/pull/4349.

Actually, it seems like the added test case doesn't pick up a thin border: https://github.com/scikit-image/scikit-image/blob/9e537e6eb58ba60543820702b77cbb84dd2a8c45/skimage/exposure/tests/test_exposure.py#L365

## Way to reproduce

```python
from skimage import exposure, util, data
from skimage.color import rgb2gray

# same as test case
img = rgb2gray(util.img_as_float(data.astronaut()))
adapted = exposure.equalize_adapthist(img, kernel_size = 11)
```

```python
fig,axs = plt.subplots(1,2)
axs[0].imshow(img)
axs[1].imshow(adapted)
axs[0].set_title('input')
axs[1].set_title('CLAHE')
plt.show()
```

![image](https://user-images.githubusercontent.com/12528388/79749654-e05a8c00-830f-11ea-9ad0-1788684b7138.png)

Using e.g. `kernel_size = 13` makes the artefact stronger:

![image](https://user-images.githubusercontent.com/12528388/79749773-10a22a80-8310-11ea-9bc7-9f7dc6189161.png)

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.8.2 | packaged by conda-forge | (default, Mar 23 2020, 17:55:48)
[Clang 9.0.1 ]
macOS-10.14.6-x86_64-i386-64bit
scikit-image version: 0.17.dev0
numpy version: 1.18.2
```

PR #2761, opened 2017-08-19
Generalised existing CLAHE implementation to support n dimensions

m-albert:
## Description
CLAHE (https://en.wikipedia.org/wiki/Adaptive_histogram_equalization) is well defined for n dimensions and I extended the existing implementation in a straight forward manner.

  - only change in main function equalize_adapthist: handling of default argument
  - changed: _clahe and interpolate naturally extended to nd
  - unchanged: functions clip_histogram and map_histogram

First (and only) tests I performed:
- using the new equalize_adapthist on a 2d np.uint16 image yields the same result as the old equalize_adapthist
- new equalize_adapthist yields sensible results in 3d

## Checklist
- [ ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [x] Unit tests

## References
- the unwanted artefact reported in #2219 is still present (EDIT: fixes #2219)

## For reviewers
- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.

Issue #2219, opened 2016-07-24
equalize_adapthist leaves artefacts when image size is not a multiple of tile size


anntzer:
Python 3.5.2/Arch Linux/skimage 0.12.3

equalize_adapthist leaves "artefacts" when a dimension of the image is not a multiple of the kernel size:

```
from matplotlib import pyplot as plt
import numpy as np
from skimage.exposure import equalize_adapthist
img = (np.add.outer(np.sin(np.arange(126)), np.arange(128))
       + np.random.randint(3, size=(126, 128))).astype(int)
fig, axs = plt.subplots(1, 2)
axs[0].imshow(img, cmap="gray", interpolation="none")
axs[1].imshow(equalize_adapthist(img), cmap="gray", interpolation="none")
plt.show()
```

![adapthist](https://cloud.githubusercontent.com/assets/1322974/17085201/89f25158-5187-11e6-9b45-2d7007219cbe.png)
See the line at the top and the strip at the bottom of the second image.

Possibly related to #757?

anntzer:
FWIW I now have an implementation of CLAHE which does not use any interpolation at all.  For reasonably sized images such as this one, it is essentially instantaneous.

Would there be interest in merging it?

soupault:
I'd definitely be glad to have a look.

anntzer:
See https://github.com/anntzer/clahe.  The implementation should be relatively transparent: we uniquify the values to make histogramming easier, and progressively update the histogram, and compute the CLAHE values.  The only slightly tricky part is the contrast-limiting itself.
Edges are handled by padding with the border value.

JDWarner:
@anntzer that looks rather elegant!

anntzer:
I'll let you discuss the API changes :-)

Note that I use the "standard" definition of clip limit (>1, instead of <1 as is currently the case), see https://github.com/scikit-image/scikit-image/issues/2214#issuecomment-234841754.
For simplicity (mostly) and performance (less so), I only do one round of clipping, instead of repeating it until the contrast is actually limited by the specified value -- IMO there isn't much benefit in actually enforcing the exact clip limit.
Also I gave a try to "multiplicative" instead of "additive" redistribution of the clipped values (both are available), not an expert there but the original paper suggests that the latter (which is what's used in general) works better.

b52:
What's the state on this one? This is a really annoying bug and it would be nice to see it fixed sooner rather than later, especially when there is a working code base available.

[...]

PR #4598, opened 2020-04-20
Extended existing `equalize_adapthist` implementation (CLAHE) to N dimensions

m-albert:
## Description

This PR naturally extends the **existing** implementation of `equalize_adapthist` to arbitrary dimensions.

This is a follow up of https://github.com/scikit-image/scikit-image/pull/2761 (chose to open a new PR instead of rebasing old feature branch) and it fixes https://github.com/scikit-image/scikit-image/issues/4597 (image border artefacts, see also https://github.com/scikit-image/scikit-image/issues/2219#issuecomment-516526979).

Summary of PR code changes:
- small change in main function `equalize_adapthist`
- functions `_clahe` and `interpolate` naturally extended to nd
- unchanged: functions `clip_histogram` and `map_histogram`

Tests:
  - added a test to check successful extension to 3d:
    - show that N=3 is processed as such
    - show that the central plane of a processed 3d stack of a given 2d image is similar to the processed 2d image
  - changed the target value of an existing assert due to the mentioned artifact not occurring anymore
  - all other existing tests pass (showing the new implementation just extends the old one)
  - modified existing testing of border artefacts (which was not catching still existing border artefacts https://github.com/scikit-image/scikit-image/issues/4597) and extended it to check a variety of kernel_sizes

Example output in 3D using `napari`:

```python
from skimage import exposure
import numpy as np
import napari
from scipy import ndimage

# create 3d data
a = 5
np.random.seed(100)
im = np.random.randint(0, 1000, (a, a, a)).astype(np.float64) / 1000.
im = ndimage.zoom(im, 50, order=2)

# apply uneven intensity gradient
sigmoid = 1/4 + 1/2 * (1 / (1 + np.exp(3 * np.linspace(1, 0, im.shape[0]) - 0.5)))
im *= sigmoid

# recover local contrast with CLAHE
imc = exposure.equalize_adapthist(im, 55, clip_limit=0.1)

with napari.gui_qt():
    viewer = napari.Viewer(ndisplay=3)
    viewer.theme = 'light'

    # add orig and processed volumes
    gamma = 1.5
    viewer.add_image(im, name='orig', gamma=gamma, colormap='viridis')
    viewer.add_image(imc, name='clahe', gamma=gamma, colormap='viridis')

    # draw bounding box around data
    verts = np.array([[i, j, k] for i in [0,1] for j in [0,1] for k in [0, 1]]).astype(np.float32)
    lines = [np.array([i, j])*im.shape[0] for i in verts for j in verts if np.allclose(np.linalg.norm(i - j), 1)]
    viewer.add_shapes(lines[:], shape_type='line', edge_color='gray')
```

original image             |  3D `equalize_adapthist`
:-------------------------:|:-------------------------:
  |  

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -65,6 +65,10 @@ def equalize_adapthist(image, kernel_size=None,
        - The image is converted back to RGB space and returned
     * For RGBA images, the original alpha channel is removed.
 
+    .. versionchanged:: 0.17
+        The values returned by this function are slightly shifted upwards
+        because of an internal change in rounding behavior.
+
     References
     ----------
     .. [1] http://tog.acm.org/resources/GraphicsGems/
@@ -75,7 +79,9 @@ def equalize_adapthist(image, kernel_size=None,
         return img_as_float(image)  # convert to float for consistency
 
     image = img_as_uint(image)
-    image = rescale_intensity(image, out_range=(0, NR_OF_GRAY - 1))
+    image = np.round(
+        rescale_intensity(image, out_range=(0, NR_OF_GRAY - 1))
+    ).astype(np.uint16)
 
     if kernel_size is None:
         kernel_size = (image.shape[0] // 8, image.shape[1] // 8)
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -150,12 +150,12 @@ def _clahe(image, kernel_size, clip_limit, nbins):
     # Calculate graylevel mappings for each contextual region
     for inds in np.ndindex(*ns):
 
-        region = tuple([slice(inds[dim] * steps[dim], (inds[dim] + 1) *
-                              steps[dim])
+        region = tuple([slice(inds[dim] * steps[dim], (inds[dim] + 1)
+                              * steps[dim])
                         for dim in range(image.ndim)])
         sub_img = image[region]
 
-        if clip_limit > 0.0:  # Calculate actual cliplimit
+        if clip_limit > 0.0:  # Calculate actual clip limit
             clim = int(clip_limit * sub_img.size / nbins)
             if clim < 1:
                 clim = 1
@@ -169,9 +169,7 @@ def _clahe(image, kernel_size, clip_limit, nbins):
         hist = map_histogram(hist, 0, NR_OF_GRAY - 1, sub_img.size)
         map_array[inds] = hist
 
-
     # Interpolate greylevel mappings to get CLAHE image
-
     offsets = [0] * image.ndim
     lowers = [0] * image.ndim
     uppers = [0] * image.ndim
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -180,7 +180,7 @@ def _clahe(image, kernel_size, clip_limit, nbins):
                   for i, s in zip(inds, steps)]
 
         offsets = [np.ceil(s / 2.) if not i else s
-                   for i, s, n in zip(inds, steps, ns)]
+                   for i, s in zip(inds, steps)]
 
         slices = [slice(int(st), int(min([st + o, sh])))
                   for st, o, sh in zip(starts, offsets, image.shape)]
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -219,14 +219,15 @@ def _clahe(image, kernel_size, clip_limit, nbins):
     # rebuild result image from blocks
     result = result.reshape(blocks_flattened_shape)
     blocks_axis_rebuild_order =\
-        np.array([np.arange(0,    ndim),
+        np.array([np.arange(0, ndim),
                   np.arange(ndim, ndim * 2)]).T.flatten()
     result = np.transpose(result, axes=blocks_axis_rebuild_order)
     result = result.reshape(image.shape)
 
     # undo padding
     unpad_slices = tuple([slice(p_i, s - p_f) for p_i, p_f, s in
-                          zip(pad_start_per_dim, pad_end_per_dim, image.shape)])
+                          zip(pad_start_per_dim, pad_end_per_dim,
+                              image.shape)])
     result = result[unpad_slices]
 
     return result
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -34,9 +34,9 @@ def equalize_adapthist(image, kernel_size=None,
 
     Parameters
     ----------
-    image : (M, N[, C]) ndarray
+    image : (N1, ...,NN[, C]) ndarray
         Input image.
-    kernel_size: integer or array_like, optional
+    kernel_size: int or array_like, optional
         Defines the shape of contextual regions used in the algorithm. If
         iterable is passed, it must have the same number of elements as
         ``image.ndim`` (without color channel). If integer, it is broadcasted
@@ -50,7 +50,7 @@ def equalize_adapthist(image, kernel_size=None,
 
     Returns
     -------
-    out : (M, N[, C]) ndarray
+    out : (N1, ...,NN[, C]) ndarray
         Equalized image with float64 dtype.
 
     See Also
@@ -84,7 +84,8 @@ def equalize_adapthist(image, kernel_size=None,
     ).astype(np.uint16)
 
     if kernel_size is None:
-        kernel_size = (image.shape[0] // 8, image.shape[1] // 8)
+        kernel_size = tuple([image.shape[dim] // 8
+                             for dim in range(image.ndim)])
     elif isinstance(kernel_size, numbers.Number):
         kernel_size = (kernel_size,) * image.ndim
     elif len(kernel_size) != image.ndim:
@@ -92,7 +93,7 @@ def equalize_adapthist(image, kernel_size=None,
 
     kernel_size = [int(k) for k in kernel_size]
 
-    image = _clahe(image, kernel_size, clip_limit, nbins)
+    image = _clahe(image, kernel_size, clip_limit * nbins, nbins)
     image = img_as_float(image)
     return rescale_intensity(image)
 
@@ -102,9 +103,9 @@ def _clahe(image, kernel_size, clip_limit, nbins):
 
     Parameters
     ----------
-    image : (M, N) ndarray
+    image : (N1,...,NN) ndarray
         Input image.
-    kernel_size: 2-tuple of int
+    kernel_size: int or N-tuple of int
         Defines the shape of contextual regions used in the algorithm.
     clip_limit : float
         Normalized clipping limit (higher values give more contrast).
@@ -113,7 +114,7 @@ def _clahe(image, kernel_size, clip_limit, nbins):
 
     Returns
     -------
-    out : (M, N) ndarray
+    out : (N1,...,NN) ndarray
         Equalized image.
 
     The number of "effective" graylevels in the output image is set by `nbins`;
@@ -122,74 +123,106 @@ def _clahe(image, kernel_size, clip_limit, nbins):
     minimum and maximum value as the input image. A clip limit smaller than 1
     results in standard (non-contrast limited) AHE.
     """
-    row_step = int(image.shape[0] // np.ceil(image.shape[0] / kernel_size[0]))
-    col_step = int(image.shape[1] // np.ceil(image.shape[1] / kernel_size[1]))
+    # pad the image so that the shape in each dimension
+    # is a multiple of the relevant kernel_size
+    pad_end_per_dim = [0] * image.ndim
+    for dim in range(image.ndim):
+        while (image.shape[dim] + pad_end_per_dim[dim]) % kernel_size[dim]:
+            pad_end_per_dim[dim] += 1
 
-    nr = image.shape[0] // row_step
-    nc = image.shape[1] // col_step
+    image = np.pad(image,
+                   [[0, pad_end_per_dim[dim]] for dim in range(image.ndim)],
+                   mode='reflect',
+                   )
+
+    ns = [int(np.ceil(image.shape[dim] / kernel_size[dim]))
+          for dim in range(image.ndim)]
+
+    steps = [int(np.floor(image.shape[dim] / ns[dim]))
+             for dim in range(image.ndim)]
 
     bin_size = 1 + NR_OF_GRAY // nbins
     lut = np.arange(NR_OF_GRAY)
     lut //= bin_size
 
-    map_array = np.zeros((nr, nc, nbins), dtype=int)
+    map_array = np.zeros(tuple(ns) + (nbins,), dtype=int)
 
     # Calculate graylevel mappings for each contextual region
-    i0 = 0
-    for r in range(nr):
-        i1 = i0 + row_step
-        j0 = 0
-        for c in range(nc):
-            j1 = j0 + col_step
-            sub_img = image[i0:i1, j0:j1]
+    for inds in np.ndindex(*ns):
 
-            if clip_limit > 0.0:  # Calculate actual cliplimit
-                clim = max(int(clip_limit * sub_img.size), 1)
+        region = tuple([slice(inds[dim] * steps[dim], (inds[dim] + 1) *
+                              steps[dim])
+                        for dim in range(image.ndim)])
+        sub_img = image[region]
+
+        if clip_limit > 0.0:  # Calculate actual cliplimit
+            clim = int(clip_limit * sub_img.size / nbins)
+            if clim < 1:
+                clim = 1
+        else:
+            clim = NR_OF_GRAY  # Large value, do not clip (AHE)
+
+        hist = lut[sub_img.ravel()]
+        hist = np.bincount(hist)
+        hist = np.append(hist, np.zeros(nbins - hist.size, dtype=int))
+        hist = clip_histogram(hist, clim)
+        hist = map_histogram(hist, 0, NR_OF_GRAY - 1, sub_img.size)
+        map_array[inds] = hist
+
+
+    # Interpolate greylevel mappings to get CLAHE image
+
+    offsets = [0] * image.ndim
+    lowers = [0] * image.ndim
+    uppers = [0] * image.ndim
+    starts = [0] * image.ndim
+    prev_inds = [0] * image.ndim
+
+    for inds in np.ndindex(*[ns[dim] + 1 for dim in range(image.ndim)]):
+
+        for dim in range(image.ndim):
+            if inds[dim] != prev_inds[dim]:
+                starts[dim] += offsets[dim]
+
+        for dim in range(image.ndim):
+            if dim < image.ndim - 1:
+                if inds[dim] != prev_inds[dim]:
+                    starts[dim + 1] = 0
+
+        prev_inds = inds[:]
+
+        # modify edges to handle special cases
+        for dim in range(image.ndim):
+            if inds[dim] == 0:
+                offsets[dim] = np.ceil(steps[dim] / 2.0)
+                lowers[dim] = 0
+                uppers[dim] = 0
+            elif inds[dim] == ns[dim]:
+                offsets[dim] = np.ceil(steps[dim] / 2.0)
+                lowers[dim] = ns[dim] - 1
+                uppers[dim] = ns[dim] - 1
             else:
-                clim = NR_OF_GRAY  # Large value, do not clip (AHE)
+                offsets[dim] = steps[dim]
+                lowers[dim] = inds[dim] - 1
+                uppers[dim] = inds[dim]
 
-            hist = np.bincount(lut[sub_img.ravel()], minlength=nbins)
-            hist = clip_histogram(hist, clim)
-            hist = map_histogram(hist, 0, NR_OF_GRAY - 1, sub_img.size)
-            map_array[r, c] = hist
-            j0 = j1
-        i0 = i1
+        maps = []
+        for edge in np.ndindex(*([2] * image.ndim)):
+            maps.append(map_array[tuple([[lowers, uppers][edge[dim]][dim]
+                                         for dim in range(image.ndim)])])
 
-    # Interpolate graylevel mappings to get CLAHE image
-    rstart = 0
-    for r in range(nr + 1):
-        cstart = 0
-        rU = max(0, r - 1)
-        rB = min(nr - 1, r)
-        if r in [0, nr]:  # special case: top and bottom rows
-            r_offset = row_step // 2
-        else:  # default values
-            r_offset = row_step
+        slices = [np.arange(starts[dim],
+                            np.min([starts[dim] + offsets[dim],
+                                    image.shape[dim]]))
+                  for dim in range(image.ndim)]
 
-        rslice = slice(rstart, rstart + r_offset)
+        interpolate(image, slices[::-1], maps, lut)
 
-        for c in range(nc + 1):
-            cL = max(0, c - 1)
-            cR = min(nc - 1, c)
-            if c in [0, nc]:  # special case: left and right columns
-                c_offset = col_step // 2
-            else:  # default values
-                c_offset = col_step
+    # undo padding
+    depad_slices = tuple([slice(0, image.shape[dim] - pad_end_per_dim[dim])
+                          for dim in range(image.ndim)])
 
-            cslice = slice(cstart, cstart + c_offset)
-
-            mapLU = map_array[rU, cL]
-            mapRU = map_array[rU, cR]
-            mapLB = map_array[rB, cL]
-            mapRB = map_array[rB, cR]
-
-            interpolate(image, cslice, rslice,
-                        mapLU, mapRU, mapLB, mapRB, lut)
-
-            cstart += c_offset  # set pointer on next matrix */
-
-        rstart += r_offset
-
+    image = image[depad_slices]
     return image
 
 
@@ -275,17 +308,16 @@ def map_histogram(hist, min_val, max_val, n_pixels):
     return out.astype(int)
 
 
-def interpolate(image, xslice, yslice,
-                mapLU, mapRU, mapLB, mapRB, lut):
+def interpolate(image, slices, maps, lut):
     """Find the new grayscale level for a region using bilinear interpolation.
 
     Parameters
     ----------
     image : ndarray
         Full image.
-    xslice, yslice : slice
-       Slices of the region.
-    map* : ndarray
+    slices : list of array-like
+       Indices of the region.
+    maps : list of ndarray
         Mappings of graylevels from histograms.
     lut : ndarray
         Maps grayscale levels in image to histogram levels.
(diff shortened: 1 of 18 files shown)


VincentStimper:
Nice work, @m-albert! However, there are two main issues in your implementation. The first thing is the padding. It does not make sense to only pad the image at one side since this would treat image borders asymmetrically. Furthermore, the padding should be more generous to simplify the interpolation process. This issue is discussed in [this paper](https://ieeexplore.ieee.org/document/8895993). Secondly, transforming the image by doing a for loop over the kernel blocks in the image is inefficient. The transformation can be written as array operations which can be run in parallel with numpys native functions. I'll submit a PR based on this one once it is through.

m-albert:
@VincentStimper Thanks for your input :)

For this PR I tried to stick to the already existing implementation of CLAHE to
1) avoid introducing breaking changes
2) reuse existing tested code

> It does not make sense to only pad the image at one side since this would treat image borders asymmetrically. Furthermore, the padding should be more generous to simplify the interpolation process.

This would introduce a breaking change w.r.t. the existing 2D implementation. I agree however that a symmetrical treatment of the borders makes sense. Of course this could be avoided by calling different implementations for `n=2` and `n>2`, however that would have the disadvantage of inconsistency. It might be that `scikit-image` has a guideline or good practice statement on how to deal with this sort of situations (didn't find yet).

> The transformation can be written as array operations which can be run in parallel with numpys native functions.

Again the idea was to stick to the existing implementation. However I'm sure the code can be optimised and squeezing most out of efficient numpy vectorizations, ufuncs etc. sounds great. Do you have suggestions for non-breaking optimizations?

skimage/exposure/_adapthist.py
```diff
@@ -116,74 +117,104 @@ def _clahe(image, kernel_size, clip_limit, nbins):
     minimum and maximum value as the input image. A clip limit smaller than 1
     results in standard (non-contrast limited) AHE.
     """
-    row_step = int(image.shape[0] // np.ceil(image.shape[0] / kernel_size[0]))
-    col_step = int(image.shape[1] // np.ceil(image.shape[1] / kernel_size[1]))
+    # pad the image so that the shape in each dimension
+    # is a multiple of the relevant kernel_size
+    pad_end_per_dim = [0] * image.ndim
+    for dim in range(image.ndim):
+        while (image.shape[dim] + pad_end_per_dim[dim]) % kernel_size[dim]:
```
emmanuelle, line 124: you can compute this without the while loop I think: 
```
pad_end_per_dim[dim] = (kernel_size[dim] - image.shape[dim] % kernel_size[dim]) %kernel_size[dim]
```
not tested but it should work :-)

skimage/exposure/tests/test_exposure.py
```diff
@@ -324,7 +324,7 @@ def test_adapthist_grayscale():
     adapted = exposure.equalize_adapthist(img, kernel_size=(57, 51),
                                           clip_limit=0.01, nbins=128)
     assert img.shape == adapted.shape
-    assert_almost_equal(peak_snr(img, adapted), 102.078, 3)
+    assert_almost_equal(peak_snr(img, adapted), 100.158, 3)
```
emmanuelle, line 327: why would the value change here?

skimage/exposure/tests/test_exposure.py
```diff
@@ -362,16 +362,56 @@ def test_adapthist_alpha():
...
+    # maximize difference between orig and processed img
+    img /= 100.
+    img[img.shape[0] // 2, img.shape[1] // 2] = 1.
+
+    # check borders are processed for different kernel sizes
+    border_index = -1
+    for kernel_size in range(51, 71, 2):
+        adapted = exposure.equalize_adapthist(img, kernel_size, clip_limit=0.5)
+        # Check last columns are processed
+        assert norm_brightness_err(adapted[:, border_index],
```
emmanuelle, line 410: I don't understand these tests here. Shouldn't we compare the same columns on the two images? I wonder if there was a bug here (not introduced by your PR, but let's fix it if it is indeed a bug).

skimage/exposure/_adapthist.py
m-albert: You're right of course!

skimage/exposure/tests/test_exposure.py
m-albert: As mentioned in the description, the PR fixes border artefacts which are gone here but present in master. This would be reflected in this value no?

skimage/exposure/tests/test_exposure.py
m-albert: You're completely right! I think when I added the minus to the border_index value I forgot to take it away from the index into the original image. Good catch!

m-albert: There was a problem in this test before, however it was not the comparison of different columns (which I had introduced). Namely that column -42 was compared (together with a slightly misleading `width` variable) and that hadn't catched the remaining border artefact in some of the last columns.

skimage/exposure/tests/test_exposure.py
m-albert: By contrast, in the test `test_adapthist_alpha`, where the image size is a multiple of the kernel size and both the nd branch and master don't exhibit artefacts, values stay the same.

emmanuelle:
@VincentStimper you are completely right regarding the padding. Probably we'll merge this PR without fixing this bug in the interest of adding the nd feature without modifying the existing behaviour, but we'll need to fix this at some point. Could you please open an issue explaining what should be done?

skimage/exposure/_adapthist.py
```diff
@@ -122,74 +123,103 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+
+        if clip_limit > 0.0:  # Calculate actual clip limit
+            clim = int(clip_limit * sub_img.size)
+            if clim < 1:
+                clim = 1
+        else:
+            clim = NR_OF_GRAY  # Large value, do not clip (AHE)
+
+        hist = lut[sub_img.ravel()]
+        hist = np.bincount(hist)
```
emmanuelle, line 165: why did you change the existing code `np.bincount(hist, minlength=nbins)` ? It looked ok to me.

emmanuelle:
Our tests are not very demanding for this function, so could you please execute the example of https://scikit-image.org/docs/dev/auto_examples/color_exposure/plot_equalize.html#sphx-glr-auto-examples-color-exposure-plot-equalize-py on your branch and paste here the image so that we can check visually that nothing changed?

skimage/exposure/_adapthist.py
```diff
@@ -122,74 +123,103 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+        map_array[inds] = hist
+
+    # Interpolate greylevel mappings to get CLAHE image
+    offsets = [0] * image.ndim
+    lowers = [0] * image.ndim
+    uppers = [0] * image.ndim
+    starts = [0] * image.ndim
+    prev_inds = [0] * image.ndim
+
+    for inds in np.ndindex(*[ns[dim] + 1 for dim in range(image.ndim)]):
```
emmanuelle, line 178: So if I understand well `ns[dim]` is the number of cells / blocks for a given dimension. Do you really need to add 1 to this number? If yes please explain :-).

skimage/exposure/_adapthist.py
```diff
@@ -122,74 +123,103 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+
+        for dim in range(image.ndim):
+            if inds[dim] != prev_inds[dim]:
+                starts[dim] += offsets[dim]
+
+        for dim in range(image.ndim):
```
emmanuelle, line 184: could you please add an inline comment explaining why this block is needed? I don't understand.

emmanuelle:
Thank you for your PR @m-albert ! Today I had more time to review it. I asked for clarifications in places but globally it looks good to me.

skimage/exposure/_adapthist.py
```diff
@@ -298,19 +327,35 @@ def interpolate(image, xslice, yslice,
...
-                       + x_coef * mapRB[im_slice]))
-    view[:, :] = new / (x_size * y_size)
+
+    new = np.zeros_like(view, dtype=int)
+    for iedge, edge in enumerate(np.ndindex(*([2] * image.ndim))):
+        edge = edge[::-1]
+        new += \
+            np.product([[inv_coeffs, coeffs][edge[dim]][dim]
+                        for dim in range(image.ndim)], 0)\
+            * maps[iedge][im_slice]
```
rfezzani, line 357:
```suggestion
        new += (np.product([[inv_coeffs, coeffs][e][dim]
                            for dim, e in enumerate(edge)], 0)
                * maps[iedge][im_slice])
```

skimage/exposure/_adapthist.py
m-albert: You're right, the existing line here is better and should stay! It's a change from three months ago (https://github.com/scikit-image/scikit-image/commit/834f8b6701ec55394f0e48d59a1524a4bbada14a) which happened after I had copied the code to adapt it. This was actually some years ago already and recently I went through new commits but seem to have missed that one. Just changed the line back to master. Thanks!

skimage/exposure/_adapthist.py
```diff
@@ -122,74 +123,103 @@ def _clahe(image, kernel_size, clip_limit, nbins):
     minimum and maximum value as the input image. A clip limit smaller than 1
     results in standard (non-contrast limited) AHE.
     """
-    row_step = int(image.shape[0] // np.ceil(image.shape[0] / kernel_size[0]))
-    col_step = int(image.shape[1] // np.ceil(image.shape[1] / kernel_size[1]))
+    # pad the image so that the shape in each dimension
+    # is a multiple of the relevant kernel_size
+    pad_end_per_dim = [0] * image.ndim
+    for dim in range(image.ndim):
+        pad_end_per_dim[dim] = (kernel_size[dim] - image.shape[dim] % kernel_size[dim]) % kernel_size[dim]
```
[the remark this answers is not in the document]
m-albert: Thanks for these suggestions on the use of list comprehensions, I incorporated them all and marked the relevant conversations as resolved.

skimage/exposure/_adapthist.py
```diff
@@ -122,74 +123,103 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+        if clip_limit > 0.0:  # Calculate actual clip limit
+            clim = int(clip_limit * sub_img.size)
+            if clim < 1:
+                clim = 1
+        else:
+            clim = NR_OF_GRAY  # Large value, do not clip (AHE)
+
+        hist = lut[sub_img.ravel()]
+        hist = np.bincount(hist)
+        hist = np.append(hist, np.zeros(nbins - hist.size, dtype=int))
```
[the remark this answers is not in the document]
m-albert: Agreed! As commented below the comment of @emmanuelle, this change had been made in master which I accidentally hadn't carried along.

m-albert:
> Our tests are not very demanding for this function, so could you please execute the example of https://scikit-image.org/docs/dev/auto_examples/color_exposure/plot_equalize.html#sphx-glr-auto-examples-color-exposure-plot-equalize-py on your branch and paste here the image so that we can check visually that nothing changed?

master             |  PR branch
:-------------------------:|:-------------------------:
  |

skimage/exposure/_adapthist.py
```diff
@@ -298,19 +327,35 @@ def interpolate(image, xslice, yslice,
...
-    a submatrix of the image. This is done by a bilinear interpolation between
-    four different mappings in order to eliminate boundary artifacts.
+    a submatrix of the image. This is done by bilinear interpolation between
+    2^image.ndim different adjacent mappings in order to eliminate boundary
+    artifacts.
     """
-    view = image[yslice, xslice]
-    y_size, x_size = view.shape
+    norm = np.product([slices[dim].size
+                       for dim in range(image.ndim)])  # Normalization factor
```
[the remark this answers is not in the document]
m-albert: Good point. Changed from passing arrays to passing slices to `interpolate` now, as this should be clearer and more efficient. Should also be clear from the docstring.

skimage/exposure/_adapthist.py
m-albert: Thanks for the feedback, this part of the function was very hard to understand. I rewrote it in a hopefully much clearer way.

skimage/exposure/_adapthist.py
```diff
@@ -162,64 +157,50 @@ def _clahe(image, kernel_size, clip_limit, nbins):
             clim = NR_OF_GRAY  # Large value, do not clip (AHE)
 
         hist = lut[sub_img.ravel()]
-        hist = np.bincount(hist)
+        hist = np.bincount(hist, minlength=nbins)
         hist = np.append(hist, np.zeros(nbins - hist.size, dtype=int))
```
rfezzani, line 161: This line is no more needed since `hist.size == nbins` now ;-)
```suggestion
```

skimage/exposure/_adapthist.py
```diff
@@ -162,64 +157,50 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+    # determine num of blocks to be processed separately in each dim:
+    # generally n+1, n if last block only processes padded pixels
+    ns_proc = [n + 1 if (np.ceil(s / 2.) + (n - 1) * s) <= (sh - pe)
+               else n for sh, pe, s, n
+               in zip(image.shape, pad_end_per_dim, steps, ns)]
+
+    for inds in np.ndindex(*ns_proc):
+
+        # define slices for each dim
+        starts = [int(i > 0) * np.ceil(s / 2.) + np.max([0, i - 1]) * s
```
rfezzani, line 180: Am I wrong or `i` can't be negative?
```suggestion
        starts = [np.ceil(s / 2.) + (i - 1) * s if i > 0 else 0
```

skimage/exposure/_adapthist.py
```diff
@@ -162,64 +157,50 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+                  for i, s in zip(inds, steps)]
+
+        offsets = [np.ceil(s / 2.) if not i else s
+                   for i, s, n in zip(inds, steps, ns)]
+
+        slices = [slice(int(st), int(np.min([st + o, sh])))
```
rfezzani, line 186:
```suggestion
        slices = [slice(int(st), int(min(st + o, sh)))
```

skimage/exposure/_adapthist.py
```diff
@@ -162,64 +157,50 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+                  for st, o, sh in zip(starts, offsets, image.shape)]
+
+        # define neighboring contextual regions
+        lowers = [np.max([0, i - 1]) for i in inds]
```
rfezzani, line 190:
```suggestion
        lowers = [max(0, i - 1) for i in inds]
```

skimage/exposure/_adapthist.py
```diff
@@ -162,64 +157,50 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+        uppers = [np.min([n - 1, i]) for i, n in zip(inds, ns)]
```
rfezzani, line 191:
```suggestion
        uppers = [min(n - 1, i) for i, n in zip(inds, ns)]
```

skimage/exposure/_adapthist.py
m-albert: The need for `n+1` is indeed not exactly straight forward: As you say, `ns[dim]` is the number of contextual regions for each dim (to calculate graylevel mappings). The number of blocks to process separately for interpolation is however different from this number in the general case, as becomes clear from [this](https://en.wikipedia.org/w/index.php?title=Adaptive_histogram_equalization&oldid=936814673#Efficient_computation_by_interpolation) wikipedia schematic (which I also linked in a code comment to clarify). Depending on how many pixels are padded at the end of the dim (to have the contextual regions well defined), `ns[dim]` blocks can be sufficient to process all pixels within the shape of the input image. Otherwise `ns[dim]+1` blocks are needed.

The 2d version in `master` always processes `ns[dim]+1` blocks, however thanks to your comment I went through it and realised the last block can in many cases be dropped (as it only processes pixels which will be discarded during unpadding. Thanks :) Let me know if you agree and what you think!

skimage/exposure/_adapthist.py
```diff
@@ -122,74 +123,103 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+    for inds in np.ndindex(*[ns[dim] + 1 for dim in range(image.ndim)]):
+
+        for dim in range(image.ndim):
+            if inds[dim] != prev_inds[dim]:
+                starts[dim] += offsets[dim]
+
+        for dim in range(image.ndim):
+            if dim < image.ndim - 1:
+                if inds[dim] != prev_inds[dim]:
+                    starts[dim + 1] = 0
```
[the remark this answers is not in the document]
m-albert: Clearly too convoluted, rewrote this part. Thanks!

skimage/exposure/_adapthist.py
```diff
@@ -326,36 +308,32 @@ def interpolate(image, slices, maps, lut):
...
+    region = tuple([s for s in slices])
+    view = image[region]
 
     # interpolation weight matrices
-    coeffs = np.meshgrid(*tuple([np.arange(slices[dim].size)
-                                 for dim in range(image.ndim)]), indexing='ij')
-    coeffs = [coeff.transpose() for coeff in coeffs]
+    coeffs = np.meshgrid(*tuple([np.arange(s) for s in view.shape[::-1]]),
+                         indexing='ij')
+    coeffs = [np.transpose(c) for c in coeffs]
```
rfezzani, line 322: I think it is equivalent to :
```suggestion
    coeffs = np.meshgrid(*tuple([np.arange(s) for s in view.shape[::-1]]),
                         indexing='xy')
```

skimage/exposure/_adapthist.py
```diff
@@ -326,36 +308,32 @@ def interpolate(image, slices, maps, lut):
...
 
-    inv_coeffs = [np.flip(coeffs[dim], axis=image.ndim - dim - 1) + 1
-                  for dim in range(image.ndim)]
-
-    region = tuple([slice(int(slices[dim][0]), int(slices[dim][-1] + 1))
-                    for dim in range(image.ndim)][::-1])
-    view = image[region]
+    inv_coeffs = [np.flip(c, axis=image.ndim - dim - 1) + 1
```
rfezzani, line 324:
```suggestion
    inv_coeffs = [np.flip(c, axis=-dim) + 1
```

skimage/exposure/_adapthist.py
```diff
@@ -326,36 +308,32 @@ def interpolate(image, slices, maps, lut):
...
-            * maps[iedge][im_slice]
-
-    new = (new / norm).astype(view.dtype)
-    view[::] = new
+        result += (np.product([[inv_coeffs, coeffs][e][dim]
+                               for dim, e in enumerate(edge[::-1])], 0)
+                   * maps[iedge][im_slice])
+
+    # normalize
+    result = result / np.product(view.shape)
```
rfezzani, line 336:
```suggestion
    result /= view.size
```

skimage/exposure/_adapthist.py
```diff
@@ -306,13 +287,14 @@ def map_histogram(hist, min_val, max_val, n_pixels):
...
-    """Find the new grayscale level for a region using bilinear interpolation.
+    """Find the new grayscale level for a region
+    using multilinear interpolation.
 
     Parameters
     ----------
     image : ndarray
         Full image.
-    slices : list of array-like
+    slices : list of slices
```
rfezzani, line 297: can't this variable be renamed to `slices_list` for clarity?

skimage/exposure/_adapthist.py
m-albert: True, the conditional statement is better than the awkward casting to `int` :)

skimage/exposure/_adapthist.py
m-albert: Hmm I'm a bit confused about this. I get an error when using the `xy` line. https://docs.scipy.org/doc/numpy/reference/generated/numpy.meshgrid.html says:

> In the 2-D case with inputs of length M and N, the outputs are of shape (N, M) for ‘xy’ indexing and (M, N) for ‘ij’ indexing. In the 3-D case with inputs of length M, N and P, outputs are of shape (N, M, P) for ‘xy’ indexing and (M, N, P) for ‘ij’ indexing.

Seems like the `ij` indexing might be better for the general n-dim case?

skimage/exposure/_adapthist.py
m-albert: Should be `...axis=-dim - 1)` no?

skimage/exposure/_adapthist.py
rfezzani: You are right, I only tested for 2D case, sorry :pray:

skimage/exposure/_adapthist.py
m-albert: Wasn't aware of this attribute! Let's assume I'm the only one :P

skimage/exposure/_adapthist.py
m-albert: Done.

skimage/exposure/_adapthist.py
rfezzani: Looks much better indeed ;)

m-albert:
Thanks @emmanuelle and @rfezzani for the code review and useful suggestions and comments! It's such a fun experience to get targeted feedback :)

It became clear that parts of the code were difficult to read and very convoluted. I restructured the parts you commented on quite a bit and think it's more concise now :)

Be aware of two further changes I made and let me know if you should disagree:

First, thanks to @emmanuelle's comment I realised in some cases the processing of the final blocks is exclusively affecting padded pixels. Therefore I changed the way the blocks are iterated (see code below) which should make processing slightly faster in some cases compared to `master`, see more details in this comment https://github.com/scikit-image/scikit-image/pull/4598#discussion_r416194634
```python
    # determine num of blocks to be processed separately in each dim:
    # generally n+1, n if last block only processes padded pixels
    ns_proc = [n + 1 if (np.ceil(s / 2.) + (n - 1) * s) <= (sh - pe)
               else n for sh, pe, s, n
               in zip(image.shape, pad_end_per_dim, steps, ns)]

    for inds in np.ndindex(*ns_proc):
```

Second (somewhat related), thanks to a comment by @rfezzani I switched from passing numpy arrays to passing slices to `interpolate` (as also done in 2d `master`). As there is no problem with slices overshooting the array they index, it becomes redundant to restrict the processed slice in case of `inds[dim]`==`ns[dim]` to half the step width. Therefore I removed the special treatment in this case (2d `master` does this differently).
```python
        # offsets = [np.ceil(s / 2.) if i in [0, n]  else s
        #            for i, s, n in zip(inds, steps, ns)]
        offsets = [np.ceil(s / 2.) if not i else s
                   for i, s in zip(inds, steps)]
```

Okay, thought it cannot harm to lay out these two changes here in case somebody has comments :)

skimage/exposure/tests/test_exposure.py
```diff
@@ -370,7 +370,7 @@ def test_adapthist_grayscale():
     adapted = exposure.equalize_adapthist(img, kernel_size=(57, 51),
                                           clip_limit=0.01, nbins=128)
     assert img.shape == adapted.shape
-    assert_almost_equal(peak_snr(img, adapted), 102.019, 3)
+    assert_almost_equal(peak_snr(img, adapted), 100.140, 3)
```
jni, line 373: This seems like a major drop... Do you know why there is a difference?

jni:
Sorry, I missed the earlier reviews, and went through the code blind to the earlier comments. It's funny, I noticed the asymmetrical padding but decided to ignore it since it was there already in the 2D implementation. Typically, yes, we branch out the code in 2D and 3D cases if we think there is a significant advantage to the new method. However, in this case I thought that with reflect boundary handling the difference would be minimal in most cases.

Personally, I would be ok with adding symmetric padding and having a breaking change. My understanding was that the asymmetric padding was due to a reference implementation, but if the method itself doesn't prescribe that, then I would be happy to treat it as a bug and just update it. What do others at @scikit-image/core think?

At any rate, I think the code is in a good spot now, thank you to everyone involved in the iterations so far. Like @VincentStimper there were a couple of places where I thought vectorisation might help, but I don't think it's necessary to get this PR in, especially since the number of blocks is much smaller than the number of pixels.

Thanks again!

skimage/exposure/tests/test_exposure.py
emmanuelle: @jni I made the same comment, I copy here Marvin's answer (which is one of the "resolved" blocks): "As mentioned in the description, the PR fixes border artefacts which are gone here but present in master. This would be reflected in this value no? By contrast, in the test test_adapthist_alpha, where the image size is a multiple of the kernel size and both the nd branch and master don't exhibit artefacts, values stay the same."

m-albert: Thanks a lot @emmanuelle for forwarding the comment :)

skimage/exposure/_adapthist.py
```diff
@@ -122,73 +123,82 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
+
+        region = tuple([slice(i * s, (i + 1) * s)
+                        for i, s in zip(inds, steps)])
+
+        sub_img = image[region]
+
+        if clip_limit > 0.0:  # Calculate actual clip limit
+            clim = int(clip_limit * sub_img.size)
+            if clim < 1:
+                clim = 1
```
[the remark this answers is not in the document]
m-albert: Thanks @jni, incorporated your suggested change, makes it clearer and less convoluted :)

m-albert:
Thanks a lot @jni and @emmanuelle for your comments and thoughts!

> Like @VincentStimper there were a couple of places where I thought vectorisation might help

You're completely right and @VincentStimper had already mentioned it. I just had a further iteration with the code and put some effort into vectorising many of the for loops. The results are the same and speed is improved by orders of magnitude especially for small kernel sizes :)

> Typically, yes, we branch out the code in 2D and 3D cases if we think there is a significant advantage to the new method.

That makes sense.

> However, in this case I thought that with reflect boundary handling the difference would be minimal in most cases.

I agree that different boundary handling / padding would minimally affect the CLAHE results. In my understanding the main advantage of using symmetric padding (we're speaking about the same thing here, not the pad mode but balancing the padding at the start and end of each dim right?) is that the balancing would lead to less total required padding for vectorised execution.

The code that I pushed now pads the image such that the result in 2D is identical to the previous implementation (but faster).

> It'd be great to implement symmetric padding in a future PR if you're interested ;-)

This would be a straight-forward change to the current implementation and could potentially squeeze some more speed out of it (especially for large kernel sizes). We could of course also add symmetric padding to this PR, but as you say it probably makes sense to wrap up this one first and start a fresh one, also with a more targeted description (as also this discussion has become a bit lengthy by now).

m-albert:
Oh and another thing: Would you be interested in adding an example illustration of the new functionality?

In any case we don't have to delay this PR for that, but I was thinking that something like this could both illustrate the usefulness of clahe and create awareness that multidimensional code for it is available in `skimage`.

image


(this is a matplotlib 3d scatterplot, I know it looks somewhat ugly :P)

jni:
@m-albert that 3D example is gorgeous! I'd definitely want that in our gallery. Would you want to add that to this PR or do it in a new one? This is ready as far as I'm concerned!

m-albert:
> @m-albert that 3D example is gorgeous! I'd definitely want that in our gallery. Would you want to add that to this PR or do it in a new one?

@jni Great! Then I'd add it in a new one after cleaning up its code.

From my side I should be done with this PR :)

jni:
:tada: Thank you @m-albert and everyone else who participated!

PR #4604, opened 2020-04-21
Update min required sphinx version for sphinx-copybutton

sciunto:
## Description

I discovered that our min sphinx version is not satisfying since the recent addition of sphinx copybutton. It leads to this sort of error:

```Running Sphinx v1.3
making output directory...
Adding copy buttons to code blocks...

Exception occurred:
  File "/home/fr/.local/share/virtualenvs/sk/lib/python3.8/site-packages/sphinx_copybutton/__init__.py", line 25, in setup
    app.add_js_file('clipboard.min.js')
AttributeError: 'Sphinx' object has no attribute 'add_js_file'
The full traceback has been saved in /tmp/sphinx-err-y2eoe9d4.log, if you want to report the issue to the developers.

```

The problem is solved for sphinx >= 1.8.

@scikit-image/core We may want to build the doc with the min requirements. If we want it, let's do it in another PR.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,6 +1,4 @@
-# Sphinx 1.7.8 has a bug when building on cached envs
-# https://github.com/sphinx-doc/sphinx/issues/5361
-sphinx>=1.3,!=1.7.8
+sphinx>=1.8
 numpydoc>=0.9
 sphinx-gallery>=0.3.1
 sphinx-copybutton


requirements/docs.txt
```diff
@@ -1,6 +1,4 @@
-# Sphinx 1.7.8 has a bug when building on cached envs
-# https://github.com/sphinx-doc/sphinx/issues/5361
-sphinx>=1.3,!=1.7.8
+sphinx>=1.8
```
soupault, line 1: If we apply the 24-month rule (somewhat proposed in https://github.com/scikit-image/scikit-image/issues/4100#issuecomment-522331398), we should go with 1.7.4 (release date Apr 25, 2018). 1.8.0 was released on Sep 12th, 2018.

I'm personally fine with having 1.8.0, since this a pure doc dependency, and we can be more agressive with them. Just wanted to make sure that we are on the same page regarding those policies :slightly_smiling_face:.

sciunto: There is also @emmanuelle practical argument, which is: are we going to prevent people from using scikit-image if we increase the min version? (See https://github.com/scikit-image/scikit-image/pull/4568#issuecomment-611654645). In this case likely not. :)

Also, if we decide to set 1.7.4, we need to revert and hold on the plugin. Here, I loop back on the fact that we must extensively test our library against all the min versions.

sciunto: However, you raise an important point @soupault we probably need to think about a more solid policy on this, especially for scikit-image 1.0.

soupault: I feel I want to write a SKIP on this topic! Give me a day.

jni: @soupault remember that SKIPs are for controversial proposals. =P Start with an RFC, and if everyone agrees, we can adopt the policy. =)

sciunto: @jni, @soupault is working by anticipation :laughing:

PR #4609, opened 2020-04-22
Fix register_translation warning message

rfezzani:
## Description

I forgot to modify the deprecation warning message of `register_translation` in #4502 when we renamed it :pray:


sciunto:
Would have been puzzling :)

Issue #1396, opened 2015-02-27
Region property mapper


stefanv:
It is often desired to map region properties back to the original image.  E.g., calculate the area of each region, and then fill in that value into the array (or a new array).  This would be a useful utility function, and we could potentially do it faster than the user via Cython.

ahojnnes:
What do you mean by "fill in"?

stefanv:
I mean creating a map of a certain property, so that when you do map[r, c]
you get the corresponding property of the image at position r, c.

ahojnnes:
Makes sense, but let’s make this a goal for 0.12… at some point we should finally release 0.11 :-)

vighneshbirodkar:
We can do it in a way that functions like `label2rgb` can be a subset of this more generic property mapper

stefanv:
@vighneshbirodkar I like that tsuggestion

Issue #1349, opened 2015-01-22
Consider reimplementing ``segmentation.relabel_sequential`` using ``np.digitize``


stefanv:
Basic idea:

```
def remap_range(image):
    image = image.astype(int)
    unique_values = np.unique(image)
    index = np.digitize(image.reshape(-1,), unique_values) - 1

    return index.reshape(image.shape)
```

spark1729:
I am new to scikit-image development and have recently been trying to understand the code base. When I came across this issue, I looked up at ``relabel_sequential ``, in the ``_join.py`` in ``segmentation`` module. I tried to understand the above mentioned implementation, I found that the subtraction of 1 in the ``digitize`` line was producing wrong indices. ``np.digitize `` directly gives the matrix `` relabeled  `` when implemented.

This is the re-modeled code that I tried out:

    def relabel_sequential(image, offset=1):
        m = image.max()
        if not np.issubdtype(image.dtype, np.int):
            new_type = np.min_scalar_type(int(m))
            image = image.astype(new_type)
            m = m.astype(new_type)  # Ensures m is an integer
    
        labels = np.unique(image)
        index = np.digitize(image.reshape(-1,), unique_values) 
        relabeled =  index.reshape(image.shape)
   
        forward_map = np.zeros(m + 1, int)
        forward_map[labels0] = np.arange(offset, offset + len(labels0))
        if not (labels == 0).any():
            labels = np.concatenate(([0], labels))
        inverse_map = np.zeros(offset - 1 + len(labels), dtype=np.intp)
        inverse_map[(offset - 1):] = labels
    
        return relabeled, forward_map, inverse_map
    
Are we to implement the find ``forward_map`` and ``inverse_map`` using ``relabeled``?

spark1729:
I would really appreciate if anyone could point out any issues and guide me forward

soupault:
Whoa! That is a very cool function!

sinha-kirti:
Is anyone working on it at present? I would like to attempt it.

jni:
@sinha-kirti there is a PR that has not received any attention in many months: https://github.com/scikit-image/scikit-image/pull/1531 You could try to start from that branch and:

- time the new version and the old version — is either faster?
- try to simplify the code
- rebase it on master

Then you can make a new PR with both the original commits and yours, to supersede the existing PR.

Thanks!

sinha-kirti:
@jni Thanks for the consideration. I'll get started on it from today.

[...]

PR #4612, opened 2020-04-24
Allow very large values in relabel_sequential

VolkerH:
## Description

This PR introduces:
 
* an ehancement (I might even go as far as calling it a bug-fix) to `relabel_sequential`,  and 
* a generic vay to remap integer arrays (that can also remap from integer to other types)

Background:

**Storage requirements in `relabel_sequential` currently scale with maxim value in array not with the size of the array:**

The current implementation of  `relabel_sequential` uses numpy's fancy indexing in a very clever way. As it leverages numpy's array operators it is very fast. However, the current implementation requires building a LUT as a numpy array that scales (in memory requirement) with the value of the largest label. This is a long-standing problem mentioned in https://github.com/scikit-image/scikit-image/issues/1349#issuecomment-296676180. Depending on the exact value in the array this can lead  to `MemoryError` or `ValueError` exceptions. I have documented these failure modes in  [this notebook  under the heading Storage Requirements](https://github.com/VolkerH/PythonSnippets/blob/master/relabel_label_images/skimage_relabel_sequential_issues.ipynb).

**Cython implementation implements a sparse LUT using a hashmap**

To address the undesired memory scaling behaviour, a new function `map_array` is introduced that maps one array to another in which the LUT is implemented as a hashmap. Numpy does not provide a hash-table data structure and using a Python dict would be too slow. Instead, an `unordered_map`  from the C++ STL is used with Cython. Initial benchmarking on my machine suggests that this is not dramatically slower than the fany indexing (see [this notebook for my initial experiments, including alternative implementations using numba](https://github.com/VolkerH/PythonSnippets/blob/master/NumbaReassignPixels/numba_reassign_experiments.ipynb) for additional background.

`relabel_sequential` is reimplemented using the new `map_array`.

**Other uses: e.g. creating value-maps**
I have recently played around with some visualizations for measurements returned by `regionprops_table` to create value maps, [see here for an example](https://github.com/napari/napari/issues/998). This has been suggested for skimage in #1396 and can be trivially implemented using `map_array`.

There is still plenty of work to do in terms of code tidying, documentation and testing. 

Thanks to @jni for some help and encouragement with this.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -232,7 +232,11 @@ class ArrayMap:
         self._max_lines = 4
 
     def __array__(self, dtype=None):
-        """Return an array that behaves like the arraymap when indexed."""
+        """Return an array that behaves like the arraymap when indexed.
+        
+        This array can be very large: it is the size of the largest value
+        in the ``in_vals`` array, plus one.
+        """
         if dtype is None:
             dtype = self.out_values.dtype
         output = np.zeros(np.max(self.in_values) + 1, dtype=dtype)
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -134,8 +134,14 @@ def relabel_sequential(label_field, offset=1):
 
 
 def map_array(input_arr, input_vals, output_vals, out=None):
-    """
-    input_arr: input array (np.ndarray)
+    """Map values from input array from input_vals to output_vals.
+
+    Parameters
+    ----------
+    input_arr : array of int
+        The input label image.
+    input_vals : array of int
+        The values to map from.
     input_vals: 1d array of input values (integer)
     output_vals: 1d array of output values
     out: the output array. Created if not provided
@@ -166,7 +172,7 @@ class ArrayMap:
         return f'ArrayMap({repr(self.inval)}, {repr(self.outval)})'
 
     def __str__(self):
-        if len(self.inval) <= self._max_lines
+        if len(self.inval) <= self._max_lines:
             rows = range(len(self.inval))
             string = '\n'.join(
                 ['ArrayMap:'] +
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -168,6 +168,14 @@ class ArrayMap:
         self.outval = outval
         self._max_lines = 4
 
+    def __array__(self, dtype=None):
+        """Return an array that behaves like the arraymap when indexed."""
+        if dtype is None:
+            dtype = self.outval.dtype
+        output = np.zeros(np.max(self.inval) + 1, dtype=dtype)
+        output[self.inval] = self.outval
+        return output
+
     def __repr__(self):
         return f'ArrayMap({repr(self.inval)}, {repr(self.outval)})'
 
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -183,6 +183,9 @@ class ArrayMap:
             )
         return string
 
+    def __call__(self, arr):
+        return self.__getitem__(arr)
+
     def __getitem__(self, arr):
         return map_array(arr, self.inval, self.outval)
 
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -203,7 +203,7 @@ class ArrayMap:
     array if the values in the ``indices`` array are large.
 
     >>> values = np.array([0.25, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.0])
-    >>> indices = np.array([[0, 0, 10], [0, 10, 10]]
+    >>> indices = np.array([[0, 0, 10], [0, 10, 10]])
     >>> values[indices]
     array([[0.25, 0.25, 1.  ],
            [0.25, 1.  , 1.  ]])
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -186,15 +186,11 @@ def map_array(input_arr, input_vals, output_vals, out=None):
     input_arr = input_arr.reshape(-1)
     if out is None:
         out = np.empty_like(input_arr, dtype=output_vals.dtype)
-    else:
-        out_original = out
-        out = out_original.reshape(-1)
-        if out.strides[0] != min(out_original.strides):
-            # reshaping forced a copy
-            raise ValueError(
-                'If out array is provided, it should be either contiguous '
-                'or 1-dimensional.'
-            )
+    elif not (out.flags['FORC'] or out.ndim == 1):
+        raise ValueError(
+            'If out array is provided, it should be either contiguous '
+            'or 1-dimensional.'
+        )
 
     _map_array(input_arr, out, input_vals, output_vals)
     return out.reshape(orig_shape)
@@ -290,4 +286,3 @@ class ArrayMap:
         return map_array(
             arr, self.in_values.astype(arr.dtype, copy=False), self.out_values
         )
-
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -191,6 +191,13 @@ def map_array(input_arr, input_vals, output_vals, out=None):
             'If out array is provided, it should be either contiguous '
             'or 1-dimensional.'
         )
+    elif out.shape != orig_shape:
+        raise ValueError(
+            'If out array is provided, it should have the same shape as '
+            'the input array.'
+        )
+    else:
+        out = out.reshape(-1)
 
     _map_array(input_arr, out, input_vals, output_vals)
     return out.reshape(orig_shape)
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -149,25 +149,39 @@ def map_array(input_arr, input_vals, output_vals, out=None):
 
     Parameters
     ----------
-    input_arr : array of int
+    input_arr : array of int, shape (M[, N][, P][, ...])
         The input label image.
-    input_vals : array of int
+    input_vals : array of int, shape (N,)
         The values to map from.
-    input_vals: 1d array of input values (integer)
-    output_vals: 1d array of output values
-    out: the output array. Created if not provided
+    output_vals : array, shape (N,)
+        The values to map to.
+    out: array, same shape as `input_arr`
+        The output array. Will be created if not provided. It should
+        have the same dtype as `output_vals`.
+
+    Returns
+    -------
+    out : array, same shape as `input_arr`
+        The array of mapped values.
     """
 
-    # We want to reshape to 1D to make the numba loop as simple
-    # as possible
+    # We ravel the input array for simplicity of iteration in Cython:
     orig_shape = input_arr.shape
-    # numpy doc for ravel says 
+    # NumPy docs for `np.ravel()` says:
     # "When a view is desired in as many cases as possible, 
     # arr.reshape(-1) may be preferable."
     input_arr = input_arr.reshape(-1)
     if out is None:
-        out= np.empty_like(input_arr, dtype=output_vals.dtype)
-    out = out.reshape(-1)
+        out = np.empty_like(input_arr, dtype=output_vals.dtype)
+    else:
+        out_original = out
+        out = out_original.reshape(-1)
+        if out.strides[0] != min(out_original.strides):
+            # reshaping forced a copy
+            raise ValueError(
+                'If out array is provided, it should be either contiguous '
+                'or 1-dimensional.'
+            )
 
     _map_array(input_arr, out, input_vals, output_vals)
     return out.reshape(orig_shape)
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -101,6 +101,13 @@ def relabel_sequential(label_field, offset=1):
     >>> relab, fw, inv = relabel_sequential(label_field)
     >>> relab
     array([1, 1, 2, 2, 3, 5, 4])
+    >>> print(fw)
+    ArrayMap:
+      1 → 1
+      5 → 2
+      8 → 3
+      42 → 4
+      99 → 5
     >>> np.array(fw)
     array([0, 1, 0, 0, 0, 2, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0,
@@ -240,6 +247,12 @@ class ArrayMap:
     >>> in_indices = np.array([0, 10])
     >>> out_values = np.array([0.25, 1.0])
     >>> values = ArrayMap(in_indices, out_values)
+    >>> values
+    ArrayMap(array([ 0, 10]), array([0.25, 1.  ]))
+    >>> print(values)
+    ArrayMap:
+      0 → 0.25
+      10 → 1.0
     >>> indices = np.array([[0, 0, 10], [0, 10, 10]])
     >>> values[indices]
     array([[0.25, 0.25, 1.  ],
@@ -277,7 +290,7 @@ class ArrayMap:
         return f'ArrayMap({repr(self.in_values)}, {repr(self.out_values)})'
 
     def __str__(self):
-        if len(self.in_values) <= self._max_lines:
+        if len(self.in_values) <= self._max_lines + 1:
             rows = range(len(self.in_values))
             string = '\n'.join(
                 ['ArrayMap:'] +
@@ -285,7 +298,7 @@ class ArrayMap:
             )
         else:
             rows0 = list(range(0, self._max_lines // 2))
-            rows1 = list(range(-self.max_lines // 2, None))
+            rows1 = list(range(-self._max_lines // 2, 0))
             string = '\n'.join(
                 ['ArrayMap:'] +
                 [f'  {self.in_values[i]} → {self.out_values[i]}'
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -131,11 +131,24 @@ def relabel_sequential(label_field, offset=1):
     else:
         out_vals = np.arange(offset, offset+len(in_vals))
     input_type = label_field.dtype
-    # current version can return signed (if it fits in input dtype and that one
-    # is signed) but will return unsigned if a dtype change is necessary
+
+    # Some logic to determine the output type:
+    #  - we don't want to return a smaller output type than the input type,
+    #  ie if we get uint32 as labels input, don't return a uint8 array.
+    #  - but, in some cases, using the input type could result in overflow. The
+    #  input type could be a signed integer (e.g. int32) but
+    #  `np.min_scalar_type` will always return an unsigned type. We check for
+    #  that by casting the largest output value to the input type. If it is
+    #  unchanged, we use the input type, else we use the unsigned minimum
+    #  required type
     required_type = np.min_scalar_type(out_vals[-1])
-    output_type = (input_type if input_type.itemsize >= required_type.itemsize
-                   else required_type)
+    if input_type.itemsize < required_type.itemsize:
+        output_type = required_type
+    else:
+        if input_type.type(out_vals[-1]) == out_vals[-1]:
+            output_type = input_type
+        else:
+            output_type = required_type
     out_array = np.empty(label_field.shape, dtype=output_type)
     out_vals = out_vals.astype(output_type)
     map_array(label_field, in_vals, out_vals, out=out_array)
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -101,13 +101,13 @@ def relabel_sequential(label_field, offset=1):
     >>> relab, fw, inv = relabel_sequential(label_field)
     >>> relab
     array([1, 1, 2, 2, 3, 5, 4])
-    >>> fw
+    >>> np.array(fw)
     array([0, 1, 0, 0, 0, 2, 0, 0, 3, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4, 0,
            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5])
-    >>> inv
+    >>> np.array(inv)
     array([ 0,  1,  5,  8, 42, 99])
     >>> (fw[label_field] == relab).all()
     True
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -1,5 +1,6 @@
 import numpy as np
 from .._shared.utils import deprecated
+from ._remap import _map_array
 
 
 def join_segmentations(s1, s2):
@@ -117,25 +118,56 @@ def relabel_sequential(label_field, offset=1):
     array([5, 5, 6, 6, 7, 9, 8])
     """
     offset = int(offset)
-    if offset <= 0:
-        raise ValueError("Offset must be strictly positive.")
-    if np.min(label_field) < 0:
-        raise ValueError("Cannot relabel array that contains negative values.")
-    max_label = int(label_field.max()) # Ensure max_label is an integer
-    if not np.issubdtype(label_field.dtype, np.integer):
-        new_type = np.min_scalar_type(max_label)
-        label_field = label_field.astype(new_type)
-    labels = np.unique(label_field)
-    labels0 = labels[labels != 0]
-    new_max_label = offset - 1 + len(labels0)
-    new_labels0 = np.arange(offset, new_max_label + 1)
-    output_type = label_field.dtype
-    required_type = np.min_scalar_type(new_max_label)
-    if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
-        output_type = required_type
-    forward_map = np.zeros(max_label + 1, dtype=output_type)
-    forward_map[labels0] = new_labels0
-    inverse_map = np.zeros(new_max_label + 1, dtype=output_type)
-    inverse_map[offset:] = labels0
-    relabeled = forward_map[label_field]
-    return relabeled, forward_map, inverse_map
+    # current version can return signed (if it fits in input dtype and that one is signed)
+    # but will return unsigned if a dtype change is necessary
+    in_vals = np.unique(label_field)
+    out_vals = np.arange(offset, offset+len(in_vals))
+    input_type = label_field.dtype
+    required_type = np.min_scalar_type(out_vals[-1]) # what happens to signed/unsigned ?
+    output_type = (input_type if input_type.itemsize > required_type.itemsize
+                   else required_type)
+    out_array = np.empty(label_field.shape, dtype=output_type)
+    map_array(label_field, in_vals, out_vals, out=out_array)
+    fw_map = ArrayMap(in_vals, out_vals)
+    inv_map = ArrayMap(out_vals, in_vals)
+    return out_array, fw_map, inv_map
+
+
+def map_array(input_arr, input_vals, output_vals, out=None):
+    """
+    input_arr: input array (np.ndarray)
+    input_vals: 1d array of input values (integer)
+    output_vals: 1d array of output values
+    out: the output array. Created if not provided
+    """
+
+    # We want to reshape to 1D to make the numba loop as simple
+    # as possible
+    orig_shape = input_arr.shape
+    # numpy doc for ravel says 
+    # "When a view is desired in as many cases as possible, 
+    # arr.reshape(-1) may be preferable."
+    input_arr = input_arr.reshape(-1)
+    if out is None:
+        out_arr = np.empty_like(input_arr, dtype=output_vals.dtype)
+    out_arr = out_arr.reshape(-1)
+
+    _map_array(input_arr, out_arr, input_vals, output_vals)
+    return out_arr.reshape(orig_shape)
+
+
+
+class ArrayMap:
+
+    def __init__(self, inval, outval):
+        self.inval = inval
+        self.outval = outval
+
+    def inverse -> returns ArrayMap 
+        check bijunctive
+
+    def __getitem__(self, arr):
+        output = np.empty(arr.shape, self.outval.dtype)
+        map_array(arr, output, self.inval, self.outval)
+        return output
+        
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -196,28 +196,28 @@ def map_array(input_arr, input_vals, output_vals, out=None):
     # arr.reshape(-1) may be preferable."
     input_arr = input_arr.reshape(-1)
     if out is None:
-        out = np.empty_like(input_arr, dtype=output_vals.dtype)
+        out = np.empty(orig_shape, dtype=output_vals.dtype)
     elif out.shape != orig_shape:
         raise ValueError(
             'If out array is provided, it should have the same shape as '
             f'the input array. Input array has shape {orig_shape}, provided '
             f'output array has shape {out.shape}.'
         )
-    else:
-        try:
-            out.shape = (-1,)  # no-copy reshape/ravel
-        except AttributeError:  # if out strides are not compatible with 0-copy
-            raise ValueError(
-                'If out array is provided, it should be either contiguous '
-                f'or 1-dimensional. Got array with shape {out.shape} and '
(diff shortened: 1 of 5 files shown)


VolkerH:
Initial PR mentioned other uses for ArrayMap. We have decided to not include these in this pull request, but focus just on `relabel_sequential`.

jni:
@scikit-image/core this is ready for review! This code will allow a very easy fix for #1396, but since that will require adding a new API, which will require more discussion, we thought we should get this in first since it is already a major improvement.

CC @uschmidt83 you might be interested in this change. =)

skimage/segmentation/_join.py
```diff
@@ -116,26 +117,177 @@ def relabel_sequential(label_field, offset=1):
...
+            raise ValueError(
+                'If out array is provided, it should be either contiguous '
+                'or 1-dimensional.'
+            )
+
+    _map_array(input_arr, out, input_vals, output_vals)
+    return out.reshape(orig_shape)
+
+
+class ArrayMap:
```
rfezzani, line 203: I may be missed something, but is the definition of the `ArrayMap` object necessary?

skimage/segmentation/_join.py
```diff
@@ -116,26 +117,177 @@ def relabel_sequential(label_field, offset=1):
...
+    if input_type.itemsize < required_type.itemsize:
         output_type = required_type
-    forward_map = np.zeros(max_label + 1, dtype=output_type)
-    forward_map[labels0] = new_labels0
-    inverse_map = np.zeros(new_max_label + 1, dtype=output_type)
-    inverse_map[offset:] = labels0
-    relabeled = forward_map[label_field]
-    return relabeled, forward_map, inverse_map
+    else:
+        if input_type.type(out_vals[-1]) == out_vals[-1]:
```
rfezzani, line 148: Isn't it equivalent to 
```suggestion
        if np.can_cast(out_vals.dtype, input_type):
```
?

skimage/segmentation/_join.py
```diff
@@ -116,26 +117,177 @@ def relabel_sequential(label_field, offset=1):
...
+        out = out_original.reshape(-1)
+        if out.strides[0] != min(out_original.strides):
+            # reshaping forced a copy
+            raise ValueError(
+                'If out array is provided, it should be either contiguous '
+                'or 1-dimensional.'
+            )
+
+    _map_array(input_arr, out, input_vals, output_vals)
+    return out.reshape(orig_shape)
```
rfezzani, line 200: We should also check that `out.size == input_arr.size` if `out is not None`.

skimage/segmentation/_remap.pyx
```diff
@@ -0,0 +1,26 @@
...
+    for i in range(n_map):
+        lut[inval[i]] = outval[i]
+    # apply the map to the array
+    n_array = inarr.shape[0]
+    # The prange option gave some compilation warnings
+    #  "Unsigned index type not allowed before OpenMP 3.0"
+    # and didn't seem to be any faster
+    # for i in prange(n_array, nogil=True): #
+    for i in range(n_array):
+        outarr[i] = lut[inarr[i]]
```
rfezzani, line 26:
```suggestion
        outarr[i] = lut[inarr[i]]

```

skimage/segmentation/tests/test_join.py
```diff
@@ -87,7 +87,7 @@ def test_relabel_sequential_offset5_with0():
 
 
 def test_relabel_sequential_dtype():
-    ar = np.array([1, 1, 5, 5, 8, 99, 42, 0], dtype=float)
+    ar = np.array([1, 1, 5, 5, 8, 99, 42, 0], dtype=np.uint8)
```
rfezzani, line 90: Should we raise an error if input dtype is not integer?

skimage/segmentation/_join.py
```diff
@@ -116,26 +117,177 @@ def relabel_sequential(label_field, offset=1):
...
+        if input_type.type(out_vals[-1]) == out_vals[-1]:
+            output_type = input_type
+        else:
+            output_type = required_type
+    out_array = np.empty(label_field.shape, dtype=output_type)
+    out_vals = out_vals.astype(output_type)
+    map_array(label_field, in_vals, out_vals, out=out_array)
+    fw_map = ArrayMap(in_vals, out_vals)
+    inv_map = ArrayMap(out_vals, in_vals)
+    return out_array, fw_map, inv_map
```
rfezzani, line 157: Docstrings are no more correct since `fw_map` and `inv_map` are no more numpy arrays

skimage/segmentation/_remap.pyx
```diff
@@ -0,0 +1,26 @@
+# distutils: language = c++
+
+from libcpp.unordered_map cimport unordered_map
+from cython.parallel import prange
```
rfezzani, line 4:
```suggestion
```

skimage/segmentation/_remap.pyx
```diff
@@ -0,0 +1,26 @@
...
+@cython.boundscheck(False)  # Deactivate bounds checking
+@cython.wraparound(False)   # Deactivate negative indexing
+def _map_array(np_anyint[:] inarr, np_numeric[:] outarr,
+               np_anyint[:] inval, np_numeric[:] outval):
+    # build the map from the input and output vectors
+    cdef size_t i, n_map, n_array
+    cdef unordered_map[np_anyint, np_numeric] lut
+    n_map = inval.shape[0]
+    for i in range(n_map):
+        lut[inval[i]] = outval[i]
```
rfezzani, line 18: Should we check that `inval` only contains positive values?

skimage/segmentation/_remap.pyx
```diff
@@ -0,0 +1,26 @@
...
+    # apply the map to the array
+    n_array = inarr.shape[0]
+    # The prange option gave some compilation warnings
+    #  "Unsigned index type not allowed before OpenMP 3.0"
```
rfezzani, line 22: may be declaring `inarr` as `np_uints` removes these warnings... BTW, are `inarr` values allowed to be negative?

skimage/segmentation/_join.py
jni: @rfezzani I think the docstring does a decent job explaining things? It is a way to preserve the array indexing remapping API without creating enormous arrays.

skimage/segmentation/_join.py
jni: @rfezzani no, because if the input array is int32 (signed) and the output values can fit in that, we want to keep that type, but `np.can_cast(np.uint32, np.int32)` is False.

skimage/segmentation/_join.py
jni: Well, this fails naturally and spectacularly in that case, so the check is built-in. ;) But, sure, the error message can be clearer if we do our own check. =)

skimage/segmentation/_remap.pyx
jni: They are not allowed to be negative but they will often be a signed type because that is what NumPy uses for indexing.

skimage/segmentation/tests/test_join.py
jni: We don't have a strict policy on this. Often, if the docstring specifies the dtype, we consider that sufficient warning. In this case, the Cython code will fail. Happy to go either way, but I worry about functions that have a big chunk of validity checking at the top. It's a distraction.

skimage/segmentation/setup.py
```diff
@@ -23,6 +24,9 @@ def configuration(parent_package='', top_path=None):
                          include_dirs=[get_numpy_include_dirs()])
     config.add_extension('_slic', sources=['_slic.c'],
                          include_dirs=[get_numpy_include_dirs()])
+    config.add_extension('_remap', sources='_remap.cpp',
+                         include_dirs=[get_numpy_include_dirs()],
+                         language='c++', extra_compile_args=['-std=c++0x'])
```
jni, line 29: Just btw, this arg is required because `std::unordered_map` is not enabled by default in some older versions of gcc, and we were hitting this is the Linux builds.

https://stackoverflow.com/a/3973692/224254

I'll add a comment.

skimage/segmentation/_join.py
rfezzani: Well, I see that `out_vals.dtype` is always `int64`. I think that `np.can_cast(out_vals[-1], input_type)` do the trick then.

skimage/segmentation/_join.py
rfezzani: OK, my question then was not clear, sorry. I reformulate: can't we use a function here instead of the class definition?

VolkerH: The class implements the `__array__` method for backward compatibility as `relabel_sequential` returns the forward and backward transformations in the form of ndarrays that are used as lookup tables. 
If we just returned simple functions for forward and backward transformations existing code that relies on arrays being returned would no longer work.
Basically we need the function to be triggered when `[ ]` indexing is used.

rfezzani: OK, I was in fact missing something =), but I think that this approach is not addressing the problem of large memory allocation: it is simply postponed to the implementation of  the `__array__` method!

skimage/segmentation/_join.py
```diff
@@ -116,26 +117,172 @@ def relabel_sequential(label_field, offset=1):
...
+
+    def __array__(self, dtype=None):
+        """Return an array that behaves like the arraymap when indexed.
+        
+        This array can be very large: it is the size of the largest value
+        in the ``in_vals`` array, plus one.
+        """
+        if dtype is None:
+            dtype = self.out_values.dtype
+        output = np.zeros(np.max(self.in_values) + 1, dtype=dtype)
```
rfezzani, line 251: This array may be huge if `np.max(self.in_values)` is large....

jni: yes, but that is what this PR is fixing... We are only providing `__array__` as a compatibility shim for existing users of array indexing. ie this PR strictly *decreases* the surface area of this problem.

skimage/segmentation/_join.py
rfezzani: Yes, it is a simple `elif` statement more above ;-)

jni:
@rfezzani I have no understanding about why testing is failing since https://github.com/scikit-image/scikit-image/pull/4612/commits/433c64eefcd1851f1a2904bd778274c63523a72f ... Do you???

rfezzani:
> @rfezzani I have no understanding about why testing is failing since 433c64e ... Do you???

Me neither :-( Can it be related to the new Cython alpha release?

jni:
@rfezzani @VolkerH the latest commit should fix the issue, plus add that extra check you requested @rfezzani! ;) Still need to add a few tests to cover all the possible errors being raised, plus the str/repr of the ArrayMap class.

skimage/segmentation/_join.py
```diff
@@ -116,26 +117,179 @@ def relabel_sequential(label_field, offset=1):
...
+
+    # We ravel the input array for simplicity of iteration in Cython:
+    orig_shape = input_arr.shape
+    # NumPy docs for `np.ravel()` says:
+    # "When a view is desired in as many cases as possible, 
+    # arr.reshape(-1) may be preferable."
+    input_arr = input_arr.reshape(-1)
+    if out is None:
+        out = np.empty_like(input_arr, dtype=output_vals.dtype)
+    elif not (out.flags['FORC'] or out.ndim == 1):
```
rfezzani, line 189: I wonder if the check for `out` to be 1D is relevant since it is reshaped later :-/

VolkerH: If it is already 1d we don't care if it is contiguous. I think Cython takes care of iterating over non-contiguous arrays correctly. 

TBH, I prefer the original code: reshape first, complain if we got a copy (rather than a view) afterwards. It makes the intention clearer. The point is simply that we need a 1D - view into the source array to modify it in place with the Cython code. If numpy returns a copy instead (for whatever reason, maybe this could happen even if `out.flags['FORC']` is `True` if some other conditions are satisfied) we would not catch it with the logic that is in place now.

VolkerH:
Am I right that these latest (partial) test failures are unrelated to this code? Looks like some issue with pip?

jni:
@VolkerH yes, the failures are due to #4624

skimage/segmentation/_join.py
jni: @VolkerH you are right. Indeed I had to look up what `out.flags['FORC']` meant. Having said that, the earlier test was not exactly comprehensive either and would not have caught a copy if the array was contiguous to begin with. Perhaps we should use `np.may_share_memory`.

rfezzani: @VolkerH, I proposed the modifications above simply to reflect the raised error message, nothing more... If the test isn't appropriate, it may be means that the error message is not correct!
Lookin at note section in the [`np.reshape`](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html?highlight=reshape#numpy-reshape) documentation, one solution may be:
```python
try:
    out_view = out.view()
    out_view.shape = (-1, )  # this will fail if a copy is necessary
except AttributeError:
    raise ValueError("An appropriate error message")
```

skimage/segmentation/tests/test_join.py
emmanuelle: An error raised by us is probably easier to understand than a failure of Cython... I'd be :+1: for raising on float.

skimage/segmentation/_join.py
rfezzani: OK, I see...

Can't we consider using a `return_sparse_maps` (or equivalent) argument:
- if `False`, the function returns the dense `fw_map` and `inv_map` as actually,
- if `True`, the sparse version of `fw_map` and `inv_map` in any of the `scipy.sparse` available sparse matrix formats.

and in the documentation we show how remapping can be made with both options:

```python
>>> from scipy import sparse as Sp
>>> from skimage.segmentation import relabel_sequential
>>> # Dense version
>>> label_field = np.array([1, 1, 5, 5, 8, 99, 42])
>>> relab, fw, inv = relabel_sequential(label_field)
>>> (fw[label_field] == relab).all()
True
>>> (inv[relab] == label_field).all()
True
>>> # Sparse version
>>> sp_fw = Sp.lil_matrix(fw)
>>> sp_inv = Sp.lil_matrix(inv)
>>> (sp_fw[0, label_field] == relab).all()
True
>>> (sp_inv[0, relab] == label_field).all()
True
```

skimage/segmentation/_join.py
```diff
@@ -116,26 +117,179 @@ def relabel_sequential(label_field, offset=1):
...
+    #  unchanged, we use the input type, else we use the unsigned minimum
+    #  required type
+    required_type = np.min_scalar_type(out_vals[-1])
+    if input_type.itemsize < required_type.itemsize:
         output_type = required_type
-    forward_map = np.zeros(max_label + 1, dtype=output_type)
-    forward_map[labels0] = new_labels0
-    inverse_map = np.zeros(new_max_label + 1, dtype=output_type)
-    inverse_map[offset:] = labels0
-    relabeled = forward_map[label_field]
```
rfezzani, line 140: Following my [suggestion](https://github.com/scikit-image/scikit-image/pull/4612/files#r417282497), the only changes are:
```python
    if return_sparse_map:
        forward_map = Sp.csr_matrix((new_labels0, labels0, [0, max_label + 1]),
                                    shape=(1, max_label + 1))
        inverse_map = Sp.csr_matrix((labels0, range(offset, new_max_label + 1), [0, new_max_label + 1]),
                                    shape=(1, new_max_label + 1))
        relabeled = forward_map[0, label_field]
    else:
        forward_map = np.zeros(max_label + 1, dtype=output_type)
        forward_map[labels0] = new_labels0
        inverse_map = np.zeros(new_max_label + 1, dtype=output_type)
        inverse_map[offset:] = labels0
        relabeled = forward_map[label_field]
```
What do you think?

skimage/segmentation/_join.py
VolkerH: I don't think that is necessary. The `__array__` method of  `ArrayMap` is really only needed for backwards compatibiliy. If the memory wasn't a problem in theses historic cases it should not be now. In the future, the forward and backward mappings should simply be invoked as if it were fancy indexing (the `__getitem__` method takes care of that and if memory-efficient) or like a function call (`__call__`).

BTW, I had looked whether the original approach of using an array as LUT could be retained with `scipy.sparse` arrays before setting out on using Cython. I hadn't tried it but IIRC I read in the documentation that the sparse arrays don't support fancy indexing (or need to be converted to numpy arrays first).

jni: @rfezzani I still prefer the ArrayMap solution, for the following reasons:

(1) the sparse relabeling is actually much slower (>3x) even than the Cython code, which is already a bit slower than the NumPy code for moderately-sized images.
(2) we are trying to move away from `return_*` arguments
(3) we will need to deprecate the current default in favour of a new default, which is a pain
(4) the code is clever (and I do like fancy uses of csr matrices, I do it all the time!), but it is less readable/understandable than the Cython code, which is a straightforward loop
(5) @VolkerH is right, the fancy indexing with the sparse arrays only works for 1D arrays, so relabeling with sparse needs to go through a ravel and a reshape.

The dtype of the matrices also needs to be tweaked, too.

skimage/segmentation/_join.py
jni: @rfezzani no, the class allows relabeling with `[]` using Cython. It never needs to instantiate the arrays.

skimage/segmentation/_join.py
jni: Whoa. Nice idea, @rfezzani!

skimage/segmentation/_join.py
jni: True, will fix.

skimage/segmentation/_remap.pyx
jni: Actually, `lut` is an unordered_map (ie a dictionary), so it doesn't matter. This will work with negative values just as well as with positive ones.

skimage/segmentation/_join.py
rfezzani: I am confused here. If the `__array__`  method definition is required, the `MemoryError` pointed out in the PR description is not addressed...

skimage/segmentation/_join.py
rfezzani: No problem, but:

(1) I will have to do some more experiments to be convinced =p,
(2) I agree, the `return_*` arguments are not a good choice, but a better wording is may be possible,
(3) No deprecation is needed: if you see my [comment](https://github.com/scikit-image/scikit-image/pull/4612#discussion_r417581951), the actual behavior can be kept as default and the sparse arrays advised only for advanced use, when memory can be an issue
(4) this is subjective :)
(5) that's a major drawback, but if we consider that sparse matrices are only for advanced use, just few lines of documentation can suffices =p

The dtype is not much problematic with matrices then with arrays like it is the case in the actual implementation.

BTW, what prevents `relabel_sequence` from processing large values is the huge memory use that it implies, so if the `__array__` method is required and called in the `ArrayMap` approach, I wonder if the issue is addressed.

jni:
This now has 100% test coverage and all the sanity checks requested by @emmanuelle and @rfezzani.

jni:
Fun fact: I was concerned that this would fail catastrophically when attempting to map values not defined in the lut, but actually, the C++ standard map acts like a Python defaultdict and inserts 0s when a new key is encountered!

http://www.cplusplus.com/reference/unordered_map/unordered_map/operator[]/

This suits us fine. =)

skimage/segmentation/_join.py
```diff
@@ -116,26 +115,195 @@ def relabel_sequential(label_field, offset=1):
...
-        label_field = label_field.astype(new_type)
-    labels = np.unique(label_field)
-    labels0 = labels[labels != 0]
-    new_max_label = offset - 1 + len(labels0)
-    new_labels0 = np.arange(offset, new_max_label + 1)
-    output_type = label_field.dtype
-    required_type = np.min_scalar_type(new_max_label)
-    if np.dtype(required_type).itemsize > np.dtype(label_field.dtype).itemsize:
+    offset = int(offset)
+    in_vals = np.unique(label_field)
```
rfezzani, line 123: It seems that `out_array` is almost accessible via the `return_inverse ` of `np.unique`:
```python
>>> import numpy as np
>>> from skimage.segmentation import relabel_sequential
>>> 
>>> offset = 2
>>> label_field = np.array([[1, 1, 5, 5, 8, 99, 42],
...                         [1, 0, 0, 5, 99, 99, 42]])
>>> relab, fw, inv = relabel_sequential(label_field, offset=offset)
>>> 
>>> in_vals, out_array = np.unique(label_field, return_inverse=True)
>>> 
>>> out_array = out_array.reshape(label_field.shape) + offset - (0 in in_vals)
>>> out_array[label_field == 0] = 0
>>> relab
array([[2, 2, 3, 3, 4, 6, 5],
       [2, 0, 0, 3, 6, 6, 5]])
>>> out_array
array([[2, 2, 3, 3, 4, 6, 5],
       [2, 0, 0, 3, 6, 6, 5]])
```
These [3 lines](https://github.com/scikit-image/scikit-image/pull/4612/files#diff-09660f97c6b4c3dc5d143d3be12e8aaaR150-R152) are no more needed.

skimage/segmentation/_join.py
VolkerH: You are correct, one can still trigger a `MemoryError` if one wants to instantiate the lookup-table as a numpy array. However, instantiating the lookup-table is not required to access the functionality of relabeling. The same functionality is provided (without memory issues) using the `__getitem__` and `__call__` interfaces.

Example:
```
labels =  np.array([1, 1e16], dtype=np.int64)

# 1. call relabel_sequential
relab, fw, inv = relabel_sequential(labels) # does not allocate tons of memory

# 2. use the returned forward transform to relabel the array again using fancy indexing:
_relab = fw[labels] # does not allocate tons of memory ... is handled by __getitem__

# 3. use the returned forward transform to relabel the array again using call interdace:
_relab = fw(labels) # does not allocate tons of memory ... is handled by __call__

# 4. turn the returned forward transform into a numpy array 
fw_lut = np.narray(fw) # this will now try and allocate a huge array !
```

Use case `4.` is the only one where we still are likely to get `MemoryErrors` if we have large values in the input array. So in that sense one can still potentially hit this problem in the current code. However, this functionality is exclusively for backwards compatibility, in case anyone has code that used the returned transformations as an array (and not just for fancy indexing) for whatever reason. 
We provide the functionality for relabeling another array using the `fw` or `inv` transforms using methods `2.,3.`.

I agree that this is not immediately obvious. To address this I have the following suggestions:
1. spell out very clearly in the docstring that `np.array(fw)` is not recommended, or
1. raise a warning in `__array__`  that states that this is not recommended and for backwards compatibility only, or
1. remove the `__array__` method altogether, which may be a breaking change but will likely not affect a lot of use cases.

skimage/segmentation/_join.py
VolkerH: As you mentioned, this is partly subjective, but just point 3 is a deal-breaker for me:

3. If I write a program the uses `relabel_sequential` I don't necessarily know what sort of maximum labels someone might provide as input data. Therefore I would have to add code that finds the maximum label for each provided input image and then either set an arbitrary threshold or check how much memory is available to decide whether to call it with the option `return_sparse` or not. 

In my opinion, it should simply work independently of the maximum label value without having to provide such an option.

skimage/segmentation/_join.py
jni: @rfezzani Fascinating!!!

However, if we are building the map arrays anyway, this is actually faster than the overhead of running np.unique with `return_inverse=True`:

```python
In [3]: %timeit np.unique(seg)                                                     
3.83 ms ± 229 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [4]: %timeit np.unique(seg, return_inverse=True)                                
7.12 ms ± 399 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [5]: %timeit j.relabel_sequential(seg)                                          
6.2 ms ± 36.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [6]: reseg, fw, inv = j.relabel_sequential(seg)                                 
In [7]: %timeit fw[seg]                                                            
2.33 ms ± 152 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [8]: %timeit np.empty_like(seg)                                                 
547 ns ± 6.14 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Having said that, this would have been a useful refactor before this Cython code. :sweat_smile:

rfezzani:
`ArrayMap` are fine for me, but, as @VolkerH mentioned

> ... in case anyone has code that used the returned transformations as an array (and not just for fancy indexing) for whatever reason.

 it introduces a clear breaking change, and we are usually more careful in this case. Can't we think of a scenario with a deprecation cycle?

emmanuelle:
I finally took the time to read the entire PR and the comments. Would it be possible @VolkerH to return for now the arrays for the inverse map, unless a flag (for example `sparse`) is True, with the default value of the flag being False for backwards compatibility reasons. And to start a deprecation cycle to adopt `sparse=True` as the default in two releases. This way we can get the memory-saving behaviour by setting a flag but we still abide by our compatibility/deprecation policy. What do you think @jni @rfezzani ?

rfezzani:
I am +1 for @emmanuelle suggestion.

jni:
@emmanuelle my idea for this PR, which I suggested to @VolkerH, was to make a mock array-like object that would minimize, if not eliminate altogether, downstream issues with the API breakage. The problem I have with the `sparse=True` keyword is that it commits us to four versions of useless arguments — two with `sparse=None -> False`, and another two with `sparse=None -> True` but deprecated, because we don't want to have non-sparse in the future. @JDWarner has spoken out against this kind of transition in the past and I do agree with him at this point.

I just did a [github code search for relabel_sequential](https://github.com/search?p=3&q=relabel_sequential+-filename%3Atest_rag.py+-filename%3A_join.py+-filename%3A__init__.py+language%3APython&type=Code) (it includes some filename exclusions to avoid vendored copies of skimage), and find confirmed my suspicion that the vast majority of uses throw away the forward and inverse maps. In fact, I have to go to page 4 before I find [one example](https://github.com/mpicbg-csbd/stardist/blob/d4c983984c94ef43a83e9aa559ebbaa77d09fd8b/stardist/matching.py) (@uschmidt83's stardist), and in that example, our `__getitem__` implementation will continue to work.

In all, I went through 12 pages of results, and found just 4 cases where they didn't throw out the forward and inverse maps. This perhaps indicates that we should get rid of the forward and inverse maps altogether, but that is probably best reserved for the 1.0 transition, with the `function.extra` syntax.

Of the remaining 4 cases, one is the `stardist` case above (covered by current functionality). Another two require the addition of a `__len__` attribute but would otherwise work ([here](https://github.com/ishanivyas/DCNF-FCSP/blob/c961a1d7ef7bbaf3451d95e32f4c3c8851712d56/superpixel.py) and [here](https://github.com/MATSEAusbildung-RWTHAachen/Clusterman/blob/331557fcbcf6b481ded9096b0bb22d505e771489/histoComp.py))

Finally, on page 6, I see that @Borda in PyImSegm tries to [fill our LUT](https://github.com/Borda/pyImSegm/blob/d44a02aac1334802f7f9be0625441b7f4242dc1d/experiments_segmentation/run_compute_stat_annot_segm.py). This is one case where a conversion to numpy array is indeed needed. This could actually be done by creating a `__setitem__` method, at which point the full array is instantiated, cached, filled with the existing values, and then filled with the requested values. At any rate, it would be interesting to see what @Borda's preference would be for this PR. I think that the new ArrayMap object could simplify his code, which I believe is there as a workaround for negative labels?

As a side note, I also found [this example](https://github.com/constantinpape/cluster_tools/blob/d1bb7c1b1e517b8c603fed3028a0f8b530c01705/cluster_tools/features/region_features.py) of @constantinpape avoiding `relabel_sequential` because of the large array problem. It's unlikely that others avoiding our code would actually note it down so usefully. =)

So, in short, the code as is would improve performance for 117/120 users without any break whatsoever. A very simple fix covers 2 of the remaining cases, and a slightly harder but still simple fix means that 120/120 surveyed uses of this function would not see a breaking change. Given all this, I think it's a pretty compelling case to not have a 4-release long deprecation cycle...

jni:
@emmanuelle @rfezzani I've tried to alleviate issues with API breakage. Currently, the new code covers 120/120 surveyed uses of `relabel_sequential`. I'm happy to go over another 12 pages of results tomorrow. I agree that API breakage is a concern, and this search was indeed a useful exercise that helped me find some extra corner cases! However, I also think that we can mitigate the risk and it is overall better for the vast majority of users to have improved default performance now.

constantinpape:
> As a side note, I also found [this example](https://github.com/constantinpape/cluster_tools/blob/d1bb7c1b1e517b8c603fed3028a0f8b530c01705/cluster_tools/features/region_features.py) of @constantinpape avoiding `relabel_sequential` because of the large array problem. It's unlikely that others avoiding our code would actually note it down so usefully. =)

Great to see this happen! For me this actually is one of the major reasons to keep using `vigra`, which (currently) has a more efficient implementation. (As a side note, in the example you found, I can't use `vigra` either, because it assigns the new consecutive ids in scan order and not in the order of old labels, as is necessary in the function at hand).

One more comment though: I find getting the forward and inverse map very useful. At least for me, there is often the situation where I relabel some small array (node labels used for a graph based segmentation) and then need to apply this relabeling to some much larger array (the full pixel-wise segmentation). Is this still part of the new API or are you deprecating forward and backward map?

emmanuelle:
@jni I'm quite impressed by your efforts to check how the community uses this function, and also happy that it could lead to improvements for the PR. Given the fact that the `ArrayMap` now quacks really well like a numpy array, I'm going to approve!

VolkerH:
Thanks for reviewing @emmanuelle and @rfezzani  and thanks @jni for burning the midnight oil to do the sleuth work  which substantiates my suspicion that the forward and backward maps weren't actually used widely. With the `__len__` and `__setitem__` we are nearly there to reimplement `ndarrays` :-) . 
I have nothing to add to @jni 's comments.

@constantinpape, there are still forward and inverse maps returned, just in the form of an `ArrayMap` class and @jni has done most of the work that makes sure `ArrayMap` behaves pretty much the same as the arrays returned so far, minus the memory overhead (except and unavoidably when converting to an actual ndarray via the `__array__` method).

jni:
@constantinpape (:wave:!)

> Is this still part of the new API or are you deprecating forward and backward map?

To clarify my comment, the forward and backward maps will never disappear. In the next release, they will be (:crossed_fingers:) the ArrayMap objects proposed in this PR, which achieve the exact same thing as the current fw/inv maps but in memory proportional to the number of distinct labels, rather than the max label. In skimage 1.0, we *may* switch to the syntax `relabeled = relabel_sequential(labels)` and `relabeled, fw, inv = relabel_sequential.extra(labels)`, or similar. This has not been decided yet, just the leading of several proposals.

jni:
So, I continued my search today and found [someone](https://github.com/michielkleinnijenhuis/EM/blob/8372d2863c81969d6f8144e45f6f09d769c5d861/wmem/mergeblocks.py) *slicing* into the forward map and updating it with `+=`. They are relabeling over MPI and combining the fw maps somehow :exploding_head:. The EM segmentation crowd is very creative with this function. =D My latest update fixes that use case. I also verified that `np.save()` works!

@VolkerH yes, we are now very close to recreating NumPy arrays. :joy:

constantinpape:
@jni @VolkerH Thanks for clarifying. That's totally fine for my purposes.

jni:
After searching through 24 pages, I found just one more uncovered case: boolean indexing. This now covers 240/240 surveyed uses of relabel_sequential on GitHub.

JDWarner:
I'm going to go ahead and merge this so we can have it for 0.17.  The only CI failures appear to be pre-builds with Cython the culprit.

rfezzani:
Good job, thank you for all your efforts @VolkerH and @jni. I am happy to see this in 0.17 ;)

grlee77:
> After searching through 24 pages, I found just one more uncovered case: boolean indexing. This now covers 240/240 surveyed uses of relabel_sequential on GitHub.

Wow, thanks for doing this extensive work to ensure compatibility!

VolkerH:
Thanks everyone for reviewing and commenting and thanks to @jni for his work on taking this from my proof of concept to something that is ready to merge.

skimage/segmentation/tests/test_join.py
```diff
@@ -162,3 +181,81 @@ def test_relabel_sequential_already_sequential(offset, with0,
...
+    relabeled, fw, inv = relabel_sequential(ar)
+    fw[72] = 6
+    assert fw[72] == 6
+
+
+def test_arraymap_update():
+    in_values = np.unique(np.random.randint(0, 200, size=5))
+    out_values = np.random.random(len(in_values))
+    m = ArrayMap(in_values, out_values)
+    image = np.random.randint(1, len(in_values), size=(512, 512))
```
rfezzani, line 245: @jni, shouldn't you use `max(in_values)` instead of `len(values)`?

stefanv:
Thanks, all, this is quite an epic piece of work.  I love the simplicity of the C++ code too.

skimage/segmentation/tests/test_join.py
jni: You're absolutely right! (Well, +1). I meant to write `len(m)`. Will write a PR.

jni: @rfezzani see #4645 

Thank you!

Issue #4596, opened 2020-04-20
`exposure.equalize_adapthist` can fail for constant inputs due to incorrect behavior in `rescale_intensity`


m-albert:
## Description

`exposure.rescale_intensity` does not rescale the image to the range defined in `out_range` when min and max are equal.

This leads to `equalize_adapthist` to fail for many cases of constant inputs.

(In principle one wouldn't apply `equalize_adapthist` in the case of constant inputs, but probably it shouldn't fail, e.g. when processing a constant chunk of a large dask array).

## Way to reproduce
```python
from skimage import exposure
import numpy as np

ar = np.array([300,300])
ar_rescaled = exposure.rescale_intensity(ar, out_range=(0,255))
print(ar_rescaled)
```
> [76500 76500]

This should be within (0,255).

Here's how to make `exposure.equalize_adapthist` fail:
```python
img = np.zeros((8, 8)) + 0.1
img = img.astype(np.float64)
adapted = exposure.equalize_adapthist(img, 3) # yields exception
```

> IndexError: index 26214 is out of bounds for axis 0 with size 16384

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```
3.8.2 | packaged by conda-forge | (default, Mar 23 2020, 17:55:48)
[Clang 9.0.1 ]
macOS-10.14.6-x86_64-i386-64bit
scikit-image version: 0.17.dev0
numpy version: 1.18.2
```

m-albert:
It seems the problem is that when `imin == imax`, `rescale_intensity` doesn't do more than clipping the image to the `in_range`:

https://github.com/scikit-image/scikit-image/blob/37ca7deb15c1dd4d89b7c7e777a32909122009b5/skimage/exposure/exposure.py#L358-L362

In `equalize_adapthist`, images are therefore not properly rescaled to `(0, NR_OF_GRAY - 1)` here:
https://github.com/scikit-image/scikit-image/blob/37ca7deb15c1dd4d89b7c7e777a32909122009b5/skimage/exposure/_adapthist.py#L77-L78

However, correct scaling is assumed here:
https://github.com/scikit-image/scikit-image/blob/37ca7deb15c1dd4d89b7c7e777a32909122009b5/skimage/exposure/_adapthist.py#L126
[...]

PR #4615, opened 2020-04-24
Fixed wrong behaviour of `exposure.rescale_intensity` for constant input.

m-albert:
## Description

This closes https://github.com/scikit-image/scikit-image/issues/4596.

For a constant image, the output of `exposure.rescale_intensity` could lie outside of what is defined using out_range. The proposed fix in this case directly clips the input image to the output range (and converts to the output dtype). This is consistent with intended behaviour judging from already existing tests (test_rescale_same_values).

Importantly, this bug was leading to an exception in `equalize_adapthist` for constant input images in many cases.

Tests:
- add test to check that `rescale_intensity` behavior is fixed
- add test to check fixed `equalize_adapthist` behavior when providing constant input

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -391,10 +391,17 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
     >>> rescale_intensity(image, out_range=(0, 127))
     array([  0. ,  63.5, 127. ])
 
+    If the input image is constant, the output will be clipped directly to the
+    output range:
+    >>> image = np.array([130, 130, 130], dtype=np.int8)
+    >>> rescale_intensity(image, out_range=(0, 127))
+    array([ 127, 127, 127], dtype=int8)
+
     To get the desired range with a specific dtype, use ``.astype()``:
 
     >>> rescale_intensity(image, out_range=(0, 127)).astype(np.int8)
     array([  0,  63, 127], dtype=int8)
+
     """
     if out_range in ['dtype', 'image']:
         out_dtype = _output_dtype(image.dtype.type)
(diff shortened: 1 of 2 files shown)


skimage/exposure/exposure.py
```diff
@@ -406,7 +412,7 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
                                             clip_negative=(imin >= 0)))
 
     # Fast test for multiple values, operations with at least 1 NaN return NaN
-    if np.isnan(imin + imax + omin + omax):
+    if np.any(np.isnan([imin, imax, omin, omax])):
```
jni, line 415: Why did you make this change?

jni: I’m not too fussed honestly but if you make the change you should remove the comment above it. =)

m-albert: True that the change is unrelated to the issue this PR fixes, but I thought it might be too small for an issue by itself. My problem with the old line was that it could lead to an overflow warning (commented in the commit message). As these are just four scalars there shouldn't be a large cost to creating the array.

You're absolutely right that the comment above it should be removed :) I guess there's no need for an additional explanation of what the line does, as the warning below explains it. Thanks for pointing that out!

m-albert:
Sorry for the unrelated change! See https://github.com/scikit-image/scikit-image/pull/4615#discussion_r415080861 for what motivated it (it's also mentioned in the commit message). Should I open a new issue, or just leave it unchanged to not confuse things? I'm new to contributing to such a large project and happy to learn how to best deal with these things :)

skimage/exposure/exposure.py
```diff
@@ -406,7 +413,7 @@ def rescale_intensity(image, in_range='image', out_range='dtype'):
                                             clip_negative=(imin >= 0)))
 
     # Fast test for multiple values, operations with at least 1 NaN return NaN
```
emmanuelle:
```suggestion
    
```

emmanuelle:
thank you @m-albert, merging!

Issue #4618, opened 2020-04-26
flood_fill and flood are both in segmentation and morphology


emmanuelle:
The functions `flood_fill` and `flood` are both in the segmentation and in the morphology submodules, and neither is deprecated. I don't think we should keep the two versions in the API, but that we should deprecate one of them (probably keep segmentation?).

@JDWarner

rfezzani:
This reminds me this discussion https://github.com/scikit-image/scikit-image/pull/4400#issuecomment-577231251 ;-)

soupault:
I remember these two original arguments by @JDWarner - https://github.com/scikit-image/scikit-image/pull/3245#issuecomment-405047996, https://github.com/scikit-image/scikit-image/pull/3245#issuecomment-455918092. To me, both of them still sound reasonable.

jni:
Yeah, this is the recurring problem of tags vs folders. For a while, we decided to treat our modules as tags (watershed was also in both morphology and segmentation), as a solution to the problem that even advanced users such as myself were unable to find said watershed (after my proposal of a flat namespace was rejected. ;)

I think our latest consensus is that we want our modules to reflect *purpose* rather than the discipline from which they stem. So I think flood should probably go in segmentation, if we pick one.

PR #4619, opened 2020-04-26
added one example to flood fill tutorial

emmanuelle:
This PR results partly from a misunderstanding: since the flood fill function is both in morphology and segmentation (see #4618), I had mistakenly thought that there was no gallery example for `morphology.flood_fill` (from looking at the API doc and their mini-galleries) and I started writing an example.

Of course an excellent tutorial already exists. Maybe it's still interesting to add my example which uses the HSV space, so I still opened this PR for discussion.

@JDWarner


alexdesiqueira:
Thanks @emmanuelle, I like the idea of examples in different color spaces. LGTM!

doc/examples/segmentation/plot_floodfill.py
```diff
@@ -142,3 +140,47 @@
...
+img = data.astronaut()
+img_hsv = color.rgb2hsv(img)
+
+img_hsv_copy = np.copy(img_hsv)
+# Fill hue channel with a different color using flood fill
+_ = flood_fill(img_hsv_copy[..., 0], (313, 160), 0.5,
+               tolerance=0.016, in_place=True)
+
+# Return mask for post-processing
+mask = flood(img_hsv[..., 0], (313, 160), tolerance=0.016)
```
soupault, line 166: I feel this post-processing may unnecessarily complicate the example. And since the flood actually propagates into the areas beyond the suit, showing the patched mask contour may be a bit unfair :sweat_smile: . Do you think we could show here only the `flood_fill` result? Perhaphs, obtained with a higher `tolerance`, e.g. 0.03.

jni: I'm of the opposite opinion: since we are doing post-processing on the mask, let's not do the fill — only fill at the end. What do you think @emmanuelle?

soupault: @jni that would be infinitely better than my suggestion!

jni: @emmanuelle I just realised reading the comments that you are showing *both* methods, `flood_fill` and `flood`. I think it feels a bit artificial. What I was suggesting was to just use `flood`, then do `img_hsv[mask, 0] = 0.5` and `img_hsv_copy[mask_post_processed, 0], 0.5`, and show both images. I'll note that you never convert `img_hsv` back to RGB. After `img_hsv[mask, 0] = 0.5` you just never use `img_hsv` again.

emmanuelle:
@jni @soupault I've modified the example following on your comments. Thanks for the review!

PR #4640, opened 2020-05-05
Refactor ArrayMap.__setitem__ to prevent memory allocation

rfezzani:
## Description

I did not review this before the merge of #4612, sorry.
This PR removes the potential huge memory allocation when `MapArray.__array__` is called in `MapArray.__setitem__`.

## For reviewers

  later.

### DIFF
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -347,7 +347,7 @@ class ArrayMap:
         if np.isscalar(values) or len(values) == 1:
             values = np.repeat(values, len(indices))
 
-        for ind, val in zip(np.atleast_1d(indices), np.atleast_1d(values)):
+        for ind, val in zip(indices, values):
             if ind in self.in_values:
                 self.out_values[self.in_values == ind] = val
             else:
--- a/skimage/segmentation/_join.py
+++ b/skimage/segmentation/_join.py
@@ -332,9 +332,24 @@ class ArrayMap:
         return out
 
     def __setitem__(self, indices, values):
-        for ind, val in zip(indices, values):
+        if np.isscalar(indices):
+            indices = np.atleast_1d(indices)
+        elif isinstance(indices, slice):
+            start = indices.start or 0  # treat None or 0 the same way
+            stop = (indices.stop
+                    if indices.stop is not None
+                    else len(self))
+            step = indices.step
+            indices = np.arange(start, stop, step)
+        if indices.dtype == bool:
+            indices = np.flatnonzero(indices)
+
+        if np.isscalar(values) or len(values) == 1:
+            values = np.repeat(values, len(indices))
+
+        for ind, val in zip(np.atleast_1d(indices), np.atleast_1d(values)):
             if ind in self.in_values:
                 self.out_values[self.in_values == ind] = val
             else:
-                self.in_values = np.append(np.in_values, ind)
-                self.out_values = np.append(np.out_values, val)
+                self.in_values = np.append(self.in_values, ind)
+                self.out_values = np.append(self.out_values, val)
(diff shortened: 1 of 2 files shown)


skimage/segmentation/_join.py
```diff
@@ -333,8 +332,24 @@ def __getitem__(self, index):
...
+            indices = np.atleast_1d(indices)
+        elif isinstance(indices, slice):
+            start = indices.start or 0  # treat None or 0 the same way
+            stop = (indices.stop
+                    if indices.stop is not None
+                    else len(self))
+            step = indices.step
+            indices = np.arange(start, stop, step)
+        if indices.dtype == bool:
+            indices = np.flatnonzero(indices)
```
jni, line 345: I guess this whole section should be refactored as `indices = _normalize_index(indices)`, since it is repeated exactly in `__getitem__`

skimage/segmentation/_join.py
```diff
@@ -333,8 +332,24 @@ def __getitem__(self, index):
...
+
+        if np.isscalar(values) or len(values) == 1:
+            values = np.repeat(values, len(indices))
+
+        for ind, val in zip(indices, values):
+            if ind in self.in_values:
+                self.out_values[self.in_values == ind] = val
+            else:
+                self.in_values = np.append(self.in_values, ind)
+                self.out_values = np.append(self.out_values, val)
```
jni, line 355: @rfezzani the problem is that, if you look at the *two* cases out of 240 that used `setitem`, they used it for almost the entire array. Specifically, they did `fw[1:] += new_fw`. So, you end up allocating an equally huge array, and you do so very very slowly, using quadratic growth of the array. So imho this should be scrapped...

skimage/segmentation/_join.py
rfezzani: I hesitated, but I will ;)

skimage/segmentation/_join.py
rfezzani: I also though about this case and I will come with a proposal for implementing the `__iadd__` operator, since in this case we loose the sparseness of the `ArrayMap`

rfezzani:
The great advantage/justification of `ArrayMap` is its memory efficiency.
I personally support a non breaking implementation, even with degraded computation performance compared to `np.ndarray`.
We should advice users not concerned with memory efficiency to convert the `ArrayMap` to `np.ndarray`.

rfezzani:
> Specifically, they did fw[1:] += new_fw

@jni, is this use case supported when `new_fw` is a `MapArray`?

jni:
@rfezzani 

> is this use case supported when `new_fw` is a MapArray?

Whoa, a very good question! Actually not! However, the use case does not depend on that. Here it is:

https://github.com/michielkleinnijenhuis/EM/blob/8372d2863c81969d6f8144e45f6f09d769c5d861/wmem/mergeblocks.py#L213-L244

```python
def relabel_block(ds_in, maxlabel, mpi=None):
    """Relabel the labelvolume with consecutive labels.
    NOTE:
    relabel from 0, because mpi is unaware of maxlabel before gather
    """
    fw = relabel_sequential(ds_in[:])[1]

    if mpi.enabled:
        # FIXME: only terminates properly when: nblocks % size = 0

        num_labels = np.amax(fw)
        num_labels = mpi.comm.gather(num_labels, root=0)

        if mpi.rank == 0:
            add_labels = [maxlabel + np.sum(num_labels[:i])
                          for i in range(1, mpi.size)]
            add_labels = np.array([maxlabel] + add_labels, dtype='i')
            maxlabel = maxlabel + np.sum(num_labels)
        else:
            add_labels = np.empty(mpi.size)

        add_labels = mpi.comm.bcast(add_labels, root=0)
        fw[1:] += add_labels[mpi.rank]

    else:

        fw[1:] += maxlabel
        if len(fw) > 1:
            maxlabel = np.amax(fw)

    return fw, maxlabel
```

As you can see, they update with a NumPy array.

I don't think we should worry too much about the space efficiency of legacy uses, though. It is better to make sure we do *no worse* than what we are doing now.

PR #4642, opened 2020-05-05
AffineTransform: Allow a single value for 'scale' to apply to both sx & sy

perlman:
## Description

This is a small change to the AffineTransform constructor to accept a single scaling factor as an alternative to an explicit sx & sy. This is similar to transform.rescale which works with either a single scalar or one per dimension.

(This is probably not the right style to detect a single value, but sufficient for getting feedback if it should be fixed. Thanks!)

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -806,8 +806,7 @@ class AffineTransform(ProjectiveTransform):
             if translation is None:
                 translation = (0, 0)
 
-            scale = np.atleast_1d(scale)
-            if len(scale) == 1:
+            if np.isscalar(scale):
                 sx = sy = scale
             else:
                 sx, sy = scale
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -765,8 +765,8 @@ class AffineTransform(ProjectiveTransform):
     ----------
     matrix : (3, 3) array, optional
         Homogeneous transformation matrix.
-    scale : (sx, sy) as array, list or tuple, optional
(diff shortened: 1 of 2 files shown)


skimage/transform/_geometric.py
```diff
@@ -765,8 +765,8 @@ class AffineTransform(ProjectiveTransform):
     ----------
     matrix : (3, 3) array, optional
         Homogeneous transformation matrix.
-    scale : (sx, sy) as array, list or tuple, optional
-        Scale factors.
+    scale : {s as float or (sx, sy) as array, list or tuple}, optional
+        Scale factor(s). If a single value, it will be assigned to both sx and sy.
```
jni, line 769: Line too long for PEP8

```suggestion
        Scale factor(s). If a single value, it will be assigned to both
        sx and sy.
```

skimage/transform/_geometric.py
```diff
@@ -805,7 +805,12 @@ def __init__(self, matrix=None, scale=None, rotation=None, shear=None,
             if translation is None:
                 translation = (0, 0)
 
-            sx, sy = scale
+            scale = np.atleast_1d(scale)
+            if len(scale) == 1:
+                sx = sy = scale
+            else:
+                sx, sy = scale
```
jni, line 812: This is fine, but I would save a line by doing:

```suggestion
            if np.isscalar(scale):
                sx = sy = scale
            else:
                sx, sy = scale
```

PR #4644, opened 2020-05-05
Fix sphinx

### LINKED ISSUE #4616: Search links broken in dev documentation
When searching a function name in the search bar of the dev doc, many links are found but they are incorrect, since following the links result in a 404 error. 

Is this related to the new sphinx version?

emmanuelle:
Closes #4616. At the moment the search is broken in the dev documentation (including on the deployed website). We should try to fix this with the most recent sphinx but since the release is coming, this is a quick and dirty solution, pinning the sphinx version before the offending version.

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,4 +1,4 @@
-sphinx>=1.8
+sphinx>=1.8,<=2.4.4
 numpydoc>=0.9
 sphinx-gallery>=0.3.1
 sphinx-copybutton
--- a/skimage/transform/_hough_transform.pyx
+++ b/skimage/transform/_hough_transform.pyx
@@ -313,9 +313,9 @@ def _hough_line(cnp.ndarray img,
     # finally, run the transform
     cdef Py_ssize_t nidxs, nthetas, i, j, x, y, accum_idx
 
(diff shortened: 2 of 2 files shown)


emmanuelle:
I've also included my not-merged-yet commit fixing the CI pre build so that we can check whether the CI is green. I've checked that the search works locally with this version but the only way to be completely sure that it works is to merge and deploy the doc....

hmaarrfk:
Is it possible to use a not equal constraint?

emmanuelle:
@hmaarrfk it would be possible but at the moment it does not work with the whole 3.0 series. I see this fix more as a band-aid solution, and given the sphinx discussions linked to in #4616 I believe the problem is to be fixed on our side rather than on sphinx's but I'm not sure how to do it at the moment, with little time to investigate this specific problem.

sciunto:
Shall we open a ticket or add a TODO task to care about these version in the future?

emmanuelle:
Hum yes good idea @sciunto!

hmaarrfk:
Thansk for the explanation. 

I think the constrain <3.0.0. Would have made things super explicit in that case.

I was but by Sphinx 3.0 bugs too... Not fun

PR #4646, opened 2020-05-06
Expose recently introduced map_array function in skimage.util

emmanuelle:
Since this function can be useful for other tasks. I also wondered whether we should expose `ArrayMap` or not (at the moment it is not and for skimage developers it's easy to import it from its location). It's easy to change of course.


skimage/util/__init__.py
```diff
@@ -14,7 +14,7 @@
 from .unique import unique_rows
 from ._invert import invert
 from ._montage import montage
-
+from ._map_array import map_array
```
jni, line 17:
```suggestion
from ._map_array import map_array

```
PEP8

skimage/segmentation/_join.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 from .._shared.utils import deprecated
-from ._remap import _map_array
+from ..util import map_array
+from ..util._map_array import ArrayMap
```
jni, line 4:
```suggestion
from ..util._map_array import map_array, ArrayMap
```

I think it's better to use direct imports in our own code...

jni:
To clarify what @emmanuelle meant by "other tasks": this would close #1396. It needs a gallery example but it's probably not worth holding up release for it.

skimage/segmentation/_join.py
emmanuelle: oh interesting I didn't know.

rfezzani:
I am thinking about submitting a PR to create a `skimage.labelling` submodule where `skimage.morphology.label`/`skimage.measure.label`, `skimage.segmentation.relabel_sequential`, `skimage.color.colorlabel`... can take place. `MapArray` was on list too but `skimage.util` is a good place ;-).
I am waiting for the CI to be green before approving...

emmanuelle:
It is green now :-).

rfezzani:
Merged ;) Thank you @emmanuelle

jni:
@rfezzani I like your new module idea. `labeltools`? Anyway, I think the best time for this transition would be the 1.0 release. =)

rfezzani:
Yes @jni, there is no hurry ;) `labeltools` is good for me :+1:

skimage/segmentation/_join.py
jni: It's a new in-trial policy from #3397 =)

JDWarner:
Continuing the tangent, our top-level submodule naming convention is direct, not plural, and tends not to include portmanteaus.

Refactoring the label infrastructure makes some sense to me.  The best submodule name would be `skimage.label` IMHO, but `label` already being a function name clouds the picture a bit (`from skimage.label import label` feels somewhat cumbersome).

Issue #4601, opened 2020-04-20
Discuss gallery structure to bring out bioimaging examples.


mkcor:
## About

I would like to spark a discussion about an extension and possible reorganization of the [gallery](https://scikit-image.org/docs/stable/auto_examples/). Currently, the gallery is structured into families of processing tasks (filtering, feature detection, object segmentation, ...). Alternatively or additionally, we may want to offer starting points which refer to the scientific fields of application (life sciences, geospatial applications, ...).

## Life sciences

I'm working with @emmanuelle to supplement the gallery with simple processing examples of biological/biomedical images. To begin with, the examples would be small; typically, they would more straightforward than [tutorials](https://github.com/scikit-image/skimage-tutorials) used in teaching settings. The examples should look very familiar to a biologist landing on our gallery. Note that, in the big picture, this effort relates to @jni's [project via CZI](https://chanzuckerberg.com/newsroom/czi-announces-support-for-open-source-software-efforts-to-improve-biomedical-imaging/). 

## (Large) bioimaging data

A great way to make an example look familiar is to use datasets (here, biomedical images) which are well-known in the field or at least typical of the field. I believe I can use images made [available](https://cellprofiler.org/examples/#human-cells) by [CellProfiler](https://github.com/CellProfiler/CellProfiler/blob/master/LICENSE) -- I will reach out to them to see if the attribution terms also apply to the data and, if so, to make sure the attribution notice I will include suits them.

I will also be happy to add an example showcasing how to download https://github.com/scikit-image/skimage-tutorials/raw/master/images/cells.tif (~30MB dataset with 3D images of cells) on the fly (now possible thanks to the recent setup of Pooch #3945) and carry out basic processing tasks on it.

emmanuelle:
Thanks for starting this discussion @mkcor ! I think I would favor having a separate gallery page (with prominent cross-linking with the main gallery page of course) for applications, since our gallery is already quite a long page. Note that with sphinx-gallery it's possible to have several gallery pages, you just pass a list of dirs instead of a str for the `gallery_dirs` key in the sphinx-gallery config.

mkcor:
Good to know! Thanks, @emmanuelle. So... alongside 

    Download
    Gallery
    Documentation
    Community Guidelines
    Source

we would have another tab named "Life Sciences" or something like "For the life scientist" -- not that I need to picture now the final rendering at the highest level. I'll start by writing some content for this domain-specific gallery!

[...]

Issue #4326, opened 2019-11-28
Compressed TIFF files?


kmader:
## Description
I get error messages when trying to read various compressed tiff images using `skimage.io.imread` (both could be read using `PIL.Image.open`)

- PACKBITS
The first error is `ValueError: TiffPage 0: cannot decompress PACKBITS` and comes with this file https://gist.github.com/kmader/6cb706e52a58c6edc2f4c97678cea4a9/raw/a1dbd04aeb19c7d032f6a7555d4b69045fcddd3d/00215.tif 

And was found in a 3D tomography repository at https://datadryad.org/stash/dataset/doi:10.5061/dryad.c65b4j5

- LZW
The second error is `ValueError: TiffPage 0: cannot decompress LZW`
https://gist.github.com/kmader/6cb706e52a58c6edc2f4c97678cea4a9/raw/8100ea501cb1b60857891d49e0b9320766992e7f/Romundinactrpt0000.tif

## Way to reproduce
```python
from skimage.io import imread
imread('00215.tif')
```

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
# your output here
3.6.6 |Anaconda, Inc.| (default, Oct  9 2018, 12:34:16) 
[GCC 7.3.0]
Linux-4.9.0-11-amd64-x86_64-with-debian-9.9
scikit-image version: 0.16.2
numpy version: 1.16.4
```

kmader:
As I frequently use images like these in lectures (and would like to use skimage for as much as possible) I would like to fix the error, but am unsure how to best proceed.

hmaarrfk:
can you try to install tifffile?

What is the output of `conda list`???

hmaarrfk:
If you do need a newer version of tifffie than the one we vendor, I think eventually, you will find that you have many dependencies that are beyond scikit-image, e.g. jupyter, bokeh???. You might want to create an `environment.yml` file, that you can share and ask students to update as you add more dependencies
```
conda env update --file environment.yml
```
https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html

kmader:
@hmaarrfk I see this which seems like it very new.
```
tifffile                  2019.7.26                pypi_0    pypi
```

[...]

PR #4648, opened 2020-05-06
Add first biology example: segmentation of human cells with mitosis.

mkcor:
## Description

This addition falls under #4601 and aims at paving the way for bioimaging examples.

At the moment, I'm fetching the example data from a public GitHub repo; I'm happy to discuss whether we should include the image in the (newly available) pooch registry.

I would like to have a professional biologist proofread my content.

Anecdotally, I'm happy to share that I have used the 'flag' [colormap](https://matplotlib.org/3.1.1/gallery/color/colormap_reference.html) for the first time ever:

image

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,170 @@
+"""
+================================
+Segment human cells (in mitosis)
+================================
+
+In this example, we analyze a microscopy image of human cells. We use data
+provided by Jason Moffat [1]_ through
+[CellProfiler](https://cellprofiler.org/examples/#human-cells).
+
+.. [1] Moffat J, Grueneberg DA, Yang X, Kim SY, Kloepfer AM, Hinkle G, Piqani B, Eisenhaure TM, Luo B, Grenier JK, Carpenter AE, Foo SY, Stewart SA, Stockwell BR, Hacohen N, Hahn WC, Lander ES, Sabatini DM, Root DE (2006) "A lentiviral RNAi library for human and mouse genes applied to an arrayed viral high-content screen" Cell, 124(6):1283-98. DOI: [10.1016/j.cell.2006.01.040](https://doi.org/10.1016/j.cell.2006.01.040). PMID: 16564017
```
sciunto, line 10: For the DOI, please use the notation  

>   :DOI:`1234` 

Please, wrap at 80 chars also.

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,170 @@
...
+"""
+
+import matplotlib.pyplot as plt
+import numpy as np
+from scipy import ndimage as ndi
+
+from skimage import feature, filters, io, morphology, util
+
+
+image = io.imread('https://github.com/CellProfiler/examples/blob/master/ExampleHuman/images/AS_09125_050116030001_D03f00d0.tif?raw=true')
```
sciunto, line 21: To be safe, I would suggest to host our own copy. Any comment on this @scikit-image/core ?

jni: I agree but I would host it in gitlab.com/scikit-image, or a new github repo, rather than here.

mkcor: @sciunto thanks for reviewing! In the PR description, I wrote: "At the moment, I'm fetching the example data from a public GitHub repo; I'm happy to discuss whether we should include the image in the (newly available) pooch registry." <-- Do you have an opinion on the latter?

doc/examples/applications/plot_human_mitosis.py
mkcor: Oh, thanks! I should have done a recursive grep for "DOI" in `./doc/examples`... ;) Fixed.

doc/examples/applications/plot_human_mitosis.py
sciunto: This was my thought, ie include it in pooch. @emmanuelle Do you agree?

sciunto:
I'm not a biologist, but the example looks really cool :) After we make our decision on the image hosting, this is ready for me.

sciunto:
CI arror: ```Warning, treated as error:

/home/travis/build/scikit-image/scikit-image/doc/source/auto_examples/applications/plot_human_mitosis.rst:77:undefined label: sphx-glr-auto-examples-segmentation-plot-multiotsu-py (if the link has no caption the label must precede a section header)```

mkcor:
@sciunto thanks for pointing to the specific error in the CI build; I just looked up the Sphinx docs
https://www.sphinx-doc.org/en/1.7/markup/inline.html#cross-referencing-arbitrary-locations and:

* Doesn't our use case (I mean, throughout `./doc/examples/`) correspond to `:doc:` rather than `:ref:`?
* I retrieved the label reference from the URLs (https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_watershed.html#sphx-glr-auto-examples-segmentation-plot-watershed-py and https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_multiotsu.html#sphx-glr-auto-examples-segmentation-plot-multiotsu-py, respectively);
* only now did I notice that the existing examples used a different syntax, with underscores and dot (matched in cec55753c91b60112e6efc2c46f2013b4e1182a0);
* I wonder why the warning went off with `sphx-glr-auto-examples-segmentation-plot-multiotsu-py` but not with `sphx-glr-auto-examples-segmentation-plot-watershed-py`: I hope it's not because the [watershed example](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_watershed.html) contains a (unrelated) reference label (i.e., Wikipedia)...

doc/examples/applications/plot_human_mitosis.py
emmanuelle: ok, let's host in https://gitlab.com/scikit-image/data as suggested by @jni. That will be our first image in this repo :-). When you do this @mkcor please add the origin of this image in the Readme of the repo. Then for adding the image to the registry you can copy what's done for the `cells.tif` dataset.

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,176 @@
...
+
+#####################################################################
+# This image is a TIFF file. If you run into issues loading it, please
```
emmanuelle, line 30: This is a bit sad :-(. @mkcor added this comment because she ran in the same issue as in #4326. If we store our own version of this file should we convert it to png (hiding the tiff problem under the rug is not great either...)

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,176 @@
...
+# consider using ``external.tifffile.imread`` or following:
+# https://github.com/scikit-image/scikit-image/issues/4326#issuecomment-559595147
+
+fig, ax = plt.subplots()
+ax.imshow(image, cmap='gray')
```
emmanuelle, line 35: Could you please add a title like "Microscopy image of human cells"?

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,176 @@
...
+# Cellular biology uses the mitotic index to quantify cell division (and,
+# hence, cell population growth). By definition, it is the ratio of cells in
+# mitosis (over the total number of cells). To analyze the above image,
+# we are thus interested in two thresholds: one separating the cells from the
+# background, the other separating the dividing nuclei (brighter spots) from
+# the cytoplasm of their respective mother cells (and, intensity-wise, from
+# the other cells). To separate these three different classes of pixels, we
+# resort to :ref:`sphx_glr_auto_examples_segmentation_plot_multiotsu.py`.
+
+thresholds = filters.threshold_multiotsu(image)
```
emmanuelle, line 56:
```suggestion
thresholds = filters.threshold_multiotsu(image, classes=3)
```

emmanuelle: Here I would specify explicitly the number of classes, even if the default parameter is used, for the sake of pedagogy (we're looking for 3 classes). Very nice use of multiotsu by the way!

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,176 @@
...
+plt.show()
+
+#####################################################################
+# Since there are touching cells, thresholding is not enough to segment all
+# the cells. If it were, we could readily compute a mitotic index for this
+# sample:
+
+cells = image > thresholds[0]
+dividing = image > thresholds[1]
+labeled_cells = morphology.label(cells)
```
emmanuelle, line 75: nowadays we recommend to use `label` from `measure` instead of `morphology`.

doc/examples/applications/plot_human_mitosis.py
jni: This should no longer be an issue if you `pip install -U tifffile`, because @cgohlke provided wheels for imagecodecs. So I would argue that we should remove the comment, and perhaps increase the tifffile version required for docs.

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,176 @@
...
+# hand, we want a smoother image, removing small spurious objects and,
+# possibly, merging clusters of neighbouring objects (some could correspond to
+# two emerging nuclei in one mother cell). In a way, the segmentation
+# challenge we are facing with dividing nuclei is the opposite of that with
+# (touching) cells.
+
+#####################################################################
+# To find suitable values for thresholds and filtering parameters, we proceed
+# by dichotomy, visually and manually.
+
```
emmanuelle, line 121: Here an alternative solution would be to use Hysteresis thresholding to keep connected components with values higher than the high multi-otsu threshold, but only if they also have pixels of higher value than an other (higher) threshold. This is a classical method to remove small spurious objects but have instead large smooth objects
```
divinding = filters.apply_hysteresis_threshold(image, thresholds[1], 1.4 * thresholds[1]), [0.5], colors="yellow"
```
The results are not completely the same as with the current method, but not being a biologist I would not know which result is the best.

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,176 @@
...
+#####################################################################
+# Segment cells
+# =============
+# To separate touching and overlapping cells, we resort to
+# :ref:`sphx_glr_auto_examples_segmentation_plot_watershed.py`.
+
+distance = ndi.distance_transform_edt(cells)
+
+local_maxi = feature.peak_local_max(distance, indices=False,
```
emmanuelle, line 152: `min_distance` is maybe easier to understand than `footprint` when the `footprint` is isotropic.

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,176 @@
...
+                                    footprint=morphology.disk(3))
```
emmanuelle, line 153:
```suggestion
                                    min_distance=7)
```

emmanuelle:
I think one of the figures could use `plt.contour` on the original image to visualize the result, it's a another way to visualize a binary array which is often useful.

emmanuelle:
Thank you very much @mkcor for this very nice tutorial! You're right that it'd be great to have a review of the approach and the text by someone from life or medical sciences. Pinging @ThomasWalter and @JDWarner, and also the friends from CellProfiler who are kind enough to let us use their image @0x00b1 @AnneCarpenter

doc/examples/applications/plot_human_mitosis.py
mkcor: Good point!

doc/examples/applications/plot_human_mitosis.py
mkcor: Right. Thanks!

doc/examples/applications/plot_human_mitosis.py
emmanuelle: on the other hand this comment is not for our build but for users who may have their own set of installed packages. Only a small fraction of users will think of consulting the doc dependencies for minimal versions when looking at a doc example.

sciunto: If an extra package is missing, then a proper message must be displayed. If it is not the case, it is a bug. If it is the case, I don't thing this comment is useful anymore.

AnneCarpenter:
It's certainly fine with the CellProfiler team to make a copy of that image if you deem that preferable.

doc/examples/applications/plot_human_mitosis.py
stefanv: I'd probably get rid of this, given that it has been fixed.  Some users may run into it initially, but that number should drop down very quickly.

doc/examples/applications/plot_human_mitosis.py
mkcor: Sure! Done 69bfa859048fe09b283678dad64ca04edce9be32.

doc/examples/applications/plot_human_mitosis.py
mkcor: Good to know! That's what I was doing in the early stages of this example, then I figured I could perhaps spare one import...

mkcor:
> I think one of the figures could use `plt.contour` on the original image to visualize the result, it's a another way to visualize a binary array which is often useful.

@emmanuelle thanks for reviewing. Oh, yes!

contour_cells

It looks like we can readily identify the cells (contours) and those in particular undergoing mitosis (filled in bright blue)?! So we could get another rough estimate of the mitotic index, just using the attributes of `ax.contour(image, origin='image')` (which is a [QuadContourSet](https://matplotlib.org/3.1.1/api/contour_api.html#matplotlib.contour.QuadContourSet))...

doc/examples/applications/plot_human_mitosis.py
mkcor: Done 126edca4bf6326f3a90b795e1caf690245a37ce5.

doc/examples/applications/plot_human_mitosis.py
mkcor: Ok, great! PR submitted over at: https://gitlab.com/scikit-image/data/-/merge_requests/2

alexdesiqueira:
@mkcor Just a side note, this looks _beautiful_! Thank you very much, Marianne!
Could I suggest you something? The red on the background is kinda eye-hurting... what if we had a white background?
Thanks again! I'll use this example often, for sure :slightly_smiling_face:

mkcor:
Thank you, @alexdesiqueira!

> The red on the background is kinda eye-hurting...

I see what you mean... Amongst colormaps readily [available](https://matplotlib.org/tutorials/colors/colormaps.html) in Matplotlib, I found that only 'flag' (and 'prism' to a lesser extent) were suitable for showing all the segmented cells. Do you have a colormap in mind, that I may have missed?

Otherwise, and maybe that's precisely what you are suggesting, I could [create](https://matplotlib.org/tutorials/colors/colormap-manipulation.html) my own colormap, based on 'flag,' where I tweak it so it 'starts at' white instead of red.

jni:
@mkcor typically segmentations with a background label are shown with the background alpha=0. @emmanuelle has a nice example [here](http://emmanuelle.github.io/a-tutorial-on-segmentation.html) showing how to use `color.label2rgb` together with `plt.cm.spectral` for displaying segmentations.

doc/examples/applications/plot_human_mitosis.py
mkcor: Ok, done c5eb4c17710584b30a094ba772fd7310b1da5a8a. I wish I could check the behaviour for real but I haven't managed: For the past half-hour I've been trying to test this in a fresh environment, but the install (of scikit-image) kept failing. I've used pip, conda, and mamba -- only the latter pointed to a timeout issue while installing, while pip and conda would 'install' apparently 'successfully' but then I would find out the install was broken when loading the package.

alexdesiqueira:
@mkcor @jni I'm very fond of `nipy_spectral` when plotting labels as well.

grlee77:
> @mkcor @jni I'm very fond of nipy_spectral when plotting labels as well.

Thanks for the suggestion. There is also the `distinguishable_colors` [function in mindboggle](https://mindboggle.readthedocs.io/en/latest/api/mindboggle.mio.html#module-mindboggle.mio.colors)

Another example is the Glasbey colormap (https://imagej.net/Glasbey or [python implementation](https://github.com/taketwo/glasbey))

emmanuelle:
We had a short discussion with @stefanv and @jni and we thought that since we are not distributing the image, it is ok to do what the example is doing at the moment, ie do `io.imread` on a URL where the image is hosted. Please add a comment with the license of the image. When/if the image is released under a license compatible with scikit-image we'll add it to the pooch repo but let's not block the example for this since the imread on URL solution is acceptable.

mkcor:
Great news! Thank you, @emmanuelle, @stefanv, and @jni! I'm currently reworking the rendering of the final images, following the colourmap suggestions offered by the reviewers.

emmanuelle:
For the next steps, I've also asked Pierre Poulain to take a quick look at the example since he knows more about cell biology than me. Maybe @ThomasWalter also if you could take a quick look it'd be awesome. There remains the CI failure due to the tifffile, for minimal dependencies. @stefanv is there a possible workaround?

mkcor:
> @mkcor @jni I'm very fond of `nipy_spectral` when plotting labels as well.

@alexdesiqueira I must be missing something because, as is, `nipy_spectral` yields gradual variations:

spectral

To highlight the segmentation of otherwise connected regions, we want the opposite, don't we? We want "a color table structured in a maximally discontinuous manner. That is, adjacent color bins are chosen to be as distinct from one another as possible." This is a quote from

> https://imagej.net/Glasbey

as shared by @grlee77 (thanks). In the [example](http://emmanuelle.github.io/a-tutorial-on-segmentation.html) by @emmanuelle shared by @jni, the spectral colourmap is used as a second step, i.e., after `color.label2rgb`, which provides the stark colour difference of adjacent regions (with labels close in value).

In 663578ecc1df48ba9bdc0889df64d56ffc724379, I have used the latter _without_ the underlying raw image (similar side-by-side visualization):

transparent_bg

and _with_ the underlying raw image (overlay):

overlay

Adding some divergence(?) with a subsequent spectral processing step does not seem necessary. But, again, I might be missing something!

cgohlke:
> There remains the CI failure due to the tifffile, for minimal dependencies

This should be fixed in tifffile-2020.5.30, which uses a pure Python PackBits decoder in case the imagecodecs package is not installed.

jni:
@mkcor 

> I must be missing something because, as is, nipy_spectral yields gradual variations:

Yeah, it needs to be used with @emmanuelle's np.random.permutation trick from her blog post. For me, the `label2rgb` default colours are too bright and there's not enough of them! We should probably change that but that should be done in another PR. =)

jni:
@cgohlke 

> This should be fixed in tifffile-2020.5.30, which uses a pure Python PackBits decoder in case the imagecodecs package is not installed.

:heart: :heart: :heart: :heart: :heart: :pray: :pray: :pray: :pray: :pray:

emmanuelle:
> This should be fixed in tifffile-2020.5.30, which uses a pure Python PackBits decoder in case the imagecodecs package is not installed.

Awesome thanks a lot @cgohlke. @mkcor you can bump the version of `tifffile` in `requirements/default.txt` to see if Travis gets completely green.

mkcor:
> For me, the label2rgb default colours are too bright and there's not enough of them! We should probably change that but that should be done in another PR. =)

@jni ok, happy to leave the improvement of these colours to subsequent iterations!

I just wanted to acknowledge @alexdesiqueira's suggestion 
> what if we had a white background?

explicitly: I guess that, if the background is fully transparent -- as conventionally done, as @jni pointed out -- it may appear either black or white, depending on a convention(?) and/or the overlay situation.

alexdesiqueira:
@jni:
> For me, the label2rgb default colours are too bright and there's not enough of them! We should probably change that but that should be done in another PR. =)

@mkcor please don't mind me then! This PR is great as is right now :slightly_smiling_face:

mkcor:
> I think one of the figures could use `plt.contour` on the original image to visualize the result, it's a another way to visualize a binary array which is often useful.

@emmanuelle done 6819653. I thought that the segment counts would be a nice transition into the mitotic index discussion.

PS: By 'binary array' I understand that you mean mono-channel image, right?

emmanuelle:
> @emmanuelle I was about to update the changelog to acknowledge the reviewers on the biology side; just to make sure, is https://github.com/scikit-image/scikit-image/blob/master/doc/release/release_dev.rst the right file for me to update?

Yes exactly! As you're editing this file you can also add an entry about your tutorial (see for example the [changelog of 0.17 as an example](https://github.com/scikit-image/scikit-image/blob/master/doc/release/release_0.17.rst))

emmanuelle:
In the gallery the thumbnail is very dark (as in the attached picture) 
image. Maybe you could take one of the other images of the tutorial like the one with contours which is on a white background? See https://github.com/scikit-image/scikit-image/blob/master/doc/examples/transform/plot_transform_types.py#L129 for how this is done. I tried with the image #2 for example 
image

emmanuelle:
The CI failure has nothing to do with this PR (I'm working on it BTW).

doc/examples/applications/plot_human_mitosis.py
```diff
@@ -0,0 +1,207 @@
...
+# label with argument `bg_label=0`.
+
+distance = ndi.distance_transform_edt(cells)
+
+local_maxi = feature.peak_local_max(distance, indices=False,
+                                    min_distance=7)
+
+markers = measure.label(local_maxi)
+
+segmented_cells = morphology.watershed(-distance, markers, mask=cells)
```
emmanuelle, line 177: please use `segmentation.watershed` to avoid the warning

emmanuelle:
> > I think one of the figures could use `plt.contour` on the original image to visualize the result, it's a another way to visualize a binary array which is often useful.
> 
> @emmanuelle done [6819653](https://github.com/scikit-image/scikit-image/commit/6819653199da2f9a0eb52eb4b75292e5c597f337). I thought that the segment counts would be a nice transition into the mitotic index discussion.
> 
> PS: By 'binary array' I understand that you mean mono-channel image, right?

The way you've used contours is cool, indeed. I was thinking more of representing the 0.5 contour of a binary 0-1 / False-True array, so that the contour is the boundary between False and True, it can be a nice viz to superimpose such a contour over the initial image so that you can see both the image and its segmentation. But never mind it's not crucial here.

doc/examples/applications/plot_human_mitosis.py
mkcor: done 5339e62c293de6803116b1e08a8e0fa684779f04

mkcor:
Thank you for the detailed review, @emmanuelle!
I will load the original image from https://gitlab.com/scikit-image/data/ via `pooch` in a subsequent PR.
I will also follow up with the scientific reviewers.

doc/release/release_dev.rst
```diff
@@ -16,6 +16,10 @@ New Features
 ------------
 
 
+A new doc tutorial presenting a cell biology example has been added to the
+gallery (#4648). The scientific content benefited from a much appreciated
+review by Pierre Poulain and Fred Bernard, both assistant professors at
+Université de Paris and Institut Jacques Monod.
```
jni, line 22:
```suggestion
- A new doc tutorial presenting a cell biology example has been added to the
  gallery (#4648). The scientific content benefited from a much appreciated
  review by Pierre Poulain and Fred Bernard, both assistant professors at
  Université de Paris and Institut Jacques Monod.
```

requirements/default.txt
```diff
@@ -4,5 +4,5 @@ matplotlib>=2.0.0,!=3.0.0
 networkx>=2.0
 pillow>=4.3.0,!=7.1.0,!=7.1.1
 imageio>=2.3.0
-tifffile>=2019.7.26
+tifffile>=2020.5.30
```
jni, line 7: Q: should we leave the original dependency but add `tifffile>=2020.5.30` to the `docs` requirements?

doc/release/release_dev.rst
mkcor: @jni I wrote it as a paragraph to duplicate this format: https://github.com/scikit-image/scikit-image/blob/master/doc/release/release_0.17.rst#new-features (that @emmanuelle linked to). There is a list of 'actual' new features (as list items / bullet points), but the 'list' of doc / gallery additions is written in a paragraph... Are we following a definite convention then? Thanks!

emmanuelle: @mkcor the usual format is to have bullet points, for the last version I had just "bunched together" a few items in a paragraph. But I think this PR deserves its own bullet point as suggested by @jni.

requirements/default.txt
emmanuelle: sounds good to me

mkcor: Ok, done.

doc/release/release_dev.rst
mkcor: All right!

emmanuelle:
thank you all!

PR #4652, opened 2020-05-07
Update version for release

emmanuelle:
Corresponds to https://github.com/scikit-image/scikit-image/blob/master/RELEASE.txt#L65 and https://github.com/scikit-image/scikit-image/blob/master/RELEASE.txt#L71 items of release process


sciunto:
Already done in #4663

PR #4661, opened 2020-05-08
Fix label2rgb

### LINKED ISSUE #4657: Bugs in label2rgb
There seem to be one or several bugs in `label2rgb`. I discovered this when building the website doc for 0.17.0 and it might be serious enough that I want to hold the release. 

For example
```
img = np.random.randint(0, 255, (10, 10), dtype=np.uint8) 
labels = np.zeros((10, 10), dtype=np.uint8)
labels[1:3, 1:3]=1
labels[6:9, 6:9] = 2
color.label2rgb(labels, image=img, bg_color=0)
```
gives a ValueError

In the gallery `segmentation/plot_label.py` example, `label2rgb` is returning an array of floats which is bizarre, and the floats are between 0 and 255.

The broken gallery examples are
 https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_join_segmentations.html#sphx-glr-auto-examples-segmentation-plot-join-segmentations-py
https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_label.html#sphx-glr-auto-examples-segmentation-plot-label-py
https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_compact_watershed.html#sphx-glr-auto-examples-segmentation-plot-compact-watershed-py

If someone has the time to investigate that'd be awesome because I'm reluctant to release with such a regression (BTW I already pushed the stable docs for this branch but we can upload a new version after when we have a fix, it's not as bad as having a buggy version on Pypi).

emmanuelle:
Closes #4657


emmanuelle:
please do not merge as I need to add tests and also it seems that the problem is not completely solved

skimage/color/tests/test_colorlabel.py
```diff
@@ -38,6 +38,17 @@ def test_wrong_kind():
         label2rgb(label, kind='foo', bg_label=-1)
 
 
+def test_uint_image():
+    img = np.random.randint(0, 255, (10, 10), dtype=np.uint8)
+    labels = np.zeros((10, 10), dtype=np.int64)
+    labels[1:3, 1:3]=1
```
emmanuelle, line 44:
```suggestion
    labels[1:3, 1:3] = 1
```

emmanuelle:
ok this is now ready for review :-)

skimage/color/tests/test_colorlabel.py
```diff
@@ -38,6 +38,17 @@ def test_wrong_kind():
         label2rgb(label, kind='foo', bg_label=-1)
 
 
+def test_uint_image():
```
emmanuelle, line 41: this test is failing on master, img was not rescaled with `img_as_float` and therefore kept a 0-255 range while the color mask to be added was in the [0-1] range so not visible when superimposed. The output was float but in [0-255] so also matplotlib would represent it completely white in the gallery examples.

PR #4663, opened 2020-05-08
Setup longdesc markdown and switch to 0.18dev

emmanuelle:
We need a flag to tell pypi that our long description is markdown otherwise pypi refuses the wheels / bundles because it is trying to write it as rst.


setup.py
```diff
@@ -191,6 +191,7 @@ def configuration(parent_package='', top_path=None):
         name=DISTNAME,
         description=DESCRIPTION,
         long_description=LONG_DESCRIPTION,
+        long_description_content_type="text/markdown"
```
jni, line 194:
```suggestion
        long_description_content_type="text/markdown",
```

sciunto:
many thanks to both of you for the release! :)

Issue #4664, opened 2020-05-08
PermissionError when creating data_dir when user doesn't have permissions


leducvin:
## Description

Version 0.16.2 worked, but 0.17.1 tries to create a directory in whatever `appdirs` thinks is an appropriate cache directory.

For my flask app's service, the user is a no-login user, and somehow this resolves to `/var/www/`.
The user doesn't have permissions to create a directory there, so the flask app crashes with a PermissionError raised from skimage.

I specified `SKIMAGE_DATADIR=/tmp/.skimage_cache`, and ended up seeing #4660, as my flask app starts multiple processes.

## Way to reproduce
```python
import appdirs
# Monkey-patch appdirs.user_cache_dir to an unwritable location
appdirs.user_cache_dir = lambda _: '/var/log'
import skimage
```

```
Cannot create data cache folder '/var/log/0.17.1'. Will not be able to download remote data files. Use environment variable 'SKIMAGE_DATADIR' to specify another directory.
Traceback (most recent call last):
  File "skimage_issue.py", line 4, in <module>
    import skimage
  File "/home/vleduc/python/virtualenvs/py37/lib/python3.7/site-packages/skimage/__init__.py", line 133, in <module>
    from .data import data_dir
  File "/home/vleduc/python/virtualenvs/py37/lib/python3.7/site-packages/skimage/data/__init__.py", line 90, in <module>
    os.makedirs(data_dir, exist_ok=True)
  File "/home/vleduc/python/virtualenvs/py37/lib/python3.7/os.py", line 213, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "/home/vleduc/python/virtualenvs/py37/lib/python3.7/os.py", line 223, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/var/log/0.17.1'
```

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.7.5 (default, Nov  7 2019, 10:50:52) 
[GCC 8.3.0]
Linux-4.15.0-99-generic-x86_64-with-Ubuntu-18.04-bionic
scikit-image version: 0.17.1
numpy version: 1.18.4
```

hmaarrfk:
[...]

Issue #4660, opened 2020-05-08
When running in multi-process Pooch tries to create a directory that already exist


dashesy:
## Description

Previous version was good, but I cannot use new scikit-learn in a docker when running first time in multi-processing setting. Pooch tries to create same directory in first run in __init__ and that results in crash.
It is not a good idea to try to create directory in __init__, specially without any lock. Even with the lock you should be careful not to create the lock on a filesystem that does not support file locks.

## Way to reproduce
1. In a docker, when skimage is never imported before
2. Import skimage in multiple processes at once
3. This is a race condition on which process creates the cache first

```python
# Place the full code we need to recreate your issue here
from skimage import transform
```

```text
File "/miniconda/envs/py37/lib/python3.7/site-packages/skimage/__init__.py", line 133, in <module>
    from skimage import transform
  File "/miniconda/envs/py37/lib/python3.7/site-packages/skimage/__init__.py", line 133, in <module>
    from .data import data_dir
  File "/miniconda/envs/py37/lib/python3.7/site-packages/skimage/data/__init__.py", line 86, in <module>
    from .data import data_dir
  File "/miniconda/envs/py37/lib/python3.7/site-packages/skimage/data/__init__.py", line 86, in <module>
    urls=registry_urls,
  File "/miniconda/envs/py37/lib/python3.7/site-packages/pooch/core.py", line 371, in create
    urls=registry_urls,
  File "/miniconda/envs/py37/lib/python3.7/site-packages/pooch/core.py", line 371, in create
    path = make_local_storage(path, env, version)
  File "/miniconda/envs/py37/lib/python3.7/site-packages/pooch/utils.py", line 256, in make_local_storage
    path = make_local_storage(path, env, version)
    os.makedirs(path)  File "/miniconda/envs/py37/lib/python3.7/site-packages/pooch/utils.py", line 256, in make_local_storage

  File "/miniconda/envs/py37/lib/python3.7/os.py", line 221, in makedirs
    os.makedirs(path)
  File "/miniconda/envs/py37/lib/python3.7/os.py", line 221, in makedirs
        mkdir(name, mode)mkdir(name, mode)

FileExistsErrorFileExistsError: : [Errno 17] File exists: '/root/.cache/scikit-image/0.17.0'
[Errno 17] File exists: '/root/.cache/scikit-image/0.17.0'
Traceback (most recent call last):
  File "/miniconda/envs/py37/lib/python3.7/runpy.py", line 193, in _run_module_as_main
[...]

PR #4666, opened 2020-05-09
Optional pooch dependency

### LINKED ISSUE #4665: pooch is a hard requirement
## Description

I thought we wanted pooch to be a soft requirement.
Do you want me to make a PR Todo that.

It seems that got lost in a few rebases maybe. Maybe I'm forgetting a discussion 

## Version information
0.17.1

hmaarrfk:
## Description

Makes pooch an optional dependency. This was at least in my original design plan. Not sure why it never got in, I'll sift through the comments again.

TODO: Pooch contribution guide.

When merged backport with:
```
```

Note that this does NOT address #4660 or #4664

Closes  #4665 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -129,6 +129,10 @@ def _fetch(data_filename):
     KeyError:
         If the filename is not known to the scikit-image distribution.
 
+    ModuleNotFoundError:
+        If the filename is known to the scikit-image distribution but pooch
+        is not installed.
+
     ConnectionError:
         If scikit-image is unable to connect to the internet but the
         dataset has not been downloaded yet.
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -208,8 +208,8 @@ def _fetch(data_filename):
             "The requested file is part of the scikit-image distribution, "
             "but requires the installation of an optional dependency, pooch. "
             "To install pooch, use your preferred python package manager. "
-            "For example: `conda install -c conda-forge pooch`, or "
-            "`pip install -m pooch`."
+            "Follow installation instruction found at "
+            "https://scikit-image.org/docs/stable/install.html"
         )
 
     # Case 4:
@@ -256,8 +256,10 @@ def download_all(directory=None):
     This allows us to use higher quality datasets, while keeping the
     library download size small.
 
-    This function requires the installation of an optional dependency,
-    pooch, to download the full dataset.
+    This function requires the installation of an optional dependency, pooch,
+    to download the full dataset. Follow installation instruction found at
+
+        https://scikit-image.org/docs/stable/install.html
 
     Call this function to download all sample images making them available
     offline on your machine.
@@ -284,6 +286,8 @@ def download_all(directory=None):
         raise ModuleNotFoundError(
             "To download all package, scikit-image needs an optional "
             "dependency, pooch."
+            "To install pooch, follow our installation instructions found at "
+            "https://scikit-image.org/docs/stable/install.html"
         )
     # Consider moving this kind of logic to Pooch
     old_dir = image_fetcher.path
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -54,7 +54,8 @@ __all__ = ['data_dir',
            'shepp_logan_phantom',
            'stereo_motorcycle']
 
-skimage_distribution_dir = osp.join(osp.abspath(osp.dirname(__file__)), '..')
+legacy_data_dir = osp.abspath(osp.dirname(__file__))
+skimage_distribution_dir = osp.join(legacy_data_dir, '..')
 
 
 import pooch
@@ -167,6 +168,11 @@ def _fetch(data_filename):
 # Fetch all legacy data so that it is available by default
 for filename in legacy_registry:
     _fetch(filename)
+
+else:
+    # Without pooch, fallback on the standard data directory
+    # which for now, includes a few limited data samples
+    data_dir = legacy_data_dir
 
 
 def download_all(directory=None):
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -142,23 +142,18 @@ def create_image_fetcher():
 
     data_dir = osp.join(str(image_fetcher.abspath), 'data')
 
-    os.makedirs(data_dir, exist_ok=True)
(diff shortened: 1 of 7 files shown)


hmaarrfk:
@jni was there a reason why we decided to make pooch a required dependency?

https://github.com/scikit-image/scikit-image/pull/3945/commits/8e0861091341b947752f0758fabedb08d048e896

skimage/data/__init__.py
```diff
@@ -148,8 +205,20 @@ def _fetch(data_filename):
...
-    # Pooch needs to download the data. Let the image fetcher to search for our
-    # data. A ConnectionError is raised if no internet connection is available.
+    # Pooch not found.
+    if image_fetcher is None:
+        raise ModuleNotFoundError(
+            "The requested file is part of the scikit-image distribution, "
+            "but requries the installation of an optional dependency, pooch, "
+            "to be used. To install pooch, use your preffered python "
+            "package manager. Follow installation instruction found at "
+            "https://scikit-image.org/docs/stable/install.html"
```
jni, line 215:
```suggestion
            "The requested file is part of the scikit-image distribution, "
            "but requires the installation of an optional dependency, pooch. "
            "To install pooch, use your preferred python package manager. "
            "For example: `conda install -c conda-forge pooch`, or "
            "`pip install -m pooch`."
```

jni: I prefer not pointing to our instructions since they don't have any pooch-specific documentation. I would find this confusing.

skimage/data/tests/test_data.py
```diff
@@ -116,24 +107,8 @@ def test_cell():
...
-
-
-def test_data_dir():
-    astro = data.astronaut()
-    datadir = data.data_dir
-    assert 'astronaut.png' in os.listdir(datadir)
-
-
-@skipif(OFFLINE, reason='No internet connection')
-def test_download_all():
```
jni, line 134: Is there a good reason to remove this test?

skimage/data/tests/test_data.py
```diff
@@ -116,24 +107,8 @@ def test_cell():
...
 
-@skipif(OFFLINE, reason='No internet connection')
 def test_cells_3d():
     """Needs internet connection."""
     path = fetch('data/cells.tif')
     image = io.imread(path)
     assert image.shape == (60, 256, 256)
-
-
-def test_data_dir():
```
jni, line 127: Why remove this test? I feel like it's important.

jni:
@hmaarrfk it looks like I made that decision unilaterally and explicitly labeled it as such in my PR #4483, but that comment got lost in the rebase/cherry-pick dance. So it was never discussed properly. I'm happy for this PR to go in. =) Thank you!

jni:
@hmaarrfk feel free to revert the last commit, I'm not super happy with it, but I think we should be able to make the initialisation lazy, and just fall back on our legacy dir in the meantime.

hmaarrfk:
I wanted to leave the laziness to an other PR for now. Seems hard to get right

hmaarrfk:
Super thanks for the rest!

skimage/data/tests/test_data.py
hmaarrfk: Thanks, I added it back.

skimage/data/tests/test_data.py
hmaarrfk: This test really doen't play with parallel testing.... affecting the global filesystem state is not fun.

hmaarrfk: it also may induce a full download of all the files for the end users. not fun on our laptops when travelling with 4G connections.

skimage/data/__init__.py
hmaarrfk: Can we think a little longer term about this? We can work on adding instructions, lets make sure that we are doing good by future generations of developers.

hmaarrfk: I see two kinds of people:

* Advanced users: will use shortcuts to get to google, and to search `pooch python` which will give them a package, that they will know to use conda/pip automatically. These don't need any instructions.
* Novice users, that will be tempted to use `sudo pip install pooch`. 

I am trying to help the novice users by pointing them to more fleshed out instructions than a one liner (that will likely not work and ensure their frustration).

hmaarrfk: Reading over this, I strongly disagree with the instructions provided.

For example, I strongly discourage
```
conda install -c conda-forge pooch
```

I really think there should be a single place where we provide install instructions, and not adhoc like an Error message.

hmaarrfk: I raised my concerns in:
https://github.com/scikit-image/scikit-image/issues/4672

skimage/data/__init__.py
```diff
@@ -192,6 +282,13 @@ def download_all(directory=None):
     data directory by inspecting the variable `skimage.data.data_dir`.
     """
 
+    if image_fetcher is None:
+        raise ModuleNotFoundError(
+            "To download all package, scikit-image needs an optional "
```
emmanuelle, line 287:
```suggestion
            "To download all package data, scikit-image needs an optional "
```

emmanuelle:
One question related to pooch: I don't remember why we are deprecating `data.load()`? Because for image data which don't have their own function, like the `cells.tif` file, using `data.load()` will be the only way to get this image? The deprecation message mentions using `skimage.io.load` but this function does not exist. Am I missing something?

hmaarrfk:
I'm not too sure I remember why we deprecated `load`. I think the feeling was that it was a thin wrapper on `imread` and we preferred people not use it directly.

Having to use `cell()` will give you back the data, prelaoded, instead of asing you to read it again yourself.

For certain datasets, they are more complicated than "just a single image" so having a function that could return more flexible formats, like tuples was preferred.

jni:
@emmanuelle well I think the "new way" is supposed to be something like:

```python
data.download_all()
cells = io.imread(os.path.join(data.data_dir, 'data/cells.tif'))
```

which is nice because it looks more like normal IO than `data.function()`. Having said that, we should provide a public way to fetch individual images.

emmanuelle:

emmanuelle:
@jni thanks for the explanations. Yes, I agree that we should also have a way to download and open individual files. Probably not high priority since we indeed have `download_all` (which at the moment is still reasonable).

hmaarrfk:
i guess the easy way didn't work.

PR #4670, opened 2020-05-10
[WIP] Make Pooch lazily create directories for python 3.7+

hmaarrfk:
## Description

@jni just opening this up as a potential solution for #4660

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -130,7 +130,17 @@ else:
                              img_as_ubyte,
                              img_as_bool,
                              dtype_limits)
-    from .data import data_dir
     from .util.lookfor import lookfor
+
+    if sys.version_info < (3, 7):
+        from .data import data_dir
+
+def __getattr__(name):
+    if name == 'data_dir':
+        from .data import data_dir
+        return data_dir
+    else:
+        raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
+
 
 del sys
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -51,12 +51,17 @@ __all__ = ['data_dir',
            'shepp_logan_phantom',
            'stereo_motorcycle']
 
+_IMAGE_FETCHER = None
 legacy_data_dir = osp.abspath(osp.dirname(__file__))
+_data_dir = legacy_data_dir
 skimage_distribution_dir = osp.join(legacy_data_dir, '..')
 
+has_pooch = True
 try:
+    import pooch
     from pooch.utils import file_hash
 except ModuleNotFoundError:
+    has_pooch = False
     # Function taken from
     # https://github.com/fatiando/pooch/blob/master/pooch/utils.py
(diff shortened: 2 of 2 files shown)


skimage/__init__.py
```diff
@@ -130,7 +130,17 @@ def _raise_build_error(e):
                              img_as_ubyte,
                              img_as_bool,
                              dtype_limits)
-    from .data import data_dir
     from .util.lookfor import lookfor
 
+    if sys.version_info < (3, 7):
+        from .data import data_dir
+
+def __getattr__(name):
```
emmanuelle, line 138: very nice use of py3.7+ lazy loading :-).

hmaarrfk: Yeah, lets work on this in a bit. I want to get this PR to be really tiny, since it is hard to test in a pytest way.

jni:
@hmaarrfk thanks! Can you rebase now that #4666 is merged?

`lazy_pooch` is a great branch name, reminds me of my greyhound. =D This is him right now:

IMG_4310

😃

skimage/__init__.py
jni: Honestly that was my quick and dirty attempt. I am convinced there is an elegant solution here without even global variables, but it's been a long week (gave a skimage+napari workshop today) and I'm just not seeing it. I'm happy to have another pass at this in a few days after recharging my batteries, but anyone should feel free to give it a go in the meantime!

hmaarrfk: I think image_fetcher needs to be made a class.... Was thinking about this problem in the shower and you know that is where the best ideas come!

hmaarrfk:
ok i rebased, but i must admit it needs serious review in light of some organizational changes in the original branch. sorry about that one.

I'm also still not convinced that `exist_ok=True` doesn't fix things for the affected user in the original issue.

emmanuelle:
@jni @hmaarrfk if you think this PR is going to take some time to get merged in I can also release 0.17.2 with pooch an optional dependency and 0.17.3 after this one is merged. Please tell me!

jni:
@emmanuelle please do it! Optional pooch will already solve the problem for the OP of #4660 since they are not using the sample data at all.

skimage/__init__.py
jni: You're right. The whole problem is that we have to keep track of a lot of state — that's what classes are for! I was stubbornly treating the whole module as a big class.

hmaarrfk:
Definitely do it. This will help people building on automated machines for sure. Unless they use conda...

sciunto:
THe issue #4660 is solved. Is this still useful in any sense?

hmaarrfk:
I think the correct solution is a two pronged approach:

Leaving the `data_dir` pointing to the old datasets. 

Allowing users that have pooch installed use the newer, larger datasets.

I'm not super convinced, I have all the use cases right, I'll try to write a spec.

PR #4681, opened 2020-05-11
Reintroduced convert with a strong deprecation warning

### LINKED ISSUE #4680: Reintroduce `dtype.convert`
## Description

`dtype.convert` used to be a very useful function in older versions of scikit-image.

In 0.17 it was decided that it should be removed without a deprecation cycle.

Not fun.

Could we get it back in 0.17.2 with a deprecation cycle?
https://github.com/scikit-image/scikit-image/pull/4590
If so, i can make the PR.

hmaarrfk:
## Description
This PR reintroduces `convert` with a strong warning to not use it.

I know we are working toward 1.0, but this broke my code.

I guess most people use the public API, but I was not keep on making my own if/elif statement:

Closes #4680

@sciunto @emmanuelle 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/util/__init__.py
+++ b/skimage/util/__init__.py
@@ -3,7 +3,7 @@ import warnings
 import numpy as np
 from .dtype import (img_as_float32, img_as_float64, img_as_float,
                     img_as_int, img_as_uint, img_as_ubyte,
-                    img_as_bool, dtype_limits)
+                    img_as_bool, convert, dtype_limits)
 from .shape import view_as_blocks, view_as_windows
 from .noise import random_noise
 from .apply_parallel import apply_parallel
@@ -32,6 +32,7 @@ __all__ = ['img_as_float32',
            'img_as_uint',
            'img_as_ubyte',
            'img_as_bool',
+           'convert',
            'dtype_limits',
            'view_as_blocks',
            'view_as_windows',
--- a/skimage/util/__init__.py
+++ b/skimage/util/__init__.py
@@ -3,7 +3,7 @@ import warnings
 import numpy as np
 from .dtype import (img_as_float32, img_as_float64, img_as_float,
                     img_as_int, img_as_uint, img_as_ubyte,
-                    img_as_bool, convert, dtype_limits)
+                    img_as_bool, dtype_limits)
 from .shape import view_as_blocks, view_as_windows
 from .noise import random_noise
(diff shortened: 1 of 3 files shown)


alexdesiqueira:
Just out of curiosity, what code did break because of (the lack of) this, @hmaarrfk?

hmaarrfk:
i let the `convert` function do all the logic switching for me.

I guess most people do

```
if dytpe == float32:
    img_as_float32
elif dtype == float64:
    img_as_float64
```

Seems a little silly seeing as they all just call convert anyway. So i just call convert like `asdtype`

sciunto:
I think I'm -1 on this. You were not supposed to use this function, it was not in the API. The alternative, is to do what I originally did in my PR: put it in the  API.

hmaarrfk:
FWIW,after this PR it is in the API albeit with a strong warning.

jni:
@hmaarrfk I really don't understand the logic here. If we are planning to destroy the function, and you acknowledge as much in the warning, your client code will need to adapt anyway. You might as well just use the private function, with a try/except for importing `_convert`. Then at least people reading your code will know that you are playing with fire. :grimacing: 

Having said this, I don't have very strong opposition to this. But I prefer going back to the old state of semi-private rather than making it fully public. We should aim to *reduce* the 1.0 disruption, not throw in the towel and decide that APIs don't matter anymore since we are planning to break them all soon.

hmaarrfk:
These concerns are all fair. And through searching on github, it seems that i am probably the only one using this function.

I understand that in 1.0, we can remove this function.

That said, today, this is a super useful function. I routinely have

```
arr = np.zeros(...., dtype='my_type')
```
and in my loading functions
```
arr = load_my_image(...., dtype='my_type')
```
where inside `load_my_image`, we use `skimage.util.dtype.convert` to really avoid the slew of if statements.

At this stage, I could go either way.

I don't think we should spend much more reviewer cognitive task, I won't take offense if somebody simply decides to close this PR.

emmanuelle:
I feel like the current state of the PR is an acceptable trade-off, ie
- the function is still not public
- `convert` is back but calls `_convert` and issues a big warning. 
So I'm approving and merging because we should not spend too much time on this.

emmanuelle:

hmaarrfk:
thx.

PR #4339, opened 2019-12-04
Fix single precision cast to double in slic

rfezzani:
## Description

Closes #4274.

This PR the fused type `np_floats` in `_slic_cython` and care is taken in `slic` to not convert the input image to double precision if it is initially single precision.

Here are some measured timings showing the performance improvement:

```python
In [2]: from skimage.data import astronaut                                                               

In [3]: from skimage.util import img_as_float                                                            

In [4]: from skimage.segmentation import slic                                                            

In [5]: img = img_as_float(astronaut()[::2, ::2])                                                        

In [6]: img_f32 = img.astype('float32')                                                                  

In [7]: %timeit slic(img, n_segments=250, compactness=10, sigma=1)                                       
79 ms ± 248 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit slic(img_f32, n_segments=250, compactness=10, sigma=1)                                   
74.6 ms ± 132 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [9]: img = img_as_float(astronaut())                                                                  

In [10]: img_f32 = img.astype('float32')                                                                 

In [11]: %timeit slic(img, n_segments=250, compactness=10, sigma=1)                                      
309 ms ± 2.51 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [12]: %timeit slic(img_f32, n_segments=250, compactness=10, sigma=1)                                  
285 ms ± 973 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## Checklist

- [x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Gallery example in `./doc/examples` (new features only)
- [ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
- [ ] Unit tests
- [x] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

## For reviewers

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.
- [ ] Consider backporting the PR with `@meeseeksdev backport to v0.14.x`

PR #4683, opened 2020-05-11
Fix slic float32 support regression

### LINKED ISSUE #4682: slic float32 support regression
## Description

As reported by @jklaise in https://github.com/scikit-image/scikit-image/issues/4675#issuecomment-626794531, `slic` no more supports `float32`.

## Way to reproduce
```python
from skimage.segmentation import slic
import numpy as np
im = np.random.rand(28, 28).astype(np.float32)
slic(im)
```

## Version information

```python
3.8.2 (default, Mar 26 2020, 15:53:00) 
[GCC 7.3.0]
Linux-4.4.0-178-generic-x86_64-with-glibc2.10
scikit-image version: 0.16.2
numpy version: 1.18.1
```

rfezzani:
## Description

Fixes #4682 

The use of `np_floats` fused type was already introduced in #4339, but the support for `float32` was not tested :confused: 

`foat32` support is added in this PR and a test for dtype support is added.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/segmentation/slic_superpixels.py
+++ b/skimage/segmentation/slic_superpixels.py
@@ -272,14 +272,14 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
     n_centroids = centroids.shape[0]
     segments = np.ascontiguousarray(np.concatenate(
         [centroids, np.zeros((n_centroids, image.shape[3]))],
-        axis=-1))
+        axis=-1), dtype=dtype)
 
     # Scaling of ratio in the same way as in the SLIC paper so the
     # values have the same meaning
     step = max(steps)
     ratio = 1.0 / compactness
 
-    image = np.ascontiguousarray(image * ratio, dtype=np.double)
+    image = np.ascontiguousarray(image * ratio)
 
     if update_centroids:
         # Step 2 of the algorithm [3]_
--- a/skimage/segmentation/slic_superpixels.py
+++ b/skimage/segmentation/slic_superpixels.py
@@ -279,7 +279,7 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
     step = max(steps)
     ratio = 1.0 / compactness
 
-    image = np.ascontiguousarray(image * ratio)
+    image = np.ascontiguousarray(image * ratio, dtype=dtype)
 
     if update_centroids:
         # Step 2 of the algorithm [3]_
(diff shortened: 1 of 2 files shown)


skimage/segmentation/tests/test_slic.py
```diff
@@ -463,3 +464,10 @@ def test_gray_3d_mask():
     assert_equal(len(np.unique(seg)), 9)
     for s, c in zip(slices, range(1, 9)):
         assert_equal(seg[s][2:-2, 2:-2, 2:-2], c)
+
+
+@pytest.mark.parametrize("dtype", ['float32', 'float64', 'uint8', 'int'])
+def test_dtype_support(dtype):
+    img = np.random.rand(28, 28).astype(dtype)
+
+    assert slic(img).dtype == int
```
hmaarrfk, line 473:
```suggestion
    assert slic(img).dtype == np.dtype(int)
```

skimage/segmentation/slic_superpixels.py
```diff
@@ -272,14 +272,14 @@ def slic(image, n_segments=100, compactness=10., max_iter=10, sigma=0,
...
         [centroids, np.zeros((n_centroids, image.shape[3]))],
-        axis=-1))
+        axis=-1), dtype=dtype)
 
     # Scaling of ratio in the same way as in the SLIC paper so the
     # values have the same meaning
     step = max(steps)
     ratio = 1.0 / compactness
 
-    image = np.ascontiguousarray(image * ratio, dtype=np.double)
```
alexdesiqueira, line 282: @rfezzani a question: shouldn't this `dtype` continue here as `dtype=dtype`? Or else, if I understood correctly, `image` will return with `dtype=None` (from [np.ascontiguousarray()](https://numpy.org/doc/stable/reference/generated/numpy.ascontiguousarray.html) default `dtype` value) . Is that right? Thanks for clarifying!

rfezzani: That's absolutely right @alexdesiqueira, but I will set dtype explicit for caution ;)

rfezzani: Done ;)

rfezzani:
Thank you @hmaarrfk ;)

emmanuelle:
Note to self: to be backported to 0.17.x

emmanuelle:

rfezzani:
Thank you @emmanuelle

alexdesiqueira:
Thank you Riadh!

PR #4689, opened 2020-05-12
In release notes, better describe skimage's relationship to ecosystem

stefanv:
I think scikit-image can no longer accurately be described as an image processing toolbox *for* SciPy.  Rather, it is a library that relies on SciPy to provide a wide array of image processing algorithms.


emmanuelle:
Thank you @stefanv merging.

PR #4703, opened 2020-05-12
Add exception for PermissionError at data_dir creation

### LINKED ISSUE #4664: PermissionError when creating data_dir when user doesn't have permissions

rfezzani:
## Description

Fixes #4664.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -235,7 +235,8 @@ def _init_pooch():
     try:
         os.makedirs(data_dir, exist_ok=True)
     except PermissionError:
-        data_dir = tempfile.gettempdir()
+        data_dir = osp.join(tempfile.gettempdir(), 'skimage_data')
+        os.makedirs(data_dir, exist_ok=True)
     shutil.copy2(osp.join(skimage_distribution_dir, 'data', 'README.txt'),
                  osp.join(data_dir, 'README.txt'))
 
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -5,7 +5,7 @@ For more images, see
  - http://sipi.usc.edu/database/database.php
 
 """
-import sys
+import tempfile
 from warnings import warn
 import numpy as np
 import shutil
@@ -139,7 +139,6 @@ def create_image_fetcher():
 
     data_dir = osp.join(str(image_fetcher.abspath), 'data')
 
-
     return image_fetcher, data_dir
 
 
@@ -149,6 +148,7 @@ if image_fetcher is None:
     has_pooch = False
 else:
     has_pooch = True
+
 
 def _fetch(data_filename):
     """Fetch a given data file from either the local cache or the repository.
@@ -231,11 +231,14 @@ def _fetch(data_filename):
 
 
 def _init_pooch():
-    os.makedirs(data_dir, exist_ok=True)
+    global data_dir
+    try:
+        os.makedirs(data_dir, exist_ok=True)
+    except PermissionError:
+        data_dir = tempfile.gettempdir()
     shutil.copy2(osp.join(skimage_distribution_dir, 'data', 'README.txt'),
                  osp.join(data_dir, 'README.txt'))
 
-    data_base_dir = osp.join(data_dir, '..')
     # Fetch all legacy data so that it is available by default
     for filename in legacy_registry:
         _fetch(filename)


rfezzani:
@hmaarrfk, do you think that this is an acceptable solution to #4664?

hmaarrfk:
what about returning `none` in `image_fetcher` and setting deta_dir to the fallback?

hmaarrfk:
This is OK.

Truthfully, this `image_fetcher` needs to be our own class. where we can store all these global variables...

hmaarrfk:
I think you should maybe look into a solution for both this and:
https://github.com/scikit-image/scikit-image/pull/4670

at the same time.

They are likely one and the same.

skimage/data/__init__.py
```diff
@@ -231,11 +231,15 @@ def _fetch(data_filename):
 
 
 def _init_pooch():
-    os.makedirs(data_dir, exist_ok=True)
+    global data_dir
+    try:
+        os.makedirs(data_dir, exist_ok=True)
+    except PermissionError:
+        data_dir = osp.join(tempfile.gettempdir(), 'skimage_data')
```
soupault, line 238: If I am not mistaken, this is going to download all the data again and again on system reboot. I guess the question is: do we want this or do we rather raise a PermissionError. I am in favour of the latter option.

rfezzani: I think that the data are lazy loaded so it is not supposed to be downloaded at system reboot. But if the user uses any image from the `data` module in his code, it will be in fact downloaded at each system reboot.
But I agree with you, this operation must not be pass silently and at least a warning must be raised.
I am not in favor of error raising since this bug prevents the use of `skimage` and `pooch` conjointly even if the `data` module is not used by the user.

hmaarrfk: I'm leaning toward the following proposal:
* `data_dir` always points to `legacy_data_dir`.
    * This retains old behavior as expected
    * You won't be able to use `os.listdir` to find the **new** donwloaded datasets. This was always confusing anyway.
    * We should be able to remove the readme at this stage.
* New datasets, are simply unavailable without pooch.
* Large datasets are simply unavailble if we don't have a place to put them.

hmaarrfk: In this way, directories do not have to be created until we **need** them, and we only **need** it for new, functionality, that doesn't occur at import time.

The old strategy had the advantage of:
* New and old datasets are in the same location.

I don't see a way to have both advantages at once (without seriously hacking and changing the value of `data_dir` from underneath somebody)

Issue #4629, opened 2020-04-29
separate_stains output issue


asmagen:
I am using separate_stains as follows on an RGB histopathology image and the results aren't sensical because all of the values in each channel are negative.
`ihc_hax = separate_stains(rgb_original, hax_from_rgb)`

See the following input/suggestions in regards to the possible issues with the function [here](https://stackoverflow.com/questions/61092050/identifying-positive-pixels-after-color-deconvolution-ignoring-boundaries?noredirect=1#comment108080962_61092050) for the following comment given to me:
> Compare the implementation in skimage with what is described in the original paper. You'll see several errors in the implementation, most importantly the lack of a division by the max intensity. They should have used -np.log10(rgb/255) (assuming that 255 is the illumination intensity), rater than -np.log10(rgb). – Cris Luengo Apr 20 at 22:47

and the following comment from [here](https://stackoverflow.com/questions/61297083/separate-stains-output-of-one-channel-and-conversion-to-grayscale/61298083?noredirect=1#comment108453264_61298083) in regards to my observation of only negative values:
>  That makes no sense. I don’t have experience with the functions you’re using here, so cannot verify their correctness. Negative values should be fairly rare in the stain unmixing process if stain definitions used are good approximations. – Cris Luengo Apr 19 at 14:08

Reproducible example:
```
>>> import numpy as np
>>> from skimage.color import hax_from_rgb
>>> import skimage
>>> image = skimage.io.imread(work_dir+'/Figure_1.png')
>>> ihc_hax = separate_stains(image, hax_from_rgb)
>>> aec = ihc_hax[:, :, 1]
>>> np.min(aec)
-0.42166126533003073
>>> np.max(aec)
-0.26432206141264536
```

How can we address and resolve this issue?

![Figure_1](https://user-images.githubusercontent.com/11299568/80645415-134ffe80-8a39-11ea-91b8-256e2f32c17e.png)

asmagen:
Any thoughts @alexdesiqueira?

alexdesiqueira:
Hey @asmagen, thank you for your example and clarifications!
I will check our algorithm as soon as possible, based on that info. Let me look up the paper as well, and I'll get back to you with what I found, ok?
Thanks again!

asmagen:
Yes, thank you. I hope to make progress with it ASAP. There's a link to the paper in the original comment (follow the link to stack exchange).

[...]

PR #4725, opened 2020-05-15
Fixing color/colorconv/separate_stains

### LINKED ISSUE #3858: separate_stains output
## Description

The transformed images of ImageJ color deconvolution plugin and skimage.color.separate_stains() are not the same, while transformation matrices are the same. In my case, after separate_stains(img, bex_from_rgb) eosin channel looks like inverted to ImageJ and all the output values are negative.
Screenshot from 2019-04-24 11-02-34

The following lines
```python
print(img_bex.min())
print(img_bex.max())
```
outputs
```
-0.550177897460211
-0.2117949505092822
```
So how to correctly convert values after transformation to uint8?

## Way to reproduce
```python
import numpy as np
from skimage.color import separate_stains, bex_from_rgb
from matplotlib.colors import LinearSegmentedColormap
from skimage import img_as_float, img_as_ubyte
from skimage import data

import cv2

img = cv2.imread('./40x_col/Snap-450.png')
img = cv2.resize(img, (0, 0), fx=0.4, fy=0.4)
img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)

# Create an artificial color close to the original one
cmap_hema = LinearSegmentedColormap.from_list('mycmap', ['white', 'navy'])
# cmap_dab = LinearSegmentedColormap.from_list('mycmap', ['white',
#                                              'saddlebrown'])
cmap_eosin = LinearSegmentedColormap.from_list('mycmap', ['darkviolet',
                                               'white'])

img_bex = separate_stains(img, bex_from_rgb)



print(img_bex.min())
print(img_bex.max())

fig, axes = plt.subplots(2, 2, figsize=(7, 6), sharex=True, sharey=True)
ax = axes.ravel()

ax[0].imshow(img)
ax[0].set_title("Original image")

ax[1].imshow(img_bex[:, :, 0], cmap=cmap_hema)
ax[1].set_title("Hematoxylin")

ax[2].imshow(img_bex[:, :, 1], cmap=cmap_eosin)
ax[2].set_title("Eosin")

ax[3].imshow(img_bex[:, :, 2], cmap='gray')
ax[3].set_title("...")

for a in ax.ravel():
    a.axis('off')

fig.tight_layout()

plt.show()

```

## Version information
```
3.6.8 |Anaconda, Inc.| (default, Dec 30 2018, 01:22:34) 
[GCC 7.3.0]
Linux-4.15.0-47-generic-x86_64-with-debian-buster-sid
scikit-image version: 0.15.0
numpy version: 1.14.2
```
UPD
Snap-450.png
Snap-450

alexdesiqueira:
## Description

Fixes #3858 , #4629.

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1448,7 +1448,10 @@ def separate_stains(rgb, conv_matrix):
     References
     ----------
     .. [1] https://web.archive.org/web/20160624145052/http://www.mecourse.com/landinig/software/cdeconv/cdeconv.html
-    .. [2] https://github.com/jnkather/ColorDeconvolutionMatlab
+    .. [2] A. C. Ruifrok and D. A. Johnston, “Quantification of histochemical
+           staining by color deconvolution,” Anal. Quant. Cytol. Histol., vol.
+           23, no. 4, pp. 291–299, Aug. 2001.
+    .. [3] https://github.com/DIPlib/diplib/
 
     Examples
     --------
@@ -1460,13 +1463,12 @@ def separate_stains(rgb, conv_matrix):
     from ..exposure import rescale_intensity
 
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb += np.e  # avoiding log artifacts
+    rgb += np.maximum(rgb, 1E-6)  # avoiding log artifacts
+    log_adjust = -np.log10(1E-6)  # used to compensate the sum above
 
-    stains = np.reshape(-np.log(rgb), (-1, 3)) @ conv_matrix
-    stains = np.reshape(stains, rgb.shape)
-    stains = rescale_intensity(stains, out_range=(0, 1))
+    stains = np.reshape(-np.log10(rgb) / log_adjust, (-1, 3)) @ conv_matrix
 
-    return invert(stains)
+    return np.reshape(stains, rgb.shape)
 
 
 def combine_stains(stains, conv_matrix):
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1459,7 +1459,7 @@ def separate_stains(rgb, conv_matrix):
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     """
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb += np.exp(1)  # avoiding log artifacts
+    rgb += np.e  # avoiding log artifacts
 
     stains = np.reshape(-np.log(rgb), (-1, 3)) @ conv_matrix
     stains = np.reshape(stains, rgb.shape)
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1464,12 +1464,12 @@ def separate_stains(rgb, conv_matrix):
 
     rgb = _prepare_colorarray(rgb, force_copy=True)
     np.maximum(rgb, 1E-6, out=rgb)  # avoiding log artifacts
-    log_adjust = -np.log10(1E-6)  # used to compensate the sum above
+    log_adjust = -np.log(1E-6)  # used to compensate the sum above
 
-    stains = (-np.log10(rgb) / log_adjust) @ conv_matrix
+    stains = (-np.log(rgb) / log_adjust) @ conv_matrix
     stains = np.reshape(stains, rgb.shape)
 
-    return np.clip(stains, a_min=0, a_max=1)
+    return stains
 
 
 def combine_stains(stains, conv_matrix):
@@ -1513,6 +1513,9 @@ def combine_stains(stains, conv_matrix):
     References
     ----------
     .. [1] https://web.archive.org/web/20160624145052/http://www.mecourse.com/landinig/software/cdeconv/cdeconv.html
+    .. [2] A. C. Ruifrok and D. A. Johnston, “Quantification of histochemical
+           staining by color deconvolution,” Anal. Quant. Cytol. Histol., vol.
+           23, no. 4, pp. 291–299, Aug. 2001.
 
     Examples
     --------
@@ -1526,11 +1529,12 @@ def combine_stains(stains, conv_matrix):
     from ..exposure import rescale_intensity
 
     stains = _prepare_colorarray(stains)
-    logrgb2 = -np.reshape(stains, (-1, 3)) @ conv_matrix
-    rgb2 = np.power(10, logrgb2)
-    log_adjust = -np.log10(1E-6)  # used to compensate the sum above
-    return rescale_intensity(np.reshape(rgb2 * log_adjust, stains.shape),
-                             in_range=(0, 1))
+    # log_adjust here is used to compensate the sum within separate_stains().
+    log_adjust = -np.log(1E-6)
+    log_rgb = -(stains * log_adjust) @ conv_matrix
+    rgb = np.power(np.e, log_rgb)
+
+    return rescale_intensity(rgb, in_range=(0, 1))
 
 
 def lab2lch(lab):
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1467,8 +1467,9 @@ def separate_stains(rgb, conv_matrix):
     log_adjust = -np.log10(1E-6)  # used to compensate the sum above
 
     stains = np.reshape(-np.log10(rgb) / log_adjust, (-1, 3)) @ conv_matrix
+    stains = np.reshape(stains, rgb.shape)
 
-    return np.reshape(stains, rgb.shape)
+    return np.clip(stains, a_min=0, a_max=1)
 
 
 def combine_stains(stains, conv_matrix):
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -54,7 +54,6 @@ import functools
 import numpy as np
 from warnings import warn
 from scipy import linalg
-from ..exposure import rescale_intensity
 from ..util import dtype, dtype_limits, invert
 
 
@@ -1458,6 +1457,8 @@ def separate_stains(rgb, conv_matrix):
     >>> ihc = data.immunohistochemistry()
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     """
+    from ..exposure import rescale_intensity
+
     rgb = _prepare_colorarray(rgb, force_copy=True)
     rgb += np.e  # avoiding log artifacts
 
@@ -1519,6 +1520,8 @@ def combine_stains(stains, conv_matrix):
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     >>> ihc_rgb = combine_stains(ihc_hdx, rgb_from_hdx)
     """
+    from ..exposure import rescale_intensity
+
     stains = _prepare_colorarray(stains)
     logrgb2 = -np.reshape(stains, (-1, 3)) @ conv_matrix
     rgb2 = np.power(10, logrgb2)
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -54,7 +54,8 @@ import functools
 import numpy as np
 from warnings import warn
 from scipy import linalg
-from ..util import dtype, dtype_limits
+from ..exposure import rescale_intensity
+from ..util import dtype, dtype_limits, invert
 
 
 def guess_spatial_dimensions(image):
@@ -1448,6 +1449,7 @@ def separate_stains(rgb, conv_matrix):
     References
     ----------
     .. [1] https://web.archive.org/web/20160624145052/http://www.mecourse.com/landinig/software/cdeconv/cdeconv.html
+    .. [2] https://github.com/jnkather/ColorDeconvolutionMatlab
 
     Examples
     --------
@@ -1457,9 +1459,13 @@ def separate_stains(rgb, conv_matrix):
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     """
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb += 2
-    stains = np.reshape(-np.log10(rgb), (-1, 3)) @ conv_matrix
-    return np.reshape(stains, rgb.shape)
+    rgb += np.exp(1)  # avoiding log artifacts
+
+    stains = np.reshape(-np.log(rgb), (-1, 3)) @ conv_matrix
+    stains = np.reshape(stains, rgb.shape)
+    stains = rescale_intensity(stains, out_range=(0, 1))
+
+    return invert(stains)
 
 
 def combine_stains(stains, conv_matrix):
@@ -1513,8 +1519,6 @@ def combine_stains(stains, conv_matrix):
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     >>> ihc_rgb = combine_stains(ihc_hdx, rgb_from_hdx)
     """
-    from ..exposure import rescale_intensity
-
     stains = _prepare_colorarray(stains)
     logrgb2 = -np.reshape(stains, (-1, 3)) @ conv_matrix
     rgb2 = np.power(10, logrgb2)
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1463,7 +1463,7 @@ def separate_stains(rgb, conv_matrix):
     from ..exposure import rescale_intensity
 
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb += np.maximum(rgb, 1E-6)  # avoiding log artifacts
+    rgb = np.maximum(rgb, 1E-6)  # avoiding log artifacts
     log_adjust = -np.log10(1E-6)  # used to compensate the sum above
 
     stains = np.reshape(-np.log10(rgb) / log_adjust, (-1, 3)) @ conv_matrix
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1535,7 +1535,7 @@ def combine_stains(stains, conv_matrix):
     log_rgb = -(stains * log_adjust) @ conv_matrix
     rgb = np.exp(log_rgb)
 
-    return rescale_intensity(rgb, in_range=(0, 1))
+    return np.clip(rgb, a_min=0,  a_max=1)
 
 
 def lab2lch(lab):
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1527,9 +1527,8 @@ def combine_stains(stains, conv_matrix):
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     >>> ihc_rgb = combine_stains(ihc_hdx, rgb_from_hdx)
     """
-    from ..exposure import rescale_intensity
-
     stains = _prepare_colorarray(stains)
+
     # log_adjust here is used to compensate the sum within separate_stains().
     log_adjust = -np.log(1E-6)
     log_rgb = -(stains * log_adjust) @ conv_matrix
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1535,7 +1535,7 @@ def combine_stains(stains, conv_matrix):
     log_rgb = -(stains * log_adjust) @ conv_matrix
     rgb = np.exp(log_rgb)
 
-    return np.clip(rgb, a_min=0,  a_max=1)
+    return np.clip(rgb, a_min=0, a_max=1)
 
 
 def lab2lch(lab):
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1531,7 +1531,7 @@ def combine_stains(stains, conv_matrix):
     log_rgb = -(stains * log_adjust) @ conv_matrix
     rgb = np.exp(log_rgb)
 
-    return rgb
+    return rescale_intensity(rgb, in_range=(0, 1))
 
 
 def lab2lch(lab):
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1445,13 +1445,17 @@ def separate_stains(rgb, conv_matrix):
     * ``ahx_from_rgb``: Alcian Blue + Hematoxylin
     * ``hpx_from_rgb``: Hematoxylin + PAS
 
+    This implementation borrows some ideas from DIPlib [2]_, e.g. the
+    compensation using a small value to avoid log artifacts when
+    calculating the Beer-Lambert law.
+
     References
     ----------
     .. [1] https://web.archive.org/web/20160624145052/http://www.mecourse.com/landinig/software/cdeconv/cdeconv.html
-    .. [2] A. C. Ruifrok and D. A. Johnston, “Quantification of histochemical
+    .. [2] https://github.com/DIPlib/diplib/
+    .. [3] A. C. Ruifrok and D. A. Johnston, “Quantification of histochemical
            staining by color deconvolution,” Anal. Quant. Cytol. Histol., vol.
            23, no. 4, pp. 291–299, Aug. 2001.
-    .. [3] https://github.com/DIPlib/diplib/
 
     Examples
     --------
@@ -1462,9 +1466,9 @@ def separate_stains(rgb, conv_matrix):
     """
     rgb = _prepare_colorarray(rgb, force_copy=True)
     np.maximum(rgb, 1E-6, out=rgb)  # avoiding log artifacts
-    log_adjust = -np.log(1E-6)  # used to compensate the sum above
+    log_adjust = np.log(1E-6)  # used to compensate the sum above
 
-    stains = (-np.log(rgb) / log_adjust) @ conv_matrix
+    stains = (np.log(rgb) / log_adjust) @ conv_matrix
 
     return stains
 
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -54,7 +54,7 @@ import functools
 import numpy as np
 from warnings import warn
 from scipy import linalg
-from ..util import dtype, dtype_limits, invert
+from ..util import dtype, dtype_limits
 
 
 def guess_spatial_dimensions(image):
@@ -1466,7 +1466,7 @@ def separate_stains(rgb, conv_matrix):
     rgb = np.maximum(rgb, 1E-6)  # avoiding log artifacts
     log_adjust = -np.log10(1E-6)  # used to compensate the sum above
 
-    stains = np.reshape(-np.log10(rgb) / log_adjust, (-1, 3)) @ conv_matrix
+    stains = (-np.log10(rgb) / log_adjust) @ conv_matrix
     stains = np.reshape(stains, rgb.shape)
 
     return np.clip(stains, a_min=0, a_max=1)
@@ -1528,8 +1528,9 @@ def combine_stains(stains, conv_matrix):
     stains = _prepare_colorarray(stains)
     logrgb2 = -np.reshape(stains, (-1, 3)) @ conv_matrix
     rgb2 = np.power(10, logrgb2)
-    return rescale_intensity(np.reshape(rgb2 - 2, stains.shape),
-                             in_range=(-1, 1))
+    log_adjust = -np.log10(1E-6)  # used to compensate the sum above
+    return rescale_intensity(np.reshape(rgb2 * log_adjust, stains.shape),
+                             in_range=(0, 1))
 
 
 def lab2lch(lab):
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1463,7 +1463,7 @@ def separate_stains(rgb, conv_matrix):
     from ..exposure import rescale_intensity
 
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb = np.maximum(rgb, 1E-6)  # avoiding log artifacts
+    np.maximum(rgb, 1E-6, out=rgb)  # avoiding log artifacts
     log_adjust = -np.log10(1E-6)  # used to compensate the sum above
 
     stains = (-np.log10(rgb) / log_adjust) @ conv_matrix
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1529,9 +1529,9 @@ def combine_stains(stains, conv_matrix):
     # log_adjust here is used to compensate the sum within separate_stains().
     log_adjust = -np.log(1E-6)
     log_rgb = -(stains * log_adjust) @ conv_matrix
-    rgb = np.power(np.e, log_rgb)
+    rgb = np.exp(log_rgb)
 
-    return rescale_intensity(rgb, in_range=(0, 1))
+    return rgb
 
 
 def lab2lch(lab):
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -1460,14 +1460,11 @@ def separate_stains(rgb, conv_matrix):
     >>> ihc = data.immunohistochemistry()
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     """
-    from ..exposure import rescale_intensity
-
     rgb = _prepare_colorarray(rgb, force_copy=True)
     np.maximum(rgb, 1E-6, out=rgb)  # avoiding log artifacts
     log_adjust = -np.log(1E-6)  # used to compensate the sum above
 
     stains = (-np.log(rgb) / log_adjust) @ conv_matrix
-    stains = np.reshape(stains, rgb.shape)
 
     return stains
 


alexdesiqueira:
Alright everyone, hold on to your code, this will be extensive :blush: 

The original issue, https://github.com/scikit-image/scikit-image/issues/3858, presents some context. We kinda don't want (that much) negative numbers on stains. The `...` coordinate, called third component somewhere, or [residual](https://github.com/jnkather/ColorDeconvolutionMatlab) by @jnkather, is essentially noise. As I understoof, we want as little as we can on that coord.

The second issue, https://github.com/scikit-image/scikit-image/issues/4629, came with some code and context as well. In this case, @asmagen had checked for some help on [StackOverflow](https://stackoverflow.com/questions/61092050/identifying-positive-pixels-after-color-deconvolution-ignoring-boundaries). Some comments:

> @Assafb: (...) What I don't understand in that code is the line rgb += 2. I don't know where that comes from or why they use it. I'm 100% sure it is wrong. I guess they're trying to avoid log10(0), but that should be done differently. I bet this is where your negative values come from, though.

> @Assafb: Compare the implementation in skimage with what is described in the original paper. You'll see several errors in the implementation, most importantly the lack of a division by the max intensity. They should have used -np.log10(rgb/255) (assuming that 255 is the illumination intensity), rater than -np.log10(rgb).

I checked the original paper [(Ruifrok & Johnston, 2001)](https://www.ncbi.nlm.nih.gov/pubmed/11531144), and they describe that `-np.log10(rgb/255)` as `OD_C = -log10(I_C/I_0,C) = A * c_C `, with `OD` being the optical density, `A` being the amount of stain, and `c` being the absorption factor; `C` is the channel. That translates to us as `I -> rgb`, `c -> conv_matrix` in our code. Since our input image has datatype float, it does not seem necessary to divide by `I_0,C` here (max intensity in this case will be 1, but I'll get there). The log part here, `-log10(I_C/I_0,C)`, is known by them as _Beer-Lambert law_.

On [another SO question](https://stackoverflow.com/questions/61297083/separate-stains-output-of-one-channel-and-conversion-to-grayscale/),

>  The linear algebra involved in stain unmixing can produce negative values. Obviously negative stain densities do not exist. These negative values are an artefact of noise, imprecise color definitions, and the unmixing model ignoring light absorption by things other than the two stains you defined.
> It is safe to set negative pixels to zero.

This answer — and a couple of other ones — took me to [DIPlib](https://github.com/DIPlib/diplib). On `diplib/include/diplib/microscopy.h', we have:

```c++
/// The Beer-Lambert law describes how light is attenuated as it travels through an
absorbing medium. (...)
/// This function applies the mapping \f$A\f$ to the intensities \f$I\f$ in image `in`.
///
/// \f[ A = -log_{10}(I/I_0) \;, \f]
///
/// with \f$I_0\f$ the intensity of the illumination (`background`), and \f$A\f$ the
absorbance written to `out`).
///
/// `background` can be a single value or one value per tensor element (channel)
in `in`. `out` will have the same
/// number of tensor elements. (...)
///
/// `in` must be real-valued. Values outside of the range [0,`background`] will be
clipped.
/// `out` will be a floating-point type (do not force it to be an integer type, as the
rounding will destroy
/// all data).
```

I did not check the code — was not competent enough to find it —, but I got two conclusions from this documentation: `background` will be the maximum value on the range — 1 in our case, as our input is `float`; they clip the output data between `[0, 'background']`.

The ["Colour Deconvolution" Fiji plugin](https://github.com/fiji/Colour_Deconvolution/blob/master/src/main/java/sc/fiji/colourDeconvolution/StainMatrix.java) behaves more or less the same way, as seen here. Check that they add 1 to the input image, but it seems they are considering it between `(0, 255)`:

```java
double leng, A, V, C, log255=Math.log(255.0);
(...)
rgbOD[0] = rgbOD[0]+ (-((255.0*Math.log(((double)((p & 0xff0000)>>16) +1)/255.0))/log255));
rgbOD[1] = rgbOD[1]+ (-((255.0*Math.log(((double)((p & 0x00ff00)>> 8) +1)/255.0))/log255));
rgbOD[2] = rgbOD[2]+ (-((255.0*Math.log(((double)((p & 0x0000ff))     +1)/255.0))/log255));
```

I found a solution close to ours — actually based on ours, but different, from @jnkather's [Matlab code](https://github.com/jnkather/ColorDeconvolutionMatlab/blob/master/SeparateStains.m). This is something close to what I am implementing now. Instead of clipping values, it involves a solution @rfezzani tried to give before — rescaling (sorry Riadh, I didn't remember the actual PR you proposed that).

Okay, so these are the ideas I had. Using `log()` instead of `log10()` — used in Fiji's plugin and @jnkather's code — and  adding `exp(1)` instead of `2` to be "more neutral on the log", I guess.

Maybe there is something to improve on our conversion matrices — although I didn't find anything _arguably wrong_ there. [Fiji's plugin ones](https://github.com/fiji/Colour_Deconvolution/blob/master/src/main/resources/sc/fiji/colourDeconvolution/colourdeconvolution.txt) could be used as a parameter.

Lots of stuff going on here! I had _way_ more detail, but let me spare you from the suffering :smile:
Everything here is open for discussion, so please help me on making this code as great and reliable as possible. Thank you!

skimage/color/colorconv.py
```diff
@@ -1457,9 +1459,13 @@ def separate_stains(rgb, conv_matrix):
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     """
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb += 2
-    stains = np.reshape(-np.log10(rgb), (-1, 3)) @ conv_matrix
-    return np.reshape(stains, rgb.shape)
+    rgb += np.exp(1)  # avoiding log artifacts
```
hmaarrfk, line 1462:
```suggestion
    rgb += np.e  # avoiding log artifacts
```

hmaarrfk:
```
In [1]: np.e                                                                    
Out[1]: 2.718281828459045

In [2]: np.exp(1)                                                               
Out[2]: 2.718281828459045
```

hmaarrfk:
is there some kind of test you could add to solidify this change?

skimage/color/colorconv.py
```diff
@@ -1513,8 +1519,6 @@ def combine_stains(stains, conv_matrix):
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     >>> ihc_rgb = combine_stains(ihc_hdx, rgb_from_hdx)
     """
-    from ..exposure import rescale_intensity
```
hmaarrfk, line 1516: I think this was avoiding a circular import.

Maybe best to leave this to a separate PR.

alexdesiqueira: You got it.

alexdesiqueira:
> is there some kind of test you could add to solidify this change?

Not sure yet, Mark. I want something reliable first, to think on that later. I ping'ed Cris Luengo on StackOverflow, let's see what we got.

crisluengo:
The Beer-Lambert mapping is not just any logarithmic mapping, it is a specific mapping that models the physical process of light transmitted through an absorbing medium (the dyes in the tissue on the slide). These images are formed by shining light through the stained tissue and recording the amount of light that makes it through. The goal of the mapping we're implementing here is to estimate the density of the dye on each location on that tissue. This is a quantitative operation, even if the output units are arbitrary (i.e. we cannot know the absolute values, we can only compare densities across the image). Most importantly, pixels with no dye must have a value of 0 in the output.

Let's assume that the input image is in the range of 0 (black, no light) to 1 (white, brightness of the light source). This would correspond to 0 = all light absorbed, the dye is very dense, and 1 = no light is absorbed, there is no dye. So, the output should be a mapping where 0 maps to some large value and 1 maps to 0. Furthermore, due to Beer-Lambert, we know that the amount of light absorbed relates to the dye density by a power of 10 function. The proper mapping we're looking for is `output = -log10(input)`.

To avoid `log(0)` where the input is 0, one can do two things:

1. Add a small value to the input. For example `-log10(input + 1e-6)`. However, now where the input is 1, the output will not be 0 but some small negative value. The Fiji plugin uses `1/255` for this constant.

2. Set all values below some small value to that small value: `-log10(max(input, 1e-6))`. This is, IMO, the better approach. It is what I used [in DIPlib](https://github.com/DIPlib/diplib/blob/28cdcc53f0735ab474d3e6e8862f83ea1f54f8bb/src/microscopy/unmix_stains.cpp#L37) too.

It is perfectly fine to scale the output image so that the maximum possible range is [0,1], since the units are arbitrary anyway. But this scaling has to be consistent across images, so that different images can be compared: you don't want a lightly stained slide to yield the same stain densities as a darkly stained slide. So, this scaling has to use a constant, not the min and max in the image as this PR proposes (with `rescale_intensity`).

For example, we could follow the Fiji plugin idea and use 1/255 as the smallest allowed input value. This is mapped by the negative logarithm to 2.4065. Thus, we could divide the output by this value to limit it to the [0,1] range. (Note that 1 is mapped to 0, which is correct and that point should not be modified by any scaling.)

Code could look like this:

```python
rgb = np.max(rgb, 1/255)  # avoiding log artifacts
stains = np.reshape(-np.log10(rgb) / 2.4065, (-1, 3)) @ conv_matrix
return np.reshape(stains, rgb.shape)
```

---

**Note** in the proposed code:
```python
rgb += np.e                  # Causes mapping to not match physical process
stains = np.reshape(-np.log(rgb), (-1, 3)) @ conv_matrix
                             # OK to use log instead of log10, but scaling must be adjusted
stains = np.reshape(stains, rgb.shape)
stains = rescale_intensity(stains, out_range=(0, 1))
                             # Causes output values to not be comparable across images
return invert(stains)        # Not sure where this inverse comes from, should not be here
```

---

The inverse transform (`combine_stains`) must be an exact inverse of this operation, such that one can call the two in sequence and end up with the same image values (except for the 0 values being lost). One test to add here would be exactly that: two random gray-scale images, put through `combine_stains` to generate an RGB image, then put through `separate_stains` to get the two gray-scale images back. Some rounding errors are to be expected, but the two channels should be close to the original ones. You can do this test mixing three gray-scale images as well, of course.

alexdesiqueira:
@crisluengo thank you very much for your help in this issue, I really appreciate it! I've read the paper but was confused with different implementations; your comment answered all of my doubts — and you even gave the implementation as a bonus.

> For example, we could follow the Fiji plugin idea and use 1/255 as the smallest allowed input value. This is mapped by the negative logarithm to 2.4065. Thus, we could divide the output by this value to limit it to the [0,1] range. (Note that 1 is mapped to 0, which is correct and that point should not be modified by any scaling.)

One of the main points you clarified for me was the necessity of compensate the sum inside the logarithm; I _thought_ that could be the case, but didn't see how to make it work. Some details you questioned — e.g. the `invert()` within the function —  were attempts to mimic different code.

alexdesiqueira:
It seems that this behavior is already covered by tests:
```
=========================== short test summary info ============================
FAILED scikit-image/skimage/color/tests/test_colorconv.py::TestColorconv::test_hdx_rgb_roundtrip
FAILED scikit-image/skimage/color/tests/test_colorconv.py::TestColorconv::test_hdx_rgb_roundtrip_float
FAILED scikit-image/skimage/color/tests/test_colorconv.py::TestColorconv::test_hed_rgb_float_roundtrip
FAILED scikit-image/skimage/color/tests/test_colorconv.py::TestColorconv::test_hed_rgb_roundtrip
====== 4 failed, 6360 passed, 5 skipped, 34 warnings in 180.88s (0:03:00) ======
```

alexdesiqueira:
I'm checking that the function is returning values outside `(0, 1)`. They come from the multiplication with `conv_matrix`. It seems safe to clip them out... this is the behavior on most code elsewhere.

skimage/color/colorconv.py
```diff
@@ -1460,13 +1463,12 @@ def separate_stains(rgb, conv_matrix):
...
 
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb += np.e  # avoiding log artifacts
+    rgb += np.maximum(rgb, 1E-6)  # avoiding log artifacts
+    log_adjust = -np.log10(1E-6)  # used to compensate the sum above
 
-    stains = np.reshape(-np.log(rgb), (-1, 3)) @ conv_matrix
-    stains = np.reshape(stains, rgb.shape)
-    stains = rescale_intensity(stains, out_range=(0, 1))
+    stains = np.reshape(-np.log10(rgb) / log_adjust, (-1, 3)) @ conv_matrix
```
hmaarrfk, line 1469: I don't think you need this reshaping business:
```python
In [1]: np.zeros((2, 2, 3)) @ np.ones((3, 3))                                   
Out[1]: 
array([[[0., 0., 0.],
        [0., 0., 0.]],

       [[0., 0., 0.],
        [0., 0., 0.]]])

In [2]: (np.zeros((2, 2, 3)) @ np.ones((3, 3))).shape                           
Out[2]: (2, 2, 3)
```

skimage/color/colorconv.py
```diff
@@ -1456,10 +1460,16 @@ def separate_stains(rgb, conv_matrix):
     >>> ihc = data.immunohistochemistry()
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     """
+    from ..exposure import rescale_intensity
+
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb += 2
-    stains = np.reshape(-np.log10(rgb), (-1, 3)) @ conv_matrix
-    return np.reshape(stains, rgb.shape)
+    rgb = np.maximum(rgb, 1E-6)  # avoiding log artifacts
```
hmaarrfk, line 1466:
```suggestion
    np.maximum(rgb, 1E-6, out=rgb)  # avoiding log artifacts
```

hmaarrfk:
> Furthermore, due to Beer-Lambert, we know that the amount of light absorbed relates to the dye density by a power of 10 function. 

It seems to me there are two used units: Transmittance (natural log), Absorbance (log10).

That said, dividing by 2.4... makes sure to ignore both those units:

```
In [1]: np.log(5) / np.log(1E-3)                                                                         
Out[1]: -0.2329900014453396

In [2]: np.log10(5) / np.log10(1E-3)                                                                     
Out[2]: -0.23299000144533963

```

skimage/color/colorconv.py
alexdesiqueira: Isn't the explicit version better understandable?

alexdesiqueira:
@hmaarrfk I'd go for `log10`, since it's immediately relatable to the original paper.

skimage/color/colorconv.py
hmaarrfk: giving the out parameter makes this in place. I'm a big fan of doing little things like that for memory management ;)

hmaarrfk:
I'm with you on the relatability to the paper. But if you are expecting certain units to come out, the units of the current output are not what @crisluengo seemed to imply.

hmaarrfk:
The other thing is that all things being equal, I would rather have a 20% boost in performance using natural log:

These are some benchmark results on a 1MP image (1024 * 1024)

```
In [12]: a = np.full((1024 * 1024), fill_value=2.0)                                                                            

In [13]: %timeit np.log10(a)                                                                                                   
7.37 ms ± 99.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [14]: %timeit np.log(a)                                                                                                     
6.04 ms ± 74.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [15]: %timeit np.log2(a)                                                                                                    
9.29 ms ± 83 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

skimage/color/colorconv.py
alexdesiqueira: You're right, results are the same. Maybe some heritage from old times? :sweat_smile:

skimage/color/colorconv.py
hmaarrfk: Might not be so significant for 1MP arrays
```
In [16]: %timeit np.maximum(a, 0.01)                                                                                           
1.1 ms ± 8.98 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [17]: %timeit np.maximum(a, 0.01, out=a)                                                                                    
1.08 ms ± 46.6 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

```

alexdesiqueira: While I understand the appeal, I repeat the question :smile: Isn't it more readable?

hmaarrfk: I think you would really want to reconsider when your data becomes 1GB large, but i'll leave that as a *me* problem, instead of a library problem:
```
In [5]: a = np.full((1, 1024**3//8), fill_value=2, dtype='float')                                        

In [6]: %timeit np.maximum(a, 0.01, out=a)                                                               
138 ms ± 1.75 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [7]: %timeit np.maximum(a, 0.01)                                                                      
567 ms ± 198 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

alexdesiqueira: :joy: You convinced me, Mark.

hmaarrfk: yeah, the only images i care about are 1GB ;)

hmaarrfk: thx

crisluengo:
> I'm checking that the function is returning values outside (0, 1). They come from the multiplication with conv_matrix. It seems safe to clip them out... this is the behavior on most code elsewhere.

It is only the very dark pixels where this happens, and at those pixels the results are dominated by noise anyway, so I wouldn’t worry about this clipping.

crisluengo:
Regarding `log` vs `log10`: it doesn’t matter at all. It is about converting the input values to a linear space where the matrix multiplication makes sense. The scaling is irrelevant, since we can not determine absolute densities anyway. These values are all relative. We must be able to compare values across images, but we don’t get actual physical units out.

I was not aware of the  timing difference between the two. Interesting. I’d go with `log` if it’s faster.

alexdesiqueira:
> It is only the very dark pixels where this happens, and at those pixels the results are dominated by noise anyway, so I wouldn’t worry about this clipping.

```python
>>> from skimage import data 
>>> from skimage.color import separate_stains, hdx_from_rgb 
>>> ihc = data.immunohistochemistry() 
>>> ihc_hdx = separate_stains(ihc, hdx_from_rgb) 
>>> len(np.squeeze(np.where((ihc_hdx < 0) | (ihc_hdx > 1)))[0])                  
30836
```

@crisluengo yep, around 4% on our example (~31k px out of 785432. Although if we clip, it'd return a nice float in `(0, 1)` back. I wonder how to recover the original image on `combine_stains()` using the clip, though...

crisluengo:
>  I wonder how to recover the original image on combine_stains() using the clip, though...

I would do it the other way around: first use `combine_stains`, then separate them again with `separate_stains`. `combine_stains` will not need to clip, since the mapping there involves a negative exponential that maps values in the range [0,∞] to values in the range [0,1]. `combine_stains` also doesn't have the `np.maximum` call in it. Valid stain values should be able to make the round-trip unaffected. Just try to avoid values very near 1.0 in the input test images.

alexdesiqueira:
@crisluengo sorry I didn't see your last before. I was making some tests, and it seems that, without the clipping, we can still recover the original image ― close to the original up to `1E-5`, which it's pretty okay. The image is still there with the clipping, but not sufficiently close to the original anymore.

alexdesiqueira:
@hmaarrfk @crisluengo I think that's it. Please let me know if you have any considerations. Thank you for your help revising this Mark, and I truly appreciate your help Cris! Owe ya both a beer :)

skimage/color/colorconv.py
```diff
@@ -1456,10 +1460,16 @@ def separate_stains(rgb, conv_matrix):
     >>> ihc = data.immunohistochemistry()
     >>> ihc_hdx = separate_stains(ihc, hdx_from_rgb)
     """
+    from ..exposure import rescale_intensity
```
rfezzani, line 1463:
```suggestion
```
This is no more used in this function ;)

skimage/color/colorconv.py
```diff
@@ -1456,10 +1460,16 @@ def separate_stains(rgb, conv_matrix):
...
+
     rgb = _prepare_colorarray(rgb, force_copy=True)
-    rgb += 2
-    stains = np.reshape(-np.log10(rgb), (-1, 3)) @ conv_matrix
-    return np.reshape(stains, rgb.shape)
+    np.maximum(rgb, 1E-6, out=rgb)  # avoiding log artifacts
+    log_adjust = -np.log(1E-6)  # used to compensate the sum above
+
+    stains = (-np.log(rgb) / log_adjust) @ conv_matrix
+    stains = np.reshape(stains, rgb.shape)
```
rfezzani, line 1470:
```suggestion
```
This reshape is not needed I think.

alexdesiqueira:
Thank you @rfezzani!

skimage/color/colorconv.py
```diff
@@ -1516,10 +1526,12 @@ def combine_stains(stains, conv_matrix):
...
 
     stains = _prepare_colorarray(stains)
-    logrgb2 = -np.reshape(stains, (-1, 3)) @ conv_matrix
-    rgb2 = np.power(10, logrgb2)
-    return rescale_intensity(np.reshape(rgb2 - 2, stains.shape),
-                             in_range=(-1, 1))
+    # log_adjust here is used to compensate the sum within separate_stains().
+    log_adjust = -np.log(1E-6)
+    log_rgb = -(stains * log_adjust) @ conv_matrix
+    rgb = np.power(np.e, log_rgb)
```
hmaarrfk, line 1532:
```suggestion
    rgb = np.exp(log_rgb)
```


```
In [6]: %timeit np.power(np.e, a)                                               
21.3 µs ± 304 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [7]: %timeit np.exp(a)                                                       
8.12 µs ± 299 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)


```

skimage/color/colorconv.py
```diff
@@ -1516,10 +1526,12 @@ def combine_stains(stains, conv_matrix):
     from ..exposure import rescale_intensity
 
     stains = _prepare_colorarray(stains)
-    logrgb2 = -np.reshape(stains, (-1, 3)) @ conv_matrix
-    rgb2 = np.power(10, logrgb2)
-    return rescale_intensity(np.reshape(rgb2 - 2, stains.shape),
-                             in_range=(-1, 1))
```
hmaarrfk, line 1522: Is this change of behavior going to cause problems.

skimage/color/colorconv.py
```diff
@@ -1516,10 +1526,12 @@ def combine_stains(stains, conv_matrix):
...
+    # log_adjust here is used to compensate the sum within separate_stains().
+    log_adjust = -np.log(1E-6)
+    log_rgb = -(stains * log_adjust) @ conv_matrix
+    rgb = np.power(np.e, log_rgb)
+
+    return rescale_intensity(rgb, in_range=(0, 1))
```
crisluengo, line 1534: Just like in the conversion in the other direction, I would get rid of the `rescale_intensity`. We do not want scaling that depends on image content, because it makes it impossible to compare results across images.

This min-max rescaling will also make the round-trip test fail unless the input happens to span the full [0,1] range. I suggest you add a test with data that does not span the full range to verify that there is no image-content-dependent stuff going on.

skimage/color/colorconv.py
alexdesiqueira: It's not supposed to. I made some tests, and it works. Let's hope for the best :slightly_smiling_face:

skimage/color/colorconv.py
alexdesiqueira: You got it. Made some tests, it seems OK to leave it out.

alexdesiqueira: Nope... actually it will break our tests :slightly_smiling_face: it seems to a float between `(0, 1)` coming back from `combine_stains()`. I am re-running tests locally to check if this is the case.

skimage/color/colorconv.py
```diff
@@ -1448,6 +1448,10 @@ def separate_stains(rgb, conv_matrix):
     References
     ----------
     .. [1] https://web.archive.org/web/20160624145052/http://www.mecourse.com/landinig/software/cdeconv/cdeconv.html
+    .. [2] A. C. Ruifrok and D. A. Johnston, “Quantification of histochemical
+           staining by color deconvolution,” Anal. Quant. Cytol. Histol., vol.
+           23, no. 4, pp. 291–299, Aug. 2001.
+    .. [3] https://github.com/DIPlib/diplib/
```
sciunto, line 1454: Could write a word to say what's specifically useful in this project?

skimage/color/colorconv.py
```diff
@@ -1516,10 +1526,12 @@ def combine_stains(stains, conv_matrix):
     from ..exposure import rescale_intensity
 
     stains = _prepare_colorarray(stains)
-    logrgb2 = -np.reshape(stains, (-1, 3)) @ conv_matrix
-    rgb2 = np.power(10, logrgb2)
-    return rescale_intensity(np.reshape(rgb2 - 2, stains.shape),
-                             in_range=(-1, 1))
+    # log_adjust here is used to compensate the sum within separate_stains().
+    log_adjust = -np.log(1E-6)
+    log_rgb = -(stains * log_adjust) @ conv_matrix
```
rfezzani, line 1531:
```suggestion
    log_adjust = np.log(1E-6)
    log_rgb = (stains * log_adjust) @ conv_matrix
```
Can we remove this double negation?

alexdesiqueira:
```
6055 passed, 31 skipped, 12 warnings in 156.68s (0:02:36) 
```
Tests are OK here. Any hints? cc @hmaarrfk @sciunto @rfezzani

rfezzani:
@alexdesiqueira, I wonder if we shouldn't have a closer look to @crisluengo suggestion to remove the `rescale_intensity` from `combine_stain`...

rfezzani:
@alexdesiqueira, you also need to rebase if you want fix CI failures :/

crisluengo:
> @alexdesiqueira, I wonder if we shouldn't have a closer look to @crisluengo suggestion to remove the `rescale_intensity` from `combine_stain`...

As I understand it, `rescale_intensity` with `in_range=(0, 1)` doesn't actually rescale, it only clips values outside of the (0,1) range. This is OK. If my understanding is wrong, then yes, the rescaling should be removed.

alexdesiqueira:
> @alexdesiqueira, I wonder if we shouldn't have a closer look to @crisluengo suggestion to remove the rescale_intensity from combine_stain...

> As I understand it, rescale_intensity with in_range=(0, 1) doesn't actually rescale, it only clips values outside of the (0,1) range. This is OK. If my understanding is wrong, then yes, the rescaling should be removed.

@rfezzani I tested it before. Our tests will fail if we don't rescale the image, unfortunately. It expects a float going out of `combine_stains` :disappointed:
@crisluengo this one _actually_ rescales; let me test a clip function instead.

alexdesiqueira:
It seems to work properly. Thanks @rfezzani @crisluengo!

rfezzani:
Excellent @alexdesiqueira , we can remove the import of `rescale_intensity` now :tada:

rfezzani:
Thank you @alexdesiqueira ;)

rfezzani:
And of course thank you @crisluengo for your precious advises ;)

alexdesiqueira:
If it wasn't for @crisluengo we wouldn't solve this that quickly. Thank you very much!

asmagen:
@rfezzani @alexdesiqueira I'm still getting all negative values. Can you clarify what could be the issue? Did it produce mostly positive values for the image I send you earlier here?

alexdesiqueira:
@asmagen this fix will be available on skimage 0.18; to be able to use it right now, you will need to install the development version of skimage. Please refer to the [dev installation instructions](https://scikit-image.org/docs/dev/install.html#development-installation).
Regarding the test you presented on #4629, these are my outputs:
```
>>> ihc_hax = separate_stains(image, hax_from_rgb)                                                                                                                       
>>> aec = ihc_hax[:, :, 1]
>>> aec.min()                                                                                                                                                           
-0.006042518477085387
>>> aec.max()                                                                                                                                                           
0.15660479943381303
```

asmagen:
Thanks @alexdesiqueira. I did follow the installation steps and I believe the right version is installed but I still get all negative values and I'm really not sure what's wrong with the function calls or installation:

```
(saturn) jovyan@jupyter-assafmagen-2dpathflowai:~/project$ git clone https://github.com/scikit-image/scikit-image.git
Cloning into 'scikit-image'...
remote: Enumerating objects: 7, done.
remote: Counting objects: 100% (7/7), done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 72960 (delta 0), reused 1 (delta 0), pack-reused 72953
Receiving objects: 100% (72960/72960), 60.91 MiB | 45.59 MiB/s, done.
Resolving deltas: 100% (53796/53796), done.
(saturn) jovyan@jupyter-assafmagen-2dpathflowai:~/project$ cd scikit-image
(saturn) jovyan@jupyter-assafmagen-2dpathflowai:~/project/scikit-image$ pip install -e .
Obtaining file:///home/jovyan/project/scikit-image
Requirement already satisfied: numpy>=1.15.1 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from scikit-image==0.18.dev0) (1.18.1)
Requirement already satisfied: scipy>=1.0.1 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from scikit-image==0.18.dev0) (1.4.1)
Requirement already satisfied: matplotlib!=3.0.0,>=2.0.0 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from scikit-image==0.18.dev0) (3.0.3)
Requirement already satisfied: networkx>=2.0 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from scikit-image==0.18.dev0) (2.4)
Requirement already satisfied: pillow!=7.1.0,!=7.1.1,>=4.3.0 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from scikit-image==0.18.dev0) (7.1.2)
Requirement already satisfied: imageio>=2.3.0 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from scikit-image==0.18.dev0) (2.8.0)
Collecting tifffile>=2019.7.26
  Using cached tifffile-2020.5.25-py3-none-any.whl (132 kB)
Requirement already satisfied: PyWavelets>=1.1.1 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from scikit-image==0.18.dev0) (1.1.1)
Requirement already satisfied: pyparsing!=2.0.4,!=2.1.2,!=2.1.6,>=2.0.1 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from matplotlib!=3.0.0,>=2.0.0->scikit-image==0.18.dev0) (2.4.7)
Requirement already satisfied: python-dateutil>=2.1 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from matplotlib!=3.0.0,>=2.0.0->scikit-image==0.18.dev0) (2.8.1)
Requirement already satisfied: cycler>=0.10 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from matplotlib!=3.0.0,>=2.0.0->scikit-image==0.18.dev0) (0.10.0)
Requirement already satisfied: kiwisolver>=1.0.1 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from matplotlib!=3.0.0,>=2.0.0->scikit-image==0.18.dev0) (1.2.0)
Requirement already satisfied: decorator>=4.3.0 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from networkx>=2.0->scikit-image==0.18.dev0) (4.4.2)
Requirement already satisfied: six>=1.5 in /srv/conda/envs/saturn/lib/python3.6/site-packages (from python-dateutil>=2.1->matplotlib!=3.0.0,>=2.0.0->scikit-image==0.18.dev0) (1.14.0)
ERROR: pathflowai 0.1.1 has requirement click==6.7, but you'll have click 7.1.2 which is incompatible.
Installing collected packages: tifffile, scikit-image
  Attempting uninstall: tifffile
    Found existing installation: tifffile 0.15.1
    Uninstalling tifffile-0.15.1:
      Successfully uninstalled tifffile-0.15.1
  Running setup.py develop for scikit-image



Successfully installed scikit-image tifffile-2020.5.25
```

```
import skimage
from skimage.color import (separate_stains, hax_from_rgb)
​
ihc_hax = separate_stains(cropped_rgb_original, hax_from_rgb)
hema = ihc_hax[:, :, 0]
aec = ihc_hax[:, :, 1]
residual = ihc_hax[:, :, 2]
print(np.max(hema))
print(np.max(aec))
print(np.max(residual))
-0.24114280556296305
-0.2202285563618642
-0.18358928327693116
```

What else can I check or do?

hmaarrfk:
can you provide the output of 

```
import skimage
print(skimage.__version__)
```

of the system's youve installed

it should be `0.18.0.dev0`

asmagen:
It's not the right one then, it's 0.16.2
But I just clones the repo, how's that possible?

hmaarrfk:
not too sure, can you go through the common gotchas i outlined:
https://github.com/scikit-image/scikit-image/pull/4750/files#diff-26dee18463b5da190f343abcc74cf708R131

hmaarrfk:
my guess, is that you installed jupyter in the base environment, and not the `saturn` environment.

hmaarrfk:
and that you are using jupyter to test things out.

asmagen:
I see. I'm using Saturn cloud so I don't know that installation structure. Do you know how do I instruct the installation to go on jupyter?

hmaarrfk:
type
```
conda list
```

and make sure the packages you use 

But it is seems like it is a paid service, so they can probably help you  with specifics ;)

I wouldn't try to do the development build you are trying.

Rather, I outlined an other strategy where you just copy over the py file you need in the issue you raised.

asmagen:
Yeah, I'm asking them about the installation environment
But for the use of that single file kind of manually, I'm trying to do the following after downloading the file to `/home/jovyan/project/computational_imaging/py/skimage/colorconv.py`, but not sure what's the right way to import it:
```
os.chdir('/home/jovyan/project/computational_imaging/py/skimage/')
from colorconv import (separate_stains, hax_from_rgb)
​
ihc_hax = separate_stains(cropped_rgb_original, hax_from_rgb)
hema = ihc_hax[:, :, 0]
aec = ihc_hax[:, :, 1]
residual = ihc_hax[:, :, 2]
ImportError                               Traceback (most recent call last)
<ipython-input-39-7a9651ed323f> in <module>
      3 # from skimage.color import (separate_stains, hax_from_rgb)
      4 os.chdir('/home/jovyan/project/computational_imaging/py/skimage/')
----> 5 from colorconv import (separate_stains, hax_from_rgb)
      6 
      7 ihc_hax = separate_stains(cropped_rgb_original, hax_from_rgb)

~/project/computational_imaging/py/skimage/colorconv.py in <module>
     55 from warnings import warn
     56 from scipy import linalg
---> 57 from ..util import dtype, dtype_limits
     58 
     59 

ImportError: attempted relative import with no known parent package
```

hmaarrfk:
Can you try applying the change I detailed in
https://github.com/scikit-image/scikit-image/issues/4754#issuecomment-634398189

Let's continue discussing there.

I really wouldn't try to install development versions of software

hmaarrfk:
At least not ones where the python file is as self contained as the one you are trying.

asmagen:
Thanks @hmaarrfk, the import works. However, see the somewhat alarming result showing that although the values are positive, there is a huge gap between the highest value to the intermediate values, which basically renders the whole. image black even after scaling. I think the transformation may have shifted the values in an undesirable way. @alexdesiqueira might want to look too.

Thanks

Screen Shot 2020-05-28 at 8 30 47 AM

hmaarrfk:
@asmagen glad you were able to get going! 

Sad to hear you are still hitting roadblocks.

I think you would be really well served either in a separate issue or on the forum https://forum.image.sc/tags/scikit-image

asmagen:
Hi @hmaarrfk, how do I proceed with that forum? I thought it's a development issue that needs to be addressed here.
@crisluengo @alexdesiqueira do you see any issue that might cause that? Do you see the same with your tests?

jni:
@asmagen can you share the image from the notebook above? Using the png you provided in the original issue, I get totally different min/max values or histogram for hema.

hema

alexdesiqueira:
> Thanks @hmaarrfk, the import works. However, see the somewhat alarming result showing that although the values are positive, there is a huge gap between the highest value to the intermediate values, which basically renders the whole.

> Hi @hmaarrfk, how do I proceed with that forum? I thought it's a development issue that needs to be addressed here.

@asmagen I'm truly sorry you're not having the results you _wanted_, but this does not seem to be a development issue anymore. Assuming the figure you are using for test is the one in #4629, check my tests:

```
>>> from skimage import io
>>> from skimage.color import separate_stains, hax_from_rgb

>>> image = io.imread('80645415-134ffe80-8a39-11ea-91b8-256e2f32c17e.png')   
>>> ihc_hax = separate_stains(image, hax_from_rgb)                                
>>> hema = ihc_hax[:, :, 0]                                                       
>>> aec = ihc_hax[:, :, 1]                                                        
>>> residual = ihc_hax[:, :, 2]  
>>> hema.min()                                                                   
0.008154859419071136
>>> hema.min(), hema.max()                                                       
(0.008154859419071136, 0.20576427650665538)
>>> aec.min(), aec.max()                                                         
(-0.006042518477085387, 0.15660479943381303)
>>> residual.min(), residual.max()                                               
(-0.0031832712147587205, 0.10513316026296561)
```

Since this issue was resolved already, _please_ do as @hmaarrfk asked:
* Go to the forum;
* Open an account;
* Ask what is wrong with your installation there.
The forum is the correct place to give you support now.

crisluengo:
@asmagen You can just scale the image if it looks too dark: `hema *= 5` or whatever scale is suitable.

This is related to the choice of `1E-6` in the conversion. This is the minimum accepted input value, which after logarithmic mapping becomes the largest value in the image. The image is then scaled so that this largest value is 1. Taking a larger value, for example `1/256` as the ImageJ plugin uses, leads to a smaller maximum value in the mapped image, and consequently, when scaling the largest value to 1, there is less compression of gray-values. The scaling difference would be `log(1E-6) / log(1/256) = -13.8155 / -5.5452 = 2.4914`. So multiplying the output by 2.5 and clipping the result would be equivalent to choosing `1/256` instead of `1E-6` as the lower bound for the input values.

Note that all scaling is arbitrary (just as limiting floating-point images to the range [-1,1] is arbitrary). What matters is that the ratio between pixel values is equal to the ratio between amounts of stain on the slide. A pixel with a value of 0.2 represents double the amount of stain as a pixel with a value of 0.1.

asmagen:
Thanks @crisluengo. Does it make sense to use the following strategy then?

Screen Shot 2020-05-28 at 2 14 16 PM
Screen Shot 2020-05-28 at 2 02 04 PM
Screen Shot 2020-05-28 at 2 14 00 PM

asmagen:
@alexdesiqueira sure, I'll get to the forum, thanks.

PR #4730, opened 2020-05-20
Blacklist cython version 0.29.18

sciunto:
## Description

Closes #4727 

Seems that the cause comes from upstream. See https://github.com/cython/cython/issues/3622

PR #4731, opened 2020-05-20
Fix CI failures related to conversion of np.floating to dtype

rfezzani:
## Description

This PR prevent the conversion using `np.floating` as it is [deprecated](https://numpy.org/devdocs/release/1.19.0-notes.html#converting-certain-types-to-dtypes-is-deprecated) in numpy 1.19.

It fixes CI failures in #4730

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/util/dtype.py
+++ b/skimage/util/dtype.py
@@ -244,6 +244,8 @@ def _convert(image, dtype, force_copy=False, uniform=False):
         if force_copy:
             image = image.copy()
         return image
+    else:
+        dtype = np.float64
 
     if not (dtype_in in _supported_types and dtype_out in _supported_types):
         raise ValueError("Can not convert from {} to {}."
--- a/skimage/util/dtype.py
+++ b/skimage/util/dtype.py
@@ -223,7 +223,10 @@ def _convert(image, dtype, force_copy=False, uniform=False):
     """
     image = np.asarray(image)
     dtypeobj_in = image.dtype
-    dtypeobj_out = np.dtype(dtype)
+    if dtype is np.floating:
+        dtypeobj_out = np.dtype('float64')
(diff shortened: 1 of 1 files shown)


sciunto:
looks like it does not fully fixes the CI :)

skimage/util/dtype.py
```diff
@@ -244,6 +244,8 @@ def _convert(image, dtype, force_copy=False, uniform=False):
         if force_copy:
             image = image.copy()
         return image
+    else:
+        dtype = np.float64
```
lagru, line 248: Did you mean `dtypeobj_out`? I think `dtype` is never used from this point onwards...

rfezzani: Good catch, it should be `dtype_out` ;)

rfezzani:
Ci failures are fixed by #4735

jni:
To clarify my position: this PR currently takes NumPy's deprecated behaviour and hardcodes it into skimage. Instead, we should follow NumPy's guidance here. Users doing `_convert(arr, np.floating)` should get the warning (although for some reason I couldn't trigger it locally), users using `img_as_float` should get the fixed (future) behaviour.

jni:
fwiw, I *do* think we should consider making `img_as_float` default to `np.float` rather than `np.float64`. But that can be resolved another day, this matches current behaviour.

sciunto:
@scikit-image/core Do we plan to release a minor version to include this fix?

jni:
Looks like it was backported to the 0.17.x branch, so I think we could release 0.17.3 with this. @emmanuelle could you handle that? I'd be available today and/or Thursday to help if needed.

PR #2443, opened 2017-01-10
add threshold Wolf Phansalskar

sciunto:
## Description
To follow-up on #2266. This PR took time to merge and we decided to do not merge 2 of the four algorithms originally implemented. Here is the code that has not been merged, at least for memory and to decide what to do...

PR #4734, opened 2020-05-20
Add Phansalkar local threshold algorithm

kleinerELM:
Applies Phansalkar local threshold to an array. Phansalkar is a modification of Sauvola technique to deal with low contrast images.

## Description

I need the Phansalkar thresholding filter for low contrast SEM images. I found an implementation in ImageJ, which worked great. But I did not find any mentioning of that filter for scikit. I used threshold_sauvola() as a template.

Source paper: 
Phansalskar N. et al. "Adaptive local thresholding for detection of nuclei in diversity stained cytology images.", International Conference on Communications and Signal Processing (ICCSP), pp. 218-220, 2011
DOI: 10.1109/ICCSP.2011.5739305

It should be noted, that the image has to be preprocessed using equalize_adapthist():

```
from skimage import data
from skimage.exposure import equalize_adapthist
image = data.moon()
image_eq = equalize_adapthist(image)
t_phansalkar = threshold_phansalkar(image_eq, window_size=15, k=0.25, p=2.0, q=10.0)
binary_image = image_eq > t_phansalkar
```

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1219,7 +1219,7 @@ def threshold_phansalkar(image, window_size=15, k=0.25, r=None, p=2.0, q=10.0):
     References
     ----------
     .. [1] Phansalskar N. et al. "Adaptive local thresholding for detection of
-           nuclei in diversity stained cytology images.", International 
+           nuclei in diversity stained cytology images.", International
            Conference on Communications and Signal Processing (ICCSP),
            pp. 218-220, 2011
            :DOI:`10.1109/ICCSP.2011.5739305`
@@ -1229,12 +1229,13 @@ def threshold_phansalkar(image, window_size=15, k=0.25, r=None, p=2.0, q=10.0):
     >>> from skimage.exposure import equalize_adapthist
     >>> image = data.moon()
     >>> image_eq = equalize_adapthist(image)
-    >>> t_phansalkar = threshold_phansalkar(image_eq, window_size=15, k=0.25, p=2.0, q=10.0)
+    >>> t_phansalkar = threshold_phansalkar(image_eq, window_size=15, k=0.25)
     >>> binary_image = image_eq > t_phansalkar
     """
-    
+
     if r is None:
         imin, imax = dtype_limits(image, clip_negative=False)
         r = 0.5 * (imax - imin)
     m, s = _mean_std(image, window_size)
-    return m * (1 + np.power(p, (-q * m) ) + k * ((s / r) - 1))
+    return m * (1 + np.power(p, (-q * m)) + k * ((s / r) - 1))
+  
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1184,7 +1184,7 @@ def threshold_multiotsu(image, classes=3, nbins=256):
 
 def threshold_phansalkar(image, window_size=15, k=0.25, r=None, p=2.0, q=10.0):
     """Find the Phansalkar local threshold for an array.
-    
+
     Phansalkar is a modification of Sauvola thresholding designed to deal with
     low contrast images.
 
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -24,7 +24,8 @@ __all__ = ['try_all_threshold',
            'threshold_sauvola',
            'threshold_triangle',
            'apply_hysteresis_threshold',
-           'threshold_multiotsu']
+           'threshold_multiotsu',
+           'threshold_phansalkar']
 
 
 def _try_all(image, methods=None, figsize=None, num_cols=2, verbose=True):
@@ -1177,3 +1178,63 @@ def threshold_multiotsu(image, classes=3, nbins=256):
     thresh = bin_centers[thresh_idx]
 
     return thresh
+
+
+def threshold_phansalkar(image, window_size=15, k=0.25, r=None, p=2.0, q=10.0):
+    """Applies Phansalkar local threshold to an array. Phansalkar is a
+    modification of Sauvola technique to deal with low contrast images.
+    This method is using the following formula::
+        T = m(x,y) * (1 + p * exp( -q * m(x,y) ) + k * ((s(x,y) / R) - 1))
+    where m(x,y) and s(x,y) are the mean and standard deviation of
+    pixel (x,y) neighborhood defined by a rectangular window with size w
+    times w centered around the pixel. k, p and q are configurable parameters.
+    R is the maximum standard deviation of a greyscale image.
+    Parameters
+    ----------
+    image : ndarray
+        Input image.
+    window_size : int, or iterable of int, optional
+        Window size specified as a single odd integer (3, 5, 7, …),
+        or an iterable of length ``image.ndim`` containing only odd
+        integers (e.g. ``(1, 5, 5)``).
+    k : float, optional
+        Value of the positive parameter k.
+    r : float, optional
+        Value of R, the dynamic range of standard deviation.
+        If None, set to the half of the image dtype range.
+    p : float, optional
+        Value of the parameter p.
+    q : float, optional
+        Value of the parameter q.
+    Returns
+    -------
+    threshold : (N, M) ndarray
+        Threshold mask. All pixels with an intensity higher than
+        this value are assumed to be foreground.
+    Notes
+    -----
+    This algorithm is originally designed for detection of cell nuclei in low
+    contrast images. Therefore the historgram has to be equalized beforehand
+    using skimage.exposure.equalize_adapthist().
+    References
+    ----------
+    .. [1] Phansalskar N. et al. "Adaptive local thresholding for detection of
+           nuclei in diversity stained cytology images.", International 
+           Conference on Communications and Signal Processing (ICCSP),
+           pp. 218-220, 2011
+           :DOI:`10.1109/ICCSP.2011.5739305`
+    Examples
+    --------
+    >>> from skimage import data
+    >>> from skimage.exposure import equalize_adapthist
+    >>> image = data.moon()
+    >>> image_eq = equalize_adapthist(image)
+    >>> t_phansalkar = threshold_phansalkar(image_eq, window_size=15, k=0.25, p=2.0, q=10.0)
+    >>> binary_image = image_eq > t_phansalkar
+    """
+    
+    if r is None:
+        imin, imax = dtype_limits(image, clip_negative=False)
+        r = 0.5 * (imax - imin)
+    m, s = _mean_std(image, window_size)
+    return m * (1 + np.power(p, (-q * m) ) + k * ((s / r) - 1))
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -1181,8 +1181,10 @@ def threshold_multiotsu(image, classes=3, nbins=256):
 
 
 def threshold_phansalkar(image, window_size=15, k=0.25, r=None, p=2.0, q=10.0):
-    """Applies Phansalkar local threshold to an array. Phansalkar is a
-    modification of Sauvola technique to deal with low contrast images.
+    """Find the Phansalkar local threshold for an array.
+    
+    Phansalkar is a modification of Sauvola thresholding designed to deal with
+    low contrast images.
 
     This method is using the following formula::
         T = m(x,y) * (1 + p * exp( -q * m(x,y) ) + k * ((s(x,y) / R) - 1))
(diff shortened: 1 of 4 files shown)


skimage/filters/thresholding.py
```diff
@@ -1177,3 +1178,63 @@ def threshold_multiotsu(image, classes=3, nbins=256):
     thresh = bin_centers[thresh_idx]
 
     return thresh
+
+
+def threshold_phansalkar(image, window_size=15, k=0.25, r=None, p=2.0, q=10.0):
+    """Applies Phansalkar local threshold to an array. Phansalkar is a
+    modification of Sauvola technique to deal with low contrast images.
```
lagru, line 1185: There is an empty line missing between the short and the long description here and between the section further below as well. You can find [a few examples here](https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html) and in the [official reference](https://numpydoc.readthedocs.io/en/latest/) for the formatting we use throughout the library.

skimage/filters/thresholding.py
```diff
@@ -1177,3 +1178,63 @@ def threshold_multiotsu(image, classes=3, nbins=256):
...
+    Returns
+    -------
+    threshold : (N, M) ndarray
+        Threshold mask. All pixels with an intensity higher than
+        this value are assumed to be foreground.
+    Notes
+    -----
+    This algorithm is originally designed for detection of cell nuclei in low
+    contrast images. Therefore the historgram has to be equalized beforehand
+    using skimage.exposure.equalize_adapthist().
```
lagru, line 1218:
```suggestion
    using :func:`skimage.exposure.equalize_adapthist`.
```

lagru (commented):
Thank you for the contribution @kleinerELM! This looks promising and from a first look at the paper and your description this might fit with the other thresholding functions. I'll give this a more thorough review later.

skimage/filters/thresholding.py
kleinerELM: I hope i was able to resolve that issue. I added some empty lines for readabilty as sown in the examples.

skimage/filters/thresholding.py
```diff
@@ -1177,3 +1178,69 @@ def threshold_multiotsu(image, classes=3, nbins=256):
...
+           Conference on Communications and Signal Processing (ICCSP),
+           pp. 218-220, 2011
+           :DOI:`10.1109/ICCSP.2011.5739305`
+
+    Examples
+    --------
+    >>> from skimage import data
+    >>> from skimage.exposure import equalize_adapthist
+    >>> image = data.moon()
+    >>> image_eq = equalize_adapthist(image)
```
rfezzani, line 1237: Why not include this histogram equalization in the function since it seems that it is required for the method to work?

skimage/filters/thresholding.py
```diff
@@ -1177,3 +1178,69 @@ def threshold_multiotsu(image, classes=3, nbins=256):
...
+    >>> t_phansalkar = threshold_phansalkar(image_eq, window_size=15, k=0.25)
+    >>> binary_image = image_eq > t_phansalkar
+    """
+
+    if r is None:
+        imin, imax = dtype_limits(image, clip_negative=False)
+        r = 0.5 * (imax - imin)
```
rfezzani, line 1244: If the histogram equalization is integrated to the function, `imax - imin` can be replaced by `image.ptp()`

rfezzani (commented):
Thank you @kleinerELM for your contribution, I just left some comments ;)

skimage/filters/thresholding.py
```diff
@@ -1177,3 +1178,69 @@ def threshold_multiotsu(image, classes=3, nbins=256):
     thresh = bin_centers[thresh_idx]
 
     return thresh
+
+
+def threshold_phansalkar(image, window_size=15, k=0.25, r=None, p=2.0, q=10.0):
+    """Applies Phansalkar local threshold to an array. Phansalkar is a
+    modification of Sauvola technique to deal with low contrast images.
```
jni, line 1185: @kleinerELM not yet fixed — the short description should fit in one line, followed by a blank line, followed by a longer explanation. In this case:

```suggestion
    """Find the Phansalkar local threshold for an array.
    
    Phansalkar is a modification of Sauvola thresholding designed to deal with
    low contrast images.
```

skimage/filters/thresholding.py
kleinerELM: I just followed the threshold_sauvola() fuction (and most of the other thresholding functions as well). They all return a threshold mask instead of a binary image. (Is there actually any reason for that?)
Since image_eq is necessary to create the correct binary image at the end, the function should best output a binary image directly, or what do you think?

I can see one advantage not to include the histogram equalization into the thresholding function:
If you want to find the best parameters window_size, k, r, p, q for a special image set, it is much faster if you have to calculate equalize_adapthist() only once, instead every time you change a parameter.
For example:

```
def threshold_phansalkar_parameter_variation( image ):
    #equalize only once
    image_eq = equalize_adapthist(image)
    
    # list of parameters
    k_list = [0.2,0.25,0.3,0.35]
    p_list = [1.5,2.0,2.5,3.0]

    fig, ax = plt.subplots(len(p_list),len(k_list), figsize=(15, 6) )
    fig.subplots_adjust(hspace = .5)
    ax = ax.ravel()

    for j in range(len(p_list)):
        for i in range(len(k_list)):
            # calculate binary mask of thresh_phansalkar() without doing equalize_adapthist() every time
            thresh_phansalkar = threshold_phansalkar(image_eq, window_size=15, k=k_list[i], p=p_list[j], q=10.0)
            binary_phansalkar = image_eq > thresh_phansalkar

            ax[i+j*len(k_list)].imshow(binary_phansalkar, cmap=plt.cm.gray)
            ax[i+j*len(k_list)].set_title( "k = " + str( k_list[i] )+ ", p = " + str( (p_list[j]) ) )
            ax[i+j*len(k_list)].set_axis_off()

    plt.show()
```

If you think that this is too rare a usecase, then your suggestion would definitely be more user-friendly.

skimage/filters/thresholding.py
kleinerELM: Thank you for that suggestion.
I tried it for some images. And actually r will allways be 1, if the imput image is a result of the histogram equalization.
Therefore, I would suggest skipping the calculation and set `r = 1.0` in that case.

What is the advantage of image.ptp()? I tried to benchmark the function and it seems image.ptp() was a little slower on my system than the current code. But that is negligible in comparison to the thresholding itself.

jni:
@kleinerELM btw this will need some tests, and I think it would be good to add to e.g. https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_niblack_sauvola.html#sphx-glr-auto-examples-segmentation-plot-niblack-sauvola-py

It should definitely return a threshold map, not a binarised image.

kleinerELM:
> 
> 
> @kleinerELM btw this will need some tests, and I think it would be good to add to e.g. https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_niblack_sauvola.html#sphx-glr-auto-examples-segmentation-plot-niblack-sauvola-py
> 
> It should definitely return a threshold map, not a binarised image.

Thank you for your patience. I added Phansalkar local threshold to the example and fixed the missing reference in skimage/filters/__init__.py .

While I was researching how to add the test functions I stumbled across the PR #2443 . It seems there was already a discussion including the Phansalkar threshold. Phansalkar was mistyped in that discussion, which is why I did not find that issue in the fist place.
Should I continue to work on the test suite or do we drop this PR? I personally would love to see this filter in Scikit.

kleinerELM:
@jni I added test functions according to similar filter functions and they seem to work properly.
Please let me know if there's anything more I can do.

skimage/filters/thresholding.py
kleinerELM: According to @jni the function should return a threshold map. Therefore, I thing it is best to leave the function. I documented the requirements of the function in the funtion itself and the example.

skimage/filters/thresholding.py
```diff
@@ -118,10 +122,22 @@ def try_all_threshold(image, figsize=(8, 5), verbose=True):
...
+        elif ( func.__name__ == 'threshold_phansalkar' ):
+            # threshold_phansalkar requires an equalized image
+            def wrapper(im):
+                im_eq = equalize_adapthist(im)
+                return im_eq > func(im_eq)
+        else:
+            # standard case
+            def wrapper(im):
+                return im > func(im)
+
```
kleinerELM, line 140: I don't know if you would accept this as good style, but it works for thresholding algorithms which require different input variables.

skimage/filters/thresholding.py
```diff
@@ -135,10 +151,14 @@ def wrapper(im):
                            'Minimum': thresh(threshold_minimum),
                            'Otsu': thresh(threshold_otsu),
                            'Triangle': thresh(threshold_triangle),
-                           'Yen': thresh(threshold_yen)})
+                           'Yen': thresh(threshold_yen),
+                           'Local': thresh(threshold_local),
+                           'Niblack': thresh(threshold_niblack),
+                           'Sauvola': thresh(threshold_sauvola),
+                           'Phansalkar': thresh(threshold_phansalkar)})
```
kleinerELM, line 158: Even if Phansalkar will not be accepted, I think it would be helpfull, if the most available filters are included in this function.

skimage/filters/thresholding.py
kleinerELM, line 140: I don't know if you would accept this as good style, but it works for thresholding algorithms which require different input variables.

lagru: Why not check for the actual function instead of a string comparison? E.g.
```python
if func is threshold_local:
    ...
```
Also the parenthesis and extra space around the conditional expressions are a bit unconventional. I'm surprised that @pep8speaks isn't bugging you about it.

kleinerELM: Thanks for your advice, I changed it.
Actually, the parenthesis and extra space are a quirk of mine.

skimage/filters/thresholding.py
```diff
@@ -117,10 +122,22 @@ def try_all_threshold(image, figsize=(8, 5), verbose=True):
...
+            def wrapper(im):
+                im_eq = equalize_adapthist(im)
+                return im_eq > func(im_eq)
+        else:
+            # standard case
+            def wrapper(im):
+                return im > func(im)
+
         try:
             wrapper.__orifunc__ = func.__orifunc__
```
lagru, line 142: Am I correct in assuming that this line will always raise an AttributeError? At least that's what local testing tells me. Assuming you are the author, @sciunto do you still remember the intention behind this line?

I may be a good idea to use [functools.wraps](https://docs.python.org/3/library/functools.html?highlight=wraps#functools.wraps) instead; `wraps` preserves the name of the function and makes the original one accessible through the `__wrapped__` attribute.

skimage/filters/thresholding.py
```diff
@@ -124,11 +124,11 @@ def thresh(func):
         """
         Select a wrapper function to return a thresholded image. 
         """
-        if ( func.__name__ == 'threshold_local' ):
+        if func == threshold_local:
```
lagru, line 127:
```suggestion
        if func is threshold_local:
```
We want to test for the identity of the function, not for its value. A semantic but in my opinion good distinction to make. :wink:

skimage/filters/thresholding.py
```diff
@@ -124,11 +124,11 @@ def thresh(func):
...
             # threshold_local requires the additional parameter 'block_size'
             def wrapper(im):
                 return im > func(im, 15)
-        elif ( func.__name__ == 'threshold_phansalkar' ):
+        elif func == threshold_phansalkar:
```
lagru, line 131:
```suggestion
        elif func is threshold_phansalkar:
```

skimage/filters/thresholding.py
kleinerELM: Oh, I didn't know the operator. I am learning a lot more than anticipated =)

lagru:
@kleinerELM I took the liberty to commit two small suggestions. You should be able to update your local branch with `git pull origin threshold_phansalkar` (in case you didn't know).

skimage/filters/thresholding.py
kleinerELM: Can you make a code suggestion? I just followed the existing pattern here. I wanted to avoid breaking anything.

matthew-brett:

<details>

## Rationale

</details>

PR #4738, opened 2020-05-21
Add pooch in requirements/test.txt

sciunto:
## Description
I propose that we add pooch to requirements/test.txt. Some of our tests are relying on external images. 

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -5,3 +5,4 @@ pytest-cov
 pytest-localserver
 flake8
 codecov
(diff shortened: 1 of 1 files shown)


hmaarrfk:
They are not dependent though. 

They silently get skipped from a pip install. End users probably don't care.

If you use a git install, they get found from the git repo. Developers probably do care, but don't want massive second downloads multiple times.

PR #4739, opened 2020-05-21
trainable segmentation (algorithm + gallery example)

emmanuelle:
This PR is not 100% ready but I'd like to start a discussion on several points.

- should we include a function like the `trainable_segmentation` of the example in `skimage.segmentation`? This would introduce a dependency to scikit-learn, which is not an easy decision to make especially for a single function. We could also include such a function in `scikit-image-contrib` and give it some love :-).
- on the other hand, a function like `compute_features` would not introduce new dependencies.
- software such as ilastik or the weka segmentation plugin from ImageJ propose a larger number of features (in particular ImageJ, see https://imagej.net/Trainable_Weka_Segmentation#Training_features_.282D.29) probably we would need to include more features. If anybody here has experience with this and knows which features are important to include I'm interested :-).

Part of my motivation for this example comes from the fact that we now have the possibility to [draw shapes in plotly figures](https://plotly.com/python/imshow/#annotating-image-traces-with-shapes), I could also add to the example a part at the end to be run in a notebook (or in binder) where the user could draw interactively the classes and see the result.


doc/examples/segmentation/plot_trainable_segmentation.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def trainable_segmentation(img, mask, multichannel=True,
+                           intensity=True, edges=True, texture=True,
+                           sigma_min=0.5, sigma_max=16, downsample=10):
+    """
+    Segmentation using labeled parts of the image and a random forest classifier.
+    """
+    print(img.shape)
+    t1 = time()
```
emmanuelle, line 85: probably we want to remove these times but this is to show it's the same order of magnitude for computing all features, fitting the classifier (when doing aggressive subsampling, which does not change the result significantly I found) and predicting the classes of unlabeled pixels.

jni:
@emmanuelle so cool!

btw we should start hacking together again. =)

I can't do a full review right now but you could make the function take in a classifier as input. That's the great thing about the sklearn unified interface, you can drop in whatever classifier you want! And, if you take a classifier as input, we (a) don't need to depend on sklearn at all, and (b) can take in other sklearn-compatible classifiers such as skorch. =) And (c) you can add a comparison of different classifiers — as part of my research once I found that I got a drop in accuracy from 82% to 79% when going from RFs to logistic regression — in exchange for a 100x speedup! I took the speedup. =P

In general, I think it would be a pretty good idea to provide this as part of the library. As a default maybe we can implement a simple logistic regression class with pure NumPy. =)

emmanuelle:
I love the idea of passing a classifier :-). I'll update the PR and also I'll take a look with logistic regression. Here though computing the feature takes a significant fraction (50%) of the total computing time so gaining a factor of x100 on the classification part does not result in a x100 total speed-up. Which means that we need to speed-up feature computation, and therefore that we indeed need to hack together ;-).

doc/examples/segmentation/plot_trainable_segmentation.py
```diff
@@ -0,0 +1,168 @@
...
+                sigma_min=sigma_min, sigma_max=sigma_max) for dim in range(img.shape[-1]))
+        features = list(itertools.chain.from_iterable(all_results))
+    else:
+        features = _compute_features_gray(
+            img[..., dim], intensity=intensity, edges=edges, texture=texture,
+            sigma_min=sigma_min, sigma_max=sigma_max)
+    return np.array(features)
+
+
+def trainable_segmentation(img, mask, clf, multichannel=True,
```
jni, line 78: Can we rename `mask` to `labels`? mask is always a boolean array in the library.

doc/examples/segmentation/plot_trainable_segmentation.py
```diff
@@ -0,0 +1,168 @@
...
+    t1 = time()
+    features = compute_features(im, multichannel=multichannel,
+                            intensity=intensity, edges=edges, texture=texture,
+                            sigma_min=sigma_min, sigma_max=sigma_max)
+    t2 = time()
+    training_data = features[:, mask > 0].T
+    training_labels = mask[mask>0].ravel()
+    data = features[:, mask == 0].T
+    t3 = time()
+    clf.fit(training_data[::downsample], training_labels[::downsample])
```
jni, line 94: This seems unnecessary and unintuitive? Or do you just have it here so you can iterate quickly? =)

doc/examples/segmentation/plot_trainable_segmentation.py
```diff
@@ -0,0 +1,168 @@
...
+    result = np.copy(mask)
+    result[mask == 0] = labels
+    print("compute features", t2 - t1)
+    print("fit", t4 - t3)
+    print("predict", t5 - t4)
+    return result
+
+
+from time import time
+# This image is in the public domain and could be added to the Gitlab data repo
```
jni, line 107:
```suggestion
# This image is in the public domain and ~~could~~ should be added to the Gitlab data repo
```

=P

doc/examples/segmentation/plot_trainable_segmentation.py
```diff
@@ -0,0 +1,168 @@
...
+# Build a mask for training the segmentation.
+# Here we use rectangles but visualization libraries such as plotly
+# (and napari?) can be used to draw a mask on the image.
+mask = np.zeros(im.shape[:2], dtype=np.uint8)
+mask[:100] = 1
+mask[:170, :400] = 1
+mask[600:900, 200:650] = 2
+mask[330:430, 210:320] = 3
+mask[260:340, 60:170] = 4
+mask[150:200, 720:860] = 4
```
jni, line 120: I see why now. =P Perhaps we can "simulate" sparse drawing by using e.g. `skimage.draw.line`?

doc/examples/segmentation/plot_trainable_segmentation.py
```diff
@@ -0,0 +1,168 @@
...
+
+sigma_min = 1
+sigma_max = 32
+clf = RandomForestClassifier(n_estimators=100, n_jobs=-1)
+result = trainable_segmentation(im, mask, clf, multichannel=True,
+                        intensity=True, edges=False, texture=True,
+                        sigma_min=sigma_min, sigma_max=sigma_max, downsample=15)
```
jni, line 127: This is just an example, but I'm thinking about adding this to the library. In this case, I would like to separate the training and the prediction steps, since typically one trains on a few images and then predicts on many more in batch.

So, something along the lines of:

```python
clf, predicted = train_segmenter(image, labels, clf, features=feature_func, ...)  # analogous to `fit_transform`
predictions = apply_segmenter(images, clf, features=feature_func, ...)  # analogous to `predict_proba`
```

There's an issue of keeping the feature transforms "in sync", but I don't have a good answer for that... =\

doc/examples/segmentation/plot_trainable_segmentation.py
emmanuelle: https://gitlab.com/scikit-image/data/-/merge_requests/3 :-)

doc/examples/segmentation/plot_trainable_segmentation.py
emmanuelle: Could we do `clf.features_func` inside `train_segmenter` so that we can just call `clf.features_func` inside `apply_segmenter`? I don't know if it's bad practice to do this... Adding an attribute to a class instance sounds pretty innocuous to me.

emmanuelle:
So I'm currently moving this functionality to `skimage.segmentation` and I'm wondering how to write tests. Is it ok to add `scikit-learn` to our `requirements/test.txt` ? Otherwise I will have to write a dummy classifier in the test module, which is fine but a bit more work.

sciunto:
> So I'm currently moving this functionality to `skimage.segmentation` and I'm wondering how to write tests. Is it ok to add `scikit-learn` to our `requirements/test.txt` ? 

This discussion can help: https://github.com/scikit-image/scikit-image/pull/4738

@hmaarrfk suggests that if tests are skipped (and you may skip if sklearn is not installed), it is not needed to add in test.txt. Just in optional.txt then.

I think I agree with this as long as we explain it in our dev doc.

emmanuelle:
Thanks @sciunto. This can indeed be an option. Since all our modules are doctested for all builds the question is also whether I add a docstring example creating a classifier or not. Anyway let me first update the code and push it :-).

doc/examples/segmentation/plot_trainable_segmentation.py
emmanuelle: Ha ha I spoke with Gael and this type of monkey-patching is definitely a bad practice :-), all the more than scikit-learn objects erase their attribute every time that the `clone` method is called. I'll think of something else, maybe something which is object-oriented.

emmanuelle:
I'll write a naive nearest neighbor classifier in the test module which will solve the dependency problem

emmanuelle:
So the PR is not ready yet (no docstrings no tests) but I pushed this to show what a purely functional API could be. We could also have a `TrainableSegmenter` class, which would call `fit_segmenter` within its `fit` method, and a `predict` method which could predict on any new image. So people could use the function for a single object and the object for workflows with several images. Thoughts on the API?

skimage/segmentation/trainable_segmentation.py
```diff
@@ -0,0 +1,151 @@
...
+        eigvals = feature.hessian_matrix_eigvals(H_elems)
+        for eigval_mat in eigvals:
+            features.append(eigval_mat)
+    return features
+
+
+def _mutiscale_basic_features_singlechannel(
+    img, intensity=True, edges=True, texture=True, sigma_min=0.5, sigma_max=16
+):
+    """Features for a single channel image. ``img`` can be 2d or 3d.
```
jni, line 32: I think all the features you've made are nD?

Also, I would prefer passing in a sequence of sigmas, which would allow anisotropic sigmas, for example.

emmanuelle: I also think it's nD which is cool, but for the texture features I adapted the code of `filters.hessian` (not calling the function because I did not want to recompute the gaussian-filtered image since this where most of the time is spent) and the docstring of `filters.hessian` says `Defined only for 2-D and 3-D images`

doc/examples/segmentation/plot_trainable_segmentation.py
jni: Perhaps he has encountered the problem of matching the feature computation to the classifier? Is there a proper way to use sklearn pipelines for this?

rfezzani:
Thank you @emmanuelle for starting all this, I really like the idea :)

>should we include a function like the trainable_segmentation of the example in skimage.segmentation? This would introduce a dependency to scikit-learn, which is not an easy decision to make especially for a single function. We could also include such a function in scikit-image-contrib and give it some love :-).

I am more on the side of not integrating `trainable_segmentation` in `skimage` since it requires manual annotations to work...
Giving some love to `scikit-image-contrib` is a good idea ;)

emmanuelle:
@rfezzani please note that in the gallery example no manual annotations are required and that you could imagine several methods to build the `labels` array required by the API. This is also the case of other functions in the segmentation module like `random_walker` or `watershed` which require to build an array of labels.

It's also possible to put this first in `scikit-image-contrib` and see if adoption builds up, no strong opinion here.

rfezzani:
You are right @emmanuelle, the annotation is not made "manually", but the areas are carefully selected by the user :). 
Since it is already the case for other functions, I have no objection for the integration of `trainable_segmentation` in `skimage`.

Rahul1984:
Hi,

I am writing codes on image processing which actually removes model head  & added with user's head to see ,how the user looks.

Can you please help me what are libraries ,functions are needed in python ?

If you have sample code ?

Please let me know ,

Thanks in advance
Rahul

emmanuelle:
Hi @Rahul1984 I don't see the link with this PR? This is the development repo of scikit-image for discussing bugs, feature requests etc. If you need support, if you're working on scientific images you can ask on https://forum.image.sc/, if it's general computer vision / machine learning you can ask for example on the forums pytorch (https://discuss.pytorch.org/) or tensorflow (https://www.tensorflow.org/community/forums).

emmanuelle:
@jni and all, I implemented a tentative `TrainableSegmenter` class, and I pushed a variant of the gallery example using the object-oriented approach. Please comment on the API :-). (I'll need to write a lot of tests but interested in stabilizing the API first).

skimage/segmentation/trainable_segmentation.py
emmanuelle: changed for the nD, I'll take a look at the anisotropic sigmas.

emmanuelle:
Note that if you have doubts about the API, it's possible to move part of it to `skimage.future` to flag it as possibly unstable.

jni:
I do have doubts, and, sorry I have been so slow to review this. I'll give it a good go tomorrow! I still have the nagging feeling that more can be done to use scikit-learn pipelines, although the final reshape is apparently not possible to do with a pipeline...

skimage/segmentation/trainable_segmentation.py
```diff
@@ -0,0 +1,302 @@
...
+            number of dimensions compatible with ``self.features_func``.
+
+        Raises
+        ------
+        NotFittedError if ``self.clf`` has not been fitted yet (use ``self.fit``).
+        """
+        return predict_segmenter(image, self.clf, self.features_func)
+
+
+def fit_segmenter(image, labels, clf, features_func=None, downsample=10):
```
emmanuelle, line 209: @jni (and all), regarding the API, I'm thinking of changing this function to take an array of features first (the first dimension corresponding to the number of features), so that the features could be computed beforehand (useful for caching results or a warm restart of the classifier). What do you think?

emmanuelle:
@jni and all: I updated a bit the API so that it was possible to compute features separately from the predict and fit operations. 

I also accelerated the computation of features, using joblib's Parallel. Are we ok with adding joblib with our list of dependencies? It's a small dep (no mandatory dependencies outside of Python, numpy as optional dependency). 

Finally, I optimized the parameters of the estimator so that the example runs quite fast (and thanks to the max_samples parameter of the RandomForestClassifier, I could remove the `downsample` parameter of the `fit_segmenter` function).

skimage/segmentation/trainable_segmentation.py
```diff
@@ -0,0 +1,291 @@
...
+    if intensity:
+        features += all_filtered
+    if edges:
+        all_edges = Parallel(n_jobs=-1, prefer='threads')(delayed(filters.sobel)(filtered_img)
+                                for filtered_img in all_filtered)
+        features += all_edges
+    if texture:
+        all_texture = Parallel(n_jobs=-1, prefer='threads')(delayed(_texture_filter)(filtered_img)
+                                for filtered_img in all_filtered)
+        features += itertools.chain.from_iterable(all_texture)
```
VolkerH, line 57: I've just been playing around with your trainable segmentation and I like it.
Just wanted to point out that you can avoid the dependency on `joblib.Parallel` (and arguably end up with a more concise code) if you use `ThreadPoolExecutor` from `concurrent.futures` like this:

```
    with ThreadPoolExecutor() as ex:
        all_filtered = list(ex.map(lambda s: filters.gaussian(img,s), sigmas))
        features = []
        if intensity:
            features += all_filtered
        if edges:
            all_edges = list(ex.map(filters.sobel, all_filtered))
            features += all_edges
        if texture:
            all_texture = list(ex.map(_texture_filter, all_filtered))
            features += itertools.chain.from_iterable(all_texture)
        return features
```

However, I don't know what the policy on starting threads within skimage is (if there is one, @jni)

grlee77: One place we do use threading currently is in the `cycle_spin` function. [The implementation there is based on dask](https://github.com/scikit-image/scikit-image/blob/master/skimage/restoration/_cycle_spin.py#L141-L144). Dask is currently one of our optional dependencies, so the function has an equivalent single-threaded code path that is used when dask is not present.

Note that for that function, our default for `num_workers` is to use all available cores.

I think I did originally do the cycle_spin implementation using `concurrent.futures` (see #2647), but we switched to `dask` during code review. One advantage of Dask is that you can use different schedulers ('single-threaded', 'threads', 'processes') or even distributed across a number of worker nodes. Note that we do not specify the scheduler in the call to `computer` in the `cycle_spin` example, so the default threaded scheduler unless the user has [configured a different scheduler](https://docs.dask.org/en/latest/scheduling.html) prior to calling the function.

grlee77: One change relative to the time of the older `cycle_spin` discussion is that we have now dropped support for older python that didn't have `concurrent.futures` in the standard library.

jni:
@emmanuelle thanks for the updates! I would prefer to avoid the joblib dependency. As @grlee77 noted, if we need parallelism, I think dask is a better choice, since we already use it several places and our conda-forge recipe already depends on it. In this context, it might be even better, because you can use `dask.array.map_overlap(feature_function, chunked_image, boundary=5)` (or however much overlap you want), see [here](https://docs.dask.org/en/latest/array-overlap.html).

skimage/segmentation/trainable_segmentation.py
emmanuelle: thanks for the feedback here, I'll try both `concurrent.futures` and `dask` here. Given the fact that `dask` is only an optional dep, maybe it's safer to use `concurrent.futures`? I need to try it first anyway.

VolkerH: I played with dask as well this afternoon (local time) and used `dask.bag` to come up with this.

```
    import dask.bag as db
    ...
    sigmas_db = db.from_sequence(sigmas)
    all_filtered = sigmas_db.map(lambda s: filters.gaussian(img,s)).compute(scheduler='threads')
    features = []
    if intensity:
        features += all_filtered
    if edges:
        all_edges = db.from_sequence(all_filtered).map(filters.sobel).compute(scheduler='threads')
        features += all_edges
    if texture:
        all_texture = db.from_sequence(all_filtered).map(_texture_filter).compute(scheduler='threads')
        features += itertools.chain.from_iterable(all_texture)
    return features
```

Running this, I had the subjective feeling that it was slower than the `concurrent.futures` example.
When trying to benchmark this on a similar toy example, I could only confirm a very minor difference in speed, [see here](https://github.com/VolkerH/PythonSnippets/blob/master/daskbag_vs_ThreadPoolExecutor/daskbag_vs_threadpoolexecutor.ipynb).

Off topic, but given that @grlee77 commented... the original reason I looked at this was to see whether this could be quickly ported to `cupyimg` (for the features) and `cuml` (for the RandomForestClassifier). I guess it should be straight forward.

However, `dask.bag` required installation of an additional dependency `fsspec`.

grlee77: > Off topic, but given that @grlee77 commented... the original reason I looked at this was to see whether this could be quickly ported to cupyimg (for the features) and cuml (for the RandomForestClassifier). I guess it should be straight forward.

Yes, not all features have been ported to `cupyimg`, but it looks like the ones needed here have been. In that case it should be pretty straightforward to try!

emmanuelle:
@VolkerH @grlee77 I've implemented parallelism using dask if available, and concurrent.futures if not. I confirm the analysis of @VolkerH that dask is a bit slower (about 10%), but maybe it is a bit more robust, I'm not sure. I don't know if dask does anything about oversubscription for example (this is something joblib takes care of).

emmanuelle:
Should I move `multiscale_basic_features` to the `feature` module instead? I was not sure whether it could be used outside of the context of the trainable segmentation, but probably it belongs more to `feature`?

jni:
Oops apologies that I missed your latest response @emmanuelle! I definitely think feature computations should go in feature! :joy:

emmanuelle:
@jni done! One thing I'd like to know from reviewers is which gallery example makes more sense, the one with the functional API or the object-oriented one. Thanks :-)!

skimage/feature/_basic_features.py
```diff
@@ -0,0 +1,186 @@
...
+    -------
+    features : list
+        List of features, each element of the list is an array of shape as img.
+    """
+    # computations are faster as float32
+    img = np.ascontiguousarray(img_as_float32(img))
+    sigmas = np.logspace(
+        np.log2(sigma_min),
+        np.log2(sigma_max),
+        num=int(np.log2(sigma_max) - np.log2(sigma_min) + 1),
```
jni, line 83: Should we allow setting this via  `num_sigma=None`?

skimage/feature/_basic_features.py
```diff
@@ -0,0 +1,186 @@
...
+    if has_dask:
+        out_sigmas = [
+            dask.delayed(_singlescale_basic_features_singlechannel)(
+                img, s, intensity=intensity, edges=edges, texture=texture
+            )
+            for s in sigmas
+        ]
+        features = itertools.chain.from_iterable(
+            dask.compute(*out_sigmas, num_workers=num_workers)
+        )
```
jni, line 96: I think the advantage of dask here would be to use `dask.array.map_blocks` or `dask.array.map_overlap` (depending on how you want to order these things). Using delayed + compute is exactly equivalent to the threadpoolexecutor so we might as well drop the dask path?

skimage/feature/_basic_features.py
```diff
@@ -0,0 +1,186 @@
...
+    -------
+    features : np.ndarray
+        Array of shape ``(n_features,) + image.shape``
+    """
+    if not any([intensity, edges, texture]):
+        raise ValueError(
+                "At least one of ``intensity``, ``edges`` or ``textures``"
+                "must be True for features to be computed."
+                )
+    if image.ndim >= 3 and multichannel:
```
jni, line 162: I'd suggest a restructure, `if not multichannel: image = image[..., np.newaxis]`, then have a single unified code path.

skimage/segmentation/trainable_segmentation.py
```diff
@@ -0,0 +1,180 @@
...
+        ``(m_features, *labels.shape)``. If None,
+        :func:`skimage.segmentation.multiscale_basic_features` is used.
+
+    Returns
+    -------
+    output : ndarray
+        Labeled array, built from the prediction of the classifier.
+    """
+    sh = features.shape
+    features = features.reshape((sh[0], np.prod(sh[1:]))).T
```
jni, line 165: @emmanuelle why did you choose to put the features in front and not at the end? This is not cache friendly at all! Or does sklearn prefer F-ordered arrays?

Also, note that you can reshape to `(sh0, -1)`. =)

jni (commented):
@emmanuelle as you have probably `predict`ed (=P) I strongly prefer the functional interface. The only thing I might consider changing is allowing `features` to be a callable that takes an image and returns a feature image or table. So, users can either pre-compute features, or, if they are going to throw them away, just pass the feature function to the predict/fit functions.

I like the idea of putting this in future. I have a sneaking suspicion that we have not hit the optimum API yet, but in future, I'd be super happy to get this out there to users!

skimage/feature/_basic_features.py
emmanuelle: Do you mean that instead of having the embarrassingly parallel loop over sigma here, we could instead chunk the `img` array on smaller blocks and compute all the sigmas on each of these blocks using `map_overlap`? I can try if this is what you mean.

skimage/feature/_basic_features.py
emmanuelle: done!

VolkerH:
It's been a while since I played with this. I just had another read through the code and I noticed that `fit_segmenter` also predicts on the whole training image. 
One could argue that the fitting should be separate from the prediction, i.e. that `fit_segmenter` should only return the trained classifier and the user can then use that to create the segmented image. This is just may way of thinking, quite possible that the added convenience of having training and prediction on the whole sample array in one go is preferred by others.

skimage/segmentation/trainable_segmentation.py
```diff
@@ -0,0 +1,180 @@
...
+
+    Raises
+    ------
+    NotFittedError if ``self.clf`` has not been fitted yet (use ``self.fit``).
+    """
+    training_data = features[:, labels > 0].T
+    training_labels = labels[labels > 0].ravel()
+    clf.fit(training_data, training_labels)
+    data = features[:, labels == 0].T
+    predicted_labels = clf.predict(data)
```
VolkerH, line 131: see my other comment. How about leaving the prediction out of here and just returning `clf` as an alternative?

skimage/segmentation/trainable_segmentation.py
emmanuelle: hum, indeed not cache-friendly. While building the array of features it made sense to compute an array for each feature and to stack them together, hence features at the front. But I should probably use `np.ascontiguousarray` here.

emmanuelle: or do you think it would make more sense to ask that the features array is of shape `img.shape + (n_features,)`? This is possible too.

emmanuelle:
> One could argue that the fitting should be separate from the prediction, i.e. that `fit_segmenter` should only return the trained classifier and the user can then use `predict` to create the segmented image. This is just may way of thinking, quite possible that the added convenience of having training and prediction on the whole sample array in one go is preferred by others.

This is a good summary of the dilemma here I think! However, if you're passing an image with already labeled regions, the labels should be used (ie the output label should be the same as the input), which will not be always the case if `fit_segmented` is used without passing labels. Therefore, it's not just a question of syntax.

VolkerH:
I see, you don't want the classifier to contradict manual labels. I actually didn't read that far down that function.

I don't know whether the trained classifier should not be able to override the manual labels. I am actually not sure how other trainable segmentation implementations handle this.  

If the prediction does not agree very well with the labels that should also show up in the confusion matrix of the classifier. 
My approach would be not to restore the original labels in the prediction if the classifier disagrees as it hides problems with the classification. The user can always restore the label values if they desire.

skimage/feature/_basic_features.py
emmanuelle: done !

jni:
I think I agree with Volker, actually! No need to predict in the same function as fitting. On the plus side, now that it's in `future`, we can have this debate over longer timescales. =D

skimage/segmentation/trainable_segmentation.py
jni: @emmanuelle I guess you can stack directly with the features in the tail dimension, no? I'm unclear whether we should go for `(npixels, nfeatures)` or `image.shape + (nfeatures,)`. I have a slight leaning towards the former — let's maximise compatibility with sklearn, as I think that is the purpose of this module.

emmanuelle: In `predict_segmenter` you need to have `image.shape` in the shape of `features` in order to be able to reshape the labels to the correct image shape. Actually it could be possible to support several shapes, and just assume that the last dimensions is for the features. This is what I'll do.

emmanuelle:
@jni I think I addressed all the points that you and Volker raised. The only point I was not sure of is the one related to dask (please see my question above).

skimage/feature/_basic_features.py
jni: @emmanuelle that was one option, another is to create see the different sigmas as a stack of images and use `map_blocks` to process over the different "slices" of the stack. My point was less about the exact way but rather to take advantage of dask.array syntax. If we don't, then dask.bag offers nothing over concurrent.futures, does it?

Incidentally, note that (unless things have changed), the default executor for a list of dask.delayed objects is multiprocessing ([docs](https://docs.dask.org/en/latest/api.html#dask.compute) are a bit ambiguous on this), so that might be the source of the performance difference....

skimage/feature/_basic_features.py
```diff
@@ -0,0 +1,190 @@
...
+                    sigmas,
+                )
+            )
+        features = itertools.chain.from_iterable(out_sigmas)
+    return features
+
+
+def multiscale_basic_features(
+    image,
+    multichannel=True,
```
sciunto, line 119: I think that usually, we set this parameter to false. The docstring says false.

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,175 @@
...
+        Array of features, with the last dimension corresponding to the number
+        of features, and the other dimensions are compatible with the shape of
+        the image to segment, or a flattened image.
+    clf : classifier object
+        trained classifier object, exposing a ``predict`` method as in
+        scikit-learn's API, for example an instance of
+        ``RandomForestClassifier`` or ``LogisticRegression`` classifier. The
+        classifier must be already trained, for example with
+        :func:`skimage.segmentation.fit_segmenter`.
+    features_func : function, optional
```
sciunto, line 146: This parameter is missing in the signature :)

skimage/feature/_basic_features.py
```diff
@@ -0,0 +1,190 @@
...
+        neighbourhoods before extracting features.
+    sigma_max : float, optional
+        Largest value of the Gaussian kernel used to average local
+        neighbourhoods before extracting features.
+    num_sigma : int, optional
+        Number of values of the Gaussian kernel between sigma_min and sigma_max.
+        If None, sigma_min multiplied by powers of 2 are used.
+    num_workers : int or None, optional
+        The number of parallel threads to use. If set to ``None``, the full
+        set of available cores are used
```
sciunto, line 75:
```suggestion
        set of available cores are used.
```

skimage/data/tests/test_data.py
```diff
@@ -140,6 +144,11 @@ def test_lfw_subset():
     """ Test that "lfw_subset" can be loaded."""
     data.lfw_subset()
 
+@skipif(not has_pooch, reason="pooch is not installed")
```
sciunto, line 147: Can be addressed elsewhere, but if we need to skip if pooch is not installed, we must be consistent for the other online contents.

skimage/feature/_basic_features.py
```diff
@@ -0,0 +1,190 @@
...
```
sciunto, line 159:
```suggestion
        set of available cores are used.
```

skimage/feature/_basic_features.py
jni: Agreed!

skimage/future/trainable_segmentation.py
emmanuelle: oops yes, it's from an old version of the API. Thanks!

skimage/feature/_basic_features.py
emmanuelle: ok, got it, I'd need to spend a bit of time trying this out. Since I'm trying to prioritize my time for helping out with the release, maybe I could remove the dask code here for now so that we can merge the PR (unless other points are to be changed), and then if dask makes a difference here I'll integrate it in a subsequent PR?

skimage/data/tests/test_data.py
emmanuelle: indeed this was not required, I removed the skip.

skimage/feature/_basic_features.py
emmanuelle: I've removed the dask backend

emmanuelle:
@scikit-image/core please review when you can as I've addressed all review comments.

skimage/feature/_basic_features.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    if not any([intensity, edges, texture]):
+        raise ValueError(
+                "At least one of ``intensity``, ``edges`` or ``textures``"
+                "must be True for features to be computed."
+                )
+    if image.ndim < 3:
+        multichannel = False
+    if not multichannel:
+        image = image[..., np.newaxis]
```
rfezzani, line 157:
```suggestion
    image = np.atleast_3d(image)
```

skimage/data/__init__.py
```diff
@@ -1076,4 +1101,4 @@ def brain():
     .. [1] https://graphics.stanford.edu/data/voldata/
 
     """
-    return _load("data/brain.tiff")
+    return _load("data/brain.tiff")
```
rfezzani, line 1104: Missing end of line :wink:

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,170 @@
...
+    -------
+    compute_features
+    fit
+    predict
+    """
+
+    def __init__(self, clf=None, features_func=None):
+        if clf is None:
+            try:
+                from sklearn.ensemble import RandomForestClassifier
```
rfezzani, line 41: Can't we group the `sklearn` imports in the first `try: ... except: ...` block and use the `has_sklearn` variable defined there?

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,170 @@
...
+            Labeled array of shape compatible with ``image`` (same shape for a
+            single-channel image). Labels >= 1 correspond to the training set and
+            label 0 to unlabeled pixels to be segmented.
+        image : ndarray
+            Input image, which can be grayscale or multichannel, and must have a
+            number of dimensions compatible with ``self.features_func``.
+        """
+        if self.features is None:
+            self.compute_features(image)
+        output, clf = fit_segmenter(labels, self.features, self.clf)
```
rfezzani, line 75: I see in `fit_segmenter` definition that it only returns `clf`!!! Did I missed something?

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,170 @@
...
+        :func:`skimage.segmentation.fit_segmenter`.
+
+    Returns
+    -------
+    output : ndarray
+        Labeled array, built from the prediction of the classifier.
+    """
+    sh = features.shape
+    if features.ndim > 2:
+        features = features.reshape((np.prod(sh[:-1]), sh[-1]))
```
rfezzani, line 154:
```suggestion
        features = features.reshape((-1, sh[-1]))
```

rfezzani (commented):
Thank you @emmanuelle, and sorry for reviewing so late. I left some suggestions and questions :wink:

skimage/feature/_basic_features.py
```diff
@@ -0,0 +1,172 @@
...
+    img = np.ascontiguousarray(img_as_float32(img))
+    if num_sigma is None:
+        num_sigma = int(np.log2(sigma_max) - np.log2(sigma_min) + 1)
+    sigmas = np.logspace(
+        np.log2(sigma_min),
+        np.log2(sigma_max),
+        num=int(np.log2(sigma_max) - np.log2(sigma_min) + 1),
+        base=2,
+        endpoint=True,
+    )[::-1]
```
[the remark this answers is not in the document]
emmanuelle: right, thanks for the tip. In fact I'm hesitating about the order. The idea of having the largest `sigmas` first was in terms of speed optimization (with several jobs running in parallel if the last remaining one is a short one it's better than if you have to wait for a single long job to terminate). But then the order of the features is a bit counter-intuitive. In fact I could probably reorder after the computation (and add a note about why it is done this way).

skimage/feature/_basic_features.py
emmanuelle: hum, I think this would not work for a 3D image (with 3 spatial dimensions)

rfezzani: yes, I see!

skimage/future/trainable_segmentation.py
emmanuelle: good catch, thanks! This also means I need to add more tests :-).

skimage/future/trainable_segmentation.py
emmanuelle: yes good idea

skimage/data/__init__.py
emmanuelle: ok!

skimage/feature/_basic_features.py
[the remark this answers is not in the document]
emmanuelle: I ended up sorting from min to max, which is what people would expect I think.

doc/examples/segmentation/plot_trainable_segmentation.py
```diff
@@ -0,0 +1,112 @@
...
+    ax[1].set_xlabel("$\\sigma$")
+
+fig.tight_layout()
+
+##############################################################################
+# Fitting new images
+# ------------------
+#
+# If you have several images of similar objects acquired in similar conditions,
+# you can used the classifier trained with `fit_segmenter` to segment other
```
emmanuelle, line 98:
```suggestion
# you can use the classifier trained with `fit_segmenter` to segment other
```

emmanuelle:
Thank you for the review @rfezzani I think I have now addressed your comments.

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,164 @@
...
+        else:
+            self.clf = clf
+        self.features_func = features_func
+
+    def compute_features(self, image):
+        if self.features_func is None:
+            self.features_func = multiscale_basic_features
+        self.features = self.features_func(image)
+
+    def fit(self, labels, image=None):
```
rfezzani, line 55: Why `image` is set to `None` by default?

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,164 @@
...
+        image : ndarray
+            Input image, which can be grayscale or multichannel, and must have a
+            number of dimensions compatible with ``self.features_func``.
+
+        Raises
+        ------
+        NotFittedError if ``self.clf`` has not been fitted yet (use ``self.fit``).
+        """
+        if self.features_func is None:
+            self.features_func = multiscale_basic_features
```
rfezzani, line 87: `multiscale_basic_features` is not defined in this bloc...

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,164 @@
...
+                    "Please install scikit-learn or pass a classifier instance"
+                    "to TrainableSegmenter."
+                )
+        else:
+            self.clf = clf
+        self.features_func = features_func
+
+    def compute_features(self, image):
+        if self.features_func is None:
+            self.features_func = multiscale_basic_features
```
rfezzani, line 52:
```suggestion
```
This part has to be in the `__init__`

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,164 @@
...
+    -------
+    clf : classifier object
+        classifier trained on ``labels``
+
+    Raises
+    ------
+    NotFittedError if ``self.clf`` has not been fitted yet (use ``self.fit``).
+    """
+    training_data = features[labels > 0]
+    training_labels = labels[labels > 0].ravel()
```
rfezzani, line 119:
```suggestion
    mask = labels > 0
    training_data = features[mask]
    training_labels = labels[mask].ravel()
```

rfezzani (commented):
Thank you again @emmanuelle, I left some more comments. An other suggestion: can't `TrainableSegmenter.features` be a property instead of defining and using the `compute_features` method?

skimage/future/trainable_segmentation.py
emmanuelle: I'm not sure I understand `multiscale_basic_features` is imported at the beginning of the file?

skimage/future/trainable_segmentation.py
emmanuelle: Thanks for the review @rfezzani; having `features` a `property` is an interesting idea but I'm not sure how this would work out. The syntax I can think of would be like:
```python
@property
    def features(self):
        return self.features_func(self.image)

    @features.setter
    def features(self, image):
        return self.features_func(image)
```
but then in the `fit` method I would need to write (using the setter):
```python
self.features = image
```
which looks misleading to me. Or did you have something else in mind? Thanks!

skimage/future/trainable_segmentation.py
rfezzani: My mistake, I though it was defined in `__init__`... Sorry, :pray:

skimage/future/trainable_segmentation.py
rfezzani: You need to store `image` in `self` as you did in the definition of the `features` property :wink::
```python
@property
def features(self):
    return self.features_func(self.image)

def fit(self, labels, image=None):
    self.image = image
    clf = fit_segmenter(labels, self.features, self.clf)
```

but the drawback is that you end up calling `self.features_func` whenever you access `self.features` :slightly_frowning_face: .
Using `property` is probably not a good idea :slightly_smiling_face:

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        if self.features_func is None:
+            self.features_func = multiscale_basic_features
+        features = self.features_func(image)
+        return predict_segmenter(features, self.clf)
+
+
+def fit_segmenter(labels, features, clf):
+    """
+    Segmentation using labeled parts of the image and a classifier.
```
jni, line 94:
```suggestion
    """Segmentation using labeled parts of the image and a classifier.
```

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,165 @@
...
+    mask = labels > 0
+    training_data = features[mask]
+    training_labels = labels[mask].ravel()
+    clf.fit(training_data, training_labels)
+    return clf
+
+
+def predict_segmenter(features, clf):
+    """
+    Segmentation of images using a pretrained classifier.
```
jni, line 127:
```suggestion
    """Segmentation of images using a pretrained classifier.
```

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,165 @@
...
+        image : ndarray
+            Input image, which can be grayscale or multichannel, and must have a
+            number of dimensions compatible with ``self.features_func``.
+        """
+        self.compute_features(image)
+        clf = fit_segmenter(labels, self.features, self.clf)
+
+    def predict(self, image):
+        """
+        Segment new image using trained internal classifier.
```
jni, line 74:
```suggestion
        """Segment new image using trained internal classifier.
```

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,165 @@
...
+        self.features_func = features_func
+
+    def compute_features(self, image):
+        if self.features_func is None:
+            self.features_func = multiscale_basic_features
+        self.features = self.features_func(image)
+
+    def fit(self, labels, image=None):
+        """
+        Train classifier using partially labeled (annotated) image.
```
jni, line 57:
```suggestion
        """Train classifier using partially labeled (annotated) image.
```

skimage/future/trainable_segmentation.py
jni: Yes, it's puzzling... 
```suggestion
    def fit(self, labels, image):
```

skimage/future/trainable_segmentation.py
```diff
@@ -0,0 +1,165 @@
...
+except ImportError:
+    has_sklearn = False
+
+    class NotFittedError(Exception):
+        pass
+
+
+class TrainableSegmenter(object):
+    """
+    Estimator for classifying pixels.
```
jni, line 16:
```suggestion
    """Estimator for classifying pixels.
```

jni (approved):
@emmanuelle with some very minor changes I think this is ready to go in! @scikit-image/core anyone else want to approve and click the green button? =)

skimage/future/trainable_segmentation.py
emmanuelle: I don't remember, this is puzzling indeed. And to mimic scikit-learn's syntax, this should be `fit(self, image, labels)` (it's `fit(X_train, y_train)`). Let me change this ASAP.

emmanuelle:
Thanks for the suggestions @jni, they have been integrated.

jni:
I don't know how to approve this harder... :stuck_out_tongue_winking_eye:

emmanuelle:
@rfezzani is the PR ok for you now?

rfezzani (approved):
LGTM! Sorry again for the delay and thank you @emmanuelle :tada:

PR #4740, opened 2020-05-21
Unwrap decorators before resolving link to source (sphinx.ext.linkcode)

### LINKED ISSUE #2221: docs link to decorator rather than function itself
e.g. http://scikit-image.org/docs/dev/api/skimage.exposure.html#equalize-adapthist
the `[source]` link goes to the `adapt_rgb` decorator: https://github.com/scikit-image/scikit-image/blob/master/skimage/color/adapt_rgb.py#L35

A quick test locally suggests that at least Sphinx 1.4.5 on Python 3.5.2 is able to link to the inner decorator object instead.

lagru:
## Description

Ensures that the source links in the HTML documentation, generated by [sphinx.ext.linkcode](https://www.sphinx-doc.org/en/master/usage/extensions/linkcode.html), point to the source of wrapped objects rather then the source of their decorator(s).

In case of decorated functions this will still not pint to the line with the `def` statement because the inspect module seems to consider the decorator as part of the objects source.

See https://github.com/scikit-image/scikit-image/issues/2221#issuecomment-631986646.

Closes #2221 

## For reviewers

  later.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -413,6 +413,9 @@ def linkcode_resolve(domain, info):
         except:
             return None
 
+    # Strip decorators which would resolve to the source of the decorator
+    obj = inspect.unwrap(obj)
+
     try:
         fn = inspect.getsourcefile(obj)
     except:


lagru:
> In case of decorated functions this will still not pint to the line with the def statement because the inspect module seems to consider the decorator as part of the objects source.

This could be alleviated by highlighting the complete function instead of only the first line. I've already used it and found it quite pleasing so I'd be happy to propose a PR..

rfezzani:
Thank you @lagru, I was looking for this solution for a while!

rfezzani:
Shouldn't this PR be back ported?

jni:

jni:
@rfezzani why not. =D

lagru:
> I'd be in favour of your additional proposed change.

See #4746.

PR #4746, opened 2020-05-22
Highlight all source lines with HTML doc "source" links (sphinx.ext.linkcode)

lagru:
## Description

Previously only the first line of an object's source was highlighted by the links enabled by sphinx.ext.linkcode.

In comparison to `inspect.findsource`, `getsourcelines` add's the "+ 1" internally and trims unwanted lines in front of the object's source.

## For reviewers

  later.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -424,14 +424,12 @@ def linkcode_resolve(domain, info):
         return None
 
     try:
-        source, lineno = inspect.findsource(obj)
+        source, start_line = inspect.getsourcelines(obj)
     except:
-        lineno = None
-
-    if lineno:
-        linespec = "#L%d" % (lineno + 1)
-    else:
         linespec = ""
+    else:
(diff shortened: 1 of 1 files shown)


jni:
@lagru I don't understand this change. Can you elaborate on what it's doing? Perhaps with screenshots? 😊 😬

lagru:
Of course. Here is an example: 

The `[source]` link below the signature of [`skimage.color.label2rgb`](https://scikit-image.org/docs/dev/api/skimage.color.html#skimage.color.label2rgb) currently returns an [URL](https://github.com/scikit-image/scikit-image/blob/master/skimage/color/colorlabel.py#L74) that will only highlight the first line of the function (in this case the line adding the decorator):

image

With the proposed changes the new [URL](https://github.com/scikit-image/scikit-image/blob/master/skimage/color/colorlabel.py#L74-L118) highlights the full function:

image

jni:
😍

PR #4751, opened 2020-05-26
Optimize peak_local_max trivial image test

### LINKED ISSUE #3990: Trivial image check for peak_local_max
## Description

`peak_local_max`, current tests for trivial images. That is, if an image is constant, and exactly constant, then it will return that no local peaks are found. If even one pixel is off, then it will return that many local peaks were found.

I think this test is too much of an edge case, and should simply be removed possibly without deprecation.

The relevant code is:

```python
    # no peak for a trivial image
    if np.all(image == image.flat[0]):
        if indices is True:
            return np.empty((0, 2), np.int)
        else:
            return out
```
(not linking to the original code since it will be obsolete soon).

I think this is *such* an edge case that it isn't worth the cost imposed on checking equality on every other user of this function.

1. Real cameras cannot actually take images with a constant value. Either dark noise, or shot noise from incomming photons will cause natural fluctuations in an image.
2. Images that are computer generated have better ways of testing if they are of a constant value.

The users that may have trivial images in their analysis pipelines should find better ways of detecting them.

We've tried to think of better ways for testing for the trivial case, but due to the amount of options that the peak_local_max function provides, it seems like it is more complicated than it seems at first glance:

https://github.com/scikit-image/scikit-image/pull/3984#issuecomment-508541246

cc: @clementkng and @brownmk
xref: https://github.com/scikit-image/scikit-image/pull/3984

rfezzani:
## Description

Closes #3990 

Using a generator to perform the [trivial image check](https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/peak.py#L192) in `peak_local_max` makes the operation O(1) instead of O(n).

Some timings to show the difference:

```python
In [1]: x = np.random.rand(1000, 1000)                                                                                 

In [2]: %timeit np.all(x == x.flat[0])                                                                                 
718 µs ± 1.03 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [3]: %timeit all((v == x.flat[0] for v in np.nditer(x)))                                                            
6.29 µs ± 35.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [4]: x = np.ones((1000, 1000))                                                                                      

In [5]: %timeit np.all(x == x.flat[0])                                                                                 
760 µs ± 2.39 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [6]: %timeit all((v == x.flat[0] for v in np.nditer(x)))                                                            
2.3 s ± 11.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

The speedup is more then 100× for a random image!

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/peak.py
+++ b/skimage/feature/peak.py
@@ -26,8 +26,6 @@ def _get_peak_mask(image, min_distance, footprint, threshold_abs,
     Return the mask containing all peak candidates above thresholds.
     """
     if footprint is not None:
-        if footprint.size == 1 and min_distance == 1:
-            return np.ones_like(image, dtype=bool)
         image_max = ndi.maximum_filter(image, footprint=footprint,
                                        mode='constant')
     else:
@@ -39,10 +37,6 @@ def _get_peak_mask(image, min_distance, footprint, threshold_abs,
     else:
         threshold = threshold_abs
     mask &= image > threshold
-
-    # no peak for a trivial image
-    if np.count_nonzero(mask) == image.size:
-        mask[:] = False
     return mask
 
 
@@ -165,6 +159,8 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
     array([[10, 10, 10]])
 
     """
+    out = np.zeros_like(image, dtype=np.bool)
+
     threshold_abs = threshold_abs if threshold_abs is not None else image.min()
 
     if isinstance(exclude_border, bool):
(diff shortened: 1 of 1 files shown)


rfezzani:
I performed some more experiments for performance comparison:

```python
In [8]: for i in range(0, 1000000, 10000): 
   ...:     x = np.ones((1000, 1000)) 
   ...:     x[np.unravel_index(i, (1000, 1000))] = 0 
   ...:     %timeit all((v == x.flat[0] for v in np.nditer(x))) 
```

Compiling the results in a plot gives:
timings

The proposed approach only makes sense if we assume that the condition `image[idx] != image[0]` occurs  early when traversing `image` in most of the cases, otherwise, the O(n) approach is the best choice!

rfezzani:
`peak_local_max` trivial image test is better managed in #4760. This PR is no more relevant, closing ;)

PR #4753, opened 2020-05-26
Deprecate (min_distance < 1) and (footprint.size < 2) in peak_local_max

### LINKED ISSUE #4749: Deprecate edge cases in peak_local_max
## Description

While investigating for #3990, I found that `peak_local_max` manages two edge cases that can be deprecated:

- `min_distance=0` and `footprint=None`
- `footprint.size=1`

in this cases `peak_local_max` is equivalent to applying a threshold. I think that these settings must be forbidden.

## Version information

```python
[GCC 7.3.0]
Linux-4.4.0-179-generic-x86_64-with-debian-stretch-sid
scikit-image version: 0.18.dev0
numpy version: 1.18.4

```

rfezzani:
## Description

Closes #4749

In  these two deprecated cases, `peak_local_max` acts as thresholding:
```python
>>> from skimage.feature import peak_local_max
>>> img = np.random.rand(50, 50)
>>> thresh = 0.2
>>> np.all((img > thresh) == peak_local_max(img, min_distance=0, threshold_abs=thresh,
                                            indices=False, exclude_border=False))
True
>>> np.all((img > thresh) == peak_local_max(img, footprint=np.array([[1]]),
                                            threshold_abs=thresh, indices=False,
                                            exclude_border=False))
True
```

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.


skimage/feature/peak.py
```diff
@@ -163,6 +164,20 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
 
     threshold_abs = threshold_abs if threshold_abs is not None else image.min()
 
+    if min_distance < 1:
+        warn("min_distance lower then 1 is deprecated. "
```
sciunto, line 168:
```suggestion
        warn("min_distance lower than 1 is deprecated. "
```

skimage/feature/peak.py
```diff
@@ -163,6 +164,20 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
+             "In this case, peak_local_max acts as finding "
+             "image > max(threshold_abs, threshold_rel). In version 0.20 "
+             "this warning will be replaced by a ValueError.",
+             FutureWarning, stacklevel=2)
+
+    if footprint is not None and footprint.size == 1:
+        warn("footprint.size lower then 2 is deprecated. "
```
sciunto, line 175:
```suggestion
        warn("footprint.size lower than 2 is deprecated. "
```

skimage/feature/tests/test_peak.py
```diff
@@ -341,14 +342,19 @@ def test_disk(self):
         '''
         image = np.random.uniform(size=(10, 20))
         footprint = np.array([[1]])
-        result = peak.peak_local_max(image, labels=np.ones((10, 20)),
-                                     footprint=footprint,
-                                     min_distance=1, threshold_rel=0,
-                                     threshold_abs=-1, indices=False,
-                                     exclude_border=False)
+
+        with expected_warnings(["footprint.size lower then 2 is deprecated."]):
```
sciunto, line 346:
```suggestion
        with expected_warnings(["footprint.size lower than 2 is deprecated."]):
```

skimage/feature/tests/test_peak.py
```diff
@@ -341,14 +342,19 @@ def test_disk(self):
...
+            result = peak.peak_local_max(image, labels=np.ones((10, 20)),
+                                         footprint=footprint,
+                                         min_distance=1, threshold_rel=0,
+                                         threshold_abs=-1, indices=False,
+                                         exclude_border=False)
         assert np.all(result)
-        result = peak.peak_local_max(image, footprint=footprint, threshold_abs=-1,
-                                     indices=False, exclude_border=False)
+        with expected_warnings(["footprint.size lower then 2 is deprecated."]):
```
sciunto, line 353:
```suggestion
        with expected_warnings(["footprint.size lower than 2 is deprecated."]):
```

skimage/feature/tests/test_peak.py
```diff
@@ -391,7 +397,9 @@ def test_threshold_rel_default(self):
         assert_equal(peak.peak_local_max(image), [[2, 2]])
 
         image[2, 2] = 0
-        assert len(peak.peak_local_max(image, min_distance=0)) == image.size - 1
+        with expected_warnings(["min_distance lower then 1 is deprecated."]):
```
sciunto, line 400:
```suggestion
        with expected_warnings(["min_distance lower than 1 is deprecated."]):
```

sciunto (approved):

sciunto:
@scikit-image/core We need a second review over here.

@rfezzani Could you rebase please?

skimage/feature/peak.py
```diff
@@ -172,6 +173,20 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
+    if min_distance < 1:
+        warn("min_distance lower than 1 is deprecated. "
+             "In this case, peak_local_max acts as finding "
+             "image > max(threshold_abs, threshold_rel). In version 0.20 "
+             "this warning will be replaced by a ValueError.",
+             FutureWarning, stacklevel=2)
+
+    if footprint is not None and footprint.size == 1:
+        warn("footprint.size lower than 2 is deprecated. "
+             "In this cases, peak_local_max acts as finding "
```
alexdesiqueira, line 185: `"In this case (...)`

alexdesiqueira:
Thank you for that @rfezzani! I have a question, though. Wouldn't it be better for the user to still have the option to put these values, receiving a warning instead of a `ValueError`? I would personally think in putting a `Notes` section explaining what happens for `min_distance < 1`, `footprint.size < 2`, and then presenting the warnings you suggested when the user uses these arguments. My two BRL cents, though!

rfezzani:
> Wouldn't it be better for the user to still have the option to put these values, ...

That's in fact another option @alexdesiqueira, I am OK with it.
@scikit-image/core, any opinion on this: deprecation or simple warning + doc?

emmanuelle:
@rfezzani is this idea of forbidding these modes (when the deprecation cycle is complete) that it is inefficient to use `peak_local_max` for something which can be done in a faster and clearer way? 

I think I'd prefer a simple warning here, because if for some reason users want to iterate over the values of `footprint` or `min_distance` it's not very nice to have them add an if/else block to handle the corner cases (since the output of `peak_local_max` is valid).

rfezzani:
@alexdesiqueira and @emmanuelle, I replaced the deprecation by a simple `RuntimeWarning`.

skimage/feature/peak.py
grlee77: Perhaps not intentionally, but previously an array-like footprint did work. (e.g. `footprint=[[0, 1, 0], [1, 1, 1], [0, 1, 0]]`), but now this would result in an AttributeError. The docstring does indicate that `footprint` should be an ndarray, though, so perhaps it is okay to leave it as is.

If we want to maintain array-like support, use something like:
```Python
if footprint is not None:
    footprint = np.asarray(footprint)
    if footprint.size == 1:
        warn(...
```

grlee77 (approved):
one comment to consider about array-like footprint, but otherwise this seems good to go

skimage/feature/peak.py
rfezzani: Good catch, I will keep the `array-like` support :wink:

rfezzani:
Thank you @grlee77! I kept the support for array-like `footprint` as you suggested.

jni (approved):
In the long run, we should use the warnings to point to the alternative `local_maxima`. But this shouldn't hold up this improvement any longer — that can come in a later PR. Thank you @rfezzani!

rfezzani:
:tada: Thank you @jni

PR #4760, opened 2020-05-27
Refactor and fix peak_local_max

### LINKED ISSUE #2592: peak_local_max fails to enforce min_distance
## Description
I am trying to extract local maxima from a distance transform (made using openCV).  When I run: coordinates = peak_local_max(distance_xform, min_distance=100, num_peaks=12) 
The following values are in coordinates: [[2775 2647] [944 1540] [919 2420] [1486 1192] [899 2099] [900 2054] [900 2055] [900 2056] [900 2057] [900 2058] [900 2059] [900 2060]]

If you look at the distance transform image there is a line one pixel wide on the horizontal axis extending from 1922 to 2228 at 900. All pixels on this line have the same  value. It seems this causes peak_local_max to ignore min_distance and pick peaks 1 pixel adjacent to each other along this line.

I am using:
Ubuntu 14.04.2 LTS
skimage 0.9.3
Python 2.7.6

[If reporting a bug, attach the entire traceback from Python.] (No traceback: functional error)
[If proposing an enhancement/new feature, provide links to related articles, reference examples, etc.]

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [x] Code example
- [x] Relevant images (if any)
- [x] Operating system and version
- [x] Python version
- [x] scikit-image version (run `skimage.__version__`)

outfile

### LINKED ISSUE #3990: Trivial image check for peak_local_max

### LINKED ISSUE #4756: Peaks may not be detected in peak_local_max when label is provided
## Description

Because background values are implicitly set to zero ([here](https://github.com/scikit-image/scikit-image/blob/master/skimage/feature/peak.py#L215)), peaks may be not detected when input image contains negative values.

## Way to reproduce

```python
>>> from skimage.feature import peak_local_max
>>> img = -0.2 *  np.ones((5, 5))
>>> img[3, 3] = -0.1
>>> img
array([[-0.2, -0.2, -0.2, -0.2, -0.2],
       [-0.2, -0.2, -0.2, -0.2, -0.2],
       [-0.2, -0.2, -0.2, -0.2, -0.2],
       [-0.2, -0.2, -0.2, -0.1, -0.2],
       [-0.2, -0.2, -0.2, -0.2, -0.2]])
>>> label = np.ones((5, 5), int)
>>> label[:, -1] = 0
>>> label
array([[1, 1, 1, 1, 0],
       [1, 1, 1, 1, 0],
       [1, 1, 1, 1, 0],
       [1, 1, 1, 1, 0],
       [1, 1, 1, 1, 0]])
>>> peak_local_max(img)
array([[3, 3]])
>>> peak_local_max(img, labels=label)
array([], shape=(0, 2), dtype=int64)
```

## Version information

```python
3.8.2 (default, May  7 2020, 20:00:49) 
[GCC 7.3.0]
Linux-4.4.0-179-generic-x86_64-with-glibc2.10
scikit-image version: 0.17.1
numpy version: 1.18.4
```

### LINKED ISSUE #5047: feature.peak_local_max changes labels
## Description

When **non sequential** `labels` are provided as input to `peak_local_max`, **the values get altered**.

If the function needs to update the values, it ought to **copy** the matrix.

I suppose this bug will be addresses by #4760 (there is a description "labels renumbering is avoided" in that PR).  But I didn't find any explicit description of this bug, which seems serious due to its silent nature, so I thought I should document it.

## Way to reproduce
```python
import numpy as np
from skimage import feature
```

```python
im = np.zeros([10, 10])
labels = np.zeros([10, 10]).astype(int)

label = 2
for x in range(0, 10, 2):
    for y in range(0, 10, 2):
        im[x,y] = 1
        labels[x:x+1, y:y+1] = label
        label += 2
```

```python
labels
```

    array([[ 2,  0,  4,  0,  6,  0,  8,  0, 10,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
           [12,  0, 14,  0, 16,  0, 18,  0, 20,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
           [22,  0, 24,  0, 26,  0, 28,  0, 30,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
           [32,  0, 34,  0, 36,  0, 38,  0, 40,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
           [42,  0, 44,  0, 46,  0, 48,  0, 50,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0]])

```python
res = feature.peak_local_max(im, labels=labels, num_peaks_per_label=1)
```

```python
labels
```

    array([[ 1,  0,  2,  0,  3,  0,  4,  0,  5,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
           [ 6,  0,  7,  0,  8,  0,  9,  0, 10,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
           [11,  0, 12,  0, 13,  0, 14,  0, 15,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
           [16,  0, 17,  0, 18,  0, 19,  0, 20,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0],
           [21,  0, 22,  0, 23,  0, 24,  0, 25,  0],
           [ 0,  0,  0,  0,  0,  0,  0,  0,  0,  0]])

```python

```

    3.8.5 (default, Jul 21 2020, 10:48:26) 
    [Clang 11.0.3 (clang-1103.0.32.62)]
    macOS-10.15.7-x86_64-i386-64bit
    scikit-image version: 0.17.2
    numpy version: 1.19.2

```python

```

## Version information
```python
```

```python
# your output here
3.8.5 (default, Jul 21 2020, 10:48:26) 
[Clang 11.0.3 (clang-1103.0.32.62)]
macOS-10.15.7-x86_64-i386-64bit
scikit-image version: 0.17.2
numpy version: 1.19.2
```

rfezzani:
## Description

In my attempt to fix #2592, I made a review of the code and found many issues (see #4756, #4749, #4752 and https://github.com/scikit-image/scikit-image/issues/3990#issuecomment-630272333). It appeared to me that a global refactoring of `peak_local_max` is necessary.

Multiple changes are applied in this PR:

- `min_distance` is enforced in the same way as in #4218 for `corner_peak` (see [here](https://github.com/scikit-image/scikit-image/compare/master...rfezzani:Fix_peack_local_max?diff=split&expand=1#diff-30c3e44d26344e4c11a27bdd1ae43731R18))
- edge cases as described in #4749 are managed [here](https://github.com/scikit-image/scikit-image/compare/master...rfezzani:Fix_peack_local_max?diff=split&expand=1#diff-30c3e44d26344e4c11a27bdd1ae43731R43)
- trivial image case issue #3990 is managed [here](https://github.com/scikit-image/scikit-image/compare/master...rfezzani:Fix_peack_local_max?diff=split&expand=1#diff-30c3e44d26344e4c11a27bdd1ae43731R51)
- private functions `_get_threshold` and `_get_excluded_border_width` are introduced for readability
-  labels managing is optimized, in particular:
    - `labels` [renumbering](https://github.com/scikit-image/scikit-image/compare/master...rfezzani:Fix_peack_local_max?diff=split&expand=1#diff-30c3e44d26344e4c11a27bdd1ae43731L200) is avoided
    - #4756 is fixed [here](https://github.com/scikit-image/scikit-image/compare/master...rfezzani:Fix_peack_local_max?diff=split&expand=1#diff-30c3e44d26344e4c11a27bdd1ae43731R276)
    - intermediary [masks creation](https://github.com/scikit-image/scikit-image/compare/master...rfezzani:Fix_peack_local_max?diff=split&expand=1#diff-30c3e44d26344e4c11a27bdd1ae43731L223-L238) is avoided
- return strategy is simplified.

If accepted, this PR allows to deprecate `corner_peak` that becomes a duplicate of `peak_local_max`.

Fixes #2592. Fixes #4756. Fixes #4756. Closes #3990. Fixes #5047 .

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


skimage/feature/peak.py
```diff
@@ -159,98 +232,79 @@ def peak_local_max(image, min_distance=1, threshold_abs=None,
...
-            out[nd_indices] = True
-            return out
+                                                    num_peaks_per_label,
+                                                    min_distance,
+                                                    p_norm)
 
-    # Non maximum filter
-    mask = _get_peak_mask(image, min_distance, footprint, threshold_abs,
-                          threshold_rel)
+            # transform the coordinate in gloal image indices space
```
mkcor, line 284:
```suggestion
            # transform coordinates in global image indices space
```

rfezzani:
Thank you @mkcor for your review ;)

rfezzani:
Can we plan this PR for 0.18 or is the schedule too short?

skimage/feature/peak.py
```diff
@@ -14,51 +14,119 @@ def _get_high_intensity_peaks(image, mask, num_peaks):
...
+
+        indices = tree.query_ball_point(coord, r=min_distance, p=p_norm)
+        rejected_peaks_indices = set()
+        for idx, candidates in enumerate(indices):
+            if idx not in rejected_peaks_indices:
+                candidates.remove(idx)
+                rejected_peaks_indices.update(candidates)
+
+        # Remove the peaks that are too close to each other
+        coord = np.delete(coord, tuple(rejected_peaks_indices), axis=0)
```
lagru, line 31: Several suggestions here: 

- We can skip this step if `min_distance` is 1, correct? 
- Also, it's probably best to update the description of that parameter in the docstring similar to #4218? I think the part with "2 * min_distance + 1" could be clearer.
- As already suggested in #4218 it's probably best to factor this out? Maybe even to Cython to get rid of the slow loop (doesn't have to be this PR)?

rfezzani: Thank you @lagru for your review, and sorry for my late answer, I have been too much busy these last few days :pray:.

> - We can skip this step if min_distance is 1, correct?

No, we also want to remove the adjacent peaks (I may be didn't get your point here :confused:).

>  - Also, it's probably best to update the description of that parameter in the docstring similar to #4218? I think the part with "2 * min_distance + 1" could be clearer.

:+1:

 > - As already suggested in #4218 it's probably best to factor this out? Maybe even to Cython to get rid of the slow loop (doesn't have to be this PR)?

Why not, I remember that we already mentioned this ;)

rfezzani: @lagru, I factored out this part as you suggested in the `skimage._shared.coord.ensure_spacing` function, I hope you agree with the function name and its API.

lagru: Thank you. The name seems fine to me. Did you intentionally not replace the appropriate code snippet in [corner_peaks](https://github.com/scikit-image/scikit-image/blob/580629f901fbc2d1c80d80fe15825ee9c12345f0/skimage/feature/corner.py#L1083-L1097) with `ensure_spacing`?

rfezzani: With this Fix of `peak_local_max`, `corner_peaks` becomes obsolete and can directly be replaced by the former.

rfezzani:
Please see https://github.com/scikit-image/scikit-image/issues/2592#issuecomment-635624261 for illustrating new `peak_local_max` behavior :wink:

odoublewen:
@rfezzani  Many thanks for your work on this!  I just made this bug report (#5047) and wanted to make sure you noticed it.

The phrase "labels renumbering is avoided" in your initial description makes it sound like your PR will address this bug.  

If any changes at all are made to the input labels matrix, then obviously it needs to be copied.

rfezzani:
As mentioned in #5047, This PR is not affected by the label change bug. I updated the associated test (`test_input_labels`) to cover the case of "non sequential" labels.

@scikit-image/core, this PR closes 4 issues, among which 3 are bugs! Reviews are welcome :wink:. Thank you all!

skimage/feature/peak.py
```diff
@@ -15,60 +15,111 @@ def _get_high_intensity_peaks(image, mask, num_peaks):
...
+                   num_peaks_per_label=np.inf, p_norm=np.inf):
     """Find peaks in an image as coordinate list or boolean mask.
 
     Peaks are the local maxima in a region of `2 * min_distance + 1`
     (i.e. peaks are separated by at least `min_distance`).
 
-    If there are multiple local maxima with identical pixel intensities
-    inside the region defined with `min_distance`,
-    the coordinates of all such pixels are returned.
-
```
jni, line 71: Why is this behaviour changed, @rfezzani?

jni: At any rate, the behaviour in plateaus should be explained in the docstring, whatever it is.

skimage/feature/tests/test_peak.py
```diff
@@ -61,7 +61,7 @@ def test_flat_peak(self):
         image = np.zeros((5, 5), dtype=np.uint8)
         image[1:3, 1:3] = 10
         peaks = peak.peak_local_max(image, min_distance=1)
-        assert len(peaks) == 4
+        assert len(peaks) == 1
```
jni, line 64: As above. I get concerned when tests change.

In fact, although many of the behaviour changes in this PR can be considered bug fixes, I would call this one an API change and that would require a deprecation cycle. Can it be easily restored?

skimage/feature/tests/test_peak.py
```diff
@@ -351,57 +355,35 @@ def test_four_quadrants(self):
                                          exclude_border=False)
         assert np.all(result == expected)
 
-    def test_disk(self):
```
jni, line 354: Why is this test removed?

skimage/feature/tests/test_peak.py
```diff
@@ -351,57 +355,35 @@ def test_four_quadrants(self):
...
-            result = peak.peak_local_max(image, footprint=footprint,
-                                         threshold_abs=-1,
-                                         indices=False,
-                                         exclude_border=False)
-        assert np.all(result)
-
     def test_3D(self):
         image = np.zeros((30, 30, 30))
-        image[15, 15, 15] = 1
+        image[16, 16, 16] = 1
```
jni, line 360: Why did you change this?

jni (commented):
@rfezzani great work! I'm most concerned about the behaviour change for plateaus, which I think was correct. I haven't caught up on all the comments and issues surrounding this PR but I agree that it should go in 0.18. Can you tell me what you think about my comments?

skimage/feature/peak.py
rfezzani: Thank you for your review @jni. This behavior is exactly what I wanted to fix in this PR as it is reported in #2592. This explains all the reported changes in the tests below :wink:.

skimage/feature/tests/test_peak.py
rfezzani: Previously for a peak `p_0`, any other peak `p_1` verifying

```
d(p_0, p_1) < min_distance
```
was removed.
In the current implementation the condition became

```
d(p_0, p_1) <= min_distance
```
IMHO, it corresponds more to `peak_local_max` doc statement
> Peaks are the local maxima in a region of 2 * min_distance + 1 (i.e. peaks are separated by at least min_distance).

skimage/feature/tests/test_peak.py
rfezzani: `footprint` no more overwrites `min_distance`. For this test to make sens, we need to set `min_distance` to 0.

skimage/feature/tests/test_peak.py
jni: I don't think that's right? I would say that `at least x` means, `x or more`. So peaks separated by exactly x should remain.

rfezzani:
This PR fixes some cells over-segmentation of the documentation example [Segment human cells (in mitosis)](https://scikit-image.org/docs/dev/auto_examples/applications/plot_human_mitosis.html#sphx-glr-auto-examples-applications-plot-human-mitosis-py):
```python
from matplotlib.patches import Rectangle

from skimage import color, feature, filters, measure, segmentation
from skimage.data import human_mitosis


class Box(Rectangle):
    def __init__(self, pos, w, h):
        super().__init__(pos, w, h, linewidth=1, edgecolor='r',
                         facecolor='none')


image = human_mitosis()

thresholds = filters.threshold_multiotsu(image, classes=3)
cells = image > thresholds[0]
distance = ndi.distance_transform_edt(cells)

local_maxi = feature.peak_local_max(distance, indices=False,
                                    min_distance=7)

markers, cell_count = measure.label(local_maxi, return_num=True)

print(f"{cell_count} cells detected")

segmented_cells = segmentation.watershed(-distance, markers, mask=cells)

fig, ax = plt.subplots(figsize=(8, 8))
ax.imshow(color.label2rgb(segmented_cells, bg_label=0))
ax.add_patch(Box((258, 230), 15, 15))
ax.add_patch(Box((120, 78), 15, 15))
ax.add_patch(Box((6, 298), 15, 15))
ax.add_patch(Box((324, 337), 15, 15))
ax.add_patch(Box((137, 435), 15, 15))
ax.add_patch(Box((132, 460), 15, 15))
ax.add_patch(Box((106, 476), 15, 15))

ax.set_axis_off()
fig.tight_layout()
plt.show()
```

gives

### Master branch

Figure_1

### This PR

Figure_2

Boxes are plotted around fixed over-segmented cells...

skimage/feature/tests/test_peak.py
rfezzani: :facepalm:... You're right! This must be fixed!

skimage/feature/peak.py
jni: @rfezzani I've pushed a note to this effect, please pull before any further changes!

rfezzani: Cool, thx @jni :wink:

skimage/feature/tests/test_peak.py
rfezzani: I am facing a dilemma @jni :confused: . I fixed this behavior, and when I ran the tests I realized that with this convention `min_distance=1` is equivalent to `min_distance=0`!

jni (approved):
Awesome! The latest changes make me happy. =) Thanks @rfezzani!

emmanuelle:
Looks like one of the doctests is not passing, please see https://travis-ci.com/github/scikit-image/scikit-image/jobs/429507939

rfezzani:
@jni, here is the error reported by @emmanuelle: in `corner_peaks` doctest we have
```python
>>> from skimage.feature import peak_local_max
>>> response = np.zeros((5, 5))
>>> response[2:4, 2:4] = 1
>>> response
array([[0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0.],
       [0., 0., 1., 1., 0.],
       [0., 0., 1., 1., 0.],
       [0., 0., 0., 0., 0.]])
>>> peak_local_max(response)
array([[2, 2]])
```
but while with the last fix
```python
>>> peak_local_max(response)
array([[2, 2],
       [2, 3],
       [3, 2],
       [3, 3]])
```
that's the result with `min_distance=1` which is equivalent to `min_distance=0`!!!

I think that in this case, `peak_local_max(response, min_distance=1)` must be equal to `array([[2, 2]])`. Am I wrong?

rfezzani:
This is again a **pixel centred** VS **grid centred** point of view problem I think!

jni:
> This is again a pixel centred VS grid centred point of view problem I think!

I don't think so? In either case the distance to the adjacent pixels is exactly 1 or sqrt(2)?

jni:
I think the new result is what I would expect. It's unclear to me why it wasn't that before. OR, if you want, change it to use `min_distance=2` and the problem should be solved! =)

rfezzani:
> >    This is again a pixel centred VS grid centred point of view problem I think!

> I don't think so? In either case the distance to the adjacent pixels is exactly 1 or sqrt(2)?

I say that in the sens that pixel (0, 1) is adjacent to pixel (0, 2) (the distance between them is 0 because they join at (0, 0.5)). But the points (0, 1) is separated from the point (0, 2) by a distance of 1.

rfezzani:
In other words, do we consider that a pixel is a point or "a little square"? I know, [pixels are not squares](http://alvyray.com/Memos/CG/Microsoft/6_pixel.pdf), but it can be seen as such :smile:

rfezzani:
BTW, `corner_peaks` returns `array([[2, 2]])` on this example, and this PR mimic the strategy implemented in `corner_peaks`.
A wanted to propose a PR to deprecate `corner_peaks` once this one is merged...

jni:
> I say that in the sens that pixel (0, 1) is adjacent to pixel (0, 2) (the distance between them is 0 because they join at (0, 0.5)). But the points (0, 1) is separated from the point (0, 2) by a distance of 1.

Well, I think this is a can of worms that we do not want to open, since this definition no longer satisfies the definition of a [metric](https://en.wikipedia.org/wiki/Metric_(mathematics)#Definition) between pixels.

What do you think of my proposal to add `min_distance=2` to the example so we can merge this? ;)

jni:
> A wanted to propose a PR to deprecate corner_peaks once this one is merged...

I'd be into that, I was always confused by that function. But probably for 0.19 as that is a major API change.

rfezzani:
OK, let's keep the old behavior considering that consecutive pixels are separated by a distance of 1 for now :wink:.
> I'd be into that, I was always confused by that function. But probably for 0.19 as that is a major API change.

Let's go for that too, but we will then need to decide which behavior to keep (`peak_local_max` or `corner_peaks`)

jni:
Ok, as far as I can tell, the failures have nothing to do with this PR. @scikit-image/core anyone else ready to push the green button?

skimage/_shared/tests/test_coord.py
```diff
@@ -0,0 +1,43 @@
...
+    # --- Empty input
+    assert ensure_spacing([], p_norm=p) == []
+
+    # --- Verified spacing
+    coord = np.random.randn(100, 2)
+
+    # --- 0 spacing
+    assert np.array_equal(coord, ensure_spacing(coord, spacing=0, p_norm=p))
+
+    # Spacing is choosed to be half the minimum distance
```
grlee77, line 18:
```suggestion
    # Spacing is chosen to be half the minimum distance
```

grlee77 (approved):
I will approve now, but possibly add something about these changes as in the criteria:

doc/release/release_dev.rst.

rfezzani:
Thank you @grlee77, I added the changes to the release note as you advised :wink:.

grlee77 (approved):

rfezzani:
:tada: thank you all for your reviews! I am happy to see this in 0.18 :smile:

stefanv:
@rfezzani Just wanted to give you a personal shout-out: the peak finder was performing poorly, and when I looked into what should be done to fixed it I discovered you had already done the work.  Thank you very much!

Issue #2696, opened 2017-07-11
Odd keyword name `min_size` for morphology's `remove_small_holes`


stefanv:
Should the `min_size` parameter be `max_size`, or something else entirely?  It's not a very intuitive naming.  At worst, we should document this well.

/cc @mcquin

mcquin:
I think `max_size` is more intuitive than `min_size`. When I read `remove_small_x`, I think about removing x up to `max_size` rather than keeping x above `min_size`.

The word "size" is another point of confusion. It could help to define "size" as "total pixels".

kne42:
I think that as `remove_small_holes` effectively _filters_ out the blobs based upon their area, a **threshold**-themed keyword name would make the most sense, such as `area_threshold`.

PR #4761, opened 2020-05-28
Deprecate min_size in remove_small_objects

### LINKED ISSUE #4003: API inconsistancy between remove_small_holes and remove_small_objects?

rfezzani:
## Description

Closes #4003

## Checklist

  existing benchmark
- [ ] Unit tests

## For reviewers

  later.


jni:
@rfezzani I feel like we didn't have the discussion I alluded to in #4003... :joy: Does @scikit-image/core feel like this is the best name for the parameter? Clearly some didn't. Personally, I think we should think very hard about what to name these parameters, *and* possibly don't deprecate until 1.0. We already deprecated once in the opposite direction. If I was a user that had updated from the last deprecation and got this warning now, I sure as hell wouldn't be happy!

stefanv:
Area threshold seems perfectly acceptable; is it worth changing?

rfezzani:
@jni, his PR is open for discussion, @scikit-image/core can decide here which is the best argument name ;)
I personally think that API changes must be introduced ASAP to reduce the number of potential impacted users, and to reduce the amount of work for 1.0 release.
BTW, previous modification was made without any deprecation :confused:

rfezzani:
> Area threshold seems perfectly acceptable; is it worth changing?

Maybe then `min_size` can be deprecated in `remove_small_object` in favor of `area_threshold`? :slightly_smiling_face:

jni:
>  BTW, previous modification was made without any deprecation :confused:

Um, no?

https://github.com/scikit-image/scikit-image/blob/v0.14.3/skimage/morphology/misc.py#L205-L208

rfezzani:
> Um, no?

I only looked in `TODO.txt` and in `doc/api_changes.rst` and found no mention, sorry :pray: 

BTW, I also noticed that the API changes introduced in **0.17** are not documented in `doc/api_changes.rst`... That's an other PR ;)

jni:
> BTW, I also noticed that the API changes introduced in 0.17 are not documented in doc/api_changes.rst... That's an other PR ;)

Should they? I thought they lived in the release notes.

rfezzani:
> Should they?

Do we have to maintain the `doc/api_changes.rst` file then?

stefanv:
To be clear, I have no strong preference between `min_area` and `area_threshold`; just thinking what would cause minimum churn.

stefanv:
(In fact, `min_area` may be more intuitive.  Oh, but `min_area` is not even an option, I see :).

@jni When we go N-D, what do we call area: volume?  `min_volume` sounds weird in 2D.  Hence size, but that's also not quite right, is it.

rfezzani:
@stefanv, all options are open, since the intention in #4003 is to have a consistent API between `remove_small_holes` and `remove_small_objects` :)

You have a good point here! `area_threshold` is maybe not well suited in nD context...
What about `size_threshold` ?

stefanv:
The more I think about it, `min_size` is quite good: the first part describes which direction the threshold goes, and the second part that it has to do with how big it is.

jni:
@stefanv my thought exactly. :facepalm:

... ...

So, I went back to find out the original discussion, and found #2696, opened by...

:joy:

jni:
I think the confusion is with the "double negative" with `remove`... `min_size`: the smallest hole that remains, `max_size`: the largest hole that is removed.

Perhaps just `size`?

jni:
There's one more option... Go for keyword only, "natural english" usage: `remove_small_holes(image, smaller_than=45)`?

stefanv:
That reads very nicely; not sure how well that maps to the entire library, though, and may lead to confusion if there are many parameters.

rfezzani:
What about just `threshold`?

hmaarrfk:
Personally, I think there is no magical name that will convey exactly what the parameter means.

People speak different languages, come from different backgrounds.

To tune scikit-image, you will for sure have to read the manual.
This name parameter change doesn't introduce anything new, and I feel like it will just destroy the memory of the users.

It also makes manuals from 0.17 incompatible with 0.20.

Furthermore, while yes, people should have CIs to test their code, and should update frequently, the truth is you don't want to do when you are doing something useful with the library.

The status update 

> I had to change about 400 lines of code because scikit-image 0.18 decided to rename an argument

I not fun for the person to give, and is not fun to hear.

rfezzani:
> I had to change about 400 lines of code because scikit-image 0.18 decided to rename an argument

:rofl: If you are in this situation, you probably need to refactor your code.

The question is "should `remove_small_holes` and `remove_small_objects` have consistent API?". If the answer is no, there is no issue! let's close this PR and #4003 ;)

mkcor:
> (In fact, `min_area` may be more intuitive. Oh, but `min_area` is not even an option, I see :).
> 
> @jni When we go N-D, what do we call area: volume? `min_volume` sounds weird in 2D. Hence size, but that's also not quite right, is it.

`min_hyperarea` ;)
No, I'm just kidding, because I overthought this a lot when writing https://scikit-image.org/docs/0.17.x/auto_examples/segmentation/plot_regionprops_table.html
In all seriousness, I'm not suggesting any parameter name change.

Issue #4763, opened 2020-06-02
phase_cross_correlation with NaN pixels returns an invalid result


GenevieveBuckley:
## Description
The `skimage.registration.phase_cross_correlation()` function returns an invalid result if one of the input images contains NaN pixels. I think an error should be raised here instead, suggesting the user pass in a mask array to exclude the NaNs.

Alternatively, another solution could be modifying the function to handle masking the NaN pixels itself. I'd be happy to implement this as a feature.

## Way to reproduce
```python
import numpy as np
from skimage import data
from skimage.registration import phase_cross_correlation
from skimage.util import img_as_float

reference_image = img_as_float(data.camera())
EXPECTED_SHIFT = 20  # PIXELS
moving_image = np.roll(reference_image, EXPECTED_SHIFT, axis=1)

result1 = phase_cross_correlation(reference_image, moving_image)
print("Result 1 (no NaN pixels) - ", result1)

# Add a single NaN pixel to one of the images
reference_image[0][0] = np.nan

result2 = phase_cross_correlation(reference_image, moving_image)
print("Result 2 (with NaN pixels) - ", result2)

# Mask out the NaN pixels
mask = ~np.isnan(reference_image)
result3 = phase_cross_correlation(reference_image, moving_image, reference_mask=mask)

print("Result 3 (masked NaN pixels) - ", result3)
```

Which gives us this output:
```
Result 1 (no NaN pixels):-  (array([  0., -20.]), 2.1073424255447017e-08, 8.150775661402915e-18)
Result 2 (with NaN pixels) -  (array([0., 0.]), nan, nan)
Result 3 (masked NaN pixels) -  [  0. -20.]
```

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.6.8 |Anaconda, Inc.| (default, Feb 21 2019, 18:30:04) [MSC v.1916 64 bit (AMD64)]
Windows-10-10.0.17134-SP0
scikit-image version: 0.18.dev0
numpy version: 1.16.4
```

rfezzani:
Thank you @GenevieveBuckley for your report. I am pretty sure that a majority of functions in `skimage` are sensitive to NaNs... =)
[...]

PR #4767, opened 2020-06-02
Handle NaNs in phase_cross_correlation

GenevieveBuckley:
## Description

Closes https://github.com/scikit-image/scikit-image/issues/4763

I recently found that if there is a NaN in an image passed to `phase_cross_correlation()`, you get an incorrect result back (rather than an error being raised, or the NaNs being handled by the function). 

Since it seems pretty easy to be genuinely unaware that there might be a NaN pixel somewhere in your data, it seems reasonable to add a check for this and dispatch to the masked phase correlation function if needed.

## Checklist

  existing benchmark
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/registration/_phase_cross_correlation.py
+++ b/skimage/registration/_phase_cross_correlation.py
@@ -190,6 +190,13 @@ def phase_cross_correlation(reference_image, moving_image, *,
         return _masked_phase_cross_correlation(reference_image, moving_image,
                                                reference_mask, moving_mask,
                                                overlap_ratio)
+    # If input images contain NaN pixels, must mask before correlation
+    if np.isnan(reference_image).any() or np.isnan(moving_image).any():
+        reference_mask = ~np.isnan(reference_image)
+        moving_mask = ~np.isnan(moving_image)
+        return _masked_phase_cross_correlation(reference_image, moving_image,
+                                               reference_mask, moving_mask,
+                                               overlap_ratio)
 
     # images must be the same shape
     if reference_image.shape != moving_image.shape:
--- a/skimage/registration/tests/test_phase_cross_correlation.py
+++ b/skimage/registration/tests/test_phase_cross_correlation.py
@@ -84,6 +84,16 @@ def test_3d_input():
     assert_allclose(result, -np.array(subpixel_shift), atol=0.05)
 
 
+def test_input_with_nans():
+    reference_image = img_as_float(camera())
+    moving_image = np.roll(img_as_float(camera()), 20, axis=1)
+    expected = np.array([0, -20])  # so we expect to recover a -20 pixel shift
+    # Add NaN pixels
+    reference_image[0][0] = np.nan
+    result = phase_cross_correlation(reference_image, moving_image)
+    assert np.allclose(result, expected)
+
+
 def test_unknown_space_input():
     image = np.ones((5, 5))
     with testing.raises(ValueError):


alexdesiqueira:
Thank you @GenevieveBuckley, this looks great! I think we will need this on several functions in the future...

rfezzani:
As you said @alexdesiqueira, 

> we will need this on several functions in the future...

but I am not sure that we really want that :confused:.
I reiterate my [opinion](https://github.com/scikit-image/scikit-image/issues/4763#issuecomment-637367808) here: if a user suspects that his data may contain NaNs, he has to perform the check by him self!..

skimage/registration/_phase_cross_correlation.py
```diff
@@ -190,6 +190,13 @@ def phase_cross_correlation(reference_image, moving_image, *,
         return _masked_phase_cross_correlation(reference_image, moving_image,
                                                reference_mask, moving_mask,
                                                overlap_ratio)
+    # If input images contain NaN pixels, must mask before correlation
+    if np.isnan(reference_image).any() or np.isnan(moving_image).any():
```
rfezzani, line 194: `np.isnan` is called twice  on both images: once for checking and the other for creating the mask.
May be the mask can be create before checking to avoid this repetition...

alexdesiqueira:
>  I reiterate my opinion here: if a user suspects that his data may contain NaNs, he has to perform the check by him self!..
@rfezzani you have a point! Maybe we should discuss this more thoroughly?

alexdesiqueira:
> I reiterate my opinion here: if a user suspects that his data may contain NaNs, he has to perform the check by him self!..

@rfezzani you have a point! Maybe we should discuss this more thoroughly?

GenevieveBuckley:
> 
> 
> > I reiterate my opinion here: if a user suspects that his data may contain NaNs, he has to perform the check by him self!..
> 
> @rfezzani you have a point! Maybe we should discuss this more thoroughly?

Let's put this on hold until there's a decision

sciunto:
To through an idea without thinking about it deeply. Would it be possible to make a decorator?

rfezzani:
If we decide to manage NaNs more widely all over the package, a decorator is a good option, but as @hmaarrfk pointed out
> Checking for NaN is expensive on imaegs as it often requires a full pass of the memory in python.

Using a decorator means defining a generic approach to handle NaNs... I am afraid that this will consist in simply warning the user that input data contain NaNs, an no more...

emmanuelle:
Thank you @GenevieveBuckley for the PR! The current situation where the result can be incorrect without the user knowing they had a NaN is problematic. We could also have a `check_nan` boolean kwarg to toggle the check for nans on or off, and the docstring explaining that if you use the "no check" mode it's at your own risk.

emmanuelle:
The decorator idea is nice but we want to have the possibility to pass a warning message adapted to each function.

skimage/registration/_phase_cross_correlation.py
jni: Python 3.8 syntax would be so nice for this:

```python
if (reference_mask := np.isnan(reference_image)).any() ...:
```

Just doing some advertising for the future. =)

jni:
Yep, I like @emmanuelle's idea of a `check_nan=True` or `nan_safe=True` keyword argument that we use throughout the library. @hmaarrfk what do you think of this? I have a slight bias towards `True` being default but I can also be talked into `False`, I see the argument both ways.

@rfezzani 

> What if reference_mask or moving_mask is already provided, shouldn't we check for NaNs in this case too?

Yep, this is a good point. I think the correct behaviour here would be to `&` the nan mask and the provided mask together.

hmaarrfk:
@jni, the default doesn't matter to me so much. The addition of a keyword argument is fine by me.

emmanuelle:
@GenevieveBuckley could you please benchmark the execution time before and after your change for a typical image (let's say 1000^2 pixels)? If the difference is negligible maybe we don't even need a flag.

GenevieveBuckley:
I'd have to get back to you on that, I don't have asv set up right now.

jni:
@GenevieveBuckley I don't think @emmanuelle is suggesting a formal benchmark, just a quick test. The idea is, yes, `np.isnan` is not free, but *in this specific case*, since we are doing an FFT and maybe even upscaling, it might be effectively free. Anything under 5% is a shoo-in for a quick merge!

emmanuelle:
yes yes asv would be overkill here, please just write an example script which you can run on your branch and on master to compare.

GenevieveBuckley:
Ah, thanks for clarifying!

GenevieveBuckley:
The increase can be as high as 20%

Here are some results using `%timeit` in a jupyter notebook:
```

master branch  512x512 greyscale - 291 ms ± 99.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 512x512 greyscale - 298 ms ± 27.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.0240549828178693

master branch  512x512 RGB image - 1.05 s ± 72.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 512x512 RGB image - 1.29 s ± 198 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.2285714285714286

master branch  1000x1000 greyscale - 787 ms ± 85.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 1000x1000 greyscale - 944 ms ± 335 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.1994917407878019

master branch  1000x1000 RGB image - 4.47 s ± 571 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 1000x1000 RGB image - 5.12 s ± 472 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.1454138702460852

master branch  1024x1024 greyscale - 1.06 s ± 166 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 1024x1024 greyscale - 1.22 s ± 261 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.150943396226415

master branch  1024x1024 RGB image - 5.15 s ± 359 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 1024x1024 RGB image - 5.36 s ± 865 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.0407766990291263

master branch  2000x2000 greyscale - 4 s ± 492 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 2000x2000 greyscale - 4.89 s ± 915 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.2225

master branch  2000x2000 RGB image - 17.6 s ± 2.49 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 2000x2000 RGB image - 18.1 s ± 2.19 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.0284090909090908

master branch  2048x2048 greyscale - 3.64 s ± 484 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 2048x2048 greyscale - 4.62 s ± 657 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 1.2692307692307692

master branch  2048x2048 RGB image - 24.5 s ± 3.39 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature branch 2048x2048 RGB image - 20.9 s ± 1.5 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
feature-time / master-time = 0.8530612244897958

```

stefanv:
> Yep, I like @emmanuelle's idea of a `check_nan=True` or `nan_safe=True` keyword argument that we use throughout the library. @hmaarrfk what do you think of this? I have a slight bias towards `True` being default but I can also be talked into `False`, I see the argument both ways.
> 
> @rfezzani
> 
> > What if reference_mask or moving_mask is already provided, shouldn't we check for NaNs in this case too?
> 
> Yep, this is a good point. I think the correct behaviour here would be to `&` the nan mask and the provided mask together.

I wonder if this is worth it.  It's a large new API surface area, for something that the user should probably be trained to recognize or check for themselves.

jni:
@GenevieveBuckley Wowza! 20% is way above what I expected to see here! Can you share your notebook? I kinda find the results hard to believe given that `np.isnan` is about 2% of the time of `np.fft.fftn`:

```python
In [1]: image = np.random.random((2048, 2048))                                         

In [2]: %timeit np.fft.fftn(image)                                                     
153 ms ± 1.69 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit np.any(np.isnan(image))                                                
2.87 ms ± 15.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

@stefanv I get what you're saying in general for the library, and I'm in favour of that attitude e.g. when nans in the image result in a nan result (e.g. thresholding). But when they give an incorrect result, I think we have to do better than that... 

I guess an alternative to the approach here would be to ensure that the shift is `np.nan`, or raise, when the *error* is `np.nan` **at the end** of the function:

```python
In [1]: image = np.random.random((2048, 2048))                                         
In [4]: image[0, 0] = np.nan                                                           
In [10]: image2 = ndi.shift(image, (29, 10), order=0)                                  
In [12]: from skimage.registration import phase_cross_correlation                      
In [13]: shift = phase_cross_correlation(image, image2)                                
In [17]: shift, error, phasediff = phase_cross_correlation(image, image2)              
/home/jni/projects/scikit-image/skimage/registration/_phase_cross_correlation.py:104: RuntimeWarning: invalid value encountered in cdouble_scalars
  error = 1.0 - cross_correlation_max * cross_correlation_max.conj() /\
In [18]: shift, error, phasediff                                                       
Out[18]: (array([0., 0.]), nan, nan)
```

Since we would be checking a single number, that would be essentially free, and would align philosophically with @stefanv's preferences. The error message should then inform the user about how to use e.g. `reference_mask=np.isnan(reference_image)` to avoid these results.

jni:
Having said that...

```python
In [19]: phase_cross_correlation(image, image2, reference_mask=np.isnan(image), moving_
    ...: mask=np.isnan(image2))                                                        
Out[19]: array([0., 0.])
```

... wth?

GenevieveBuckley:
@jni 

1. This notebook shows how I constructed the data for benchmarking with %timeit - https://gist.github.com/GenevieveBuckley/977070acc2d169aa356073e103a864cd


2. Just guessing, but I think you might be masking the wrong way. 

`np.isnan` will return an array that is `True` where pixels are NaN **but** for `phase_cross_correlation` you want a mask array that is `False` for all NaN pixels.

> Having said that...
> 
> ```python
> In [19]: phase_cross_correlation(image, image2, reference_mask=np.isnan(image), moving_
>     ...: mask=np.isnan(image2))                                                        
> Out[19]: array([0., 0.])
> ```
> 
> ... wth?

jni:
> Just guessing, but I think you might be masking the wrong way.

🤦 you are absolutely correct! 😂

emmanuelle:
> I guess an alternative to the approach here would be to ensure that the shift is `np.nan`, or raise, when the _error_ is `np.nan` **at the end** of the function:

@GenevieveBuckley do you think this could work? If yes this would be an elegant solution, since we'd have the best of both worlds.

GenevieveBuckley:
> 
> 
> > I guess an alternative to the approach here would be to ensure that the shift is `np.nan`, or raise, when the _error_ is `np.nan` **at the end** of the function:
> 
> @GenevieveBuckley do you think this could work? If yes this would be an elegant solution, since we'd have the best of both worlds.

Yes, I think checking whether the error is `np.nan` at the end of the function and then raising an informative error to tell the user about how to use masked phase cross correlation would work well.

emmanuelle:
Thanks @GenevieveBuckley. Do you have time making this change?

GenevieveBuckley:
> Thanks @GenevieveBuckley. Do you have time making this change?

I could, but it seems that we can only do this quick check if the input argument `return_error=True` (otherwise I think we're back to checking a whole array for nan values, rather than a single number). The `return_error` keyword argument is True by default, fwiw. Is this acceptable, or is a different approach preferred?

GenevieveBuckley:
Closing this in favour of https://github.com/scikit-image/scikit-image/pull/4886 (which has the approach suggested by jni and emanuelle)

PR #4788, opened 2020-06-12
try to fix ci which is broken because of pyqt5 last version

emmanuelle:
The CI is broken on Travis, probably following the latest release of pyqt5. ~~I noticed that I could only pip install this pyqt5 release using a recent pip so this PR is an attempt to fix the CI.~~ (edit:didn't work, I just excluded the last version)


emmanuelle:
I don't really understand why the failing test is failing (seems to be a multithreading issue, can't reproduce locally) so I just excluded the last version

hmaarrfk:
the .0 release will almost surely have hiccups with pyqt ;)

jni:
Thanks @emmanuelle!

hmaarrfk:
thanks a bunch!

PR #4792, opened 2020-06-15
Remove unused variable in j invariant docs

### LINKED ISSUE #4779: sigma_range not used in denoiser calibration examples
The variable `sigma_range` is defined in both tutorials, but a different range and in-line instantiation of the sigma range is used when the `parameter_ranges` dict is defined:

https://github.com/scikit-image/scikit-image/blob/f71be82423e73cda4f3026a0eb656614db937bbc/doc/examples/filters/plot_j_invariant.py#L39-L41

https://github.com/scikit-image/scikit-image/blob/f71be82423e73cda4f3026a0eb656614db937bbc/doc/examples/filters/plot_j_invariant_tutorial.py#L44-L46 

In the latter "full" tutorial, this seems to be limited to just the wavelet denoiser calibration portion, which appears to be a copy/paste of the former "short" tutorial.

notmatthancock:
Closes #4779

### DIFF
--- a/doc/examples/filters/plot_j_invariant.py
+++ b/doc/examples/filters/plot_j_invariant.py
@@ -36,8 +36,6 @@ sigma = 0.3
 noisy = random_noise(image, var=sigma ** 2)
 
 # Parameters to test when calibrating the denoising algorithm
-sigma_range = np.arange(0.05, 0.5, 0.05)
-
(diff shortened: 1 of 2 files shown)


jni:
Thank you @notmatthancock!

emmanuelle:
Thank you @notmatthancock :-)

PR #4793, opened 2020-06-16
include all md files in manifest.in

sciunto:
## Description
I'm trying to rationalize this file.

The CODE_OF_CONDUCT.md must be included imo.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,6 @@
 include setup*.py
 include MANIFEST.in
-include README.md
-include LICENSE.txt
+include *.md
 include *.txt
 include *.rst
 recursive-include requirements *.txt


sciunto:
It is used for the sdist (at least, afaik).

emmanuelle:
an easy PR to approve and merge @scikit-image/core

alexdesiqueira:
> The CODE_OF_CONDUCT.md must be included imo.
+1 @sciunto. Thank you!

stefanv:
What is the point of having this in the source distribution? README is in there, because that is where pypi description came from. LICENSE for obvious reasons. The others? This is not where users or developers will find them. It's not a big deal either way, including a few more won't hurt.

alexdesiqueira:
I would say that it's good having this kind of info easily available "offline" — but hey, what's offline anyway :slightly_smiling_face:

sciunto:
yes, get all relevant files versioned outside git.

PR #4798, opened 2020-06-23
Remove additional "::" to make plot directive work.

### LINKED ISSUE #4540: Broken plot in userguide

dsandeep0138:
Issue here is that, after "::", there should be some code/text, but plot directive is immediately started leading to the plot directive not working. Add some code/text to make the plot code work, or simply remove "::" which doesn't have any code after.

Fixes #4540


jni:
Brilliant! Thank you @dsandeep0138!

PR #4803, opened 2020-06-26
Fix runtime warning in blob.py

### LINKED ISSUE #4729: RuntimeWarning in blob.py (division by zero)
see for example the warning in https://scikit-image.org/docs/dev/auto_examples/features_detection/plot_blob.html#sphx-glr-auto-examples-features-detection-plot-blob-py 

The warning is initiated in the `_blob_overlap` function.

Also mentioned in the forum https://forum.image.sc/t/runtime-warning-with-blob-detection-when-threshold-is-too-low/37965

dsandeep0138:
Runtime warning (division by zero) is observed during blob_overlap when the sigma of both the blobs is zero (i.e., radius of the blobs is 0. This means that the both the blobs are part of a bigger one). Therefore, the overlap can return 0 in this case.

Fixes #4729

### DIFF
--- a/skimage/feature/blob.py
+++ b/skimage/feature/blob.py
@@ -115,7 +115,9 @@ def _blob_overlap(blob1, blob2, *, sigma_dim=1):
 
     # we divide coordinates by sigma * sqrt(ndim) to rescale space to isotropy,
     # giving spheres of radius = 1 or < 1.
-    if blob1[-1] > blob2[-1]:
+    if blob1[-1] == 0 and blob2[-1] == 0:
+        return 0.0
+    elif blob1[-1] > blob2[-1]:
         max_sigma = blob1[-sigma_dim:]
         r1 = 1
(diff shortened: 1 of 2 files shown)


skimage/feature/blob.py
```diff
@@ -115,7 +115,9 @@ def _blob_overlap(blob1, blob2, *, sigma_dim=1):
 
     # we divide coordinates by sigma * sqrt(ndim) to rescale space to isotropy,
     # giving spheres of radius = 1 or < 1.
-    if blob1[-1] > blob2[-1]:
+    if blob1[-1] == 0 and blob2[-1] == 0:
```
rfezzani, line 118:
```suggestion
    if blob1[-1] == blob2[-1] == 0:
```

emmanuelle:
Thank you very much @dsandeep0138 ! merging.

Issue #4797, opened 2020-06-22
API: replace array by image in measure.find_contours()


sciunto:
## Description

See https://scikit-image.org/docs/dev/api/skimage.measure.html?highlight=find_con#skimage.measure.find_contours

The signature must be `skimage.measure.find_contours(image, level, fully_connected='low', positive_orientation='low', *, mask=None)` to comply with our policy.

PR #4806, opened 2020-06-28
API change: Replace array by image in measure.find_contour

jjerphan:
Quick fix for #4797

### DIFF
--- a/TODO.txt
+++ b/TODO.txt
@@ -70,6 +70,8 @@ Version 0.19
 Version 0.20
 ------------
 
+* In ``skimage.measure._find_contours.py``, remove the `deprecate_kwarg`
+  decorator from `find_contours`.
 
 Version 1.0
 -----------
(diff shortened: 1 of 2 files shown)


jni:
Thanks for the proposed fix @jjerphan! This is great but it requires two extra things:

- support for deprecated use of `array=` as a keyword argument to the function, so that we don't break our existing API without warning
- a TODO item under 0.20 for removing the deprecated argument.

These are actually pretty easy to achieve thanks to the [`skimage._shared.utils.deprecate_kwarg`](https://github.com/scikit-image/scikit-image/blob/477dd75aa864453bfbb11e201bd49a4b8d085dc5/skimage/_shared/utils.py#L71-L115) decorator. You can see an example use here:

https://github.com/scikit-image/scikit-image/pull/4761/files#diff-bfeb06dd4f95632ee7f7e3ae971d0065R5-R53

And you can see #4761 as an example pull request doing an API deprecation such as this one.

If you make those changes I think this will be ready to merge!

PR #4809, opened 2020-06-29
Let biomed example load image data with Pooch.

mkcor:
## Description

In this PR, we are making use of a recent (v0.17.1) feature, namely on-the-fly download of data with `pooch`.
It follows from the fact that we are [now](https://gitlab.com/scikit-image/data/-/merge_requests/2) hosting a copy of the CC0-licensed image at: https://gitlab.com/scikit-image/data

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_human_mitosis.py
+++ b/doc/examples/applications/plot_human_mitosis.py
@@ -24,9 +24,11 @@ from scipy import ndimage as ndi
 from skimage import (
     color, feature, filters, io, measure, morphology, segmentation, util
 )
+from skimage.data import image_fetcher
 
 
-image = io.imread('https://github.com/CellProfiler/examples/blob/master/ExampleHuman/images/AS_09125_050116030001_D03f00d0.tif?raw=true')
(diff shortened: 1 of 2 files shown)


jni:
> We have a gitlab page as well?

Yes, that's a new thing based on the fact that gitlab is more generous with bandwidth than github. Might be premature optimisation but we were seeing so many sporadic CI failures with pooch downloads that I wondered whether we were hitting bandwidth limits. I do also like diversifying our tools base...

jni:
Note that there is nothing compulsory about using gitlab vs another github.com/scikit-image/data repo vs really any other page on the web. Pooch only cares about stable URLs, which gitlab should be as good as github at providing.

emmanuelle:
Thank you @mkcor ! Marianne was telling me that we lack documentation about using pooch (both for devs and users), which is true. For a future PR :-)... Merging this one now.

hmaarrfk:
My goal is to help alleviate the installation guide first, then move on to the pooch guide.

I think we are doing good strides in the install guide

Issue #4466, opened 2020-02-19
Feature suggestion: passing user-defined measurement functions to regionprops


VolkerH:
I have encountered the following pattern a number of times now:

1. I segment objects and calculate regionprops and export them to a table. `regionprops_table`) already facilitates this workflow. This is often done because users at our microscopy facility ask me to write a script that calculates these features from images.
2. The users sometimes want to measure some additional property that  is not implemented in scikit-image, say for example minimum Feret diameter. What I end up doing is writing a function that calculated this property and then implementing a loop over all regions to calculate the desired feature and then somehow concatenate the results from `regionprops` (or `regionprops_table`) with the results from my own loop over the regions.  

The desired behaviour would be to only have `regionprops{_table}`loop once over all regions and have a single output table returned. 
I envisage being able to pass an optional argument to `regionprops{_table}` such as
```
additional_measurements =( (prop1_func, prop1_name, prop1_return_dtype),   (prop2_func, prop2_name, prop2_return_dtype), ... )
```

Where `prop*_func` is a function that calculates the property given a region, `prop*_name` is the name of the property and `prop*_return_dtype` specifies what one would expect the function to return as output.

Other than just looping over the regions once, this would allow 3rd-party measurement functions that may be too specific (or not of high enough code quality, or only working for a specific number of dimensions) to include in the skimage code base to be distributed separately.

Does this make sense?

jni:
I think this is a great idea! I don't love the proposed API, but I definitely think we can come up with an API I'm happy with. =)

rfezzani:
Yes, that's cool ;-). I would prefer the optional argument to be of the form `extra={'name1': func1, ...}` without caring about return dtype but it looks like a great suggestion :+1:.

VolkerH:
The API can be different, this was just a quick sketch of one option, I care more about having the functionality. Another option would be to have a function such as

```
register_measurement(func, name, ...)
```
and after registering one or many additional measurements like this they can just be used using the existing `regionprops` or `regionprops_table` functions.

jni:
@VolkerH re: registering, I really do like the idea of arbitrary regionprops plugins. We could reuse the strategy from napari/napari#908, or more lightweight, something like our io plugins.

@rfezzani as currently implemented, the dtype is important because it specifies whether the property in the table is stored as an object or as one or more numeric types. e.g. the property `.image` has a different size for each region, so you need to use an object array for that column. etc.

VolkerH:
[...]

PR #4810, opened 2020-06-30
allow passing extra measurement functions to regionprops and regionprops_table

VolkerH:
## Description

This is a WIP implements the feature request discussed here: https://github.com/scikit-image/scikit-image/issues/4466 

It extends `regionprops` and `regionprops_table` by adding an extra argument, which is an iterable of callables/functions.

A user might want to do this if a measurment they require is not included in the built-in properties. By specifying these functions users won't have to duplicate the functionality that loops over the regions. Possible use cases: 

* provide a measurement function that is too specifically focused to include in the skimage code base
* provide measurement function that is under an incompatible license
* provide a faster implementation for an existing measurement function

Each provided function must either take one (`region`) or two (`region`, `intensity_image`) (names for illustration, this is by order) arguments. 

In the issue, there was some discussion on whether the name and return dtype of the measurement function need to be passed in addition, as these are needed to build the regionprops table. Both the name of new property and the dtype are now inferred.
The name is inferred simply as the function name, and a warning is issued if there is a name clash. The return dtype is determined by passing a tiny test image to the function and inspecting the return value.

At this stage, the functionality is working but tests and examples still need to be added.

Code contributions:
Much of code in this PR was co-authored by @jni during a VS Code LiveShare session (somehow VS code only appended the co-authorship note to one of the commit messages).

## Checklist

- DONE  [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- TODO Gallery example in `./doc/examples` (new features only)
- SKIP Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
- TODO Unit tests
- DONE Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

## For reviewers

  later.


jni:
@scikit-image/core before we keep moving on this, it would be great to get some consensus on the API. Currently, it looks like this:

```python
rp = regionprops(
    label_image,
    intensity_image,
    extra_properties=(mesh_surface, porosity, average_surface_curvature),
)
```

Those properties can then be accessed with `rp[0].mesh_surface` etc. I think it's pretty nice!

stefanv:
> @scikit-image/core before we keep moving on this, it would be great to get some consensus on the API. Currently, it looks like this:
> 
> ```python
> rp = regionprops(
>     label_image,
>     intensity_image,
>     extra_properties=(mesh_surface, porosity, average_surface_curvature),
> )
> ```
> 
> Those properties can then be accessed with `rp[0].mesh_surface` etc. I think it's pretty nice!

Shouldn't that be a dict instead of a tuple?

jni:
@stefanv 

> Shouldn't that be a dict instead of a tuple?

dict mapping what to what? It's a sequence of functions. We *can* make a dict mapping name -> function, but Python makes that unnecessary/redundant since we can use `func.__name__`.

stefanv:
I guess I was thinking about lambdas, but then it's easy enough to just define a new function.

jni:
> I guess I was thinking about lambdas

Yes, I would leave this concern for future PRs. We could also accept *both* a sequence and a dict, but I fear that complicates things.

VolkerH:
added

* tests
* docstring examples

also 
* managed to shut up pep8speaks (for the moment)
* corrected one serious bug that only showed up when testing with an intensity image and at least two different labels. As the SAMPLE arrays in the tests only have a single label that was not picked up. Added a SAMPLE_MULTIPLE array in the tests.

I'm not sure whether a gallery example makes sense as there is no image to show. 

@jni, do you want to have another look before I remove the WIP and flag this as ready for review?

VolkerH:
Anyway, I think this is ready for review.

skimage/measure/_regionprops.py
```diff
@@ -166,6 +228,45 @@ def __init__(self, slice, label, label_image, intensity_image,
...
+                    raise AttributeError(
+                        f"intensity image required to calculate {attr}"
+                    )
+            elif n_args == 1:
+                return func(self.image)
+            else:
+                raise AttributeError(f"Function provided for custom regionprop"
+                                     " {attr} takes incorrect number of"
+                                     " arguments."
+                                     )
```
jni, line 264:
```suggestion
                raise AttributeError(
                    f"Function provided for custom regionprop {attr} takes {n_args} "
                    "arguments, but only 1 or 2 are supported."
                )
```

jni: (The f-ness of a string does not propagate across multiple lines, fyi.)

skimage/measure/_regionprops.py
```diff
@@ -638,33 +755,62 @@ def regionprops_table(label_image, intensity_image=None,
 
     [5 rows x 7 columns]
 
+    If we want to measure a feature that does not come as a built-in
+    property, we can define custom functions and pass them as
+    ``extra_properies``. Let's create a custom function that measures
+    the inensity quartiles in a region:
```
jni, line 761:
```suggestion
    ``extra_properies``. For example, we can create a custom function
    that measures the intensity quartiles in a region:
```

skimage/measure/_regionprops.py
```diff
@@ -638,33 +755,62 @@ def regionprops_table(label_image, intensity_image=None,
...
+
+    >>> from skimage import data, util, measure
+    >>> import numpy as np
+    >>> def quartiles(regionmask, intensity):
+    >>>     return np.percentile(intensity[regionmask], q=(25, 50, 75))
```
jni, line 766:
```suggestion
    >>> def quartiles(regionmask, intensity):
    ...     return np.percentile(intensity[regionmask], q=(25, 50, 75))
```

skimage/measure/_regionprops.py
```diff
@@ -638,33 +755,62 @@ def regionprops_table(label_image, intensity_image=None,
...
+    >>>
+    >>> image = data.coins()
+    >>> label_image = measure.label(image > 110, connectivity=image.ndim)
+    >>> props = measure.regionprops_table(label_image, intensity_image=image,
+    >>>                                   properties=('label',),
+    >>>                                   extra_properties=(quartiles,))
```
jni, line 772:
```suggestion
    >>> props = measure.regionprops_table(label_image, intensity_image=image,
    ...                                   properties=('label',),
    ...                                   extra_properties=(quartiles,))
```

skimage/measure/_regionprops.py
```diff
@@ -638,33 +755,62 @@ def regionprops_table(label_image, intensity_image=None,
...
+    >>> import pandas as pd # doctest: +SKIP
+    >>> pd.DataFrame(props).head() # doctest: +SKIP
+       label  quartiles-0  quartiles-1  quartiles-2
+0      1       117.00        123.0        130.0
+1      2       111.25        112.0        114.0
+2      3       111.00        111.0        111.0
+3      4       111.00        111.5        112.5
+4      5       112.50        113.0        114.0
```
jni, line 780:
```suggestion
           label  quartiles-0  quartiles-1  quartiles-2
    0      1       117.00        123.0        130.0
    1      2       111.25        112.0        114.0
    2      3       111.00        111.0        111.0
    3      4       111.00        111.5        112.5
    4      5       112.50        113.0        114.0
```

skimage/measure/_regionprops.py
```diff
@@ -872,6 +1024,20 @@ def regionprops(label_image, intensity_image=None, cache=True,
...
     (22.72987986048314, 81.91228523446583)
 
+    Add custom measurements by passing functions as ``extra_properties``
+    >>> from skimage import data, util
+    >>> from skimage.measure import label, regionprops
+    >>> import numpy as np
+    >>> img = util.img_as_ubyte(data.coins()) > 110
+    >>> label_img = label(img, connectivity=img.ndim)
+    >>> def pixelcount(regionmask):
+    >>>     return np.sum(regionmask)
```
jni, line 1034:
```suggestion
    >>> def pixelcount(regionmask):
    ...     return np.sum(regionmask)
```

skimage/measure/_regionprops.py
```diff
@@ -872,6 +1024,20 @@ def regionprops(label_image, intensity_image=None, cache=True,
     >>> props[0]['centroid']
     (22.72987986048314, 81.91228523446583)
 
+    Add custom measurements by passing functions as ``extra_properties``
```
jni, line 1027:
```suggestion
    Add custom measurements by passing functions as ``extra_properties``:

```

skimage/measure/_regionprops.py
VolkerH: I did not know that ! Interesting.

VolkerH:
Thanks for those changes.  Writing code in comments presents different challenges from just writing code.

jni:
Everything is passing! @scikit-image/core this is ready for review!

skimage/measure/_regionprops.py
```diff
@@ -166,6 +228,45 @@ def __init__(self, slice, label, label_image, intensity_image,
...
+                else:
+                    raise AttributeError(
+                        f"intensity image required to calculate {attr}"
+                    )
+            elif n_args == 1:
+                return func(self.image)
+            else:
+                raise AttributeError(
+                    f"Function provided for custom regionprop {attr} takes {n_args} "
+                    "arguments, but only 1 or 2 are supported."
```
rfezzani, line 263:
```suggestion
                    "Custom regionprop function's number of arguments must be 1 or 2"
                    f"but {attr} takes {n_args} arguments."
```

skimage/measure/_regionprops.py
```diff
@@ -693,7 +839,13 @@ def regionprops(label_image, intensity_image=None, cache=True,
...
             will be less trivial. For example, the new orientation is
             :math:`\frac{\pi}{2}` plus the old orientation.
-
+    extra_properties : Iterable of callables
+        Add extra property computation functions that are not included with
+        skimage. The name of the property is derived from the function name,
+        the dtype is inferred by calling the function on a small sample.
+        If the name of an extra property clashes with the name of an existing
+        property the extra property wil not be visible and a UserWarning is
+        issued.
```
rfezzani, line 848: Shouldn't we specify that the first argument of the property function must be the mask of the considered region and the optional second argument must be the corresponding image?

skimage/measure/_regionprops.py
```diff
@@ -166,6 +228,45 @@ def __init__(self, slice, label, label_image, intensity_image,
...
+        if extra_properties is None:
+            extra_properties = []
+        for func in extra_properties:
+            name = func.__name__
+            if hasattr(self, name):
+                msg = (
+                    f"Extra property '{name}' is shadowed by existing "
+                    "property and will be inaccessible. Consider renaming it."
+                )
+                warn(msg)
```
rfezzani, line 241: Can't we dynamically edit `COL_DTYPES` to avoid the handling of  dtype for extra properties in `_props_to_dict`?
```suggestion
                warn(msg)
            else:
                COL_DTYPES[func.__name__] = _infer_regionprop_dtype(
                    func,
                    intensity=intensity_image is not None,
                    ndim=self._ndim,
                )
```

skimage/measure/_regionprops.py
```diff
@@ -579,6 +689,13 @@ def regionprops_table(label_image, intensity_image=None,
         Object columns are those that cannot be split in this way because the
         number of columns would change depending on the object. For example,
         ``image`` and ``coords``.
+    extra_properties : Iterable of callables
+        Add extra property computation functions that are not included with
+        skimage. The name of the property is derived from the function name,
+        the dtype is inferred by calling the function on a small sample.
+        If the name of an extra property clashes with the name of an existing
+        property the extra property wil not be visible and a UserWarning is
+        issued.
```
rfezzani, line 698: Same [question](https://github.com/scikit-image/scikit-image/pull/4810/files#r449624730)?

skimage/measure/_regionprops.py
jni: @rfezzani  here we chose to preserve existing behaviour rather than try to improve it, which could have unintended consequences. I suggest leaving dynamic inference of existing properties for a future PR.

skimage/measure/_regionprops.py
VolkerH: yes, we should. So far that could only be derived from the examples. I just added this, happy for suggestions about how to phrase it better.

skimage/measure/_regionprops.py
VolkerH: @jni was faster than me. In general, I don't like modifying a global (in the namespace) state. That state change will persist in `COL_DTYPES` potentially leading to the "unintended consequences" that @jni mentioned.

VolkerH:
Thanks very much for the review and the suggestions @rfezzani, I made the suggested changes apart from the dynamic addition to `COL_DTYPES` (see comments there).

skimage/measure/_regionprops.py
rfezzani: The suggestion is not an improvement to existing behavior but the modification of the extra properties (new feature) dtype handeling ^^. Avoiding namespace variable modification is a good point, no problem ;)

skimage/measure/_regionprops.py
```diff
@@ -872,6 +1028,20 @@ def regionprops(label_image, intensity_image=None, cache=True,
...
     (22.72987986048314, 81.91228523446583)
 
+    Add custom measurements by passing functions as ``extra_properties``
+    >>> from skimage import data, util
+    >>> from skimage.measure import label, regionprops
+    >>> import numpy as np
+    >>> img = util.img_as_ubyte(data.coins()) > 110
+    >>> label_img = label(img, connectivity=img.ndim)
+    >>> def pixelcount(regionmask):
+    >>>     return np.sum(regionmask)
```
rfezzani, line 1038: This is causing the CI failure
```suggestion
            return np.sum(regionmask)
```

VolkerH: Thanks a lot for finding this. I didn't look at the failing Travis build log as it has been failing with matplotlib for a while and assumed that was still the case. Will commit your suggestion.
I'm curious: are the other CI builds not building documentation or why are they all give the green tick?

rfezzani: >  I'm curious: are the other CI builds not building documentation or why are they all give the green tick?

They are, and they crashed now :rofl:

skimage/measure/_regionprops.py
```diff
@@ -872,6 +1028,20 @@ def regionprops(label_image, intensity_image=None, cache=True,
...
     (22.72987986048314, 81.91228523446583)
 
+    Add custom measurements by passing functions as ``extra_properties``
+    >>> from skimage import data, util
+    >>> from skimage.measure import label, regionprops
+    >>> import numpy as np
+    >>> img = util.img_as_ubyte(data.coins()) > 110
+    >>> label_img = label(img, connectivity=img.ndim)
+    >>> def pixelcount(regionmask):
+            return np.sum(regionmask)
```
rfezzani, line 1038:
```suggestion
...         return np.sum(regionmask)
```
Now it should be OK ;)

skimage/measure/_regionprops.py
```diff
@@ -872,6 +1028,20 @@ def regionprops(label_image, intensity_image=None, cache=True,
...
     (22.72987986048314, 81.91228523446583)
 
+    Add custom measurements by passing functions as ``extra_properties``
+    >>> from skimage import data, util
+    >>> from skimage.measure import label, regionprops
+    >>> import numpy as np
+    >>> img = util.img_as_ubyte(data.coins()) > 110
+    >>> label_img = label(img, connectivity=img.ndim)
+    >>> def pixelcount(regionmask):
+...         return np.sum(regionmask)
```
rfezzani, line 1038:
```suggestion
    ...     return np.sum(regionmask)
```

rfezzani:
:tada: The CI is green again (almost :p). Merging!
Thank you again @VolkerH ^^

VolkerH:
Thank you for fixing the Sphinx build and the review @rfezzani . Credit should also go to @jni !

refack:
Hello @VolkerH,

Firstly I would like to say that I love this feature, thank you for this 🙏.
Secondly I'd like to ask if there's a reason why you chose not to pass the `RegionProperties` object to the functions? (a.k.a. `self` in the context of their evaluation)
https://github.com/scikit-image/scikit-image/blob/3a4dc584d062347d9ab6bcd0910aeb7b40bd9623/skimage/measure/_regionprops.py#L253


Just for background, for my use case I need to know the `label` of the region in order to cross corelate it.

VolkerH:
Hi @refack ,

glad you like it, appreciate the feedback that it is useful.

As for your second question: it simply did not occur to me at the time and I wanted a simple function signature for the `extra_properties` function. Someone who implements such a function, should not need to know about any other input other than the mask and potentially the intensity image for the region.

If your extra property depends on other properties calculated for the region, I am not sure whether passing `self` will be so good as I don't think it can be guarantueed that those other properties will already have been calculated. Also, passing the object might allow an incorrectly implemented `extra_properties` function to modify the object (and thus other properties). 

If your desired result depends on several properties, is it possible to calculate if from several columns of the `regionsprops_table`?

I have not looked at the regionprops code since last July (and that is the only time I looked at it) so I might be missing something. It may also be possible to add what you want without breaking things. But @jni (who deserves at least equal credit for implementing this feature) will be better positioned to answer that.

jni:
> it simply did not occur to me at the time

Me neither! But just today I found that I could have used it!

> I don't think it can be guaranteed that those other properties will already have been calculated.

Actually this is not a concern — all the properties are dynamically calculated. It would be a concern if you wanted to have chains of custom functions that depended on each other. Then you would have to make sure that you provided them in the order they are needed.

@refack I think this is certainly worth considering. Would you mind creating an issue specifically for this (e.g. "Proposal: optionally pass Regionprops object to custom properties") so that we can discuss the best way to do this? imho the argument should not be called "self" as that is a reserved-by-convention term in Python. Perhaps simply `props`?

PR #4812, opened 2020-07-01
Tweak threshold_otsu error checking - closes #4811

### LINKED ISSUE #4811: threshold_otsu check for single-colour is slow on large arrays
Currently in the threshold_otsu function, the code checks if the image has a single colour

    if image.min() == image.max():
        raise ValueError("threshold_otsu is expected to work with images "
                         "having more than one color. The input image seems "
                         "to have just one color {0}.".format(image.min()))`

However, for extremely large arrays min and max become increasingly slow, performing a basically identical sorting algorithm twice over.

I propose replacing this with a much simpler logical check, which will inevitably be more efficient

    np.all(image == image[0])

Checking with timeit:

    import timeit
    setup = '''
    import numpy as np
    
    arr = np.zeros((10000,10000))
    '''

    timeit.timeit("arr.min() == arr.max()", setup=setup, number=50)
    5.534767805999763

    timeit.timeit("np.all(arr == arr[0])", setup=setup, number=50)
    2.377352842999244

As expected, this time disparity gets greater and greater with increasing size.

Further, I disagree with the error statement. It seems very... unconfident in the code! "We think it's maybe possibly not the right thing, so let's throw an error, whatevs". If we genuinely do not have faith in what that code is doing, it should not be public code. It should be more assertive. I propose "threshold_otsu cannot threshold single-valued images", or better yet "cannot threshold single-valued images", as the former implies that other methods can threshold single-valued images - which is of course nonsense!

OGordon100:
Closes #4811

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -251,11 +251,6 @@ def threshold_otsu(image, nbins=256):
         Upper threshold value. All pixels with an intensity higher than
         this value are assumed to be foreground.
 
-    Raises
-    ------
-    ValueError
-         If ``image`` only contains a single grayscale value.
-
     References
     ----------
     .. [1] Wikipedia, https://en.wikipedia.org/wiki/Otsu's_Method
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -277,8 +277,9 @@ def threshold_otsu(image, nbins=256):
         warn(msg.format(image.shape))
 
     # Check if the image is multi-colored or not
-    if np.all(image == image.ravel()[0]):
-        raise ValueError("Cannot threshold single-valued images")
+    first_pixel = image.ravel()[0]
+    if np.all(image == first_pixel):
+        return first_pixel
 
     hist, bin_centers = histogram(image.ravel(), nbins, source_range='image')
     hist = hist.astype(float)
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -271,13 +271,13 @@ def threshold_otsu(image, nbins=256):
     -----
     The input image must be grayscale.
     """
-    if len(image.shape) > 2 and image.shape[-1] in (3, 4):
+    if image.ndim > 2 and image.shape[-1] in (3, 4):
         msg = "threshold_otsu is expected to work correctly only for " \
               "grayscale images; image shape {0} looks like an RGB image"
         warn(msg.format(image.shape))
 
     # Check if the image is multi-colored or not
-    if not np.all(image == image[0]):
+    if np.all(image == image[0]):
         raise ValueError("threshold_otsu cannot threshold single-valued images")
 
     hist, bin_centers = histogram(image.ravel(), nbins, source_range='image')
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -277,7 +277,7 @@ def threshold_otsu(image, nbins=256):
         warn(msg.format(image.shape))
 
     # Check if the image is multi-colored or not
-    if np.all(image == image[0]):
+    if np.all(image == image.ravel()[0]):
         raise ValueError("Cannot threshold single-valued images")
 
     hist, bin_centers = histogram(image.ravel(), nbins, source_range='image')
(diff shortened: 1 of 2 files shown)


skimage/filters/thresholding.py
```diff
@@ -271,16 +271,14 @@ def threshold_otsu(image, nbins=256):
...
         msg = "threshold_otsu is expected to work correctly only for " \
               "grayscale images; image shape {0} looks like an RGB image"
         warn(msg.format(image.shape))
 
     # Check if the image is multi-colored or not
-    if image.min() == image.max():
-        raise ValueError("threshold_otsu is expected to work with images "
-                         "having more than one color. The input image seems "
-                         "to have just one color {0}.".format(image.min()))
+    if np.all(image == image[0]):
```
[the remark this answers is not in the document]
jni: I prefer: 

```suggestion
    if np.all(image == image.ravel()[0]):
```

But it's a matter of taste. =)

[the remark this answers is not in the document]
rfezzani: You are right @jni, I used to prefer `ravel`, but according to the _Notes_ section in its [doc](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html)
> When a view is desired in as many cases as possible, arr.reshape(-1) may be preferable.

Otherwise `ravel` is super fine ^^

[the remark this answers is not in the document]
jni: @rfezzani 

> When a view is desired in as many cases as possible, arr.reshape(-1) may be preferable.

You already taught me this once before in the `array_map` PR and I've already forgotten it! 😅 Old habits die hard... Thanks!

emmanuelle:
See also the discussion about performance in #4811.

skimage/filters/thresholding.py
```diff
@@ -271,16 +271,15 @@ def threshold_otsu(image, nbins=256):
...
         msg = "threshold_otsu is expected to work correctly only for " \
               "grayscale images; image shape {0} looks like an RGB image"
         warn(msg.format(image.shape))
 
     # Check if the image is multi-colored or not
-    if image.min() == image.max():
-        raise ValueError("threshold_otsu is expected to work with images "
-                         "having more than one color. The input image seems "
-                         "to have just one color {0}.".format(image.min()))
+    first_pixel = image.ravel()[0]
```
emmanuelle, line 280: I feel like I'm the grumpy conservative person here but this change does not bring an improvement on my machine
```python
In [20]: arr = np.zeros((10000, 10000))   

In [21]: %timeit arr.min() == arr.max()                                                                                             
54.2 ms ± 168 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [22]: %timeit np.all(arr == arr.ravel()[0])                                                                                      
60.5 ms ± 334 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
Is it possible to see more benchmarks?

rfezzani: On my machine:
```python
In [1]: arr = arr = np.zeros((10000, 10000))                                                             

In [2]: %timeit arr.min() == arr.max()                                                                   
57.7 ms ± 119 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit np.all(arr == arr.ravel()[0])                                                            
28.9 ms ± 19.8 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

OGordon100: I've also rerun on my Linux machine, and done it backwards to ensure there's nothing clock speed related going on

```
arr = np.zeros((10000, 10000))

%timeit np.all(arr == arr.ravel()[0])
32.8 ms ± 1.42 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit arr.min() == arr.max()
43.5 ms ± 1.25 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

jni: I also see a speedup, as noted on the issue... @emmanuelle your computer is the weird one here it seems! 😬 But yes, it would be good to get to the bottom of this discrepancy. Logically, I have agree that one pass should be cheaper than two passes!

alexdesiqueira:
Thank you @OGordon100! We appreciate your help!

PR #4819, opened 2020-07-03
Simplify equivalent_diameter function

OGordon100:
The equivalent_diameter function in regionprops is a little messy, and only works with 2/3 dimensions (although I think this is a moot point as the regionprops function only applies to 2/3D images).

Regardless, it's a little messy, and the proposed PR neatens this up and removes needless if statements


alexdesiqueira:
Thank you @OGordon100!

PR #4820, opened 2020-07-03
Change feret_diameter to feret_diameter_max

### LINKED ISSUE #4817: rename `feret_diameter` -> `max_feret_diameter` ?
## Description

Sorry, I meant to comment on this while it was still an open PR, but then forgot to do so.
As it has not been in a release, maybe it is not too late to rename the new region property
`feret_diameter` to `max_feret_diameter`. 
Tagging @emmanuelle as she added it.

The simple reason being is that sometimes the minimum feret diameter is of interest and
it would be nice to be able to have naming consistency so that a `min_feret_diameter` property
can get added later.

maxfrei750:
## Description
Fixes #4817 

  later.

### DIFF
--- a/skimage/measure/_regionprops.py
+++ b/skimage/measure/_regionprops.py
@@ -736,7 +736,7 @@ def regionprops(label_image, intensity_image=None, cache=True,
     **extent** : float
         Ratio of pixels in the region to pixels in the total bounding box.
         Computed as ``area / (rows * cols)``
-    **feret_diameter** : float
+    **feret_diameter_max** : float
         Maximum Feret's diameter computed as the longest distance between
         points around a region's convex hull contour as determined by
         ``find_contours``. [5]_
--- a/skimage/measure/_regionprops.py
+++ b/skimage/measure/_regionprops.py
@@ -36,7 +36,7 @@ PROPS = {
     'EulerNumber': 'euler_number',
     'Extent': 'extent',
     # 'Extrema',
-    'FeretDiameter': 'feret_diameter',
+    'FeretDiameterMax': 'feret_diameter_max',
     'FilledArea': 'filled_area',
     'FilledImage': 'filled_image',
     'HuMoments': 'moments_hu',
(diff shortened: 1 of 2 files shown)


maxfrei750:
I thought ```feret_diameter_max``` might make more sense than ```max_feret_diameter```, because that way all potential Feret diameters would be grouped, in alphabetic property lists (e.g. documentation).

jni:
@maxfrei750 It looks like some tests are failing. I’m on my phone so I can’t investigate right now. The changes look ok to me. 

@scikit-image/core I’m agnostic on `max_feret` vs `feret_max`. max- sounds better but I agree with @maxfrei750 that the grouping with -max/-min is a nice idea. 

On the other hand, we have max-intensity, min-intensity, and mean-intensity...

Any thoughts?

maxfrei750:
@jni I already saw the errors, however I don't understand what they are caused by. There are loads of key errors like the following one in the "collection" section:
`ERROR feature/tests/test_util.py - KeyError: 'feret_diameter_max'`

Concerning `max_feret` vs `feret_max`: I agree that a consistency with `min-intensity`, etc. would be nice. However, I still like the idea of similar properties being grouped. Especially, since the Feret diameters would be mixed with the intensities, if both were to begin with min, mean, max. Therefore, IMHO it would be good to rename the intensities as well. Of course, breaking the API is quite drastic, so maybe for now, we could go with `max_feret` and change it in 1.0, along with the intensities.

grlee77:
The docstring in `regionprops` needs to be updated as well. At the moment it still states
```
    **feret_diameter** : float
        Maximum Feret's diameter computed as the longest distance between
```
which doesn't match the update name. There is some code in `_parse_docs` that parses the `regionprops` docstring and might be the source of the errors you are seeing.

maxfrei750:
@grlee77 Thanks for the hint. Changing the docstring has fixed the problems.

maxfrei750:
I'd like to finalize this PR. Any final thoughts concerning the naming?

stefanv:
I like the way it is now---it is more easily discoverable.

jni:
Looks like we have a consensus!!! Thank you @maxfrei750!

Issue #4538, opened 2020-03-30
Put our example data on a GitHub pages site


jni:
## Description

With pooch, we are able to download large datasets on demand, and no longer need to distribute them with scikit-image. We need a place to host the datasets now.

For my [blog](https://github.com/jni/ilovesymposia.com), I host on GitHub pages and add Cloudflare CDN on top of it, and this might work really well.

https://blog.cloudflare.com/secure-and-fast-github-pages-with-cloudflare/
https://www.codementor.io/@landonpatmore/how-to-setup-a-static-website-using-github-pages-and-cloudflare-with-your-own-domain-name-jb99nbuoe

We can also consider GitLab, which has more generous bandwidth limits — in fact, no bandwidth limits, just a 5GB storage limit, which should be plenty.

emmanuelle:
Hi @jni, some naive questions on my end
- why github pages instead of just a github repo, where we would put the link to the master code arborescence in the registry?
- do we think we'll have so many downloads that we'll need the CDN / SSL ?

I agree that Gitlab is an interesting alternative since the total size of the example files could exceed 1 Go. 

We should also make sure that we don't use `download_all` in our tests when the repo starts to be big...

jni:
> why github pages instead of just a github repo, where we would put the link to the master code arborescence in the registry?

(a) because github URLs are not super nice, compared to eg. `data.scikit-image.org/astronaut.png`.
(b) I don't know what github bandwidth limits are, if any, but I was suspicious that several CI runs failed after we started using pooch. I'm not sure whether they were related but I became paranoid about hitting bandwidth limits or some other kind of distribution issues, e.g. too many simultaneous hits to the same file getting an IP blocked as suspicious.

> do we think we'll have so many downloads that we'll need the CDN / SSL ?

As above, I'm not sure but I was suspicious. I don't remember the setup being hard, and if we start using large files, it might actually speed things up for data downloads around the world.

emmanuelle:
> (a) because github URLs are not super nice, compared to eg. `data.scikit-image.org/astronaut.png`.

ok but the idea is to put the URL in the registry and just use `load`, right? So we should not really care about how URLs look?

emmanuelle:
About bandwidth, we will probably want to add the files to the CI cache if this is possible, so that we don't download the files every time the doc is built (since we will probably want to write examples using our datasets).

PR #4823, opened 2020-07-06
Richardson-Lucy deconv: add optional filtering, work in > 2d

chrisroat:
## Description

Enhancement:
- Optional filter intermediate results to avoid divide-by-small numbers
- Change PSF flip to work in ndim!=2

The astronaut image w/o added noise does not work with the
standard RL algorithm.  These change fixes that, and adds
a corresponding test for the astronaut image.

## Checklist

I have added a unit test and documentation for a new optional parameter.

## For reviewers

  later.


skimage/restoration/deconvolution.py
```diff
@@ -364,10 +367,14 @@ def richardson_lucy(image, psf, iterations=50, clip=True):
     image = image.astype(np.float)
     psf = psf.astype(np.float)
     im_deconv = np.full(image.shape, 0.5)
-    psf_mirror = psf[::-1, ::-1]
+    psf_mirror = np.flip(psf)
```
emmanuelle, line 370: was this a bug which you corrected?

skimage/data/_registry.py
```diff
@@ -116,6 +116,8 @@
     "data/bw_text_skeleton.npy": "9ff4fc23c6a01497d7987f14e3a97cbcc39cce54b2b3b7ee33b84c1b661d0ae1",
     "data/_blobs_3d_fiji_skeleton.tif": "5182a2a94f240528985b8d15ec2aebbd5ca3c6b9214beff1eb6099c431e12b7b",
     "data/checker_bilevel.png": "2e207e486545874a2a3e69ba653b28fdef923157be9017559540e65d1bcb8e28",
+    "restoration/tests/astronaut_rl.npy":
```
emmanuelle, line 119: would it be possible to host this image on https://gitlab.com/scikit-image/data/ in order to limit the size of the package? See for example what has been done in https://github.com/scikit-image/scikit-image/pull/4809/

skimage/restoration/deconvolution.py
chrisroat: Yes, this was the "Change PSF flip to work in ndim!=2" part of the description

skimage/data/_registry.py
chrisroat: Is this really wise?  It feels inefficient, and trying to figure out how this will proceed seems painful:

- This requires me to setup an account in entirely different service.
- How do I run the tests prior to the data file appearing in gitlab?
- Back and forth on this PR could require additional PRs on gitlab to update the testdata.

This feels like a lot of extra effort for contributors, and will likely discourage contributions.

I will attempt it.

jni: @chrisroat you make good points! However, since we usually squash-and-merge at the end of a PR, you can leave the hosting change until the very last commit. We should definitely have better documentation and process around this, thank you for raising these issues — and for giving it a go!

jni: Also, just btw, we are having discussions about this in #4538. The *first* motivation for gitlab is that it *seemed* we were hitting github bandwidth limits with large data files being downloaded from github. But we never really verified that.

chrisroat:
Unfortunately, it does not seem simple.  My first error was fixed by digging by the code to find that I need the 'pooch' dependency.  

Then, I find the hashes are not agreeing.  If I clone the gitlab repo with the new data file, I find (as expected) that the file's hash agrees with what I had in this PR:
```
$ openssl sha256 data/astronaut_rl.npy 
SHA256(data/astronaut_rl.npy)= 3f8373e2c6182a89366e51cef6624e3625deac75fdda1079cbdad2a33322152c
```

However, when I run the test using the image fetcher, it fails.  It also is sneaky and deletes the download for safety.

```
$ pytest skimage/restoration/tests/test_restoration.py::test_richardson_lucy_filtered
...
E           ValueError: SHA256 hash of downloaded file (astronaut_rl.npy) does not match the known hash: expected 3f8373e2c6182a89366e51cef6624e3625deac75fdda1079cbdad2a33322152c but got f862d784dadd3286a60dfd79c72a718468b7eabb06bf720f3348505410a305c8. Deleted download for safety. The downloaded file may have been corrupted or the known hash may be outdated.
...
```

chrisroat:
Digging through the code, the best I can tell what's happening is that the pooch-downloaded file is only 39k, while the file is really 2MB.

chrisroat:
It's very odd.

Cloning the repo, the file shows up as 2.1M.  I also see the same figure when browsing the repo on the web.

```
[localhost] ~/tmp 
$ git clone https://gitlab.com/scikit-image/data.git
Cloning into 'data'...
remote: Enumerating objects: 71, done.
remote: Counting objects: 100% (71/71), done.
remote: Compressing objects: 100% (43/43), done.
remote: Total 71 (delta 28), reused 64 (delta 26), pack-reused 0
Unpacking objects: 100% (71/71), done.

[localhost] ~/tmp 
$ ls -l data/astronaut_rl.npy 
-rw-r--r-- 1 croat croat 2.1M Jul 10 11:25 data/astronaut_rl.npy
```

Yes, when I do a wget, it's the 39k that pooch downloads:
```
[localhost] ~/tmp 
$ wget https://gitlab.com/scikit-image/data/-/blob/master/astronaut_rl.npy
--2020-07-10 11:25:42--  https://gitlab.com/scikit-image/data/-/blob/master/astronaut_rl.npy
Resolving gitlab.com (gitlab.com)... 172.65.251.78, 2606:4700:90:0:f22e:fbec:5bed:a9b9
Connecting to gitlab.com (gitlab.com)|172.65.251.78|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/html]
Saving to: ‘astronaut_rl.npy’

astronaut_rl.npy                  [ <=>                                              ]  38.31K  --.-KB/s    in 0.003s  

2020-07-10 11:25:42 (12.1 MB/s) - ‘astronaut_rl.npy’ saved [39233]


[localhost] ~/tmp 
$ ls -l astronaut_rl.npy 
-rw-r--r-- 1 croat croat 39K Jul 10 11:25 astronaut_rl.npy
```

chrisroat:
Looking in the file, I see it's a web page, not the data.  Fixed.

chrisroat:
I looked through the failures: seems to be due to the lack of 'pooch' as a dependency in the test setup.

emmanuelle:
Thanks for bearing with us @chrisroat . We set up this pooch mechanism for good reasons (limiting the size of the package) but it definitely has its drawbacks, and it's frustrating.

skimage/data/_registry.py
emmanuelle: are tests not part of the sdist package? My initial remark was because I cared about the package size, for the repo size as you're pointing out @jni we're not so sure.

emmanuelle: ok, we do ship the test files also with the pip package, so that's the reason why we don't want to include too many files.

skimage/restoration/tests/test_restoration.py
```diff
@@ -87,6 +88,19 @@ def test_richardson_lucy():
...
 
+def test_richardson_lucy_filtered():
+    test_img_astro = rgb2gray(astronaut())
+
+    psf = np.ones((5, 5)) / 25
+    data = convolve2d(test_img_astro, psf, 'same')
+    deconvolved = restoration.richardson_lucy(data, psf, 5,
+                                              filter_epsilon=1e-6)
+
+    path = image_fetcher.fetch('restoration/tests/astronaut_rl.npy')
```
emmanuelle, line 99: Here you can use the `fetch` function from `skimage._shared.testing` as in https://github.com/scikit-image/scikit-image/blob/master/skimage/data/tests/test_data.py#L144; the function skips the test if pooch is not installed. Almost there :-).

skimage/data/_registry.py
jni: @emmanuelle but new test files can/should be managed by pooch and not shipped. We can certainly put them in this repo — as I said part of it was a hunch to avoid hitting the github bandwidth limits. All of this can be revisited easily, since it is a matter of updating some URLs.

> for the repo size as you're pointing out @jni we're not so sure.

Since some data files (e.g. cells.tif) are 30MB+, it would be pretty easy to get to a multi-100MB repo download, which is not ideal! So I still think having new data in a separate repo is a worthwhile goal.

skimage/restoration/tests/test_restoration.py
jni: @emmanuelle but do we want to skip the tests in CI? I'm kinda confused here as to why we *don't* have pooch as a test dependency, at least in *most* builds.

skimage/data/_registry.py
emmanuelle: we should document this process in the dev doc.

skimage/restoration/tests/test_restoration.py
chrisroat: It does not feel right that any test is skipped automatically if pooch is not installed -- especially if its implicit.  My preference would be to never skip and enforce pooch installation, or allow each to have a `skipif` explicitly inside so that people reading the code see the dependency.

In general, I'd be worried about long-run problems if the package depends on data downloaded outside of its own version control.

But I am happy to do whatever you feel is necessary at the moment, even if there are longer-term changes coming.  Just let me know the final decision.

chrisroat: And to give a concrete suggestion, it might be worth replacing large test files with smaller ones and/or enforcing that future test files are below some threshold in size.  I chose to use the astronaut in this test because it was already being used for a test of similar functionality.  But there is no real reason I couldn't use a smaller image.

jni: @chrisroat I suggest just add pooch to the test requirements for now. We can revisit that decision in a future PR, since there's (as you can tell) a lot of tweaking to be done about pooch.

chrisroat:
The pooch dep is now part of this PR.  The remaining failed test seems to be failing at HEAD.

emmanuelle:
don't worry about the travis CI fail, it's due to matplotlib pre-release and not related to your PR.

emmanuelle:
@jni still ok to merge this one? It's ready for me now.

jni:
Brilliant, thank you @chrisroat!

PR #4826, opened 2020-07-06
style: be explicit when stacking arrays

grlee77:
## Description

This PR relies on calling `np.stack` explicitly rather than relying on implicit stacking of lists of arrays via `np.array`. I think this "explicit is better than implicit" approach is preferable from a readability standpoint. 

There should be no change in existing behavior, it is purely a style/maintenance PR. That said, there will be a change in the contiguity of the returned type in at least one place. In `adapt_rgb.py` the current code

```Python
    return np.moveaxis(np.array(c_new), 0, -1)
```
returns an F-ordered array while the updated code

```Python
    return np.stack(c_new, axis=-1)
```
returns a C-ordered array. I don't think we typically guarantee a specific contiguity for the returned arrays, but it is something to consider.

This issue was inspired by discovering that passing lists of arrays to `array()`, or `tranpose()` does not work in CuPy. So, in addition to style/readability benefits, it may also ease potentially supporting other array module backends in the future.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/adapt_rgb.py
+++ b/skimage/color/adapt_rgb.py
@@ -76,4 +76,4 @@ def each_channel(image_filter, image, *args, **kwargs):
     """
     c_new = [image_filter(c, *args, **kwargs)
              for c in np.moveaxis(image, -1, 0)]
-    return np.moveaxis(np.array(c_new), 0, -1)
+    return np.stack(c_new, axis=-1)
--- a/skimage/color/colorconv.py
+++ b/skimage/color/colorconv.py
@@ -338,12 +338,13 @@ def hsv2rgb(hsv):
     v = arr[..., 2]
 
     hi = np.stack([hi, hi, hi], axis=-1).astype(np.uint8) % 6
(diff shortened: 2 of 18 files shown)


grlee77:
These were only the instances I came across when working on GPU conversions for  `cupyimg`. Let me do a quick grep of the codebase to see if there are any other obvious instances in functions I didn't consider there.

grlee77:
In revisiting this, implicit stacking of a list of arrays (even if they are NumPy arrays!) along axis 0 when calling 
`cupy.array` does work with recent CuPy (master) + NumPy (1.18.5). I don't think that was previously the case.

So, some of these changes are no longer strictly needed for that purpose.  I still don't think it hurts to be explicit in using `stack` instead, though.

alexdesiqueira:
@grlee77 thanks Greg!

jni:
Very lovely, thanks @grlee77!

Issue #4827, opened 2020-07-07
`equalize_adapthist` returns input image for clip value of 1.0


funkey:
## Description

A clip value of 1.0 (or larger) should perform AHE. Instead, the current implementation returns the input image (converted to float).

## Version information

```
3.7.6 (default, Jan  8 2020, 19:59:22) 
[GCC 7.3.0]
Linux-5.4.0-40-generic-x86_64-with-debian-bullseye-sid
scikit-image version: 0.18.dev0
numpy version: 1.18.1
```

PR #4828, opened 2020-07-07
Fix equalize_adapthist (CLAHE) for clip value 1.0, fix docstrings

funkey:
## Description

Addresses https://github.com/scikit-image/scikit-image/issues/4827 and fixes docstrings.

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -105,7 +105,8 @@ def _clahe(image, kernel_size, clip_limit, nbins):
     kernel_size: int or N-tuple of int
         Defines the shape of contextual regions used in the algorithm.
     clip_limit : float
-        Normalized clipping limit (higher values give more contrast).
+        Normalized clipping limit between 0 and 1 (higher values give more
+        contrast).
     nbins : int
         Number of gray bins for histogram ("data range").
 
@@ -116,8 +117,7 @@ def _clahe(image, kernel_size, clip_limit, nbins):
 
     The number of "effective" graylevels in the output image is set by `nbins`;
     selecting a small value (eg. 128) speeds up processing and still produce
-    an output image of good quality. The output image will have the same
-    minimum and maximum value as the input image. A clip limit smaller than 1
+    an output image of good quality. A clip limit of 0 or larger equal 1
     results in standard (non-contrast limited) AHE.
     """
     ndim = image.ndim
--- a/skimage/exposure/_adapthist.py
+++ b/skimage/exposure/_adapthist.py
@@ -159,7 +159,8 @@ def _clahe(image, kernel_size, clip_limit, nbins):
     if clip_limit > 0.0:
         clim = int(np.clip(clip_limit * np.product(kernel_size), 1, None))
     else:
-        clim = NR_OF_GRAY  # Large value, do not clip (AHE)
+        # largest possible value, i.e., do not clip (AHE)
+        clim = np.product(kernel_size)
 
     hist = np.apply_along_axis(np.bincount, -1, hist_blocks, minlength=nbins)
     hist = np.apply_along_axis(clip_histogram, -1, hist, clip_limit=clim)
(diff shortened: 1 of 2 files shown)


funkey:
[This test](https://github.com/scikit-image/scikit-image/blob/master/skimage/exposure/tests/test_exposure.py#L479) fails now, which was authored by @rfezzani and @sciunto. Can you confirm that this is really the expected behavior?

sciunto:
Why this test wouldn't be correct in your opinion?

funkey:
Hi François,

I mentioned that briefly in https://github.com/scikit-image/scikit-image/issues/4827, but let me provide some more details here of how I understand the `clip_limit` parameter: From the docstring and the implementation, I assume that `clip_limit` is relative to the size of the kernel (i.e., the effective, absolute clip limit is `clip_limit * np.prod(kernel_size)`). A clip limit of 0.5, for example, would limit the histogram of intensities to at most 50% of all pixels in the kernel. Excess values in the histogram are evenly distributed across the other bins, hence performing the contrast limiting (CL) part of CLAHE. A clip limit of 1.0, as I understand it, would effectively disable clipping (no bin in the histogram can have more than 100% of all pixels). However, that doesn't mean that the input image remains unchanged. Adaptive histogram equalization is still performed (i.e., AHE), replacing the intensity of each pixel with its rank divided by kernel size.

This understanding is also in line with the progression of histogram equalization over clip limits close to 1.0. Compare, for instance, clip values of 0.9, 0.95, 1.0, and 1.05. The last value performs AHE in the current implementation. The first two values perform CLAHE with decreasing effect of clipping, i.e., in the limit AHE. In this sequence, I would find it surprising if 1.0 would not perform AHE, but instead return the original image.

sciunto:
I'm afraid, as far as I remember, that we implemented the test after reading the misbehavior. It seems to me that you are correct, any other opinion @scikit-image/core ?

skimage/exposure/_adapthist.py
```diff
@@ -119,8 +117,7 @@ def _clahe(image, kernel_size, clip_limit, nbins):
 
     The number of "effective" graylevels in the output image is set by `nbins`;
     selecting a small value (eg. 128) speeds up processing and still produce
```
mkcor, line 119:
```suggestion
    selecting a small value (e.g. 128) speeds up processing and still produces
```

skimage/exposure/_adapthist.py
```diff
@@ -119,8 +117,7 @@ def _clahe(image, kernel_size, clip_limit, nbins):
...
-    an output image of good quality. The output image will have the same
-    minimum and maximum value as the input image. A clip limit smaller than 1
+    an output image of good quality. A clip limit of 0 or larger equal 1
```
mkcor, line 120: Is "a clip limit either == 0 or >=1" what is intended?

funkey: For CLAHE, the clip value would be in the open interval (0, 1). Clip values >=1 essentially disable clipping and therefore simply perform AHE. A clip value of 0 does not really make sense, so the implementation falls back to AHE in this case.

jni:
From a cursory glance, I agree with these changes and that the test should be updated. @funkey (:wave:) can you update the test accordingly?

funkey:
I believe the tests are fixed, the CircleCI error seems unrelated to me. Let me know if I can do anything else!

stefanv:
Thanks for investigating, @funkey!

skimage/exposure/_adapthist.py
emmanuelle, line 120:
```suggestion
    an output image of good quality. A clip limit of 0 or larger than or equal to 1
```

emmanuelle:
@funkey I opened a small PR in your fork with minor improvements suggested in the review of this PR.

emmanuelle (approved):
Thank you @funkey the PR now looks ready to me!

mkcor (approved):
I have left minor suggestions!

skimage/exposure/_adapthist.py
mkcor: Thanks for the very clear response!

jni:
@mkcor I'm a bit ambivalent about the suggestions so I decided to merge this so we can move on with the release. Minor wording changes in docstrings can come later. :blush: Thanks for reviewing!

mkcor:
@jni for sure! :)

PR #4837, opened 2020-07-11
Silence several warnings in the test suite

grlee77:
## Description

These span multiple modules, but it is broken into small commits that should be easy to review individually.

The first 4 commits are very simple and just silence FutureWarnings in the existing tests.

Commit (519b8106) silences a `FutureWarning`, but is also a change in boundary behavior from the old default value of "constant" to the upcoming default of "reflect". Let me know if you want to set that to "constant" instead to keep the old behavior (not sure we are even updating/maintaining the `viewer` module anymore?)

After the fixes in this PR, I am now seeing only 3 `RuntimeWarnings` locally and can look into those in a separate PR.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/corner.py
+++ b/skimage/feature/corner.py
@@ -995,7 +995,7 @@ def corner_peaks(image, min_distance=1, threshold_abs=None, threshold_rel=None,
                             threshold_abs=threshold_abs,
                             threshold_rel=threshold_rel,
                             exclude_border=exclude_border,
-                            indices=True, num_peaks=np.inf,
+                            num_peaks=np.inf,
                             footprint=footprint, labels=labels,
                             num_peaks_per_label=num_peaks_per_label)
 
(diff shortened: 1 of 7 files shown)


skimage/feature/corner.py
```diff
@@ -995,7 +995,7 @@ def corner_peaks(image, min_distance=1, threshold_abs=None, threshold_rel=None,
                             threshold_abs=threshold_abs,
                             threshold_rel=threshold_rel,
                             exclude_border=exclude_border,
-                            indices=True, num_peaks=np.inf,
```
emmanuelle, line 998: could you please explain why you had to remove this one?

grlee77: `peak_local_max` has a `remove_arg` decorator stating that this argument will be removed starting in version 0.20. From the docstring to that function:
```
    indices : bool, optional
        If True, the output will be an array representing peak
        coordinates. The coordinates are sorted according to peaks
        values (Larger first). If False, the output will be a boolean
        array shaped as `image.shape` with peaks present at True
        elements. ``indices`` is deprecated and will be removed in
        version 0.20. Default behavior will be to always return peak
        coordinates. You can obtain a mask as shown in the example
        below.
```

emmanuelle: sorry, was looking at the code in an old branch :-(.

PR #4840, opened 2020-07-11
handle color string mapping correctly

### LINKED ISSUE #4534: BUG: name of bg_color is not supported in skimage.color.label2rgb
## Description
Docstring of `label2rgb` says
```
    bg_color : str or array, optional
        Background color. Must be a name in `color_dict` or RGB float values
        between [0, 1].
```
but only array of `bg_color` is supported in the `kind` of overlay.
The reason is lack of processing for string at here
https://github.com/scikit-image/scikit-image/blob/7227346b262659103392b5701b057b776664f9aa/skimage/color/colorlabel.py#L46-L48

## Way to reproduce
we can simply reproduce the bug by using the example `doc/examples/segmentation/plot_label.py`
```python
from skimage import data
from skimage.filters import threshold_otsu
from skimage.segmentation import clear_border
from skimage.measure import label
from skimage.morphology import closing, square
from skimage.color import label2rgb


image = data.coins()[50:-50, 50:-50]

thresh = threshold_otsu(image)
bw = closing(image > thresh, square(3))

cleared = clear_border(bw)

label_image = label(cleared)
image_label_overlay = label2rgb(label_image, image=image, bg_label=0, bg_color='red')
```

## Version information
```python
```

```python
# your output here
3.6.10 | packaged by conda-forge | (default, Mar  5 2020, 09:56:10) 
[GCC Clang 9.0.1 ]
Darwin-18.7.0-x86_64-i386-64bit
scikit-image version: 0.17.dev0
numpy version: 1.18.1
```

wwymak:
## Description

Resolves #4534
I think previously the way the color isn't handled as intended in `_match_label_with_color` because if you pass `[color]` into `_rgb_vector` it's not dealing with color the way it's intended (you should only pass one color, not a list)

Also fixes (I think?) the vector size mismatch in  `label2rgb` with kind='avg' 
## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/tests/test_colorlabel.py
+++ b/skimage/color/tests/test_colorlabel.py
@@ -6,7 +6,7 @@ from skimage.color.colorlabel import label2rgb
 
 from skimage._shared import testing
 from skimage._shared.testing import (assert_array_almost_equal,
-                                     assert_array_equal, assert_warns)
+                                     assert_array_equal, assert_warns, assert_no_warnings)
 
 
 def test_deprecation_warning():
@@ -198,3 +198,11 @@ def test_negative_intensity():
     labels = np.arange(100).reshape(10, 10)
     image = np.full((10, 10), -1, dtype='float64')
     assert_warns(UserWarning, label2rgb, labels, image, bg_label=-1)
+
+
+def test_bg_color_rgb_string():
+    img = np.random.randint(0, 255, (10, 10), dtype=np.uint8)
+    labels = np.zeros((10, 10), dtype=np.int64)
+    labels[1:3, 1:3] = 1
+    labels[6:9, 6:9] = 2
+    assert_no_warnings(label2rgb,labels, image=img, bg_label=0, bg_color='red')
--- a/skimage/color/tests/test_colorlabel.py
+++ b/skimage/color/tests/test_colorlabel.py
@@ -6,7 +6,8 @@ from skimage.color.colorlabel import label2rgb
 
 from skimage._shared import testing
 from skimage._shared.testing import (assert_array_almost_equal,
-                                     assert_array_equal, assert_warns, assert_no_warnings)
+                                     assert_array_equal, assert_warns,
+                                     assert_no_warnings)
 
 
 def test_deprecation_warning():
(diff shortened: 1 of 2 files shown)


skimage/color/tests/test_colorlabel.py
```diff
@@ -198,3 +199,19 @@ def test_negative_intensity():
...
     image = np.full((10, 10), -1, dtype='float64')
     assert_warns(UserWarning, label2rgb, labels, image, bg_label=-1)
+
+
+def test_bg_color_rgb_string():
+    img = np.random.randint(0, 255, (10, 10), dtype=np.uint8)
+    labels = np.zeros((10, 10), dtype=np.int64)
+    labels[1:3, 1:3] = 1
+    labels[6:9, 6:9] = 2
+    assert_no_warnings(label2rgb, labels, image=img, bg_label=0, bg_color='red')
```
emmanuelle, line 209: would you mind making the test more specific, like for example testing that one of the background pixels is indeed red? (if it's not too complicated)

emmanuelle:
Very cool @wwymak  I think you nailed the bug beautifully! For the `'avg'` mode does this correspond to a different issue ? I asked for a minor modification of the test, I'm nitpicking a bit here :-).

wwymak:
> For the `'avg'` mode does this correspond to a different issue ?

I don't think so (or at least I didn't see an issue about this when I searched for `label2rgb` in the issues list)-- I found this bug when I was trying to fix the original bg_color string issue :)

skimage/color/tests/test_colorlabel.py
wwymak: hmm, I am not sure it's so easy to test because the background pixels isn't really red in the `overlay` mode because of https://github.com/scikit-image/scikit-image/blob/c9151096cf095afb3811c8e27d491e52f989e5fb/skimage/color/colorlabel.py#L197

Or do you think I should plug in the defaults and checked that the background pixel matches the output of that calculation referenced ^^?

wwymak:
(I am also not sure I understand the failed build so if you can point me towards what needs fixing that would be great)

grlee77:
> (I am also not sure I understand the failed build so if you can point me towards what needs fixing that would be great)

Do not worry about that failed build. It is due to an unrelated bug upstream in Matplotlib (matplotlib/matplotlib#17730).

skimage/color/tests/test_colorlabel.py
emmanuelle:
```suggestion
    output = label2rgb(labels, image=img, bg_label=0, bg_color='red', alpha=0.9)
    assert output[0, 0, 0] > 200  # red channel
```

emmanuelle: increasing `alpha` makes it easier to test the red color :-)

emmanuelle:
@scikit-image/core this bug-fix PR is ready I think, please review!

Issue #4831, opened 2020-07-08
Work on structure_tensor: gallery example, row/col notation, port to 3D


emmanuelle:
This issue is about improvements to `feature.structure_tensor`.

- [x] First of all, it does not have a gallery example. https://docs.opencv.org/master/d4/d70/tutorial_anisotropic_image_segmentation_by_a_gst.html has a nice example which could be adapted.

- [x] Second, the code of structure_tensor uses x and y instead of r and c. See https://github.com/scikit-image/scikit-image/pull/2987 for a discussion on how to adapt what has been done for `hessian_matrix` to `structure_tensor`.

- [x] Finally, https://github.com/scikit-image/scikit-image/pull/2987 provided a 3D implementation, which should be integrated into the main `structure_tensor` function. https://github.com/scikit-image/scikit-image/pull/2489/ had an nD implementation as well. Also see the [wikipedia page on structure tensor](https://en.wikipedia.org/wiki/Structure_tensor) for merging the two docstrings.

These points can be addressed in separate pull requests.

coreysharris:
I am working on this, starting with updating the row/col notation.

grlee77:
Welcome @coreysharris!

I happened to just recently come across the [structure-tensor](https://github.com/Skielex/structure-tensor) respository which also has a [3D implementation](https://github.com/Skielex/structure-tensor/blob/master/structure_tensor/st3d.py) that is MIT-licensed. I haven't looked into whether it provides any benefit over the versions already in the open PRs here. Any help sorting that out would be appreciated.

grlee77:
Stage 1 of the 3D implementation was completed during the sprint in #4841. 

@coreysharris, You had previously [indicated additional progress](https://skimage.zulipchat.com/#narrow/stream/176555-sprint/topic/structure-tensor). Is it mainly a need for test cases that is still pending?

I just wanted to check in. Feel free to open a work-in-progress PR (start the PR title with WIP: ), even if you don't have time to finish it in the near future. That will help us make sure what you have done so far does not get lost!

coreysharris:
@grlee77, sorry for the delay on this. Life gets in the way 😄

I have written test cases. They could maybe use some work, but I'm happy to open a WIP PR to get some visibility. I'll try to get it done in the next couple days.

emmanuelle:
[...]

PR #2987, opened 2018-03-18
WIP: Adds 3D structure tensor capability

scott-trinkle:
## Description
Adds `skimage.feature.structure_tensor_3D` and `skimage.feature.structure_tensor_3D_eig`, which calculates the 3x3 structure tensor from a 3D image volume, and finds its eigenvalues and eigenvectors, respectively. 

## Checklist
- [X] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)
- [X] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
- [ ] Gallery example in `./doc/examples` (new features only)
- [ ] Unit tests

## References
Closes Issue [#2972](https://github.com/scikit-image/scikit-image/issues/2972)
Mathematical description available [here](https://en.wikipedia.org/wiki/Structure_tensor#The_3D_structure_tensor)

## For reviewers

(Don't remove the checklist below.)

- [ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
- [ ] Check that new functions are imported in corresponding `__init__.py`.
- [ ] Check that new features, API changes, and deprecations are mentioned in
      `doc/release/release_dev.rst`.

PR #4841, opened 2020-07-11
Change structure_tensor to use row/col-notation instead of x/y.

coreysharris:
## Description

This PR fixes the notation discrepancy in `skimage.feature.structure_tensor` by changing from x/y-notation to row/col-notation as noted in #4831. It also begins preparations for the 3D structure tensor work done in #2987.

## Checklist

[x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
[ ] Gallery example in `./doc/examples` (new features only)
[ ] Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark
[x] Unit tests
[x] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

## For reviewers

[ ] Check that the PR title is short, concise, and will make sense 1 year
  later.
[ ] Check that new functions are imported in corresponding `__init__.py`.
[ ] Check that new features, API changes, and deprecations are mentioned in


skimage/feature/tests/test_corner.py
```diff
@@ -31,22 +32,32 @@ def im3d():
 def test_structure_tensor():
     square = np.zeros((5, 5))
     square[2, 2] = 1
-    Axx, Axy, Ayy = structure_tensor(square, sigma=0.1)
```
emmanuelle, line 34: maybe you could keep this part in the test with an explicit `order='xy'`? We're deprecating it but it's still supported at the moment so we might as well keep the test.

emmanuelle: never mind, the function you added below does just this :-). I let you resolve the convo.

skimage/feature/corner.py
```diff
@@ -67,41 +67,59 @@ def structure_tensor(image, sigma=1, mode='constant', cval=0):
...
            [0., 4., 0., 4., 0.],
            [0., 1., 0., 1., 0.],
            [0., 0., 0., 0., 0.]])
 
     """
+    if order is None:
+        if image.ndim == 2:
+            # The legacy 2D code followed (x, y) convention, so we swap the
+            # axis order to maintain compatibility with old code
+            warn('deprecation warning: the default order of the structure '
```
grlee77, line 103:
```suggestion
            FutureWarning('The default order of the structure '
```

grlee77: I think this should be a FutureWarning? In the last sentence you may want to add (Specify order="xy" to maintain the old behavior).

grlee77: (it is fine if you want to edit in a separate commit instead of merging the suggestion here)

skimage/feature/tests/test_corner.py
```diff
@@ -31,22 +32,32 @@ def im3d():
...
+                                      [0, 1, 4, 1, 0],
+                                      [0, 0, 0, 0, 0],
+                                      [0, 1, 4, 1, 0],
+                                      [0, 0, 0, 0, 0]]))
+
+
+def test_structure_tensor_orders():
+    square = np.zeros((5, 5))
+    square[2, 2] = 1
+    A_elems_default = structure_tensor(square, sigma=0.1)
```
grlee77, line 56:
```suggestion
    with pytest.warns(FutureWarning,
                      match="The default order of the structure*"):
        A_elems_default = structure_tensor(square, sigma=0.1)
```

Emma, do you know if we switched to preferring this pytest-style over using the `expected_warnings` utility? I see both styles used in this file.

skimage/feature/corner.py
```diff
@@ -67,41 +67,59 @@ def structure_tensor(image, sigma=1, mode='constant', cval=0):
...
+        else:
+            order = 'rc'
 
     image = _prepare_grayscale_input_2D(image)
 
-    imx, imy = _compute_derivatives(image, mode=mode, cval=cval)
+    derivatives = _compute_derivatives(image, mode=mode, cval=cval)
+
+    if order == 'rc':
+        derivatives = reversed(derivatives)
```
grlee77, line 116: I'm not 100% sure if this is correct in 3D/nD. We can worry about that in the 3D/nD PR, though.

For example, mode `'xy'` (vs. 'yz') in NumPy's meshgrid only swaps the order of the first two while the rest remain the same. See the [Notes section here](https://numpy.org/doc/stable/reference/generated/numpy.meshgrid.html). This would instead reverse **all** axes for "xy" vs "rc". That may be correct, but I don't recall offhand what is done in other functions in `skimage`.

grlee77: This type of potential confusion is why I am keen to see the switch to rc convention as that seems the easiest to follow

jni: @grlee77 I suggest that we do not support 'xy' for ndim>2. It is only a temporary crutch until everything is rc.

skimage/feature/corner.py
```diff
@@ -67,41 +67,59 @@ def structure_tensor(image, sigma=1, mode='constant', cval=0):
...
+    order : {'rc', 'xy'}, optional
+        This parameter allows for the use of reverse or forward order of
+        the image axes in gradient computation. 'rc' indicates the use of
+        the first axis initially (Arr, Arc, Acc), whilst 'xy' indicates the
+        usage of the last axis initially (Axx, Axy, Ayy).
 
     Returns
     -------
-    Axx : ndarray
+    Arr : ndarray
```
jni, line 78: This return value is no longer correct for ndim > 2, is it?

skimage/feature/tests/test_corner.py
jni: We usually still use `skimage._shared.testing.expected_warnings`.

jni:
@emmanuelle why is this version not 3D?

emmanuelle:
> @emmanuelle why is this version not 3D?

@jni the docstring suggests it's 2D with only Arc, etc. The code is indeed compatible with 3D but I think Corey wanted first to get the notations right and then tackle 3D in another PR.

skimage/feature/tests/test_corner.py
grlee77: Okay, I updated the suggestion to use `expected_warnings` instead of `pytest.warns`.

skimage/feature/corner.py
grlee77: >  I suggest that we do not support 'xy' for ndim>2

Okay, then perhaps we should raise an error for order = 'xy' when ndim != 2

coreysharris: Awesome! That will make the next step much easier. How do we handle this exactly? If we get an image with ndim>2 and order='xy' should we raise an error or warning or just silently ignore the option?

coreysharris:
That's right. Switching to row/col notation is almost enough to support higher dimensions, but we also need to generalize `_compute_derivatives`, along with updating docstrings and unit tests (and adding gallery examples as a bonus?).

skimage/feature/tests/test_corner.py
coreysharris: Fixed, thanks!

skimage/feature/corner.py
coreysharris: I added the note at the end, but I didn't understand the Suggested change. The category of the warning is already set to `FutureWarning`. Did you want me to raise the warning instead of using `warn`?

grlee77: Ah, nevermind. I had missed the category argument here. I will just close this as resolved

skimage/feature/corner.py
grlee77: I would raise an error

skimage/feature/corner.py
grlee77: Let's leave this one to the follow-up PR adding tests and examples for ndim > 2 support

stefanv:
🎉 Thanks @coreysharris!

PR #4843, opened 2020-07-12
Github action to build docs

### LINKED ISSUE #4805: dev doc is not deployed when pre build is failing on Travis
It happens often that the pre build fails on Travis (at the moment due to Matplotlib's RC, which is ok because RCs are meant for this). However, when one of the Travis builds fail then the docs are not deployed by Travis, which is a problem. We should change the Travis configuration so that a failing pre build does not block the rest of the builds and deployment.

jvkersch:
## Description

Closes #4805 

This PR provides a Github Action to build the development documentation upon master merge. 

Problem setting: In some cases, Travis builds fail, e.g. because they use release candidates of dependencies and expose some upstream breakage. Such PRs are still merged, and the dev documentation should be updated (automatically) upon such a merge. However, the documentation build is currently part of the `after_success:` stanza in the Travis config file, and hence doesn't get executed if some of the builds fail.

Solution: as indicated in #4805, a Github action can be made to run upon merges to master (and pushes to master, which should be rare) to build the development documentation.

Notes:

- I tried to reproduce the documentation building process accurately, but I may have overlooked something.
- For this to work, `GH_TOKEN` should be configured as a Github secret in the project settings. I don't know if that's the case.
- A sample of the output docs appears here: https://jvkersch.github.io/doc-builder-output/dev/. See #4805 for details.

## Checklist

This is mostly an internal/configuration feature.

- ~[Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)~
- ~Gallery example in `./doc/examples` (new features only)~
- ~Benchmark in `./benchmarks`, if your changes aren't covered by an
  existing benchmark~
- ~Unit tests~
- ~Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)~

## For reviewers

  later.


.travis.yml
```diff
@@ -180,4 +180,4 @@ after_success:
     - codecov
     # Prepare.release
     - doc/release/contribs.py HEAD~10
-    - bash tools/travis/deploy_docs.sh
+    # - bash tools/travis/deploy_docs.sh
```
emmanuelle, line 183: Please remove this line and `git rm` the `deploy_docs.sh` file which is not used elsewhere.

emmanuelle:
I've created a github token under my account and I have stored it under the name GH_TOKEN in Github secrets, so hopefully this should do the job?

emmanuelle:
Thanks a lot for the PR @jvkersch . I'm curious to know whether you started from a specific template ? It makes sense while reading it but I would have a hard time starting from scratch without a template.

jvkersch:
@emmanuelle Thanks for the quick review! I've removed the deploy script.

As for template, I mostly used https://github.com/napari/napari/blob/master/.github/workflows/make_release.yml. I also worked through https://lab.github.com/githubtraining/github-actions:-hello-world and assumed that I'd have to fiddle with Docker files, but fortunately that was not the case.

I think having a token under `GH_TOKEN` in secrets should do the job, but it may be better to use a deploy key instead. The tokens give access to every repo under your account (depending on scope) which may be excessive. Deploy keys should in principle give access to one repo only, so they're more finely scoped. Or maybe you have a robot user in the `scikit-image` team that can be used for deployment?

jvkersch:
@emmanuelle I've experimented a bit with [deploy keys](https://developer.github.com/v3/guides/managing-deploy-keys/#deploy-keys) and I think this can be made to work. The procedure is more cumbersome to set up, but the advantage is that it is more secure (the generated key can only be used to access one repo (the `docs` repo), rather than all of your repos). Let me know if you want to pursue this option.

Other than that, if you want to test out this action before merging the PR, I can update the trigger to run on commits to this branch. We can then push a few commits here, and see if that generates a doc build + upload.

emmanuelle:
@jvkersch sounds good, I have created a deploy key, I've added the public key as a deploy key of the `docs` repo and I've updated the github secret for GH_TOKEN with the private part. Is it what you had in mind?

emmanuelle:
I'm not so sure that a private key can be used in the same way as a github token... Please tell me if I'm missing something. The idea of running on commits is fine, I'm just cloning first the docs repo for backup as we are doing `push -f`, in case something goes wrong.

emmanuelle:
If someone has a faster Internet connexion that me for a backup of the repo, it's huge (6 Go) and I'm downloading at 200 ko/s...

grlee77:
> If someone has a faster Internet connexion that me for a backup of the repo, it's huge (6 Go) and I'm downloading at 200 ko/s...

This is the scikit-image/docs repo? I just backed it up locally via `git clone https://github.com/scikit-image/docs.git` (but it came out around 1.9 GB rather than 6?).

jvkersch:
> I'm not so sure that a private key can be used in the same way as a github token... Please tell me if I'm missing something. 

@emmanuelle I think the setup that you provided should work fine, but you're right: there's a bit more configuration on the part of the workflow to use private keys. The step that pushes to the remote repository must set up an SSH agent and add the private key to it that was provided as the `GH_TOKEN`. I've made those modifications in the last commit in this branch.

I understand the concern about force-pushing against the docs repo. I've also made a local clone (also 1.9 GB, as @grlee77 reports) and I've forked the repo as well. Actually, I've even uploaded the zipped repo to S3, just in case...

Other than that, I think we're all set to run our experiment: if you give the go ahead I will update the trigger (the branch that needs to be updated for the action to run) to point to this branch, and then we can see if docs are getting built!

emmanuelle:
Thanks a lot @jvkersch. We should be safe with three backups of the repo :-). Please update the trigger as you suggested!

jvkersch:
@emmanuelle I updated the workflow, but it errors out when trying to push the docs, with what looks like an error for `GH_TOKEN` not being set. Would you mind double-checking that `GH_TOKEN` is listed under secrets for this repo, and has an actual value?

emmanuelle:
Hi @jvkersch I had thought the secret needed to be for the whole organization. I have added the GH_TOKEN secret with the private key as a secret for scikit-image/scikit-image. Can you please push an empty commit?

jvkersch:
@emmanuelle Things still don't work, but I know what the issue is ("we're still confused, but we're confused at a higher level"). The action runs under my fork of the scikit-image repo, so doesn't see the token that you set.

In order to test that things work correctly, we can merge the code in this PR into a different, short-lived branch of the main scikit-image repo. That would allow us to test and verify that things work as expected, and we can then merge this PR into master. Would you be open to this course of action? If so, taking the code in this branch and putting it in a new branch also called `dev/github-action-doc-build` should do the trick. I can do that via a (new) PR, or someone with push privileges can it do it right away.

emmanuelle:
@jvkersch sounds good, let me push a copy of your branch to the scikit-image repo.

emmanuelle:
See #4852 and let's keep fingers crossed :-).

emmanuelle:
Closing in favor of #4852, thank you so much @jvkersch for all your hard work :-).

PR #4844, opened 2020-07-12
fix the issue where ransac fails if all the inliers are false and add tests

### LINKED ISSUE #4816: skimage.image's ransac function does not handle well when best_inliers are all False
## Description

`ransac` routine is supposed to handle well on arbitrary data input.
However, when the best sample model regards all data points as all outliers, `ransac` routine will fail immediately after this line because the `best_inliers` are all `False`:
https://github.com/scikit-image/scikit-image/blob/34c2809ab509acd4e34a0a8c5530588adf0e68e7/skimage/measure/fit.py#L873

## Way to reproduce
```python

import numpy as np
from numpy.polynomial import Polynomial as Poly
from skimage.measure import ransac

class LinearXModel(object):
    def estimate(self, data):
        self.params = Poly.fit(data[:, 1], data[:, 0], 1, domain=[-1, 1])
        return True

    def residuals(self, data):
        return abs(self.params(data[:, 1]) - data[:, 0])

def main():
    data = np.array(
      [[ 943. , 1374.5],
       [1518. , 3293.5],
       [ 749. , 3450.5],
       [2237. , 3578.5],
       [1022. , 3668.5]], dtype=np.float64)
    model, inliers = ransac(data, LinearXModel, 3, 2.2, random_state=1523427)

if __name__ == "__main__":
    main()
```

The output:
```
  File "ransac_bug.py", line 23, in <module>
    main()
  File "ransac_bug.py", line 20, in main
    model, inliers = ransac(data, LinearXModel, 3, 2.2, random_state=1523427)
  File "/home/user/miniconda3/envs/rebook/lib/python3.6/site-packages/skimage/measure/fit.py", line 874, in ransac
    best_model.estimate(*data_inliers)
  File "ransac_bug.py", line 7, in estimate
    self.params = Poly.fit(data[:, 1], data[:, 0], 1, domain=[-1, 1])
  File "/home/user/miniconda3/envs/rebook/lib/python3.6/site-packages/numpy/polynomial/_polybase.py", line 892, in fit
    res = cls._fit(xnew, y, deg, w=w, rcond=rcond, full=full)
  File "/home/user/miniconda3/envs/rebook/lib/python3.6/site-packages/numpy/polynomial/polynomial.py", line 1336, in polyfit
    return pu._fit(polyvander, x, y, deg, rcond, full, w)
  File "/home/user/miniconda3/envs/rebook/lib/python3.6/site-packages/numpy/polynomial/polyutils.py", line 668, in _fit
    raise TypeError("expected non-empty vector for x")
TypeError: expected non-empty vector for x
```

## Version information
```python
```

```python
# your output here
3.6.10 |Anaconda, Inc.| (default, May  8 2020, 02:54:21)
[GCC 7.3.0]
Linux-4.15.0-70-generic-x86_64-with-debian-buster-sid
scikit-image version: 0.17.2
numpy version: 1.18.5
```

wwymak:
## Description

Resolves #4816
The ransac function fails if there are no inliers because https://github.com/scikit-image/scikit-image/blob/1ae430495cab620a0cd067224d01b2f90d1d58e9/skimage/measure/fit.py#L871 does not handle for the case `best_inliers = [False, False, False...]`.

This PR adds a check for this case and only fit the final model if at least 1 of the array of best_inliers is True. (and also added a test to check for this bug)
 
## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/measure/fit.py
```diff
@@ -868,7 +868,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
                 break
 
     # estimate final model using all inliers
-    if best_inliers is not None:
+    if best_inliers is not None and sum(best_inliers) > 0:
```
grlee77, line 871: It will be more efficient to use `any` rather than `sum` (any will stop checking values as soon as the first True value is found)

skimage/measure/fit.py
```diff
@@ -868,7 +868,7 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
         # select inliers for each data array
         data_inliers = [d[best_inliers] for d in data]
         best_model.estimate(*data_inliers)
```
grlee77, line 874: Should we add an `else` clause to set `best_model=None` when there are no inliers to use in estimating the model? Or maybe raise a UserWarning in that case? (I am not a `ransac` user, so am not sure what the best solution is)

mkcor: I would also add an `else` clause. Morally (haven't checked for types):
```py
if any(best_inliers):
    # estimate model
else:
    best_inliers = None
    best_model = None
    # warn user
```
and update test to check for `None` return values.

skimage/measure/tests/test_fit.py
```diff
@@ -386,3 +386,11 @@ def residuals(self, data):
     data = np.arange(4)
     ransac(data, DummyModel, min_samples=3, residual_threshold=0.0,
            max_trials=10)
+
+
+def test_ransac_with_no_final_inliers():
+    data = np.random.rand(5, 2)
+    model, inliers = ransac(data, model_class=LineModelND, min_samples=3,
+                            residual_threshold=0, random_state=1523427)
+    assert_equal(inliers, None)
```
mkcor, line 395:
```suggestion
    assert inliers is None
```

skimage/measure/fit.py
emmanuelle: @grlee77 @mkcor please resolve the convo if the change is ok for you

skimage/measure/tests/test_fit.py
```diff
@@ -386,3 +386,11 @@ def residuals(self, data):
     data = np.arange(4)
     ransac(data, DummyModel, min_samples=3, residual_threshold=0.0,
            max_trials=10)
+
+
+def test_ransac_with_no_final_inliers():
+    data = np.random.rand(5, 2)
+    model, inliers = ransac(data, model_class=LineModelND, min_samples=3,
```
emmanuelle, line 393: One thing I don't understand here is that our tests should fail if a warning is not caught in a specific context (like for example https://github.com/scikit-image/scikit-image/blob/master/skimage/transform/tests/test_warps.py#L666). Are you sure that the warning is issued? Maybe I'm missing something here...

skimage/measure/fit.py
```diff
@@ -868,9 +869,13 @@ def ransac(data, model_class, min_samples, residual_threshold,
...
     # estimate final model using all inliers
-    if best_inliers is not None:
+    if best_inliers is not None and any(best_inliers):
         # select inliers for each data array
         data_inliers = [d[best_inliers] for d in data]
         best_model.estimate(*data_inliers)
+    else:
+        best_model = None
+        best_inliers = None
+        warn("User warning: no inliers found. Model not fitted")
```
emmanuelle, line 879:
```suggestion
        warn("No inliers found. Model not fitted")
```

emmanuelle: `UserWarning` is the default category of `warnings.warn` so no need to mention `UserWarning` in the message (see https://docs.python.org/2/library/warnings.html#warnings.warn).

skimage/measure/tests/test_fit.py
wwymak: hmm, I definitely see the warning I run the test

```
Testing started at 18:33 ...
/home/wwymak/anaconda3/envs/skimage-dev/bin/python /snap/pycharm-professional/206/plugins/python/helpers/pycharm/_jb_pytest_runner.py --target test_fit.py::test_ransac_with_no_final_inliers
Launching pytest with arguments test_fit.py::test_ransac_with_no_final_inliers in /home/wwymak/libraries/scikit-image/skimage/measure/tests

============================= test session starts ==============================
platform linux -- Python 3.7.6, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /home/wwymak/anaconda3/envs/skimage-dev/bin/python
cachedir: .pytest_cache
rootdir: /home/wwymak/libraries/scikit-image/skimage/measure
collecting ... collected 1 item

test_fit.py::test_ransac_with_no_final_inliers PASSED                    [100%]

=============================== warnings summary ===============================
tests/test_fit.py::test_ransac_with_no_final_inliers
  /home/wwymak/libraries/scikit-image/skimage/measure/fit.py:879: UserWarning: No inliers found. Model not fitted
    warn("No inliers found. Model not fitted")

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================= 1 passed, 1 warning in 0.19s =========================

Process finished with exit code 0
```
:thinking:

skimage/measure/fit.py
mkcor: @emmanuelle I actually don't have 'resolve' permission -- is it possible? Also, I admit that I would love to see type hints in the `ransac` function... because an object could be `not None` but not iterable either...

skimage/measure/tests/test_fit.py
mkcor: The warning is issued indeed, but (apparently) this counts as passed.  I guess that's an upstream (not directly related) issue/choice regarding whether a warning should be considered a pass or a fail. In any case, it's cleaner to catch the warning in the dedicated context, so that we have a 'plain' pass.

mkcor, line 393:
```suggestion
    with expected_warnings(['No inliers found. Model not fitted']):
        model, inliers = ransac(data, model_class=LineModelND, min_samples=3,
```

skimage/measure/tests/test_fit.py
```diff
@@ -386,3 +386,11 @@ def residuals(self, data):
...
+                            residual_threshold=0, random_state=1523427)
```
mkcor, line 394:
```suggestion
                                min_samples=3, residual_threshold=0,
                                random_state=1523427)
```

mkcor: @wwymak if you include the above two code suggestions, also include
```py
from skimage._shared._warnings import expected_warnings
```
in the imports :)

skimage/measure/tests/test_fit.py
mkcor: edit: "whether a warning ..." --> "which kind of warning exactly should be treated as an error"

mkcor:
Dear @wwymak, @emmanuelle is on vacation this week. Thank you for your patience! :pray:

alexdesiqueira:
@grlee77 could I bother you for another round of reviews here? :blush: Thanks!

skimage/measure/tests/test_fit.py
grlee77: The line break suggested here was not made, but it is within the <80 character limit as is so I will mark this as resolved

PR #4850, opened 2020-07-16
Port 3D image processing tutorial into gallery.

mkcor:
## Description

In this PR, we are porting an existing tutorial, i.e., https://github.com/scikit-image/skimage-tutorials/blob/master/lectures/three_dimensional_image_processing.ipynb, into the gallery of examples. This addition falls under #4601 and makes use of the 3D dataset available with Pooch (namely, `cells.tif`).

I have only included the first two sections of the original tutorial:

* Load and display 3D images
* Adjust exposure (in experimental images)

I think that the segmentation of cells and nuclei could be its own tutorial, and so could feature extraction. Each example should be self-contained, but could nevertheless refer to another example (e.g., this introduction to 3D image processing where we only explore slices and adjust exposure).

If core devs prefer keeping everything in one place, I'm happy to keep going and append sections to the same file.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -250,9 +250,9 @@ gamma_high = exposure.adjust_gamma(data, gamma=gamma_high_val)
 
 _, ((a, b, c), (d, e, f)) = plt.subplots(nrows=2, ncols=3, figsize=(12, 8))
 
-show_plane(a, data[32], title="Original")
-show_plane(b, gamma_low[32], title="Gamma = {}".format(gamma_low_val))
-show_plane(c, gamma_high[32], title="Gamma = {}".format(gamma_high_val))
+show_plane(a, data[32], title='Original')
+show_plane(b, gamma_low[32], title=f'Gamma = {gamma_low_val}')
+show_plane(c, gamma_high[32], title=f'Gamma = {gamma_high_val}')
 
 plot_hist(d, data)
 plot_hist(e, gamma_low)
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -100,9 +100,9 @@ def show_plane(ax, plane, cmap="gray", title=None):
 (n_plane, n_row, n_col) = data.shape
 _, (a, b, c) = plt.subplots(ncols=3, figsize=(15, 5))
 
-show_plane(a, data[n_plane // 2], title="Plane = {}".format(n_plane // 2))
-show_plane(b, data[:, n_row // 2, :], title="Row = {}".format(n_row // 2))
-show_plane(c, data[:, :, n_col // 2], title="Column = {}".format(n_col // 2))
+show_plane(a, data[n_plane // 2], title=f'Plane = {n_plane // 2}')
+show_plane(b, data[:, n_row // 2, :], title=f'Row = {n_row // 2}')
+show_plane(c, data[:, :, n_col // 2], title=f'Column = {n_col // 2}')
 
 #####################################################################
 # As hinted before, a three-dimensional image can be viewed as a series of
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -267,7 +267,8 @@ plot_hist(f, gamma_high)
 # equalization <https://en.wikipedia.org/wiki/Histogram_equalization>`_
 # improves contrast in an image by redistributing pixel intensities. The most
 # common pixel intensities get spread out, increasing contrast in low-contrast
-# areas. One downside of this approach is that it may enhance background noise.
+# areas. One downside of this approach is that it may enhance background
+# noise.
 
 equalized_data = exposure.equalize_hist(data)
 
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -266,10 +266,16 @@ plot_hist(f, gamma_high)
 
 equalized_data = exposure.equalize_hist(data)
 
+display(equalized_data)
+
+#####################################################################
+# As before, if we have a Jupyter kernel running, we can explore the above
+# slices interactively.
+
 explore_slices(equalized_data);
 
 #####################################################################
-# Let us plot the image histogram before and after histogram equalization.
+# Let us now plot the image histogram before and after histogram equalization.
 # Below, we plot the respective cumulative distribution functions (CDF).
 
 _, ((a, b), (c, d)) = plt.subplots(nrows=2, ncols=2, figsize=(16, 8))
@@ -300,4 +306,4 @@ clipped_data = exposure.rescale_intensity(
     out_range=np.float32
 )
 
-explore_slices(clipped_data);
+display(clipped_data)
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -126,6 +126,11 @@ display(data)
 # Alternatively, we can explore these planes (slices) interactively using
 # Jupyter widgets. Let the user select which slice to display and show the
 # position of this slice in the 3D dataset.
+# Note that you cannot see the Jupyter widget at work in a static HTML page,
+# as is the case in the scikit-image gallery. For the following piece of
+# code to work, you need a Jupyter kernel running either locally or in the
+# cloud: see the bottom of this page to either download the Jupyter notebook
+# and run it on your computer, or open it directly in Binder.
 
 
 def slice_in_3D(ax, i):
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -248,6 +248,7 @@ show_plane(c, gamma_high[32], title="Gamma = {}".format(gamma_high_val))
 plot_hist(d, data)
 plot_hist(e, gamma_low)
 plot_hist(f, gamma_high)
+# sphinx_gallery_thumbnail_number = 4
 
 #####################################################################
 # `Histogram
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -84,7 +84,7 @@ except TypeError as e:
     print(str(e))
 
 #####################################################################
-# Function `io.imshow` can only display grayscale and RGB(A) 2D images.
+# The `io.imshow` function can only display grayscale and RGB(A) 2D images.
 # We can thus use it to visualize 2D planes. By fixing one axis, we can
 # observe three different views of the image.
 
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -224,7 +224,10 @@ explore_slices(data);
 # ===============
 # Scikit-image's `exposure` module contains a number of functions for
 # adjusting image contrast. These functions operate on pixel values.
-# Generally, image dimensionality or pixel spacing needn't be considered.
+# Generally, image dimensionality or pixel spacing doesn't need to be
+# considered. In local exposure correction, though, one might want to
+# adjust the window size to ensure equal size in *real* coordinates along
+# each axis.
 
 #####################################################################
 # `Gamma correction <https://en.wikipedia.org/wiki/Gamma_correction>`_
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -5,21 +5,23 @@ Explore 3D images (of cells)
 
 This tutorial is an introduction to three-dimensional image processing. Images
 are represented as `numpy` arrays. A single-channel, or grayscale, image is a
-2D matrix of pixel intensities of shape `(row, column)`. We can construct a 3D
-volume as a series of 2D planes, giving 3D images the shape
-`(plane, row, column)`. A multichannel, or RGB(A), image has an additional
+2D matrix of pixel intensities of shape ``(n_row, n_col)``, where ``n_row``
+(resp. ``n_col``) denotes the number of `rows` (resp. `columns`). We can
+construct a 3D volume as a series of 2D `planes`, giving 3D images the shape
+``(n_plane, n_row, n_col)``, where ``n_plane`` is the number of planes.
+A multichannel, or RGB(A), image has an additional
 `channel` dimension in the final position containing color information.
 
 These conventions are summarized in the table below:
 
-=============== ===============================
+=============== =================================
 Image type      Coordinates
-=============== ===============================
-2D grayscale    `(row, column)`
-2D multichannel `(row, column, channel)`
-3D grayscale    `(plane, row, column)`
-3D multichannel `(plane, row, column, channel)`
-=============== ===============================
+=============== =================================
+2D grayscale    ``[row, column]``
+2D multichannel ``[row, column, channel]``
+3D grayscale    ``[plane, row, column]``
+3D multichannel ``[plane, row, column, channel]``
+=============== =================================
 
 Some 3D images are constructed with equal resolution in each dimension (e.g.,
 synchrotron tomography or computer-generated rendering of a sphere).
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -182,6 +182,7 @@ def slice_in_3D(ax, i):
     ax.set_xlabel("plane")
     ax.set_ylabel("row")
     ax.set_zlabel("col")
+ax.set_xlim(0, 100)
 
     # Autoscale plot axes
     scaling = np.array([getattr(ax,
--- a/doc/examples/applications/plot_3d_image_processing.py
+++ b/doc/examples/applications/plot_3d_image_processing.py
@@ -42,7 +42,7 @@ import numpy as np
 from scipy import ndimage as ndi
 
 from skimage import (
-    feature, filters, io, measure, morphology, segmentation, util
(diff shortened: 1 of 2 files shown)


mkcor:
@emmanuelle @alexdesiqueira @jni Following up on our sprint conversations, please have a look when you get a chance. Thanks!

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,294 @@
...
+Image type      Coordinates
+=============== ===============================
+2D grayscale    `(row, column)`
+2D multichannel `(row, column, channel)`
+3D grayscale    `(plane, row, column)`
+3D multichannel `(plane, row, column, channel)`
+=============== ===============================
+
+Some 3D images are constructed with equal resolution in each dimension (e.g.,
+computer-generated rendering of a sphere). Most experimental data are captured
```
emmanuelle, line 25:
```suggestion
synchrotron tomography or computer-generated rendering of a sphere). Most experimental data are captured
```

emmanuelle: sorry to mess up with the 80 characters here :-)

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,294 @@
+"""
+============================
+Explore 3D images (of cells)
+============================
+
+This tutorial is an introduction to three-dimensional image processing. Images
+are represented as `numpy` arrays. A single-channel, or grayscale, image is a
+2D matrix of pixel intensities of shape `(row, column)`. We can construct a 3D
```
emmanuelle, line 8: since you are porting markdown to rst, there are several places like here where you need the text to be enclosed in double backticks instead of single ones.

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,294 @@
...
+
+print("microscope spacing: {}\n".format(original_spacing))
+print("rescaled spacing: {} (after downsampling)\n".format(rescaled_spacing))
+print("normalized spacing: {}\n".format(spacing))
+
+#####################################################################
+# Let us try and visualize the (3D) image with `io.imshow`.
+
+try:
+    io.imshow(data, cmap="gray")
```
emmanuelle, line 79: this empty figure is used for the gallery thumbnail, please select another figure as in https://github.com/matplotlib/matplotlib/issues/17970

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,294 @@
...
+        ax.set_xticks([])
+        ax.set_yticks([])
+
+
+display(data)
+
+#####################################################################
+# Alternatively, we can explore these planes (slices) interactively using
+# Jupyter widgets. Let the user select which slice to display and show the
+# position of this slice in the 3D dataset.
```
emmanuelle, line 126: please add a sentence here to say that you need a Jupyter kernel to run for this to work, so people can open the notebook on binder or on their computer but it won't work in the static html page.

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,294 @@
...
+    ax.imshow(plane, cmap=cmap)
+    ax.axis("off")
+
+    if title:
+        ax.set_title(title)
+
+
+_, (a, b, c) = plt.subplots(ncols=3, figsize=(15, 5))
+
+show_plane(a, data[32], title="Plane = 32")
```
emmanuelle, line 99: maybe write
```
(l_plane, l_row, l_col) = data.shape
```
and use `l_plane // 2` etc. below?

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,294 @@
...
+            facecolors="magenta",
+            linewidths=1,
+            edgecolors="black"
+        )
+    )
+
+    ax.set_xlabel("plane")
+    ax.set_ylabel("row")
+    ax.set_zlabel("col")
+
```
emmanuelle, line 185:
```suggestion
ax.set_xlim(0, 100)

```
so that one sees better how the slice moves through the parallelepiped.

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,294 @@
...
+#####################################################################
+# `Histogram
+# equalization <https://en.wikipedia.org/wiki/Histogram_equalization>`_
+# improves contrast in an image by redistributing pixel intensities. The most
+# common pixel intensities get spread out, letting areas of lower local
+# contrast gain a higher contrast. This may enhance background noise though.
+
+equalized_data = exposure.equalize_hist(data)
+
+explore_slices(equalized_data);
```
emmanuelle, line 260: could you `display` here instead of `explore_slices`? The figure will not show in the doc html page.

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,294 @@
...
+
+vmin, vmax = np.percentile(data, q=(0.5, 99.5))
+
+clipped_data = exposure.rescale_intensity(
+    data,
+    in_range=(vmin, vmax),
+    out_range=np.float32
+)
+
+explore_slices(clipped_data);
```
emmanuelle, line 294: and here too.

emmanuelle:
Thanks @mkcor ! I left a couple of comments. I think it's very useful to port the tutorials to the main doc, one reason being that they'll go through the traditional review process! For now I'm fine with having several tutorials instead of one very big one, as long as there is proper cross-referencing between them. See for example https://github.com/scikit-image/scikit-image/blob/master/doc/examples/applications/plot_human_mitosis.py#L50 for the syntax of such links. 

At some point maybe I'll contribute a plotly paragraph to show how to visualize these data in plotly (without the need for a Jupyter kernel :-)) but I don't have the time at the moment.

doc/examples/applications/plot_3d_image_processing.py
mkcor: Oh, I didn't notice because it rendered nicely (sort of):
backticks
now I realize this is italic, not code-style!

doc/examples/applications/plot_3d_image_processing.py
mkcor: Thanks, when building the docs I've only checked the page I was working on. I didn't think of checking the resulting gallery page. I'll add this small step to my mental workflow.

doc/examples/applications/plot_3d_image_processing.py
mkcor: it's ok, I'll fix it afterwards ;)

doc/examples/applications/plot_3d_image_processing.py
mkcor: > I'll add this small step to my mental workflow.

I thought I would add it to the contributor guide: #4857

doc/examples/applications/plot_3d_image_processing.py
mkcor: Good point, hard-coding isn't justified here. Done 9bb3123fa8. I have used notation `(n_plane, n_row, n_col)` though, it comes more natural to me.

doc/examples/applications/plot_3d_image_processing.py
mkcor: Done b5e6e4d327f. I hesitated quite a bit, because the 3rd figure is much more representative of 'exploring 3D images' but the 30 slices are way too tiny on a thumbnail... I thus chose the 4th figure, which is about adjusting exposure but it *looks* nice on a thumbnail. Plus it shows cells pretty clearly, which may catch the eye of a biologist or two :-)

doc/examples/applications/plot_3d_image_processing.py
emmanuelle: indeed :-).

emmanuelle: but are you ok changing the single backticks to double ones?

mkcor: I'm definitely ok, in the sense that I don't have the strongest opinion. But I have hesitations in terms of consistency and pedagogy: Words in code style should refer to words... in code. For example, "the shape `(plane, row, column)`" (lines 9--10) should match the code we have line 98:
```py
- (n_plane, n_row, n_col) = data.shape
+ (plane, row, column) = data.shape
```
There is some confusion because we are using "(plane, row, column)" to mean:
- the number of planes, rows, columns (shape);
- the dimensions (axes) as in line 34;
- the indices (coordinates) along the plane, row, column dimensions as in the summary table, line 16.
In this context, I think that using italic style for one of the meanings not found in code would make sense.

doc/examples/applications/plot_3d_image_processing.py
mkcor: Sure! Done 0e58bb1e -- did I even go into too many details?

doc/examples/applications/plot_3d_image_processing.py
mkcor: Done c1efdd404.

doc/examples/applications/plot_3d_image_processing.py
mkcor: Ok, I went for the following: e8c23f90. I hope it's not too heavy in terms of wording...
In the table, coordinates use the square-bracket notation, as they should in general and recalling/announcing code line 103--105.
Alternatively, I considered keeping the italic style to describe dimensions:
dimensions
which could also make sense.

mkcor:
> I think it's very useful to port the tutorials to the main doc, one reason being that they'll go through the traditional review process!

Absolutely!

> For now I'm fine with having several tutorials instead of one very big one, as long as there is proper cross-referencing between them.

Ok.

> See for example https://github.com/scikit-image/scikit-image/blob/master/doc/examples/applications/plot_human_mitosis.py#L50 for the syntax of such links.

Looks like it's not the link you meant to share :thinking:

> At some point maybe I'll contribute a plotly paragraph to show how to visualize these data in plotly (without the need for a Jupyter kernel :-)) but I don't have the time at the moment.

Oh, good idea!!

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,311 @@
...
+def show_plane(ax, plane, cmap="gray", title=None):
+    ax.imshow(plane, cmap=cmap)
+    ax.axis("off")
+
+    if title:
+        ax.set_title(title)
+
+
+(n_plane, n_row, n_col) = data.shape
+_, (a, b, c) = plt.subplots(ncols=3, figsize=(15, 5))
```
mkcor, line 101: I'm confused because the current [guidelines](https://scikit-image.org/docs/stable/contribute.html#guidelines) say "Examples in the gallery should have a maximum figure width of 8 inches" but we keep using widths of 15, 16 inches throughout examples (not only this one)... :thinking: @emmanuelle

doc/examples/applications/plot_3d_image_processing.py
jni: Looks great to me!

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,311 @@
...
+    )
+
+    ax.set_xlabel("plane")
+    ax.set_xlim(0, 100)
+    ax.set_ylabel("row")
+    ax.set_zlabel("col")
+
+    # Autoscale plot axes
+    scaling = np.array([getattr(ax,
+                                "get_{}lim".format(dim))() for dim in "xyz"])
```
jni, line 198:
```suggestion
                                f'get_{dim}lim')() for dim in "xyz"])
```

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+explore_slices(data);
+
+#####################################################################
+# Adjust exposure
+# ===============
+# Scikit-image's `exposure` module contains a number of functions for
+# adjusting image contrast. These functions operate on pixel values.
+# Generally, image dimensionality or pixel spacing needn't be considered.
```
jni, line 227:
```suggestion
# Generally, image dimensionality or pixel spacing doesn't need to be
# considered. In local exposure correction, though, one might want to
# adjust the window size to ensure equal size in *real* coordinates along
# each axis.
```

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,311 @@
...
+gamma_low = exposure.adjust_gamma(data, gamma=gamma_low_val)
+
+gamma_high_val = 1.5
+gamma_high = exposure.adjust_gamma(data, gamma=gamma_high_val)
+
+_, ((a, b, c), (d, e, f)) = plt.subplots(nrows=2, ncols=3, figsize=(12, 8))
+
+show_plane(a, data[32], title="Original")
+show_plane(b, gamma_low[32], title="Gamma = {}".format(gamma_low_val))
+show_plane(c, gamma_high[32], title="Gamma = {}".format(gamma_high_val))
```
jni, line 255:
```suggestion
show_plane(a, data[32], title='Original')
show_plane(b, gamma_low[32], title=f'Gamma = {gamma_low_val}')
show_plane(c, gamma_high[32], title=f'Gamma = {gamma_high_val}')
```

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,311 @@
...
+plot_hist(e, gamma_low)
+plot_hist(f, gamma_high)
+# sphinx_gallery_thumbnail_number = 4
+
+#####################################################################
+# `Histogram
+# equalization <https://en.wikipedia.org/wiki/Histogram_equalization>`_
+# improves contrast in an image by redistributing pixel intensities. The most
+# common pixel intensities get spread out, letting areas of lower local
+# contrast gain a higher contrast. This may enhance background noise though.
```
jni, line 267:
```suggestion
# common pixel intensities get spread out, increasing contrast in low-contrast
# areas. One downside of this approach is that it may enhance background noise.
```

doc/examples/applications/plot_3d_image_processing.py
```diff
@@ -0,0 +1,311 @@
...
+d.set_title("Histogram equalization CDF")
+
+#####################################################################
+# Most experimental images are affected by salt and pepper noise. A few bright
+# artifacts can decrease the relative intensity of the pixels of interest. A
+# simple way to improve contrast is to clip the pixel values on the lowest and
+# highest extremes. Clipping the darkest and brightest 0.5% of pixels will
+# increase the overall contrast of the image.
+
+vmin, vmax = np.percentile(data, q=(0.5, 99.5))
```
jni, line 303: Personally, I prefer using quantiles, though it's a personal preference, consider this comment optional. =)

```suggestion
vmin, vmax = np.quantile(data, q=(0.005, 0.995))
```

doc/examples/applications/plot_3d_image_processing.py
emmanuelle: I have no idea of where this 8 inches come from :-).

emmanuelle: The rule of thumb should be to always take a look at the generated html...

mkcor: > The rule of thumb should be to always take a look at the generated html...

Sure. So maybe this line
https://github.com/scikit-image/scikit-image/blob/ebb53a67f255129c796b21138d248b398199e108/CONTRIBUTING.txt#L418
could be removed or edited.

doc/examples/applications/plot_3d_image_processing.py
mkcor: @jni style question: I learnt Python using single quotes by default, resorting to double quotes only when necessary. Then `black` influenced me into double quotes... Do you have a strong leaning?

doc/examples/applications/plot_3d_image_processing.py
mkcor: Ok, I just wanted to be a little posh... :laughing:

doc/examples/applications/plot_3d_image_processing.py
mkcor: I went for a popularity contest to decide and percentile won. ;-)
I searched for `"cell biology" "percentile"` vs `"cell biology" "quantile"`.

doc/examples/applications/plot_3d_image_processing.py
jni: Yes, my strong leaning is that [black is the devil](https://github.com/napari/napari/pull/1451#issuecomment-658002959), and your history reinforces that opinion. =P

More seriously: I feel very strongly that the original PEP8 had it right (single quotes by default), and black really made the wrong decision here, as with their default line length. They essentially had to admit as such, adding config options for both of these in a package generally devoid of config options.

In scikit-image, when in conflict, follow PEP8, not black.

I *will* also say that we have not actually enforced string formatting at all. I only changed these because I was changing them to f-strings anyway. I wouldn't stress about going back and changing existing stuff.

mkcor: Ok, thanks for sharing! =)

Yes, yes, I got that the suggested change was about f-strings and not so much about quotes. But this side 'issue' had been on my mind, so I'm happy to be more informed now. Thanks.

doc/examples/applications/plot_3d_image_processing.py
emmanuelle: I also did not know that numpy had a quantile function :-). (but I'm happy that I learned this, since I also prefer a bit quantile to percentile, but I'm fine with letting the pop contest decide.)

emmanuelle:
Merging, thanks a lot @mkcor!

PR #4852, opened 2020-07-17
Build and deploy dev/master documentation using GitHub Actions

### LINKED ISSUE #4805: dev doc is not deployed when pre build is failing on Travis

emmanuelle:
Closes #4805.
Follow-up of #4843 but with a branch in the scikit-image repo so that we can test the workflow.


emmanuelle:
Argh the GH action keeps failing with "Error loading key "(stdin)": invalid format". I've been trying to add the private key in the sciki-image/scikit-image secrets with and without the "begin" and "end" lines, with a new line... Is this the same error which you were having @jvkersch or another one (in which case we would undoubtedly be making progress :-))?

emmanuelle:
Lo and behold, we've tested the github action with @jvkersch (with commit d172b74 which used this branch instead of master to see if the deploy worked), and it works (you can see in the online dev doc that the mitosis tutorial by @mkcor is online now)! For future reference, the authentication problems were due to the fact so that you need a specific formatting of RSA key which did not correspond to the default of `ssh-keygen` on my Ubuntu machine. The successful command to generate the key pair was 
```
 ssh-keygen -m PEM -t rsa -b 4096 -C "your_email@example.com"
```
Note that this PR is using a branch pushed to the scikit-image repo, which we don't do usually, so we'll have to make sure to delete that branch.

@jvkersch could you please check that my last commit (switching back to master for the branch) is ok? And then @skimage/core is think this one is ready!

jvkersch:
@emmanuelle Looks good! 

One suggestion, if you're up for it, is to rename the `secrets.GH_TOKEN` into something more descriptive like `secrets.DEPLOY_KEY_PRIVATE`. That would help future generations of developers with troubleshooting ;)

emmanuelle:
> One suggestion, if you're up for it, is to rename the `secrets.GH_TOKEN` into something more descriptive like `secrets.DEPLOY_KEY_PRIVATE`. That would help future generations of developers with troubleshooting ;)

good idea :-). Done!

emmanuelle:
@skimage/core anything else I should do before merging this one?

sciunto:
@scikit-image/core  anything else I should do before merging this one? [former ping by @emmanuelle didn't work)

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,51 @@
...
+  build:
+    name: Build docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
```
sciunto, line 18: Just a question, what happens if python 3.7 is not shipped in ubuntu-lastest (which will be true a day or another)?

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,51 @@
...
+          python -m pip install -r requirements/docs.txt
+      - name: Install package
+        run: |
+          python setup.py install
+      - name: Push docs
+        run: |
+          eval "$(ssh-agent -s)"
+          ssh-add - <<< "${DEPLOY_KEY}"
+          git config --global user.email "travis@travis-ci.com"
+          git config --global user.name "Travis Bot"
```
sciunto, line 34: Shouldn't be attributed to GH-action instead? (Or am I lost?)

.github/workflows/build-docs.yml
sciunto: I just want to know which indicator we are supposed to monitor...

emmanuelle: the build fails :-). And then we change the config. But it would stop early enough not to push a wrong version of the doc, if this is your concern.

.github/workflows/build-docs.yml
jni: I agree!

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements/build.txt
+          python -m pip install -r requirements/default.txt
+          python -m pip install -r requirements/optional.txt
+          python -m pip install -r requirements/docs.txt
+      - name: Install package
+        run: |
+          python setup.py install
```
jni, line 28: Could we use `python -m pip install .` here? That will future-proof us for when we switch to pyproject.toml etc.

jni:
@emmanuelle this is so cool! Have you explored whether there is a way to make the docs visible on each PR, though?

.github/workflows/build-docs.yml
emmanuelle: ok! done.

.github/workflows/build-docs.yml
emmanuelle: done!

emmanuelle:
@sciunto @jni I made the requested changes, then I temporarily checked the action by having it run on this branch (it works, as you can see from https://github.com/scikit-image/docs/tree/gh-pages), and then I switched back to master in the last commit. I hope the PR is ready now! Don't worry about the CircleCI failure, the config file is in #4881

PR #4861, opened 2020-07-28
Update min req for pillow due to CVE-2020-10379 and co.

sciunto:
## Description

See https://github.com/scikit-image/scikit-image/network/alert/requirements/default.txt/Pillow/open

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


rfezzani:
I agree with this change, I approve and wait few times before merge if anyone have objection ;)

hmaarrfk:
I think we are quick to forget how hard it is for us to bump our own dependencies.

Is this really a dependency we want to impose on others?

Security depends on your threat model. What if users decide they are not beholden to this vulnerability?
There will always be a new security vulnerability. We should patch them, but maybe not impose them?

soupault:
I am also slightly against bumping the dependency here. Pillow 7.1.2 was released in April 2020. Since it is a mandatory dependency, such agressive bumping can lead to some Python distribution channels (e.g. Ubuntu 16.04/18.04) being unable to catch up. Of course, the recommended channels are pypi and conda, but this is something we should consider while making a decision here. On the other hand, out-of-bounds access is a severe threat.

The best thing would probably be to wait for the Pillow team to respond and backport the fix. It shows to me that the recommended version now is 7.1.0 - https://github.com/scikit-image/scikit-image/network/alert/requirements/default.txt/Pillow/open 🤔 , so they are, apparently, taking some actions.

sciunto:
Please, note that we blacklisted 0.7.0 and 0.7.1.

sciunto:
> I am also slightly against bumping the dependency here. Pillow 7.1.2 was released in April 2020. Since it is a mandatory dependency, such agressive bumping can lead to some Python distribution channels (e.g. Ubuntu 16.04/18.04) being unable to catch up. 

I'm not sure to understand this argument. If we peak about pillow and skimage installed from package manager, either the maintainers update/fix pillow regarding the security issue and the problem does not exist any more, or they disregard, and nothing will change.

To me, the purpose of this patch is to target users who install scikit-image in virtual env to make sure they do not use an outdated version. These users have the right to expect that if a security issue is known and fixes, updating scikit-image will include the fix.

stefanv:
I don't think we need to worry about Debian/Ubuntu; those packagers are clever enough to know that their (probably patched) version of pillow can be used instead of our requirement.

emmanuelle:
> I don't think we need to worry about Debian/Ubuntu; those packagers are clever enough to know that their (probably patched) version of pillow can be used instead of our requirement.

Should we ask @yarikoptic for a confirmation here?

@sciunto, you're a packager for Arch Linux if I remember well, would this work also for Arch?

sciunto:
I'm not active anymore, but it wouldn't be a problem.

sciunto:
To summarize, on the "pro" side, we have @lagru @rfezzani @stefanv and myself. On the hesitant side, we have @hmaarrfk @soupault Perhaps other @scikit-image/core members want to chime in?

emmanuelle:
From https://pypi.org/project/Pillow/#history, there is little chance that the Pillow team will backport the fix to a pre-7.x version. So we are left with two options:
- bumping the dep
- just issuing a warning advising users to upgrade their Pillow if they can, when Pillow is imported and corresponds to a version older than 7.1.2. 
Out-of-bounds is bad but github labels the issue as "moderate severity" so I'd be in favor of option 2.

sciunto:
Option 2 works for me too.

jni:
In private conversations, @stefanv told me he had become a little more hesitant since his post. So I think we should go for option 2.

Thank you @sciunto for your patience! Do you want to update the PR? I personally think we should warn upon importing of `skimage.io`.

sciunto:
I hope to have to time next week. It's on my todo list :) I also prefer this strategy now.

sciunto:
Here is the new way of handling this guy ;)

skimage/io/collection.py
```diff
@@ -10,6 +10,17 @@
...
 
+# Check CVE-2020-10379
+from distutils.version import LooseVersion
+if LooseVersion(PIL.__version__) < LooseVersion('7.1.0'):
+    from warnings import warn
+    warn('Your installed pillow version is < 7.1.0. '
+         'Several security issues (CVE-2020-11538, '
+         'CVE-2020-10379, CVE-2020-10994, CVE-2020-10177) '
+         'have been fixed in pillow 7.1.0 or higher. '
+         'We recommand to upgrade this library.',
```
lagru, line 21:
```suggestion
         'We recommend to upgrade this library.',
```

skimage/io/_plugins/pil_plugin.py
```diff
@@ -7,6 +7,17 @@
...
 
+# Check CVE-2020-10379
+from distutils.version import LooseVersion
+if LooseVersion(pil_version) < LooseVersion('7.1.0'):
+    from warnings import warn
+    warn('Your installed pillow version is < 7.1.0. '
+         'Several security issues (CVE-2020-11538, '
+         'CVE-2020-10379, CVE-2020-10994, CVE-2020-10177) '
+         'have been fixed in pillow 7.1.0 or higher. '
+         'We recommand to upgrade this library.',
```
lagru, line 18:
```suggestion
         'We recommend to upgrade this library.',
```

skimage/io/_plugins/pil_plugin.py
```diff
@@ -7,6 +7,17 @@
...
+         stacklevel=2)
```
lagru, line 19: This warning is not called inside a function but on the module level. Is `stacklevel=2` still correct?

hmaarrfk: I don't think stack level does anything at the import level. Pretty annoying. 

But since the resolution is at system level, not code level, I think it is ok.

PR #4862, opened 2020-07-28
Add default level parameter in measure.find_contours

### LINKED ISSUE #4813: Make level parameter optional in measure.find_contours
## Description

I think that we could put a default value to level in find_contours. I would suggest (image.max() - image.min())/2. This would make this function able to be used with an image and I think this default value is meaningful.

I'm open to discussion.

rubywerman:
## Description
Fixes #4813
This PR adds the median as the default level parameter in `find_contours`. After reading the discussion in #4813 and looking at different image arrays [see [notebook](https://github.com/rubywerman/notebooks/blob/master/plot_contours.ipynb)], I thought the median could be a good default value. 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -25,7 +25,7 @@ def find_contours(image, level=None,
         Input image in which to find contours.
     level : float
         Value along which to find contours in the array. By default, the level
-        is set to the median value of the array.
+        is set to (max(image) - min(image)) / 2
     fully_connected : str, {'low', 'high'}
          Indicates whether array elements below the given level value are to be
          considered fully-connected (and hence elements above the value will
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -9,7 +9,7 @@ _param_options = ('high', 'low')
 
 
 @deprecate_kwarg({'array': 'image'}, removed_version="0.20")
-def find_contours(image, level='median',
+def find_contours(image, level=None,
                   fully_connected='low', positive_orientation='low',
                   *,
                   mask=None):
@@ -140,7 +140,7 @@ def find_contours(image, level='median',
         if not np.can_cast(mask.dtype, bool, casting='safe'):
             raise TypeError('Parameter "mask" must be a binary array.')
         mask = mask.astype(np.uint8, copy=False)
-    if level == 'median':
+    if level is None:
         level = np.median(image)
     segments = _get_contour_segments(image.astype(np.double), float(level),
                                      fully_connected == 'high', mask=mask)
--- a/skimage/measure/_find_contours.py
+++ b/skimage/measure/_find_contours.py
@@ -23,7 +23,7 @@ def find_contours(image, level=None,
     ----------
     image : 2D ndarray of double
         Input image in which to find contours.
-    level : float
+    level : float, optional
         Value along which to find contours in the array. By default, the level
         is set to (max(image) - min(image)) / 2
     fully_connected : str, {'low', 'high'}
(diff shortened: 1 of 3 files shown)


skimage/measure/_find_contours.py
```diff
@@ -9,7 +9,7 @@
 
 
 @deprecate_kwarg({'array': 'image'}, removed_version="0.20")
-def find_contours(image, level,
+def find_contours(image, level='median',
```
rfezzani, line 12: Setting the default to a string may suggest that other string values are accepted ('mean' for example)
```suggestion
def find_contours(image, level=None,
```

rubywerman:
Thank you for your feedback @rfezzani! I updated the parameter as you suggested :)

emmanuelle:
Thanks for the PR @rubywerman! The discussion about which default level to choose was not completely settled in #4813, there is also the possibility in taking (max - min)/2. The latter choice is mostly useful for binary arrays while the median makes more sense for float arrays. I'd argue that binary arrays are the typical case where one does not want to specify the value of the level, because there is really only one value which makes sense. So I have a slight preference for (max - min)/2 but the median also makes sense.

Please also add a test which executes `find_contours` without passing the level value, so that the new API is tested.

rubywerman:
Thanks for the feedback @emmanuelle! I will add the test. Also, could it be beneficial to have two default values, depending on whether it's a binary or float array?

jni:
I agree with @sciunto that having different behaviour based on the input type isn't ideal. I have commented on the original issue, but my suggestion would be to have `np.mean(np.quantile(image, (0.005, 0.995)))` as the default. Thoughts?

emmanuelle:
I have reconsidered my initial doubts. Given the fact that there is the possibility to pass a value, and that for a binary image other contours than 0.5 (in fact anything strictly between 0 and 1) will also good give results, we have to accept that there are several possible good solutions and just choose one of them. The one you suggest @jni is a good one, in fact the original median value proposed by @rubywerman is also good (and simple to explain).

skimage/measure/tests/test_find_contours.py
```diff
@@ -130,3 +130,40 @@ def test_invalid_input():
         find_contours(r, 0.5, 'foo', 'bar')
     with testing.raises(ValueError):
         find_contours(r[..., None], 0.5)
+
```
emmanuelle, line 133: thanks for adding all these tests :-)

skimage/measure/_find_contours.py
```diff
@@ -24,7 +24,8 @@ def find_contours(image, level,
     image : 2D ndarray of double
         Input image in which to find contours.
     level : float
```
sciunto, line 26:
```suggestion
    level : float, optional
```

skimage/morphology/convex_hull.py
```diff
@@ -128,8 +128,8 @@ def convex_hull_object(image, *, connectivity=2):
     objects may overlap in the result. If this is suspected, consider using
     convex_hull_image separately on each object or adjust ``connectivity``.
     """
-    if image.ndim > 2:
-        raise ValueError("Input must be a 2D image")
+    if image.ndim > 3:
+        raise ValueError("Input must be a 2D or 3D image")
```
sciunto, line 132: Why this change in this PR?

rubywerman: my mistake– it shouldn't. fixing now

sciunto: Thank you!

sciunto:
I approve it as well, I'm just waiting for the CI. Thank you for your contribution.

skimage/morphology/convex_hull.py
emmanuelle: I thought @rubywerman was fixing #4865 at the same time :-) (but it can be in another PR, with a test ;-))

emmanuelle:
Merging, thank you @rubywerman !

Issue #4537, opened 2020-03-30
Feature request: mesh decimation algorithm


emmanuelle:
As discussed on https://forum.image.sc/t/marching-cubes-element-size/35656/ it would be a nice addition to have post-processing tools for meshes generated by the marching cube algorithms. 

The VTK algorithm (Decimate Pro filter) is based on http://marchingcubes.org/images/4/41/DecimationOfTriangleMeshes.pdf. 

In the meantime we could also link to [this Mayavi example](https://docs.enthought.com/mayavi/mayavi/auto/example_julia_set_decimation.html#example-julia-set-decimation) in the marching cubes docstring.

clementkng:
@emmanuelle Hi, I'd be interested in trying this! I'm currently looking through the links and papers above, and had a couple of high level questions:

1. What would the API for introducing mesh decimation look like? In the image.sc post above, it seems like the mesh decimation could be run as part of the `marching_cubes_lewiner` method, so I think one possibility would be to pass this in as an optional argument, which would have termination conditions and other info necessary for mesh decimation. How does that sound?

2. Can we use the vtkDecimatePro code directly? Or is it supposed to simply be an inspiration for our implementation?

3. Should I start off by linking the Mayavi example so scikit-image users know that some mesh decimation algorithm is coming?

emmanuelle:
Hi @clementkng this is awesome! 

1. The API should become clearer when implementing the filter I think. At the moment I don't know how many parameters should be exposed from the vtk filter, if it's just the target reduction fraction or if we need other parameters (the Mayavi example uses a `feature_angle` parameter, I'm not sure of what this means). We could start with having just one single `target_reduction` parameter in the `marching_cubes` function, because it's easier to enrich an API later on than to deprecate exposed parameters. 
I also don't think that we should expose a standalone function for mesh reduction because this falls outside the scope of scikit-image, but this can be subject to debate.

 2. I don't think we can reuse directly the vtk code because i) it's C++ code and ii) it's hardly meant to be read by humans, so it does not satisfy our criteria of only Python/Cython and easily maintainable code. So we would need a full reimplementation, probably in Cython. I believe this represents a fair amount of work :-).

3. A small PR with a link to the Mayavi example in the `marching_cubes` docstring would be very welcome, because I think it will take some time to implement the decimation algorithm.

clementkng:
[...]

PR #4872, opened 2020-07-31
DOC: Clarify condition on unique vertices returned by marching cubes

clementkng:
## Description

`marching_cubes` can actually return non unique vertices if the `allow_degenerate` argument is set to `True`. This is because if `allow_degenerate` is set to `True`, then degenerate triangles are not removed from the mesh, such as zero-area triangles. These triangles have an effect of adding duplicate vertices to the outputted `verts` array. I've added a clarification in the documentation so people won't be confused seeing non unique vertices in their output.

Addresses comments in #4537.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_marching_cubes_classic.py
+++ b/skimage/measure/_marching_cubes_classic.py
@@ -33,7 +33,9 @@ def marching_cubes_classic(volume, level=None, spacing=(1., 1., 1.),
     -------
     verts : (V, 3) array
         Spatial coordinates for V unique mesh vertices. Coordinate order
-        matches input `volume` (M, N, P).
+        matches input `volume` (M, N, P). If ``allow_degenerate`` is set to
+        True, then the presence of degenerate triangles in the mesh can make
+        this array appear to have duplicate vertices.
     faces : (F, 3) array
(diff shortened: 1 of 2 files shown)


skimage/measure/_marching_cubes_classic.py
```diff
@@ -33,7 +33,9 @@ def marching_cubes_classic(volume, level=None, spacing=(1., 1., 1.),
     -------
     verts : (V, 3) array
         Spatial coordinates for V unique mesh vertices. Coordinate order
-        matches input `volume` (M, N, P).
+        matches input `volume` (M, N, P). If ``allow_degenerate`` is set to
+        True, then the presence of degenerate triangles in the mesh can make
+        this array appear to have duplicate vertices.
```
jni, line 38: Appear to have, or have?

clementkng: "have", good catch. This is because V simply refers to the number of output vertices, and there are duplicates in the V vertices outputted. Setting `allow_degenerate` to False would reduce the `verts` list to only have unique vertices.

PR #4873, opened 2020-07-31
Remove unmaintained wiki page link in contributor guidelines

sciunto:
## Description

CC @emmanuelle 

looks like it is not the best idea to put this wiki page on the main entry point for new contributors.


jni:
Thank you @sciunto, nicely spotted. =)

PR #4877, opened 2020-07-31
Use matplotlib's infinite axline to demonstrate hough transform

stefanv:
This requires matplotlib 3.3.

It does prevent problems when `np.sin(angle)` is zero.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,6 +1,6 @@
 numpy>=1.16.5
 scipy>=1.2.3
-matplotlib>=3.0.3
+matplotlib>=3.3
 networkx>=2.2
 pillow>=5.4.0,!=7.1.0,!=7.1.1
 imageio>=2.4.1
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,6 +1,6 @@
 numpy>=1.16.5
 scipy>=1.2.3
-matplotlib>=3.3
+matplotlib>=3.0.3
 networkx>=2.2
 pillow>=5.4.0,!=7.1.0,!=7.1.1
 imageio>=2.4.1
--- a/doc/examples/edges/plot_line_hough_transform.py
+++ b/doc/examples/edges/plot_line_hough_transform.py
@@ -104,14 +104,13 @@ ax[1].set_ylabel('Distance (pixels)')
 ax[1].axis('image')
 
 ax[2].imshow(image, cmap=cm.gray)
-origin = np.array((0, image.shape[1]))
-for _, angle, dist in zip(*hough_line_peaks(h, theta, d)):
-    y0, y1 = (dist - origin * np.cos(angle)) / np.sin(angle)
-    ax[2].plot(origin, (y0, y1), '-r')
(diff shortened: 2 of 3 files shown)


grlee77:
I think this actually needs Matplotib >= 3.3!. When I tried with 3.1.1 I got `AttributeError: 'AxesSubplot' object has no attribute 'axline'`

There are `axhline` and `axvline` in 3.0, but not `axline`.

grlee77:
Matplotlib 3.0 is nearly two years old, so that seems reasonable to require. Matplotlib 3.3, though, was just released 15 days ago! Do you think it is worth having two code paths in the example until we make 3.3 a requirement or should we hold off on this change for now?

tacaswell:
Yeah, this is a 3.3 feature.

stefanv:
Woops, meant to write 3.3 ;)

stefanv:
I think using Arch linux has messed with me.

What?! Not running latest? Get at it!

hmaarrfk:
Can we change the dependency for the docs instead?
https://github.com/scikit-image/scikit-image/blob/master/requirements/docs.txt

We already require a newer version of matplotlib for docs than for the build.

emmanuelle:
I agree with Mark here: if we bump a dependency for a version which is only 2 weeks old, it should be the docs dependencies, not the package ones.

stefanv:
Sure, that's fine. But I may hold off until 3.3 is widely available to users, otherwise this won't run for most readers.

requirements/default.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.15.1
 scipy>=1.0.1
-matplotlib>=2.0.0,!=3.0.0
+matplotlib>=3.3,!=3.0.0
```
sciunto, line 3:
```suggestion
matplotlib>=3.3
```

madphysicist:
FWIW, I'm all for this change.

requirements/default.txt
sciunto: Would you accept this change and would you like to rebase @stefanv ?

I'm OK to bump for a 6 months old version.

sciunto (approved):

jni:
I'm happy to merge ASAP if we change this to happen with the docs requirement rather than the default ones.

stefanv:
@rfezzani I was wondering if you'd mind taking a look at this one.

stefanv:
OK, should be good to go.

requirements/default.txt
```diff
@@ -1,6 +1,6 @@
 numpy>=1.16.5
 scipy>=1.2.3
-matplotlib>=3.0.3
+matplotlib>=3.0.1
```
rfezzani, line 3: Is this change intentional?

rfezzani (approved):
I just wonder if you meant matplotlib version decrease, otherwise I am :+1: for the introduced changes. Thank you @stefanv

requirements/default.txt
stefanv: Woops, no, thanks for catching that; fixed.

PR #4879, opened 2020-08-01
new matomo config

emmanuelle:
Our free trial account had expired, I created a new one.

@jni @grlee77 if this PR is merged quickly we can have the stats for Monday before the grant deadline :-)


emmanuelle:
The code block is what Matomo asks to paste.

Issue #4083, opened 2019-08-05
Explore performance of parallel Richardson-Lucy deconvolution


jni:
## Description

In [this thread](https://github.com/dask/dask-blog/issues/27#issuecomment-518020228), @mrocklin notes that scikit-image's Richardson-Lucy implementation is significantly slower than ITK's when run in parallel over many blocks. We should explore:

- why that is
- whether the [performance heuristics](https://github.com/scikit-image/scikit-image/blob/3ee0dbae978df083252ad168b5986a6f9f19aae6/skimage/restoration/deconvolution.py#L370-L377) about whether to switch to FFTs vs direct convolution needs to be updated/changed/removed for parallel workflows or for newer dependency versions. (I think ndi.convolve has its own internal logic to do this!)
- whether it can be improved by e.g. using gputools FFT and/or convolution when that library and a GPU (dedicated or integrated) is available.

In the past @hmaarrfk has done really well at figuring this stuff out. ;)

mrocklin:
@jni asked if I had code to help test performance here.  Rather than use Dask, I would recommend using a simple thread pool, just to keep things isolated.  The approach I recommend is the same as in this notebook: https://nbviewer.jupyter.org/gist/mrocklin/ec745d6c2a12dddddb125ef460a4da76#So-how-well-does-skimage.filters.gaussian-parallelize-normally?

grlee77:
One way to accelerate the existing function in SciPy 1.4+ (at least for cases where FFT-based convolution get selected) is to enable multi-threaded FFTs in SciPy):

Here is a concrete example:

```Python
from time import time
from skimage import color, data, restoration
from scipy.signal import convolve2d
camera = data.camera() / 255.

# create test data and psf
camera = np.tile(camera, (8, 8))
psf = np.ones((8, 8), dtype=np.float32) / 64
camera = convolve2d(camera, psf, 'same')
camera += 0.05 * camera.std() * np.random.standard_normal(camera.shape)

iterations = 10

# default, single-threaded FFTs
tstart = time()
deconvolved = restoration.richardson_lucy(camera, psf, iterations)
dur = time() - tstart
print(f"duration (default FFT) = {dur:0.4} s")

# workers=-1 -> use max. # of threads
tstart = time()
with scipy.fft.set_workers(-1):
    deconvolved = restoration.richardson_lucy(camera, psf, iterations)
dur = time() - tstart
print(f"duration (threaded FFT) = {dur:0.4} s")

[...]

PR #4880, opened 2020-08-02
Richardson-Lucy deconvolution: allow single-precision computation

grlee77:
## Description

This PR improves the performance of deconvolution by keeping the computations in single-precision when the input image is single-precision.

It also avoids unnecessary copies of the input arrays by adding `copy=False` to existing `.astype()` calls.

This can give up to a factor 2 performance improvement when calling Richardson Lucy either single-threaded or in various multi-threaded scenarios.  See the timings for double-precision in https://github.com/scikit-image/scikit-image/issues/4083#issuecomment-667611962 vs. single-precision in https://github.com/scikit-image/scikit-image/issues/4083#issuecomment-667612287.

If this seems reasonable to others, I can also adapt the example from the comments into an ASV benchmark.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/restoration/tests/test_restoration.py
+++ b/skimage/restoration/tests/test_restoration.py
@@ -1,5 +1,3 @@
-import itertools
-
 import numpy as np
 import pytest
 from scipy.signal import convolve2d
@@ -91,13 +89,8 @@ def test_richardson_lucy():
     np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3)
 
 
-@pytest.mark.parametrize(
-    'dtype_image, dtype_psf',
-    itertools.product(
-        [np.float32, np.float64],
-        [np.float32, np.float64],
-    )
-)
+@pytest.mark.parametrize('dtype_image', [np.float32, np.float64])
+@pytest.mark.parametrize('dtype_psf', [np.float32, np.float64])
 def test_richardson_lucy_filtered(dtype_image, dtype_psf):
     if dtype_image == np.float64:
         atol = 1e-8
--- a/skimage/restoration/tests/test_restoration.py
+++ b/skimage/restoration/tests/test_restoration.py
@@ -1,4 +1,7 @@
+import itertools
+
 import numpy as np
+import pytest
 from scipy.signal import convolve2d
 from scipy import ndimage as ndi
 from skimage._shared.testing import fetch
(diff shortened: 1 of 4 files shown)


skimage/restoration/tests/test_restoration.py
```diff
@@ -88,17 +91,31 @@ def test_richardson_lucy():
     np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3)
 
 
-def test_richardson_lucy_filtered():
+@pytest.mark.parametrize(
```
emmanuelle, line 94: nitpicking here, but I believe
```
@pytest.mark.parametrize('dtype_image', [np.float32, np.float64])
@pytest.mark.parametrize('dtype_psf, [np.float32, np.float64])
def test_richardson_lucy_filtered(dtype_image, dtype_psf):
```
does exactly this, and is a bit more readable

emmanuelle:
Thanks for the PR! An ASV benchmark would be great if you find the time :-).

skimage/restoration/tests/test_restoration.py
grlee77: Yeah, I got used to the other style from the CuPy test suite, but agree that the above is simpler!

grlee77:
I added some simple ASV benchmarks for `richardson_lucy`. There is also a separate commit that fixed the existing restoration benchmarks, which were currently failing to run due to a misnamed variable!

As expected, float32 operation is now faster. This PR also reduced peak memory use a bit in the float64 case by avoiding internal copies.

## Result of benchmarks for release 0.17.2
```
[  0.00%] ·· Benchmarking existing-py_home_lee8rx_miniconda3_envs_pyir_bin_python3.7
[ 40.00%] ··· Running (benchmark_restoration.DeconvolutionSuite.time_denoise_nl_means_f32--)..
[ 60.00%] ··· benchmark_restoration.DeconvolutionSuite.peakmem_denoise_nl_means_f32                       293M
[ 70.00%] ··· benchmark_restoration.DeconvolutionSuite.peakmem_richardson_lucy_f64                        293M
[ 80.00%] ··· benchmark_restoration.DeconvolutionSuite.peakmem_setup                                      124M
[ 90.00%] ··· benchmark_restoration.DeconvolutionSuite.time_denoise_nl_means_f32                    2.32±0.01s
[100.00%] ··· benchmark_restoration.DeconvolutionSuite.time_richardson_lucy_f64                     2.33±0.02s
```

## Result of benchmarks for this PR
```
[  0.00%] ·· Benchmarking existing-py_home_lee8rx_miniconda3_envs_pyir_bin_python3.7
[ 40.00%] ··· Running (benchmark_restoration.DeconvolutionSuite.time_denoise_nl_means_f32--)..
[ 60.00%] ··· benchmark_restoration.DeconvolutionSuite.peakmem_denoise_nl_means_f32                       200M
[ 70.00%] ··· benchmark_restoration.DeconvolutionSuite.peakmem_richardson_lucy_f64                        276M
[ 80.00%] ··· benchmark_restoration.DeconvolutionSuite.peakmem_setup                                      124M
[ 90.00%] ··· benchmark_restoration.DeconvolutionSuite.time_denoise_nl_means_f32                    1.64±0.01s
[100.00%] ··· benchmark_restoration.DeconvolutionSuite.time_richardson_lucy_f64                     2.30±0.01s
```

grlee77:
For the benchmarks, I ended up making a new `DeconvolutionSuite` class since the `setup` method is different to the one in the existing `RestorationSuite` class (need to store a PDF and convolve it with the input image).

grlee77:
I think maybe the doc failure might be because I haven't rebased on master since that PR was merged?

jni:
@grlee77 why is time RL f32 not shown in your benchmarks result?

grlee77:
> @grlee77 why is time RL f32 not shown in your benchmarks result?

It is there, but was misnamed as "time_denoise_nl_means_f32" due to a copy paste error that was later fixed in bf38f4c.

I will edit the comment above to reflect the correct naming.

jni:
@grlee77 this is then ready after that fix and a rebase/merge on master!

alexdesiqueira:
Thank you @grlee77!

PR #4881, opened 2020-08-02
build doc on circle ci and link artifact

emmanuelle:
The purpose of this PR is to have circleCI build the doc so that reviewers can check the output of the doc build without having to checkout the pr branch.

~~work in progress...~~


emmanuelle:
The circleCI part works well, you can check the artifacts tab in circleCI to check that you can access the doc. I was trying to set up a github action to link directly the index.html from the PR but it does not seem to work. @glemaitre since I stole your config file from the covid dashboard maybe you would know why :-). Anyway, we might be happy with just the artifacts tab of circle CI.

glemaitre:
From what I remember, you need to add a file to mention where the changed will be stored: https://github.com/scikit-learn/scikit-learn/pull/14731/files

Would this be the problem?

emmanuelle:
Thanks @glemaitre. The scikit-learn repo does not use a Github action to display the artifact link, so probably it's not needed to add a file in `.github/workflows`? I tried adding an `artifact_path` file with the path of the index of the doc but it does not seem to change what appears in the list of checks. I'll keep on investigating, thanks for your help!

emmanuelle:
I registererd the app https://github.com/apps/circleci-artifacts-redirector/ to `scikit-image/scikit-image`, maybe this will help.

emmanuelle:
ok it works now. I had overlooked that the artifact registration app only listened to jobs with specific names.

sciunto:
awesome, let's merge this! Thank you @emmanuelle !

soupault:
Wow, this is great! Thank you @emmanuelle !

PR #4886, opened 2020-08-05
Raise ValueError if NaN found in phase_cross_correlation result

GenevieveBuckley:
## Description

Closes https://github.com/scikit-image/scikit-image/issues/4763

As requested by @emmanuelle in https://github.com/scikit-image/scikit-image/pull/4767#issuecomment-668082059

This PR raises a `ValueError` if a NaN value is found in the error or phase_diff, indicating that there were unexpected NaN values in the input data. 

Note: this check *only* occurs if the keyword argument `return_error=True`, which is the default value for this kwarg in `phase_cross_correlation`. If `return_error=False` then there is no check, and it's possible to get a garbage shift result of zero back if the input data accidentally contained NaNs. This is because it's computationally expensive to check a large array for NaN values, which is what we'd have to do if `return_error=False`. Still, it seems helpful if we can catch at least some of these mistakes and redirect the user to the masked cross correlation functionality instead.

## Checklist

- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/registration/__init__.py
+++ b/skimage/registration/__init__.py
@@ -1,9 +1,7 @@
 from ._optical_flow import optical_flow_tvl1
 from ._phase_cross_correlation import phase_cross_correlation
-from ._masked_phase_cross_correlation import cross_correlate_masked
 
 __all__ = [
     'optical_flow_tvl1',
-    'phase_cross_correlation',
-    'cross_correlate_masked'
+    'phase_cross_correlation'
     ]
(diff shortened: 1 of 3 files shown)


emmanuelle:
A possibility would be to add a note in the docstring mentioning that if don't return the error you are supposed to check that there are no nans (or to use a mask).

GenevieveBuckley:
> A possibility would be to add a note in the docstring mentioning that if don't return the error you are supposed to check that there are no nans (or to use a mask).

The docstring already has info about the `reference_mask` and `moving_mask` keyword arguments, so I don't think there's much extra benefit to adding this (although I do appreciate your suggestion). I think that catching some of the errors some of the time, is as you say, a clear improvement over the existing situation so it'll be fine.

PR #4893, opened 2020-08-06
added script for plausible analytics

emmanuelle:
At the moment we have a script corresponding to Matomo Analytics, where we have a 21-day trial plan. However, Juan recently discovered Plausible Analytics, which is an open source alternative to Matomo. Both solutions are open-source and don't sell your data. Plausible seems to be cheaper.

This PR adds the script for plausible. The idea is to compare both solutions and to take a paid plan before the end of the trial plan.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,12 +24,6 @@ jobs:
       - store_artifacts:
           path: doc/build/html
           destination: doc/build/html
-      # Persists generated documentation so that it can be attached and deployed
-      # in the 'deploy' step.
-      # - persist_to_workspace:
-      #    root: doc/build/html
-      #    paths: .
-
 
 workflows:
   version: 2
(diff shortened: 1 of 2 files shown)


jni:
Why did everything break?

emmanuelle:
> Why did everything break?

Excellent question :-). I'm investigating.

emmanuelle:
@jni I think it was random CircleCI failures... it seems to be ok now.

jni:
They *are* random... And somewhat concerning:

https://travis-ci.org/github/scikit-image/scikit-image/jobs/715749296#L4329-L4359

anyway do you want to uncomment the lines again? =)

emmanuelle:
> They _are_ random... And somewhat concerning:
> 
> https://travis-ci.org/github/scikit-image/scikit-image/jobs/715749296#L4329-L4359
> 
> anyway do you want to uncomment the lines again? =)

In fact I'll remove them for good if you don't mind, because it seems they are not necessary for visualizing the artifact (I copied these lines from another project).

emmanuelle:
Anything else I need to do here? I'd like to merge this one soon so that we have some overlap between the free plans of matomo and plausible, in order to compare.

PR #4897, opened 2020-08-07
Fix so docs properly render

smsaladi:
This is a small fix so that the docs properly render in html (right now the code section is interpreted as text):
https://scikit-image.org/docs/dev/api/skimage.measure.html?highlight=region%20props#regionprops


sciunto:
perfect.

Issue #4902, opened 2020-08-08
slic is much slower when using a mask


emmanuelle:
```python
from skimage import segmentation, data
img = data.astronaut()
mask = np.ones(img.shape[:2], dtype=np.bool)
mask[0, 0] = False
%timeit _ = segmentation.slic(img, start_label=1)
```
Output
```
224 ms ± 2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
But
```python
%timeit _ = segmentation.slic(img, mask=mask, start_label=1)
```
Output
```
1.22 s ± 23.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
The difference is very large! I wonder if it's possible to reduce it.

grlee77:
I looked at #3120 earlier today and did not see an immediate way to speed up the Cython code, but did end up speeding up coming up with a way to accelerate the `_get_mask_centroids` a lot. I will open a PR with what I have to discuss.

grlee77:
I came across the following repository that has a fast, MIT-licensed Cython implementation, but I don't think it supports using a mask.
https://github.com/Algy/fast-slic

It may be worth looking into that and seeing what tricks lead to the claimed performance: "Fast-slic can process 1280x720 image stream at 60fps"

jni:
@grlee77 this issue was solved by #4903 (thank you!). Should we close it and create a separate one for the fast-slic audit?

grlee77:
> @grlee77 this issue was solved by #4903 (thank you!). Should we close it and create a separate one for the fast-slic audit?

sure I will open a new issue

PR #4903, opened 2020-08-09
accelerate the masked slic superpixels algorithm

grlee77:
## Description

This PR is related to #4902. For the image of shape (400, 400, 100) used in the benchmarks (`with multichannel=False`), I get a time of 59 seconds prior to this change where >60% of the time is taken up by the call to `kmeans2` in `_get_mask_centroids`. 

Two ways to make the kmeans2 call run faster are:
1.) use a subset of the full set of points for the first argument.
2.) lower the number of k-means iterations from SciPy's default value of 10.

This gives a reduction from 59.4 s to 19.9 s to run the `time_mask_slic` benchmark. Before this change ~60% of the execution time is in `_get_mask_slic` while afterwards it is around 5%.

After this change the `slic` output is still deterministic, but the starting centroids would not be identical to prior releases. If that is a problem, we could add a flag to only enable this acceleration on request.

One unresolved question is how many iterations and what density of samples is most appropriate. I set `dense_factor=10` empirically. It may be possible to use a somewhat lower value, but there is very little further performance gain, at least for the benchmark case. Also, with `dense_factor` of 1 or 2 I sometimes see warnings like:

```
scipy/cluster/vq.py:574:
UserWarning: One of the clusters is empty. Re-run kmeans with a different initialization.
  warnings.warn("One of the clusters is empty. "
```

For the `mask_time_slic` benchmark case there are 25,992,000 nonzero voxels in the mask, but after the change in this PR we will only use a much smaller subset of these in `kmeans2`. After this change, for the benchmark example, ``n_dense = (dense_factor**ndim) * n_centroids`` is only 100,000 out of 25,992,000 points. 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_segmentation.py
+++ b/benchmarks/benchmark_segmentation.py
@@ -20,7 +20,6 @@ class SlicSegmentation:
             self.slic_kwargs = {}
 
     def time_slic_basic(self):
-
         segmentation.slic(self.image, enforce_connectivity=False,
                           multichannel=False, **self.slic_kwargs)
 
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -27,6 +27,11 @@ Improvements
   single-precision when the input image is single-precision. This can give a
   substantial performance improvement when working with single precision data.
 
+- The performance of the SLIC superpixels algorithm
+  (``skimage.segmentation.slice``) was improved for the case where a mask
+  is supplied by the user (#4903). The specific superpixels produced by
+  masked SLIC will not be identical to those produced by prior releases.
+
 API Changes
 -----------
 
--- a/skimage/segmentation/slic_superpixels.py
+++ b/skimage/segmentation/slic_superpixels.py
@@ -12,7 +12,7 @@ from ..util import img_as_float, regular_grid
 from ..color import rgb2lab
 
 
-def _get_mask_centroids(mask, n_centroids):
+def _get_mask_centroids(mask, n_centroids, multichannel):
     """Find regularly spaced centroids on a mask.
 
     Parameters
@@ -48,7 +48,8 @@ def _get_mask_centroids(mask, n_centroids):
     # somewhat arbitrarily choose dense_factor=10 to make the samples
     # 10 times closer together along each axis than the n_centroids samples.
     dense_factor = 10
-    n_dense = (dense_factor ** mask.ndim) * n_centroids
+    ndim_spatial = mask.ndim - 1 if multichannel else mask.ndim
+    n_dense = int((dense_factor ** ndim_spatial) * n_centroids)
     if len(coord) > n_dense:
         # subset of points to use for the k-means calculation
         # (much denser than idx, but less than the full set)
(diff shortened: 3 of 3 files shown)


grlee77:
Results of the masked slic example before and after this PR. Runtime for the masked 2D case decreases from 545 ms to 303 ms, so less than the nearly factor 3 runtime decrease for the 3D benchmark, but still a substantial reduction.

mask_slic example on master:
mask_slic_master

vs. this PR (note similar qualitative superpixels in the masked case, but not the same ones as in current master)
mask_slic_pr

emmanuelle:
Thanks, this is a very smart optimization :-). Could you please:
- add an item in the `release_dev.rst` (probably under Improvements)
- add a test so that the two cases of the new if/else block are covered by tests?
When I was profiling the module yesterday I thought maybe it would also be possible to gain a little my doing early stopping (small number of iterations) in this part of the code
```python
if update_centroids:
        # Step 2 of the algorithm [3]_
        _slic_cython(image, mask, segments, step, max_iter, spacing,
                     slic_zero, ignore_color=True,
                     start_label=start_label)
```
but I'm not so sure of what this part does (it looks to me like a simple kmeans if color is not taken into account), so not 100% sure.

grlee77:
Actually, I think we can also get a large speedup by changing the SLIC search window area to match the size in the original publication. Specifically, do you know why the following code is +/- `2*step` and not just +/- `step`?

https://github.com/scikit-image/scikit-image/blob/66f78a4a3aeadc423f4d2389dce4b6666ea50a2f/skimage/segmentation/_slic.pyx#L141-L146

If you look at Figure 2 [in this paper by the original authors](https://infoscience.epfl.ch/record/177415/files/Superpixel_PAMI2011-2.pdf) you can see that the search window is size `2*step` per size, but our code is `4*step` per side. Changing that would make a factor of `2**ndim` reduction in the number of times the inner loop is called.

grlee77:
> When I was profiling the module yesterday I thought maybe it would also be possible to gain a little my doing early stopping (small number of iterations) in this part of the code

Yes, I think we could do that. I was playing with it on an example and the centroids actually don't converge until many iterations (>150 for the `plot_mask_slic` example data), but we don't need to reach convergence since the real `slic` call below is going to modify the shapes substantially anyways. Also, we definitely shouldn't need a 4*step search window size (see above) for these initial iterations even if we choose to keep that for the main `slic` iterations below.

emmanuelle:
About the size of the window, this is a very interested comment. Maybe the idea was to allow more convoluted superpixels, which follow gradients in a better way, but 
- if one wants to have elongated / convoluted superpixels, one should use felzenswalb instead (for example)
- a consequence of our generous window size is that superpixels are often not connected, and that there is a lot of work to do in the `enforce_connectivity` step. See for example below 
superpixels_before_connectivity

So I really think it's worth changing the window size to the one of the original article. Probably in an other PR since it will be a bit more controversial than this one.

emmanuelle:
@grlee77 the PR is already a big improvement, we can leave further optimization (reducing the size of the search window as in the original article). If you have the time to add a line in the release notes it'd be awesome, otherwise I can do it in the "catching-up on release notes" PR.

emmanuelle (approved):
Thanks @grlee77 the PR is ready to go as far as I'm concerned! Now we need a second review :-).

emmanuelle:
ping @scikit-image/core

jni (approved):

jni:
Thank you @grlee77 and thank you @emmanuelle for the ping. =)

Issue #4407, opened 2020-01-22
array or ndarray as argument type documentation?


rfezzani:
## Description

I found that argument type documentation is not consistent for numpy arrays over `skimage` functions docstings:

- `array` is used 198 times,
- `ndarray` 118 times,
- and `numpy array` 5 times.

Should we think about defining a convention for this?

Here are the cleaned output of `grep`:
[array.txt](https://github.com/scikit-image/scikit-image/files/4098948/array.txt)
[ndarray.txt](https://github.com/scikit-image/scikit-image/files/4098949/ndarray.txt)

grlee77:
I see use both `array` and `ndarray` used with the docstrings of NumPy and SciPy themselves, although `ndarray` seems to be more common. I don't see a clear preference for either case specified in the [NumPy documenting guidelines](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard)

The other case you didn't mention is `array_like` which should technically be used rather than `ndarray` when the function accepts other types of sequences as well (e.g. it may call `np.asarray` or `np.atleast1d`, etc. internally).

rfezzani:
Thank you @grlee77 for linking to numpy doc guidelines. I did not considered `array_like` since it is used to describe arguments that can be converted to `ndarray` (i.e. scalar types, sequence types) according to these guidelines.
My preference goes to `ndarray`, since it is the name of the numpy array class while `array` is a function for numpy array creation.

jni:
In my head, `ndarray` meant "a NumPy ndarray object", while `array` meant "array-like". Perhaps we should use that convention or similar? Specifically, I am thinking that any function that uses Cython needs an ndarray, but most that don't can still work with CuPy arrays or dask arrays. It would be great if we could document this distinction somehow.

rfezzani:
I like your point of view @jni, but why not use the more conventional `array_like` instead of `array` then?
If we decide to go this way, we will have to call `np.asarray` whenever an argument is declared as `array_like`. This make me think of the `matrix` issue #4383.

jni:
Well, it's actually a slightly different issue. We *don't* want to call `np.asarray` on a Dask array. We just want to call `np.sum`, `np.max`, `arr @ vals`, etc as usual, and let [`__array_function__`](https://numpy.org/neps/nep-0018-array-function-protocol.html) and related protocols do their magic.

This is annoyingly subtle! LOL

rfezzani:
[...]

PR #4912, opened 2020-08-13
Use sparse cross-correlation to accelerate local thresholding functions

zeroth:
## Description

correlate_sparse function computes valid cross-correlation, this function is *fast* when `kernel` is large with many zeros.

this dramatically speeds up sauvola and niblack thresholding for large kernel sizes and is equally fast for smaller ones. 

@jni developed this code in [Numba in Skan](https://github.com/jni/skan/blob/a367746c814c98124b2e5817cda2f3d97c22d9e7/skan/vendored/thresholding.py#L27) out of necessity, so this brings the skimage version into line with that.

I've added a benchmark to verify this.
- For 2d Images:
    Without sparse correlate  **780 ms**.
    With Sparse Correlate  **601ms**.

- For 3d Image:
    Without Sparse Correlate **1min 28s**.
    With Sparse Correlate  **809ms**.

## For reviewers

  later.


emmanuelle:
Thank you for the PR @zeroth! A meta-question before starting a proper review: would it make sense contributing this algorithm to scipy instead? Usually the more upstream the better, if it makes sense for the upstream project of course.

emmanuelle:
pinging @grlee77 about the interest for scipy.

grlee77:
Yeah the current SciPy ndimage convolution code is mainly good for small, densely populated kernels. If the kernels are larger and not sparsely populated than FFT-based convolution is a reasonable alternative. The approach here does make sense for sparsely populated medium to large kernels. 

I guess the question regarding whether this should go into scipy.ndimage rather than here would be whether it is sufficiently broad in application? I am not sure how common such filters are in practice. I do appreciate you pointing out the use cases here for the niblack and sauvola thresholding methods.  Here is another publication I had seen with [sets of nd filters that each have only two non-zero entries](https://arxiv.org/abs/1708.08421). If we can make an argument that this is of general interest, then SciPy could be a reasonable fit. 

I do like the relative simplicity of the Cython code here as opposed to most of the C code underlying scipy.ndimage.

I had actually recently implemented this same sort of thing in a PR to CuPy, but we ended up not merging it (cupy/cupy#2842). This is in part because such a feature was not currently in SciPy and CuPy tends to stick closely to the NumPy and SciPy APIs.  It was also a bit slower in the fully dense case and we thought perhaps the use case for highly sparse kernels was a bit too niche to warrant supporting two parallel kernel generation codes.

benchmarks/benchmark_threshold_sauvola.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+from skimage import filters
+
+class ThresholdSuite:
+    """Benchmark for transform routines in scikit-image."""
+
+
+    def setup(self):
+        self.image = np.zeros((2000, 2000))
+        self.image3D = np.zeros((300, 300, 30))
```
jni, line 10: This shape is a Matlab-ism. =P Prefer (30, 300, 300) in NumPy. =)

benchmarks/benchmark_threshold_sauvola.py
```diff
@@ -0,0 +1,25 @@
...
+
+        idx = np.arange(500, 700)
+        idx3D = np.arange(100, 100)
+
+        self.image[idx[::-1], idx] = 255
+        self.image[idx, idx] = 255
+
+        self.image3D[idx3D[::-1], idx3D, idx3D] = 255
+        self.image3D[idx3D, idx3D, idx3D] = 255
```
jni, line 19: If you are going to use 255 as the value, make sure that `np.zeros` is called with `dtype=np.uint8`. Otherwise, use 1.0 here. See https://scikit-image.org/docs/dev/user_guide/data_types.html for details.

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,69 @@
+import numpy as np
+from ._sparse_cy import _correlate_sparse_offsets
+
+
+def broadcast_mgrid(arrays):
```
jni, line 5: This definitely needs a docstring with examples. I wrote this and reading the code I can't figure out what it does. 😂 There's a lesson here that everyone needs to learn multiple times in their coding life. 😂 

iirc it's like `np.meshgrid(index='ij', sparse=True)`, but broadcasting the resulting arrays together.

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,69 @@
...
+    ----------
+    image : array of float, shape (M, N,[ ...,] P)
+        The input array. If mode is 'valid', this array should already be
+        padded, as a margin of the same shape as kernel will be stripped
+        off.
+    kernel : array of float, shape (Q, R,[ ...,] S)
+        The kernel to be correlated. Must have the same number of
+        dimensions as `padded_array`. For high performance, it should
+        be sparse (few nonzero entries).
+    mode : string, optional
```
jni, line 33: @scikit-image/core one Q is whether we want to take *scipy* modes as input and translate to the NumPy modes. My instinct says yes since this is meant to be a "mirror" of `ndi.correlate`.

Does anyone have a time machine so we can undo the divergence between the modes? =P

jni:
@emmanuelle @grlee77 in an ideal world this would indeed happen upstream in ndimage, but imho that is a much bigger battle. The change here is not invasive at all and I'd like to get it in locally before contemplating an upstream port. I have literally never managed to go beyond skin-deep into the ndimage code, despite several attempts.

zeroth:
@jni I have updated the code with your recommendation please have a look .

grlee77:
@jni, I am in agreement about going ahead with this here and can help review it now. We can revisit for SciPy later if desired.

> I have literally never managed to go beyond skin-deep into the ndimage code, despite several attempts.

I would recommend keeping this in Cython if submitted for consideration within ndimage. Although most of ndimage is in C, `ndimage.label` is actually Cython and IIRC that was encouraged as easier to maintain.

skimage/filters/_sparse.py
grlee77: It does seem that we have been using the SciPy mode names within `skimage.filters`, so I agree that we should also do that here for consistency.

Unfortunately we previously made the opposite decision for the `skimage.transform` functions `warp`, `reshape` and `resize`! (`warp` actually uses a [utility to convert names from numpy.pad to ndimage equivalents](https://github.com/scikit-image/scikit-image/blob/59fd9682aad41a1f69a35f05d2a7e15ebdfb8b29/skimage/transform/_geometric.py#L9) so it can `ndi.map_coordinates` in some cases).

grlee77: There are more extension modes in pad than in ndimage, but just supporting the subset corresponding to those in ndimage should be sufficient here. The three that change name from numpy->ndimage are:
pad name | ndimage name
--------------|---------------------
nearest | edge
symmetric | reflect
reflect | mirror

The two that are the same for both are `wrap` and `constant`

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+    # memory-efficient version of numpy.mgrid
+    corner_multi_indices = np.broadcast_arrays(
+                        *np.meshgrid(
+                            *[np.arange(i) for i in result.shape], 
+                            indexing='ij', 
+                            sparse=True
+                            )
+                        )
```
grlee77, line 57:
```suggestion
    corner_multi_indices = np.meshgrid(*[np.arange(i) for i in result.shape], indexing='ij')
```

Isn't this just the same as calling meshgrid alone, but with the default `sparse=False`?

grlee77: Actually, you can leave `sparse=True`. The sparse arrays can be used as input to `np.ravel_multi_index` and it will take care of the broadcasting.

grlee77, line 57: I at first thought this was similar to just calling meshgrid with `sparse=False`, but on closer inspection I see that `broadcast_arrays` creates arrays that appear dense when you look at their shape, but are not actually consuming more memory due to a trick with stride=0, so this approach is preferable. 

I think it is simpler to just pass the list of sparse arrays produced by `meshgrid` directly to `np.ravel_multi_index` and allow it to handle the broadcasting, though.

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,65 @@
...
+
+    corner_indices = np.ravel_multi_index(
+        corner_multi_indices, padded_image.shape
+    ).ravel().astype(np.intp)
```
grlee77, line 61:
```suggestion
    ).ravel().astype(np.intp, copy=False)
```

avoid making a copy if possible

skimage/filters/_sparse.py
jni: oooooh I was unaware that `ravel_multi_index` would handle the broadcasting! Just verified it does! Nice! NumPy is always so full of pleasant surprises. =D

@zeroth this means you can remove `np.broadcast_arrays(*)`, ie `corner_multi_indices = np.meshgrid(...)`

skimage/filters/_sparse.py
jni: > Note that although the name reflect is present in both, the behavior is not equivalent!

I know. 🤦 Again: time machine. =P

skimage/filters/_sparse.py
zeroth: @jni Tested and updated. Please have a look.

skimage/filters/_sparse.py
zeroth: @grlee77 thanks for the tip :)

grlee77:
I just noticed that the benchmark here is added as a new file with a name corresponding to a specific function, but all other files in that folder are named after modules. For that reason, I would move this benchmark into the existing `benchmark_filters.py` file instead.

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,88 @@
...
+
+    See ``scipy.ndimage.correlate`` for a description of cross-correlation.
+
+    Parameters
+    ----------
+    image : array of float, shape (M, N,[ ...,] P)
+        The input array. If mode is 'valid', this array should already be
+        padded, as a margin of the same shape as kernel will be stripped
+        off.
+    kernel : array of float, shape (Q, R,[ ...,] S)
```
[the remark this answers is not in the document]
jni: @sciunto `ndarray of float, shape ...`? ie I don't think we've made up our mind about whether to specify the dtype, but I think we should, it's very useful.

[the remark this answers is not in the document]
sciunto: I put some hope in signature annotations, but for numpy, it does not work afaik. I agree that this information is useful.

Suggestion: `ndarray, dtype float, shape (Q, R,[ ...,] S)` ?

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,88 @@
...
+        inside the original data.
+
+    Returns
+    -------
+    result : array of float, shape (M, N,[ ...,] P)
+        The result of cross-correlating `image` with `kernel`. If mode
+        'valid' is used, the resulting shape is (M-Q+1, N-R+1,[ ...,] P-S+1).
+    """
+    if not isinstance(kernel, np.ndarray):
+        msg = '`correlate_sparse` Kernal must be an numpy array object.'
```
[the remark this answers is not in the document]
sciunto: I updated the change.

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,100 @@
...
+            msg = ('Window size for `threshold_sauvola` or '
+                   '`threshold_niblack` must not be even on any dimension. '
+                   'Got {}'.format(axis_sizes))
+            raise ValueError(msg)
+
+
+def _to_np_mode(mode):
+    """Convert from `scipy.ndimage.correlate` mode name 
+    to the corresponding numpy.pad  mode."""
+    mode_translation_dict = dict(nearest = 'edge', reflect='symmetric',
```
jni, line 29:
```suggestion
    mode_translation_dict = dict(nearest='edge', reflect='symmetric',
```

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,100 @@
...
+        if axis_size % 2 == 0:
+            msg = ('Window size for `threshold_sauvola` or '
+                   '`threshold_niblack` must not be even on any dimension. '
+                   'Got {}'.format(axis_sizes))
+            raise ValueError(msg)
+
+
+def _to_np_mode(mode):
+    """Convert from `scipy.ndimage.correlate` mode name 
+    to the corresponding numpy.pad  mode."""
```
jni, line 28:
```suggestion
    """Convert padding modes from `ndi.correlate` to `np.pad`."""
```

skimage/filters/tests/test_correlate.py
```diff
@@ -0,0 +1,21 @@
+import numpy as np
+from skimage._shared import testing
+from skimage._shared.testing import assert_equal, assert_almost_equal
+from skimage._shared.testing import assert_array_equal
+from scipy import ndimage as ndi
+from skimage.filters import correlate_sparse
+
+def test_correlate_sparse():
```
jni, line 8: @zeroth could you use `skimage._shared.testing.parametrize` (example use [here](https://github.com/scikit-image/scikit-image/blob/bab6e3b92596bdf020c5a402000112b39579b06e/skimage/filters/tests/test_window.py#L17-L20)) to test all available ndimage.correlate modes? *and* additionally, test `valid` separately since that requires different handling.

Finally, you should also use `testing.raises` to test what happens when you give an even-sized kernel.

jni (approved):
@scikit-image/core imho this is ready to merge! 🎉 Since I co-authored bits of the code I'll let someone else review + merge.

skimage/filters/_sparse_cy.pyx
```diff
@@ -0,0 +1,24 @@
...
+    cdef Py_ssize_t i
+    cdef Py_ssize_t indices_len = indices.shape[0]
+    cdef Py_ssize_t value_len = values.shape[0]
+    cdef Py_ssize_t off
+    cdef np_floats val
+    cdef Py_ssize_t vindex
+    
+    for vindex in range(value_len):
+        off = offsets[vindex]
+        val = values[vindex]
```
rfezzani, line 20:
```suggestion
    for off, val in zip(offsets, values):
```

rfezzani (commented):
Thank you @zeroth, I just have a side question: can't we release the GIL in `_correlate_sparse_offsets`?

skimage/filters/_sparse_cy.pyx
jni: @rfezzani good points! One should check that the above suggestion does not harm the benchmarks.

@zeroth you can release the gil by wrapping the loop code in a

```python
with nogil:
    do_something...
```

block

zeroth: Hi @rfezzani and @jni,
Quick update,
if I just update the current code to  `for off, val in zip(offsets, values):`  there no performance hit.
But, if I add 
```
with nogil:
        for off, val in zip(offsets, values):
```
Then there is a compile error 
```
with nogil:
        for off, val in zip(offsets, values):
                                            ^
skimage\filters\_sparse_cy.pyx:23:37: Converting to Python object not allowed without gil
```
I think making the thread-safe is priority hence I have just added `with nogil`. 
Let me know your thoughts.

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,99 @@
...
+            mode=np_mode,
+        )
+    indices = np.nonzero(kernel)
+    offsets = np.ravel_multi_index(indices, padded_image.shape).astype(np.intp)
+    values = kernel[indices].astype(padded_image.dtype)
+
+    result = np.zeros(
+        [a - b + 1 for a, b in zip(padded_image.shape, kernel.shape)],
+        dtype=padded_image.dtype,
+    )
```
rfezzani, line 85: Aren't `result` and `image` shapes equal? why not simply
```suggestion
    result = np.zeros_like(image)
```

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Returns
+    -------
+    result : array of float, shape (M, N,[ ...,] P)
+        The result of cross-correlating `image` with `kernel`. If mode
+        'valid' is used, the resulting shape is (M-Q+1, N-R+1,[ ...,] P-S+1).
+    """
+    if not isinstance(kernel, np.ndarray):
+        msg = '`correlate_sparse` kernel must be an numpy array object.'
+        raise ValueError(msg)
```
rfezzani, line 66: Can't we accept array like inputs for `kernel`?
```suggestion
    kernel = np.asarray(kernel)
```

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # memory-efficient version of numpy.mgrid
+    corner_multi_indices = np.meshgrid(*[np.arange(i) for i in result.shape],
+                                       indexing='ij',
+                                       sparse=True
+                                       )
+
+    corner_indices = np.ravel_multi_index(
+        corner_multi_indices, padded_image.shape
+    ).ravel().astype(np.intp, copy=False)
```
rfezzani, line 95:
```suggestion
    ).astype(np.intp, copy=False).reshape(-1)
```

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,99 @@
...
+
+    _correlate_sparse_offsets(padded_image.ravel(), corner_indices,
+                              offsets, values, result.ravel())
```
rfezzani, line 98: If we simply need a view, `np.ravel` [doc](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html) advices to use `.reahspae(-1)`:
```suggestion
    _correlate_sparse_offsets(padded_image.reshape(-1), corner_indices,
                              offsets, values, result.reshape(-1))
```

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,99 @@
...
+                                 mirror='reflect')
+    if mode in mode_translation_dict:
+        mode = mode_translation_dict[mode]
+    return mode
+
+
+def correlate_sparse(image, kernel, mode='reflect'):
+    """Compute valid cross-correlation of `padded_array` and `kernel`.
+
+    This function is *fast* when `kernel` is large with many zeros.
```
rfezzani, line 38:
```suggestion
    This function is well suited when `kernel` is large with many zeros.
```

rfezzani (commented):
Thank you for investigating the `nogil` point @zeroth.
I have an other question: do we want to support kernels with even shapes? Should we then consider using the `origin` argument like in `scipy.ndimage.correlate`?

skimage/filters/_sparse.py
zeroth: @rfezzani we had this discussion before and based on #4407  we decided to go for ndarray. https://github.com/scikit-image/scikit-image/pull/4912#discussion_r475174333 @jni please correct me if I am wrong.

skimage/filters/_sparse.py
zeroth: @rfezzani `result.shape == image.shape` that's for almost all the modes except `valid`. @jni I think we can do this unless I am missing something, what do you think?

jni:
@rfezzani I think that can come in a later PR.

skimage/filters/_sparse.py
jni: As you say, this expression is valid in all cases, while the proposed expression won't work when `mode='valid'`. So we might as well keep the current expression?

skimage/filters/_sparse.py
jni: We can leave this decision for later. Honestly the entire ecosystem is having massive headaches with "array like" when it comes to typing, so I think I'd rather move to the future where everyone is responsible for providing actual NumPy arrays.

Again, we can revisit this decision later, I don't think it should hold up the PR.

skimage/filters/_sparse.py
rfezzani: Do I missed something? `valid` is not a valid mode for [ndi.correlate](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.correlate.html)

skimage/filters/_sparse.py
jni: @rfezzani I kinda like fast better. Let's be bold, then if someone complains that their implementation is faster, we can copy it. =P

rfezzani:
> @rfezzani I think that can come in a later PR.

I agree, but for now we may reject even shape kernels or add a side note in the doc describing the behavior of the function in this case...

skimage/filters/_sparse.py
rfezzani: Well, I assume that no one will complain in any case :smile:.
I just feel that "fast" is not precise, that's it....

skimage/filters/_sparse.py
```diff
@@ -0,0 +1,99 @@
...
+    -------
+    result : array of float, shape (M, N,[ ...,] P)
+        The result of cross-correlating `image` with `kernel`. If mode
+        'valid' is used, the resulting shape is (M-Q+1, N-R+1,[ ...,] P-S+1).
+    """
+    if not isinstance(kernel, np.ndarray):
+        msg = '`correlate_sparse` kernel must be an numpy array object.'
+        raise ValueError(msg)
+
+    if mode == 'valid':
```
zeroth, line 68: @rfezzani afaik there is no mode name `valid` in `ndi.correlate` we have this specific mode check here.

skimage/filters/_sparse.py
rfezzani: > Honestly the entire ecosystem is having massive headaches with "array like" 

Let's assume then that `kernel` is an array and avoid the check. I don't see such validation else where in `skimage`...

skimage/filters/_sparse.py
rfezzani: Oh, I definitely missed something :sweat_smile:...
This is may be opinionated, but I vote for removing this option! It comes from [`scipy.signal.convolve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.convolve.html), and I personally  don't like it: a user needing the valid part only have to crop the output by him self!

jni: @rfezzani in this case it is incredibly useful because we don't want to pad the image if the user has already padded it. Without it, we *have* to pad.

jni:
@rfezzani 

> for now we may reject even shape kernels

We do!

https://github.com/scikit-image/scikit-image/pull/4912/files#diff-a86dd6897f0acc876eda03ec11dbeebbR6-R23

skimage/filters/_sparse.py
jni: > I just feel that "fast" is not precise

Neither is well-suited. ;)

skimage/filters/thresholding.py
```diff
@@ -899,10 +881,10 @@ def _mean_std(image, w):
...
 
     total_window_size = np.prod(w)
-    sum_full = ndi.correlate(integral, kern, mode='constant')
-    m = crop(sum_full, pad_width) / total_window_size
-    sum_sq_full = ndi.correlate(integral_sq, kern, mode='constant')
-    g2 = crop(sum_sq_full, pad_width) / total_window_size
+    sum_full = correlate_sparse(integral, kern, mode='valid')
+    m = sum_full / total_window_size
+    sum_sq_full = correlate_sparse(integral_sq, kern, mode='valid')
+    g2 = sum_sq_full / total_window_size
```
jni, line 887: @rfezzani over here 👆. If we don't have mode valid, we pad a potentially very large volume with a potentially very large pad, and compute a bunch of useless values, only to remove them at the end.

skimage/filters/_sparse.py
jni: @rfezzani https://github.com/scikit-image/scikit-image/pull/4912/files#r479006134

skimage/filters/_sparse.py
jni: I just merged this suggestion. The problem with assuming is that the code will fail in a rather obscure way (list object has no attribute shape).

skimage/filters/thresholding.py
rfezzani: I see, now my next question is: is the acceleration coming from the kernel sparseness or from the `"valid"` mode? Can't we then simply use `scipy.signal.correlate` here instead of `ndi.correlate`?

rfezzani:
> We do!

Excellent, I missed this part too, sorry :) It is fine then!

jni:
The Travis failures don't appear to be related to this PR. Something about texlive *again*. 🤦‍♂

skimage/filters/thresholding.py
jni: No, it's the sparseness. For large kernels, SciPy (rightly) switches to FFT, which is much more expensive than 4 (2D) or 8 (3D) simple passes over the data.

sciunto:
> The Travis failures don't appear to be related to this PR. Something about texlive _again_. man_facepalming

This time, it seems to be a timeout issue.

skimage/filters/_sparse.py
rfezzani: We may be can find a better wording then :smile:

jni:
@scikit-image/core I think this PR is ready to merge, anyone else want to review?

skimage/filters/_sparse.py
stefanv: I'm good with fast for now.

skimage/filters/_sparse_cy.pyx
```diff
@@ -0,0 +1,25 @@
...
+    cdef Py_ssize_t value_len = values.shape[0]
+    cdef Py_ssize_t off
+    cdef np_floats val
+    cdef Py_ssize_t vindex
+    
+    with nogil:
+        for vindex in range(value_len):
+            off = offsets[vindex]
+            val = values[vindex]    
+            # this loop order optimises cache access, gives 10x speedup
```
stefanv, line 22: Not sure I believe that.

stefanv:
Fantastic work, thank you @zeroth!

Request: please remove the 10x statement in a follow-up PR unless it is accurate.

skimage/filters/_sparse_cy.pyx
jni: LOL It was 10x on the hardware in which I first tested this. (2015 Macbook.) On more modern hardware (my X1C gen6), it's a 4x speedup, which you can see for yourself (and I think did see for yourself? 😂) live in my [SciPy 2019 skan talk](https://www.youtube.com/watch?v=0pUPNMglnaE) (see minute 20:35 onwards). I don't have any reason to believe Cython and Numba would behave differently here.

The point of the comment is to prevent people from swapping the loop to the more "intuitive" order, so I think it should be kept in *some* form ("*up to* 10x speedup"?).

jni: Performance is also dependent on the size of the dataset and whether it's 2D/3D (more cache misses in 3D).

jni: Also for anyone who might be interested: the repo for that talk: https://github.com/jni/skan-talk-scipy-2019

stefanv: Fine with me, thanks Juan!

Issue #3927, opened 2019-05-24
camera(man) image must be removed for license issues


carandraug:
I am interested on the use of the `camera.png` image. The file is at `skimage/data/camera.png`. However, I am unable to find copyright information about this file on the scikit-image repository.

My understanding is that this image is copyright of MIT and one would need permission to use and redistribute it.

jni:
Hmm, I couldn't find any info on the image online either. As policy we only include CC0/Public domain images, but I don't know the source here. @stefanv @ahojnnes @JDWarner ?

matthew-brett:
I couldn't find any license information, except that the image is one of the most common of the images used for image processing demonstrations.

https://www.quora.com/What-are-the-most-common-example-images-used-in-image-processing-and-computer-vision

You can find quite a few papers using this image to demonstrate image processing algorithms.

I saw the claim that the image belongs to MIT, but nothing to support it:

https://www.npmjs.com/package/cameraman

carandraug:
> I saw the claim that the image belongs to MIT, but nothing to support it:

On a Matlab installation, where this image is named "cameraman", in the file `toolbox/images/imdata/Contents.m`, it says the following "Copyright Massachusetts Institute of Technology.  Used with permission."

matthew-brett:
Aha - then I guess we have to ask the Mathworks who they got permission from.

stefanv:
That copyright claim is validated by Steve Eddins in [this entertaining blog post](https://blogs.mathworks.com/steve/2006/10/17/the-story-behind-the-matlab-default-image/) (see the comments).

I will write MIT and ask.

stefanv:
I'm afraid we'll have to start looking for a new image to replace our existing cameraman image.  Does anyone have an antique camera and a suit, so they can go and pose on the lawn? :)

Peter Bebergal of MIT wrote:

> After much research,  it turns out that MIT cannot actually be certain the image
is owned by MIT and as such cannot be released under a CC license. The most we
can do is say that by whatever rights we do have in the image, we can grant its
use for the express purpose you need. While MIT has been giving permission in
the past, it is not clear we should have done so without this disclaimer.

I then asked:

> Thank you very much for taking the time to investigate. Did you happen upon any
clues as to whom may have taken or provided the original?
>
> When you say "for the express purpose you need": as a library, our purpose is to
let our users use images without having to fear any copyright infringement. As
far as MIT is concerned, I then understand this to mean that they will not take
[...]

PR #4913, opened 2020-08-13
Replacing data/camera.png due to copyright

alexdesiqueira:
## Description

Removing `data/camera.png` (a.k.a. cameraman) due to copyright issues. Solves #3927.


grlee77:
This image is pretty widely used in lot of test cases, including some with some pre-computed results such as:
https://github.com/scikit-image/scikit-image/blob/59fd9682aad41a1f69a35f05d2a7e15ebdfb8b29/skimage/data/_registry.py#L121-L124

I am wondering if we can keep using it in the test suite for that purpose, but just not keep it as part of the public API or demos in the documentation?  Wouldn't that usage fit within the non-profit restrictions of the license?

alexdesiqueira:
@grlee77 indeed, I'm seeing that now! :sweat_smile:

> I am wondering if we can keep using it in the test suite for that purpose, but just not keep it as part of the public API or demos in the documentation?

Personally, I'd not be comfortable with that, Greg. It would give us some work (by Tupã, I know :joy:), but IMHO it's not like we don't have alternatives around with less restrictive licenses. Of course, my two BRL cents.

grlee77:
> Personally, I'd not be comfortable with that, Greg. It would give us some work (by Tupã, I know joy), but IMHO it's not like we don't have alternatives around with less restrictive licenses. Of course, my two BRL cents.

Sure, in most cases it is probably trivial to switch. For the few with precomputed outputs, I guess we can just regenerate those for a different image and update the registry.

alexdesiqueira:
@grlee77 yeah, I was thinking about that today. @stefanv wanted a nice image to take cameraman's place, but we didn't get it yet... I'll search for some viable candidates today.

stefanv:
Best to just bite the bullet, I think. It wouldn't be the end of the world to use it internally, but it is shouldn't be too hard to switch.

alexdesiqueira:
Alright, based on the discussion we had on #4914, `data.camera` will point to a new [camera.png](http://krvarshney.github.io/kush_varshney.png).

emmanuelle:
I was surprised that all tests passed with the new image but there is a catch: pooch uses the data files in master because of https://github.com/scikit-image/scikit-image/blob/master/skimage/data/__init__.py#L120, so that the old image is used (as you can see from https://225-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/data/plot_general.html#sphx-glr-auto-examples-data-plot-general-py for example).

As a workaround, could you please modify https://github.com/scikit-image/scikit-image/blob/master/skimage/data/__init__.py#L120 so that it points to your repo and branch? Then we can see if tests pass and the doc looks good, and of course we'll revert the code in `data` before merging.

alexdesiqueira:
Sure thing @emmanuelle, thanks for the catch! You're right, I think it's very unlikely the tests will pass for the new image :slightly_smiling_face:

skimage/data/__init__.py
```diff
@@ -117,7 +117,7 @@ def create_image_fetcher():
     # This helps pooch understand that it should look in master
     # to find the required files
     pooch_version = __version__.replace('.dev', '+')
-    url = "https://github.com/scikit-image/scikit-image/raw/{version}/skimage/"
+    url = "https://github.com/alexdesiqueira/scikit-image/tree/goodbye_cameraman/skimage"  # test url; checking new image
```
emmanuelle, line 120:
```suggestion
    url = "https://github.com/alexdesiqueira/scikit-image/raw/goodbye_cameraman/skimage/"  # test url; checking new image
```

emmanuelle:
I pushed a commit to your branch because the URL where to find the files was not the correct one I think, let's keep fingers crossed this one works!

emmanuelle:
ok, you also need to modify the hash in the registry (see the error in CI).

emmanuelle:
We're making progress, now the next thing is to fix the 25 failing tests...
```
FAILED exposure/tests/test_exposure.py::test_equalize_masked - assert 1.23262...
FAILED feature/tests/test_canny.py::TestCanny::test_use_quantiles - Assertion...
FAILED filters/tests/test_ridges.py::test_2d_cropped_camera_image - ValueErro...
FAILED filters/tests/test_ridges.py::test_3d_cropped_camera_image - ValueErro...
FAILED filters/tests/test_thresholding.py::test_otsu_camera_image - assert 91...
FAILED filters/tests/test_thresholding.py::test_li_camera_image - assert 69.7...
FAILED filters/tests/test_thresholding.py::test_yen_camera_image - assert 199...
FAILED filters/tests/test_thresholding.py::test_isodata_camera_image - assert...
FAILED filters/tests/test_thresholding.py::test_threshold_minimum - Assertion...
FAILED io/tests/test_collection.py::TestImageCollection::test_concatenate - V...
FAILED io/tests/test_io.py::test_imread_file_url - assert (245, 327) == (512,...
FAILED measure/tests/test_structural_similarity.py::test_gaussian_mssim_vs_IPOL
FAILED measure/tests/test_structural_similarity.py::test_gaussian_mssim_vs_author_ref
FAILED measure/tests/test_structural_similarity.py::test_gaussian_mssim_and_gradient_vs_Matlab
FAILED measure/tests/test_structural_similarity.py::test_mssim_vs_legacy - As...
FAILED metrics/tests/test_simple_metrics.py::test_PSNR_vs_IPOL - AssertionErr...
FAILED metrics/tests/test_structural_similarity.py::test_gaussian_structural_similarity_vs_IPOL
FAILED metrics/tests/test_structural_similarity.py::test_gaussian_mssim_vs_author_ref
FAILED metrics/tests/test_structural_similarity.py::test_gaussian_mssim_and_gradient_vs_Matlab
FAILED metrics/tests/test_structural_similarity.py::test_mssim_vs_legacy - As...
FAILED restoration/tests/test_restoration.py::test_wiener - AssertionError: 
FAILED restoration/tests/test_restoration.py::test_unsupervised_wiener - Asse...
FAILED restoration/tests/test_restoration.py::test_richardson_lucy - Assertio...
FAILED util/tests/test_random_noise.py::test_salt_and_pepper - assert (3052 /...
FAILED util/tests/test_random_noise.py::test_clip_speckle - assert (1.2706687...
```
Note that the tests probably fail on your machine too, it will be easier to fix them and run the tests locally.

alexdesiqueira:
I'm on it :slightly_smiling_face: It seems we used `cameraman` too much... :joy:

sciunto:
how is it going @alexdesiqueira with this PR?

rfezzani:
> now the next thing is to fix the 25 failing tests...

What about using [`imageio`](https://imageio.readthedocs.io/en/stable/standardimages.html) in the tests to load our original cameraman image?

```python
img = imageio.imread("imageio:camera.png")
```

sciunto:
> > now the next thing is to fix the 25 failing tests...
> 
> What about using [`imageio`](https://imageio.readthedocs.io/en/stable/standardimages.html) in the tests to load our original cameraman image?
> 
> ```python
> img = imageio.imread("imageio:camera.png")
> ```

I think they will remove the image sooner or later for the same reason...

grlee77:
> What about using imageio in the tests to load our original cameraman image?

It is also available from our PyWavelets dependency via `pywt.data.camera()`, but I plan to replace it in PyWavelets as well.

rfezzani:
> I think they will remove the image sooner or later for the same reason...

You are right https://github.com/imageio/imageio/issues/452 :smile:

emmanuelle:
@alexdesiqueira do you need any help/input for this PR?

alexdesiqueira:
@emmanuelle sorry I didn't work here these days. I'll come back to it next week — I would ask some help to redo the pre-calculated values we have on tests, but I'll check it properly.

alexdesiqueira:
Hi everyone, let me discuss some issues I'm having while tackling this.

* On `skimage/metrics/tests/test_simple_metrics.py::test_PSNR_vs_IPOL`:

```
    def test_PSNR_vs_IPOL():
        # Tests vs. imdiff result from the following IPOL article and code:
        # https://www.ipol.im/pub/art/2011/g_lmii/
        p_IPOL = 22.4497
        p = peak_signal_noise_ratio(cam, cam_noisy)
>       assert_almost_equal(p, p_IPOL, decimal=4)
E       AssertionError: 
E       Arrays are not almost equal to 4 decimals
E        ACTUAL: 22.383828063078575
E        DESIRED: 22.4497

test_simple_metrics.py:23: AssertionError
```
* On `skimage/metrics/tests/test_structural_similarity.py::test_gaussian_structural_similarity_vs_IPOL`:

```
    def test_gaussian_structural_similarity_vs_IPOL():
        # Tests vs. imdiff result from the following IPOL article and code:
        # https://www.ipol.im/pub/art/2011/g_lmii/
        mssim_IPOL = 0.327309966087341
        mssim = structural_similarity(cam, cam_noisy, gaussian_weights=True,
                                      use_sample_covariance=False)
>       assert_almost_equal(mssim, mssim_IPOL, decimal=3)
E       AssertionError: 
E       Arrays are not almost equal to 3 decimals
E        ACTUAL: 0.3501368923026972
E        DESIRED: 0.327309966087341

test_structural_similarity.py:166: AssertionError
```

Not sure if `p_IPOL` and `mssim_IPOL` are kinda fixed for all kinds of images, or it's supposed to be specific to the old `cameraman`. I took a quick look at the paper but didn't figured out that much; asking @rbavery for help.

* On `skimage/metrics/tests/test_structural_similarity.py::test_gaussian_mssim_vs_author_ref`:

```
    def test_gaussian_mssim_and_gradient_vs_Matlab():
        # comparison to Matlab implementation of N. Avanaki:
        # https://ece.uwaterloo.ca/~nnikvand/Coderep/SHINE%20TOOLBOX/SHINEtoolbox/
        # Note: final line of ssim_sens.m was modified to discard image borders
    
        ref = np.load(fetch('data/mssim_matlab_output.npz'))
        grad_matlab = ref['grad_matlab']
        mssim_matlab = float(ref['mssim_matlab'])
    
        mssim, grad = structural_similarity(cam, cam_noisy, gaussian_weights=True,
                                            gradient=True,
                                            use_sample_covariance=False)
    
>       assert_almost_equal(mssim, mssim_matlab, decimal=3)
E       AssertionError: 
E       Arrays are not almost equal to 3 decimals
E        ACTUAL: 0.3501368923026972
E        DESIRED: 0.3273142956733568

test_structural_similarity.py:198: AssertionError
```

The toolbox address is not available anymore (and I don't have access to MATLAB). @rbavery could you please help me generating the new values we'd need to use here?


* On `skimage/metrics/tests/test_structural_similarity.py::test_gaussian_mssim_and_gradient_vs_Matlab`

```
    def test_gaussian_mssim_vs_author_ref():
        """
        test vs. result from original author's Matlab implementation available at
        https://ece.uwaterloo.ca/~z70wang/research/ssim/
    
        Matlab test code:
           img1 = imread('camera.png')
           img2 = imread('camera_noisy.png')
           mssim = structural_similarity_index(img1, img2)
        """
        mssim_matlab = 0.327314295673357
        mssim = structural_similarity(cam, cam_noisy, gaussian_weights=True,
                                      use_sample_covariance=False)
>       assert_almost_equal(mssim, mssim_matlab, decimal=10)
E       AssertionError: 
E       Arrays are not almost equal to 10 decimals
E        ACTUAL: 0.3501368923026972
E        DESIRED: 0.327314295673357

test_structural_similarity.py:182: AssertionError
```

* On `skimage/metrics/tests/test_structural_similarity.py::test_mssim_vs_legacy`:

```
    def test_mssim_vs_legacy():
        # check that ssim with default options matches skimage 0.11 result
        mssim_skimage_0pt11 = 0.34192589699605191
        mssim = structural_similarity(cam, cam_noisy)
>       assert_almost_equal(mssim, mssim_skimage_0pt11)
E       AssertionError: 
E       Arrays are not almost equal to 7 decimals
E        ACTUAL: 0.3561711557677345
E        DESIRED: 0.3419258969960519

test_structural_similarity.py:208: AssertionError
```

Having issues trying to install scikit-image 0.11.3:

`ERROR: Command errored out with exit status 1: /home/alex/anaconda3/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-qtr94oes/scikit-image/setup.py'"'"'; __file__='"'"'/tmp/pip-install-qtr94oes/scikit-image/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-cnxgrhng/install-record.txt --single-version-externally-managed --compile --install-headers /home/alex/anaconda3/include/python3.7m/scikit-image Check the logs for full command output.`

Is this test still necessary? Checking with @rbavery @sciunto.
Thank you for your help so far; I'll come back with other errors I cannot fix by my own.

alexdesiqueira:
Now to `skimage/restoration` :slightly_smiling_face: 

* On `skimage/restoration/tests/test_restoration.py::test_wiener()`:

```
    def test_wiener():
        psf = np.ones((5, 5)) / 25
        data = convolve2d(test_img, psf, 'same')
        np.random.seed(0)
        data += 0.1 * data.std() * np.random.standard_normal(data.shape)
        deconvolved = restoration.wiener(data, psf, 0.05)
    
        path = fetch('restoration/tests/camera_wiener.npy')
>       np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3)
E       AssertionError: 
E       Not equal to tolerance rtol=0.001, atol=0
E       
E       (shapes (245, 327), (512, 512) mismatch)
E        x: array([[ 0.156946,  0.213657,  0.302087, ...,  0.286404,  0.195821,
E                0.142914],
E              [ 0.291271,  0.396997,  0.53115 , ...,  0.375351,  0.274008,...
E        y: array([[0.125916, 0.183368, 0.302236, ..., 0.323755, 0.220084, 0.146606],
E              [0.222835, 0.297109, 0.433168, ..., 0.432838, 0.313082, 0.232841],
E              [0.342223, 0.437642, 0.596713, ..., 0.58443 , 0.436913, 0.343966],...
```

* On `skimage/restoration/tests/test_restoration.py::test_unsupervised_wiener()`:

```
    def test_unsupervised_wiener():
        psf = np.ones((5, 5)) / 25
        data = convolve2d(test_img, psf, 'same')
        np.random.seed(0)
        data += 0.1 * data.std() * np.random.standard_normal(data.shape)
        deconvolved, _ = restoration.unsupervised_wiener(data, psf)
    
>       np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3)
E       AssertionError: 
E       Not equal to tolerance rtol=0.001, atol=0
E       
E       (shapes (245, 327), (512, 512) mismatch)
E        x: array([[ 0.146846,  0.204145,  0.288669, ...,  0.285808,  0.192781,
E                0.098453],
E              [ 0.260007,  0.366386,  0.553619, ...,  0.343528,  0.215041,...
E        y: array([[0.110767, 0.15033 , 0.288193, ..., 0.330331, 0.206819, 0.162408],
E              [0.221344, 0.276947, 0.425761, ..., 0.435131, 0.279585, 0.241978],
E              [0.352406, 0.465364, 0.648465, ..., 0.632352, 0.454249, 0.351247],...
```

* On `skimage/restoration/tests/test_restoration.py::test_richardson_lucy()`:

```
    def test_richardson_lucy():
        psf = np.ones((5, 5)) / 25
        data = convolve2d(test_img, psf, 'same')
        np.random.seed(0)
        data += 0.1 * data.std() * np.random.standard_normal(data.shape)
        deconvolved = restoration.richardson_lucy(data, psf, 5)
    
        path = fetch('restoration/tests/camera_rl.npy')
>       np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3)
E       AssertionError: 
E       Not equal to tolerance rtol=0.001, atol=0
E       
E       (shapes (245, 327), (512, 512) mismatch)
E        x: array([[0.019313, 0.068293, 0.177045, ..., 0.147891, 0.054079, 0.015531],
E              [0.065458, 0.229525, 0.600645, ..., 0.508423, 0.186367, 0.053415],
E              [0.170681, 0.595333, 1.      , ..., 1.      , 0.497023, 0.142383],...
E        y: array([[0.017209, 0.062213, 0.169346, ..., 0.150776, 0.058305, 0.017192],
E              [0.059233, 0.213159, 0.581141, ..., 0.539677, 0.208713, 0.061162],
E              [0.150825, 0.546864, 1.      , ..., 1.      , 0.550621, 0.15953 ],...
```

Let me ask y'all, how were `camera_wiener.npy`, `camera_unsup.npy`, and `camera_rl.npy` calculated? It'd be great if @forieux could let us know, but it was a while ago.  I remember discussing it with @grlee77 as well...
Thanks again for your help!

grlee77:
For `structural_similarity` (SSIM), those are values I generated some years ago and are specific to the original camera image. I can try to regenerate for the new camera image. Note that for the Matlab value as used there, it was not based on the SSIM that is now included in Matlab itself, but was based on a Matlab implementation by the authors of the original SSIM publication.

grlee77:
> For structural_similarity (SSIM), those are values I generated some years ago and are specific to the original camera image. I can try to regenerate for the new camera image. Note that for the Matlab value as used there, it was not based on the SSIM that is now included in Matlab itself, but was based on a Matlab implementation by the authors of the original SSIM publication.

Another alternative to regenerating these with IPOL and Matlab implementations is to just replace those test cases with ones that use a stored value from the current scikit-image implementation as that one has already been validated. Basically, just run our current `peak_signal_noise_ratio` and `structural_similarity` on the new camera image and store those results for testing future versions against. That will still allow us to make sure future changes don't unintentionally change the result.

grlee77:
I see we already have a `test_mssim_vs_legacy`, so I would just update that one to store the current result from scikit-image master on the new camera image and then that will be the new "legacy" value we test against.

grlee77:
Here are PSNR and SSIM results vs. IPOL for the new cameraman image (The only change to IPOL's source was to print more digits).

### New metrics results for scikit-image

```Python
import numpy as np
import skimage.io
from skimage import data

np.random.seed(5)
cam = data.camera()
sigma = 20.0
cam_noisy = np.clip(cam + sigma * np.random.randn(*cam.shape), 0, 255)
cam_noisy = cam_noisy.astype(cam.dtype)
skimage.io.imsave('/tmp/cam_new_noisy.png', cam_noisy)  # for testing with IPOL

mssim = structural_similarity(cam, cam_noisy, gaussian_weights=True, use_sample_covariance=False)
psnr = peak_signal_noise_ratio(cam, cam_noisy)
print(f"MSSIM = {mssim}")
print(f"PSNR = {psnr}")
```

```
MSSIM = 0.3501368923026972
PSNR = 22.383828063078575
```


### IPOL Result
```
# ./imdiff -m mssim ~/camera.png ~/cam_new_noisy.png 
0.350132882595062
# ./imdiff -m psnr ~/camera.png ~/cam_new_noisy.png 
22.383828134978579
```

grlee77:
For wiener, richardson-lucy, etc. I would also just replace those stored files with new ones generated from the current implementation for this image. It will serve the same purpose going forward to make sure that we don't unintentionally change the behavior.

alexdesiqueira:
Thanks for your help @grlee77!

> Another alternative to regenerating these with IPOL and Matlab implementations is to just replace those test cases with ones that use a stored value from the current scikit-image implementation as that one has already been validated. Basically, just run our current peak_signal_noise_ratio and structural_similarity on the new camera image and store those results for testing future versions against. That will still allow us to make sure future changes don't unintentionally change the result.

I think this would introduce redundancy in one or two tests,; maybe I'm wrong. I'll check! Thank you again!

grlee77:
> I think this would introduce redundancy in one or two tests,; maybe I'm wrong. I'll check! Thank you again!

Yes, I would just remove the MATLAB comparison unless we regenerate a new one. I think that is pretty low priority though as comparing to both IPOL and the existing implementation should be sufficient.

rbavery:
Hi @alexdesiqueira, my contribution to this module was mostly in pushing through an existing PR and organizing the tests, docs and module, not writing these tests themselves.  I think the values for the metrics `p_IPOL` and `mssim_IPOL` are specific to each comparison, so if the camera man image changes I think the test would need to change. I don't have access to MATLAB so unfortunately I can't help with regenerating the similarity metrics with MATLAB.

jni:
Yeah I agree with @grlee77, we don't need MATLAB to create a reference anymore, we just need to check against our current implementation, which we already established matches MATLAB.

@alexdesiqueira I just noticed your branch name 😂

alexdesiqueira:
Thank you for your help @rbavery, we appreciate it!
I'll follow what @grlee77 proposed and create a reference against ourselves then. Thanks y'all :slightly_smiling_face:

alexdesiqueira:
Fingers crossed :slightly_smiling_face: :crossed_fingers:

emmanuelle:
Some builds are passing now, almost there :-). New patches need to be generated for `plot_glcm.py` I think.

alexdesiqueira:
Somehow I can't run the doctests in my PC... :thinking:

alexdesiqueira:
Ah. The happiness. :slightly_smiling_face:

skimage/data/__init__.py
```diff
@@ -117,7 +117,7 @@ def create_image_fetcher():
     # This helps pooch understand that it should look in master
     # to find the required files
     pooch_version = __version__.replace('.dev', '+')
-    url = "https://github.com/scikit-image/scikit-image/raw/{version}/skimage/"
```
jni, line 120: Does this need changing before merging @alexdesiqueira?

skimage/data/__init__.py
```diff
@@ -337,12 +337,16 @@ def _load(f, as_gray=False):
...
+    Can be used for segmentation and denoising examples.
 
     Returns
     -------
     camera : (512, 512) uint8 ndarray
         Camera image.
+
+    Notes
+    -----
+    No copyright restrictions. CC0 by the photographer (Lav Varshney).
```
jni, line 349: Can you add a `.. versionchanged:: 0.18` here explaining the change and linking to the issue? "This image was changed due to copyright restrictions. See [1]_ for more information.", and a References section for [1].

Here's a versionchanged example for reference:

https://github.com/scikit-image/scikit-image/blob/83e857ca8f974bb560315a00989673b4b52f8a93/skimage/exposure/exposure.py#L346-L348

jni (approved):
@alexdesiqueira AWESOME, thank you! I'm very excited, I've just asked one question and one made a suggestion that we include a `versionchanged` tag in the docstring.

Possibly it should also be added as an API change in the release notes...

rfezzani (commented):
Excellent @alexdesiqueira, thank you! Last thing I think is making a last pass on gallery examples :wink:, [here](https://538-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/color_exposure/plot_tinting_grayscale_images.html#sphx-glr-auto-examples-color-exposure-plot-tinting-grayscale-images-py) for example:
Capture du 2020-09-17 09-38-59

skimage/data/__init__.py
alexdesiqueira: It does @jni, we'll get back to the original one when the PR is ready :slightly_smiling_face:

alexdesiqueira:
@rfezzani sure thing :slightly_smiling_face: this will take a little while, though.

grlee77:
Thanks for being persistent on working out all these cases!

I see one more, where the elliptical mask coordinates should be rescaled a bit to remain within the image bounds and give a more similar mask to the master branch. I hadn't realized that this new camera image was a bit smaller rather than (512, 512) like the old one, but that seems to be the source of the difference. See the figures at the bottom of:

[dev docs](https://scikit-image.org/docs/stable/auto_examples/registration/plot_masked_register_translation.html#sphx-glr-auto-examples-registration-plot-masked-register-translation-py) vs. [current artifact](https://539-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/registration/plot_masked_register_translation.html#sphx-glr-auto-examples-registration-plot-masked-register-translation-py)

jni:
@alexdesiqueira @grlee77 

> I hadn't realized that this new camera image was a bit smaller rather than (512, 512) 

Me neither, especially since the docstring still says (512, 512)! 😅 I actually think we should resize the new image to match the old one — changing the output size of the function seems like a big deal. I realise this will be super annoying @alexdesiqueira ... 😬 So I definitely won't insist on it, but worth a shot — maybe not that many things will change since the histogram will be the same...

alexdesiqueira:
> I actually think we should resize the new image to match the old one — changing the output size of the function seems like a big deal.

I love you @jni, but I wish you've seen this sooner! :laughing: 
Check what `transform.resize(data.camera(), output_shape=(512, 512))` gives us:
Screenshot from 2020-09-17 17-24-44
I kinda don't like it since it stretches the image too much; feels poor quality to me.
Is that okay for you? If yes, I'll use it.

jni:
Oh no, we should definitely not change the aspect ratio: first crop to a square, then resize.

If only we had seam carving still! 😂

Do we have a higher-quality source image than what you've included so far?

jni:
CC @krvarshney — could we have a higher quality version of the source image?

alexdesiqueira:
> Oh no, we should definitely not change the aspect ratio: first crop to a square, then resize.

```
>>> data.camera().shape
(245, 327)
>>> cam = transform.resize(data.camera()[:, :245], output_shape=(512, 512))
```

Check it out @jni. We'd lose some structures on the right, but I think the image looks nice:

Screenshot from 2020-09-18 05-41-09

grlee77:
@alexdesiqueira , Is this (512, 512) version in c219a95 now cropped from a higher resolution source? I can regenerate the SSIM/PSNR values for you again assuming this is the version we are going to go with.

alexdesiqueira:
Hey @grlee77,
>  Is this (512, 512) version in c219a95 now cropped from a higher resolution source?

@jni was talking about using neural nets to improve the image quality, so I used [bigjpg](https://bigjpg.com/) to increase its size with an acceptable quality:
1. Cropped the original image to a square one (245x245);
2. Used bigjpg to increase the image to a 980x980;
3. Rescaled it to 512x512. :cold_sweat: 

How does this one feel?

>  I can regenerate the SSIM/PSNR values for you again assuming this is the version we are going to go with.

I was going to ask you that right now. Thank you for your help!

grlee77:
> @jni was talking about using neural nets to improve the image quality, so I used bigjpg to increase its size with an acceptable quality:
...
> How does this one feel?

I think that visually it looks pretty good. Did you not hear anything back regarding the higher resolution source or did they confirm it is unavailable? The image statistics may differ a bit from a natural image, but I don't think that should matter for use in the test cases and our demos.

alexdesiqueira:
>  Did you not hear anything back regarding the higher resolution source or did they confirm it is unavailable? The image statistics may differ a bit from a natural image, but I don't think that should matter for use in the test cases and our demos.

We didn't hear from them. I think this will be the best we can do with what we have right now.

alexdesiqueira:
The following tests in `filters/rank/tests/test_rank` are breaking:
```
FAILED test_rank.py::TestRank::test_compare_8bit_vs_16bit[autolevel] - AssertionError: 
FAILED test_rank.py::TestRank::test_compare_8bit_vs_16bit[equalize] - AssertionError: 
FAILED test_rank.py::TestRank::test_compare_8bit_vs_16bit[subtract_mean] - AssertionError: 
```
Not sure why... is this new image "broken" somehow? :slightly_smiling_face:

doc/examples/registration/plot_masked_register_translation.py
```diff
@@ -82,10 +82,10 @@
 image = data.camera()
 shift = (-22, 13)
 
-rr1, cc1 = draw.ellipse(259, 248, r_radius=125, c_radius=100,
+rr1, cc1 = draw.ellipse(79, 128, r_radius=75, c_radius=50,
                         shape=image.shape)
 
-rr2, cc2 = draw.ellipse(300, 200, r_radius=110, c_radius=180,
+rr2, cc2 = draw.ellipse(150, 100, r_radius=60, c_radius=90,
                         shape=image.shape)
```
jni, line 89: Do these need to be changed back? 😬

skimage/data/__init__.py
jni: Lovely update, thank you! 😊

doc/examples/registration/plot_masked_register_translation.py
alexdesiqueira: It'd be better, right? :sweat_smile:

skimage/restoration/tests/test_restoration.py
```diff
@@ -63,7 +61,7 @@ def test_image_shape():
     point[2, 2] = 1.
     psf = ndi.gaussian_filter(point, sigma=1.)
     # image shape: (45, 45), as reported in #1172
-    image = skimage.img_as_float(camera()[110:155, 225:270]) # just the face
+    image = util.img_as_float(camera()[110:155, 225:270]) # just the face
```
jni, line 64: Is this still just the face? 😂

alexdesiqueira: Probably not! :laughing:

jni:
> The following tests in filters/rank/tests/test_rank are breaking:
> Not sure why... is this new image "broken" somehow?

For reference, here's a link to the failing tests:

https://travis-ci.org/github/scikit-image/scikit-image/jobs/729487096#L4324-L4423

And the test itself:

https://github.com/scikit-image/scikit-image/blob/ebf8aa35052e0a0cc0a2d1a7a91f2e3b7ffaa6f2/skimage/filters/rank/tests/test_rank.py#L347-L357

I have to agree with @alexdesiqueira, this is absolutely bizarre. From my reading, we are revealing something is actually broken in the rank filters, and all we have done here is expose a pre-existing error. The failures are off by a maximum of 1, so I think it probably has to do with a `<`/`<=` somewhere. 

My suggestion for now is to relax the test to assert diff is at most 1, and create an issue linking to this comment that we need to investigate.

Comments from @scikit-image/core are welcome!

doc/examples/features_detection/plot_glcm.py
```diff
@@ -36,14 +36,14 @@
...
-grass_locations = [(474, 291), (440, 433), (466, 18), (462, 236)]
+grass_locations = [(153, 200), (175, 225), (200, 250), (175, 300)]
 grass_patches = []
 for loc in grass_locations:
     grass_patches.append(image[loc[0]:loc[0] + PATCH_SIZE,
                                loc[1]:loc[1] + PATCH_SIZE])
 
 # select some patches from sky areas of the image
-sky_locations = [(54, 48), (21, 233), (90, 380), (195, 330)]
+sky_locations = [(25, 25), (50, 200), (75, 250), (100, 275)]
```
jni, line 46: Last Q: are these correct now?

Other than that, this is ready, if we can get someone else to agree on what to do about the failing rank filter tests. Thank you @alexdesiqueira!!!

JDWarner:
I don't want to rain on this parade, but unless the need and timing is extremely urgent I suggest reshooting the proposed replacement image (reasoning to follow).  I'm not sure where this was shot, but it seems like the same field and buildings still exist so this could be done.  I am firmly :-1: on using super-resolution deep-learning techniques. The latter will introduce subtle artifacts that compromise its role as a test image.  Only a downsampled/downscaled version of a high resolution original should be considered.

We're talking about changing one of the fundamental test images widely used in image processing and our replacement should be held to the highest standard.  Test images are technically excellent images which show a diversity of regions of contrast and texture, chosen as likely representative of the majority of images.  The original meets that bar.

Beyond that, the framing and posing of this replacement is not great and in my humble opinion, makes is not as well suited as a test image.  The original has a narrow field of view, essentially taken by telephoto lens, with the camera taking the image far from the subject.  This allows the majority of the man - including his dark colored slacks - and also the majority of the tripod to be in the field of view, but critically the dark coated man does not dominate the field and the buildings take a significant proportion of the frame.  Different scales of detail are present, with multiple levels of contrast, and detailed regions probably take up 25%+ of the frame.

The new version is shot with a wider lens, from close by, resulting in de-magnification of the distant buildings and magnification of the cameraman.  A great deal of the valuable contrast from the background structures is lost.  This same effect results in the proposed new version also containing relatively less of the dark coated man in frame (no slacks), but the wider lens at closer range magnifies the cameraman and depending on where you crop it to square over 1/3 of the image area is the dark coat.  Certain features related to positioning are lost, for example, the dark gloved right hand should be positioned wholly over the coat, because maintaining contrast from near-black coat to a distinct slightly less dark region of the globe is important.  The flowing gradation in the original man's slacks is not present, which is a key textural feature.  The tripod branches out lower, so the linear foreground of the tripod does not intersect with the detail in the buildings, another key textural feature.  There is far less detail overall; the overwhelming majority of the image is relatively flat sky, dark coat, and grass.

In short, the original is a good _test image_ because of sharpness, contrast (at multiple scales), and a diversity of textural features.  The proposed modern take - even if we had the original and didn't need to super-resolution it - falls short technically and I'm uncomfortable presenting it as a replacement.

alexdesiqueira:
> I suggest reshooting the proposed replacement image (reasoning to follow). I'm not sure where this was shot, but it seems like the same field and buildings still exist so this could be done.

@JDWarner this image is from ~2006, I'm not sure the original photographer would be able to do that :sweat_smile: This is possibly the only version of that image.

>  I am firmly -1 on using super-resolution deep-learning techniques. The latter will introduce subtle artifacts that compromise its role as a test image. Only a downsampled/downscaled version of a high resolution original should be considered.
> (...)
> The proposed modern take - even if we had the original and didn't need to super-resolution it - falls short technically and I'm uncomfortable presenting it as a replacement.

I agree with and understand you Josh, and we raised these concerns during last meeting (this Mon) as well. However, we are discussing this image as a substitute for a while already — please check #3927 and #4914 —, and it seems that `cameraman`'s (lack of) copyright is a huge deal breaker for us.
Nevertheless, could you please suggest a better alternative? We are out of options here. :slightly_smiling_face: Or, @scikit-image/core are we still using this one as the current `camera` image? Please let me know.

JDWarner:
I like the idea of recreating the original and explicitly licensing it CC0, given we should maintain a meaningful result behind `data.cameraman()` - we just need to make sure it sets a similarly high standard.  I have a great deal of images I'd be happy to license CC0, some which are technically good enough to be considered test images, but I don't think I have any which would make sense named `cameraman`.

Do we know where these were taken?  I wager there are scikit-image users close by...

stefanv:
I think it's unlikely we're going to get this re-shot, but @jni can probably ask Lav for the original and/or how easy it is to get this redone.

Josh, I see your points but I don't think these old test images were ever that amazing; they're pretty low resolution, and often folks scanned them in from photos, magazines, etc.

I agree that we can come up with better test images and, by all means, we should add those.

We've had similar discussions around the texture images we used. In that case the author did go to a lot of trouble to get those images perfect, so those are harder to replace.

JDWarner:
In case it helps, after some digging I am fairly sure the cameraman image was shot from Briggs Field at MIT looking almost directly Northeast toward the Great Dome and central campus.

I have one connection at MIT I will reach out to.  If anyone else has connections it might be fruitful to getting a more ideal reproduction.  I'd be happy to advise anyone about how to shoot it.

jni:
I'm with @stefanv. I don't think it's that important to get this perfect. Though I admire you for choosing such a flat hill as this field to die on @JDWarner! 😂 (👋 btw!) People complained when we replaced Lena with Astronaut, with similar points, but I don't think that we have set back the field of computer vision all that much. People are using these images for cute demos, not to demonstrate state of the art image processing — that ship has sailed, and big datasets like BSDS, imagenet etc. are now the norm. None of our documentation depends on the factors that you mention, @JDWarner.

We also already have a better-quality test image in `data.eagle()`. My suggestion to move forward is to add notes in the docstring that: (a) this image was generated with the procedure described by @alexdesiqueira, so it might exhibit artifacts (incidentally, it would be then an interesting case for methods that attempt to *detect* such artifacts) and (b) data.eagle is to be preferred as a test image in 2020; and then merge this. We can always revisit our test set for skimage 1.0.

JDWarner:
Lenna -> Astronaut was actually an upgrade from the standpoint of a test image in my opinion.  That's probably now among if not the single best test image we currently ship, actually.

Individual image characteristics matter less than they once did, especially for deep learning work and other capitalized "Big Data" projects.  But these images do get heavy use in image processing publications, and their use isn't just cute - for or better or worse these sort of test images are quite important in the review process.

Convolutional neural net assisted image reconstruction is trying to enter the medical imaging space, and it's outright dangerous.  I am quite familiar with super-resolution techniques and they can look great, but they are _not_ real.  I realize this was thought to be a deal breaker at the time of #3927 , but if we're considering dealing in disclaimers, why not just keep the original and disclaim with the NC clause and whatever license MIT requires?

stefanv:
Please don't use any kind of strange algorithm on the image; this will only introduce inexplicable artifacts in outputs. If anything, we should see if the original is available in RAW.

stefanv:
Josh, unfortunately fetching the original cameraman with Pooch is not an option, since we are still redistributing copyrighted material when we have agreed to only use freely modifiable images in our data package.

jni:
Ok so, after a private conversation with @stefanv, the decision is that we will either wait for @lrvarshney or @krvarshney to provide a higher-resolution version, or we find a different image, with preference for the former. Sorry for the churn, @alexdesiqueira! But at least now you can relax for a bit while we wait. :grimacing:

sciunto:
For the second option, we have the eagle (just a note for those who did not necessarily followed all the recent PRs).
See https://scikit-image.org/docs/dev/auto_examples/data/plot_general.html#sphx-glr-auto-examples-data-plot-general-py

stefanv:
> Ok so, after a private conversation with @stefanv, the decision is that we will either wait for @lrvarshney or @krvarshney to provide a higher-resolution version, or we find a different image, with preference for the former. Sorry for the churn, @alexdesiqueira! But at least now you can relax for a bit while we wait.

Not that we get to make decisions, but that's our recommendation :)

alexdesiqueira:
@scikit-image/core pardon my language everyone, but we re-deciding anything at this point kinda feels like a lack of respect of one's work. We invested a substantial amount of time discussing what image should be used, with _several_ core devs involved (please see #3927 and the other PRs after that); we decided this should be the image used, and I put the pressure to have this PR ready for 0.18. I've rewritten this PR _twice_ already, involving more people's work, reviews and help (thanks everyone!), as you can read throughout this PR. Many person-hours were invested here. I agree with all the points raised — I even raised some of them myself —, but when we say we decide something, we have to have some backbone to go with that up to some place, as bad as this decision seems to be — even more when we consider someone else's work. I personally don't care that much, and please don't let this influence your stand on this PR, but please let this be an example on how we need to treat the work of other contributors.
Love you all :heart: :blush: And some of you owe me a beer! :joy: :wink:

stefanv:
Alex, sorry: I think there must have been a misunderstanding. The suggestion (at least from my side) was simply to ask the photographer for a higher resolution photo (he is taking a look).

Perhaps I failed to take into account work that would result from using a higher resolution image, but that was not the intention.

alexdesiqueira:
@stefanv the comment wasn't aimed to the final decision :slightly_smiling_face: please don't take it personally.

>  The suggestion (at least from my side) was simply to ask the photographer for a higher resolution photo (he is taking a look).

@jni proposed this before in the thread, and I fully agree with it. We agreed to wait until Mon PST evening for an answer, though. I didn't want to get in the way of 0.18 with this. We (Juan and I) proposed the super resolution using DL, and I came with the solution. That was my fault, I take total responsibility and apologize for it. However, we've been discussing this — the image, how to do it, how to solve the issues — for quite a while already.

The issue for me is that, IMHO, our process of contributing fails when `i` (`i` being a large number, in this case) people discuss a strategy and the `i+1` person can pull the plug and stop the machines on something that was already discussed and in the end already. This happened twice only in this PR (@jni asked for a different size, @JDWarner :-1: the image. Love you guys, no hard feelings at all; just stating my point :slightly_smiling_face:).

For Odin, I personally didn't even like the image :joy: However, we (as a group) proposed other alternatives and decided on this one. I was more inclined to use something like [this one](https://www.flickr.com/photos/wyncliffe/13184340485/in/photolist-m64c6g-Em5r69-7vTNoP-7vWip6-7vYEkh-DQtK3Q-EeHwkV-7vUvPz-DQtHds-N6Z6Kb-N5DTnQ-HU1XHm-4jYwkq-YrYtxv-2erywby-2fQvkh6-oa8UP4-o8oEoN-o8oyaW-nSVCBd-nSVWjf-nSVzYC-occ3Te-ocbKWP-4GD3dQ-nSWCL9-nSWUX5-oa8xGK-nSW6jm-oahSg3-oaqm6t-occbpM-nSVL69-oa7fi8-4wXFJ8-nSX1jC-nSWDim-oairFQ-nSWuc5-nSWsrw-oa8EVx-nSWqMC-oajUW7-ocbGpZ-FMEzhS-JHZweF-JYBPFP-23vdtmM-A4aZNA-LiHU64) @sciunto suggested — imagine the examples on textures and such! However, we are discussing the clarity on what we ask on our issues recently, and if I may point y'all to https://github.com/scikit-image/scikit-image/issues/4914#issuecomment-674540222, we decided on 1. what work should be done and 2. when it would need it to be done as well (as in **"Milestone: 0.18"**; label **"priority: critical"**).

Again, _please_ don't take this as a lament directed to anyone in special. I feel we could do better in the future as a group. Sorry if this is getting too heavy; I don't want y'all to leave this with feeling bad or angry at me, so please check out [this red panda](https://giphy.com/gifs/adorable-red-panda-woodland-park-zoo-UQ7uIiy1BYNox0F0l7) to lift your spirits a little bit :slightly_smiling_face: 

red_panda

stefanv:
What's the copyright on that panda image. Looks pretty good ;)

No offense taken, Alex, and sorry for piping up at the last minute. Too many emails nowadays and hard to keep up.

What can I say: consensus seeking can be frustrating, and requires patience and kindness. You are right: our contributors are our most important assets and we should treat them as such.

There is a fine balance with these things: you want to move them forward in a pragmatic way, but you also want to do the best job you can under the circumstances.

So, it comes down on when to push back hard. I think we can afford pushing back a bit less and letting PRs flow more easily. The only reason I would be a bit cautious in this case is because the decision really is quite irreversible (not entirely, but pretty much).

As such, I'm asking no more work, just a little patience.

I loved Josh's analysis of the image (gotta admit, he has a fine eye for these things!) but I also disagree with him that those factors are important here (I feel we can easily add additional images later and it would be *great* if someone went with lenses and tripods and did a special capture that mimics the original scene---holiday project, any East Coast person?).

So, long way to say: I appreciate all your efforts, I think the SR approach was a neat and creative idea but perhaps a bit too adventurous for what we do, and if you can kindly hold off just a tiny while until we hear back from the photographer, hopefully we'll close this down rapidly.

JDWarner:
Alex, I do apologize.  My time for the package has not been ideal due to working in medicine and other pressures in this strange era.  I briefly chimed in way back during the initial discussion and try to do a review of larger things before they get merged.  I've been taking a closer look as 0.18 is coming up soon, and this came back to my attention yesterday.  Be assured it has nothing to do with disrespect for anyone's time or work.

I've seen the original cameraman in quite a few papers I have personally reviewed, and know the image well.  The proposed replacement is inferior, even if we had a native higher resolution version.  

It seems satisfying release pressure is conflicting with the ability to ensure we ship the best test image.  Those competing priorities are unfortunate.  **Open question**: I expect someone will be able to produce a more faithful and superior test image based on the original cameraman before 0.19.  Is the pressure on this one issue such that we want to 
1. Ship an inferior test image, or
2. Ship an inferior test image and likely change it yet again in the next major point release, or
3. Keep the original cameraman for now, skip a major point release on Debian, and provide a definitive one-time change in 0.19?

krvarshney:
All of the original full resolution cameraman images that @lrvarshney and I took in February 2006 are now available here under CC0 license: http://krvarshney.github.io/cameraman.zip

alexdesiqueira:
@JDWarner don't worry at all :slightly_smiling_face: I understand life is not easy these days, and hope you're safe out there.

I'm very fond of the third point you put:

> Keep the original cameraman for now, skip a major point release on Debian, and provide a definitive one-time change in 0.19?

Maybe we could solve this with an [I cut, you choose](https://en.wikipedia.org/wiki/Divide_and_choose) strategy, Josh :joy: @krvarshney provided all images they had (thank you very much!):

> All of the original full resolution cameraman images that @lrvarshney and I took in February 2006 are now available here under CC0 license: https://krvarshney.github.io/cameraman.zip

Could you please check if is there an image from this package that suits us? I think the closest one would be `IMG_0172.JPG`. If not, how about one of @sciunto's alternatives on #4914?

Thank you for your help!

JDWarner:
All right, thank you very much @krvarshney! 

I had a look through these, made minor rotation to align the horizon and then a couple square cropped versions of what I think are the two best options (all tweaks with GIMP 2.10, linear mode image scaling after crop, lossless PNG output).  This one was the original proposed replacement, now at 512x512 with minor tweaks and revised square crop (derived from IMG_0178.jpg):
IMG_0178_linear

However, this one would be my preference (derived from IMG_0167.jpg, which was a color image; Luma mode desaturation used otherwise same methods).  It is from a location closer to the campus, with the photographer and buildings more balanced and better overall detail in frame.  The original IMG_0167 had the buildings in background out of focus, but with linear rescaling during a 4x reduction in linear dimensions the whole frame looks sharp.  This, to me, serves the purpose best and I would be comfortable with its use as a test image:
IMG_0167_linear

I release these derivatives also under CC0.

grlee77:
> However, this one would be my preference (derived from IMG_0167.jpg, which was a color image; Luma mode desaturation used otherwise same methods)

Thanks @JDWarner. This version looks a bit better to me as well. I did not think our intention was to advocate for the image's use as a test image for academic publications, but rather just as a reasonable image for use in our demos and test suite. It is nice to have one that is suitable for both purposes, though.

I agree that the current cameraman image is widely used in the image processing literature, but I am sure most publications using it did not obtain it via scikit-image and it will still be easily obtained elsewhere for those who want it for comparison to previously published values on the same image.

stefanv:
Nice, thank you @JDWarner!  Thanks for being so pragmatic and helpful on this.

alexdesiqueira:
> However, this one would be my preference (derived from IMG_0167.jpg, which was a color image; Luma mode desaturation used otherwise same methods). (...) This, to me, serves the purpose best and I would be comfortable with its use as a test image

Thank you @JDWarner, it looks great! I will use this one then, if that's okay with everyone.
@grlee77 could you help me calculating the coefficients for this image? Thank you!

grlee77:
> @grlee77 could you help me calculating the coefficients for this image? Thank you!

Yes. Thank you for your persistence on this PR!

If I use `imread` on the image saved from @JDWarner's comment I get a (512, 512, 4) image where the first 3 channels are identical and the alpha channel is 255 everywhere. I have attached here a version where I extracted channel 0 and saved to a new PNG via `imsave`.
camera_newest

Using this (512, 512) image I get the following values for use in the test cases

**skimage v0.17.2:**
MSSIM = 0.35796369678883744
PSNR = 22.409352228719225

**IPOL:**
MSSIM: 0.357958555221558
PSNR: 22.409352306975116

JDWarner:
Sorry about the alpha channel, GIMP was set to automatic rather than explicit RGB and for... reasons... it defaulted to 8 bit RGBA despite no information being in the alpha channel.

Use the last version posted by @grlee77.

jni:
I've had chats with @stefanv, @JDWarner, and @alexdesiqueira and we've come to a consensus on the second option, based on IMG_0167. If anyone objects, speak now (within about 12h I reckon 😂) or forever hold your peace.

Honestly I do like the variety of textures in this new option. @alexdesiqueira next time we meet I promise we'll make all the pain worth your while!!! 😂

alexdesiqueira:
> @alexdesiqueira next time we meet I promise we'll make all the pain worth your while!!! :joy:

:beers: :beers: :beers: :laughing:

alexdesiqueira:
It seems that this image doesn't return Otsu == MultiOtsu. There are some strange things on the dask tests as well. I wasn't having these issues before...

```
============================================================================== short test summary info ==============================================================================
FAILED test_thresholding.py::test_threshold_minimum - AssertionError: 
FAILED test_thresholding.py::test_bimodal_multiotsu_hist - assert 102 == array([103])
FAILED test_thresholding.py::test_thresholds_dask_compatibility[threshold_otsu-90-92] - ValueError: operands could not be broadcast together with shapes (254,) (255,)
FAILED test_thresholding.py::test_thresholds_dask_compatibility[threshold_isodata-90-92] - ValueError: operands could not be broadcast together with shapes (254,) (255,)
```

JDWarner:
Most errors seem to be around histogram dependent functions and the actual range of data values in this particular crop at this exposure - because I deliberately did absolutely no exposure tweaking - is [1, 255] and not [0, 255].  If you load it up and check `im.min()` you get 1.  The legacy cameraman scan image did have values spanning the full 8-bit range of [0, 255].

I will find some 1 values in the image and set one to 0, and post that version.  I'm pretty sure that will sort most if not all of these failures.  

Hopefully these issues are only due to test assumptions with the prior cameraman using the full range.  We should of course gracefully support this sort of thing, because far from all images are perfectly exposed.

JDWarner:
Turned out there was exactly one pixel == 1 in the whole image, but it was surrounded by several == 2 (the area was shadow between the lower two coat buttons).  The pixel which was 1 is now 0, and a couple surrounding pixels have been made 1.  The histogram is now full range, as the prior cameraman was.  Hopefully this resolves the remaining failures.

Felt the need to do this manually as curves or other multiplicative exposure tweaks or histogram equalizations generally create histogram abnormalities/gaps elsewhere we do not want.

cameraman_fullrange

JDWarner:
As a couple pixels have minimally changed, we may need new SSIM coefficients @grlee77.

alexdesiqueira:
Thank you very much @JDWarner! I'll check what we have now.

alexdesiqueira:
Kindly pinging @grlee77 :slightly_smiling_face: Could you help us with the coefficients for the latest image? Thanks!

grlee77:
> Kindly pinging @grlee77 slightly_smiling_face Could you help us with the coefficients for the latest image? Thanks!

Here you go!

**IPOL results**
SSIM: 0.357959091663361
PSNR: 22.409353363576034

alexdesiqueira:
All green again :tada: Gotta point the first reviewers one more time (@emmanuelle @rfezzani @jni). Thanks again @grlee77 and @JDWarner for all your help!

jni:
Whoo! I updated plot_glcm with the new sky and grass locations. The other example looks ok, not great, but master also looks ok, not great, so I say leave it for now. My approval from before still stands. Thank you everyone who has contributed to this (a lot of people now! 😅) and particularly @alexdesiqueira for the hard work here! I'm really happy with the end result.

jni:
@scikit-image/core the travis failure is unrelated and should not hold up a merge. =)

skimage/data/__init__.py
sciunto: We need an update here :)

sciunto:
there is one line to be changed before merging.

rfezzani (approved):
Thank you @alexdesiqueira for your hard work on this :clap:. Since the CI is green now, I approve this PR, but I think that we need to improve some gallery [examples](https://628-2014929-gh.circle-artifacts.com/0/doc/build/html/api/skimage.data.html#examples-using-skimage-data-camera) involving `data.camera`. Some examples where tuned for the old image. To illustrate this, please see:
- Multi-Otsu Thresholding: [new](https://628-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/segmentation/plot_multiotsu.html#sphx-glr-auto-examples-segmentation-plot-multiotsu-py) vs [old](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_multiotsu.html#sphx-glr-auto-examples-segmentation-plot-multiotsu-py)
- Markers for watershed transform: [new](https://628-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/segmentation/plot_marked_watershed.html#sphx-glr-auto-examples-segmentation-plot-marked-watershed-py) vs [old](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_marked_watershed.html#sphx-glr-auto-examples-segmentation-plot-marked-watershed-py)
- Flood Fill: [new](https://628-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/segmentation/plot_floodfill.html#sphx-glr-auto-examples-segmentation-plot-floodfill-py) vs [old](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_floodfill.html#sphx-glr-auto-examples-segmentation-plot-floodfill-py)
- Rank filters: [new](https://628-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/applications/plot_rank_filters.html#sphx-glr-auto-examples-applications-plot-rank-filters-py) vs [old](https://scikit-image.org/docs/stable/auto_examples/applications/plot_rank_filters.html#sphx-glr-auto-examples-applications-plot-rank-filters-py)

But this can be addressed in different PRs :wink:

sciunto:
> 
> But this can be addressed in different PRs 

+100

alexdesiqueira:
> But this can be addressed in different PRs :wink:

Sure thing :) I will open an issue to deal with them later; my repo is breaking already, I'd need a rebase... :joy:

sciunto:
looks like there are some issues with checksums on azure, but not on travis...

alexdesiqueira:
> looks like there are some issues with checksums on azure, but not on travis...

@sciunto the errors are due to last commit, since we don't have that image on base yet (please see [3f24084](https://github.com/scikit-image/scikit-image/pull/4913/commits/3f24084a9f9d27a36897a2258c25bca8468aaa71)).

emmanuelle:
Yes a8cc0a9 was green.

sciunto:
Of course... I'm stupid.

alexdesiqueira:
Thanks everyone, we made it! :tada:

hmaarrfk:
awesome job!

stefanv:
Thank you very much, @alexdesiqueira !

Issue #4914, opened 2020-08-13
Proposing a substitute for cameraman


alexdesiqueira:
## Description

Hey @scikit-image/core,
I am dealing with #3927 on #4913. We wanted a substitute for cameraman for a while, and I propose this image:

![eagle](https://user-images.githubusercontent.com/5776022/90188433-4dd37200-dd70-11ea-9deb-3a1ac195c49a.png)

@daymachado took it; she's releasing it under CC0.
What do you think? Is it a good substitute?

alexdesiqueira:
Its histogram:
![histogram_eagle](https://user-images.githubusercontent.com/5776022/90192854-a27aeb00-dd78-11ea-8f6d-62c247e0f7ca.png)

stefanv:
I like it! Although, perhaps that should become `eagle`, and cameraman should go with depreciation?

matthew-brett:
How about http://krvarshney.github.io/kush_varshney.png as suggested in https://twitter.com/stefanvdwalt/status/1136039155119472640

alexdesiqueira:
@matthew-brett thank you for remembering that discussion, Matt! That's indeed a nice candidate as well.
How could we proceed, Stéfan?

- Would you like to use @krvarshney's camera (http://krvarshney.github.io/kush_varshney.png) as `camera.png`?
- If yes: how about `eagle`? Would that still be a good fit?

stefanv:
@krvarshney's is a good replacement for camera, then we can add eagle separately. Presuming that is an eagle :)

sciunto:
I see in @alexdesiqueira image an eagle glaring the cameraman who is imposing a NC usage of himself. :)

alexdesiqueira:
We can always name it `eagle_or_hawk` :wink:

jni:
`or_falcon`!

How about `data.bird_of_prey_of_some_sort`? =P

jni:
@alexdesiqueira The more I think about it the more I love that picture. In addition to being beautiful, it seems so great for so many examples — contrast adjustment, circular and line hough transforms, corner detection... Probably multi-otsu thresholding...! So thanks for the nomination!

hmaarrfk:
I thought the conclusion from the whole MIT cameraman copyright was that they were OK with us using it. We mostly have to document their emails.

sciunto:
> How about `data.bird_of_prey_of_some_sort`? =P

[...]

PR #4922, opened 2020-08-16
Add new high resolution grayscale image data/eagle.png

alexdesiqueira:
## Description

Adding `data/eagle.png`. Please check the discussion on #4914.

## Checklist

- [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py) :heavy_check_mark: 
  existing benchmark
- Unit tests :heavy_check_mark: 
- Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/) :heavy_check_mark: 

## For reviewers

  later.


emmanuelle:
Thanks @alexdesiqueira. Can you please add the image to the registry https://github.com/scikit-image/scikit-image/blob/master/skimage/data/_registry.py? Also the image is large enough (2.3 Mo) that maybe we want to put it in the Gitlab repo rather than increasing the package size by 2.3 Mo? Please see the doc section in https://scikit-image.org/docs/dev/contribute.html?highlight=pooch

sciunto:
+1 with @emmanuelle

sciunto:
The image is merged in gitlab.

alexdesiqueira:
Thank you @sciunto!

sciunto:
Green lights, I approve!

skimage/data/tests/test_data.py
```diff
@@ -68,6 +68,14 @@ def test_coffee():
     data.coffee()
 
 
+def test_eagle():
+    """ Test that "eagle" image can be loaded. """
+    path = fetch('data/eagle.png')
```
emmanuelle, line 73: wouldn't `eagle = data.eagle()` work here?

emmanuelle:
Could you please add this image to https://github.com/scikit-image/scikit-image/blob/master/doc/examples/data/plot_general.py ?

sciunto:
> Could you please add this image to https://github.com/scikit-image/scikit-image/blob/master/doc/examples/data/plot_general.py ?

Good point! :)

skimage/data/tests/test_data.py
sciunto: In this case, the test `test_cells_3d` must be changed as well...

sciunto: However, my guess was to avoid to use the `_load` function to test more directly the image.

alexdesiqueira: Yeah, I tried to be consistent with `test_cells_3d`. I could change both if you'd like @emmanuelle.

emmanuelle: no pb, it's fine for the test.

jni (approved):

skimage/data/tests/test_data.py
emmanuelle: the `test_cells_3d` test is a bit different because there is no corresponding function for the dataset in `skimage.data`. Here more code would be tested by calling `data.eagle()`. But it's not very important. the PR could also be merged as it.

alexdesiqueira: I can change only this one, then :slightly_smiling_face:

emmanuelle:
Thanks @alexdesiqueira ! I'll merge when the CI has run.

alexdesiqueira:
Thank you for your help @emmanuelle and @sciunto!

Issue #4911, opened 2020-08-12
OpenMP build not supported for MSVC


FirefoxMetzger:
## Description

While working on https://github.com/scikit-image/scikit-image/pull/4851 I noticed that I got compiler warnings about the flag `-fopenmp` being unknown and, accordingly, didn't have openMP working. I'm compiling with msvc, so in my case, the flag should be `/openmp`.

I've updated the setup.py to assign the correct flag if the compiler is `msvc`:
https://gist.github.com/FirefoxMetzger/b10a215204732ebf0a2cc34b15b89039
(in particular line 99ff.)

This works in my case, but I didn't think about it too deeply. If there is interest to incorporate this change, I'll open a PR.

## Way to reproduce
```
python setup.py build_ext
```
Somewhere in the build logs:
```
cl : Command line warning D9002 : ignoring unknown option '-fopenmp'
```

## Version information
```python
>>> import sys; print(sys.version)
3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)]
>>> import platform; print(platform.platform())
Windows-10-10.0.17134-SP0
>>> import skimage; print("scikit-image version: {}".format(skimage.__version__))
scikit-image version: 0.18.dev0
>>> import numpy; print("numpy version: {}".format(numpy.__version__))
numpy version: 1.19.1
```

emmanuelle:
Thanks @FirefoxMetzger ! A PR solving this would definitely be welcome!

PR #4924, opened 2020-08-17
OpenMP support in MSVC

FirefoxMetzger:
## Description
Related Issue: https://github.com/scikit-image/scikit-image/issues/4911

## For reviewers

  later.


setup.py
```diff
@@ -57,15 +57,12 @@
 def openmp_build_ext():
     from numpy.distutils.command.build_ext import build_ext
 
-    compile_flags = ['-fopenmp']
-    link_flags = ['-fopenmp']
-
     code = """#include <omp.h>
     int main(int argc, char** argv) { return(0); }"""
```
FirefoxMetzger, line 61: This could be refactored further and we could remove the function entirely

emmanuelle: could you please explain how?

emmanuelle:
Thanks @FirefoxMetzger ! Our Appveyor CI platform uses Windows so maybe it's possible to add a test which would check that extensions were compiled with openmp? The easiest check is maybe to add a test in your rolling ball PR to be sure that the code runs faster with 2 threads than one...

setup.py
FirefoxMetzger: @emmanuelle See https://github.com/scikit-image/scikit-image/pull/4924/commits/085e8008e345676c9b78c17eacafb657b2b4aa08

Essentially we group the import with the remaining imports, make the code string local to the class, remove the wrapping function and pass the class to `build_ext`

FirefoxMetzger:
@emmanuelle The problem I see with such a test is that it would be inconsistent. If there is no OpenMP then both versions (with one thread and with many) compile to exactly the same code, and it becomes a coin toss if the test will fail or pass; we would obviously like it to always fail. Similarly, the test can fail on a whim if there is OpenMP, but the computer only has a single core (e.g. a VM).

What might be done is to specify `\WX` as a compiler flag (and potentially linker flag), which will treat warnings as errors and will fail loudly at compile time.

If that is not an option, there might be an angle via `cimport openmp` which may fail at compile time if it can't find OpenMP, but I didn't think this through at all so that might not work.

mkcor:
Poking @sciunto and @lagru because I've seen you on infrastructure type of issues. Not sure: I'm trying to do my part as the person in charge this week! Please let me know if this move wasn't relevant.

jni:
@FirefoxMetzger man, the git diff seems to have a lot of noise, with lines that changed in indentation but not semantics... Could you summarise the changes here? I would actually encourage you to add comments in the code itself to make it clearer to those of us who haven't really followed this compiler logic. Is this class structure standard? If so, can you link to the distutils doc that mentions it?

Also NumPy distutils is [getting deprecated](https://github.com/numpy/numpy/issues/17620), so it would be good to add a comment linking to the above issue above the import.

FirefoxMetzger:
There is two changes:

(1) I've added a check for the compiler being used (msvc vs other) and depending on the used compiler I set the appropriate OpenMP flag.

(2) I refactored the script. The function `openmp_build_ext` was wrapping the class `ConditionalOpenMp` and returning it when called (the class not the instance). I removed the wrapping function because it wasn't needed and saves one level of indentation. I also did some other basic gardening tasks like choosing descriptive variable names where applicable.

jni:
Thanks @FirefoxMetzger 

> (2) I refactored the script. The function openmp_build_ext was wrapping the class ConditionalOpenMp and returning it when called (the class not the instance). I removed the wrapping function because it wasn't needed and saves one level of indentation. I also did some other basic gardening tasks like choosing descriptive variable names where applicable.

In general, it's preferable to keep these things as separate commits or even separate PRs, because they make review more difficult. It's certainly contributed to why this PR has sat here for so long when it really only adds an if/else check! That core is very simple and should have just been shooed in. Sorry that it's taken so long!

jni (approved):

jni:
... And it also makes it more likely to result in merge conflicts. 🤦 I don't trust myself to do the right thing with this one (#5059), could you have a look @FirefoxMetzger? 🙏

FirefoxMetzger:
@jni Yes, I was planning to refactor the code and make it cleaner (+ resolving the merge conflict). I remember encountering the merge conflict a while ago on my local version. I have now forgotten why I didn't push the change here back then.

I'm currently stuck in another project (parsing audio data and converting it into a more standard format), which is surprisingly fiesty and stubborn, i.e., it just doesn't work as I want it to (yet). Once I fixed that, my plan is to fix this PR and then look at the new API over at `imageio`. My current aim is to have it done by the end of the week. Then again, I was thinking this last week, too.

FirefoxMetzger:
Until we manage to resolve #5091 I won't be able to test the test failure on 3.9. The error seems a bit arbitrary though, as if the `expected` array doesn't get parsed to a numpy array but remains a binary string.

The docs seem to have timed out during build; could somebody manually trigger the tests again (@jni ) to check if it replicates? (It timed out in other current PRs, too)

emmanuelle:
@FirefoxMetzger I restarted the doc build but as you say it happens on many PRs so don't worry about it.

Issue #4925, opened 2020-08-17
`width` and `height` swapped in `morphology.rectangle`


FirefoxMetzger:
https://github.com/scikit-image/scikit-image/blob/20297dca8c797ca91ad515d0cd8d6f738a4e8b9e/skimage/morphology/selem.py#L57

I just noticed that the rectangle structuring element behaves inconsistently. The line should be:

```
return np.ones((height, width), dtype=dtype)
```

or the variables should be renamed.

jni:
@FirefoxMetzger This is a duplicate of #4125. See the discussion in the #4126 (stalled PR) for the proposed fix. We'd certainly welcome the proper deprecation fix!

In the meantime I'll close this as duplicate, though. =) Thank for reporting!

PR #1280, opened 2014-12-13
WIP: Describe array order terminology in 2D, 2D+c, 3D, 3D+c arrays

jni:
Still working on it, but just thought I'd check this in so that we can tag it as 0.11.

PR #4926, opened 2020-08-17
swap rectangle variable names

FirefoxMetzger:
## Description
Closes: #4125
Related: #4925, stalled PR #4126

@jni suggested an alternative signature of the form `rectangle(nrows, ncols, dtype=np.uint8, *, height=None, width=None):` combined with a depreciation warning (in #4126 ). I can change to this signature; I don't understand how `height=...` and `width=...` will ever be triggered since the original arguments weren't keyword arguments. 

`nrows` and `ncols` is a valid candidate for naming (not sure if it is as/more intuitive than `height` and `width` though). Is there a larger effort to switch to this nomenclature?

## For reviewers

  later.


grlee77:
>Is there a larger effort to switch to this nomenclature?

Yes. I was not too involved in the effort, so perhaps others will comment more, but I believe much of the relevant past discussion is in #1170 and implementation in #1280

alexdesiqueira:
Thanks for that, @FirefoxMetzger!

> @jni suggested an alternative signature of the form rectangle(nrows, ncols, dtype=np.uint8, *, height=None, width=None): combined with a depreciation warning (in #4126 ). 

I'm +1 for `rows`, `cols` or `nrows`, `ncols`, but I'd ask for @jni to weigh in. This seems like a task for the Guardian of Dimensions :wink:

skimage/morphology/selem.py
```diff
@@ -29,18 +29,18 @@ def square(width, dtype=np.uint8):
     return np.ones((width, width), dtype=dtype)
 
 
-def rectangle(width, height, dtype=np.uint8):
+def rectangle(nrows, ncols, dtype=np.uint8):
```
jni, line 32: @FirefoxMetzger this also needs the new deprecated, keyword-only arguments mentioned in the earlier issues. The reason is that, even though the values work as positional arguments, users could have used:

```python
rectangle(width=5, height=8)
```

and that code would now break.

skimage/morphology/selem.py
```diff
@@ -29,18 +29,18 @@ def square(width, dtype=np.uint8):
...
     belongs to the neighborhood.
 
     Parameters
     ----------
-    width : int
-        The width of the rectangle.
     height : int
         The height of the rectangle.
+    width : int
+        The width of the rectangle.
```
jni, line 43: These arguments need to be renamed in the docstring

jni (changes requested):
@FirefoxMetzger thanks for tackling this! See comments below — we still need the deprecated (and incorrect) arguments for backwards-compatibility. They should emit a deprecation warning if used, then assigned to their historical values (`nrows=width`, `ncols=height`). And yes, we are indeed moving to a rows/columns nomenclature throughout the library.

skimage/morphology/selem.py
FirefoxMetzger: > rectangle(width=5, height=8)

I actually didn't believe this was possible until I tried it in a shell xD In retrospect, it makes a lot of sense that something like this would be possible in python, but I never even conceptualized this and - accordingly - didn't quite understand why such a test would be necessary. 

I guess you always learn something new; thanks for that.

emmanuelle:
Thanks @FirefoxMetzger ! Could you please add a line in the release notes (in the API changes section) `doc/releases/release_dev.rst` ?

FirefoxMetzger:
Done. I didn't find the location where the docs generate the file (didn't search very hard), so I didn't check how it renders. @emmanuelle Is it generated when building the docs?

skimage/morphology/selem.py
```diff
@@ -29,31 +29,67 @@ def square(width, dtype=np.uint8):
...
+        warnings.warn("The usage of width is depreciated and the argument will"
+                      " be removed in a future version. Use"
+                      " `rectangle(nrows, ncols)` instead.",
+                      DeprecationWarning)
+
+    if height is not None:
+        ncols = height
+
+        import warnings
+        warnings.warn("The usage of height is depreciated and the argument"
```
jni, line 82:
```suggestion
        warnings.warn("The usage of height is deprecated and the argument"
```

skimage/morphology/selem.py
```diff
@@ -29,31 +29,67 @@ def square(width, dtype=np.uint8):
...
+        import warnings
+        warnings.warn("The usage of width is depreciated and the argument will"
+                      " be removed in a future version. Use"
+                      " `rectangle(nrows, ncols)` instead.",
+                      DeprecationWarning)
+
+    if height is not None:
+        ncols = height
+
+        import warnings
```
jni, line 81: Move this import to the top of the file?

skimage/morphology/selem.py
```diff
@@ -29,31 +29,67 @@ def square(width, dtype=np.uint8):
...
     """
+
+    if width is not None:
+        nrows = width
+
+        import warnings
+        warnings.warn("The usage of width is depreciated and the argument will"
+                      " be removed in a future version. Use"
+                      " `rectangle(nrows, ncols)` instead.",
+                      DeprecationWarning)
```
jni, line 76: We use `FutureWarning` by default because DeprecationWarnings are usually not displayed by Python.

jni (changes requested):
@FirefoxMetzger thank you! Just a few more changes! 😬

skimage/morphology/selem.py
```diff
@@ -29,31 +29,67 @@ def square(width, dtype=np.uint8):
...
+        The number of columns of the rectangle.
 
     Other Parameters
     ----------------
     dtype : data-type
         The data type of the structuring element.
+    width : int, depreciated
+        The number of rows of the rectangle.
+    height : int, depreciated
+        The number of columns of the rectangle.
```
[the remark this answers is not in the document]
FirefoxMetzger: I'm not sure about this suggestion, @jni . Your suggestion would mirror the docstring for the old version, but that one is faulty (as per #4125 ). Should we keep the wrong documentation, or update the docs to match what actually happens?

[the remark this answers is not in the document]
jni: oooooooh! I take it back. You're right, but I do prefer (height width) as an order, so I would additionall change rows/columns between the two descriptions.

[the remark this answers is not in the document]
FirefoxMetzger: Can be done. Should we then also switch the order of the keyword arguments to `rectangle(..., height=None, width=None`)?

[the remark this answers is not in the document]
rfezzani: Jumping into the discussion! Can't we use the `skimage._shared_.utils.deprecate_kwarg` decorator here? It avoids the side note and the explicit managements of `None`!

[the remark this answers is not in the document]
FirefoxMetzger: @rfezzani nice suggestion. I updated the PR.

skimage/morphology/selem.py
```diff
@@ -29,32 +32,46 @@ def square(width, dtype=np.uint8):
...
 
     Other Parameters
     ----------------
     dtype : data-type
         The data type of the structuring element.
+    height : int, depreciated
+        The number of columns of the rectangle.
+    width : int, depreciated
+        The number of rows of the rectangle.
```
rfezzani, line 57:
```suggestion

```

skimage/morphology/selem.py
```diff
@@ -29,32 +32,46 @@ def square(width, dtype=np.uint8):
...
         A structuring element consisting only of ones, i.e. every
         pixel belongs to the neighborhood.
 
+    Notes
+    -----
+    ``nrows`` and ``ncols`` have None specified as a default value. This exists
+    for backwards compatibility with scikit-image v0.18 and older. You must
+    specify values for both variables, and an exception will be raised
+    otherwise.
+
```
rfezzani, line 71:
```suggestion
```

skimage/morphology/selem.py
```diff
@@ -29,32 +32,46 @@ def square(width, dtype=np.uint8):
     return np.ones((width, width), dtype=dtype)
 
 
-def rectangle(width, height, dtype=np.uint8):
+@deprecate_kwarg({"height": "ncols", "width": "nrows"},
+                 removed_version="0.20.0")
+def rectangle(nrows=None, ncols=None, dtype=np.uint8):
```
rfezzani, line 37: Specifying a default value is no more needed with the decorator
```suggestion
def rectangle(nrows, ncols, dtype=np.uint8):
```

skimage/morphology/selem.py
```diff
@@ -1,6 +1,9 @@
 import numpy as np
 from scipy import ndimage as ndi
+import warnings
```
rfezzani, line 3:
```suggestion
```
Unnecessary import ;)

doc/release/release_dev.rst
```diff
@@ -44,6 +47,8 @@ Deprecations
 - ``skimage.feature.structure_tensor_eigvals`` has been deprecated and will be
   removed in version 0.20. Use ``skimage.feature.structure_tensor_eigenvalues``
   instead.
+- In ``skimage.morphology.selem.rectangle`` the arguments ``width`` and 
+  ``height`` have been depreciated. Use ``nrow`` and ``ncol`` instead.
```
rfezzani, line 51:
```suggestion
  ``height`` have been deprecated. Use ``nrow`` and ``ncol`` instead.
```

rfezzani (commented):
Excellent, thank you @FirefoxMetzger !

skimage/morphology/selem.py
FirefoxMetzger: I'm not sure if this should be removed in this PR or at a later date. We are changing 10-year-old code, so some folks may wonder what happened to `width` and `height` and it seems polite to at least mention that they have been deprecated inside the docstring. Hence, I would remove this once skimage moves to a version that removes `width` and `height`.

Perhaps we could add a note instead of these parameters? I do admit that it depends on how you guys handle deprecations.

rfezzani: I prefer a note since we are deprecating these parameters and don't want new users to use them...

jni (approved):
This is great now!

rfezzani:
That's excellent! Thank you again @FirefoxMetzger :tada:

PR #4943, opened 2020-08-26
Fix draw.polygon segfault caused by 0d input

### LINKED ISSUE #4938: skimage.draw.polygon segmentation fault with np.nan
## Description
the function skimage.draw.polygon throws a segmentation fault instead of python exception when receiving np.nan

## Way to reproduce
```python
skimage.draw.polygon(np.nan, np.nan)
```

## Version information
```python
```

```python
# your output here

```

rfezzani:
## Description

Fixes #4938 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/draw/tests/test_draw.py
```diff
@@ -214,6 +214,12 @@ def test_polygon_exceed():
     assert_array_equal(img, img_)
 
 
+def test_polygon_0d_input():
```
sciunto, line 217: Please, add a comment with the issue number for future reference.

sciunto:
@rfezzani Could you confirm that this shape is the only one that fails in the draw submodule?

sciunto (approved):

rfezzani:
Thank you @sciunto.
No I can't confirm that `draw.polygon` is the only affected shape function, I simply searched if there is any `np.asanyarray` call in the `draw/_draw.pyx` file...

jni (approved):

jni:
Thanks @rfezzani!

PR #4947, opened 2020-08-28
modify development version string to allow use with NumpyVersion

grlee77:
## Description

We use NumpyVersion to compare version strings in some benchmarks. This currently fails on the master branch because '0.18.dev0' is not considered a valid version string by `NumpyVersion`. '0.18.0.dev0' works okay. 

If there is any reason not to change this version scheme we will instead have to change the version comparisons in the benchmark suite. Example usage in the benchmarks: https://github.com/scikit-image/scikit-image/blob/9e418094e9b7d80cf82b1090a93a16128f8f0a05/benchmarks/benchmark_segmentation.py#L3-L17

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/RELEASE.txt
+++ b/RELEASE.txt
@@ -65,8 +65,11 @@ access), specifically to:
 - On the master branch, update the version number in ``skimage/__init__.py``
   to the next ``.dev0`` version, commit, and push. This should follow PEP440
   meaning that the appropriate version number would look something like
-  ``0.20.dev0`` with the period between ``20`` and ``dev`` and a trailing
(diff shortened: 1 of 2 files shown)


sciunto (approved):

sciunto:
I think it's a typo...

hmaarrfk:
Link to the numpy version API

https://numpy.org/doc/stable/reference/generated/numpy.lib.NumpyVersion.html

I'm OK with this. I just wanted to update the release instructions.

Issue #4918, opened 2020-08-14
Memory corruption in flood_fill


mgmalheiros:
## Description

When using `flood_fill`, there is a consistent double free error and a dying kernel (as shown in the output from Jupyter) when using a negative row coordinate for the point where the fill starts.

## Way to reproduce

```python
import numpy as np
from skimage import segmentation

a = np.random.randint(0, 2, (100,100))

#for r in range(a.shape[0]):
#    segmentation.flood_fill(a, (r,  0), 1, in_place=True)
#    segmentation.flood_fill(a, (r, a.shape[1]-1), 1, in_place=True)
#    segmentation.flood_fill(a, (r, -1), 1, in_place=True)
    
for c in range(a.shape[1]):
#    segmentation.flood_fill(a, ( 0, c), 1, in_place=True)
#    segmentation.flood_fill(a, (a.shape[0]-1, c), 1, in_place=True)
    segmentation.flood_fill(a, (-1, c), 1, in_place=True)
```
All commented lines work, except the last one, which consistently triggers the error. The `flood` function also fails likewise with `segmentation.flood(a, (-1, c))`.

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.7.7 (default, May  7 2020, 21:25:33) 
[GCC 7.3.0]
Linux-5.4.0-42-generic-x86_64-with-debian-bullseye-sid
scikit-image version: 0.17.2
numpy version: 1.19.1
```
This error was present in version 0.16.2 too.

jni:
Yikes! Thank you @mgmalheiros for reporting this!

I think we can manually normalise the input coordinates. There might even be a NumPy function for doing this.

@emmanuelle I think a fix for this should be merged and backported before we release 0.17.3, hopefully this week. =)

emmanuelle:
I don't think there is a NumPy function for this, since it's as easy as `index % img.shape[0]` etc., i.e. you use the modulo operator. I therefore flagged this PR as "good first issue". I think my plate is a bit too full to release 0.17.3 this coming week and the next week I'm away on vacation, so there is a two-week window to address this issue :-).

FirefoxMetzger:
I tried to reproduce this just now, and it runs without error on `0.18.dev0`

Code:

```
[...]

PR #4948, opened 2020-08-29
Fix corrupt floodfill

FirefoxMetzger:
## Description

Closes: https://github.com/scikit-image/scikit-image/issues/4918

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/_flood_fill.py
+++ b/skimage/morphology/_flood_fill.py
@@ -236,8 +236,7 @@ def flood(image, seed_point, *, selem=None, connectivity=None, tolerance=None):
 
     seed_point_as_array = np.asarray_chkfinite(seed_point)
     if not np.all(
-            (0 <= seed_point_as_array) &
-            (seed_point_as_array < image.shape)):
+            (0 <= seed_point_as_array) & (seed_point_as_array < image.shape)):
         raise IndexError("seed_point lies outside the image.")
 
     seed_value = image[seed_point]
--- a/skimage/morphology/_flood_fill.py
+++ b/skimage/morphology/_flood_fill.py
@@ -224,6 +224,12 @@ def flood(image, seed_point, *, selem=None, connectivity=None, tolerance=None):
         image = np.ascontiguousarray(image)
         order = 'C'
 
+    seed_point_as_array = np.asarray_chkfinite(seed_point)
+    if not np.all(
+            (0 <= seed_point_as_array) &
+            (seed_point_as_array < image.shape)):
+        raise IndexError("seed_point lies outside the image.")
+
     seed_value = image[seed_point]
 
     # Shortcut for rank zero
@@ -242,7 +248,7 @@ def flood(image, seed_point, *, selem=None, connectivity=None, tolerance=None):
     working_image = _fast_pad(image, image.min(), order=order)
 
     # Stride-aware neighbors - works for both C- and Fortran-contiguity
-    ravelled_seed_idx = np.ravel_multi_index([i+1 for i in seed_point],
+    ravelled_seed_idx = np.ravel_multi_index([i + 1 for i in seed_point],
                                              working_image.shape, order=order)
     neighbor_offsets = _offsets_to_raveled_neighbors(
         working_image.shape, selem, center=((1,) * image.ndim), order=order)
--- a/skimage/morphology/_flood_fill.py
+++ b/skimage/morphology/_flood_fill.py
@@ -234,12 +234,8 @@ def flood(image, seed_point, *, selem=None, connectivity=None, tolerance=None):
     except TypeError:
         seed_point = (seed_point,)
 
-    seed_point_as_array = np.asarray_chkfinite(seed_point)
-    if not np.all(
-            (0 <= seed_point_as_array) & (seed_point_as_array < image.shape)):
-        raise IndexError("seed_point lies outside the image.")
-
     seed_value = image[seed_point]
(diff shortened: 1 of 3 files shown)


sciunto (approved):

skimage/morphology/_flood_fill.py
```diff
@@ -236,13 +234,20 @@ def flood(image, seed_point, *, selem=None, connectivity=None, tolerance=None):
     except TypeError:
         seed_point = (seed_point,)
 
+    seed_point_as_array = np.asarray_chkfinite(seed_point)
+    if not np.all(
+            (0 <= seed_point_as_array) & (seed_point_as_array < image.shape)):
+        raise IndexError("seed_point lies outside the image.")
```
jni, line 240: @FirefoxMetzger I'd prefer to give users the convenience of using negative indexing. Additionally, I don't think `chkfinite` is needed since the indices should be int anyway. I'd suggest:

```suggestion
    seed_point = tuple(np.asarray(seed_point) % working_image.shape)
```

... But that needs to go after the `working_image` definition, below.

FirefoxMetzger: @jni I saw the suggestion in the issue. I think it's a bit strange to use `%`, because it allows you to specify numbers completely outside the image like `seed_point = 2.5 * np.asarray(image.shape, dtype=np.intp)`, but why not.

FirefoxMetzger: Oh, also shouldn't it be `image.shape` instead of `working_image.shape` as we want to avoid the seed being located on the padding, which is what causes the segfault in the first place?

jni: > also shouldn't it be image.shape instead of working_image.shape as we want to avoid the seed being located on the padding, which is what causes the segfault in the first place?

... 🤦 of course! Well spotted. =)

> I think it's a bit strange to use %, because it allows you to specify numbers completely outside the image like `seed_point = 2.5 * np.asarray(image.shape, dtype=np.intp)`

Sure, and for those cases I wouldn't be opposed to a check that ensures that the seed is in `[-image.shape, image.shape)`, but I think that's a rather minor concern. I generally think undocumented behaviour is ok, as long as the failure mode is not catastrophic (such as a segfault! 😂).

skimage/morphology/_flood_fill.py
```diff
@@ -236,13 +234,16 @@ def flood(image, seed_point, *, selem=None, connectivity=None, tolerance=None):
     except TypeError:
         seed_point = (seed_point,)
 
+    seed_value = image[seed_point]
```
jni, line 237: @FirefoxMetzger actually, this will fail with out-of-range seed points, so I think we're covered. =)

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -219,7 +219,7 @@ def test_selem():
                          [0, 255, 255, 255, 255, 255],
                          [0, 255, 255, 255, 255, 255],
                          [0, 255, 255, 255, 255, 255],
-                         [0,   0,   0,   0,   0,   0]], dtype=np.uint8)
+                         [0, 0, 0, 0, 0, 0]], dtype=np.uint8)
```
jni, line 222: Could you please revert this change?

jni:
```suggestion
                         [0,   0,   0,   0,   0,   0]], dtype=np.uint8)
```

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -230,11 +230,11 @@ def test_selem():
     output = flood_fill(np.zeros((5, 6), dtype=np.uint8), (1, 4), 255,
                         selem=selem)
 
-    expected = np.array([[  0,   0,   0,   0,   0,   0],
-                         [255, 255, 255, 255, 255,   0],
-                         [255, 255, 255, 255, 255,   0],
-                         [255, 255, 255, 255, 255,   0],
-                         [255, 255, 255, 255, 255,   0]], dtype=np.uint8)
```
jni, line 237: And here. (Sorry, this suggestion is harder to do on the web interface 😬)

skimage/morphology/tests/test_flood_fill.py
```diff
@@ -275,5 +275,23 @@ def test_f_order(tolerance):
     np.testing.assert_array_equal(expected, mask)
 
 
+def test_out_of_bounds_seed_point():
```
jni, line 278: I'd rename this test, `test_negative_indexing_seed_point`

jni (approved):
@FirefoxMetzger other than some very minor formatting/naming suggestions, this is ready for me!

The Travis failures appear unrelated. One is a sporadic failure in the [rank filters tests](https://travis-ci.org/github/scikit-image/scikit-image/jobs/724676450#L4318-L4458), the other is the [mac build hanging](https://travis-ci.org/github/scikit-image/scikit-image/jobs/724676453) which we have encountered elsewhere.

emmanuelle:
Also approving once's @jni's small comments have been addressed.

FirefoxMetzger:
@jni @emmanuelle done. Let's wait for the tests to pass and then this *should* be good.

FirefoxMetzger:
@jni LGTM

emmanuelle:
Thank you @FirefoxMetzger , merging!

jni:

jni:
Dang. LOL

PR #4957, opened 2020-09-03
PERF: faster Hog computation

WeiChungChang:
Current Hog takes use of 3 loops, ex: for 9 bins, 128 * 128 cells.

outer loop - go through each bins.
middle loop - go through each rows.
inner loop - go through each columns.
    
However, 2 loops should be beter:
outer loop - go through each rows.
inner loop - go through each columns.
then divide orientation by number_of_orientations_per_180 and put result into corresponding bin.

Ex, if number_of_orientations_per_180 is 20 (divide 180 degree by 9 bins so each bin's range is 20 degree);
if current pixel has orientation = 45, it locates at (45/20) = 2nd bin.
if current pixel has orientation = 18, it locates at (18/20) = 0th bin.
and so on.

The experiment shows this way can speed up the calculation **about 10% ~ 20%** of whole flow.
Since the execution time is dominated by (1) calculation bin distribution for each cell, and (2) normalization,
**it implies for (1)only we speed up to 20~40%(may by more)**.

Here provided a test patch and attached a file to measure the difference.
It shows the difference is small enough to be ignored.

Roughly test with script attached shows the difference between original flow and refined flow is very small (use sum(abs(diff)) and for 4K image it is far smaller than 0.01).

The proposal patch keeps original flow, so a performance comparison can be executed by calling hog(....., **fast=True**) to select 2 loops method.

Please consider to rework current logic to provide better performance.

### DIFF
--- a/skimage/feature/_hoghistogram.pyx
+++ b/skimage/feature/_hoghistogram.pyx
@@ -6,6 +6,37 @@ import numpy as np
 cimport numpy as cnp
 cnp.import_array()
 
+cdef float cell_hog_fast(double[:, ::1] magnitude,
+                    double[:, ::1] orientation,
+                    int r_i, int c_i,
+                    int cell_columns, int cell_rows,
+                    int column_index, int row_index,
+                    int size_columns, int size_rows,
+                    int range_rows_start, int range_rows_stop,
+                    int range_columns_start, int range_columns_stop,
+                    int number_of_orientations,
+                    float number_of_orientations_per_180,
+                    cnp.float64_t[:, :, :] orientation_histogram) nogil:
+    cdef int cell_column, cell_row, cell_row_index, cell_column_index
+    cdef float total = 0.
+    cdef int idx
+
+    for cell_row in range(range_rows_start, range_rows_stop):
+        cell_row_index = row_index + cell_row
+        if (cell_row_index < 0 or cell_row_index >= size_rows):
+            continue
+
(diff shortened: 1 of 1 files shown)


skimage/feature/_hoghistogram.pyx
```diff
@@ -127,12 +159,13 @@ def hog_histograms(double[:, ::1] gradient_columns,
     range_columns_start = -range_columns_stop
     number_of_orientations_per_180 = 180. / number_of_orientations
 
-    with nogil:
-        # compute orientations integral images
-        for i in range(number_of_orientations):
+    if fast == 1:
+        print("use fast")  
+    else:
+        print("use original")
```
jni, line 165:
```suggestion
```

jni (changes requested):
@WeiChungChang thanks for this contribution! Apologies that you haven't received a response yet.

On first glance this is great! I'd like to request a few things:

- you mentioned an attached script but I don't see one. Rather than a script, though, it would be great (actually necessary) to add a test function to `skimage/feature/tests/test_hog.py` that checks that the two paths are close for a range of example images.
- you mention speed, it would so it would be great to see some benchmarks for this, again, both with fast and slow paths. See [this section of the contributing guide](https://scikit-image.org/docs/dev/contribute.html#benchmarks) for more!

After that I think we should be good — maybe some small tweaks to the code but nothing major.

Thank you again!

matthew-brett:

<details>

## Rationale

</details>

PR #4964, opened 2020-09-08
Delegate standard images load to imageio

rfezzani:
## Description

I suggest here to rely on `imageio` to load the standard test images that are common to both libraries. This will make `skimage` package a little bit lighter :wink:.
I started with `astraunot.png`, and if you agree with this idea, I will extend this approach to the other images.

### Candidate images

- astronaut.png
- text.png
- chessboard_GRAY.png
- coins.png
- moon.png
- page.png
- horse.png
- clock_motion.png
- ihc.png
- chelsea.png
- coffee.png
- hubble_deep_field.jpg

camera.png is also candidate, but waiting for #4913 and https://github.com/imageio/imageio/issues/452 to be fixed :wink:

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


sciunto:
It's a good idea to start along this way. I'm +1

sciunto:
PS : looks like you have unrelated commits in this branch...

rfezzani:
> PS : looks like you have unrelated commits in this branch...

Yeah, I messed up with my master branch  :cry: , but changes have been reverted...

jni:
Does imageio download the images or are they distributed with the package? I don't think we can assume an internet connection for a lot of our stuff.

Other than that I love this. I love it enough that I might consider ignoring the internet issue. 😂 @hmaarrfk might be the most opposed to that...?

jni:
@rfezzani would you be able to rebase those commits onto the current master? It shouldn't be too hard with cherry-pick:

```
git branch -D master
git fetch upstream master
git checkout --track master  # now your master matches ours
git checkout -b new-delegate-astronaut-iio
git cherry-pick f473a9e~..Delegate_astraunot_to_imageio
git branch -D Delegate_astraunot_to_imageio
git checkout -b Delegate_astraunot_to_imageio
git push origin --force-with-lease --set-upstream Delegate_astraunot_to_imageio
```

(Or you might have to replace your actual latest commit hash in the cherry-pick, not sure.)

Currently the commit history is a complete mess and I'd be worried that someone clicks "merge" rather than "squash" without thinking. 😬

sciunto:
> Does imageio download the images or are they distributed with the package? I don't think we can assume an internet connection for a lot of our stuff.
> 

It is downloaded from github (https://github.com/imageio/imageio/blob/master/imageio/core/fetching.py#L41) Not from our repository, would have been so funny ;D

However, I'm not sure to get the interrogation here. I think this debate was appropriate when we decided to use pooch. But, now, part of our date are online, and we will make examples based on them. What difference do you make @jni?

The question, to me, would rather be: are imageio's images cached or not? If they are, fine, if they aren't, we must integrate in our pooch thing.

Other question: do we aim to master the data we use? If imageio's images are unavailable, we do not have a hand to fix it. If it is on pooch/gitlab, we can take decision.

rfezzani:
> ... Not from our repository, would have been so funny ;D

:rofl: 

> The question, to me, would rather be: are imageio's images cached or not?

They are cached in `~/.imageio/images` :tada: 

> Other question: do we aim to master the data we use? If imageio's images are unavailable, we do not have a hand to fix it. If it is on pooch/gitlab, we can take decision.

That's an excellent remark. In an ideal world, the `scikit-image-data` package would be a common dependency of `scikit-image` and `imageio` to manage example/test images.

sciunto:
> > ... Not from our repository, would have been so funny ;D
> 
> rofl
> 
> > The question, to me, would rather be: are imageio's images cached or not?
> 
> They are cached in `~/.imageio/images` tada
> 
> > Other question: do we aim to master the data we use? If imageio's images are unavailable, we do not have a hand to fix it. If it is on pooch/gitlab, we can take decision.
> 
> That's an excellent remark. In an ideal world, the `scikit-image-data` package would be a common dependency of `scikit-image` and `imageio` to manage example/test images.

Yes, we can make this proposition.

jni:
Two things:

- yes, we are adding online datasets now. But our historical datasets are available with the package. (Right? Or did we deprecate this?)

- I've actually discussed with @almarklein the possibility of rolling imageio into the skimage org. Back in Feb he was open to the idea. The world is very different now though!

rfezzani:
> But our historical datasets are available with the package. (Right? Or did we deprecate this?)

I would love to deprecate `skimage.data` :grin:

hmaarrfk:
> I would love to deprecate skimage.data grin

Yeah, but there are use cases you aren't entirely consider. We were specifically looking for a solution where scikit-image's data was living as files. There some reassurance to this.

> Right? Or did we deprecate this?

We did not deprecate this. This is still very much a requirement, and part of what makes our pooch solution so challenging to implement. It isn't a bad thing, it just is what is it.

rfezzani:
@hmaarrfk, I see `skimage.data` as a test/doc requirement and not a "hard" requirement, that's why I am insisting that much for its deprecation + move to a separate package but I am may be not considering all the use cases :confused:

hmaarrfk:
I am beginning to learn (in my own projects), that having a few files that are in a familiar format is critical to a good healthy project.

@stefanv had opinions on this during the pooch PR. I think his concerns of having real files are valid.

rfezzani:
> that having a few files that are in a familiar format is critical to a good healthy project

I am not against this principle, I just don't understand why we are distributing these files! I think that they can live in a separate project dedicated to their distribution...

hmaarrfk:
They are rather historical, and sake for the sake of change isn't really a good thing. 

There is a simplicity in terms of just shipping sample data. You can be sure that the user's internet connection won't interfere.

I was running a demo once, and the motion stage I needed to use required an internet connection to refresh the internal database. Not fun.......

Does it:
* Reduce the overall repo size: no. Git doesn't work that way.
* Reduce the wheel size (arguably what most users use): I don't think the reduction is significant.
* Increase complexity for historically used samples. Yes.

The introduction of pooch had a strong value: namely the ability to add much larger datasets.

Basically:
* I wouldn't depend on imageio to maintain the datasets we expect for examples.
* I don't see a huge size value in removing the small sample datasets.

Note that we don't ship any test dataset. The idea is that the test suite is elaborate enough. Users that want the full suite should use pooch.

rfezzani:
Being conservative isn't a good thing either :wink:. I am OK for these images to be shipped somewhere, but I am not sure that `scikit-image` is the right place.

> Reduce the wheel size (arguably what most users use): I don't think the reduction is significant.

Current wheel size built from master is 29.1Mo and from this branch is 26.6Mo, the reduction is not really negligible...

hmaarrfk:
> Current wheel size built from master is 29.1Mo and from this branch is 26.6Mo, the reduction is not really negligible...

And the binary packages from conda-forge are 10Mo https://anaconda.org/conda-forge/scikit-image/files

There are many ways to achieve archive and runtime size reduction, I don't think forcing the reduction of 2Mo of images is really the way to go given the usability tradeoff.

The thing is that "2Mo is much easier to download during installation time" than "200Ko that you need to download during runtime". This is because you might be at home when you install, and on the train when you run example code.

If you are packaging things yourself, for a project where that 2Mo happens to make a difference, and you are asking for scikit-image to change their packaging structure, then I think adding

```python
if os.env.get('SCIKIT_IMAGE_PACKAGE_LEGACY_DATA', True) :
    config.add_data_files(*legacy_datasets)
```
https://github.com/scikit-image/scikit-image/blob/c76fc455493157d55f5eb50786b744daf3327b14/skimage/data/setup.py#L14

is the only change you need. The pooch system will fetch the data for you from the online store transparently instead of relying on imageio. FWIW, imageio also fetches images from github raw: https://github.com/imageio/imageio/blob/master/imageio/core/fetching.py#L62

I don't think many packagers will go for this, unless we provide them with a `scikit-image-data` data package, but it might be worth a pull request.

almarklein:
@jni 
> I've actually discussed with @almarklein the possibility of rolling imageio into the skimage org. Back in Feb he was open to the idea. The world is very different now though!

Still open to that idea :)

jni:
Thanks @almarklein for chiming in! We can then continue that discussion in a separate thread. =)

In the meantime, I have to say I am quite swayed by @hmaarrfk's arguments — as someone who does spend a lot of time on trains! Or did, anyway. 😅 And I agree with him that 2MB is not that amazing a return for the trouble.

Perhaps this is something we can revisit for 1.0.

@stefanv @emmanuelle thoughts here?

rfezzani:
@hmaarrfk, I am not arguing strictly for the purpose of this PR, but more for completely dropping `skimage.data` as you may be guessed :wink:.

I like the idea of packages doing one thing and doing it well. `scikit-image` is an image processing package, I consider that efforts must be dedicated to developing and maintaining image processing algorithms. You can realize how much "side" tasks (IO, visualization and data) requires efforts to `scikit-image` community by looking in our issue tracker!

I strongly believe that any other responsibility (IO, visualization and data) must be delegated to other packages.

When you work in regulated contexts (medical software), you need to manage the [SOUP](https://en.wikipedia.org/wiki/Software_of_unknown_pedigree) involved in your development, their issues must be tracked, analyzed... This a huge task, and believe me, when you are in this situation, you restrict as much as possible your external dependencies!..

Restricting `scikit-image` package dependency may help I think its adoption in this kind of context.

Now, `scikit-image` requires
- `pooch` only for the `skimage.data`,
- `matplotlib` mainly for `skimage.viewer`,
- `imageio`, `tifffile` and others for `skimage.io`
- ...

I think you got the idea :smile:

hmaarrfk:
> but more for completely dropping skimage.data as you may be guessed

sciunto:
> When you work in regulated contexts (medical software), you need to manage the [SOUP](https://en.wikipedia.org/wiki/Software_of_unknown_pedigree) involved in your development, their issues must be tracked, analyzed... This a huge task, and believe me, when you are in this situation, you restrict as much as possible your external dependencies!..

There is already all you need on pypi like [this](https://pypi.org/project/OnionPy/), [this](https://pypi.org/project/leek/) and [this](https://pypi.org/project/potato/) :) 

To be serious, the argument about traveling coders was one I advocated when we debated on pooch (but, at the same time, I was in for the size issues). Now that we have part of our data online and that, in the future, we will mix up examples using shipped and online datasets, I do not see the strength of this argument anymore.

Also, this makes me think that we need (if doesn't already exist) a `fetch_all` function for coders on rails ;)
In the doc, we can explain the installation procedure + how to cache the dataset.

emmanuelle:
@sciunto we do have `download_all` is this what you meant?

sciunto:
> @sciunto we do have `download_all` is this what you meant?

yes! See #4971

PR #4966, opened 2020-09-09
Accelerate gamma_adjust for uint8 images

rfezzani:
## Description

This PR proposes a LUT based approach for gamma correction of uint8 images. It allows for considerable speedup:

### Master branch

```python
In [1]: from skimage.exposure import adjust_gamma                                                       

In [2]: img = np.ones((1024, 1024, 3), dtype='uint8')                                                   

In [3]: %timeit adjust_gamma(img)                                                                       
10.7 ms ± 3.68 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### Current branch

```python
In [1]: from skimage.exposure import adjust_gamma                                                       

In [2]: img = np.ones((1024, 1024, 3), dtype='uint8')                                                   

In [3]: %timeit adjust_gamma(img)                                   
6.87 ms ± 9.45 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_exposure.py
+++ b/benchmarks/benchmark_exposure.py
@@ -10,7 +10,8 @@ from skimage import exposure
 class ExposureSuite:
     """Benchmark for exposure routines in scikit-image."""
     def setup(self):
-        self.image = img_as_float(data.moon())
+        self.image_u8 = data.moon()
+        self.image = img_as_float(self.image_u8)
         self.image = rescale(self.image, 2.0, anti_aliasing=False)
         # for Contrast stretching
         self.p2, self.p98 = np.percentile(self.image, (2, 98))
@@ -32,3 +33,7 @@ class ExposureSuite:
         # Running it 10 times to achieve significant performance time.
         for i in range(10):
             result = exposure.histogram(self.image)
+
+    def time_gamma_adjust_u8(self):
+        for i in range(10):
+            _ = exposure.adjust_gamma(self.image_u8)
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -496,14 +496,16 @@ def adjust_gamma(image, gamma=1, gain=1):
     dtype = image.dtype.type
 
     if dtype is np.uint8:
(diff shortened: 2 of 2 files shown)


sciunto:
Nice! I think it would be too much work to add a benchmark for this function: https://github.com/scikit-image/scikit-image/blob/master/benchmarks/benchmark_exposure.py

emmanuelle (approved):
Really cool! The performance gain was even bigger on my machine (15 vs 8.5 ms so factor of 1.76), and I also tested with an array of random uint and observed the same gain. A benchmark could be nice indeed, but we ca also merge the PR as it is.

jni (approved):
I'm also ok with merging as is but would also like a benchmark.

One comment about the code is it'd be better to write:

```python
if dtype is np.uint8:
    out = ...
else:
    _assert_non_negative(image)
    scale = ...
    out = ...
return out
```

It's one of NASA's rules for bug-free code to have only one return statement per function. I'm happy to stray from this when the returns are close to each other and well structured (`if ...: return something; else: return something_else`). Here, though, there's a return high up in the function and then the function continues...

But, I'm nitpicking, if you get a chance, I'd appreciate the changes, otherwise, we'll merge before release. =)

emmanuelle:
There seems to be something wrong with Travis, but it's not specific to this PR, it also happens on recent commits in other PRs.

emmanuelle:
> There seems to be something wrong with Travis, but it's not specific to this PR, it also happens on recent commits in other PRs.

There has been a release of dask on Friday, maybe this is why? It's probably related to one of the optional deps since only those builds are failing.

emmanuelle:
In #4973 I try to fix the CI problem, let's see if it works.

jni:
Awesome, thanks @emmanuelle! Since this is otherwise ready, I suggest we merge?

emmanuelle:
Sure, merging! Thanks @rfezzani .

Issue #4969, opened 2020-09-10
Tackle remaining todo tasks for 0.18


sciunto:
## Description

* In ``skimage.features.corner_peaks``, remove the warning.
    * In the following docstrings, remove the explicit setting of ``threshold_rel=0``
        * ``skimage/feature/brief.py::BRIEF``
        * `` skimage/feature/corner.py::corner_harris``
        * `` skimage/feature/corner.py::corner_shi_tomasi``
        * `` skimage/feature/corner.py::corner_foerstner``
        * `` skimage/feature/corner.py::corner_fast``
        * `` skimage/feature/corner.py::corner_subpix``
        * `` skimage/feature/corner.py::corner_peaks``
        * `` skimage/feature/corner.py::corner_orientations``
    * In ``skimage/feature/orb.py::_detect_octage`` remove explicitely setting
      ``threshold_rel=0``.
* In ``skimage/restoration/_denoise.py`` remove warning regarding
  ``rescale_sigma``.
    * In ``skimage/restoration/_cycle_spin.py::cycle_spin`` unskip the doctest
      now that the denoise warning is gone.

PR #4974, opened 2020-09-14
Remove warnings/explicit settings on feature, restoration

alexdesiqueira:
## Description

Solves #4969.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -38,6 +38,17 @@ API Changes
 - ``skimage.restoration.richardson_lucy`` returns a single-precision output
   when the input is single-precision. Prior to this release, double-precision
   was always used.
+- The explicit setting ``threshold_rel=0` was removed from the Examples of the
+  following docstrings: ``skimage.feature.BRIEF``,
+  ``skimage.feature.corner_harris``, ``skimage.feature.corner_shi_tomasi``,
+  ``skimage.feature.corner_foerstner``, ``skimage.feature.corner_fast``,
+  ``skimage.feature.corner_subpix``, ``skimage.feature.corner_peaks``,
+  ``skimage.feature.corner_orientations``, and
+  ``skimage.feature._detect_octave``.
+- In ``skimage.restoration._denoise``, the warning regarding
+  ``rescale_sigma=None`` was removed.
+- In ``skimage.restoration._cycle_spin``, the ``# doctest: +SKIP`` was removed.
+
 
 Bugfixes
 --------
@@ -59,7 +70,6 @@ Deprecations
   instead.
 - In ``skimage.morphology.selem.rectangle`` the arguments ``width`` and 
   ``height`` have been deprecated. Use ``nrow`` and ``ncol`` instead.
-
 
 Contributors to this release
 ----------------------------
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -38,17 +38,10 @@ API Changes
 - ``skimage.restoration.richardson_lucy`` returns a single-precision output
   when the input is single-precision. Prior to this release, double-precision
   was always used.
-- The explicit setting ``threshold_rel=0` was removed from the Examples of the
-  following docstrings: ``skimage.feature.BRIEF``,
-  ``skimage.feature.corner_harris``, ``skimage.feature.corner_shi_tomasi``,
-  ``skimage.feature.corner_foerstner``, ``skimage.feature.corner_fast``,
-  ``skimage.feature.corner_subpix``, ``skimage.feature.corner_peaks``,
-  ``skimage.feature.corner_orientations``, and
-  ``skimage.feature._detect_octave``.
-- In ``skimage.restoration._denoise``, the warning regarding
-  ``rescale_sigma=None`` was removed.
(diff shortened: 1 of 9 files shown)


rfezzani (approved):
Thank you @alexdesiqueira :wink:

jni (approved):
Awesome, thanks @alexdesiqueira! Could you add these to the API changes of the dev release notes?

We should really implement towncrier...

skimage/feature/corner.py
```diff
@@ -1052,18 +1050,12 @@ def corner_peaks(image, min_distance=1, threshold_abs=None, threshold_rel=None,
...
 
     """
     if threshold_rel is None:
         threshold_rel = 0.1
-        warn("Until version 0.16, threshold_rel was set to 0.1 by default. "
-             "Starting from version 0.16, the default value is set to None. "
-             "Until version 0.18, a None value corresponds to a threshold "
-             "value of 0.1. The default behavior will match "
-             "skimage.feature.peak_local_max. To avoid this warning, set "
-             "threshold_rel=0.", category=FutureWarning, stacklevel=2)
```
jni, line 1066: My reading of this is actually that the new default should be 0? Then *that* is what must go in the API change... Right?

skimage/restoration/_denoise.py
```diff
@@ -815,14 +815,6 @@ def denoise_wavelet(image, sigma=None, wavelet='db1', mode='soft',
...
 
     if rescale_sigma is None:
-        msg = (
-            "As of scikit-image 0.16, automated rescaling of sigma to match "
-            "any internal rescaling of the image is performed. Setting "
-            "rescale_sigma to False, will disable this new behaviour. To "
-            "avoid this warning the user should explicitly set rescale_sigma "
-            "to True or False."
-        )
-        warn(msg, FutureWarning, stacklevel=2)
```
jni, line 825: Also here: change the default value to `rescale=True` in the function signature.

jni (changes requested):
@alexdesiqueira I misread this PR in my earlier review. If I'm not mistaken, `threshold_rel=0` should be the new default value for all the feature.functions, and `rescale_sigma=True` should go directly in the function signature for cycle spin.

alexdesiqueira:
I'd ask for @sciunto's input as well; maybe I missed something, but #4969 reads different to me.

skimage/feature/corner.py
sciunto: In the signature, threshold_rel must be None. No modification of this variable must be done in this function. Therefore, the if statement has to be removed. This is to match feature.peak_local_max.

sciunto:
In addition, please remove the corresponding lines in TODO.txt :)

alexdesiqueira:
Alright @jni @sciunto that should suffice.
Please let me know what to do next. Maybe state that `rescale_sigma=True` on API changes?

doc/release/release_dev.rst
```diff
@@ -38,6 +38,17 @@ API Changes
 - ``skimage.restoration.richardson_lucy`` returns a single-precision output
   when the input is single-precision. Prior to this release, double-precision
   was always used.
+- The explicit setting ``threshold_rel=0` was removed from the Examples of the
+  following docstrings: ``skimage.feature.BRIEF``,
+  ``skimage.feature.corner_harris``, ``skimage.feature.corner_shi_tomasi``,
+  ``skimage.feature.corner_foerstner``, ``skimage.feature.corner_fast``,
+  ``skimage.feature.corner_subpix``, ``skimage.feature.corner_peaks``,
+  ``skimage.feature.corner_orientations``, and
+  ``skimage.feature._detect_octave``.
```
jni, line 47: This is not an API change.

skimage/feature/corner.py
jni: So, then, in API changes, we should list, "The default value of threshold_rel in feature.corner has changed from 0.1 to None, which corresponds to letting feature.peak_local_max decide on the default. That is currently equivalent to threshold_rel=0."

Something like that? The point is that the same function call (`corner_peaks(arr)`) will return different things between 0.17 and 0.18. That is what absolutely needs documenting in the release notes.

doc/release/release_dev.rst
```diff
@@ -38,6 +38,17 @@ API Changes
...
+- In ``skimage.restoration._denoise``, the warning regarding
+  ``rescale_sigma=None`` was removed.
+- In ``skimage.restoration._cycle_spin``, the ``# doctest: +SKIP`` was removed.
```
jni, line 50: Nor are these two. :joy:. See my comment below. Hopefully it is clear that only one API change has happened here!

jni (changes requested):
@alexdesiqueira one tiny fix! Also, my suggestion about adding `.. versionchanged::`?

jni:
> Also, my suggestion about adding `.. versionchanged::`?

LOL, I just remembered that was in a different PR :joy: Should apply here also though? :grimacing:

alexdesiqueira:
> Nor are these two. joy. See my comment below. Hopefully it is clear that only one API change has happened here!

@jni hope this is it! Please check again :blush:

jni (approved):

jni:
Boom. 😃

PR #4981, opened 2020-09-23
3D rank filters (ctd', follow up of 4435)

emmanuelle:
I created a new PR since I could not push commits to the branch of #4435. 

Benchmarks have been separated in two classes and take less time now (for the 3d suite, it takes about 1s per filter for the 128^3 shape. I could also reduce to 64^3).


JDWarner:
This looks extremely clean.  I'll let the original PR reviewers chime in rather than doing a review myself, but I love the generalization to 3D and have immediate use for it.

grlee77 (approved):
Thanks for fixing up the benchmarks, Emma. This looks ready to go

grlee77:
The only suggestion from #4435 that has not been addressed was to make the shape non-isotropic in at least one of the test cases. 

The example on the shape (10, 256, 256) brain does cover this to some extent and was what revealed the initial bug that I had fixed.

xialuoke4062:
> The only suggestion from #4435 that has not been addressed was to make the shape non-isotropic in at least one of the test cases.
> 
> The example on the shape (10, 256, 256) brain does cover this to some extent and was what revealed the initial bug that I had fixed.

Thank you emmanuelle for the benchmark fix! 
@grlee77  We made the shape non-isotropic in two test cases in #4435 and they pass the tests. Please let us know if there is anything else that we should edit regarding the test cases. Thank you!

emmanuelle:
I'm closing this PR in favor of #4435 since the commits from this PR have been merged there.

PR #5006, opened 2020-10-04
Refactoring: threshold methods with histogram parameters

haesleinhuepf:
added method threshold_otsu_from_histogram as shortcut in case the histogram is known in advance

## Description
Hi all, hi @jni ,

this PR-draft refactors thresholding_otsu method so that one can call it also with a histogram instead of an image. I would like to use it in a [GPU-accelerated image processing library](https://github.com/clEsperanto/pyclesperanto_prototype) where we can determine the histogram on the GPU, but lack threshold-method implementations.

I would wonder what you think about this implementation. If it makes sense, I would offer to implement the same strategy for other threshold methods.

Best,
Robert

## Checklist

[x] [Docstrings for all functions](https://github.com/numpy/numpy/blob/master/doc/example.py)
[ ] Gallery example in `./doc/examples` (new features only)
[ ] Benchmark in `./benchmarks`, if your changes aren't covered by an existing benchmark
[ ] Unit tests
[ ] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

## For reviewers

  later.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -293,9 +293,9 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
     hist : array, or 2-tuple of arrays, optional
-        Histogram from which to determine the threshold, and optionally a corresponding array
-        of bin center intensities. Alternatively, only the histogram can be
-        passed.
+        Histogram from which to determine the threshold, and optionally a
+        corresponding array of bin center intensities.
+        An alternative use of this function is to pass it only hist.
 
     Returns
     -------
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -237,6 +237,7 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
 
 def threshold_otsu(image=None, nbins=256, *, hist=None):
     """Return threshold value based on Otsu's method.
+
     Either image or hist must be provided. In case hist is given, the actual
     histogram of the image is ignored.
 
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -257,7 +257,7 @@ def _validate_image_histogram(image, hist, nbins=None):
     counts : 1D array of float
         Each element is the number of pixels falling in each intensity bin.
     bin_centers : 1D array
-        The value corresponding to the center of each intensity bin.
+        Each element is the value corresponding to the center of each intensity bin.
 
     Raises
     ------
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -293,7 +293,7 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
     hist : array, or 2-tuple of arrays, optional
-        Histogram to determine the threshold from and a corresponding array
+        Histogram from which to determine the threshold, and optionally a corresponding array
         of bin center intensities. Alternatively, only the histogram can be
         passed.
 
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -345,8 +345,7 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
     variance12 = weight1[:-1] * weight2[1:] * (mean1[:-1] - mean2[1:]) ** 2
 
     idx = np.argmax(variance12)
-
-    threshold = bin_centers[:-1][idx]
+    threshold = bin_centers[idx]
 
     return threshold
 
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -317,16 +317,22 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
     return threshold
 
 
-def threshold_yen(image, nbins=256):
+def threshold_yen(image=None, nbins=256, *, hist=None):
     """Return threshold value based on Yen's method.
+    Either image or hist must be provided. In case hist is given, the actual
+    histogram of the image is ignored.
 
     Parameters
     ----------
-    image : (N, M) ndarray
+    image : (N, M) ndarray, optional
         Input image.
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
+    hist : array, or 2-tuple of arrays, optional
+        Histogram to determine the threshold from and a corresponding array
+        of bin center intensities. Alternatively, only the histogram can be
+        passed.
 
     Returns
     -------
@@ -352,14 +358,38 @@ def threshold_yen(image, nbins=256):
     >>> thresh = threshold_yen(image)
     >>> binary = image <= thresh
     """
-    hist, bin_centers = histogram(image.ravel(), nbins, source_range='image')
+
+    if image is None and hist is None:
+        raise Exception("Either image or hist must be provided.")
+
+    if hist is not None:
+        if isinstance(hist, (tuple, list)):
+            counts, bin_centers = hist
+        else:
+            counts = hist
+            bin_centers = np.arange(counts.size)
+    else:
+        if image.ndim > 2 and image.shape[-1] in (3, 4):
+            msg = "threshold_otsu is expected to work correctly only for " \
+                  "grayscale images; image shape {0} looks like an RGB image"
+            warn(msg.format(image.shape))
+
+        # Check if the image is multi-colored or not
+        first_pixel = image.ravel()[0]
+        if np.all(image == first_pixel):
+            return first_pixel
+
+        counts, bin_centers = histogram(image.ravel(), nbins, source_range='image')
+
+    counts = counts.astype(float)
+
     # On blank images (e.g. filled with 0) with int dtype, `histogram()`
     # returns ``bin_centers`` containing only one value. Speed up with it.
     if bin_centers.size == 1:
         return bin_centers[0]
 
     # Calculate probability mass function
-    pmf = hist.astype(np.float32) / hist.sum()
+    pmf = counts.astype(np.float32) / counts.sum()
     P1 = np.cumsum(pmf)  # Cumulative normalized histogram
     P1_sq = np.cumsum(pmf ** 2)
     # Get cumsum calculated from end of squared array:
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -248,7 +248,7 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
-    hist : (array, array)  tuple, optional
+    hist : array, or 2-tuple of arrays, optional
         Histogram to determine the threshold from and a corresponding array
         of bin center intensities. Alternatively, only the histogram can be
         passed.
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -235,8 +235,7 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
 
     return thresh_image - offset
 
-
-def threshold_otsu(image=None, nbins=256, hist=None, bin_centers=None):
+def threshold_otsu(image=None, nbins=256, *, histogram=None):
     """Return threshold value based on Otsu's method.
     Either image or hist must be provided. In case hist is given, the actual
     histogram of the image is ignored.
@@ -248,11 +247,10 @@ def threshold_otsu(image=None, nbins=256, hist=None, bin_centers=None):
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
-    hist : optional
-        Histogram to determine the threshold from.
-    bin_centers : optional
-        Intensity values for each histogram bin, corresponding to the given
-        histogram
+    histogram : (array, array)  tuple, optional
+        Histogram to determine the threshold from and a corresponding array
+        of bin center intensities. Alternatively, only the histogram can be
+        passed.
 
     Returns
     -------
@@ -278,7 +276,13 @@ def threshold_otsu(image=None, nbins=256, hist=None, bin_centers=None):
     if image is None and hist is None:
         raise Exception("Either name or hist must be provided.")
 
-    if hist is None:
+    if histogram is not None:
+        if isinstance(histogram, (tuple, list)):
+            hist, bin_centers = histogram
+        else:
+            hist = histogram
+            bin_centers = np.arange(counts.size)
+    else:
         if image.ndim > 2 and image.shape[-1] in (3, 4):
             msg = "threshold_otsu is expected to work correctly only for " \
                   "grayscale images; image shape {0} looks like an RGB image"
@@ -307,10 +311,7 @@ def threshold_otsu(image=None, nbins=256, hist=None, bin_centers=None):
 
     idx = np.argmax(variance12)
 
-    if bin_centers is None:
-        threshold = idx
-    else :
-        threshold = bin_centers[:-1][idx]
+    threshold = bin_centers[:-1][idx]
 
     return threshold
 
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -279,6 +279,42 @@ def threshold_otsu(image, nbins=256):
         return first_pixel
 
     hist, bin_centers = histogram(image.ravel(), nbins, source_range='image')
+
+    return threshold_otsu_from_histogram(hist, bin_centers)
+
+def threshold_otsu_from_histogram(hist, bin_centers):
+    """Return threshold value based on Otsu's method dermined from a histogram.
+
+    Parameters
+    ----------
+    hist : optional
+        Histogram to determine the threshold from.
+    bin_centers : optional
+        Intensity values for each histogram bin, corresponding to the given
+        histogram
+
+    Returns
+    -------
+    threshold : float
+        Upper threshold value. All pixels with an intensity higher than
+        this value are assumed to be foreground.
+
+    References
+    ----------
+    .. [1] Wikipedia, https://en.wikipedia.org/wiki/Otsu's_Method
+
+    Examples
+    --------
+    >>> from skimage.data import camera
+    >>> from skimage.exposure import histogram
+    >>> from skimage.filters import threshold_otsu_from_histogram
+    >>> image = camera()
+    >>> nbins = 256
+    >>> hist, bin_centers = histogram(image.ravel(), nbins, source_range='image')
+    >>> thresh = threshold_otsu_from_histogram(hist, bin_centers)
+    >>> binary = image <= thresh
+    """
+
     hist = hist.astype(float)
 
     # class probabilities for all possible thresholds
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -720,22 +720,29 @@ def threshold_li(image, *, tolerance=None, initial_guess=None,
     return threshold
 
 
-def threshold_minimum(image, nbins=256, max_iter=10000):
+def threshold_minimum(image=None, nbins=256, max_iter=10000, *, hist=None):
     """Return threshold value based on minimum method.
 
-    The histogram of the input ``image`` is computed and smoothed until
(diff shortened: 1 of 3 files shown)


emmanuelle:
Hi @haesleinhuepf thanks for starting this discussion! I'll let other devs chime in, but my gut feeling is that rather than adding one function to the API we could just add one keyword argument `histogram` (which would default to None) to the existing function. I see the interest of precomputing the histogram also for comparing different thresholding functions which require the histogram. 

@sciunto you've worked extensively on thresholding functions what do you think of these additions?

jni:
I think the same as @emmanuelle: let's add `histogram=None` to the functions that benefit from a pre-computed histogram. We could potentially then also make `image=None`, and make sure one of the two is provided, with preference given to the histogram.

haesleinhuepf:
Hey @emmanuelle and @jni ,

at the beginning I thought so, too. But then I saw that we need the histogram and the bin_centers array. Otherwise, we can only return the bin_index  and not the corresponding grey value. I didn't know how force the user to provide both together or both none. Assert statements came to my mind. Any better idea how to achieve this? 

Thanks for the feedback! 😃

jni:
A great point that I had overlooked! 😅 

My preferred option given this would be to provide a tuple of `counts, bin_centers`, same as returned by `exposure.histogram`. Perhaps we also support a single array and then use the bin index (`bin_centers = np.arange(len(counts))`), which is convenient for uint16 and uint8 images.

haesleinhuepf:
> Perhaps we also support a single array and then use the bin index (`bin_centers = np.arange(len(counts))`), which is convenient for uint16 and uint8 images.

Just to make sure I understand correctly: That would mean the threshold_otsu method returns a conceptually different thing (histogram index versus grey value) depending on input parameters. If that's ok, I'm happy to implement it.

sciunto:
I'm +1 with you @emmanuelle and this can be applied to several of our functions as well.

jni:
> a conceptually different thing 

It's not, the point is that for integer-valued images they are one and the same thing! So it is a shortcut for users of those images. If someone decides to misuse the shortcut, that's on them! 😜

At any rate, you can implement the tuple input version first and later we can decide if we want to provide this shortcut. 😊

haesleinhuepf:
Alright @emmanuelle  and @jni ,

I just implemented it like you suggested. The function header is now
```
def threshold_otsu(image=None, nbins=256, hist=None, bin_centers=None)
```

If prefers the hist over the image as @jni suggested and returns the bin index in case no bin_centers are given. The only thing I didn't manage is the tuple-thingy @jni suggested. I assume one needs to add some more brakets to the  function defintion, but I'm not sure how this works with bin_centers being optional. You know, I just recently arrived in snake county. ;-)

Thanks again for your feedback. I appreciate it :-)

Let me know what you think!

Cheers,
Robert

jni:
@haesleinhuepf I think you were expecting more magic than Python provides here. 😂  I merely meant that a single argument, `histogram=`, takes in as input a tuple of arrays, which you unpack inside the function:

```python
def threshold_otsu(image=None, nbins=256, *, histogram=None):
    if histogram is not None:
        counts, bin_centers = histogram
    else:
        counts, bin_centers = exposure.histogram(image, nbins)
```

or 

```python
def threshold_otsu(image=None, nbins=256, *, histogram=None):
    if histogram is not None:
        if isinstance(histogram, (tuple, list)):
            counts, bin_centers = histogram
        else:
            counts = histogram
            bin_centers = np.arange(counts.size)
    else:
        counts, bin_centers = exposure.histogram(image, nbins)
```

skimage/filters/thresholding.py
```diff
@@ -235,17 +235,22 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
 
     return thresh_image - offset
 
-
-def threshold_otsu(image, nbins=256):
+def threshold_otsu(image=None, nbins=256, *, histogram=None):
```
emmanuelle, line 238: Please find another name here as it conflicts with the `histogram` function imported from `exposure` at the top of the file (another solution would be to use it as `exposure.histogram` in the whole file, but it's probably better to choose another name, maybe just `hist`?)

haesleinhuepf: Good catch! It was called `hist` earlier. So we can stick to that shorter name.

haesleinhuepf:
Hey @emmanuelle and @jni ,

this should ne it. Please let me know if there is anything else that should be adapted. If you like it, I'd be happy to do the same change for the other thresholding methods :-)

skimage/filters/thresholding.py
```diff
@@ -235,17 +235,22 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
 
     return thresh_image - offset
 
-
-def threshold_otsu(image, nbins=256):
+def threshold_otsu(image=None, nbins=256, *, hist=None):
     """Return threshold value based on Otsu's method.
```
jni, line 239:
```suggestion
    """Return threshold value based on Otsu's method.

```

skimage/filters/thresholding.py
```diff
@@ -268,33 +273,46 @@ def threshold_otsu(image, nbins=256):
     -----
     The input image must be grayscale.
     """
-    if image.ndim > 2 and image.shape[-1] in (3, 4):
-        msg = "threshold_otsu is expected to work correctly only for " \
-              "grayscale images; image shape {0} looks like an RGB image"
-        warn(msg.format(image.shape))
+    if image is None and hist is None:
+        raise Exception("Either name or hist must be provided.")
```
jni, line 277:
```suggestion
        raise Exception("Either image or hist must be provided.")
```

skimage/filters/thresholding.py
```diff
@@ -268,33 +273,46 @@ def threshold_otsu(image, nbins=256):
...
+    mean2 = (np.cumsum((counts * bin_centers)[::-1]) / weight2[::-1])[::-1]
 
     # Clip ends to align class 1 and class 2 variables:
     # The last value of ``weight1``/``mean1`` should pair with zero values in
     # ``weight2``/``mean2``, which do not exist.
     variance12 = weight1[:-1] * weight2[1:] * (mean1[:-1] - mean2[1:]) ** 2
 
     idx = np.argmax(variance12)
+
     threshold = bin_centers[:-1][idx]
```
jni, line 314:
```suggestion
    threshold = bin_centers[idx]
```

I know it's not your code but might as well do some cleanup! This should clearly not affect the output!

skimage/filters/thresholding.py
```diff
@@ -235,17 +235,22 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
 
     Parameters
     ----------
-    image : (N, M) ndarray
+    image : (N, M) ndarray, optional
         Grayscale input image.
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
+    hist : (array, array)  tuple, optional
```
jni, line 250:
```suggestion
    hist : array, or 2-tuple of arrays, optional
```

jni (approved):
Thank you @haesleinhuepf! For me this is ready! I'm excited to have this functionality throughout the module!

I've made some very minor formatting wording suggestions — please apply at your leisure. :blush:

jni:
@scikit-image/core anyone else want to push the green button on this one? 😊

haesleinhuepf:
> @scikit-image/core anyone else want to push the green button on this one? 😊

before you push any button: May I also adapt the other thresholding methods within this PR?

jni:
@haesleinhuepf as you wish, but I am very much a "bird in the hand" kinda guy. 😂

jni:
Actually it occurs to me that we should probably add a test for this code! 😂

jni:
@haesleinhuepf would you mind adding that test? It should be the same as the test for otsu, but computing the histogram outside of the function call.

haesleinhuepf:
Sure! Will add it/them [here](https://github.com/scikit-image/scikit-image/blob/master/skimage/filters/tests/test_thresholding.py#L222)

jni:
@haesleinhuepf I've refactored the copy-pasted code into its own function, to avoid repeating code too much. =)

FYI you should generally not make PRs from master, but rather from a specific branch. This is because it is now impossible for you to have your master branch match ours after this PR merges. It's not a huge deal, just a bunch of branch renaming, but it's simpler to just start from a separate branch to begin with. =)

@scikit-image/core this is ready to merge imho, if anyone has some review cycles available!

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
+        Histogram, either a 1D counts array, or an array of counts together
+        with an array of bin centers.
+    nbins : int, optional
+        The number of bins with which to compute the histogram, if `hist` is
+        None.
+
+    Returns
+    -------
+    counts : 1D array of float
+        The number of pixels falling in each intensity bin
```
mkcor, line 258:
```suggestion
        Each element is the number of pixels falling in each intensity bin.
```

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
+    bin_centers : 1D array
+        The value corresponding to the center of each intensity bin.
```
mkcor, line 260:
```suggestion
        Each element is the value corresponding to the center of each intensity bin.
```

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
+        counts, bin_centers = histogram(
+                image.ravel(), nbins, source_range='image'
+            )
+    return counts.astype(float), bin_centers
+
+
+def threshold_otsu(image=None, nbins=256, *, hist=None):
     """Return threshold value based on Otsu's method.
 
+    Either image or hist must be provided. In case hist is given, the actual
```
mkcor, line 285:
```suggestion
    Either image or hist must be provided. If hist is provided, the actual
```
I find it more direct this way.

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
+    hist : array, 2-tuple of array, or None
+        Histogram, either a 1D counts array, or an array of counts together
+        with an array of bin centers.
+    nbins : int, optional
+        The number of bins with which to compute the histogram, if `hist` is
+        None.
+
+    Returns
+    -------
+    counts : 1D array of float
```
mkcor, line 257: :thinking: Counts are integers... I can see you are casting them into floats line 279. But the `counts` returned by `exposure.histogram` have dtype `int`...

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
     Parameters
     ----------
-    image : (N, M) ndarray
+    image : (N, M) ndarray, optional
         Grayscale input image.
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
+    hist : array, or 2-tuple of arrays, optional
+        Histogram to determine the threshold from and a corresponding array
```
mkcor, line 296:
```suggestion
        Histogram to determine the threshold from, and possibly a corresponding array
```

skimage/filters/thresholding.py
```diff
@@ -268,46 +318,54 @@ def threshold_otsu(image, nbins=256):
...
     Parameters
     ----------
-    image : (N, M) ndarray
+    image : (N, M) ndarray, optional
         Input image.
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
+    hist : array, or 2-tuple of arrays, optional
+        Histogram to determine the threshold from and a corresponding array
```
mkcor, line 366: I would make the same suggestion as above. Also, I wonder if 'Alternative' is clear enough (it seems like we're referring to what was just said). I would rephrase along the lines of: "An alternative usage of this threshold function is to pass it only a histogram, not an image."

mkcor (commented):
@jni I left a few comments, but I'm not far from approving!
PS: I have taken the liberty of committing the tiny cleanup suggestion.

skimage/filters/thresholding.py
jni: @mkcor yes, but basically every calling function was then converting them to float, so I thought that conversion might as well get rolled into here.

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
+    if hist is not None:
+        if isinstance(hist, (tuple, list)):
+            counts, bin_centers = hist
+        else:
+            counts = hist
+            bin_centers = np.arange(counts.size)
+    else:
+        counts, bin_centers = histogram(
+                image.ravel(), nbins, source_range='image'
+            )
```
mkcor, line 278: Are the above two lines extra-indented, or is it just GitHub's rendering?

skimage/filters/thresholding.py
mkcor: Ok!

skimage/filters/thresholding.py
jni: this was my editor's indentation choice and I liked it! And it's PEP8 compliant. ;)

jni: (Did I already mention I hate black? =P)

skimage/filters/thresholding.py
jni: Even better I think:

```suggestion
        Histogram from which to determine the threshold from, and optionally a corresponding array
```

But this might need reflowing for PEP8.

mkcor: Agreed! Oops, almost: remove second 'from' (:

jni: Oops indeed! Edited. :facepalm:

skimage/filters/thresholding.py
mkcor: > this was my editor's indentation choice and I liked it! And it's PEP8 compliant. ;)

No, it's not! :confounded: See https://github.com/scikit-image/scikit-image/pull/5006#issuecomment-703691052:

* In the file skimage/filters/thresholding.py:

    Line 277:17: E126 continuation line over-indented for hanging indent
    Line 278:13: E121 continuation line under-indented for hanging indent

mkcor: > (Did I already mention I hate black? =P)

This I know! :rofl: I was surprised to see double quotes [here](https://github.com/scikit-image/scikit-image/pull/5006/files#diff-4f1fba6d7478009483dadb139f489db58130185867229af13354ae80a1b2ca3aR267). :wink:

skimage/filters/thresholding.py
```diff
@@ -268,46 +318,54 @@ def threshold_otsu(image, nbins=256):
...
-    image : (N, M) ndarray
+    image : (N, M) ndarray, optional
         Input image.
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
+    hist : array, or 2-tuple of arrays, optional
+        Histogram to determine the threshold from and a corresponding array
+        of bin center intensities. Alternatively, only the histogram can be
+        passed.
```
mkcor, line 368:
```suggestion
        Histogram from which to determine the threshold, and optionally a
        corresponding array of bin center intensities.
        An alternative use of this function is to pass it only hist.
```

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
-    image : (N, M) ndarray
+    image : (N, M) ndarray, optional
         Grayscale input image.
     nbins : int, optional
         Number of bins used to calculate histogram. This value is ignored for
         integer arrays.
+    hist : array, or 2-tuple of arrays, optional
+        Histogram from which to determine the threshold, and optionally a corresponding array
+        of bin center intensities. Alternatively, only the histogram can be
+        passed.
```
mkcor, line 298:
```suggestion
        Histogram from which to determine the threshold, and optionally a
        corresponding array of bin center intensities.
        An alternative use of this function is to pass it only hist.
```

mkcor (commented):
@haesleinhuepf please note, in case you were going to commit on your branch again, that I have since committed a suggestion on it. Before resuming work locally, you would therefore need to sync up your branch with a `git pull`.

skimage/filters/thresholding.py
mkcor: @jni PEP8 ^^

skimage/filters/thresholding.py
haesleinhuepf: Maybe these double quote come from that Java developer who sent this PR ;-)

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
     return thresh_image - offset
 
 
-def threshold_otsu(image, nbins=256):
+def _validate_image_histogram(image, hist, nbins=None):
+    """Ensure that either image or hist were given, return valid histogram.
+
+    If hist is given, image is ignored.
```
alexdesiqueira, line 242: Could you please put this sentence in a `Notes` section?

skimage/filters/thresholding.py
```diff
@@ -236,16 +236,66 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
+        counts, bin_centers = histogram(
+                image.ravel(), nbins, source_range='image'
+            )
+    return counts.astype(float), bin_centers
+
+
+def threshold_otsu(image=None, nbins=256, *, hist=None):
     """Return threshold value based on Otsu's method.
 
+    Either image or hist must be provided. If hist is provided, the actual
```
alexdesiqueira, line 285: Same here. Could this sentence be in `Notes`?

skimage/filters/thresholding.py
```diff
@@ -268,46 +318,54 @@ def threshold_otsu(image, nbins=256):
...
-    threshold = bin_centers[:-1][idx]
+    threshold = bin_centers[idx]
+
     return threshold
 
 
-def threshold_yen(image, nbins=256):
+def threshold_yen(image=None, nbins=256, *, hist=None):
     """Return threshold value based on Yen's method.
+    Either image or hist must be provided. In case hist is given, the actual
```
alexdesiqueira, line 355: Same here: this sentence in `Notes` :)

skimage/filters/thresholding.py
```diff
@@ -368,16 +427,23 @@ def threshold_isodata(image, nbins=256, return_all=False):
     For integer images, the above equality holds to within one; for floating-
     point images, the equality holds to within the histogram bin-width.
 
+    Either image or hist must be provided. In case hist is given, the actual
```
alexdesiqueira, line 430: See previous :)

skimage/filters/thresholding.py
```diff
@@ -641,22 +707,29 @@ def threshold_li(image, *, tolerance=None, initial_guess=None,
...
 
 
-def threshold_minimum(image, nbins=256, max_iter=10000):
+def threshold_minimum(image=None, nbins=256, max_iter=10000, *, hist=None):
     """Return threshold value based on minimum method.
 
-    The histogram of the input ``image`` is computed and smoothed until
-    there are only two maxima. Then the minimum in between is the threshold
-    value.
+    The histogram of the input ``image`` is computed if not provided and
```
alexdesiqueira, line 713: Both sentences here in `Notes`, maybe?

alexdesiqueira (commented):
Danke schön, @haesleinhuepf! I put some ideas around; I think it'd be nicer if we maintain the details on a `Notes` section in each doc, uncluttering the headers a little. How does that sound?
Could we also use `should` where we are using `must`, maybe? :slightly_smiling_face: 
Thanks again!

skimage/filters/thresholding.py
stefanv: These should be in parameter descriptions.  "Must" is a good word to use when, well, you *must* provide the argument for it to work :)

skimage/filters/thresholding.py
stefanv: Black can be told to leave quotes alone; that's my preference.

stefanv (approved):
@haesleinhuepf Thank you for your PR!

This is a good improvement; if folks feel strongly about any of the tweaks raised here, let's address them in a follow-up PR.

PR #5012, opened 2020-10-09
Add gallery example for structure_tensor... in 3D!

mkcor:
## Description

Addresses part of #4831.
With this PR, I'm following up on https://github.com/scikit-image/scikit-image/issues/4831#issuecomment-705450778. 

@emmanuelle I tried telling a story but, in terms of applications, it's more didactic than scientific.
@GenevieveBuckley I used a small region of the kidney image!

I wanted to include some 3D rendering with Plotly, like:
3d_isotropy
but it would only spit out a JSON:
```json
{'text/html': '<div>            <script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-AMS-MML_SVG"></script><script type="text/javascript">if (window.MathJax) {MathJax.Hub.Config({SVG: {font: "STIX-Web"}});}</script>                <script type="text/javascript">window.PlotlyConfig = {MathJaxConfig: \'local\'};</script>\n        <script src="https://cdn.plot.ly/plotly-latest.min.js"></script>                <div id="de5ac8b7-d7bc-41f1-a094-bc28e8cdaa06" class="plotly-graph-div" style="height:525px; width:100%;"></div>            <script type="text/javascript">                                    window.PLOTLYENV=window.PLOTLYENV || {};                                    if (document.getElementById("de5ac8b7-d7bc-41f1-a094-bc28e8cdaa06")) {                    Plotly.newPlot(                        "de5ac8b7-d7bc-41f1-a094-bc28e8cdaa06",                        [{"opacity": 0.5, "slices": {"z": {"locations": [5], "show": true}}, "type": "volume", "value": [774, 566, 384, 333, ...
```
(plus it would be too heavy).

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


mkcor:
@jni iirc you would recommend these changes: 1a5887b873ccce2d8fd3ad20876d18f8294fdddf
What's the name for this way of formatting strings again? Thanks!

jni:
Yes I would and they are [f-strings](https://www.python.org/dev/peps/pep-0498/)

mkcor:
@scikit-image/core please review when you get a chance!

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+
+#####################################################################
+# Let us pick a specific region, which shows relative X-Y isotropy. In
+# contrast, the gradient is quite different (and, for that matter, weak) along
+# Z.
+
+sample = data[5:13, 380:410, 370:400, 1]
+
+
+def display(im3d, cmap='gray', step=3):
```
rfezzani, line 67: Defining the `display` function is not necessary since it is used only once...

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+    _, axes = plt.subplots(nrows=1, ncols=3, figsize=(16, 8))
+
+    it = 0
+    for ax, image in zip(axes.flatten(), im3d[::step]):
+        ax.imshow(image, cmap=cmap, vmin=v_min, vmax=v_max)
+        ax.set_title(f'Plane = {5 + it * step}')
+        ax.set_xticks([])
+        ax.set_yticks([])
+        it += 1
```
rfezzani, line 76:
```suggestion
    for it, (ax, image) in enumerate(zip(axes.flatten(), im3d[::step])):
        ax.imshow(image, cmap=cmap, vmin=v_min, vmax=v_max)
        ax.set_title(f'Plane = {5 + it * step}')
        ax.set_xticks([])
        ax.set_yticks([])
```

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+display(sample)
+
+#####################################################################
+# To view the sample data in 3D, run the following code:
+
+# import plotly.graph_objects as go
```
rfezzani, line 84: Can't we use a code block here:
```
.. code-block:: python
```
?

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+#####################################################################
+# We can then compute the eigenvalues of the structure tensor.
+
+eigen = feature.structure_tensor_eigenvalues(A_elems)
+eigen.shape
+
+#####################################################################
+# Where is the largest eigenvalue?
+
+coords = np.where(eigen == eigen.max())
```
rfezzani, line 131:
```suggestion
coords = np.unravel_index(eigen.argmax(), eigen.shape)
```

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+#####################################################################
+# In contrast, the third-largest eigenvalues are one order of magnitude
+# smaller.
+
+eigen[2, coords[1], 20:23, 12:14]
+
+#####################################################################
+# As expected, the region about voxel ``(Z, Y, X) = coords[1:]`` is markedly
+# anisotropic in 3D: There is an order of magnitude between the third-largest
+# eigenvalues on one hand, and the largest and second-largest eigenvalues on
```
rfezzani, line 156:
```suggestion
# eigenvalue on one hand, and the largest and second-largest eigenvalues on
```

rfezzani (approved):
Thank you @mkcor! I just left some minor suggestions :wink:

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: Oh, yes, thanks! Not only do we get the expected rendering, but readers can also use the copy-to-clipboard button. Done d843572952caf7b1c4b5e7070aba87121f253894.

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: The third-largest eigenvalues are `eigen[2, coords[1], 20:23, 12:14]` (six values) mentioned above.

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: You're right. It doesn't even work well for readers to reuse (say, changing the `step` value), because the number of rows and columns are hard-coded for the subplots! Fixed in 808b9a7f2ba129556aad6d68f367903afd647a72.

PS: thanks for suggesting `enumerate()` to get a counter in the loop! :woman_facepalming:

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,160 @@
...
+
+coords = np.unravel_index(eigen.argmax(), eigen.shape)
+assert coords[0] == 0  # by definition
+coords
+
+#####################################################################
+# We are looking at a local property. Let us consider a tiny neighbourhood
+# of this maximum in the X-Y plane.
+
+eigen[0, coords[1], 20:23, 12:14]
```
grlee77, line 135: shouldn't we use `coords` instead of these hardcoded numbers (20:23, 12:14) here?
```Python
ycoords = slice(coords[2] - 1, coords[2] + 2)
zcoords = slice(coords[3] - 1, coords[3] + 2)

eigen[0, coords[1], ycoords, zcoords]
```

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,160 @@
...
+
+#####################################################################
+# As expected, the region about voxel ``(Z, Y, X) = coords[1:]`` is markedly
+# anisotropic in 3D: There is an order of magnitude between the third-largest
+# eigenvalues on one hand, and the largest and second-largest eigenvalues on
+# the other hand.
+# This region is 'somewhat isotropic' in the X-Y plane: There is a factor of
+# (only) ~3 between the second-largest and largest eigenvalues.
+# This is definitely compatible with what we are seeing in the image, i.e., a
+# stronger gradient roughly along X and a weaker gradient perpendicular to it.
```
grlee77, line 157: This sentence sounds kind of like we are implying the entries in `eigvals` correspond to X,Y, Z axis, but that isn't really the case. From just the `eigvals` array, we don't know the directions of the corresponding eigenvectors.

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,160 @@
...
+    zmin=v_min,
+    zmax=v_max,
+    labels={'x': 'Y', 'y': 'X', 'color': 'intensity'},
+    title='Interactive view of bottom slice of sample data.'
+)
+
+#####################################################################
+# Therefore, we choose a 'width' of 3 for the window function.
+
+sigma = 3
```
grlee77, line 115: sigma of 3 seems pretty large, particularly for the first axis which is only size 8 here. To make the result less dependent on boundary conditions, I would use something a bit smaller (maybe 1.0-1.5 or so?)

This is the Gaussian kernel that would be used by SciPy's `gaussian_filter` for sigma=3:
```Python
from scipy.ndimage.filters import _gaussian_kernel1d
sigma = 3
truncate = 4
kernel = _gaussian_kernel1d(sigma, order=0, radius=int(truncate*sigma + 0.5))
plt.figure(); plt.plot(kernel, 'k.-')
```
g


also, The `structure_tensor` docstring indicates `sigma` should be a float, but it looks like the function can accept a unique sigma per axis (e.g. `sigma=(3, 3, 1)`). If that is intentional, we should update the docstring to mention it.

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,160 @@
...
+eigen = feature.structure_tensor_eigenvalues(A_elems)
+eigen.shape
+
+#####################################################################
+# Where is the largest eigenvalue?
+
+coords = np.unravel_index(eigen.argmax(), eigen.shape)
+assert coords[0] == 0  # by definition
+coords
+
```
grlee77, line 130: I think it would be nice to also plot all 3 eigenvalues for a single slice (e.g. 1 subplot per eigenvalue, using uniform brightness scale across subplots).

grlee77 (changes requested):
Thanks for working on this @mkcor! I have some comments, but think the general approach here is fine and this dataset should be okay to demonstrate the basic principles

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,160 @@
+"""
+==========================================
+Estimate isotropy in a 3D microscopy image
```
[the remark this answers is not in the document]
mkcor: Right, good point. In geology (which was my major when I started university), people say 'sub-isotropic' so we could think of it as "how isotropic? almost" but -- yes. :sweat_smile:

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -0,0 +1,160 @@
...
+
+#####################################################################
+# What exactly are the shape and size of our 3D multichannel image?
+
+print(f'number of dimensions: {data.ndim}')
+print(f'shape: {data.shape}')
+print(f'dtype: {data.dtype}')
+
+#####################################################################
+# For the purposes of this tutorial, we shall consider only the second colour
```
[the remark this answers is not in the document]
mkcor: Yes, because I'm the author of this other demo and I use Canadian spelling! :sweat_smile: Ok, I won't argue.

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: Oops! Yes, indeed, although in this case the order of coordinates is different. It goes like: `eigen[i, zcoords, ycoords, xcoords]`, with `i = 0, 1, 2`, is/are the `(i + 1)`th largest eigenvalue(s) at position/neighbourhood `(xcoords, ycoords, zcoords)`.

Thanks to your review, I've noticed something inconsistent in my narrative: I'm showing the bottom plane (z = 0) but then I'm describing the situation at z = 3, since that's where the overall largest eigenvalue is found. Fixing now!

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: Absolutely. Fixed in 521d4d235e86ec64e709e586fc864850b45044fb. Should we draw the gradient directions on the image? As in, roughly, manually, with a red marker?

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: Done in 521d4d235e86ec64e709e586fc864850b45044fb.

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: Right, fixed in 521d4d235e86ec64e709e586fc864850b45044fb.

> The structure_tensor docstring indicates sigma should be a float, but it looks like the function can accept a unique sigma per axis (e.g. sigma=(3, 3, 1)). If that is intentional, we should update the docstring to mention it.

PS: Nice find! TODO...

mkcor: Done over at #5196.

doc/examples/applications/plot_3d_structure_tensor.py
grlee77: I was suggesting a figure something like this, which shows the spatial distribution of the eigenvalues. Do you think it is helpful?


```Python
fig, axes = plt.subplots(1, 3)
vmax = eigen[0][coords[1]].max()
axes[0].imshow(eigen[0][coords[1]], vmin=0, vmax=vmax, cmap=plt.cm.viridis)
axes[0].set_title('eigenvalue 0')
axes[1].imshow(eigen[1][coords[1]], vmin=0, vmax=vmax, cmap=plt.cm.viridis)
axes[1].set_title('eigenvalue 1')
axes[2].imshow(eigen[2][coords[1]], vmin=0, vmax=vmax, cmap=plt.cm.viridis)
axes[2].set_title('eigenvalue 2')
for ax in axes:
    ax.set_axis_off()
```
evs

grlee77 (approved):
Thanks @mkcor. I am approving this now, but please see one other suggestion regarding potentially showing the eigenvalue images themselves.

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: I'm aware I hadn't addressed all of your review comments -- looking now! Thanks.

mkcor: Yes, I think it is helpful: It makes it easier to follow the rest of the story. I've added the plot in d82d754. I couldn't resist but use `facet_col` with plotly's imshow... :heart: @emmanuelle :wink:

rfezzani:
Thank you @mkcor :tada:.

PR #5013, opened 2020-10-10
Added README to binary (wheel) distribution

jasmcaus:
Added the README.md file to the binary wheel distribution

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,3 @@
 [metadata]
+description_file = README.md
 license_file = LICENSE.txt


hmaarrfk (approved):

jni (approved):
LGTM, thank you!

jni:
I'm just going to let this sit another day to give more devs a chance to weigh in, but I don't anticipate that it will be controversial.

hmaarrfk (changes requested):
Can you provide us with some documentation that states that this is the currest best practice to include this kind of long description?

hmaarrfk:
It just seems like there are a bunch of non-standard keys.

https://github.com/pypa/setuptools/blob/master/setup.cfg#L23
https://stackoverflow.com/questions/60084128/does-description-file-in-setup-cfg-section-metadata-have-any-effect

hmaarrfk:
Maybe we want to focus on https://github.com/scikit-image/scikit-image/pull/5015 instead?

jasmcaus:
This PR is addressed in #5015. Closing this now.

jni:
> I don't anticipate that it will be controversial.

Never make a prediction, especially about the future. =P

PR #5015, opened 2020-10-11
Updates to setup.py

jasmcaus:
A quick summary of what's changed:
- a `settings.ini` file was created which contains all the package information (distname, maintainer, email, etc) and configurations were parsed based off these configurations
- `LONG_DESCRIPTION` was refactored to be on one line (imported the `io` package for this)
- Python check for >=3.6.0 was placed just after the imports to make the most logical sense


setup.py
```diff
@@ -106,21 +114,20 @@ def build_extensions(self):
...
-
+# Grabbing __version__
 with open('skimage/__init__.py', encoding='utf-8') as fid:
     for line in fid:
         if line.startswith('__version__'):
             VERSION = line.strip().split()[-1][1:-1]
             break
 
-
+# Parsing requirements file
```
hmaarrfk, line 124: Is this comment necessary? The function name contains the exact same English root words.
```suggestion
```

settings.ini
```diff
@@ -0,0 +1,10 @@
+[DEFAULT]
```
hmaarrfk, line 1: Does an other package use `settings.ini` file for this?

Maybe we should just expand on setup.cfg

https://github.com/pypa/setuptools/blob/master/setup.cfg

jasmcaus: I know Numpy does, so too my package Caer, but I suppose there isn't much difference using a `settings.ini` file over expanding over the `setup.cfg`. Needless, I've removed `settings.ini` and moved all configurations to `setup.cfg`.

setup.cfg
```diff
@@ -1,2 +1,10 @@
 [metadata]
-license_file = LICENSE.txt
```
hmaarrfk, line 2: license_file should be kept.

settings.ini
hmaarrfk: Can we reference numpy's `settings.ini` or `setup.cfg` file by any chance? We typically follow their lead on certain things like this.

jasmcaus: I must have got mixed up with another package. Working with far too many these days! [Fastai](https://github.com/fastai/fastai/blob/master/settings.ini) has a `settings.ini` file to separate out meta information from `setup.py`. While they use a `settings.ini` file, I suggest sticking with setup.cfg (since that's where the configuration variables are defined).

setup.py
```diff
@@ -1,48 +1,70 @@
...
+
+# All settings are in configs.ini
+config = ConfigParser(delimiters=['='])
+config.read('setup.cfg')
+cfg = config['metadata']
+
+
+DISTNAME = cfg['name']
+DESCRIPTION = cfg['description']
+LONG_DESCRIPTION = io.open('README.md', encoding='utf-8').read()
```
hmaarrfk, line 39: The old way used to ensure that the file was closed. Can you point to documentation that states that this would be closed after the read is complete?

setup.cfg
```diff
@@ -1,2 +1,10 @@
 [metadata]
+name = scikit-image
+description = Image Processing in Python
+author = Stefan van der Walt
+author_email = stefan@sun.ac.za
+url = https://scikit-image.org
+git_url = https://github.com/scikit-image/scikit-image
+download_url = https://scikit-image.org/docs/stable/install.html
 license_file = LICENSE.txt
```
hmaarrfk, line 9:
```suggestion
license_file = LICENSE.txt
license = Modified BSD
```
I would keep the old text as well. Including the line `license_file` was simply the only way I found to include the license file in the source distribution.

Didn't you want to include the README file in the wheel distribution?  You had added an other line to this in your other PR

setup.py
```diff
@@ -1,48 +1,70 @@
...
-}
-
-with open('README.md', encoding='utf-8') as f:
-    LONG_DESCRIPTION = f.read()
-
-if sys.version_info < (3, 6):
-
-    error = """Python {py} detected.
-
+    error = f"""Python {'.'.join([str(v) for v in sys.version_info[:3]])}
```
hmaarrfk, line 20: you can't have fstrings before python 3.6. So when this error is hit, it need to be in syntax that is compatible with old versions of python.

Namely, we are still interested in providing this error message to those that are still using python 2.7 for whatever reason.

I think it is best to revert.

setup.py
```diff
@@ -1,48 +1,70 @@
...
     sys.exit(1)
 
+
+# All settings are in configs.ini
+config = ConfigParser(delimiters=['='])
+config.read('setup.cfg')
+cfg = config['metadata']
+
+
+DISTNAME = cfg['name']
```
hmaarrfk, line 37: I'm not sure you need to parse anything yourself.

https://github.com/pypa/setuptools/blob/master/setup.py

They don't do any of this parsing. I think setuptools will do this automatically for you if you don't specify the parameters.

Honestly. This kind of change is somewhat of a side-step. There are advantages to "declarative" build specifications, but I'm wondering if there is any measurable advantage to this change other than "keeping up with a shiny new tool from setuptools"

hmaarrfk:
I think I've given this a thorough pass.

Generally, we are quite weary of making such drastic changes to setup.py since the changes are only really visible one we've uploaded to pypi.

This is quite costly since a release process involves many parties.

Generally, I would like to revert to the more succinct and targeted suggestion you had

https://github.com/scikit-image/scikit-image/pull/5013

Would that be OK?

setup.py
jasmcaus: [Yes](https://stackoverflow.com/questions/51556149/does-with-io-open-close-the-file-automatically), it closes automatically

setup.py
jasmcaus: Setuptools used a different way of parsing the configurations, of which I am unaware of. However, I have seen a number of popular packages use this method (which the PR is about). Honestly, there is no measurable advantage, but it helps neatly separate out the configuration variables.

jasmcaus:
I doubt this PR proffers such a drastic change to the setup. The original information remains the same; the only difference is, perhaps, where the configuration variables are located (see my previous comment). Overall, this does not (and will not) have such a huge impact, but is a neat way of, again, separating out the configuration variables.

hmaarrfk:
You are right that it is a `neat` way of doing something. 

Unfortunately, that isn't always the goal of the project. Rather, bugs in the setup.py will only becaught by our end users once the whole release has been completed (which only happens something like twice a year).

Oddly, I'm torn by two things:
1. I would rather not really change how setup.py is done. This has serious implications in the build and release process, something that isn't clear until release time.
2. I think you aren't taking this PR far enough. See for example https://github.com/scipy/oldest-supported-numpy where the setup.py file is essentially blank.

If this PR is merged, I would rather converge toward 2. On the other hand, if nothing needs to be changed, I would rather move forward with the suggestions in #5013.

jni:
Hi @jasmcaus, thanks for contributing!

I'm mostly ok with this, but I sort of agree with @hmaarrfk that this could/should/might as well go further at this point. We can actually put the trove classifiers in the setup.cfg, and more, see e.g. napari's [setup.cfg](https://github.com/napari/napari/blob/master/setup.cfg) and [setup.py](https://github.com/napari/napari/blob/master/setup.py). 

I *don't* think this is untestable before release — we can build wheels and source distributions locally and try them out. But, since we are about to do a release, we should probably stick with the current setup until 0.18 is out (very shortly! 😬), then press on with these changes.

My proposed course of action is:
- release 0.18
- merge this PR more or less as is, after some local testing — I don't think it's a big deal and it takes a small step in the right direction
- add the further changes proposed by @hmaarrfk and other projects' configs in follow-ups. Of course @jasmcaus if you get some of those changes in on this PR in the meantime, even better!

Hopefully that balances all the concerns nicely enough...?

jasmcaus:
@jni that's perfect! Will update the PR with new changes

hmaarrfk:
I'm more concerned as to how the packages will appear to users on pypi. 

I think there is a pypi testing website that we should likely start using.

jni:
@hmaarrfk 

> I'm more concerned as to how the packages will appear to users on pypi.

I don't think this PR affects that? Am I missing something?

jni:
Also, is there something I should be worried about with https://pypi.org/project/napari/?

jasmcaus:
I highly doubt that this will affect the package metadata on PyPi. I use this same technique in my package [caer](https://pypi.org/project/caer/) and there is no difference if the configurations were hard-coded in.

jasmcaus:
> Also, is there something I should be worried about with https://pypi.org/project/napari/?

Setuptools used a different way of parsing the configurations, of which I am unaware of. However, I have seen a number of popular packages use this method (which the PR is about).

hmaarrfk:
Pypi pulls the metadata from setup and setup.cfg

It's not a big deal. They should in theory be equivalent. 

But in theory, theory and practice are the same.

PR #5016, opened 2020-10-12
Bump numpy versions to match scipy (kinda)

hmaarrfk:
## Description
I'm waiting to hear back on the reasons why scipy held back bumping AIX numpy requirements.

https://github.com/scipy/scipy/issues/12940

Hopefully this should help with the failures on the minimum version requirement.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,13 +3,11 @@ requires = [
     "wheel",
     "setuptools",
     "Cython>=0.29.18",
-    # AIX is IBM's variant of UNIX; Scipy uses special handling for it.
-    # Therefore we decided to handle it in the same way.
-    # https://github.com/scipy/scipy/pull/10431#issue-295816438
-    "numpy==1.15.1; python_version=='3.6' and platform_system!='AIX'",
-    "numpy==1.15.1; python_version=='3.7' and platform_system!='AIX'",
-    "numpy==1.17.3; python_version>='3.8' and platform_system!='AIX'",
-    "numpy==1.16.0; python_version=='3.6' and platform_system=='AIX'",
-    "numpy==1.16.0; python_version=='3.7' and platform_system=='AIX'",
-    "numpy==1.17.3; python_version>='3.8' and platform_system=='AIX'",
+    # We follow scipy for much of these pinnings
+    # https://github.com/scipy/scipy/blob/master/pyproject.toml
+    "numpy==1.16.5; python_version=='3.6'",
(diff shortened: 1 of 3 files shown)


jni:
@hmaarrfk coolio, let us know when this is ready for review/merge.

hmaarrfk:
rgrommers replied and indicated that the current difference between AIX and non AIX might have been an oversight.

This is ready now.

jni:
> This is ready now.

Travis disagrees. 😂 But actually I don't think it's your fault, I think plotly isn't happy with minimum requirements. @emmanuelle the [error](https://travis-ci.org/github/scikit-image/scikit-image/jobs/735359633#L6501-L6504) is:

```pytb
  File "doc/examples/segmentation/plot_regionprops.py", line 98, in <module>
    fig = px.imshow(img, binary_string=True)
TypeError: imshow() got an unexpected keyword argument 'binary_string'
```

I think this feature was only added in the most recent version of plotly, right? Do we need to update our minimum plotly version requirement in the docs requirements?

The other error is also in the docs, a failed download from pivchallenge.org, so I'm approving this.

hmaarrfk:
The CIs have been really problematic. I thought I had fixed them yesterday with an other PR.

emmanuelle:
@jni I opened #5021 re plotly.

In this PR, why do we need to pin exact versions and not just specify a minimum version? Sorry @hmaarrfk if this has already been explained somewhere else.

hmaarrfk:
Pyproject toml specifies *build time* dependencies for those building from source using pip tools.

Users on the platforms we support will never interact with this.

However, some users of pip on exotic platforms have asked for it to be kept up to date,and we have obliged.
There, we want to instruct pip to create an environment similar to the ones we have in scikit-image wheels, that is, oldest numpy that we support, before taking the result and installing it in their virtual env satisfying the greater than dependencies we specify.

jni:
> The CIs have been really problematic.

Yeah, let's turn them off, seemed like a good idea at the time but they're just annoying. 🤣 

More seriously, I'm happy to merge this as is, or do you want to rebase on master now that #5021 is in?

emmanuelle:
@jni in some projects on Github there is an "update branch" button which allows to merge master to the PR branch, and it's very convenient. Maybe it's available only for branches of the same repo and not on a fork? I don't really understand which repos / PRs have this button.

emmanuelle:
Thanks @hmaarrfk merging!

Issue #5019, opened 2020-10-13
MemoryError in computing Convex Hull of a large 3D image


kushaangupta:
## Description
While following the discussion [here](https://forum.image.sc/t/convex-hull-of-3d-image-with-sparsely-dispersed-particles/43903), following error was discovered in `skimage.morphology.convex_hull_image`.

## Way to reproduce
[Source](https://forum.image.sc/t/convex-hull-of-3d-image-with-sparsely-dispersed-particles/43903/4)
```python
import numpy as np
from skimage.morphology import convex_hull_image

tmp = np.random.randn(100, 1024, 1024) > 0
ch = convex_hull_image(tmp)
```

Output:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\user\AppData\Local\Programs\Python\Python38\lib\site-packages\skimage\morphology\convex_hull.py", line 89, in convex_hull_image
    coords_in_hull = np.all(hull.equations[:, :ndim].dot(gridcoords) +
MemoryError: Unable to allocate 67.2 GiB for an array with shape (86, 104857600) and data type float64
```

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)]
Windows-10-10.0.16299-SP0
scikit-image version: 0.17.2
numpy version: 1.18.5
```

jni:
I think this was closed by #5020. Anyone stumbling on the same issue please open a new issue, as further fixes will require additional code.

PR #5020, opened 2020-10-14
ENH: Reduce memory use for convex_hull_image by checking hull inequalities in sequence

kushaangupta:
## Description

Fixes issue #5019 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/convex_hull.py
+++ b/skimage/morphology/convex_hull.py
@@ -19,12 +19,12 @@ def _offsets_diamond(ndim):
 
 
 def _check_coords_in_hull(gridcoords, hull_equations, tolerance):
-    r"""Checks all the co-ordinates for inclusiveness in the convex hull.
+    r"""Checks all the coordinates for inclusiveness in the convex hull.
 
     Parameters
     ----------
     gridcoords : (M, N) ndarray
-        Co-ordinates of ``N`` points in ``M`` dimensions.
+        Coordinates of ``N`` points in ``M`` dimensions.
     hull_equations : (M, N) ndarray
         Hyperplane equations of the facets of the convex hull.
     tolerance : float
@@ -40,10 +40,10 @@ def _check_coords_in_hull(gridcoords, hull_equations, tolerance):
 
     Notes
     -----
-    Check inclusiveness of co-ordinates in convex hull requires intermediate
-    calculations of dot products which are memory-intensive. Thus, the convex
-    hull equations are checked individually with all co-ordinates to keep
-    within the memory limit.
+    Checking the inclusiveness of coordinates in a convex hull requires
+    intermediate calculations of dot products which are memory-intensive.
+    Thus, the convex hull equations are checked individually with all
+    coordinates to keep within the memory limit.
 
     References
     ----------
@@ -55,13 +55,13 @@ def _check_coords_in_hull(gridcoords, hull_equations, tolerance):
     coords_in_hull = np.ones(n_coords, dtype=np.bool_)
 
     # Pre-allocate arrays to cache intermediate results for reducing overheads
-    dot_array = np.zeros(n_coords, dtype=np.float64)
-    test_ineq_temp = np.zeros(n_coords, dtype=np.float64)
-    coords_single_ineq = np.zeros(n_coords, dtype=np.bool_)
+    dot_array = np.empty(n_coords, dtype=np.float64)
+    test_ineq_temp = np.empty(n_coords, dtype=np.float64)
+    coords_single_ineq = np.empty(n_coords, dtype=np.bool_)
 
     # A point is in the hull if it satisfies all of the hull's inequalities
     for idx in range(n_hull_equations):
-        # Tests a hyperplane equation on all co-ordinates of volume
+        # Tests a hyperplane equation on all coordinates of volume
         np.dot(hull_equations[idx, :ndim], gridcoords, out=dot_array)
         np.add(dot_array, hull_equations[idx, ndim:], out=test_ineq_temp)
         np.less(test_ineq_temp, tolerance, out=coords_single_ineq)
--- a/skimage/morphology/convex_hull.py
+++ b/skimage/morphology/convex_hull.py
@@ -40,9 +40,10 @@ def _check_coords_in_hull(gridcoords, hull_equations, tolerance):
 
     Notes
     -----
-    The check requires intermediate calculation of dot product which is
-    memory-intensitve. Thus, volume of ``gridcoords`` is broken down into
-    ``chunks`` to keep within the memory limit.
+    Check inclusiveness of co-ordinates in convex hull requires intermediate
+    calculations of dot products which are memory-intensive. Thus, the convex
+    hull equations are checked individually with all co-ordinates to keep
+    within the memory limit.
 
     References
     ----------
@@ -50,27 +51,19 @@ def _check_coords_in_hull(gridcoords, hull_equations, tolerance):
 
     """
     ndim, n_coords = gridcoords.shape
-    coords_in_hull = np.zeros(n_coords)
-    chunk_size = n_coords
+    n_hull_equations = hull_equations.shape[0]
+    coords_in_hull = np.ones(n_coords, dtype=np.bool_)
+
+    # Pre-allocate arrays to cache intermediate results for reducing overheads
+    dot_array = np.zeros(n_coords, dtype=np.float64)
+    test_ineq_temp = np.zeros(n_coords, dtype=np.float64)
 
     # A point is in the hull if it satisfies all of the hull's inequalities
-    def _test_ineq(coords):
-        return np.all(hull_equations[:, :ndim].dot(coords)
-                      + hull_equations[:, ndim:] < tolerance, axis=0)
-
-    # Find the right chunk size
-    while chunk_size:
-        try:
-            tested_chunk = _test_ineq(gridcoords[:, :chunk_size])
-            coords_in_hull[:chunk_size] = tested_chunk
-            break
-        except MemoryError:
-            chunk_size = (chunk_size + 1) // 2
-
-    # Apply the test in chunks
-    for idx in range(chunk_size, n_coords, chunk_size):
-        tested_chunk = _test_ineq(gridcoords[:, idx: idx + chunk_size])
-        coords_in_hull[idx: idx + chunk_size] = tested_chunk
+    for idx in range(n_hull_equations):
+        # Tests a hull equation on all co-ordinates of volume
+        np.dot(hull_equations[idx, :ndim], gridcoords, out=dot_array)
+        np.add(dot_array, hull_equations[idx, ndim:], out=test_ineq_temp)
+        coords_in_hull *= test_ineq_temp < tolerance
 
     return coords_in_hull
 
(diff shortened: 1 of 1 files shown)


skimage/morphology/convex_hull.py
```diff
@@ -86,8 +112,8 @@ def convex_hull_image(image, offset_coordinates=True, tolerance=1e-10):
         gridcoords = np.reshape(np.mgrid[tuple(map(slice, image.shape))],
                                 (ndim, -1))
         # A point is in the hull if it satisfies all of the hull's inequalities
```
jni, line 114: This comment should probably be moved to `_test_ineq`, above.

skimage/morphology/convex_hull.py
```diff
@@ -18,6 +18,32 @@ def _offsets_diamond(ndim):
     return offsets
 
 
+def _check_coords_in_hull(gridcoords, hull_equations, tolerance):
```
jni, line 21: It would be great to have a docstring here that briefly explains why this is needed (it's a memory-hungry method) and what it does. It can include a link to the issue, for example.

jni:
Thank you for this PR @kushaangupta! It's great when issues get resolved as soon as they're raised! 😊 As mentioned in the comments above, it would be awesome if you could add some docstrings that explain what this fix is doing. It would also be great to add a test, although maybe it should be off by default.

jni:
> off by default.

**skipped** by default, sorry

skimage/morphology/convex_hull.py
kushaangupta: Done

skimage/morphology/convex_hull.py
kushaangupta: Done

kushaangupta:
> Thank you for this PR @kushaangupta! It's great when issues get resolved as soon as they're raised! 😊 As mentioned in the comments above, it would be awesome if you could add some docstrings that explain what this fix is doing. It would also be great to add a test, although maybe it should be off by default.

Thanks a lot for the appreciation and being a great help resolving my doubts @jni !

I'm not sure how I should add test cases for this function. The example [here](https://forum.image.sc/t/convex-hull-of-3d-image-with-sparsely-dispersed-particles/43903/4) tests the whole skimage.morphology.convex_hull_image function. Should I randomly generate Convex hull's hyperplane equations?

jni:
You can copy the lines from that function into a test to create a "big" hull:

https://github.com/scikit-image/scikit-image/blob/ef5ac57b1585d0236916bf22dbfa1278f0d0e0d5/skimage/morphology/convex_hull.py#L61-L80

In fact, the hull itself is actually quite small, 50 planes or so based on the error message — you could generate enough random points in a volume of that size, and compute the hull of those points.

> Is it okay that the functionality for finding the chunk size relies on while loop with break?

This is a good point of self-review! 😂 I think it is worth fixing for sure! Checking for a chunk size of 1 should do it!

To be honest, one thing that *could* be worthwhile is to time it in this form vs directly jumping to a chunk size of 1 — I suspect that, given the number of points being checked relative to the number of equations, it's just not worth stepping down the chunk size slowly and we should go straight to solving one inequality at a time. How does that sound?

Actually, if you go straight to that (never try the giant volume), you don't even need to write a new test — the code will be covered by existing tests!

kushaangupta:
> To be honest, one thing that _could_ be worthwhile is to time it in this form vs directly jumping to a chunk size of 1 — I suspect that, given the number of points being checked relative to the number of equations, it's just not worth stepping down the chunk size slowly and we should go straight to solving one inequality at a time. How does that sound?
> 
> Actually, if you go straight to that (never try the giant volume), you don't even need to write a new test — the code will be covered by existing tests!

I believe there's another complication to determine a chunk size. I copied my function into Colab notebook and found that the runtime crashes due to excessive memory usage even though try-except clause is present. 

My rationale for determining highest possible chunk_size within the memory limits was to get most out of numpy's greater speed, compared to looping in Python.

If we have to keep the chunk size of 1 only, won't it be better to change the helper function to Cython? (for faster iterations)

jni:
The point is that the number of equations is tiny -- in Volker's example, 50? So the Python for-loop overhead cost will be tiny. I think a for-loop is ~200￼￼￼ns, so 50 loops would be 10ms. I think even allocating the output array will take longer.

kushaangupta:
It depends highly on the dataset. I was working with a image with 384 equations. After removing the while loop and setting `chunk_size = 1` took ~5 minutes to compute convex hull image, but setting `chunk_size = 5` halved the runtime. Greater chunk size does seem to be much efficient.

Setting `chunk_size` to a constant can improve efficiency, but I don't think it's good to hard-code one into the function. Can we decide it dynamically while being conscious of the memory limits.

Or should I just set `chunk_size = 1` and remove the while loop?

jni:
@kushaangupta That's interesting, thanks for doing that test! I also find it surprising. I think the reason is not because of the for-loop overhead, but because each equation test actually results in the creation of very large arrays. So you are spending a lot of time allocating and deallocating memory for huge arrays.

I just realised that we are chunking along the coordinates axis, rather than along the equations, which changes my calculations. There's obviously a lot more coordinates than there are equations, in most cases at least, so you are doing a lot more iteration in Python.

So, my proposal:

- use chunk size of 1 anyway — 2x is actually not a huge factor for the complexity gain, and anyway, we can optimise later.
- potentially, chunk along equations rather than coordinates. Then on each loop you do `coords_in_hull *= test_ineq(hull_equations[i])`
- optimise memory allocation by pre-allocating all the arrays you need and using the `out=` keyword argument of NumPy functions, e.g.

```python
np.add(
    np.dot(hull_equation[:ndim], coords, out=dot_array),
    hull_equation[ndim:],
    out=test_ineq_temp,
)
```

I think that will eliminate most of the speedup you see from a large chunk size.

kushaangupta:
@jni I have made the changes accordingly.

skimage/morphology/convex_hull.py
```diff
@@ -18,6 +18,56 @@ def _offsets_diamond(ndim):
...
+    # Pre-allocate arrays to cache intermediate results for reducing overheads
+    dot_array = np.zeros(n_coords, dtype=np.float64)
+    test_ineq_temp = np.zeros(n_coords, dtype=np.float64)
+
+    # A point is in the hull if it satisfies all of the hull's inequalities
+    for idx in range(n_hull_equations):
+        # Tests a hull equation on all co-ordinates of volume
+        np.dot(hull_equations[idx, :ndim], gridcoords, out=dot_array)
+        np.add(dot_array, hull_equations[idx, ndim:], out=test_ineq_temp)
+        coords_in_hull *= test_ineq_temp < tolerance
```
jni, line 66: This could also use a pre-allocated array using `np.less(test_ineq_temp, tolerance, out=coords_single_ineq)`. But this is less readable so we should make sure there is a measurable improvement in performance before doing this.

jni (approved):
Wonderful @kushaangupta! Does this work for your original problem and how does it compare in performance to the test case you mentioned? (5min vs 2min30s)?

Anyway for me the code looks beautiful and I'm happy to get this in!

skimage/morphology/convex_hull.py
kushaangupta: > Does this work for your original problem and how does it compare in performance to the test case you mentioned? (5min vs 2min30s)?

Yes, and it took on average ~14 minutes. Also, one time there was a MemoryError of allocating space for an array with dimensions (ndim, n_coords) while computing the dot product, but I wasn't able to reproduce the error.

kushaangupta: > But this is less readable so we should make sure there is a measurable improvement in performance before doing this.

It took 18 seconds less for my test case

jni: > it took on average ~14 minutes

Yikes, that's not so great! 😂 It's ok though, I'm happy to get this merged like this and work on optimising the performance in a subsequent PR.

I suspect it might be worth applying the `np.less()` approach I suggested here for a gain in performance. Right now, for every loop, you need to allocate a massive array for `test_ineq_temp < tolerance`, and then, because you are hitting your RAM limit, deallocate it.

You can try to verify this with [`line_profiler`](https://github.com/pyutils/line_profiler).

Again, performance fix can come before or after this PR, I don't think it's important compared to getting this working.

@scikit-image/core anyone else up for approving this?

kushaangupta: Wait, I'll add the `np.less()` approach.

skimage/morphology/convex_hull.py
```diff
@@ -18,6 +18,58 @@ def _offsets_diamond(ndim):
     return offsets
 
 
+def _check_coords_in_hull(gridcoords, hull_equations, tolerance):
+    r"""Checks all the co-ordinates for inclusiveness in the convex hull.
```
grlee77, line 22:
```suggestion
    r"""Checks all the coordinates for inclusiveness in the convex hull.
```

skimage/morphology/convex_hull.py
```diff
@@ -18,6 +18,58 @@ def _offsets_diamond(ndim):
...
+
+    Parameters
+    ----------
+    gridcoords : (M, N) ndarray
+        Co-ordinates of ``N`` points in ``M`` dimensions.
```
grlee77, line 27:
```suggestion
        Coordinates of ``N`` points in ``M`` dimensions.
```

skimage/morphology/convex_hull.py
```diff
@@ -18,6 +18,58 @@ def _offsets_diamond(ndim):
...
+    coords_in_hull : ndarray of bool
+        Binary 1D ndarray representing points in n-dimensional space
+        with value ``True`` set for points inside the convex hull.
+
+    Notes
+    -----
+    Check inclusiveness of co-ordinates in convex hull requires intermediate
+    calculations of dot products which are memory-intensive. Thus, the convex
+    hull equations are checked individually with all co-ordinates to keep
+    within the memory limit.
```
grlee77, line 46:
```suggestion
    Checking the inclusiveness of coordinates in a convex hull requires
    intermediate calculations of dot products which are memory-intensive. 
    Thus, the convex hull equations are checked individually with all
    coordinates to keep within the memory limit.
```

skimage/morphology/convex_hull.py
```diff
@@ -18,6 +18,58 @@ def _offsets_diamond(ndim):
...
+
+    """
+    ndim, n_coords = gridcoords.shape
+    n_hull_equations = hull_equations.shape[0]
+    coords_in_hull = np.ones(n_coords, dtype=np.bool_)
+
+    # Pre-allocate arrays to cache intermediate results for reducing overheads
+    dot_array = np.zeros(n_coords, dtype=np.float64)
+    test_ineq_temp = np.zeros(n_coords, dtype=np.float64)
+    coords_single_ineq = np.zeros(n_coords, dtype=np.bool_)
```
grlee77, line 60:
```suggestion
    dot_array = np.empty(n_coords, dtype=np.float64)
    test_ineq_temp = np.empty(n_coords, dtype=np.float64)
    coords_single_ineq = np.empty(n_coords, dtype=np.bool_)
```

empty can save time by not initializing to zeros

grlee77 (changes requested):
This looks good to me aside from some minor suggestions

kushaangupta:
> This looks good to me aside from some minor suggestions

Thanks for your suggestions @grlee77. I've made the changes accordingly

grlee77 (approved):

grlee77:
@jni, I have now approved if you want to do a final check and merge

jni:
Awesome! Thank you @kushaangupta for contributing this and iterating with us! :tada:

Issue #4814, opened 2020-07-01
IndexError when calling hough_line_peaks


Mark531:
## Description
Sometimes, when the list of tested angles passed to Hough_line is too narrow, hough_line_peaks triggers an IndexError.

## Way to reproduce
```
tested_angles = np.array([np.pi / 2])
h, theta, d = hough_line(image, theta=tested_angles)
accum, angles, dists = hough_line_peaks(h, theta, d, threshold=2)
```
triggers
```
Traceback (most recent call last):
  File "<ipython-input-147-b875ad1f50fd>", line 3, in <module>
    accum, angles, dists = hough_line_peaks(h, theta, d, threshold=2)
  File "C:\Users\Mark4\anaconda3\lib\site-packages\skimage\transform\hough_transform.py", line 63, in hough_line_peaks
    num_peaks=num_peaks)
  File "C:\Users\Mark4\anaconda3\lib\site-packages\skimage\feature\peak.py", line 316, in _prominent_peaks
    img_max[ycoords_nh, xcoords_nh] = 0
IndexError: index -9 is out of bounds for axis 1 with size 1
```

But
```
tested_angles = np.linspace(np.pi / 2 - 0.01, np.pi / 2  + 0.01, 100)
h, theta, d = hough_line(image, theta=tested_angles)
accum, angles, dists = hough_line_peaks(h, theta, d, threshold=2)
```
works.

I don't understand what the problem is, it seems that no line is found with the given angle. But in that case, the Hough accumulator should stay to 0. I uploaded the image that is needed to reproduce the issue. It can be loaded with `numpy.load`.

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
3.7.7 (default, May  6 2020, 11:45:54) [MSC v.1916 64 bit (AMD64)]
Windows-10-10.0.18362-SP0
scikit-image version: 0.16.2
numpy version: 1.18.1
```
[image.zip](https://github.com/scikit-image/scikit-image/files/4861273/image.zip)

jni:
I can reproduce this on master with just a random image:

```python
import numpy as np
from skimage.transform import hough_line, hough_line_peaks

image = np.random.random((512, 512))
tested_angles = np.array([np.pi / 2])
[...]

Issue #5022, opened 2020-10-15
Store pivchallenge data on gitlab


jni:
## Description

I'm seeing a fair number of failed builds due to pivchallenge.org rejecting http requests. I think it is not designed to get hammered by CI builds! 😂 We should contact the authors (this is on my to-do, haven't gotten around to it yet) to ask whether we could host that image ourselves.

example build:

https://travis-ci.org/github/scikit-image/scikit-image/jobs/735724132

```pytb
/home/travis/build/scikit-image/scikit-image/doc/examples/registration/plot_opticalflow.py failed to execute correctly: Traceback (most recent call last):
  File "/home/travis/build/scikit-image/scikit-image/doc/examples/registration/plot_opticalflow.py", line 92, in <module>
    image0 = imageio.imread("http://www.pivchallenge.org/pub/B/B001_1.tif")
  File "/home/travis/venv/lib/python3.6/site-packages/imageio/core/functions.py", line 265, in imread
    reader = read(uri, format, "i", **kwargs)
  File "/home/travis/venv/lib/python3.6/site-packages/imageio/core/functions.py", line 186, in get_reader
    return format.get_reader(request)
  File "/home/travis/venv/lib/python3.6/site-packages/imageio/core/format.py", line 170, in get_reader
    return self.Reader(self, request)
  File "/home/travis/venv/lib/python3.6/site-packages/imageio/core/format.py", line 221, in __init__
    self._open(**self.request.kwargs.copy())
  File "/home/travis/venv/lib/python3.6/site-packages/imageio/plugins/tifffile.py", line 222, in _open
    self._f = f = open(self.request.get_local_filename(), "rb")
  File "/home/travis/venv/lib/python3.6/site-packages/imageio/core/request.py", line 377, in get_local_filename
    shutil.copyfileobj(self.get_file(), file)
  File "/home/travis/venv/lib/python3.6/site-packages/imageio/core/request.py", line 353, in get_file
    self._file = urlopen(self.filename, timeout=float(timeout))
  File "/home/travis/venv/lib/python3.6/site-packages/imageio/core/util.py", line 34, in urlopen
    return urlopen(*args, **kwargs)
  File "/opt/python/3.6.3/lib/python3.6/urllib/request.py", line 223, in urlopen
    return opener.open(url, data, timeout)
  File "/opt/python/3.6.3/lib/python3.6/urllib/request.py", line 526, in open
    response = self._open(req, data)
[...]

PR #5024, opened 2020-10-18
Fix IndexError when calling hough_line_peaks with too few angles

cjblocker:
## Description

`hough_line_peaks` has a default parameter `min_angle=10` which does not make sense if the second dimension of the accumulator array is less than 10. In such cases, it would raise an IndexError.

Fixes gh-4814 with the single line fix suggested [in the issue's comments](https://github.com/scikit-image/scikit-image/issues/4814#issuecomment-654698200).

## For reviewers

  later.

### DIFF
--- a/skimage/transform/hough_transform.py
+++ b/skimage/transform/hough_transform.py
@@ -58,6 +58,7 @@ def hough_line_peaks(hspace, angles, dists, min_distance=9, min_angle=10,
     """
     from ..feature.peak import _prominent_peaks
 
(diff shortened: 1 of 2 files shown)


jni (approved):
Thanks @cjblocker! This looks good to me, the travis failure is caused by #5022 and unrelated to these changes.

jni:
And thank you in particular for the regression test! :tada:

rfezzani (approved):
Thank you @cjblocker, and sorry for the delay.

Issue #3830, opened 2019-04-07
Examples wanted for the gallery (again)


emmanuelle:
This issue is a follow-up of #1231.

Some functions of the API are not illustrated in gallery examples, which makes them harder to discover and understand for beginners. To know which examples are missing, the API doc http://scikit-image.org/docs/dev/api/api.html has links to gallery examples (in a mini gallery) for each function. Functions without a mini-gallery do not appear in a gallery example. 

This issue is a good place to discuss about future examples to be added.

Functions without a gallery example (list to be completed):

- [ ] `measure.points_in_poly`

jni:
Thanks for flagging this, @emmanuelle! I would say that we should make #3384 and remote data a prerequisite for this, so that new examples written use more scientifically relevant data.

PR #5029, opened 2020-10-21
Fix point_in_polygon

### LINKED ISSUE #3892: some edge points not included in points_in_poly result
## Description
I've been trying to rasterize points in a polygon (like `grid_points_in_poly`) but have been getting unintuitive/incorrect results due to certain edge points being deemed not inside the polygon by your algorithm.

Specifically, the problem seems to arise because the points are counted inclusively only on the lesser side of the equation (`min <= point < max`).

This is most obvious when checking if a polygon's vertices are counted as inside the polygon:
```python
In [1]: from skimage.measure import points_in_poly

In [2]: verts = [(0, 0), (1, 0), (0, 1)]

In [3]: points_in_poly(verts, verts)
Out[3]: array([ True, False, False])
```

I wanted to check in if this is intentional or a bug since this is not the output I would expect.

rfezzani:
## Description

Fixes #3892 and probably some bugs in the `draw` module...

See @hmaarrfk https://github.com/scikit-image/scikit-image/issues/3892#issuecomment-490714349 for explanations about the proposed solution.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/_shared/geometry.pyx
```diff
@@ -2,54 +2,96 @@
...
+    xp, yp : np_floats array
         Coordinates of polygon with length nr_verts.
-    x, y : double
+    x, y : np_floats
         Coordinates of point.
+
+    Returns
+    -------
+    c : unsigned char
+        Point relative position to the polygone O: outside, 1: inside,
```
jni, line 25:
```suggestion
        Point relative position to the polygon O: outside, 1: inside,
```

skimage/_shared/geometry.pyx
```diff
@@ -2,54 +2,96 @@
...
+           Second Edition, Cambridge Unversity Press, Chapter 7
     """
     cdef Py_ssize_t i
-    cdef unsigned char c = 0
-    cdef Py_ssize_t j = nr_verts - 1
+    cdef Py_ssize_t nr_verts = xp.shape[0]
+    cdef np_floats x0, x1, y0, y1, eps
+    cdef unsigned int l_cross = 0, r_cross = 0
+
+    # Tolerence for vertices labelling
```
jni, line 38:
```suggestion
    # Tolerance for vertices labelling
```

skimage/draw/_random_shapes.py
```diff
@@ -402,22 +404,23 @@ def random_shapes(image_shape,
...
             shape_generator = random.choice(SHAPE_CHOICES)
         else:
             shape_generator = SHAPE_GENERATORS[user_shape]
-        shape = (min_size, max_size)
         for _ in range(num_trials):
             # Pick start coordinates.
-            column = random.randint(image_shape[1])
-            row = random.randint(image_shape[0])
+            column = random.randint(max(1, image_shape[1] - min_size))
+            row = random.randint(max(1, image_shape[0] - min_size))
```
jni, line 416: Why do you start at 1 and not 0?

skimage/draw/tests/test_random_shapes.py
```diff
@@ -33,6 +33,7 @@ def test_generates_correct_bounding_boxes_for_rectangles():
     assert (crop >= 0).all() and (crop < 255).all()
 
     # The crop is complete.
+    print(bbox)
```
jni, line 36:
```suggestion
```

jni (approved):
I'm happy with this, the code is much simpler, thank you @rfezzani! I've only made small typo fixes and asked one question about `random_shapes`.

Some of the changes to our reference data are a little worrisome, but I think the changes are correct. Perhaps @stefanv would like to have a closer look here.

skimage/draw/_random_shapes.py
rfezzani: when only the `low` parameter is provided to [`randint`](https://docs.scipy.org/doc/numpy-1.15.1/reference/generated/numpy.random.randint.html) it acts as `high` :upside_down_face:

rfezzani:
Thank you @jni for your review :slightly_smiling_face:

stefanv:
Nice work, @rfezzani!

It's a big ask but, if you have some bandwidth and since this is fresh in your mind, would you mind glancing over https://github.com/scikit-image/scikit-image/pull/931 ?  It's been so long, but still something I think about.

stefanv:
I'm a bit worried about the changes in test results; these algorithms are extremely sensitive, so a person has to check every result very carefully, and often calculate them by hand or with other software to verify.  Any sense why the results changed?

jni:
@stefanv 

> Any sense why the results changed?

These changes ensure that the coordinates corresponding to the vertices of the polygon are included. If you look at [these changes](https://github.com/scikit-image/scikit-image/pull/5029/files#diff-1041f849bdf51f62396c2e4edeb7263210a1a3ea6872c0200d48632278d41930L488-R489), for example, the two added points are actually in the input. I quickly checked the other additions and they made sense to me — they are bang on the polygon lines. So I think these results are actually an improvement.

skimage/draw/_random_shapes.py
jni: Bleugh. I misread. LOL My preference would be to add a 0 as `low` anyway to make things explicit and easier to parse. But I am open to being ignored on this particular point. 😂

rfezzani:
@stefanv, I will try to find some time to dig into #931 :wink:

skimage/_shared/geometry.pyx
```diff
@@ -2,54 +2,96 @@
...
+cnp.import_array()
 
-cdef inline unsigned char point_in_polygon(Py_ssize_t nr_verts, double *xp,
-                                           double *yp, double x,
-                                           double y) nogil:
-    """Test whether point lies inside a polygon.
+
+cdef unsigned char point_in_polygon(np_floats[::1] xp, np_floats[::1] yp,
+                                    np_floats x, np_floats y) nogil:
+    """Test relative point position to a polygone.
```
emmanuelle, line 13:
```suggestion
    """Test relative point position to a polygon.
```

emmanuelle:
Thank you @rfezzani ! This looks good to me, so let's merge. I noted that we don't have a gallery example for `points_in_poly`... (I added an item about it in #3830)

stefanv:
Nice work, @rfezzani !

PR #5040, opened 2020-10-28
Specify whether core dev should merge right after second approving review.

mkcor:
## Description

@grlee77 resubmitting #5039 over here.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/source/core_developer.md
+++ b/doc/source/core_developer.md
@@ -125,6 +125,16 @@ While we collectively "own" any patches (and bugs!) that become part
 of the code base, you are vouching for changes you merge.  Please take
 that responsibility seriously.
 
+In practice, if you are the second core developer reviewing and approving a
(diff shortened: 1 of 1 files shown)


grlee77 (approved):
Thanks, @mkcor!

rfezzani (approved):

rfezzani:
Thank you @mkcor! It is also sometimes good to not directly merge when the PR collects to quick approvals... In this case it is good to give a chance to other reviewers to speak up...
Like this PR for example :smile:...

emmanuelle:
Ha ha @rfezzani. I think this one is not too controversial so I'm gonna merge. If somebody disagrees we can always consider a follow-up PR. In general I don't think we are too trigger-happy about merging PRs, we err on the opposite side of letting PRs open for a long time.

PR #5046, opened 2020-11-02
[WIP] Drop python 3.6 from the build matrix

hmaarrfk:
## Description

This does the technical work to drop python 3.6 from the build matrix.
Follow recommendations from NEP29 proactively: https://numpy.org/neps/nep-0029-deprecation_policy.html

It is probably best to focus the discussion on the mailing list:

Mailing list discussion: https://mail.python.org/archives/list/scikit-image@python.org/thread/X5XBWFD2DYKXC2XYBCY6Z5D7FCHTMKEA/

NumPy mailing list discussion: https://mail.python.org/pipermail/numpy-discussion/2020-November/081127.html

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -73,7 +73,7 @@ Improvements
 - ``measure.label`` has been accelerated for boolean input images, by using
   ``scipy.ndimage``'s implementation for this case (#4945).
 - ``util.apply_parallel`` now works with multichannel data (#4927).
-
+- ``skimage.feature.peak_local_max`` supports now any Minkowski distance.
 
 
 API Changes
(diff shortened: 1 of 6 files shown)


soupault:
+1 if the numpy team falls back to NEP29 policy. Although, not going to lie, to me 3.6 sounds quite recent.

hmaarrfk:
It seems that many of our minimum pins would need updating too because they don't exist as wheels for python 3.7.

This tool can help refresh dependencies for nep29 conformity
pypi.org/project/nep29

emmanuelle:
I think you also need to modify the `.appveyor.yml`. Probably we want to merge this PR after the `0.18.x` branch is made?

hmaarrfk:
image

Numpy just decided to drop Python 3.6. If this PR is ready in time approved and all, I wouldn't "wait" to merge it in.

jni:
I agree with @hmaarrfk, now that NumPy is going to follow NEP-29 strictly also, I think we should just drop it — the drop date was June so we are well past it.

@hmaarrfk Do you need any help getting this to pass? Is this PR dependent on the other build matrix PRs in flight? I don't have a good overview right now.

hmaarrfk:
I would like to help the pthon 3.8 pr get through, then do a straight version bump on all python versions with NEP29 updates.

I was holding off until the numpy decision went through

PR #5052, opened 2020-11-05
update trove classifiers and tests for Python 3.9 + fix pytest config

terencehonles:
## Description

This change adds the trove classifiers for Python 3.8 & 3.9, updates the CI to test Python 3.9, and updates the pytest config to not try running the test setup as a test itself.

## For reviewers

  later.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,7 +8,13 @@ environment:
     - PYTHON: C:\Python37-x64
     - PYTHON: C:\Python38
     - PYTHON: C:\Python38-x64
-    - PYTHON: C:\Python37-x64
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+      PYTHON: C:\Python39
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
+      PYTHON: C:\Python39-x64
+    # build pre release packages on Python 3.8 since it has been out long
+    # enough for wheels to be built for packages that need to be compiled.
+    - PYTHON: C:\Python38-x64
       PIP_FLAGS: --pre
 
 matrix:
@@ -29,7 +35,7 @@ install:
         throw "There are newer queued builds for this pull request, failing early." }
 
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-  - "python -m pip install --retries 3 -U pip"
+  - "python -m pip install -U pip"
 
   # Check that we have the expected version and architecture for Python
   - "python --version"
@@ -37,15 +43,15 @@ install:
   - "pip --version"
 
   # Install the build and runtime dependencies of the project.
-  - pip install %PIP_FLAGS% --retries 3 -r requirements/default.txt
-  - pip install %PIP_FLAGS% --retries 3 -r requirements/build.txt
+  - pip install %PIP_FLAGS% -r requirements/default.txt
+  - pip install %PIP_FLAGS% -r requirements/build.txt
   - python setup.py bdist_wheel bdist_wininst
   - ps: "ls dist"
 
   # Install the generated wheel package to test it.
   - pip install %PIP_FLAGS% --no-index --find-links dist/ scikit-image
   # Install the test dependencies
-  - pip install %PIP_FLAGS% --retries 3 -r requirements/test.txt
+  - pip install %PIP_FLAGS% -r requirements/test.txt
 
 # Not a .NET project, we build scikit-image in the install step instead
 build: false
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   doc:
     docker:
-      - image: circleci/python:3.8.2
+      - image: circleci/python:3.9.0
     steps:
       - checkout
       - run:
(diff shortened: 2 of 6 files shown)


terencehonles:
It appears the azure pipelines were not configured correctly and looking at the logs the Python version specified in the UsePythonVersion task did not match the Python used in the later stages. I adjusted the pipelines to explicitly refer to the `$(stage.pythonLocation)` variable (using a bash variable for brevity) and also updated bash to be run with `set -e` to abort on errors.

It looks like there are still some issues with the Python 3.9 runners and I'll see if I can figure out what the issue is tomorrow.

emmanuelle:
Thanks a lot @terencehonles ; there seem to be indeed a few problems with some of our dependencies (Pywavelets, SimpleITK).

hmaarrfk:
Could you revert the portion 3/.9 changes? This pr looks good otherwise and it l would be a Shane to hold it up.

grlee77:
> Thanks a lot @terencehonles ; there seem to be indeed a few problems with some of our dependencies (Pywavelets, SimpleITK).

We do not currently have Python 3.9 wheels for PyWavelets, but I don't think it should be too much work to get those out of it is holding things up here. I can take a look sometime today/tomorrow

terencehonles:
> > Thanks a lot @terencehonles ; there seem to be indeed a few problems with some of our dependencies (Pywavelets, SimpleITK).
> 
> We do not currently have Python 3.9 wheels for PyWavelets, but I don't think it should be too much work to get those out of it is holding things up here. I can take a look sometime today/tomorrow

Thanks @grlee77 the wheels will build on Linux, but it looks like there's something missing on Windows. If you get to it great, otherwise I was going to look at making sure it has a pyproject.toml as I noticed there were warnings for that from pip. Looking at the error on Azure Pipelines I don't think adding the pyproject.toml will actually resolve the issue though so either building a wheel so it doesn't need to be built here or identifying what needs to be fixed on Windows would be great.

terencehonles:
> Could you revert the portion 3/.9 changes? This pr looks good otherwise and it l would be a Shane to hold it up.

@hmaarrfk I put the non Python 3.9 changes in #5054

hmaarrfk:
Thanks @terencehonles I think you were hitting problems with compiling `imread`. That is kinda standard and why we placed them as "extra", we should be able to disable them with a flag.

terencehonles:
> Thanks @terencehonles I think you were hitting problems with compiling `imread`. That is kinda standard and why we placed them as "extra", we should be able to disable them with a flag.

Would you prefer I disable the optional section rather than building them from source? I believe I identified the couple lines to enable that.

terencehonles:
Nevermind on that question, it's pretty slow to build and it doesn't really make sense to make the build take that long.

grlee77:
It looks like we also still have Appveyor builds (`.appveyor.yml`)? I just added 3.9 support on Appveyor for PyWavelets and I think you have to specify a "Visual Studio 2019" image as done here:

https://github.com/PyWavelets/pywt/blob/e2b05eaf849d516be02a187fd014d42f8df940d5/appveyor.yml#L18-L19

I will look into the Azure failure now and see if it is a similar issue there

grlee77:
Azure passed just by restarting the job. It looks the change here now need to be rebased, though

terencehonles:
> It looks like we also still have Appveyor builds (`.appveyor.yml`)? I just added 3.9 support on Appveyor for PyWavelets and I think you have to specify a "Visual Studio 2019" image as done here:
> 
> https://github.com/PyWavelets/pywt/blob/e2b05eaf849d516be02a187fd014d42f8df940d5/appveyor.yml#L18-L19
> 
> I will look into the Azure failure now and see if it is a similar issue there

It looks like I hadn't pushed my change which update appveyor.yml. I've updated it, but was going to follow up on all/any build errors. Thanks for the reference to the PyWavelets change

terencehonles:
> Azure passed just by restarting the job. It looks the change here now need to be rebased, though

It looks like some of the job runners for Azure Pipelines haven't been refreshed for the new build tools. I'm not sure if there's a flag you can specify to pick the specific version.

It looks like you would need to specify https://github.com/actions/virtual-environments/blob/14b4cec214c5a521821055add04699f48e72bf75/images/linux/Ubuntu1804-README.md which is image version 20201012.1 or greater

emmanuelle (approved):
Thanks @terencehonles looks good to me! We cannot test the Travis changes any more but your changes will be useful to port as we are migrating to Github Actions.

.appveyor.yml
```diff
@@ -29,23 +35,23 @@ install:
...
+  - pip install %PIP_FLAGS% -r requirements/default.txt
+  - pip install %PIP_FLAGS% -r requirements/build.txt
   - python setup.py bdist_wheel bdist_wininst
   - ps: "ls dist"
 
   # Install the generated wheel package to test it.
   - pip install %PIP_FLAGS% --no-index --find-links dist/ scikit-image
   # Install the test dependencies
-  - pip install %PIP_FLAGS% --retries 3 -r requirements/test.txt
+  - pip install %PIP_FLAGS% -r requirements/test.txt
```
jni, line 54: Why are these removed?

terencehonles: The [default now is 5](https://pip.pypa.io/en/stable/reference/pip/#cmdoption-retries) and I don't think the author wanted *less* retries

hmaarrfk:
Cool!

PR #5054, opened 2020-11-05
fix Azure Pipelines, pytest config, and trove classifiers for Python 3.8

terencehonles:
## Description

Azure Pipelines did not appear to be running the correct Python versions because they were running on a Windows image and the UsePythonVersion task did not set the path correctly for running bash scripts. This change uses the variable produced by UsePythonVersion to specify the full path to Python.

This change also updates the trove classifiers to specify Python 3.8 which is already supported and tested, and updates the pytest config to not try running the test setup as a test itself.

## For reviewers

  later.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -47,11 +47,12 @@ api:
 
 release_notes:
 	@echo "Copying release notes"
-	@cat `$(PYTHON) -c \
-		"from pathlib import Path; from skimage import io; \
-		 l = list(str(e) for e in Path('release').glob('release_*.*.*')); \
-		 l.sort(key=io.collection.alphanumeric_key); print(l[-1])"` \
-		| tail -n +4 > release/_release_notes_for_docs.rst
+	@tail -n +4 release/release_$$(\
+		ls release/release_*.*.* \
+			| sed s:release/release_:: \
+			| sort -nrt. -k1,1 -k2,2 \
+			| head -n 1) \
+		> release/_release_notes_for_docs.rst
 
 html: api release_notes
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(DEST)/html
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,6 @@ addons:
     # https://github.com/matplotlib/matplotlib/pull/13412
     - qtbase5-dev
 
-stage: Comprehensive tests
-
 stages:
     # Do the style check and a single test job, don't proceed if it fails
     - name: Initial tests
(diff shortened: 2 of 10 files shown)


hmaarrfk (approved):

terencehonles:
[Small updates](https://github.com/scikit-image/scikit-image/compare/8daefc7529e48db7d2ba173f5d7d3ca837986887..c6116a60a2c25e700c74dbd365d1fbe03eb02820):
- Trying to install pip with get-pip.py does not seem like it's needed now using the right Python
- I also removed `--retries 3` from pip since the default is 5 and the original author was probably trying to make sure retries were tried rather than limiting the number of retries.

terencehonles:
Removing the `-q` flag from pip as that's not really the best practice on a CI where you might need to investigate an issue.

terencehonles:
@hmaarrfk I have an idea on how to fix Travis, and can possibly look at CircleCI, but do you want to re-review this and possibly merge this change in now? I'm iterating with Travis on #5052 to allow this PR to be merged independently if that makes sense.

hmaarrfk:
i have no idea why travis isn't working.... so weird.

doc/Makefile
```diff
@@ -47,11 +47,12 @@ api:
 
 release_notes:
 	@echo "Copying release notes"
-	@cat `$(PYTHON) -c \
-		"from pathlib import Path; from skimage import io; \
-		 l = list(str(e) for e in Path('release').glob('release_*.*.*')); \
-		 l.sort(key=io.collection.alphanumeric_key); print(l[-1])"` \
-		| tail -n +4 > release/_release_notes_for_docs.rst
+	@tail -n +4 release/release_$$(\
```
emmanuelle, line 50: Could you please explain why you made this change? I understand that it's cleaner to use shell commands instead of calling python here but on the other hand folks here are more familiar with Python so the "ugly" version with Python might be easier to maintain. Not a strong feeling here, just trying to decide what is best.

emmanuelle (approved):
I left a small comment related to shell commands, but otherwise the PR looks good to me, thank you very much @terencehonles. Travis's failure is not related to this PR and might be solved by #5055.

doc/Makefile
terencehonles: Yeah, I made this change because the shell expansion of `$(PYTHON)` is getting mangled and was breaking the build. The code is being executed on Windows by a variation of Bash (it seems to be different per Python version), but since Python is coming from Windows it needs to be specified as a full path to the version of Python so it's not picked up by bash. The issue is coming from either Make or Bash, but the Python code is also "not very straight forward" as to what it is doing (maybe this should be not an inline command) and I figured it would not be a big deal to change it to a roughly equally "hard to read" shell command which would keep everything using whatever version of Bash this was running on.

hmaarrfk:
Why has travis wait improved been removed?

terencehonles:
> Why has travis wait improved been removed?

@hmaarrfk it hasn't, it was just moved (and installed a little later) it's basically the same change as @emmanuelle did in #5055

terencehonles:
Just in case you were commenting on https://travis-ci.com/github/scikit-image/scikit-image/jobs/430301019 I didn't realize we were running so close to the cap, but 50 minutes is the max time a job can run and travis-wait-improved will not help

See: https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts

hmaarrfk:
Ok thanks, good from me.

emmanuelle:
About the 50 minutes, probably we must try to reduce the length of these builds... This will be for future work. Thanks a lot @terencehonles , merging!

terencehonles:
> About the 50 minutes, probably we must try to reduce the length of these builds... This will be for future work. Thanks a lot @terencehonles , merging!

No prob, it looks like the example testing should be probably moved to their own jobs/stages and that can probably save a decent amount of time by running those in parallel rather than in series.

hmaarrfk:
This is awesome. Super excited to have this so that we can actually think about a clean transition to 3.9.

PR #5055, opened 2020-11-06
debug travis build

emmanuelle:
Travis OSX builds are failing with mysterious error messages such as "/usr/local/opt/python/bin/python3.7: bad interpreter: No such file or directory", I'm trying to debug this here but would welcome also ideas if it rings a bell for some folks.

### DIFF
--- a/tools/travis/osx_install.sh
+++ b/tools/travis/osx_install.sh
@@ -7,8 +7,8 @@ brew tap homebrew/homebrew-cask
 brew cask install basictex
 
 export PATH="$PATH:/Library/TeX/texbin"
-travis-wait-improved --timeout 30m sudo tlmgr --verify-repo=none update --self
-travis-wait-improved --timeout 30m sudo tlmgr --verify-repo=none install ucs dvipng anyfontsize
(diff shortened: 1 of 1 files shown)


emmanuelle:
ok, removing two calls to `travis-wait-improved` seem to remove the weird Travis Error. @sciunto I feel bad about removing a recent change of yours which did solve some CI problems at the time (timeout) but on the other hand these failures are very annoying.

terencehonles:
@emmanuelle I've also fixed this by changing which Python OS X is using in #5052 (I just pushed an update which should hopefully fix the new Azure Pipeline jobs I added, but also take care of OS X)

You can see just the specific change here https://github.com/scikit-image/scikit-image/pull/5052/commits/683d13efbc15625c5133a45854dc39a1ed487e57

emmanuelle:
@terencehonles ok thanks. Your fix is definitely better than mine, so I should close this PR. I'm not just sure when we can merge #5052 because of the remaining failures on py3.9. , on the other hand #5054 also includes the same commit fixing Travis and seems ready to be merged. Therefore, closing this PR!

PR #5056, opened 2020-11-07
Clarify disk documentation inconsistency regarding 'shape'

### LINKED ISSUE #5038: disk Inconsistency between `shape=None` and `shape=actual_shape`
## Description

The shape parameter in [disk documentation](https://scikit-image.org/docs/dev/api/skimage.draw.html#disk) (and maybe others using the same machinery) seems misleading to me. Based on the following phrasing, I would expect `shape=None` and the `shape=dimensions_of_my_disk` to lead to equal results. (but it doesn't)

```
Image shape which is used to determine the maximum extent of output pixel coordinates. This is useful for disks that exceed the image size. If None, the full extent of the disk is used. Must be at least length 2. Only the first two values are used to determine the extent of the input image.
```

## Way to reproduce
This MWE shows that using `shape=None` the negative indices are consumed by numpy without trouble and therefore make it a circular canvas where the disk can roll over (see figure below)
```python
from skimage.draw import disk

fig, axes = plt.subplots(1, 2)
for ax, shape in zip(axes, [None, (10, 10)]):
    img = np.zeros((10, 10), dtype=np.uint8)
    rr, cc = disk((1, 1), 5, shape=shape)
    img[rr, cc] = 1
    ax.imshow(img)
```
image

## Version information

```
3.8.5 | packaged by conda-forge | (default, Aug 29 2020, 00:43:28) [MSC v.1916 64 bit (AMD64)]
Windows-10-10.0.18362-SP0
scikit-image version: 0.17.2
numpy version: 1.19.1
```

jjerphan:
## Description

Fixes #5038.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/draw/draw.py
+++ b/skimage/draw/draw.py
@@ -205,7 +205,26 @@ def disk(center, radius, *, shape=None):
 
     Examples
     --------
+    >>> import numpy as np
     >>> from skimage.draw import disk
+    >>> shape = (4, 4)
+    >>> img = np.zeros(shape, dtype=np.uint8)
+    >>> rr, cc = disk((0, 0), 2, shape=shape)
+    >>> img[rr, cc] = 1
+    >>> img
+    array([[1, 1, 0, 0],
+           [1, 1, 0, 0],
+           [0, 0, 0, 0],
+           [0, 0, 0, 0]], dtype=uint8)
+    >>> img = np.zeros(shape, dtype=np.uint8)
+    >>> # Negative coordinates in rr and cc perform a wraparound
+    >>> rr, cc = disk((0, 0), 2, shape=None)
+    >>> img[rr, cc] = 1
+    >>> img
+    array([[1, 1, 0, 1],
+           [1, 1, 0, 1],
+           [0, 0, 0, 0],
+           [1, 1, 0, 1]], dtype=uint8)
     >>> img = np.zeros((10, 10), dtype=np.uint8)
     >>> rr, cc = disk((4, 4), 5)
     >>> img[rr, cc] = 1
--- a/skimage/draw/draw.py
+++ b/skimage/draw/draw.py
@@ -190,11 +190,11 @@ def disk(center, radius, *, shape=None):
     radius : double
         Radius of disk.
     shape : tuple, optional
-        Image shape which is used to determine the maximum extent of output
-        pixel coordinates. This is useful for disks that exceed the image
-        size. If None, the full extent of the disk is used.  Must be at least
-        length 2. Only the first two values are used to determine the extent of
-        the input image.
+        Image shape as a tuple of size 2 which is used to determine the maximum
+        extent of output pixel coordinates. This is useful for disks that
+        exceed the image size. If None, the full extent of the disk is used.
+        and the shape might result in negative coordinates and wraparound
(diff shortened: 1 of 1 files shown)


skimage/draw/draw.py
```diff
@@ -190,11 +190,11 @@ def disk(center, radius, *, shape=None):
     radius : double
         Radius of disk.
     shape : tuple, optional
-        Image shape which is used to determine the maximum extent of output
-        pixel coordinates. This is useful for disks that exceed the image
-        size. If None, the full extent of the disk is used.  Must be at least
-        length 2. Only the first two values are used to determine the extent of
-        the input image.
+        Image shape as a tuple of size 2 which is used to determine the maximum
```
alexdesiqueira, line 193: `as a tuple of size 2, used to determine`

skimage/draw/draw.py
```diff
@@ -190,11 +190,11 @@ def disk(center, radius, *, shape=None):
...
+        extent of output pixel coordinates. This is useful for disks that
+        exceed the image size. If None, the full extent of the disk is used.
+        and the shape might result in negative coordinates and wraparound
```
alexdesiqueira, line 196: `The shape might result in`

alexdesiqueira (commented):
Thank you @jjerphan! I have two small suggestions. Besides that, I'm not that comfortable with "_the shape might result in negative coordinates and wraparound_," but let's see what @jni would like to add. Thanks again!

alexdesiqueira:
Also, is this what you would expect to see here @massich? (glad to see you around!)

jjerphan:
@alexdesiqueira: looks like checks aren't passing for weird reasons.

alexdesiqueira:
The error seems related to OOM :thinking: I'll try restarting the CI.

skimage/draw/draw.py
```diff
@@ -190,11 +190,11 @@ def disk(center, radius, *, shape=None):
...
-        Image shape which is used to determine the maximum extent of output
-        pixel coordinates. This is useful for disks that exceed the image
-        size. If None, the full extent of the disk is used.  Must be at least
-        length 2. Only the first two values are used to determine the extent of
-        the input image.
+        Image shape as a tuple of size 2, used to determine the maximum
+        extent of output pixel coordinates. This is useful for disks that
+        exceed the image size. If None, the full extent of the disk is used.
+        The  shape might result in negative coordinates and wraparound
+        behaviour.
```
massich, line 197: I think that the confusing part of the wording was the `full extend of the disk is used` and the clarifying sentence makes sense but only 'cos I've seen where those negative indices are coming from.  I would keep this (or anything similar) but I would add a similar example to #5038 to compare `shape=None` to a tuple that matches shape `shape=(shape[0], shape[1])`

jjerphan: Let me know what you think in the example introduced by 3076a61.

jjerphan: Let me know what you think of the example introduced by 3076a61.

jni (approved):

jni:
Thank you @jjerphan!

jjerphan:
You're welcome!

Issue #4836, opened 2020-07-11
pyiodide compatibility


emmanuelle:
See the issue in pyiodide https://github.com/iodide-project/pyodide/issues/253. It's a very interesting project and it'd be worth giving it a try.

rth:
Thanks for opening this issue! Assuming scikit-image only uses Cython and that openmp can be disabled, I think it should be not too difficult to add scikit-image to pyodide.

As an example the setup for scikit-learn is here https://github.com/iodide-project/pyodide/blob/master/packages/scikit-learn/meta.yaml and it should be somewhat comparable I think.

rth:
Thanks for investigating this issue @chigozienri! Regarding the OpenMP error, the issue is that as far as I understand scikit-image will try optionally build with OpenMP ([here](https://github.com/scikit-image/scikit-image/blob/967886138a28727f954048b8a8ce1c295d857476/setup.py#L66)). In the context of pyodide this wouldn't work, because,
 - it would do a first compilation pass with gcc on the host system in which case OpenMP will work fine
 - then it would take the resulting log with compilations commands and run them with emscripten /emcc in which case those openmp includes will fail.

A temporary workaround could be to add a patch to put,
```py
    class ConditionalOpenMP(build_ext):

        def can_compile_link(self):
             return False
```
in [setup.py](https://github.com/scikit-image/scikit-image/blob/967886138a28727f954048b8a8ce1c295d857476/setup.py#L69) (see how patches are specified in other `meta.yaml` packages).

I'm not sure if there is anything we could do to avoid having to patch in the long term, maybe adding a specific env variable to scikit-image? Unfortunately adding conditions along the lines of,
```py
        def can_compile_link(self):
             if platform.system() == "Emscripten":
                 return False
```
to `setup.py` also won't work as the first pass is with gcc, though  maybe this is something that we should try to improve on the pyodide side.

bio43:
Hi @rth @chigozienri , 
I'm very interested in this feature. Do you have an open PR or is there any way I can help contribute to @rth 's suggested fix.
Thanks!

grlee77:
@rth, I think using an environment to optionally disable OpenMP would be an easy solution on our end. Perhaps modifying the top of `ConditionalOpenMP` as:

```Python
    class ConditionalOpenMP(build_ext):

        def can_compile_link(self):
            if 'SKIMAGE_DISABLE_OPENMP' in os.environ:
                return False
```

[...]

PR #5059, opened 2020-11-11
MAINT Build fix for pyodide

rth:
Adds a small fix to build scikit-image in pyodide, to avoid patching the `setup.py`, as done in https://github.com/iodide-project/pyodide/pull/733. ~~There the challenge is more with building pillow than scikit-image.~~

Contributes toward https://github.com/scikit-image/scikit-image/issues/4836

See [detecting Pyodide at build time](https://pyodide.readthedocs.io/en/latest/faq.html#how-to-detect-that-code-is-run-with-pyodide) for more information.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,6 +67,10 @@ def openmp_build_ext():
 
         def can_compile_link(self):
 
+            if "PYODIDE_PACKAGE_ABI" in os.environ:
+                # pyodide doesn't support OpenMP
+                return False
+
             cc = self.compiler
             fname = 'test.c'
             cwd = os.getcwd()


jni (approved):
Cool, thanks @rth! Happy with this change. If there is some testing we can do to ensure this stays put, we'd be happy to put it in, but at any rate I'm happy to merge this as-is.

rth:
> If there is some testing we can do to ensure this stays put

Thanks @jni ! I can't really think of an easy and reliable way of testing it. I think not having a test for this in `setup.py` is probably fine. Even if this logic accidentally changes in the future, and the fix no longer work, once scikit-image is packaged, we will just patch it temporary in pyodide and submit a fix here.

emmanuelle:
Thank you @rth, merging!

PR #5062, opened 2020-11-12
Optimize ensure_spacing

### LINKED ISSUE #5061: ensure_spacing function may be slow
## Description

The `skimage._shared.coord.ensure_spacing` function introduced in #4760 may be slow when the number of input points is large.

## Way to reproduce
```python
In [1]: import numpy as np

In [2]: from skimage._shared.coord import ensure_spacing

In [3]: for n in range(500, 10001, 500):
    ...:     print(f"n = {n}")
    ...:     x = np.random.rand(n, 2)
    ...:     %timeit ensure_spacing(x, 0.2)
    ...: 
n = 500
27.6 ms ± 857 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
n = 1000
64.7 ms ± 1.93 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
n = 1500
105 ms ± 2.27 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
n = 2000
158 ms ± 1.18 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
n = 2500
213 ms ± 7.32 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 3000
299 ms ± 11.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 3500
378 ms ± 3.05 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 4000
428 ms ± 11.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 4500
554 ms ± 14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 5000
647 ms ± 9.42 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 5500
777 ms ± 7.05 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 6000
870 ms ± 13.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 6500
1.03 s ± 9.56 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 7000
1.17 s ± 12.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 7500
1.33 s ± 28 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 8000
1.45 s ± 18.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 8500
1.63 s ± 36.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 9000
1.87 s ± 50.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 9500
2.07 s ± 23.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
n = 10000
2.29 s ± 51.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

rfezzani:
## Description

Fixes #5061. This PR proposes to process input points by batch. Using this strategy timing is considerably reduced:

Figure_3

## For reviewers

  later.


jni (commented):
@rfezzani thanks for this, sorry for the long silence!

If I'm reading this correctly, this means that the cKDTree construction is quadratic, and you can get it cheaper by removing nearby points each time, so you never build a tree with all the points in it?

I do wonder how this affects the worst case, though: evenly spaced points just beyond min_distance away, so you don't actually remove anything in each batch... 

One idea would be to instead double the batches starting from 10 up to the total size *in the original space of points*. So you build with the top 10, 20, 40, ... n peaks, removing points each time. That way, in the worst case, right before doing all the points you only do half the points, which is 1/4 the time, so you are protected from a dramatic blow-up of time in the worst case.

Thoughts?

rfezzani:
No worry @jni, we all are so busy these days :wink:.

You are absolutely right, it's an excellent suggestion! the strategy that you propose also has the advantage of removing the `batch_size` parameter :smile:!

skimage/_shared/coord.py
```diff
@@ -24,28 +24,72 @@ def ensure_spacing(coord, spacing=1, p_norm=np.inf):
...
+    spacing : float
+        the maximum allowed spacing between the points.
+    p_norm : float
+        Which Minkowski p-norm to use. Should be in the range [1, inf].
+        A finite large p may cause a ValueError if overflow can occur.
+        ``inf`` corresponds to the Chebyshev distance and 2 to the
+        Euclidean distance.
+    min_split_size : int
+        Minimum split size used to process ``coord`` by batch to save
+        memory. If None, the memory saving strategy is not applyed.
```
jni, line 68:
```suggestion
        memory. If None, the memory saving strategy is not applied.
```

jni (approved):
@rfezzani in the nick of time for 0.18.0, maybe! 😅 Do you want to check benchmarks with this new strategy both in a normal case and in the previous worst-case of evenly-spaced points? But I'm happy as is, thank you! 🎉

rfezzani:
Hi @jni, thank you for your review! I measured the performances using

```python
In [1]: from skimage._shared.coord import ensure_spacing                                                 

In [2]: for f, x in [("random", np.random.rand(10000, 2)), 
   ...:              ("worst", np.arange(20000).reshape((-1, 2)))]: 
   ...:     print(f) 
   ...:     for s in [None, 100, 500, 1000]: 
   ...:         print(f"s = {s}") 
   ...:         for n in range(500, 10001, 500): 
   ...:             print(f"n = {n}") 
   ...:             %timeit ensure_spacing(x[:n], 0.2, min_split_size=s) 
   ...:         print(4*'\n')
```

and obtained

Figure_1

Unfortunately, the new approach is twice as slow as the previous approach on the random points, but it is ~3x faster on worst case `min_split_size = 500` which should be the default parameter I think.

jni:
Thanks @rfezzani! I can't say I know which approach is better to have, but we should have one of them and the newer one seems like an ok choice for now — we can tweak later. @scikit-image/core anyone else available to review+merge this?

alexdesiqueira:
Thank you @rfezzani! I'll merge, based on @jni arguments; I like it so far.

rfezzani:
Thank you @alexdesiqueira !

jni:

Issue #5066, opened 2020-11-15
OSX build in Travis is taking too long


emmanuelle:
See for example https://travis-ci.com/github/scikit-image/scikit-image/jobs/432959421

In particular executing the `osx_install.sh` takes about 15 minutes to execute.-

emmanuelle:
Closing since we switched to Github Actions.

PR #5067, opened 2020-11-15
reduce OSX build time so that Travis is happy

emmanuelle:
Solves #5066 (or tries to).

### DIFF
--- a/tools/travis/osx_install.sh
+++ b/tools/travis/osx_install.sh
@@ -3,8 +3,6 @@ set -ex
 
 brew update
 brew install ccache
-brew tap homebrew/homebrew-cask
-brew cask install basictex
 
 # Set up virtualenv on OSX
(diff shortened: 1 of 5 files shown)


tools/travis/osx_install.sh
```diff
@@ -3,25 +3,12 @@ set -ex
 
 brew update
 brew install ccache
-brew tap homebrew/homebrew-cask
```
emmanuelle, line 6: probably we don't need the two lines above either?

emmanuelle:
This is now ready for review. I'm glad Travis is green again. The biggest change of this PR is to remove some unnecessary installations is `osx_install.sh`, like latex etc. (I guess this was used when the doc was built on osx but this is not the case now).

rfezzani (approved):
:tada: Thank you @emmanuelle!

jni (approved):
Yesss! 😄 🎉 🥳 Thank you so much @emmanuelle!

Issue #5036, opened 2020-10-27
Discussion: deprecating (most of) skimage.io in favour of imageio


jni:
## Description

This is meant to supersede the discussion in #1605, which contains a lot of outdated information.

In 2020, here are the pros for replacing skimage.io with imageio:
- distributing wheels is easy now
- imageio has feature parity with `skimage.io`. In fact, I would be surprised if 99% of our `io.imread` calls didn't go straight through to imageio — at least if we exclude tifffile.
- [this post](https://blog.danallan.com/posts/2020-03-07-reader-protocol/) by @danielballan, for those that haven't seen it, gives a rough outline for implementing a single protocol for lazy image reading in Python. imageio is a natural place to implement such work, and `skimage.io` is at best redundant, at worst, another place for things to fall out of sync.

Things that I think need to change in imageio for us to completely drop `skimage.io`:
- imageio.imread does return a fancy array wrapper rather than a straight numpy array, which breaks the scikit-image API. However, this is somewhat necessary for imageio to provide metadata. Not sure where to meet on this one.
- imageio.imread only gets a single plane out of a tiff file, preferring to leave volume loading to `imageio.volread`. imho we should aim to be dimension-agnostic. @almarklein is this something you foresee could change? See e.g. #4138.

In my opinion, the existence of both skimage.io and imageio is redundant, and it makes sense to separate the two functionalities (IO vs processing/analysis) into two separate packages.

@emmanuelle argues for keeping imread. I'm ok with that but I also would contemplate removing it for 1.0.

sciunto:
+1 with @emmanuelle

rfezzani:
Thank you @jni for opening this, you can't imagine how enthusiastic I am about dropping `skimage.io`! :smiley:

I already shared my feelings about this, and I am convinced that `skimage` can gain maturity and stability if it focuses on it's core functionality: image processing!

There is plenty of excellent python packages dedicated to IO and viewing. We must advertise for them instead of proposing yet another implementation with all the induced maintenance...

@sciunto, the `skimage` deprecation procedure takes 2 releases before completely removing the deprecated part from the package. I think that this is enough time to teach beginners how to properly read/write images without our `skimage.io` module :wink:

almarklein:
> this post by @danielballan, for those that haven't seen it, gives a rough outline for implementing a single protocol [...]

[...]

PR #5074, opened 2020-11-22
github actions for tests

emmanuelle:
Since we don't have Travis any more, this is a WIP PR to have a configuration file to test scikit-image with different environments. Later on I'll open another PR in order to build wheels (which will require a different configuration file because of different numpy versions etc.).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,7 @@ jobs:
             platform_id: macosx_x86_64
         
     if: matrix.os == 'macos-latest'
-        env:
+        env: |
             CC: /usr/bin/clang
             CXX: /usr/bin/clang++
             CPPFLAGS: "$CPPFLAGS -Xpreprocessor -fopenmp"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,16 +1,7 @@
 # Workflow to build and test wheels
 name: Run tests
 
-on:
-  push:
-    branches:
-      - master
-      - gh-ci
-  pull_request:
-    branches:
-      - master
-      - gh-ci
-
+on: [push, pull request]
 
 jobs:
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,14 +19,32 @@ jobs:
         python-version: [3.6, 3.7, 3.8]
         PIP_FLAGS: [""]
         MINIMUM_REQUIREMENTS: [0]
+        QT: ["PyQt5"]
+        WITH_PYSIDE: [0]
+        BUILD_DOCS: [1]
+        PYTHONOPTIMIZE: [0]
+        TEST_EXAMPLES: [0]
+        OPTIONAL_DEPS: [0]
         include:
           - platform_id: manylinux_x86_64
             python-version: 3.7
             MINIMUM_REQUIREMENTS: 1
-            PIP_FLAGS: ""
+          - platform_id: manylinux_x86_64
+            python-version: 3.7
+            MINIMUM_REQUIREMENTS: 1
+            OPTIONAL_DEPS: 1
           - platform_id: manylinux_x86_64
             python-version: 3.8
             PIP_FLAGS: "--pre"
+          - platform_id: manylinux_x86_64
+            python-version: 3.7
+            BUILD_DOCS: 1
+            OPTIONAL_DEPS: 1
+          - platform_id: manylinux_x86_64
+            python-version: 3.7
+            PYTHONOPTIMIZE: 2
+            BUILD_DOCS: 0
+
         
 
         
@@ -41,7 +59,8 @@ jobs:
 
       - name: Build package
         env: 
-            VERSION: ${{ matrix.python-version}} 
+            VERSION: ${{ matrix.python-version }} 
+            PYTHONOPTIMIZE: ${{ matrix.PYTHONOPTIMIZE }} 
         run: |
             set -ex
             source tools/github/before_install.sh
@@ -49,13 +68,27 @@ jobs:
             pip install -vv -e .;
 
       - name: Run tests
+        env: 
+            PIP_FLAGS: ${{ matrix.PIP_FLAGS }} 
+            MINIMUM_REQUIREMENTS: ${{ matrix.MINIMUM_REQUIREMENTS }} 
+            QT: ${{ matrix.QT }} 
+            WITH_PYSIDE: ${{ matrix.WITH_PYSIDE }} 
+            BUILD_DOCS: ${{ matrix.BUILD_DOCS }}
+            TEST_EXAMPLES: ${{ matrix.TEST_EXAMPLES }}
+            OPTIONAL_DEPS: ${{ matrix.OPTIONAL_DEPS }}
         run: |
             pip install $PIP_FLAGS -r requirements/test.txt
             export MPL_DIR=`python -c 'import matplotlib; print(matplotlib.get_configdir())'`
             mkdir -p ${MPL_DIR}
             touch ${MPL_DIR}/matplotlibrc
-            pip install -r ./requirements/optional.txt
-            pip install -r ./requirements/extras.txt
+            if [[ "${OPTIONAL_DEPS}" == "1" ]]; then
+                pip install -r ./requirements/optional.txt
+                if [[ "${EXTRA_DEPS}" != "0" ]]; then
+                # Extra deps need compilation, and it may not always be possible to
+                # compile them easily on all platforms
+                    pip install -r ./requirements/extras.txt
+                fi
+            fi
             source tools/github/script.sh
 
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,9 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: [3.6, 3.7, 3.8]
-        bitness: 64
         include:
-          # Run 32 and 64 bit version in parallel for Linux and Windows
           - platform_id: manylinux_x86_64
             python-version: 3.7
             MINIMUM_REQUIREMENTS: 1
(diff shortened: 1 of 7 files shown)


emmanuelle:
Let's see if Github Actions notices the new workflow. For now it's not doing anything.

emmanuelle:
@scikit-image/core  I could use some help here for the OSX configuration. The build fails probably because some os the compilation options set in the "OSX step" has not the right value. I could either go back using the `multibuild` tool which we used with Travis, or read the code of `multibuild` to select the right values (the one I put for now where taken from scikit-learn and https://github.com/scikit-learn/scikit-learn/blob/master/build_tools/github/build_wheels.sh). My idea was to simplify the CI config as much as possible, but maybe we are better off using `multibuild`. Any advice here will be welcome!

jni:
I think [this failure](https://github.com/scikit-image/scikit-image/pull/5074/checks?check_run_id=1455724123#step:6:614) is a pooch error, but not sure...? Let's see on this current build...

jni:
I don't understand why [this error](https://github.com/scikit-image/scikit-image/pull/5074/checks?check_run_id=1456148207#step:5:344) is happening:

```pytb
E     File "/opt/hostedtoolcache/Python/3.7.9/x64/lib/python3.7/site-packages/dask/__init__.py", line 17
E       import dask.async
E                       ^
E   SyntaxError: invalid syntax
```

This is a minreq build and the installed dask version is 0.15, which includes https://github.com/dask/dask/pull/2318 (see also the [release notes for 0.15](https://docs.dask.org/en/latest/changelog.html#id329)).

At any rate, per NEP-29 we should depend on Dask 0.20 (released Nov 9 2018), so I will try bumping that.

Other issues were caused by the pytest minimum version actually being incorrect given the APIs we use.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,175 @@
+# Workflow to build and test wheels
+name: Run tests
```
terencehonles, line 2:
```suggestion
name: Test
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,175 @@
...
+
+on: push
+
+jobs:
+
+  test_skimage_linux:
+    name: Run tests linux-cp${{ matrix.python-version }}-${{ matrix.OPTIONS_NAME }}
```
terencehonles, line 9:
```suggestion
    name: linux-cp${{ matrix.python-version }}-${{ matrix.OPTIONS_NAME }}
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,175 @@
...
+                # compile them easily on all platforms
+                    pip install -r ./requirements/extras.txt
+                fi
+            fi
+            source tools/github/script.sh
+
+
+
+  test_skimage_macos:
+    name: Run tests macos-cp${{ matrix.python-version }}
```
terencehonles, line 110:
```suggestion
    name: macos-cp${{ matrix.python-version }}
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,175 @@
...
+        python-version: [3.6, 3.7, 3.8]
+        PIP_FLAGS: [""]
+        MINIMUM_REQUIREMENTS: [0]
+        QT: ["PyQt5"]
+        WITH_PYSIDE: [0]
+        BUILD_DOCS: [1]
+        PYTHONOPTIMIZE: [0]
+        TEST_EXAMPLES: [0]
+        OPTIONAL_DEPS: [0]
+        OPTIONS_NAME: [""]
```
terencehonles, line 28: This ends up being a little confusing in the output:

```
Run tests / Run tests linux-cp3.6- (push)
                                  ^
```

I'm not sure where it's used and why the default is blank (perhaps "default" might be better?)

terencehonles (commented):
I would probably make these shorter and less redundant because they will be joined by " / "

Screenshot from 2020-11-25 16-24-44

jni:
Can anyone figure out why pytest 3.6.4 is [getting installed](https://github.com/scikit-image/scikit-image/pull/5074/checks?check_run_id=1456417550#step:5:209) in the minimum requirements??? (I just checked and saw that pytest-cov 2.6 [does *not* pin Pytest](https://github.com/pytest-dev/pytest-cov/blob/v2.6.0/setup.py#L123), so it's not that...)

.github/workflows/tests.yml
jni: @terencehonles thanks for noticing! I don't know about this and would rather not mess with it until after this PR is merged. 😬 As it is painful enough to iterate here...

jni:
Ah, [found it](https://github.com/scikit-image/scikit-image/pull/5074/checks?check_run_id=1456417550#step:5:134), it's [astropy 2.0.12](https://github.com/astropy/astropy/blob/v2.0.12/pip-requirements). Changing to 2.0.10 should fix 🤞 . Hold please!

requirements/optional.txt
```diff
@@ -1,8 +1,8 @@
 SimpleITK
-astropy>=1.2.0
+astropy>=2.0.10
```
hmaarrfk, line 2:
```suggestion
astropy>=3.1.2
```

```
22:32 $ nep29 astropy
| version |    date    |
|---------|------------|
|   4.2   | 2020-11-25 |
|   4.1   | 2020-10-21 |
|   4.0   | 2019-12-17 |
|   3.2   | 2019-06-10 |
|   3.1   | 2018-12-06 |
```

requirements/optional.txt
```diff
@@ -1,8 +1,8 @@
...
 qtpy
 pyamg
-dask[array]>=0.15.0,!=2.17.0
+dask[array]>=0.20.0,!=2.17.0
```
hmaarrfk, line 5:
```suggestion
dask[array]>=1.0.0,!=2.17.0
```

```
22:33 $ nep29 dask
| version |    date    |
|---------|------------|
|  2.30.0 | 2020-10-06 |
|  2.29.0 | 2020-10-02 |
|  2.28.0 | 2020-09-26 |
|  2.27.0 | 2020-09-19 |
|  2.26.0 | 2020-09-11 |
|  2.25.0 | 2020-08-28 |
|  2.24.0 | 2020-08-22 |
|  2.23.0 | 2020-08-14 |
|  2.22.0 | 2020-07-31 |
|  2.21.0 | 2020-07-17 |
|  2.20.0 | 2020-07-03 |
|  2.19.0 | 2020-06-19 |
|  2.18.0 | 2020-06-06 |
|  2.17.0 | 2020-05-27 |
|  2.16.0 | 2020-05-08 |
|  2.15.0 | 2020-04-25 |
|  2.14.0 | 2020-04-03 |
|  2.13.0 | 2020-03-25 |
|  2.12.0 | 2020-03-06 |
|  2.11.0 | 2020-02-19 |
|  2.10.0 | 2020-01-28 |
|  2.9.0  | 2019-12-06 |
|  2.8.0  | 2019-11-14 |
|  2.7.0  | 2019-11-08 |
|  2.6.0  | 2019-10-16 |
|  2.5.0  | 2019-09-27 |
|  2.4.0  | 2019-09-13 |
|  2.3.0  | 2019-08-16 |
|  2.2.0  | 2019-07-31 |
|  2.1.0  | 2019-07-08 |
|  2.0.0  | 2019-06-25 |
|  1.2.0  | 2019-04-12 |
|  1.1.0  | 2019-01-18 |
|  1.0.0  | 2018-11-28 |
```

requirements/test.txt
```diff
@@ -1,9 +1,7 @@
-# pytest 3.7.3 is broken on macOS, see:
-# https://github.com/pytest-dev/pytest/issues/3888
-pytest!=3.7.3
-pytest-cov
+pytest>=5.2.0
+pytest-cov>=2.6.0
```
hmaarrfk, line 2:
```suggestion
pytest-cov>=2.7.0
```

```
22:34 $ nep29 pytest-cov
| version |    date    |
|---------|------------|
|  2.10.0 | 2020-06-12 |
|  2.9.0  | 2020-05-22 |
|  2.8.0  | 2019-10-04 |
|  2.7.0  | 2019-05-03 |
```

emmanuelle:
Note that I had to bump the astropy min requirements because otherwise their wheel would not build (don't know exactly why but some version problem with Cython I believe). Since this is an optional requirement and for a part of the code which we will remove soon I think (see #5036 for the discussion on io), I hope this is ok.

.github/workflows/tests.yml
emmanuelle: @terencehonles it is used to have a different name for each job. "default" is a good idea, thanks!

jni (approved):

jni:
@emmanuelle well we bumped it even further so 🤷 =)

Anyway imho this should be merged! Do you think it's ready @emmanuelle?

emmanuelle:
@jni I added python 3.9 and also configured the macos options to be the same one we used to have on Travis. 

One thing which I have not configured yet is the `INSTALL_FROM_SDIST`  to force using pooch, I think we need one build with it.

emmanuelle:
ok all builds are passing :-). @scikit-image/core please review! This is a quite a big PR since all of our testing infrastructure is transferred from Travis to Github Actions.

jni:
Awesome!

tools/github/before_install.sh
```diff
@@ -0,0 +1,77 @@
...
+export PIP_DEFAULT_TIMEOUT=60
+
+# This URL is for any extra wheels that are not available on pypi.  As of 14
+# Jan 2017, the major packages such as numpy and matplotlib are up for all
+# platforms.  The URL points to a Rackspace CDN belonging to the scikit-learn
+# team.  Please contact Olivier Grisel or Matthew Brett if you need
+# permissions for this folder.
+EXTRA_WHEELS="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
+WHEELHOUSE="--find-links=$EXTRA_WHEELS"
+
```
grlee77, line 13: We don't have this Rackspace storage anymore, but it doesn't look like we really need it for this script anyway since these two environment variables don't appear to be used below?

I had setup Travis and Anaconda to upload wheels to anaconda over at scikit-image-wheels recently. To use those we would configure `EXTRA_WHEELS="https://pypi.anaconda.org/multibuild-wheels-staging/label/main/simple/"`.

grlee77: My understanding is that this extra wheel URL would previously have been used to find prerelease wheels for testing? It looks like now prereleases from PyPI are used instead via `pip --pre`.

grlee77 (approved):
This looks great. There are still a couple of Python 3.6 cases that I thought we had planned to drop? I also noticed a couple of outdated environment variables in the `before_install.sh` script.

I am fine with merging as-is and trimming those in a follow up or even after the 0.18 release, though.

emmanuelle:
ok, just pushed a commit removing the unused environment variables. 

Work left for another PR: remove 3.6 from the matrix, and remove travis configuration files

grlee77:
This test failure may be the same one resolved by cd1c34c in #5080?

emmanuelle:
Maybe. It's a Heisenbug which appears from time to time. I just restarted the build, if it does not work I'll cherry-pick the commit from #5080.

grlee77:
Looks good now, so in it goes. Thanks!

PR #5079, opened 2020-11-25
Some minor tweaks to CI

jni:
## Description

These are just minor follow-up tweaks on top of #5052:
- Circle-CI is just for doc builds. I think we should keep it on 3.8 until wheels for 3.9 are available for all our dependencies.
- ~~I think it's better to keep the `--retries` on pip installs on Azure, though happy to be overruled there! (?)~~ **Update:** undid this, see discussion below.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,7 +35,7 @@ install:
         throw "There are newer queued builds for this pull request, failing early." }
 
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-  - "python -m pip install --retries 3 -U pip"
+  - "python -m pip install -U pip"
 
   # Check that we have the expected version and architecture for Python
   - "python --version"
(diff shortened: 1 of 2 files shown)


terencehonles:
I followed up on your comment here https://github.com/scikit-image/scikit-image/pull/5052#discussion_r530510930, but the default retries is 5 and it's probably better to leave this the default unless there are problems with retrying that much and it really should be 3.

hmaarrfk:
Do you want to rework the python 3.6 builds on Travis?

I thought we were going to drop 3.6. I would keep 1 token build for pypy support (maybe).

jni:
@terencehonles sorry for jumping the gun! Is the 5 retries default new or did we just bungle that from the beginning??? 😂

@hmaarrfk Travis is getting jettisoned altogether... Do you mean to move some of the builds somewhere? @emmanuelle is doing that in #5074...

terencehonles:
> @terencehonles sorry for jumping the gun! Is the 5 retries default new or did we just bungle that from the beginning??? 

Not sure on that, but it looks like it switched from 3 to 5 about 6 years ago :sweat_smile: 

https://github.com/pypa/pip/commit/15b2258769987870b02c5c1feea9f131b9f13b66

jni:
Yikes! On our end it dates from #3072 but there wasn't any rationale or discussion. I'll revert! Thank you @terencehonles!

hmaarrfk:
maybe it is time to retire appveyor too?

hmaarrfk:
as for retry. The first pip that is installed was likely not up to date. I must have erroneously added the argument even to the updated pips,

alexdesiqueira (approved):

alexdesiqueira:
Thank you @jni!

Issue #5053, opened 2020-11-05
dropping 3.6 and a few other things


jarrodmillman:
Now that it looks like NumPy will officially drop Python 3.6 for 1.20, I think it would be nice if other important libraries get on the same page.  Pandas and Matplotlib have already dropped 3.6.  I am thinking about dropping 3.6 for the next NetworkX release.  It would be nice if scikit-image also did so.

So I took a look to see if I could tell whether you had decided to drop 3.6 yet and  I noticed that

1. you don't test 3.8 on travis
2. you don't list support for 3.8 in setup.py (see  #5052)
3. your badges on your README say that the forum resource not found and codecov is 0%.
4. it looks like you have not yet dropped support for 3.6. (see #5046)

I assume the travis issue is b/c you are planning on dropping it.

Would you consider dropping 3.6 now?  Having numpy, pandas, matplotlib, networkx, and scikit-image on the same page would be nice and maybe that would motivate scipy and scikit-learn to do the same.

hmaarrfk:
Many of the core devs were active on the discussion to drop Numpy 1.20 from 3.6. 

I think there are many things we need to do to release a wheel for 0.18 and I don't think we should slow it down for any kind of hiccups.

The omission of python 3.8 was an oversight, and we are working on fixing it in 
https://github.com/scikit-image/scikit-image/pull/5054

Once we have the infrastructure for 3.8 setup, my goal is the following:
1. Bump all builds configuration from 3.6 -> 3.7, 3.7 -> 3.8, 3.8->3.9
2. Update all the dependencies to match nep29: https://pypi.org/project/nep29/  -- Many of our minimum dependencies don't have wheels for python 3.7, this makes it virtually impossible for us to build a test system.
2. Maybe keep a token 3.6 build for pypy support -- I would rather keep a pypy build for pypy support.
3. Update the wheel matrix to drop wheels for 3.6.

CC: @terencehonles i know you've been working on adding 3.9 support to our test matrix, and I think dropping 3.6 makes it much easier to think about which configurations you want to add for 3.9.

hmaarrfk:
I'm also in support of releasing a 0.18 Python 3.6 wheel since it just seems like too much infrastructure change at the last minute.

stefanv:
I would strongly consider dropping Python 3.6 to be in sync with the rest of the community and with the NumPy NEP.

terencehonles:
[...]

PR #5080, opened 2020-11-25
Add cibuildwheel action from mpl

jni:
## Description

@emmanuelle here's a start! 😊 Don't forget to delete the branch after merging!

@hmaarrfk if you have any comments on this they would sure be appreciated!!! 🙏 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -32,7 +32,24 @@ jobs:
         run: |
           python -m pip install cibuildwheel==1.6.3
 
-      - name: Build wheels for CPython 3.9
+      - name: Build wheels for CPython 3.9 and Mac OS
+        if: matrix.os == "macos-latest"
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: "cp39-*"
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
+          CIBW_MANYLINUX_I686_IMAGE: manylinux1
+          CIBW_BEFORE_BUILD: pip install certifi numpy==1.19.3
+          CC: /usr/bin/clang
+          CXX: /usr/bin/clang++
+          CPPFLAGS: "-Xpreprocessor -fopenmp"
+          CFLAGS: "-Wno-implicit-function-declaration -I/usr/local/opt/libomp/include"
+          CXXFLAGS: "-I/usr/local/opt/libomp/include"
+          LDFLAGS: "-Wl,-rpath,/usr/local/opt/libomp/lib -L/usr/local/opt/libomp/lib -lomp"
+
+      - name: Build wheels for CPython 3.9 (Linux and Windows)
+        if: matrix.os != "macos-latest"
         run: |
           python -m cibuildwheel --output-dir dist
         env:
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -1,0 +1,89 @@
+name: Build CI wheels
+
+on:
+  push:
+    branches:
+      - master
+      - v[0-9]+.[0-9]+.x
+    tags:
+      - v*
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, windows-latest, macos-latest]
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.7'
+
+      - name: Install cibuildwheel
+        run: |
+          python -m pip install cibuildwheel==1.6.3
+
+      - name: Copy setup.cfg to configure wheel
+        run: |
+          cp setup.cfg.template setup.cfg
+
+      - name: Build wheels for CPython 3.9
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: "cp39-*"
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
+          CIBW_MANYLINUX_I686_IMAGE: manylinux1
+          CIBW_BEFORE_BUILD: pip install certifi numpy==1.19.3
+          MPL_DISABLE_FH4: "yes"
+
+      - name: Build wheels for CPython
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: "cp3?-*"
+          CIBW_SKIP: "cp35-* cp36-* cp39-*"
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
+          CIBW_MANYLINUX_I686_IMAGE: manylinux1
+          CIBW_BEFORE_BUILD: pip install certifi numpy==1.16
+          MPL_DISABLE_FH4: "yes"
+
+      - name: Build wheels for CPython 3.6
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: "cp36-*"
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
+          CIBW_MANYLINUX_I686_IMAGE: manylinux1
+          CIBW_BEFORE_BUILD: pip install certifi numpy==1.16
+          MPL_DISABLE_FH4: "yes"
+        if: >
+          startsWith(github.ref, 'refs/heads/v3.3') ||
+          startsWith(github.ref, 'refs/tags/v3.3')
+
+      - name: Build wheels for PyPy
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: "pp3?-*"
+          CIBW_BEFORE_BUILD: pip install certifi numpy==1.16
+        if: >
+          runner.os != 'Windows' && (
+          startsWith(github.ref, 'refs/heads/v3.3') ||
+          startsWith(github.ref, 'refs/tags/v3.3') )
+
+      - name: Validate that LICENSE files are included in wheels
+        run: |
+          python ./ci/check_wheel_licenses.py
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: wheels
+          path: ./dist/*.whl
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -59,7 +59,7 @@ jobs:
           CIBW_MANYLINUX_I686_IMAGE: manylinux1
           CIBW_BEFORE_BUILD: pip install certifi numpy==1.19.3
           CIBW_TEST_REQUIRES: pytest pooch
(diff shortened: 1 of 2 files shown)


emmanuelle:
We probably want to run tests using the wheels after creating them.

hmaarrfk:
I know it is tempting to merge the two repos, but there is a real concern in debugging the wheel builder, which could give you version 0.18.5 pretty quickly once it bugs out on releasing 0.18.0.

Is that something we are OK with?

I mean, I would be happy if we:

1. Forked `towncrier` because it hasn't been released in a while: https://github.com/twisted/towncrier/issues/269
2. Start using towncrier.
3. Start releasing more often.

But if we aren't going to release more often, maybe stick with the wheel builder repo?

emmanuelle:
@hmaarrfk could you please elaborate a bit more? We started discussing this in #5045, but I don't understand why we can't have a config file for testing the package and another one for building the wheels, which can have different compilers options, versions of numpy etc. Please tell me what I'm missing :-).

emmanuelle:
@jni the wheel building part is now working (I had to separate Mac for some specific compiler options), I've started adding CIBW env variables in order to test the wheels.

jni:
@emmanuelle, I think @hmaarrfk's concern is that if you tag the repo, and it builds but fails, you then have to tag again to try again. But that happened to us on the other repo anyway, and I think that concern can be addressed quite effectively with release candidates.

Additionally, @hmaarrfk I know I was originally skeptical of your frequent releases proposal, but having experienced a year of doing it with napari, I'm a convert! So I absolutely want to use this infrastructure to start doing more frequent releases.

jni:
Additionally, @hmaarrfk, I would say that I'm pretty ok with 0.18.5 as a concept. LOL

jni:
This test seems a bit too stringent:

https://github.com/scikit-image/scikit-image/pull/5080/checks?check_run_id=1456068555#step:6:910

```pytb
    @test_parallel(warnings_matching=['Possible precision loss'])
    def check():
        expected = self.refs_3d[filter]
        result = getattr(rank, filter)(self.volume, self.selem_3d)
>       assert_array_equal(expected, result)
E       AssertionError: 
E       Arrays are not equal
E       
E       Mismatched elements: 13 / 1000 (1.3%)
E       Max absolute difference: 4.4408921e-16
E       Max relative difference: 1.7611122e-16
E        x: array([[[2.      , 2.321928, 2.321928, 2.321928, 2.321928, 2.321928,
E                2.321928, 2.321928, 2.321928, 2.      ],
E               [2.321928, 2.584963, 2.584963, 2.584963, 2.584963, 2.584963,...
E        y: array([[[2.      , 2.321928, 2.321928, 2.321928, 2.321928, 2.321928,
E                2.321928, 2.321928, 2.321928, 2.      ],
E               [2.321928, 2.584963, 2.584963, 2.584963, 2.584963, 2.584963,...
```

😂 I'll update to almost_equal...

jni:
Ok we are in business people!!! 

Screen Shot 2020-11-26 at 12 39 55 pm

Screen Shot 2020-11-26 at 12 36 40 pm

Anyone willing to push the green button? =D

There are various questions, such as whether we want to build manylinux1 vs manylinux2010 wheels, but these should be better addressed in follow-up PRs, I think!

.github/workflows/cibuildwheel.yml
```diff
@@ -0,0 +1,113 @@
...
+          python -m pip install cibuildwheel==1.6.3
+
+      - name: Build wheels for CPython 3.9 and Mac OS
+        if: matrix.os == 'macos-latest'
+        run: |
+          brew install libomp
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: "cp39-*"
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
```
hmaarrfk, line 42: what is this? Where does it get used?

.github/workflows/cibuildwheel.yml
```diff
@@ -0,0 +1,113 @@
...
+      - name: Build wheels for CPython (Linux and Windows)
+        if: matrix.os != 'macos-latest'
+        run: |
+          python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BUILD: "cp3?-*"
+          CIBW_SKIP: "cp35-* cp36-* cp39-*"
+          CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
+          CIBW_MANYLINUX_I686_IMAGE: manylinux1
+          CIBW_BEFORE_BUILD: pip install certifi numpy==1.16
```
hmaarrfk, line 92: Why do we install numpy first? We have things pinned in pyproject.toml ;) don't we?

.github/workflows/cibuildwheel.yml
```diff
@@ -0,0 +1,113 @@
...
+          CIBW_TEST_REQUIRES: pytest pooch pytest-localserver pytest-faulthandler
+          CIBW_TEST_COMMAND: pytest --pyargs skimage
+
+
+
+      - name: Build wheels for CPython 3.6
```
hmaarrfk, line 98: I thought we were dropping this?

.github/workflows/cibuildwheel.yml
jni: manylinux1 is the type of Linux wheel that gets produced (see my comment above). I guess [cibuildwheel](https://cibuildwheel.readthedocs.org) uses this env variable to determine how to build the wheels.

.github/workflows/cibuildwheel.yml
```diff
@@ -0,0 +1,113 @@
...
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v2
+        name: Install Python
+        with:
+          python-version: '3.7'
+
+      - name: Install cibuildwheel
+        run: |
+          python -m pip install cibuildwheel==1.6.3
```
hmaarrfk, line 33: This is amazing!! Why are we pinned at 1.6.3 though?

.github/workflows/cibuildwheel.yml
jni: Yes but @emmanuelle wasn't quite committed to that, partly because of your earlier skepticism in #5053. I would be happy to drop this. I'd still like to get @emmanuelle's blessing though, so perhaps that can happen in a followup PR?

.github/workflows/cibuildwheel.yml
jni: That's because matplotlib pinned it 😂 

https://github.com/matplotlib/matplotlib/blob/2303c2b4afa7ecd1c3c576377c77dc53e0c86273/.github/workflows/cibuildwheel.yml

I'd like to take an "if it ain't broke" approach here for now. We can become adventurous after 0.18?

.github/workflows/cibuildwheel.yml
jni: In theory! I'd love to change this in a follow-up PR though, because the testing cycle is so long on changes here. Perhaps raise an issue with these tweaks to the build process to investigate once this is in?

.github/workflows/cibuildwheel.yml
jni: Actually, scrap that: this particular section is only triggered on 0.17 (see IF below). So I think it's appropriate to keep it.

jni: You can see in my screenshot that no 3.6 wheels were built.

.github/workflows/cibuildwheel.yml
hmaarrfk: ok

emmanuelle: `1.6.4` is the most recent version and `1.6.3` dates back to Oct 12, so quite recent.

emmanuelle:
Regarding tags, I have to admit that for 0.17 I had to delete tags because of such problems. My approach is that we should not touch a tag once files have been pushed to pypi but that it's not so bad to change the tag before that. But of course I realize this is not completely orthodox, so this needs to be discussed!

emmanuelle:
Also as it is written now, wheels will be built when new commits are pushed to master or release branches. Maybe we can tag after checking that the build went smoothly?

emmanuelle:
@jni not all builds were tested so I added the environment variables for testing all wheels. Also I tried to remove this strange installation of a pinned numpy and of certify, hopefully we don't need it.

jni:
Ok thanks @emmanuelle. One concern I have is that this is a lot of tests now, might we hit gh-actions timeout limits? I also would like to consider testing the wheels in a totally different action, or after the artifact is uploaded. It's frustrating at release time to be blocked by things that you know are unrelated, e.g. failure of pooch to download some data.

emmanuelle:
Hum it's quite long indeed. With tests being configured in environment variables I'm not sure it's easy to have the tests run after uploading artifacts. However, probably we could have more jobs in parallel by having the python version as an element of the matrix (I'm not sure why py3.9 is separated from py 3.8 and 3.7 at the moment, it seems to be the same configuration).

Also when do we want to build the wheels: not on every commit I guess? On release tags? On pushes to master and release branches? (that's what scikit-learn does)

jni:
> Also when do we want to build the wheels: not on every commit I guess? On release tags? On pushes to master and release branches? (that's what scikit-learn does)

I think this is already properly configured: on tags, pushes to master, pushes to release branches, and pushes to the "cibuildwheel" branch, which we can use whenever we want to make changes.

> Hum it's quite long indeed.

Personally I think it's overkill, but I don't mind leaving it for now, getting this merged, getting 0.18 out, and then tweaking for speed. I think we have done enough fiddling that I hope 0.18 will be quite smooth now!

jni:
@scikit-image/core anyone up for some approvals/merges? ;)

grlee77:
> Regarding tags, I have to admit that for 0.17 I had to delete tags because of such problems. My approach is that we should not touch a tag once files have been pushed to pypi but that it's not so bad to change the tag before that. But of course I realize this is not completely orthodox, so this needs to be discussed!

I have also overwritten tags in this manner for PyWavelets before (always before pushing anything to PyPI and typically within a short time frame of < 1 hour from tag creation).

.github/workflows/cibuildwheel.yml
grlee77: It seems we probably do want to add more recent manylinux builds fairly soon. The following recent blog post may be useful: https://labs.quansight.org/blog/2020/11/manylinux1-is-obsolete-manylinux2010-is-almost-eol-what-is-next/

grlee77 (approved):
I don't have GitHub Actions expertise or `cibuildwheel` experience, but the script is pretty understandable and the wheel artifacts look correct to me.

grlee77:
The documentation failure looks like a timeout that can be ignored

alexdesiqueira:
I'll try re-running it @grlee77.

alexdesiqueira:
@grlee77 I'll try re-running it.

emmanuelle:
The CircleCI failure is indeed quite common, unfortunately. It would be worth investigating it but it's not related at all to the PR.

alexdesiqueira:
@jni :joy:

jni:
Not sure if you were waiting for AppVeyor but I think that queue is like 700 builds long at this point so I wasn't about to. 😂

alexdesiqueira:
I was still waiting for it. Good call, then :joy:

PR #5082, opened 2020-11-26
peak_local_max: remove deprecated `indices` argument from examples

### LINKED ISSUE #5078: misleading commentary for peak_local_max example
In the example `plot_peak_local_max.py`, the intro at the top states:

> The ``peak_local_max`` function returns the coordinates of local peaks (maxima)
in an image. __A maximum filter is used for finding local maxima. This operation
dilates the original image and merges neighboring local maxima closer than the
size of the dilation. Locations where the original image is equal to the
dilated image are returned as local maxima.__

Everything after the first sentence is either wrong or misleading.  The output of `maximum_filter()`, called `image_max`, is not used for locating the maxima.  It is only used for making a panel in the visualization, and its utility is not evident. 

I suspect this was left over from some previous approach.

https://github.com/scikit-image/scikit-image/blob/ad737acfc8515eb17f7ac731aa92ad77918c3b8b/doc/examples/segmentation/plot_peak_local_max.py#L20-L22

odoublewen:
@mkcor  This PR updates examples to demonstrate the creation of a mask using the same method currently employed inside the `peak_local_mask` function, namely: `peak_mask[tuple(peak_idx.T)] = True`

As it turns out, the description in `plot_peak_local_max.py` was not as misleading as I initially thought.  The text is copied verbatim from the "Note" in the function docstring, where is is more clear that it's talking about the internal algorithm.   I think adding the word "Internally" to this note and to the example makes this a little more clear.

I think the main improvement here is updating the example in `skimage/feature/peak.py`, such that it demonstrates how to elegantly create a mask (when the indices contain more than one element!) and without using the deprecated `indices=False` method.

I built the docs on my laptop and my changes look good to me in the resultant HTML.

By the way, the plot_watershed.py example has an empty plot for "distances" -- this [exists on the live docs](https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_watershed.html#sphx-glr-auto-examples-segmentation-plot-watershed-py) as well as this PR.  Just noticed this but I could look into it and fix it if you'd like me to roll that into this PR.

## Description

Closes #5078 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_human_mitosis.py
+++ b/doc/examples/applications/plot_human_mitosis.py
@@ -172,10 +172,10 @@ print(cleaned_dividing.max())
 
 distance = ndi.distance_transform_edt(cells)
 
-local_maxi = feature.peak_local_max(distance, indices=False,
-                                    min_distance=7)
-
-markers = measure.label(local_maxi)
+local_max_coords = feature.peak_local_max(distance, min_distance=7)
+local_max_mask = np.zeros(distance.shape, dtype=bool)
+local_max_mask[tuple(local_max_coords.T)] = True
+markers = measure.label(local_max_mask)
 
 segmented_cells = segmentation.watershed(-distance, markers, mask=cells)
 
--- a/doc/examples/segmentation/plot_peak_local_max.py
+++ b/doc/examples/segmentation/plot_peak_local_max.py
@@ -4,9 +4,9 @@ Finding local maxima
 ====================
 
 The ``peak_local_max`` function returns the coordinates of local peaks (maxima)
-in an image. A maximum filter is used for finding local maxima. This operation
-dilates the original image and merges neighboring local maxima closer than the
-size of the dilation. Locations where the original image is equal to the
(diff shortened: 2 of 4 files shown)


jni (approved):
This is great! Thank you @odoublewen!

jni:
> By the way, the plot_watershed.py example has an empty plot for "distances" -- this exists on the live docs as well as this PR. Just noticed this but I could look into it and fix it if you'd like me to roll that into this PR.

Let's keep PRs focused on fixing one thing at a time. It makes things easier to review, both now and when investigating changes in the future. Thank you for this contribution @odoublewen!

By the way, we generally recommend that you make PRs from a new branch rather than master, so that it's easy to keep your master branch synchronised with our own.

Thanks again!

odoublewen:
> By the way, we generally recommend that you make PRs from a new branch rather than master, so that it's easy to keep your master branch synchronised with our own.

Will do for future contributions, thanks for the reminder.

PR #5083, opened 2020-11-27
Fix default level computation for find_contours

jni:
## Description

I noticed that the computation for the default find_contours level added in #4862 was actually incorrect. It should not be half the range, but the mean of the min and max. I've added a test to check for this. The test fails on master but passes here.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,7 +1,7 @@
 # Workflow to build and test wheels
 name: Test
 
-on: push
+on: [push, pull_request]
 
 jobs:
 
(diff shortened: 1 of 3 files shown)


jni:
Side note: looks like our gh tests are not properly configured! "on: push" presumably needs to be changed...

jni:
Boom, fixed! =)

jni:
@scikit-image/core this should go into the 0.18 release which is otherwise ready to go! Please review if you have a few mins! 😊 I might tag the release candidate anyway and backport this to the branch.

sciunto (approved):

emmanuelle (approved):
Good catch @jni!

jni:

PR #5090, opened 2020-11-30
BUG: random_walker: fix import path for umfpack 

grlee77:
## Description

`random_walker` seems to have a code path to allow use of UMFPACK, but it does not seem to get used even if `scikit-umfpack` is installed. The issue seems to be that the `umfpack` module is not visible via `scipy.sparse.linalg.dsolve`, but only within the `scipy.sparse.linalg.dsolve.linsolve` submodule. As far as I could tell, it has been this way in SciPy for many years. I verified locally that the tests still pass when `scikit-umfpack` is installed and added a new import test, although I don't think we ever run with `scikit-umfpack` on CI currently.

I thought about adding `scikit-umfpack` to `requirements/optional.txt`, but it does not have wheels on PyPI for recent Python or for Windows, so I have left it out for now. 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/segmentation/random_walker_segmentation.py
+++ b/skimage/segmentation/random_walker_segmentation.py
@@ -20,7 +20,7 @@ from .._shared.utils import warn
 # https://groups.google.com/d/msg/scikit-image/FrM5IGP6wh4/1hp-FtVZmfcJ
 # https://stackoverflow.com/questions/13977970/ignore-exceptions-printed-to-stderr-in-del/13977992?noredirect=1#comment28386412_13977992
 try:
-    from scipy.sparse.linalg.dsolve import umfpack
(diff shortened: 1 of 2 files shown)


emmanuelle:
Thanks @grlee77. I even wonder if we still need this warning. I was the one to play initially with the random walker algorithm, and to notice that things could be faster with umfpack. But I guess most people don't have it installed, and with the default mode using the Jacobi preconditioner maybe we could get rid of this warning altogether?

emmanuelle (approved):

jni (approved):

jni:
I agree with your comments @emmanuelle but something for future generations to worry about. =P

PR #5097, opened 2020-12-03
Improvements to CircleCI build: no parallelization and caching

emmanuelle:
This PR aims at reducing the frequent failures for the doc-build action with CircleCI.

The `EOFError` seem to be related to the coupled used of sphinx + multiprocessing so let's build with `j=1`. 

I also propose to cache some stuff (data files + requirements) but the difference is probably small.

PR #5099, opened 2020-12-04
disable Travis-CI builds

grlee77:
We do not have sufficient free credits to run this test matrix on their new billing policy.

We may want to re-enable for non-x86 architectures such as AArch64 going forward

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


grlee77:
Not sure if we want to keep this file around for future use or just rely on the git history for that? I think we may want to enable it again for ARM only going forward and perhaps only in the wheels repo. I think all repos under a single organization share a common pool of credits.

emmanuelle:
I also had the impression that credits are shared within an org, so :+1: for disabling Travis in this repo. I'd be in favor of removing the file altogether and relying on `git history`, but it's not a strong preference so I'm fine with merging the PR. I'm just a bit confused about the CI: it looks like all builds were cancelled.

grlee77:
I manually cancelled them since there was nothing new to test

emmanuelle (approved):
Thanks @grlee77, approved! Regarding CI, I also opened #5097 yesterday. Maybe we'll have green PRs soon ;-).

emmanuelle:
Oh and there are also the files in `tools/travis` which could be removed. (some of `tools/travis/notes.txt` could be useful for the other CI tools). Can be done in another PR ;-).

jni (approved):

jni:
Thank you @grlee77! `--pre` build failures should be fixed in #5103

PR #5101, opened 2020-12-04
Implement lazy submodule importing

stefanv:
**Note:** This is an experimental PR, so that we can discuss the idea and get feedback on its implementation.

This allows scikit-image to be used as follows:

```
import skimage as ski
ski.filters.gaussian(...)
```

The lazy loading strategy could also be implemented inside submodules,
to speed up imports of that submodule.  E.g., we could do:

```
sla = lazy.require('scipy.linalg')
```

The linalg module is only loaded once used in the code, avoiding the
standard workaround of having to move the import inside of a function
call.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/__init__.py
+++ b/skimage/__init__.py
@@ -68,10 +68,36 @@ dtype_limits
 
 """
 
-import sys
-
-
 __version__ = '0.18.0.dev0'
+
+__all__ = [
+    'color',
+    'data',
+    'draw',
+    'exposure',
+    'external',
+    'feature',
+    'filters',
+    'future',
+    'graph',
+    'io',
+    'measure',
+    'metrics',
+    'morphology',
+    'registration',
+    'restoration',
+    'segmentation',
+    '_shared',
+    'transform',
+    'util',
+    'viewer'
+]
+
+from .util.lazy import install_lazy
+__getattr__, __dir__ = install_lazy(__name__, __all__)
+
+
+import sys
 
 from ._shared.version_requirements import ensure_python_version
 ensure_python_version((3, 5))
--- a/skimage/util/lazy.py
+++ b/skimage/util/lazy.py
@@ -1,0 +1,39 @@
+# Adapted from https://stackoverflow.com/a/51126745/214686
+
+import sys
+import importlib
+import importlib.util
+
+
+def install_lazy(module_name, submodules):
+    def __getattr__(name):
+        if name in submodules:
+            lazy_mod = require(f'skimage.{name}')
+            return lazy_mod
+        else:
+            raise AttributeError(f'No skimage attribute {name}')
+
+
+    def __dir__():
+        return submodules
+
+    return __getattr__, __dir__
+
+
+def require(fullname):
+    if fullname in sys.modules:
+        return sys.modules[fullname]
+
+    spec = importlib.util.find_spec(fullname)
+    try:
+        module = importlib.util.module_from_spec(spec)
+    except:
+        raise ImportError(f'Could not lazy import module {fullname}') from None
+    loader = importlib.util.LazyLoader(spec.loader)
+
+    sys.modules[fullname] = module
+
+    # Make module with proper locking and get it inserted into sys.modules.
+    loader.exec_module(module)
+
+    return module


stefanv:
Our test runners are unhappy, but I swear it works locally ;)  Clearly some details to sort out around what numpy.distutils expects.

stefanv:
Tests passed for the previous commit; now added lazy functions for `skimage.filters`.

stefanv:
3.6 is expected to fail.  3.8 failures are unrelated.  Still waiting on the others.

hmaarrfk:
Does PyPy support `__getattr__`? I think that is why some people have been keeping 3.6 alive. is there a way to upgrade your `lazy` module to fallback on Python 3.6 behavior?

jarrodmillman:
I haven't tested PyPy's 3.7 implementation yet, since I've only been using it as part of our CI testing.  Fortunately, GH actions will soon support testing with PyPy3.7:
- https://github.com/actions/setup-python/pull/161
- https://foss.heptapod.net/pypy/pypy/-/issues/3354

I thought you were dropping support for Python 3.6 according to NEP 29.  Are you still planning to support Python 3.6 for the next release?  Or is the concern mostly related to CI testing?  Will this still be an issue once GH actions sorts things out or do you need another CI system to support PyPy3.7?

Out of curiosity, do you have many users using PyPy?  We try to support it for NetworkX, but we've never had a user ask us anything about it.  And none of our developers use it.  Do you have a way to tell if scikit-image is being run on PyPy in the wild?  If so, I would love to know so I can track it for NetworkX.

(I spend a lot of time trying to sort out CI stuff for PyPy and we've spent some time jumping through hoops trying to ensure we mostly work with PyPy.  It would be nice to know how many people (if any) are benefiting from the extra work.)

jarrodmillman:
I am pretty sure this requires https://www.python.org/dev/peps/pep-0562/, which is Python 3.7+.

stefanv:
This requires 3.7, and only works with 3.7 up.  Python 3.6 will no longer be supported in the next release.  I don't know whether this works on PyPy---I haven't tested.  But, I also don't know whether scikit-image works on PyPy.

rossbar (commented):
This looks really promising. The "demo" case in `skimage/filters/__init__.py` is great for showing how it's a radical departure from the way imports are currently specified (and arguably looks even better).

One thing that came to mind when thinking about how other libraries could adopt this is how modules that already define a `__getattr__` could be accommodated. Huge :+1: on the idea and thanks so much for working this up, I'm looking forward to learning more!

stefanv:
@rossbar, good question.  The easiest way would be to implement `__getattr__` as follows:

```
lazy_getattr, lazy_dir = install_lazy(...)

def __getattr__(name):
    if name in some_collection_I_want_to_handle:
        ...
    else:
        lazy_getattr(name)

def __dir__():
    my_list = [...]
    return lazy_dir() + my_list
```

skimage/util/lazy.py
```diff
@@ -0,0 +1,65 @@
...
+                if name in funcs:
+                    submod = importlib.import_module(
+                        f'{module_name}.{mod}'
+                    )
+                    return getattr(submod, name)
+        else:
+            raise AttributeError(f'No {module_name} attribute {name}')
+
+    def __dir__():
+        return submodules + all_funcs
```
Erotemic, line 44: It might be (very slightly) better to create a variable `__all__ =  submodules + all_funcs` and return that here via closure rather than execute the extra list concatenation each time `__dir__` is called. You're already doing this in the return statement so you could simply return `__all__` as well.

skimage/util/lazy.py
```diff
@@ -0,0 +1,65 @@
...
+    if submodules is None:
+        submodules = []
+
+    all_funcs = []
+    for mod, funcs in submod_funcs.items():
+        all_funcs.extend(funcs)
+
+    def __getattr__(name):
+        if name in submodules:
+            lazy_mod = require(f'skimage.{name}')
```
Erotemic, line 31: I really don't like how the name of the package is hard coded here. I'm not sure if there is an easy way around it, but I would prefer an implementation of this utility that could easily be plucked out of skimage and moved to importlib itself (or another third-party utility library).  I'm not sure what the best way to do that is ATM, but I'm just throwing that out there.

skimage/util/lazy.py
```diff
@@ -0,0 +1,65 @@
...
+            return lazy_mod
+        elif name in all_funcs:
+            for mod, funcs in submod_funcs.items():
```
Erotemic, line 34: There should definitely be a reverse lookup table here so you don't have to execute a loop every time a new attribute is accessed. 

Something like:

```python
# On the call to ``install_lazy``
name_to_submod = {func: mod for mod, funcs in submod_funcs.items() for func in funcs}
```

Then use it as 

```python
elif name in name_to_submod :
    mod = name_to_submod[name]
    submod = importlib.import_module(f'{module_name}.{mod}')
    return getattr(submod, name)
```

skimage/util/lazy.py
```diff
@@ -0,0 +1,65 @@
...
+
+    if submodules is None:
+        submodules = []
+
+    all_funcs = []
+    for mod, funcs in submod_funcs.items():
+        all_funcs.extend(funcs)
+
+    def __getattr__(name):
+        if name in submodules:
```
Erotemic, line 30: Cast `submodules` to a `set` before doing this for an `O(1)` test instead of an `O(N)` test.

skimage/filters/__init__.py
```diff
@@ -1,68 +1,29 @@
...
-           'threshold_multiotsu',
-           'apply_hysteresis_threshold',
-           'rank',
-           'unsharp_mask',
-           'window']
+from ..util import lazy
+__getattr__, __dir__, __all__ = lazy.install_lazy(
+    __name__,
+    submodules=['rank'],
+    submod_funcs={
```
Erotemic, line 5: I would rename this to `submod_attrs`. There is not guarantee that the values in this dictionary are referencing functions. They are attributes of the submodule. They could really be anything, even other submodules. I think the name "attr" would be better.

Erotemic (commented):
There are my 2cents.

Erotemic:
@jarrodmillman just pointed me here, and I'm liking the idea of this PR a lot. I left some comments about the implementation of `lazy_init`.

I'm going to probably spend a day or so working on adding my own variant of this functionality to my [mkinit package](https://github.com/Erotemic/mkinit). I'm hoping to add a formater that generates a `__getattr__` function or something like the way the `lazy_init` function is used here for a module. Assuming that works out, that should make it very easy to port to a new lazy init style. Ideally you will just write `mkinit <path-to-init.py> --lazy` and it will write the appropriate code to stdout. I'll comment again after I take a look at that.

Erotemic:
### Update on mkinit 

I've started playing around with this in mkinit. I have a working version. I've update the README that walks through a simple example that illustrates what mkinit does: https://github.com/Erotemic/mkinit/tree/dev/0.3.0#the-pitch This ends with a discussion of lazy importing. Note that I haven't finalized this feature yet and its progress is being tracked in https://github.com/Erotemic/mkinit/pull/10

With these new changes I ran: `mkinit skimage/filters/__init__.py --relative --lazy` which generated this file: 
https://gist.github.com/Erotemic/2b2b38bc87386d97ef91ede3b349f5d6

The tests seem to work, but it exposes many more attributes than the current init.py file. This can be controlled. By specifying the `__submodules__ : List[str]` attribute in the `__init__.py`, only modules specified in that variable will have themselves and their attributes exposed. Also, if the `__all__` variable in a submodule is specified, then only those attributes are exposed via mkinit.

I find it slightly weird that the only module currently exposed is ``rank`` and all other exposed names are attributes of submodules. There are mkinit CLI flags to only expose modules or only expose attributes, but mixing requires a bit of manual specification in the header of the `__init__.py` file (mkinit only clobbers the end of the file after all comments / special directives are read).

### General note on `__getattr__` efficiency 

On a more general note, inside the `__getattr__` function I'm specifying `globals()[name] = attr`, which will replace the more expensive function call with a dictionary lookup on all subsequent times lazy-attributes are accessed. Note, I do believe this only works in the case where `__getattr__` is defined in the module itself. Of course we could pass `globals()` to `lazy_install` or get fancy with `inspect` and stack frames if `lazy_install` is defined outside of this module.

stefanv:
@Erotemic Thanks for your feedback! I like the optimizations you propose and will integrate them as soon as I get back after the break. Generalizing this for use in other packages is in scope too.

Erotemic:
I've released the new version of [mkinit](https://pypi.org/project/mkinit/) 0.3.0 on pypi that contains the option to generate lazy imports. 

For now, you have to use the in-file boilerplate (i.e. every `__init__.py` defines its own `lazy_import` function). However, I have infrastructure setup so you will be able to pass in an option `--lazy_boilerplate`, which is a string that can contain custom code that overwrites how `lazy_import` is defined. (Thus if you had an internal definition you could just specify an import statement to get rid of all of the extra boilerplate).

Again, the current version might expose more names than the existing API, but that can be fixed by specifying the `__all__` attribute in each submodule to define what the user-facing API should be. (IMO that's better practice).

If you only want certain submodules exposed you can specify special global attributes:

```
    `__submodules__` - indicates the list of submodules to be
        introspected, if unspecified all submodules are introspected.

    `__protected__` -  Protected modules are exposed, but their attributes are not.

    `__private__` - Private modules and their attributes are not exposed.
```

Each of these global attributes should be a list of strings: (e.g.  `['{modname1'}, ...]` )

Effectively, `__submodules__` is a passlist, `__private__` is a blocklist, and `__protected__` will add a submodule to the API, but not its attributes. (Useful for forcing things like `numpy.linalg`). 

Note, if you use mkinit with `-w` specified (i.e. it will overwrite the file), it will only overwrite contents AFTER all lines involved in defining these variables and ALL comments have been parsed. So, you can perform any custom logic you want, just place a comment after it and the autogenerated code will always be written after it. (you can also specify `--diff`) to look at what mkinit would do.

skimage/util/lazy.py
stefanv: Done

skimage/util/lazy.py
stefanv: Fixed

skimage/util/lazy.py
stefanv: Done

skimage/util/lazy.py
stefanv: Done

skimage/filters/__init__.py
stefanv: Done

stefanv:
@Erotemic Thank you again for your feedback; I've incorporated it, and simplified the importing mechanism to avoid one level of indirection.  I'd appreciate it if you could take another look.

skimage/util/lazy.py
```diff
@@ -0,0 +1,52 @@
...
+
+    attr_to_modules = {
+        attr: mod for mod, attrs in submod_attrs.items() for attr in attrs
+    }
+
+    __all__ = submodules | attr_to_modules.keys()
+
+    def __getattr__(name):
+        if name in submodules:
+            return importlib.import_module(f'{module_name}.{name}')
```
Erotemic, line 40: In a previous version of this there was a "require" logic. It did stuff like `importlib.util.find_spec` and making `importlib.util.LazyLoader(spec.loader)`. I'm not sure if that is necessary or not. If its not necessary, great, I like getting rid of superfluous code, but I'm curious why it was there to begin with and what made you remove it?

skimage/util/__init__.py
```diff
@@ -14,10 +14,16 @@
 from .unique import unique_rows
 from ._invert import invert
 from ._montage import montage
-from ._map_array import map_array
 from ._label import label_points
 
 
+# Temporary workaround
```
Erotemic, line 20: Temporary workaround for what? Are you just testing a lazy load in skimage.util? Or is this working around a bigger issue. Is this a WIP? If this is intended to be merged, more context would be nice.

skimage/filters/__init__.py
```diff
@@ -23,49 +23,33 @@
...
-           'threshold_niblack',
-           'threshold_sauvola',
-           'threshold_triangle',
-           'threshold_multiotsu',
-           'apply_hysteresis_threshold',
-           'rank',
-           'unsharp_mask',
-           'window']
+from ..util import lazy
+print('module:', __name__)
```
Erotemic, line 27: Don't forget to remove this.

skimage/__init__.py
```diff
@@ -68,10 +68,35 @@
...
+    'transform',
+    'util',
+    'viewer'
+]
+
+from .util.lazy import install_lazy
+__getattr__, __dir__, _ = install_lazy(__name__, __all__)
 
-__version__ = '0.19.0.dev0'
+import sys
```
Erotemic, line 99: Is `import sys` needed here? I don't see any explicit usage of it.

Erotemic:
@stefanv I took a look at this and added a few comments.

Something else I saw is that you aren't adding the returned attribute to the globals dictionary. I would modify `install_lazy` to accept `globals()` as a parameter, and then when you return a lazy attribute, add it to the `globals()` dictionary so you don't incur function overhead when you subsequently access the variable. E.g:


```python

def install_lazy(module_name, submodules=None, submod_attrs=None, globals_=None):
    ...
    def __getattr__(name):
        ...
           attr = importlib.import_module(f'{module_name}.{name}')
           ...
           attr = getattr(submod, name
       ... 
       globals_[name] = attr
       return attr
    ...

```

replace the returns by assigning the variable to `attr`. 

You could even use the `inspect` module to grab the globals from the parent frame if they aren't passed in:

```python
def get_stack_frame(n=0, strict=True):
    """
    Gets the current stack frame or any of its ancestors dynamically

    Args:
        n (int): n=0 means the frame you called this function in.
                 n=1 is the parent frame.
        strict (bool): (default = True)

    Returns:
        frame: frame_cur

    Example:
        >>> frame_cur = get_stack_frame(n=0)
        >>> print('frame_cur = %r' % (frame_cur,))
        >>> assert frame_cur.f_globals['frame_cur'] is frame_cur
    """
    frame_cur = inspect.currentframe()
    # Use n+1 to always skip the frame of this function
    for ix in range(n + 1):
        frame_next = frame_cur.f_back
        if frame_next is None:  # nocover
            if strict:
                raise AssertionError('Frame level %r is root' % ix)
            else:
                break
        frame_cur = frame_next
    return frame_cur

...

if globals_ is None:
    globals_ = get_stack_frame(n=1).f_globals
```


(Side note I'm going to release 0.3.1 of mkinit later today https://github.com/Erotemic/mkinit/pull/11 is there any sort of integration with that that we should consider here?)

skimage/util/lazy.py
```diff
@@ -0,0 +1,52 @@
...
+    if submodules is None:
+        submodules = set()
+    else:
+        submodules = set(submodules)
+
+    attr_to_modules = {
+        attr: mod for mod, attrs in submod_attrs.items() for attr in attrs
+    }
+
+    __all__ = submodules | attr_to_modules.keys()
```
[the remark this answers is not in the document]
stefanv: Modifying this to be a list, just to be safe.

skimage/util/lazy.py
stefanv: I think that was primarily because of the evolution of the idea—I started with lazy module imports, and then added the other layers on top.

With `__dir__` in place, you can inspect the module, and only on `__getattr__` should the module be loaded.  The loaded module has the same setup, so it won't be expensive to import either.

You have a keen eye, I must say!

skimage/util/__init__.py
stefanv: OK, I updated the documentation string here to be more detailed.

skimage/__init__.py
stefanv: Moved to where it is used.

skimage/__init__.py
```diff
@@ -68,10 +68,35 @@
...
+    'restoration',
+    'segmentation',
+    '_shared',
+    'transform',
+    'util',
+    'viewer'
+]
+
+from .util.lazy import install_lazy
+__getattr__, __dir__, _ = install_lazy(__name__, __all__)
```
hmaarrfk, line 97: Does this work with Python 3.5?

jni: No, but we only support 3.7+?

stefanv: I presume it works but it doesn't do anything.  But Python 3.5 is ancient news now.

stefanv:
@Erotemic I'm not quite sure how to handle the globals dict correctly.  Your suggestion makes me think that `globals()[lib] = module` is not enough, and that you need to track the globals of the parent frame.  Could you explain to me why that is necessary? I will be working on this actively the next few days; sorry for the long delay between your review and my response.

skimage/__init__.py
hmaarrfk: I'm asking this in the context that on line 77 /102 we have the lines `ensure_python_version((3, 5))

We tried to ensure that we had a nice message to the end users, instead of a scary python stack trace.

image

Erotemic:
@stefanv `globals()[lib] = module` is enough, if the call to `globals()` is in the `__init__` file, but it is in the `util_lazy` file instead, which means when you call it you will get `util_lazy`'s globals instead of the `__init__` file's `globals()`.

I would recommend explicitly passing globals() as an argument to the function. The inspect method was just showing how you could access the caller's globals, but I think just forcing it to be passed isn't a big deal.

----

Either that or you could use `mkinit` to just generate the file for every directory in the module. 

In version >=0.3.1  running 

```bash 
# Install mkinit
pip install mkinit

# Point to the path to the `__init__` file to autogenerate use --diff to view the changes
mkinit ~/code/scikit-image/skimage/transform/__init__.py --lazy --diff

# Use -w to accept the changes
mkinit ~/code/scikit-image/skimage/transform/__init__.py --lazy -w

```

results in 

```python
def lazy_import(module_name, submodules, submod_attrs):
    import importlib
    name_to_submod = {
        func: mod for mod, funcs in submod_attrs.items()
        for func in funcs
    }

    def __getattr__(name):
        if name in submodules:
            attr = importlib.import_module(
                '{module_name}.{name}'.format(
                    module_name=module_name, name=name)
            )
        elif name in name_to_submod:
            submodname = name_to_submod[name]
            module = importlib.import_module(
                '{module_name}.{submodname}'.format(
                    module_name=module_name, submodname=submodname)
            )
            attr = getattr(module, name)
        else:
            raise AttributeError(
                'No {module_name} attribute {name}'.format(
                    module_name=module_name, name=name))
        globals()[name] = attr
        return attr
    return __getattr__


__getattr__ = lazy_import(
    __name__,
    submodules={
        'finite_radon_transform',
        'hough_transform',
        'integral',
        'pyramids',
        'radon_transform',
        'setup',
        'tests',
    },
    submod_attrs={
        'finite_radon_transform': [
            'frt2',
            'ifrt2',
        ],
        'hough_transform': [
            'hough_circle',
            'hough_circle_peaks',
            'hough_ellipse',
            'hough_line',
            'hough_line_peaks',
            'label_distant_points',
            'probabilistic_hough_line',
        ],
        'integral': [
            'integral_image',
            'integrate',
        ],
        'pyramids': [
            'pyramid_expand',
            'pyramid_gaussian',
            'pyramid_laplacian',
            'pyramid_reduce',
        ],
        'radon_transform': [
            'iradon',
            'iradon_sart',
            'order_angles_golden_ratio',
            'radon',
        ],
        'setup': [
            'base_path',
            'configuration',
        ],
        'tests': [
            'setup',
            'teardown',
        ],
    },
)


def __dir__():
    return __all__

__all__ = ['base_path', 'configuration', 'finite_radon_transform', 'frt2',
           'hough_circle', 'hough_circle_peaks', 'hough_ellipse', 'hough_line',
           'hough_line_peaks', 'hough_transform', 'ifrt2', 'integral',
           'integral_image', 'integrate', 'iradon', 'iradon_sart',
           'label_distant_points', 'order_angles_golden_ratio',
           'probabilistic_hough_line', 'pyramid_expand', 'pyramid_gaussian',
           'pyramid_laplacian', 'pyramid_reduce', 'pyramids', 'radon',
           'radon_transform', 'setup', 'teardown', 'tests']
```

(Use `mkinit --help` to get information about special variables you can define to modify the autogeneration behavior; also not it  only overwrites part of the file that looks like it was generated, so you can add custom code in the `__init__` and still use mkinit).

But the util_lazy variant does have its own advantages, so no need to use mkinit if you don't want to, but I think it also helps illustrates behavior of `globals`.

------

and because `lazy_import` is defined in the `__init__` module itself you don't have to worry about `globals()` being called in the wrong frame. But if you move that function to a different file, you would need to pass them as a parameter as I described at the top of my previous comment.

stefanv:
@Erotemic Is there any reason we *have* to use`globals` though.  We could keep a dictionary of all imported modules around and check that before doing an implicit import.  Also, are we sure that `import_module` doesn't already do a check like that?  I.e., did you benchmark to see whether there is a noticeable slowdown?

Erotemic:
If there is a reason, I don't recall it clearly. I have a feeling there might be, but its not a strong feeling. I do know that I never benchmarked it. If there is a way to avoid calling globals, that would be best.

If you want to benchmark it, I would recommend using my timerit library to help with benchmarking. Its much more scriptable than timeit.

stefanv:
> If you want to benchmark it, I would recommend using my [timerit](https://github.com/Erotemic/timerit) library to help with benchmarking. Its much more scriptable than timeit.

I often need something like `Timerit`, although the following API feels a bit more intuitive:

```
with Timerit() as timer:
    for _ in timer:
        some_func()
```

Erotemic:
Fortunately, it's a variable so you can absolutely use that syntax :stuck_out_tongue_winking_eye:. I should add an example using `_` in the README because it is probably more idiomatic. But personally, I usually prefer giving things names; I think it emphasizes that it is an object with attributes. The same thing is true of an instance of `Timerit`. If you reuse the same `Timerit` instance via the `Timerit.reset(label)` method then look at the `rankings` properties it will compare the times from all the instances of unique labels you passed to reset. 

Here is an example (which also makes use of ubelt.repr2) to do this:

```python
import timerit
import ubelt as ub

ti = timerit.Timerit(100, bestof=10, verbose=2)


params = [
    {'num': 100},
    {'num': 1000},
    {'num': 10000},
]


for param in params:

    paramkey = ub.repr2(param, itemsep='', kvsep='=', sv=1, sk=1, nobr=1)

    for _ in ti.reset(f'for loop + append  - {paramkey}'):
        with _:
            accum = []
            for i in range(param['num']):
                accum.append(i ** 2)

    for _ in ti.reset(f'list comprehension - {paramkey}'):
        with _:
            accum = [i ** 2 for i in range(param['num'])]


print('ti.rankings = {}'.format(ub.repr2(ti.rankings, nl=2, align=':', precision=6)))
```

And the results look like this:
```
Timed for loop + append  - num=100, for: 100 loops, best of 10
    time per loop: best=27.951 µs, mean=35.451 ± 8.4 µs
Timed list comprehension - num=100, for: 100 loops, best of 10
    time per loop: best=22.394 µs, mean=22.503 ± 0.1 µs
Timed for loop + append  - num=1000, for: 100 loops, best of 10
    time per loop: best=288.862 µs, mean=297.921 ± 4.4 µs
Timed list comprehension - num=1000, for: 100 loops, best of 10
    time per loop: best=230.884 µs, mean=232.965 ± 2.5 µs
Timed for loop + append  - num=10000, for: 100 loops, best of 10
    time per loop: best=2.998 ms, mean=3.033 ± 0.0 ms
Timed list comprehension - num=10000, for: 100 loops, best of 10
    time per loop: best=2.303 ms, mean=2.369 ± 0.0 ms
ti.rankings = {
    'mean'    : {
        'list comprehension - num=100,'  : 0.000023,
        'for loop + append  - num=100,'  : 0.000035,
        'list comprehension - num=1000,' : 0.000233,
        'for loop + append  - num=1000,' : 0.000298,
        'list comprehension - num=10000,': 0.002369,
        'for loop + append  - num=10000,': 0.003033,
    },
    'mean+std': {
        'list comprehension - num=100,'  : 0.000023,
        'for loop + append  - num=100,'  : 0.000044,
        'list comprehension - num=1000,' : 0.000235,
        'for loop + append  - num=1000,' : 0.000302,
        'list comprehension - num=10000,': 0.002395,
        'for loop + append  - num=10000,': 0.003068,
    },
    'mean-std': {
        'list comprehension - num=100,'  : 0.000022,
        'for loop + append  - num=100,'  : 0.000027,
        'list comprehension - num=1000,' : 0.000230,
        'for loop + append  - num=1000,' : 0.000294,
        'list comprehension - num=10000,': 0.002342,
        'for loop + append  - num=10000,': 0.002999,
    },
    'min'     : {
        'list comprehension - num=100,'  : 0.000022,
        'for loop + append  - num=100,'  : 0.000028,
        'list comprehension - num=1000,' : 0.000231,
        'for loop + append  - num=1000,' : 0.000289,
        'list comprehension - num=10000,': 0.002303,
        'for loop + append  - num=10000,': 0.002998,
    },
}
```

I think it serves as fairly readable, scriptable, and reproducible proof that of any benchmark claim. I probably should add this example to the docs.

stefanv:
Ah, yes, I see why it needs to be that way around; there is no other way of trigger the iteration time measurement!

Erotemic:
@stefanv Something that might be useful is adding some environment variable that forces all lazy modules to be eagerly imported. I'll probably modify mkinit to respect some environ like 'EAGER_IMPORT' or something. That will make it easier to debug import failures and also find what imports are slow.

stefanv:
It's a good idea, but I think the libraries would have to do that as a conditional around the replacement of __getattr__.

Erotemic:
You call a function in each `__init__` to define `__getattr__` and in that function you get a list of all the values you *could* call it with. So, just call it with those values before you return it.

stefanv:
So, you mean use the same machinery, just do them "one shot" instead of as they're needed?

Erotemic:
Yes, something like this:

```python
def lazy_import(module_name, submodules, submod_attrs):
    import importlib
    import os
    name_to_submod = {
        func: mod for mod, funcs in submod_attrs.items()
        for func in funcs
    }

    def __getattr__(name):
        if name in submodules:
            attr = importlib.import_module(
                '{module_name}.{name}'.format(
                    module_name=module_name, name=name)
            )
        elif name in name_to_submod:
            submodname = name_to_submod[name]
            module = importlib.import_module(
                '{module_name}.{submodname}'.format(
                    module_name=module_name, submodname=submodname)
            )
            attr = getattr(module, name)
        else:
            raise AttributeError(
                'No {module_name} attribute {name}'.format(
                    module_name=module_name, name=name))
        globals()[name] = attr
        return attr

    if os.environ.get('EAGER_IMPORT', ''):
        for name in name_to_submod.values():
            __getattr__(name)

        for attrs in submod_attrs.values():
            for attr in attrs:
                __getattr__(attr)
    return __getattr__
```

One of the gripes I get from people who use my code is that I tend to do a lot of imports lazily, and they just want to do it all at the beginning so they can debug dependency issues in their system. While I think lazy is on average better, I do see the utility in being able to do an eager import without much fuss (e.g. setting an environment variable).

stefanv:
Discussion of the SPEC: https://discuss.scientific-python.org/t/spec-1-lazy-loading-for-submodules/25/3

skimage/__init__.py
grlee77: Yeah, it seems the `ensure_python_version` call below should either be removed or updated to use `(3, 7)` instead of `(3, 5)`. If updated, it may need to be called prior to `install_lazy`?

stefanv: Ah, thanks, I get what you're saying now @hmaarrfk !

stefanv:
@Erotemic I have now added the `EAGER_IMPORT` environment variable.  I had to fix a bunch of circular imports that never surfaced until this PR :thinking:

`EAGER_IMPORT` would probably be enabled for one of the entries in the test matrix; it caught a typo in LPIFilter2D, that I wouldn't have seen otherwise.

stefanv:
Does anyone know what might be up with `linux-cp3.7-mini-req`?

stefanv:
OK, looks like external is not being picked up by the sdist.

stefanv:
Failure is due to timeout on Windows builder.

skimage/filters/__init__.py
```diff
@@ -22,50 +22,33 @@
...
-           'threshold_triangle',
-           'threshold_multiotsu',
-           'apply_hysteresis_threshold',
-           'rank',
-           'unsharp_mask',
-           'window']
+__getattr__, __dir__, __all__ = lazy.install_lazy(
+    __name__,
+    submodules={'rank'},
+    submod_attrs={
```
grlee77, line 30: I see that specifying the attributes here seems to also have a desirable effect that when the user is in an interactive IPython session and does
```Python
skimage.filters.<TAB>
``` 
only the functions and not the submodule names show up in the list of autocompletions.

On the contrary, for `skimage.metrics`, I see the submodules `set_metrics` and `simple_metrics` show up in the list of autocompletion suggestions. Do you recommend extending this approach there or should we instead/also rename `set_metrics` -> `_set_metrics`, etc?

grlee77 (approved):
This looks good to me and worked nicely in local testing. I mainly just had a question on the use of `submod_attrs` and possibly renaming some submodules that were probably intended to be private, but were not named with a leading underscore.

skimage/filters/__init__.py
stefanv: I think this is an advantage of this approach, yes!  You can present whatever facade you want, despite how you implement things in the background.  Users were *never* meant to see our sub-submodules, and like you say the only way to suggest that to them before was to name them with an underscore.

stefanv:
@grlee77 I added back the lazy loading proxy (it was useful to NetworkX, and the pattern they want it use it for is pretty pervasive).   I'm unsure on the name, in case you have suggestions. I also updated the docstring of the main function.

I mention this to you so you have a chance to withdraw your approval, in case you don't like the new changes!

skimage/_shared/lazy.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Typically, modules import submodules and attributes as follows::
+
+      import mysubmodule
+      import anothersubmodule
+
+      from .foo import someattr
+
+    The idea is to replace a module's `__getattr__`, `__dir__`, and
+    `__all__`, such that all imports work exactly the way the would
```
grlee77, line 18:
```suggestion
    `__all__`, such that all imports work exactly the way they did
```

stefanv:
@jni Given that you used this in napari, would you like to give it a look-over?  I updated the docstrings, if you want to take that over to Napari.

stefanv:
I modified `lazy.load` so that it raises a custom exception (that contains the module name being loaded).  I also renamed `install_lazy` to `lazy.install` for consistency.

skimage/__init__.py
```diff
@@ -68,13 +68,38 @@
...
+    'util',
+    'viewer'
+]
+
 from ._shared.version_requirements import ensure_python_version
-ensure_python_version((3, 5))
+ensure_python_version((3, 7))
+
+from ._shared import lazy
+__getattr__, __lazy_dir__, _ = lazy.install(__name__, __all__)
```
Erotemic, line 98: In this case "__all__" happens to be only a list of submodules, so this works, but I think it makes things confusing. If someone unfamiliar with the code looked at this it may appear that either `lazy.install` is able to figure out what each item in `__all__` means, or something else.

I think what would be a more intuitive way to write this code is if the first `__all__` variable was renamed to `__submodules__` or `submodules` to make it clear that it is simply a list of submodules, and not necessarily any imported attributes. Then change the `_` return to either be `__all__` or rename it to `__lazy_all__`, and so something similar like you are doing with `__dir__`.

(Or maybe, you only return `lazy.install` only returns the `__getattr__` and a `__lazy_all__` variable and it is the user's resposibility to setup `__dir__` and `__all__`, because you are kind of doing that already?)

Erotemic: Personally, I don't like the name `install`, that evokes some sort of installation process to me (like apt install or running a windows installer). Obviously we can't use `lazy.import`, but are we fixed on using the `install` terminology? It feels like `lazy_import`, `lazyimport`, or something else might be better.

Erotemic: Maybe `lazy.expose`?

stefanv: I think anything that doesn't re-use the module name is fine; so `lazy.expose` would work (although `expose` has a bit of a weird connotation in general).  @jarrodmillman Naming thoughts?

stefanv: I think the `__dir__` return is convenient, but it is not usable in all cases.  The naming change suggestions are sensible, I will update those.

jarrodmillman: How about `lazy.attach`?  (I also thought about, but didn't like as much: `lazy.setup`, `lazy.manifest`, `lazy.teether`, `lazy.connect`).

stefanv: I've renamed it to `lazy.attach` for now.

Erotemic: Maybe we need make PEP for Python 3.11 to add a `__call__` to modules so we can just do `lazy()` :smile:. I would like to find a better name than `attach`, but I don't have better ideas at the moment. In this spirit of brainstorming I'll just throw out ideas, some are probably bad: `init`, `new`, `module`, `define`, `delayed_import`, `inject`, `load`. I'm actually kind of partial to `lazy.load(...)`, 4 characters, reasonably clear, alteration, no duplication.

stefanv: Except that lazy.load is now used for lazy loading of external models, so then we'd need an alternative name for that function.

I used 'require' originally after Javascript and emacs.

Erotemic: I like `require`.

stefanv: Currently, there is a clear distinction between the functionality of the two functions: `attach` configures a bunch of things, whereas `load` does an actual import.  If we can find other words to convey that, we can revisit.

stefanv: I like the `__call__` on the module idea, btw :D

dschult: The [difference between](https://stackoverflow.com/questions/19198166/whats-the-difference-between-a-module-and-a-library-in-python) `library`, `package` and `module` is subtle, but I think it is useful when talking about this kind of namespace management.  The goal of the `__init__.py` module in a `library` is to create a `package` with a namespace that includes `subpackages` and attributes from `subpackages`. Technically `module` is a Python object, though it is often used to refer to a file containing the code for this object.  To help organize modules and provide a naming hierarchy, Python has a concept of packages.  As a rough analogy, modules are like files while packages are like directories. The `__init__.py` module in the package determines the namespace for the package. The term subpackage is used similarly to the term subdirectory for file systems.  I'm thinking that a module is defined in a single file, a package is held in a directory with an `__init__.py` file to load more than one module.  That package could then be a subpackage of another package, etc.  In the current code the term `module` is used to refer to many different kinds of objects -- though they are technically all python objects, the distinction between code from a single file and code from a directory is useful here.

module <-> python file
package <-> python directory (with __init__.py file)
subpackage <-> python directory within another python directory (like subdirectory)

So, for the name of the `attach` function, I like `lazy.package_structure`.  That is, this function constructs the package structure for this package. We give it a name for the package, a list of subpackages and attributes to include in the namespace and it lazily adds those entries to the namespace. 

The name is definitely longer than `attach`, and it is not a verb, so maybe there is a better name. But by not being a verb, it makes clear that this is where we set up the structure for the namespace of this package.

stefanv: I agree that the distinction is important, but I am not sure that `lazy.package_structure` captures that correctly.  `lazy.attach` is, indeed, used to set up members of a module, not a package, and can be used by many modules and submodules *within* a package.  `package_structure` would refer to the overarching package (thus, multiple modules / submodules).

skimage/_shared/lazy.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return __getattr__, __dir__, list(__all__)
+
+
+class LazyImportError(ImportError):
+    def __init__(self, module, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.module = module
+
+
```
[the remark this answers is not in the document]
stefanv: Removed.

skimage/_shared/lazy.py
```diff
@@ -0,0 +1,133 @@
...
+          spla = lazy_import('scipy.linalg')  # import scipy.linalg as spla
+
+    Returns
+    -------
+    pm : importlib.util._LazyModule
+        Proxy module.  Can be used like any regularly imported module.
+
+    """
+    if fullname in sys.modules:
+        return sys.modules[fullname]
```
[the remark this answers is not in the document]
stefanv: I changed it to use this form.

stefanv:
Thank you for the thorough review, @dschult!  I committed most of your suggestions, and incorporated the code changes you recommended.  I am not quite sure about the documentation language yet, so I left that until we have time to discuss further.

skimage/filters/__init__.py
```diff
@@ -1,74 +1,31 @@
...
-           'window',
-           ]
+__getattr__, __dir__, __all__ = lazy.attach(
+    __name__,
+    submodules={'rank'},
+    submod_attrs={
+        'lpi_filter': ['inverse', 'wiener', 'LPIFilter2D'],
+        '_gaussian': ['gaussian', '_guess_spatial_dimensions',
+                      'difference_of_gaussians'],
+        'edges': ['sobel', 'sobel_h', 'sobel_v',
```
hmaarrfk, line 10: My general worry about things like this, is that it breaks people's internal knowledge of what python is according to other tutorials.

The barrier to contribution is thus increased from:
* Someone that knows how to contribute a good algorithm
to
* Someone who knows all the funny tricks used at scikit-image to get marginal gains.

Currently the import time is dominated by two packages:

image

```
# You might need to redirect 1 not 2, but for some reason I couldn't redirect both...
python -X importtime -c "import skimage" 2>import.log
tuna import.log
```

hmaarrfk: From that graph, we are limited by importing numpy, and requests from pooch. 

I think our efforts would be better spent making pooch more "on demand" and biting the bullet on numpy. Numpy seems pretty standard so far as all these libraries go.

hmaarrfk: Leaving it here for further exploration, but this PR is pretty specific about a submodule that is pretty slow to load.

I think a fair benchmark would be to compare the import time of a few key functions:

```
python -X importtime -c "import numpy; import pooch; from skimage.filters.edges import sobel" 2>import.log && tuna import.log
```
Should be able to help you compare the import time of the specific edge filter, when all other functions are "lazy"

jni: @hmaarrfk although I totally get your concerns, I think the benefits are bigger than you're suggesting, because now people are able to `import skimage; result = skimage.filters.gaussian(image)`. I think that actually *helps* a lot of beginners who are not aware that submodules don't get imported automatically.

jni (approved):
Is this ready as far as you're concerned @stefanv? It's 👍 from me! (though we should talk through @hmaarrfk's concerns before merging)

skimage/filters/__init__.py
hmaarrfk: that is an important benefit

hmaarrfk:
i think there is a simpler way to get the one benefit you listed. but i need to think about it more.

this shouldn't stop others from merging but i would be a little more concerned in propagating this pattern too far.

stefanv:
It might be worth reading through the [SPEC](https://github.com/scientific-python/specs/blob/f719b41c4cdb9c880f0602f3d3b9535321be074a/spec-0001.md) for motivation.  I am definitely not trying to force this onto scikit-image, but I think the benefits are:

a) Easy interactive exploration of the entire API without special imports
b) Developers don't have to jump through special hoops to ensure that modules are written to have small import times (see also the provided lazy external module importer, which I haven't yet used in this PR)

This all happens transparently to the user (if not entirely to developers who add new submodules).

@hmaarrfk The external module importer may well address the pooch import times.

@jni I'm pretty happy with this for now.  @dschult made some great suggestions which I incorporated.  There were some documentation language fixes that I wasn't sure about; maybe you can help me cast an eye over those?

hmaarrfk:
Thanks for the reading material @stefanv very informative!

I can't say i love how the file `skimage/filters/__init__.py`` reads, but i've tried to use lazy importing before and the other organization I came up with before read even more poorly.

(Y)

hmaarrfk (approved):

stefanv:
Thanks, Greg, sorry that I didn't get to the editing of the commit message quickly enough!

PR #5103, opened 2020-12-05
Replace np.bool, np.int, np.float with np.bool_, np.int_, np.float_

jni:
## Description

This should fix CI failures with numpy 1.20rc

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_metrics.py
+++ b/benchmarks/benchmark_metrics.py
@@ -9,8 +9,8 @@ except ImportError:
 
 class SetMetricsSuite(object):
     shape = (6, 6)
-    coords_a = np.zeros(shape, dtype=np.bool)
-    coords_b = np.zeros(shape, dtype=np.bool)
+    coords_a = np.zeros(shape, dtype=np.bool_)
+    coords_b = np.zeros(shape, dtype=np.bool_)
 
     def setup(self):
         try:
--- a/benchmarks/benchmark_metrics.py
+++ b/benchmarks/benchmark_metrics.py
@@ -9,8 +9,8 @@ except ImportError:
 
 class SetMetricsSuite(object):
     shape = (6, 6)
-    coords_a = np.zeros(shape, dtype=np.bool_)
-    coords_b = np.zeros(shape, dtype=np.bool_)
+    coords_a = np.zeros(shape, dtype=bool)
+    coords_b = np.zeros(shape, dtype=bool)
 
     def setup(self):
         try:
--- a/benchmarks/benchmark_restoration.py
+++ b/benchmarks/benchmark_restoration.py
@@ -130,7 +130,7 @@ class RollingBall(object):
     peakmem_rollingball.param_names = ["radius"]
 
     def time_rollingball_nan(self, radius):
-        image = data.coins().astype(np.float_)
+        image = data.coins().astype(float)
         pos = np.arange(np.min(image.shape))
         image[pos, pos] = np.NaN
         restoration.rolling_ball(image, radius=radius, nansafe=True)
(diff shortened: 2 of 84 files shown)


jni:
@scikit-image/core This is ready for review! I don't think I've ever seen so much green! 😍 💚 💚 💚

alexdesiqueira (approved):
Huge work @jni, looks great :) Thank you!

jni:
Since [some people on the NumPy core team](https://github.com/stefanv) are not aware that NumPy deprecated `np.{bool,float,int}`, here's a link to one of the CI failures + warning soups that motivated this PR. 😂 

https://github.com/scikit-image/scikit-image/runs/1497565948

hmaarrfk:
Did you use a regexp for this?

jni:
@hmaarrfk VSCode global find and replace. I followed each with replacing e.g. `np.bool__` with `np.bool_`, which took care of the existing `np.bool_`s.

jni:
@stefanv informs me that `dtype=float` is perfectly acceptable. Nevertheless, the Python typing fan in me prefers `np.float_` because that is, in fact, a dtype.

jni:
Update: @stefanv's position is actually stronger: he things `dtype=np.{bool,int,float}_` is an antipattern, and we should use `dtype={bool,int,float}`. He has asked @seberg, who is leading NumPy's dtype extension efforts, to confirm this. I'll update the PR accordingly.

hmaarrfk:
I just admit the release notes were not very clear about which one people should use

jni:
Looking green again, finally.

emmanuelle:
Thanks @jni! Merging.

seberg:
Thanks for updating quickly :). Just in case it seems like it.  While I *like* this deprecation (and think cleaning it up in `skimage` is good), it is not actually important for dtypes.
This is just the first release were we can deprecate these confusing names, so if this is creating too much noise downstream or for end-users, I am happy to pull it out.

hmaarrfk:
I have to say, i got into the habbit of doing `np.int` in my own projects, and i'm kinda troubled by the need to go and undo all of this.

Not really sure what the goal is, but yeah, it will be pretty painful to change everybody's habbits.

Is the goal to stop people from doing `from numpy import *`, if so, something should catch that and warn against that.

seberg:
@hmaarrfk the use is OK for `dtype=np.int`, what is a bit troubling is the expectation to get a NumPy scalar for `np.int(1)`. I do think we had issues on this, but while it seems clear that we don't want users to write `np.int`, I am not sure that this particular change is a good use of our "annoy the user base" chip allocation :).

Maybe bring it up on the mailing list?

stefanv:
@jni Interesting side-effect of seach-and-replace :)

```
    if (input_dtype in (bool, bool) or out_dtype in (bool, bool)):
```

jni:

jni:
@seberg well it was very surprising to me when I found out that `np.float is float`. For the longest time I thought that `np.float` was equivalent to "whatever the default float value is on my platform", and considered it best practice to use that instead of plain `float`. 😅 I think that is a common misconception. In my ideal world `np.int` would become `np.int_` and so on. That would use up even more downstream annoyance credits but would get us to a nice place.

jni:
@stefanv that is perfectly cromulent code. =P

jni:
Started a thread on the NumPy mailing list as suggested by @seberg

emmanuelle:
Argh I had overseen that the minimal requirements jobs fail on GHA. @jni is this known / taken care of in another PR?

emmanuelle:
See #5109 for the CI failure.

PR #5104, opened 2020-12-05
Use v{} for version tags with pooch

jni:
## Description

The pooch URL was incorrect for released version tags given our convention to use v0.18.0 rather than just 0.18.0 for releases.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -125,7 +125,8 @@ def create_image_fetcher():
     # This helps pooch understand that it should look in master
     # to find the required files
     pooch_version = __version__.replace('.dev', '+')
-    url = "https://github.com/scikit-image/scikit-image/raw/{version}/skimage/"
+    url = ("https://github.com/scikit-image/scikit-image/raw/"
+           "v{version}/skimage/")
 
     # Create a new friend to manage your sample data storage
     image_fetcher = pooch.create(
--- a/skimage/data/__init__.py
+++ b/skimage/data/__init__.py
@@ -125,8 +125,12 @@ def create_image_fetcher():
     # This helps pooch understand that it should look in master
     # to find the required files
     pooch_version = __version__.replace('.dev', '+')
-    url = ("https://github.com/scikit-image/scikit-image/raw/"
-           "v{version}/skimage/")
+    if '+' in pooch_version:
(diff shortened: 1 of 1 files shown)


jni:
Reference thread on Zulip: https://skimage.zulipchat.com/#narrow/stream/175598-general/topic/release.200.2E18/near/218161832

But, looks like this works for versions but not master... Hold please...

emmanuelle (approved):
Thanks @jni. pre failures have been addressed in #5103.

emmanuelle:
We'll need a second person to review/approve/merge here.

skimage/data/__init__.py
```diff
@@ -125,7 +125,12 @@ def create_image_fetcher():
     # This helps pooch understand that it should look in master
     # to find the required files
     pooch_version = __version__.replace('.dev', '+')
-    url = "https://github.com/scikit-image/scikit-image/raw/{version}/skimage/"
+    if '+' in pooch_version:
```
hmaarrfk, line 128: why is thel ogic different?

jni: @hmaarrfk for master, the data lives in `github.com/scikit-image/scikit-image/raw/master/skimage/...`, while for version tags, it's in `github.com/scikit-image/scikit-image/raw/v0.18.0/skimage/...`, for example.

Technically, the approach is not quite correct: the v0.18.x branch will still look in master. This happens to work, but might not always. For now, it gets us unstuck. In the future, maybe we move to setuptools_scm, which would solve a lot of issues more elegantly.

emmanuelle: This, and pooch knows how to convert internally `version` to `master` so that in the url it's `master`, I guess?

emmanuelle:
@scikit-image/core please review/merge, this is a fix needed for the release.

mkcor (approved):

emmanuelle:
Thank you all, merging.

mkcor:
@emmanuelle I understand that the CI failure is resolved by #5109, right?

emmanuelle:
@mkcor by #5103 (which itself is fixed by #5109)

jni:

Issue #4971, opened 2020-09-11
Write a piece of documentation on data.download_all() in the install procedure


sciunto:
## Description

In https://scikit-image.org/docs/stable/install.html
write about https://scikit-image.org/docs/stable/api/skimage.data.html#download-all

jni:
Good idea @sciunto! =)

One idea is whether it would be possible to add an optional extra, `pip install scikit-image[data]`, that runs download all after installation. That would make it a very nice user experience I think!

sciunto:
That would be cool. I'm searching how to achieve this, and I found this discussion that seems to indicate it is not possible: https://stackoverflow.com/questions/55737009/how-to-get-extras-from-pip-install-in-custom-cmdclass-command

My understanding is that to do what you suggest, data must be a package.

Have you ever done that @jni?

sciunto:
All I'm able to do is to download file with `python setup.py ...`.

jni:
@sciunto ah, very sad. No, I had not done it, I just thought it might be possible with setup.py.

emmanuelle:
That said, moving our data to a skimage_data package and having it an `extras_require` dependency of `skimage["data"]` could also be a solution for #4964. But then in the setup.py of `skimage_data`, would we be able to call `download_all`?

PR #5105, opened 2020-12-05
Add data optional reqs to allow pip install scikit-image[data]

### LINKED ISSUE #4890: Sporadic pooch failures in CI
## Description

Sometimes, the hash of the downloaded files doesn't match the expected hash, see e.g.

https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=3987&view=logs&j=bc76316b-1dff-5b05-0ff1-0b85ce9c0f8b&t=dc06b1ce-f667-542e-9f00-54b6784edf0a&l=6713

Other builds in the same group passed fine, so I don't think that this is an issue with the data. Rather, it seems pooch doesn't know to retry if it's failed. (Or even if it's failed.) @leouieda do you think this is something that pooch should handle, or should we somehow wrap it to include retries?

Incidentally maybe this means the earlier errors we were seeing had nothing to do with github vs gitlab bandwidth. 😬 Or, they did but we are hitting gitlab bandwidth limits also. I don't know, this is hard. 😭

jni:
## Description

This also bumps pooch to 1.3.0 and uses retry_if_failed=3.

Closes #4890 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


jni:
@scikit-image/core I'd like to get this in and backported to 0.18 before 0.18.0, since one of our headline features is that we ship with many new datasets — but not without installing pooch first.

stefanv:
What is this for, and where is that documented and advertised for users?

jni:
If you `pip install scikit-image==0.18.0rc0`, then do `data.eagle()`, you get a runtime error because you don't have pooch installed. Same for `data.cells3d`, `data.skin`, `data.vortex`, and probably a growing number of such datasets. This lets you install `pip install scikit-image[data]` so you can get those seamlessly. I think it's nicer than telling people to `pip install scikit-image pooch`.

It will be documented in the release notes, but I want to merge all the 0.18 PRs first and then update the release notes in one go. Currently the release notes are not in master (see #5106), so updating them here is tricky to say the least.

sciunto:
https://github.com/scikit-image/scikit-image/issues/4971

emmanuelle:
The 3.8 failure is a Heisenbug and the pre failures have been addressed in #5301. @jni how can you test that `pip install scikit-image[data]` works from your branch? Is the answer "it's not possible before it's on pypi" ?

emmanuelle:
>  @jni how can you test that `pip install scikit-image[data]` works from your branch? Is the answer "it's not possible before it's on pypi" ?

OK I didn't know that 
```python
pip3 install scikit-image[optional]
```
worked, and it does (just tested it). Pretty convenient! So I'm now confident that this PR will work.

One thing is that people might expect that `pip install scikit-image[data]` will also download the data (whereas it just installs pooch). But I think it's ok as long as we document this at some point :-).

emmanuelle (approved):

jni:
> One thing is that people might expect that pip install scikit-image[data] will also download the data (whereas it just installs pooch). But I think it's ok as long as we document this at some point :-).

Yeah, @sciunto investigated this at some point and found that it was more or less impossible currently. 

@emmanuelle also,

> how can you test that pip install scikit-image[data] works from your branch? Is the answer "it's not possible before it's on pypi" ?

You can also do `pip install -e .[data]` for a local dir and it should work! Even for installing from github it works:

```
pip install -U git+https://github.com/jni/scikit-image.git@add-data-optional-reqs#egg=scikit-image[data]
```

grlee77:
> One thing is that people might expect that pip install scikit-image[data] will also download the data (whereas it just installs pooch). But I think it's ok as long as we document this at some point :-).

agreed. From the issue title, I thought this was going to be downloading the data. I still think it is worth adding, though.

grlee77 (approved):

jni:
Thanks @emmanuelle and @grlee77 for review! I'll work on docs in some follow-up PRs after 0.18.0!

stefanv:
I'm confused by this PR; why isn't pooch just listed as a dependency of scikit-image?

jni:
@hmaarrfk voiced valid concerns about having scikit-image, an image processing library, now depend on the whole web stack, just to download sample data (which many users don't touch, especially in production). I agree with that decision tbh.

stefanv:
The whole "web stack" being, in this case, `requests`?

jni:
Haha indeed. But also whatever they might (decide to) depend on.

jni:
Also `packaging` and `appdirs`. iirc at napari we've had issues with the appdirs dependency in the past.

https://github.com/fatiando/pooch/blob/master/requirements.txt

stefanv:
So, instead of depending on a tiny handful of packages (currently *one*), we introduce a) breakage in default installs and b) abstract notion that most newcomers have no idea about.

I feel strongly negative about this call.

rfezzani:
I like the idea of restricting `skimage` dependencies as much as possible, so I am :+1: with making `pooch` optional.

But I understand @stefanv unwillingness since an extra manipulation is required (`pip install pooch` or `pip install scikit-image[data]`) to have access to the new available examples (`data.eagle`, `data.cells3d`, `data.skin`, `data.vortex`).

All these considerations can be avoided once the whole `skimage.data` submodule becomes optional :wink: !

stefanv:
A good solution here would have been for `scikit-image[minimal]` to give a stripped down skimage, but this is [not currently possible with setuptools](https://github.com/pypa/setuptools/issues/1139).

My main concern is with newcomers, who may install e.g. Anaconda, and not know about packaging.  It is these users who want to run example galleries, and now can't without re-installing (which they don't necessarily know how to do).

@jni Pointed out that this has been "broken" for all of the 0.17 release already.  So, if we haven't had any complaints during that cycle, it softens my concern.

rgommers:
> My main concern is with newcomers, who may install e.g. Anaconda, and not know about packaging. It is these users who want to run example galleries, and now can't without re-installing (which they don't necessarily know how to do).

This seems like a matter of getting Anaconda to include `pooch` in its default set. Did anyone make an attempt at this, or has there been any user confusion since this was merged?

jni:
Hello @rgommers! I'd say user confusion has been minimal, but it would nevertheless be a good plan to try to get pooch into Anaconda. (So long as no animals are harmed in the making of these distributions! 😂) If there is a place for me to vote for its inclusion, count me in!

rgommers:
This is the place for feature/package requests: https://github.com/ContinuumIO/anaconda-issues/

jni:
@rgommers do you know which other libraries in our ecosystem use pooch? Also, what brought you to this PR in the first place?

PR #5107, opened 2020-12-05
Fix compilation error in XCode 12

### LINKED ISSUE #5051: pip install error: implicit declaration of function 'unwrap2D' is invalid in C99
## Description
I'm getting an installation error on MacOS using pip

## Way to reproduce
```python
pip install scikit-image
```

## Version information
```python
3.9.0 (default, Oct 27 2020, 14:15:17) 
[Clang 12.0.0 (clang-1200.0.32.21)]
macOS-10.15.7-x86_64-i386-64bit
numpy version: 1.19.4
```

skimage/restoration/_unwrap_2d.c:2285:9: error: implicit declaration of function 'unwrap2D' is invalid in C99 [-Werror,-Wimplicit-function-declaration]

See the attached log file the for the full dump:
[scikit_install_log.txt](https://github.com/scikit-image/scikit-image/files/5492160/scikit_install_log.txt)

jni:
## Description

Closes #5051 

@david-a-joy thank you for the fix! Could you check that this branch compiles as-is?

Does anyone have suggestions for how we can test this? Is there a flag we can pass to our older compilers to make them error out on the missing declarations?

At any rate even without a test I think we should merge this, since we have seen many people hitting this issue.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/restoration/_unwrap_2d.pyx
+++ b/skimage/restoration/_unwrap_2d.pyx
@@ -11,6 +11,18 @@ cdef extern from *:
                   int wrap_around_x, int wrap_around_y,
                   char use_seed, unsigned int seed) nogil
 
+
+cdef extern from "unwrap_2d_ljmu.h":
+    void unwrap2D(
+            double *wrapped_image,
+            double *UnwrappedImage,
+            unsigned char *input_mask,
+            int image_width, int image_height,
+            int wrap_around_x, int wrap_around_y,
+            char use_seed, unsigned int seed
+            )
+
+
 def unwrap_2d(double[:, ::1] image,
               unsigned char[:, ::1] mask,
               double[:, ::1] unwrapped_image,
--- a/skimage/restoration/_unwrap_2d.pyx
+++ b/skimage/restoration/_unwrap_2d.pyx
@@ -3,14 +3,6 @@
 # cython: nonecheck=False
 # cython: wraparound=False
 
-cdef extern from *:
-    void unwrap2D(double* wrapped_image,
-                  double* unwrapped_image,
-                  unsigned char* input_mask,
-                  int image_width, int image_height,
-                  int wrap_around_x, int wrap_around_y,
-                  char use_seed, unsigned int seed) nogil
-
 
 cdef extern from "unwrap_2d_ljmu.h":
     void unwrap2D(
@@ -20,7 +12,7 @@ cdef extern from "unwrap_2d_ljmu.h":
             int image_width, int image_height,
             int wrap_around_x, int wrap_around_y,
             char use_seed, unsigned int seed
-            )
+            ) nogil
 
 
 def unwrap_2d(double[:, ::1] image,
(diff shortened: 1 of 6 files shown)


hmaarrfk (approved):

jni:
The `--pre` failures are addressed in #5103

david-a-joy:
Oops, looks like we also need to include the new headers in the relevant cython files:
 
"_unwrap_2d.pyx":
> cdef extern from "unwrap_2d_ljmu.h":
>      void unwrap2D(double* wrapped_image,
>                    double* unwrapped_image,
>                    unsigned char* input_mask,

"_unwrap_3d.pyx":
> cdef extern from "unwrap_3d_ljmu.h":
>    void unwrap3D(double* wrapped_volume,
>                  double* unwrapped_volume,
>                  unsigned char* input_mask,

With that, everything compiles successfully on cython 0.29.21 and cython 3.0a6 (Mac OS Catalina, Python 3.9.0, clang 12.0.0). I get what I think are unrelated test failures ([errors_cython0-29.txt](https://github.com/scikit-image/scikit-image/files/5647643/errors_cython0-29.txt)), but everything in the restoration module passes.

---

As far as testing, both clang and gcc should take "-std=c99" flag that should reproduce the failure (no idea about msvc, a quick google suggests maybe not?). I tried to naively add that switch to extra_compile_args in "setup.py" but it breaks when compiling stuff with clang++ so I think something fancier is needed.

jni:
Thanks @david-a-joy! For some reason I'm having trouble getting my local compilation working for unrelated reasons, so I haven't been able to test it out. 😬

emmanuelle:
@david-a-joy any idea of why the compilation fails on all jobs?

david-a-joy:
I think the cdefs got added to the wrong files? They go in the Cython "_unwrap_2d.pyx" and "_unwrap_3d.pyx" files not in the C "unwrap_2d_ljmu.c" and "unwrap_3d_ljmu.c" files...

emmanuelle:
Thanks @david-a-joy ! So should the cdefs replace the existing blocks like

```python
cdef extern from *:
    void unwrap2D(double* wrapped_image,
                  double* unwrapped_image,
                  unsigned char* input_mask,
                  int image_width, int image_height,
                  int wrap_around_x, int wrap_around_y,
                  char use_seed, unsigned int seed) nogil
```

or should they be added to the file, leaving the existing ones?

david-a-joy:
They should replace them. The old line 

> cdef extern from *:

means that the functions are assumed to be defined in a header already included elsewhere (say if there was a "skimage.h" that had the headers for all c functions used by scikit-image). We don't have that, so cython needs to be pointed to the header files explicitly, i.e:

> cdef extern from "unwrap_2d_ljmu.h":

I don't know what happens if you try to do two cdef extern blocks that define the same function, but my guess is that it will make cython mad.

emmanuelle:
Thanks a lot for your help @david-a-joy ! I'm trying to get this PR to the finish line while Australia sleeps, so that Juan can concentrate on other things for the release :-).

jni:
Awesome, thanks @emmanuelle for picking up the slack! If anyone else from @scikit-image/core could merge this (failure is what @emmanuelle fixed in #5109), it would be much appreciated! 🙏

grlee77:
Looks good. Thanks!

PR #5109, opened 2020-12-06
revert bool back to np.bool_ for np.issubdtype calls

emmanuelle:
Follow-up of #5103 in order to fox the test failures with minimal requirements (numpy 1.16.5)


jni (approved):
Test failure is a pooch failure I think (misread numpy array). Thank you for fixing this @emmanuelle!

alexdesiqueira:
Thank you @emmanuelle!

PR #5114, opened 2020-12-08
Deprecate high_threshold and low_threshold in canny

rfezzani:
## Description

This PR implements my first [suggestion](https://github.com/scikit-image/scikit-image/pull/3070#issuecomment-583325610) concerning #3070 to help passing a callable to select the thresholds.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


stefanv:
@rfezzani What is the motivation behind unifying the two keywords?  Aesthetically, this may be a bit cleaner---but it is still an API change.

jni:
I agree with @stefanv, I don't think this deprecation is worthwhile. Instead, I suggest only making high_threshold and low_threshold *optional*, and adding a new keyword-only `thresholds` that can be a callable.

rfezzani:
> Instead, I suggest only making high_threshold and low_threshold optional, and adding a new keyword-only thresholds that can be a callable.

It is indeed an other option :smile:. I essentially wanted to move forward #3070, which is an excellent addition BTW I think :wink:

PR #5115, opened 2020-12-08
change plausible script to track outbound links

emmanuelle:
A small change to the plausible.io javascript script; see https://docs.plausible.io/outbound-link-click-tracking for the explanation.

This would in particular allow to see whether people click on the user forum links.


jni (approved):

jni:

PR #5117, opened 2020-12-09
Remove Python 3.6 support

### LINKED ISSUE #5053: dropping 3.6 and a few other things

jni:
## Description

Closes #5053, see discussion there.

Currently, if we release 0.18 without this PR, users on py3.6 will need to compile from source, since we are not building 3.6 wheels. Given the discussion in #5053 and elsewhere in the community, I think we should pull the bandaid and remove 3.6 support altogether from this release.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,6 @@ requires = [
     "Cython>=0.29.18",
     # We follow scipy for much of these pinnings
     # https://github.com/scipy/scipy/blob/master/pyproject.toml
-    "numpy==1.16.5; python_version=='3.6'",
     "numpy==1.16.5; python_version=='3.7'",
     "numpy==1.17.3; python_version=='3.8'",
     # do not pin numpy on future versions of python to avoid incompatible numpy and python versions
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ requires = [
     # https://github.com/scipy/scipy/blob/master/pyproject.toml
     "numpy==1.16.5; python_version=='3.7'",
     "numpy==1.17.3; python_version=='3.8'",
+    "numpy==1.19.3; python_version=='3.9'",
     # do not pin numpy on future versions of python to avoid incompatible numpy and python versions
-    "numpy; python_version>='3.9'",
+    "numpy; python_version>='3.10'",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [build-system]
 requires = [
     "wheel",
-    "setuptools",
+    "setuptools<=51.0.0",
     "Cython>=0.29.18",
     # We follow scipy for much of these pinnings
     # https://github.com/scipy/scipy/blob/master/pyproject.toml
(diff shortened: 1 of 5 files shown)


emmanuelle (approved):
ok, let's do it!

pyproject.toml
```diff
@@ -5,7 +5,6 @@ requires = [
     "Cython>=0.29.18",
     # We follow scipy for much of these pinnings
     # https://github.com/scipy/scipy/blob/master/pyproject.toml
-    "numpy==1.16.5; python_version=='3.6'",
     "numpy==1.16.5; python_version=='3.7'",
     "numpy==1.17.3; python_version=='3.8'",
```
emmanuelle, line 9: shouldn't we add 3.9 to this file? Not directly related to the PR of course.

jni: It is added below under a comment, I did wonder the same thing. 😊 I must say I don't understand why it's not pinned but I presume the scipy folks know what they're doing!

hmaarrfk: you might as well pin to 1.19.4 since that is the only current version that has a build on pypi

hmaarrfk: and leave it like that for pythyon 3.10

grlee77: There is a pin to 1.19.3 for 3.9 in https://github.com/scipy/oldest-supported-numpy. My understanding from the about section of that package is that we could just include `oldest-supported-numpy` within our pyproject.toml instead of maintaining all these special cases ourselves.

grlee77: would have to check with the scipy team about why we don't use it in SciPy itself, though

grlee77: There is an open PR at https://github.com/scipy/scipy/pull/13217 adding the Python 3.9 pin in SciPy itself. See also about pinning setuptools<=51.0 in that issue.

hmaarrfk: Oldest supported numpy doesn't adhere to nep29. 

https://github.com/scipy/oldest-supported-numpy/issues/11

jni: Thanks, both. I don't understand the story with setuptools (is `<=51.0.0` necessary rather than `<=51`?), nor why they are only doing this on the 1.6 branch, but I've implemented the above-linked changes. Please merge or modify as you see fit...

hmaarrfk:
Great!

jni:

PR #5120, opened 2020-12-09
example to show the 3d dataset

emmanuelle:
This new gallery example shows the 3d dataset which has been recently introduced.

See for example https://1138-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/data/plot_3d.html#sphx-glr-auto-examples-data-plot-3d-py

PR #5126, opened 2020-12-10
Proofread gallery example for rank filters.

mkcor:
Follows from https://github.com/scikit-image/scikit-image/pull/5120#discussion_r540132671.

## Description

Some links to the API reference and other docs were broken or outdated.

I removed the last part of a sentence ("e.g. salt and pepper noise typically does not distort the gray-level") because I didn't understand it. I'm happy to re-add it, maybe as a separate sentence, if someone can explain it to me.

I also removed an extra `[1]_` pointer.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_rank_filters.py
+++ b/doc/examples/applications/plot_rank_filters.py
@@ -97,7 +97,13 @@ plt.tight_layout()
 # pixel wide), a small filter radius is sufficient. As the radius
 # increases, objects with bigger sizes get filtered as well, such as the
 # camera tripod. The median filter is often used for noise removal because it
-# preserves borders.
+# preserves borders. For example, consider noise which is located only on a
+# few pixels in the entire image, as is the case with salt-and-pepper noise
+# [2]_: the median filter will ignore the noisy pixels, for they will appear
+# as outliers; thus, it will not change significantly the median of a group of
+# local pixels, in contrast to what a moving average filter would do.
+#
+# .. [2] https://en.wikipedia.org/wiki/Salt-and-pepper_noise
 #
 # Image smoothing
 # ===============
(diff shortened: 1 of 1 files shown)


hmaarrfk:
https://en.wikipedia.org/wiki/Salt-and-pepper_noise

mkcor:
@hmaarrfk thanks, but my lack of understanding isn't about salt-and-pepper per se (by now I've stumbled across it more than once in the scikit-image docs, so I've had opportunities to look it up). It's about the whole phrasing (parsing). Currently, the whole sentence reads: "The median filter is often used for noise removal because borders are preserved and e.g. salt and pepper noise typically does not distort the gray-level." As an independent proposition, I would fix it into "Salt-and-pepper noise typically does not distort the gray-level histogram." But I'm not sure how it's supposed to work with the previous proposition.

doc/examples/applications/plot_rank_filters.py
```diff
@@ -93,12 +93,11 @@
...
-# pixel wide), a small filter radius is sufficient. As the radius is
-# increasing, objects with bigger sizes are filtered as well, such as the
-# camera tripod. The median filter is often used for noise removal because
-# borders are preserved and e.g. salt and pepper noise typically does not
-# distort the gray-level.
+# The added noise is efficiently removed, as the image defaults are small (1-\
+# pixel wide), a small filter radius is sufficient. As the radius
+# increases, objects with bigger sizes get filtered as well, such as the
+# camera tripod. The median filter is often used for noise removal because it
+# preserves borders.
```
hmaarrfk, line 100: Maybe something like

Maybe something like:

For example, consider noise that is located on only a few pixels in the entire image, as is the case for salt and pepper noise (reference to wikipedia), the median filter will ignore the noisy pixels since they will appear as outliers, and thus not significantly change the median of a group of local pixels in contrast to a moving average filter.

mkcor: I see. Thanks a lot, @hmaarrfk, I'll include your version.

mkcor: @hmaarrfk check out 67d2aaac340709f7 ;-)

jni (approved):

jni:

PR #5143, opened 2020-12-16
Fix large arrays labelling

### LINKED ISSUE #5124: Possible bug in label connected regions in large volume crushes all kernel
## Description
I have used several times the label function of scikit but this is the largest volume I have used so far. I tried the label with connectivity = 3 but the kernel was crushing repeatly, then changed it to connectivity =2 and after taking almost 80 GB of RAM was able to throw a result. Then I compaired it with the label feature of ndimage library, it finished faster and not taking even 20 GB of RAM. Out of curiosity, visualize both of them using the napari library.

In the video below the tag "labels[1]" is the result of the ndimage label, while "label" tag is the scikit-image label output.
https://youtu.be/waXdlIVgk2M
I uploaded to youtube to make it more accesible

## Way to reproduce

#LIBRARIES
### work in jupyterlab
import Pred_tomo as ptom
import pandas as pd
import numpy as np
from skimage.measure import regionprops
from pathlib import Path
from PIL import Image
import napari
from scipy.ndimage import label, generate_binary_structure
%gui qt5

## Code to open a tif volume as numpy array from path..etc I get a numpy array called "prob_l"
prob_l.shape          #output (7443, 1771, 234)

## Compute the labels with the 2 libraries
import skimage.measure
skimage_labels_p_l = skimage.measure.label(prob_l,connectivity=2)
labels_p_l = label(prob_l)  #NDimage label

##### ptom.axis_napari is a simple func I defined to have the same coordinate system in napari visualizer than Imagej.
'''
def axis_napari(array):
    array = np.transpose(array, (2,0,1))
    return(array)
'''
viewer = napari.view_labels(ptom.axis_napari(skimage_labels_p_l))
viewer.add_labels(ptom.axis_napari(labels_p_l[0]))

## Version information
```python
3.7.6 (default, Jan  8 2020, 20:23:39) [MSC v.1916 64 bit (AMD64)]
Windows-10-10.0.18362-SP0
scikit-image version: 0.16.2
numpy version: 1.18.1
```

rfezzani:
## Description

Fixes #5124 by replacing `long` by `size_t` to prevent overflow in array indexing.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_ccomp.pyx
+++ b/skimage/measure/_ccomp.pyx
@@ -15,7 +15,7 @@ cdef DTYPE_t BG_NODE_NULL = -999
 cdef struct s_shpinfo
 
 ctypedef s_shpinfo shape_info
-ctypedef long (* fun_ravel)(long, long, long, shape_info *) nogil
+ctypedef size_t (* fun_ravel)(size_t, size_t, size_t, shape_info *) nogil
 
 
 # For having stuff concerning background in one place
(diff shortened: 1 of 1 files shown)


jni (approved):
Good thinking @rfezzani!

alexdesiqueira:
Thank you @rfezzani!

jni:

PR #5145, opened 2020-12-17
label2rgb performance fix

grlee77:
## Description

I came across an inappropriate call to built-in `max` rather than `numpy.max` on arrays. This is slower on NumPy arrays, but where it would really cause problems is if using CuPy or device arrays because it will lead to separate host->device transfers for each element!

Here is an example illustrating ~ 200 fold performance difference for NumPy arrays on the CPU

```Python
import numpy as np

a = np.arange(10000000)

%timeit np.sum(a)
# gives: 6.62 ms ± 64.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit sum(a)
# gives: 1.25 s ± 42 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/colorlabel.py
+++ b/skimage/color/colorlabel.py
@@ -188,7 +188,7 @@ def _label2rgb_overlay(label, image=None, colors=None, alpha=0.3,
     if len(mapped_labels_flat) == 0:
         return image
 
-    dense_labels = range(max(mapped_labels_flat) + 1)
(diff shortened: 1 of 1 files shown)


jni (approved):

rfezzani:
Thank you @grlee77 :wink:

PR #5156, opened 2020-12-23
Add saturation parameter to color.label2rgb

charlielito:
## Description

The function `skimage.color.label2rgb` transforms always the input image to grayscale. There are cases where one wants to keep the image as it is (rgb). I added a flag `force_grayscale` to control this behavior. What do you think? I know the name could be improved :P

### DIFF
--- a/skimage/color/colorlabel.py
+++ b/skimage/color/colorlabel.py
@@ -74,7 +74,7 @@ def _match_label_with_color(label, colors, bg_label, bg_color):
 @change_default_value("bg_label", new_value=0, changed_version="0.19")
 def label2rgb(label, image=None, colors=None, alpha=0.3,
               bg_label=-1, bg_color=(0, 0, 0), image_alpha=1, kind='overlay',
-              saturation=0):
+              *, saturation=0):
     """Return an RGB image where color-coded labels are painted over the image.
 
     Parameters
@@ -161,6 +161,9 @@ def _label2rgb_overlay(label, image=None, colors=None, alpha=0.3,
         The result of blending a cycling colormap (`colors`) for each distinct
         value in `label` with the image, at a certain alpha value.
     """
+    if not 0 <= saturation <= 1:
+        warn(f"saturation must be in range [0, 1], got {saturation}")
+
     if colors is None:
         colors = DEFAULT_COLORS
     colors = [_rgb_vector(c) for c in colors]
--- a/skimage/color/colorlabel.py
+++ b/skimage/color/colorlabel.py
@@ -73,7 +73,8 @@ def _match_label_with_color(label, colors, bg_label, bg_color):
 
 @change_default_value("bg_label", new_value=0, changed_version="0.19")
 def label2rgb(label, image=None, colors=None, alpha=0.3,
-              bg_label=-1, bg_color=(0, 0, 0), image_alpha=1, kind='overlay'):
+              bg_label=-1, bg_color=(0, 0, 0), image_alpha=1, kind='overlay',
+              force_grayscale=True):
     """Return an RGB image where color-coded labels are painted over the image.
 
     Parameters
@@ -102,6 +103,9 @@ def label2rgb(label, image=None, colors=None, alpha=0.3,
         and overlays the colored labels over the original image. 'avg' replaces
         each labeled segment with its average color, for a stained-class or
         pastel painting appearance.
+    force_grayscale : bool, optional
+        Whether or not to convert the input image to grayscale. By default it 
+        converts the image to grayscale before making the overlay.
 
     Returns
     -------
@@ -111,7 +115,7 @@ def label2rgb(label, image=None, colors=None, alpha=0.3,
     """
     if kind == 'overlay':
         return _label2rgb_overlay(label, image, colors, alpha, bg_label,
-                                  bg_color, image_alpha)
+                                  bg_color, image_alpha, force_grayscale)
     elif kind == 'avg':
         return _label2rgb_avg(label, image, bg_label, bg_color)
     else:
@@ -119,7 +123,8 @@ def label2rgb(label, image=None, colors=None, alpha=0.3,
 
 
 def _label2rgb_overlay(label, image=None, colors=None, alpha=0.3,
-                       bg_label=-1, bg_color=None, image_alpha=1):
+                       bg_label=-1, bg_color=None, image_alpha=1,
+                       force_grayscale=True):
     """Return an RGB image where color-coded labels are painted over the image.
 
     Parameters
@@ -142,6 +147,9 @@ def _label2rgb_overlay(label, image=None, colors=None, alpha=0.3,
         between [0, 1].
     image_alpha : float [0, 1], optional
         Opacity of the image.
+    force_grayscale : bool, optional
+        Whether or not to convert the input image to grayscale. By default it 
+        converts the image to grayscale before making the overlay.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


jni:
Thanks for the contribution @charlielito! I agree that we could work a bit on the parameter name but otherwise I think this is worth adding... Maybe there's a better way, e.g. adding a new mode? I wonder if anyone from @scikit-image/core has any suggestions?

jni:
e.g. `mode='overlay-rgb'`?

mkcor:
> e.g. `mode='overlay-rgb'`?

Do you mean a new `kind`? If so, yes, I would welcome this addition:
```py
kind : string, one of {'avg', 'overlay', 'overlay-rgb'}
```

mkcor:
> `kind='overlay-rgb'`

Dear @charlielito, are you happy to make the proposed changes or would you like us to take over from here?

charlielito:
> > `kind='overlay-rgb'`
> 
> Dear @charlielito, are you happy to make the proposed changes or would you like us to take over from here?

I'll make the changes and also make the tests pass :) I'll come later to you folks, thanks for the recommendation

mkcor:
@charlielito wonderful! Thank you so much again :smile:

charlielito:
@mkcor @jni already made the changed you proposed with test

jni (approved):
Great, lgtm now @charlielito! Thank you for this improvement!

grlee77 (changes requested):
Thanks, this looks pretty good. I have just a couple of minor docstring comments and one question/suggestion about avoiding calling `gray2rgb` on RGB or RGBA inputs.

skimage/color/colorlabel.py
```diff
@@ -164,11 +173,13 @@ def _label2rgb_overlay(label, image=None, colors=None, alpha=0.3,
         if image.min() < 0:
             warn("Negative intensities in `image` are not supported")
 
-        if image.ndim > label.ndim:
-            image = img_as_float(rgb2gray(image))
+        image = img_as_float(image)
+        if image.ndim > label.ndim and not preserve_rgb:
+            image = rgb2gray(image)
+            image = gray2rgb(image)
         else:
```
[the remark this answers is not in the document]
charlielito: Yes, I didn't see that case

mkcor (commented):
Thank you, @charlielito! I have left several suggestions for you to include.

stefanv:
Why not add a desaturation parameter, by which you can control where between fully saturated (original RGB) and fully desaturated (gray image) the drawn upon image is?

jni:
> Why not add a desaturation parameter

It's a very good suggestion tbh, though it's a fair bit more work. @charlielito what do you think about it, and do you have the bandwidth to implement it? We can finish the PR for you if you are busy now — I acknowledge that this PR has dragged on for a while now, even though there was the Christmas break in the middle.

Anyway, to implement it I think we would:
- go back to just 'overlay' and 'average',
- add a 'saturation=0.' parameter with values between 0. and 1.
- then, instead of using rgb2gray, we would do rgb2hsv, then multiply the 1st channel (s for saturation) by the given saturation value, then convert back with hsv2rgb.

To reiterate, if you want to apply those changes, that'd be great, if you're too busy, we can push this over the line.

Thank you!

skimage/color/tests/test_colorlabel.py
```diff
@@ -206,8 +206,9 @@ def test_bg_color_rgb_string():
     labels = np.zeros((10, 10), dtype=np.int64)
     labels[1:3, 1:3] = 1
     labels[6:9, 6:9] = 2
-    output = label2rgb(labels, image=img, alpha=0.9, bg_label=0, bg_color='red')
-    assert output[0, 0, 0] > 0.9 # red channel
+    output = label2rgb(labels, image=img, alpha=0.9,
+                       bg_label=0, bg_color='red')
+    assert output[0, 0, 0] > 0.9  # red channel
```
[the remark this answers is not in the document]
charlielito: `black` formating :sweat_smile:

stefanv:
Nice work, @charlielito!  I think you can save a bit of memory by using `hsv[..., 1] *= saturation` instead of the current form, but otherwise this is great.

stefanv (approved):

skimage/color/tests/test_colorlabel.py
```diff
@@ -216,3 +217,22 @@ def test_avg_with_2d_image():
...
+                    bg_label=0, saturation=1)
+    # check that rgb part of input image is preserved, where labels=0
+    assert_array_almost_equal(rgb_img[:3, :3] * (1 - alpha), rgb[:3, :3])
+
+    # now check with saturation=0 that does not match with rgb values
+    rgb = label2rgb(labels, image=rgb_img, alpha=alpha,
+                    bg_label=0, saturation=0)
+
+    with testing.raises(AssertionError):
+        assert_array_almost_equal(rgb_img[:3, :3] * (1 - alpha), rgb[:3, :3])
```
rfezzani, line 238:
```suggestion
    assert_array_almost_equal(rgb_img[:3, :3] * (1 - alpha), rgb[:3, :3])
```

skimage/color/colorlabel.py
```diff
@@ -73,7 +73,8 @@ def _match_label_with_color(label, colors, bg_label, bg_color):
 
 @change_default_value("bg_label", new_value=0, changed_version="0.19")
 def label2rgb(label, image=None, colors=None, alpha=0.3,
-              bg_label=-1, bg_color=(0, 0, 0), image_alpha=1, kind='overlay'):
+              bg_label=-1, bg_color=(0, 0, 0), image_alpha=1, kind='overlay',
+              saturation=0):
```
rfezzani, line 77: Shouldn't `saturation` be a keyword only argument?
```suggestion
              *, saturation=0):
```

rfezzani (commented):
It looks good, thank you @charlielito! I just left some minor comments

skimage/color/tests/test_colorlabel.py
charlielito: The `with testing.raises(AssertionError):` is needed here, because here the we are testing that the arrays are not equal

skimage/color/colorlabel.py
charlielito: Why?

skimage/color/tests/test_colorlabel.py
rfezzani: You are right, but this test looks weird then :confused:.
Can't we refactor it in a separate function `test_zero_saturation` asserting that when `saturation` is 0 the returned image is the gray level values where `labels`=0?

skimage/color/colorlabel.py
rfezzani: Sorry, this question was for @scikit-image/core :). We usually make new arguments keyword only...

skimage/color/tests/test_colorlabel.py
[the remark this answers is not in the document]
mkcor: Oh, ok! :wink:

skimage/color/colorlabel.py
mkcor: Yes, I think `saturation` should be a keyword-only argument. And should we raise a warning if it is used while `kind='avg'?

I would also expect a code unit later on to raise an error or, at least, a warning if `saturation < 0` or `saturation > 1`.

skimage/color/tests/test_colorlabel.py
charlielito: Sure, this makes sense

skimage/color/colorlabel.py
charlielito: Just added a warning to check the saturation correct range (also a unit test) and `saturation` to be a keyword-only argument

charlielito: Other parameters like `image_alpha`, `alpha`, etc are not used when `kind='avg'` and there is no warning

jni:
Thank you @charlielito! 🎉

rfezzani:
Excellent, thank you @charlielito :wink:!

PR #5162, opened 2020-12-28
WIP: Performance improvements for Canny feature detection

grlee77:
## UPDATE

I have marked this as WIP due to a similar preceding PR at #4342. It probably makes more sense to merge this PR into that one.

## Description

This PR makes multiple independent changes to the Canny feature detection code to improve performance. It will be easiest to review the commits in isolation as they each modify independent things. The table below shows the benchmark result for each commit here.

time_canny (ms) | commit
-----------|-------
1160   | 59889f5 (same as v0.18.1)
1090   | 769235e
1020   | 5757c11
988     | 5ab8429
791     | a69f6a8
22 | CuPy-based implementation equivalent to a69f6a8 (not in this PR)

Commits 5ab8429 and a69f6a8 only help for the case when `mask=None`. Overall, the reduction in runtime for the benchmark is 32% in the unmasked case and 12% in the masked case.

Commit 5757c11 gives only a minor improvement here, but makes an order of magnitude difference if using CuPy for the array module (`cupyx.scipy.ndimage.sum` is not very efficient for a large number of very small labeled regions). The poor performance of this function in a CuPy-based implementation was the reason I started looking at performance here in the first place. 

Commit a69f6a8 provides a substantial benefit, but this commit does result in small differences near the image edge as generated via results from the following script. Qualitatively, they look equivalent to me, but the results are not pixelwise identical to before near the edge.

```Python
import numpy as np
from skimage import color, data, feature, util

image = color.rgb2gray(data.astronaut())
result = feature.canny(image, sigma=1.5, use_quantiles=True)
```
canny_compare

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/benchmarks/benchmark_feature.py
+++ b/benchmarks/benchmark_feature.py
@@ -2,20 +2,16 @@
 # https://asv.readthedocs.io/en/latest/writing_benchmarks.html
 import numpy as np
 from scipy import ndimage as ndi
-from skimage import feature, util
+from skimage import color, data, feature, util
 
 
 class FeatureSuite:
     """Benchmark for feature routines in scikit-image."""
     def setup(self):
-        self.image = np.zeros((640, 640))
-        self.image[320:-320, 320:-320] = 1
-
-        self.image = ndi.rotate(self.image, 15, mode='constant')
-        self.image = ndi.gaussian_filter(self.image, 4)
-        self.image += 0.2 * np.random.random(self.image.shape)
-
-        self.image_ubyte = util.img_as_ubyte(np.clip(self.image, 0, 1))
+        # Use a real-world image for more realistic features, but tile it to
+        # get a larger size for the benchmark.
+        self.image = np.tile(color.rgb2gray(data.astronaut()), (4, 4))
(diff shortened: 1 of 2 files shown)


grlee77:
The CI failure is unrelated.

rfezzani:
Hi @grlee77, I think that I implemented some of the improvement proposed in this PR in #4342, plus a global refactoring that improves code readability I think. May be can we merge these two PRs?

grlee77:
> Hi @grlee77, I think that I implemented some of the improvement proposed in this PR in #4342, plus a global refactoring that improves code readability I think. May be can we merge these two PRs?

Thanks for pointing out that PR. I had not noticed it previously, but should have done a search here first! I do like the clarity of the larger refactoring in yours. Let me review it and then see about adding any needed commits to your branch (it does not appear to have a corresponding change to 5757c11, which was the key change here on the GPU side).

rfezzani:
:+1: Yes I noticed your excellent use of `np.unique` and wanted to implement it in #4342 :wink:

grlee77:
I have now added the performance improvements here to #4342, so will close this PR.

PR #5163, opened 2020-12-30
Single out docs-only PRs in review process.

mkcor:
## Description

@grlee77 updating the guidelines as per https://github.com/scikit-image/scikit-image/pull/5120#issuecomment-752209390.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


emmanuelle:
Thanks @mkcor! I'm wondering whether to merge your PR as is, or to specify "non-controversial pull requests" (meaning that for a long doc PR with a lot of stuff, it is still better to have two reviews). Thoughts?

mkcor:
> Thanks @mkcor! I'm wondering whether to merge your PR as is, or to specify "non-controversial pull requests" (meaning that for a long doc PR with a lot of stuff, it is still better to have two reviews). Thoughts?

I agree with you, @emmanuelle... It might be better not to merge this PR at all then, and keep the guidelines valid in the general case. I think it's not great to use the term "non-controversial" in a short bullet-point list, because it is subject to judgement. On the contrary, "docs-only" is objective, but we do want two reviewers if the "docs-only" PR is long and/or complex, indeed... So I would suggest using our judgement just like we've done so far. Feel free to close!

emmanuelle:
@mkcor not completely sure here, I think core devs are not always aware (or they forget) that they can merge small doc PRs they are happy with... Let's wait for others to chime in and in any case, thanks for bringing up this point.

jni:
I'm happy to go one way or the other here...

CONTRIBUTING.txt
```diff
@@ -190,8 +190,9 @@ Guidelines
   `standard <https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard>`_ as NumPy and SciPy.
 * For new functionality, always add an example to the gallery (see
   :ref:`Sphinx-Gallery<sphinx_gallery>` below for more details).
-* No changes are ever committed without review and approval by two core
-  team members.  Ask on the
+* No changes are ever merged without review and approval by two core team members.
+  There is an exception to this rule: pull requests which affect only the documentation
+  require review and approval by only one core team member. Reach out on the
```
grlee77, line 195:
```suggestion
  require review and approval by only one core team member in most cases. If the maintainer feels the changes are large or likely to be controversial, two reviews should still be encouraged. Reach out on the
```

grlee77: I suggested some text here along the lines of Emma's suggestion.

mkcor:
Sorry, this almost fell in the cracks! I have included @grlee77's suggestion, which seems consensual.

grlee77:
The other case I would encourage merging fairly quickly after a single review are minor fixes needed to restore CI to a working state (e.g. #5315, #5319 for two recent examples)

mkcor:
@grlee77 good point! Added this other case (96850b161fdd0491f024b61bd0ebbe9d5ed2748e). I really appreciate your foolproof radar! :grin:

grlee77 (approved):
Thanks for updating. This looks ready to me

jni (approved):

PR #5167, opened 2020-12-30
Fix return dtype of _label2rgb_avg function

### LINKED ISSUE #5166: DOCS: Return dtype of _label2rgb_avg function
## Description

With skimage release 0.18.0, and in particular after PR #4840, there is a mismatch between the return dtype of function `_label2rgb_avg` in the `color/colorlabel` module: it is stated to be the same dtype as the image, but using `np.zeros` instead of `np.zeros_like` makes the return type always to be float

alessiamarcolini:
## Description

Fixes #5166

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/color/colorlabel.py
+++ b/skimage/color/colorlabel.py
@@ -223,7 +223,7 @@ def _label2rgb_avg(label_field, image, bg_label=0, bg_color=(0, 0, 0)):
     out : array, same shape and type as `image`
         The output visualization.
     """
-    out = np.zeros(label_field.shape + (3,))
+    out = np.zeros(label_field.shape + (3,), dtype=image.dtype)
     labels = np.unique(label_field)
     bg = (labels == bg_label)
     if bg.any():


emmanuelle:
Thanks for noticing this @alessiamarcolini (and great to see you here :-))! I would advocate that #4840 introduced a bug when using `np.zeros` without fixing the dtype, and rather than fixing the doc, we should add `dtype=image.dtype` inside the `np.zeros` function, in order to preserve the dtype. What do you think?

alessiamarcolini:
Hi @emmanuelle, thank you for the quick reply! I agree with you, I'm going to implement your suggestion and change the code instead

jni (approved):
very nice! Do we want to add a test to make sure this is working in the future?

Looking at the code that follows, we'll probably have something like:

```python
out[mask] = \  # a uint8 array
        color  # a float64 array
```

which works fine but might be an anti-pattern. I wonder if we want to do some explicit casting, e.g.:

```python
out[mask] = np.asarray(color).astype(image.dtype, casting='unsafe')
```

This is exactly equivalent code, since `unsafe` is the default, but the intent is explicit in the code and it protects us from any potential changes in the default value in the future.

But anyway, this is already a good fix so happy to approve as-is, just adding stuff to discuss. 😊

emmanuelle:
I'm going to merge the PR as it is since it fixes a bug, and @alessiamarcolini if you want to address Juan's ideas (a test and explicit casting), feel free to open a new PR :-).

PR #5170, opened 2021-01-04
threshold_local: n-dimensional support and anisotropic block_size

grlee77:
## Description

`threshold_local` appears to be the only function in `skimage.filters.thresholding` that does not support n-dimensional inputs. I have added n-d support in this PR. This was fairly easy because the underlying `scipy.ndimage` functions used were already n-dimensional.

I was not sure about good 3d test cases, but decided to just replicate the existing 2d test cases via tiling so that the same expected results could be reused.

Allowing anisotropic `block_size` can be useful to enable `block_size = 1` for any non-spatial dimensions as an easy way to batch threshold across multiple channels or timepoints.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/thresholding.py
+++ b/skimage/filters/thresholding.py
@@ -153,8 +153,8 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
 
     Parameters
     ----------
-    image : (N, M) ndarray
-        Input image.
+    image : (N, M[, ..., P]) ndarray
+        Grayscale input image.
     block_size : int
         Odd size of pixel neighborhood which is used to calculate the
         threshold value (e.g. 3, 5, 7, ..., 21, ...).
@@ -186,7 +186,7 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
 
     Returns
     -------
-    threshold : (N, M) ndarray
+    threshold : (N, M[, ..., P]) ndarray
         Threshold image. All pixels in the input image higher than the
         corresponding pixel in the threshold image are considered foreground.
 
@@ -206,7 +206,6 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
     if block_size % 2 == 0:
         raise ValueError("The kwarg ``block_size`` must be odd! Given "
                          "``block_size`` {0} is even.".format(block_size))
-    check_nD(image, 2)
     thresh_image = np.zeros(image.shape, 'double')
     if method == 'generic':
         ndi.generic_filter(image, param, block_size,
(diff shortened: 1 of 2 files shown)


jni (approved):
Awesome! 🎉

skimage/filters/thresholding.py
```diff
@@ -203,29 +203,37 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
+        masks = {}
+        for b in set(block_size):
+            masks[b] = 1. / b * np.ones((b,))
+        # # separation of filters to speedup convolution
+        b = block_size[0]
+        ndi.convolve1d(
+            image, masks[b], axis=0, output=thresh_image, mode=mode, cval=cval)
+        for ax, b in zip(range(1, image.ndim), block_size[1:]):
+            ndi.convolve1d(thresh_image, masks[b], axis=ax,
+                           output=thresh_image, mode=mode, cval=cval)
```
rfezzani, line 236:
```suggestion
        kernels = {b: np.full(b, 1/b) for b in set(block_size)}
        # separation of filters to speedup convolution
        thresh_image = image.copy()
        for ax, size in enumerate(block_size):
            ndi.convolve1d(thresh_image, kernels[b], axis=ax,
                           output=thresh_image, mode=mode, cval=cval)
```
Can't we here directly use the [`ndi.uniform_filter`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.uniform_filter.html) that is also implemented using the kernel separation?

skimage/filters/thresholding.py
```diff
@@ -203,29 +203,37 @@ def threshold_local(image, block_size, method='gaussian', offset=0,
...
-        mask = 1. / block_size * np.ones((block_size,))
-        # separation of filters to speedup convolution
-        ndi.convolve1d(image, mask, axis=0, output=thresh_image, mode=mode,
-                       cval=cval)
-        ndi.convolve1d(thresh_image, mask, axis=1, output=thresh_image,
-                       mode=mode, cval=cval)
+        masks = {}
+        for b in set(block_size):
+            masks[b] = 1. / b * np.ones((b,))
+        # # separation of filters to speedup convolution
```
mkcor, line 230:
```suggestion
        # separation of filters to speed up convolution
```

skimage/filters/thresholding.py
grlee77: yes, I we can just make a single call to `uniform_filter`. thanks

skimage/filters/thresholding.py
grlee77: good suggestion, but this line ended up being removed altogether when switching from `convolve1d` to `uniform_filter`

rfezzani (approved):
Thank you @grlee77 :tada:

PR #5171, opened 2021-01-04
fast, non-Cython implementation for correlate_sparse

grlee77:
## Description

This PR replaces the recently added `correlate_sparse` Cython version with a simple slicing-based variant. This was done so that this function can be used with other array types implementing the NumPy API, such as CuPy. There should be no change in behavior or API.

A secondary benefit is that this also profiled modestly faster on 2d and 3d cases. See the `threshold_sauvola` benchmarks for v0.18.1 vs. this PR below. I tried out varying sparsity levels as well and wasn't able to find a case where it is worse than the current implementation.

### Benchmark Results

**v0.18.1**
· Discovering benchmarks
[  0.00%] ·· Benchmarking existing-py_home_lee8rx_miniconda3_envs_skimage_dev_bin_python3.7
[ 25.00%] ··· benchmark_filters.ThresholdSauvolaSuite.time_sauvola                                                                               252±0ms
[ 50.00%] ··· benchmark_filters.ThresholdSauvolaSuite.time_sauvola_3d                                                                            458±0ms

**This PR**
· Discovering benchmarks
[  0.00%] ·· Benchmarking existing-py_home_lee8rx_miniconda3_envs_skimage_dev_bin_python3.7
[ 25.00%] ··· benchmark_filters.ThresholdSauvolaSuite.time_sauvola                                                                               188±0ms
[ 50.00%] ··· benchmark_filters.ThresholdSauvolaSuite.time_sauvola_3d                                                                            387±0ms

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/filters/_sparse.py
```diff
@@ -32,6 +31,89 @@ def _to_np_mode(mode):
...
+            raise ValueError(msg)
+
+
+def _to_np_mode(mode):
+    """Convert padding modes from `ndi.correlate` to `np.pad`."""
+    mode_translation_dict = dict(nearest='edge', reflect='symmetric',
+                                 mirror='reflect')
+    if mode in mode_translation_dict:
+        mode = mode_translation_dict[mode]
+    return mode
```
rfezzani, line 59:
```suggestion

```

`_validate_window_size` and `_to_np_mode` are already defined above :wink:.

BTW, `_to_np_mode` is already defined in `skimage.filters._sparse` and the opposite function `_to_ndimage_mode` is defined in `skimage.transform._geometric` and used in `skimage.tranform._warp`. These two functions can be added to `skimage._shared` module.

skimage/filters/_sparse.py
```diff
@@ -70,28 +152,18 @@ def correlate_sparse(image, kernel, mode='reflect'):
...
-    values = kernel[indices].astype(padded_image.dtype)
 
-    result = np.zeros(
-        [a - b + 1 for a, b in zip(padded_image.shape, kernel.shape)],
-        dtype=padded_image.dtype,
+    indices = np.nonzero(kernel)
+    values = kernel[indices].astype(padded_image.dtype, copy=False)
+    indices = list(zip(*indices))
+    kernel_indices_and_values = [(idx, v) for idx, v in zip(indices, values)]
+    if (0, ) * kernel.ndim not in indices:
```
rfezzani, line 163:
```suggestion
    if kernel.reshape(-1)[0] == 0:
```

rfezzani (commented):
Thank you @grlee77, I like the idea of reducing code base :wink:

skimage/filters/_sparse.py
grlee77: I don't think this is equivalent? What I want is if the sparse kernel does not have an entry at the `(0,) * ndim` corner location, then an entry should be added so that there will not be a shift in the output of `_correlate_sparse`. (Internal to `_correlate_sparse`, the output array is initialized from this first point, which needs to be the corner location in order to avoid an unexpected spatial shift in the output).

See if e4cff50 makes things any clearer

rfezzani: Isn't `(0,) * ndim` the index of the first entry of `kernel`? `kernel.reshape(-1)[0]` is the first entry of `kernel`. If it is `0` then `(0,) * ndim` is not in `indices`.

grlee77: Yes, I agree now that what you suggest is equivalent and would be fine with changing to that.

skimage/_shared/utils.py
```diff
@@ -415,3 +415,21 @@ def _validate_interpolation_order(image_dtype, order):
              FutureWarning, stacklevel=2)
 
     return order
+
+
+def _to_np_mode(mode):
```
rfezzani, line 420: I suggested this changes for an other PR, but it is fine by me here :slightly_smiling_face:

skimage/filters/_sparse.py
```diff
@@ -23,13 +22,59 @@ def _validate_window_size(axis_sizes):
...
+
+    Notes
+    -----
+    This function only returns results for the 'valid' region of the
+    convolution, and thus `out` will be smaller than `image` by an amount
+    equal to the kernel size along each axis.
+    """
+    idx, val = kernel_indices_and_values[0]
+    # implementation assumes this corner is first in kernel_indices_in_values
+    assert tuple(idx) == (0, ) * image.ndim
```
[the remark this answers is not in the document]
emmanuelle: ValueError maybe?

emmanuelle (approved):
Thanks @grlee77 ! I had the same comment as @rfezzani with respect to assertion vs error but it's not in a function of the public API and the public-API function calling the helper function will not cause problems, so this should not block the PR. Also, I noted that `correlate_sparse` is in the public API but it's not use in any gallery example, I'm not sure it is very discoverable for users (but once again this should not block the PR).

skimage/filters/_sparse.py
grlee77: On closer look, I would prefer to leave this check as-is so that it avoids a device->host transfer if using a CuPy backend. (For CuPy, `kernel.reshape(-1)[0]` will give a 0-dimensional `cupy.ndarray` rather than a host scalar and comparing to 0 requires overhead of transferring that value from the GPU back to the host to do the comparison)

skimage/_shared/utils.py
grlee77: This was also added in #5173, so I will probably have to rebase whichever one is merged second.

skimage/filters/_sparse.py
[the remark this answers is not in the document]
grlee77: I added a RuntimeError

skimage/filters/thresholding.py
```diff
@@ -944,20 +944,27 @@ def _mean_std(image, w):
...
+            (indices, (-1) ** (image.ndim % 2 != np.sum(indices) % 2))
+        )
 
     total_window_size = np.prod(w)
-    sum_full = correlate_sparse(integral, kern, mode='valid')
-    m = sum_full / total_window_size
-    sum_sq_full = correlate_sparse(integral_sq, kern, mode='valid')
-    g2 = sum_sq_full / total_window_size
+    kernel_shape = tuple(_w + 1 for _w in w)
+    m = _correlate_sparse(integral, kernel_shape, kernel_indices_and_values)
```
jni, line 963: As a minor point, from an API niceness standpoint I'd rather `kernel_indices` and `kernel_values` be two separate arrays that are zipped inside the function. Not critical though.

jni (approved):
Just some minor comments from me. Thanks @grlee77, very nice!

skimage/filters/_sparse.py
```diff
@@ -23,13 +22,59 @@ def _validate_window_size(axis_sizes):
...
+    in _mean_std, then the view created here will match the size of the
+    original image.
+    """
+    sl_shift = tuple([slice(c, s - (w_ - 1 - c))
+                      for c, w_, s in zip(idx, kernel_shape, padded.shape)])
+    v = padded[sl_shift]
+    if val == 1:
+        return v
+    elif val == -1:
+        return -v
```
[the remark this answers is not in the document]
grlee77: I suspected it may be faster, but hadn't timed it. I tried just now and it was faster on very small arrays, but the same on large arrays, so probably not important to have it as a special case

```Python
import numpy as np
a = np.ones((4096, 4096))

%timeit (-a)
22.6 ms ± 130 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit (-1 * a)
22.6 ms ± 33.4 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

a = np.ones((16, 16))

%timeit (-a)
400 ns ± 2.4 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit (-1 * a)
913 ns ± 3.76 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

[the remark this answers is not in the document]
grlee77: The same relative performance pattern occurs for CuPy as reported for NumPy above

skimage/filters/_sparse.py
```diff
@@ -23,13 +22,59 @@ def _validate_window_size(axis_sizes):
...
+    -----
+    This function only returns results for the 'valid' region of the
+    convolution, and thus `out` will be smaller than `image` by an amount
+    equal to the kernel size along each axis.
+    """
+    idx, val = kernel_indices_and_values[0]
+    # implementation assumes this corner is first in kernel_indices_in_values
+    assert tuple(idx) == (0, ) * image.ndim
+    out = _get_view(image, kernel_shape, idx, val)
+    if not out.flags.owndata:
```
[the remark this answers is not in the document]
grlee77: I am fine with that and don't think it will substantially reduce performance, but given the reliance on `scipy.ndimage` elsewhere in the library for many things related to filtering/morphology/labeling (as well as usage of FFTs or sparse matrices in various places), we will already not be able to realistically do many things purely with the Array API. I think NumPy+SciPy, Dask+dask-image and CuPy are the only implementations that provide most of the needed features currently.

grlee77:
The CI failures seen here should be resolved by #5175

skimage/filters/_sparse.py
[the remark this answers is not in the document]
jni: Well, dask arrays don't have .flags either...

```python
In [1]: arr = da.random.random((5,))
In [2]: arr.flags
AttributeError                            Traceback (most recent call last)
<ipython-input-2-c469d1119f64> in <module>
----> 1 arr.flags

AttributeError: 'Array' object has no attribute 'flags'
```

skimage/filters/_sparse.py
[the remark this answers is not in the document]
jni: Coolio, let's remove that special case then?

stefanv:
Thanks, Greg!

PR #5172, opened 2021-01-04
Stain unmixing fix

### LINKED ISSUE #5164: RGB to HED conversion after version 0.17
## Description
After upgrading our [project](https://github.com/histolab/histolab) environment to version 0.18.x we have noticed that the RGB to HED conversion behaves unexpectedly according to the biological meaning of the HED color space. We have done some research and carefully read the [PR](https://github.com/scikit-image/scikit-image/pull/4725) changing the stain deconvolution method, however the HED conversion in version 0.17.x seemed to be more appropriate; in particular the H channel highlights the nuclei while the DAB channel highlights the brown stain:

**version 0.17.x**
image

We are not able to find the same biological meaning in version 0.18.x:

**version 0.18.x**
image

Regards,
Nicole, Alessia, Ernesto (histolab team)

crisluengo:
## Description

Closes #5164

- Clip negative values in output of RGB to HED conversion (actually in `skimage.color. separate_stains`).
- Fixed stain separation tests accordingly:
   - Round trips start with stains, such that the roundtrip does not require negative values. Unmixing an RGB image can lead to negative values, which we clip in this fix. The round-trip is now impossible. But if we start with a stains image, convert it to RGB, and then unmix that RGB image, we can have a round-trip without errors. The stains image must be dim enough such that adding stains together doesn't saturate a channel in the RGB image.
   - Using a blue+red+orange stain combination, the test previously used a 2-stain combination that made it impossible to do a correct roundtrip: one of the stain channels was multiplied by 0, and could therefore never be recovered. I picked the first 3-stain combination I found in the existing set that wasn't HED.
- Improves IHC stain separation example:
    - Proper display of unmixed stains requires applying the non-linear mapping in `combine_stains`.
    - The contrast stretch ignores 1% of pixel values, to ensure a clear image.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/color_exposure/plot_ihc_color_separation.py
+++ b/doc/examples/color_exposure/plot_ihc_color_separation.py
@@ -19,35 +19,38 @@ Diaminobenzidine (DAB) which gives a brown color.
        American Society of Cytology, vol. 23, no. 4, pp. 291-9, Aug. 2001.
 
 """
+import numpy as np
 import matplotlib.pyplot as plt
 
 from skimage import data
-from skimage.color import rgb2hed
-from matplotlib.colors import LinearSegmentedColormap
+from skimage.color import rgb2hed, hed2rgb
 
-# Create an artificial color close to the original one
-cmap_hema = LinearSegmentedColormap.from_list('mycmap', ['white', 'navy'])
-cmap_dab = LinearSegmentedColormap.from_list('mycmap', ['white',
-                                             'saddlebrown'])
-cmap_eosin = LinearSegmentedColormap.from_list('mycmap', ['darkviolet',
-                                               'white'])
-
+# Example IHC image
 ihc_rgb = data.immunohistochemistry()
+
+# Separate the stains from the IHC image
 ihc_hed = rgb2hed(ihc_rgb)
 
+# Create an RGB image for each of the stains
+null = np.zeros_like(ihc_hed[:, :, 0])
+ihc_h = hed2rgb(np.stack((ihc_hed[:, :, 0], null, null), axis=-1))
+ihc_e = hed2rgb(np.stack((null, ihc_hed[:, :, 1], null), axis=-1))
+ihc_d = hed2rgb(np.stack((null, null, ihc_hed[:, :, 2]), axis=-1))
+
+# Display
 fig, axes = plt.subplots(2, 2, figsize=(7, 6), sharex=True, sharey=True)
 ax = axes.ravel()
 
 ax[0].imshow(ihc_rgb)
 ax[0].set_title("Original image")
 
-ax[1].imshow(ihc_hed[:, :, 0], cmap=cmap_hema)
+ax[1].imshow(ihc_h)
 ax[1].set_title("Hematoxylin")
 
-ax[2].imshow(ihc_hed[:, :, 1], cmap=cmap_eosin)
-ax[2].set_title("Eosin")
+ax[2].imshow(ihc_e)
+ax[2].set_title("Eosin")  # Note that there is no Eosin stain in this image
 
-ax[3].imshow(ihc_hed[:, :, 2], cmap=cmap_dab)
+ax[3].imshow(ihc_d)
 ax[3].set_title("DAB")
 
 for a in ax.ravel():
@@ -59,12 +62,15 @@ fig.tight_layout()
 ######################################################################
 # Now we can easily manipulate the hematoxylin and DAB "channels":
 
-import numpy as np
 from skimage.exposure import rescale_intensity
 
 # Rescale hematoxylin and DAB signals and give them a fluorescence look
-h = rescale_intensity(ihc_hed[:, :, 0], out_range=(0, 1))
-d = rescale_intensity(ihc_hed[:, :, 2], out_range=(0, 1))
+h = rescale_intensity(ihc_hed[:, :, 0], out_range=(0, 1),
+                      in_range=(0, np.percentile(ihc_hed[:, :, 0], 99)))
+d = rescale_intensity(ihc_hed[:, :, 2], out_range=(0, 1),
+                      in_range=(0, np.percentile(ihc_hed[:, :, 2], 99)))
+
+# Put the two channels into an RGB image as green and blue channels
 zdh = np.dstack((np.zeros_like(h), d, h))
 
 fig = plt.figure()
--- a/doc/examples/color_exposure/plot_ihc_color_separation.py
+++ b/doc/examples/color_exposure/plot_ihc_color_separation.py
@@ -71,7 +71,7 @@ d = rescale_intensity(ihc_hed[:, :, 2], out_range=(0, 1),
                       in_range=(0, np.percentile(ihc_hed[:, :, 2], 99)))
 
 # Put the two channels into an RGB image as green and blue channels
-zdh = np.dstack((np.zeros_like(h), d, h))
+zdh = np.dstack((null, d, h))
 
 fig = plt.figure()
 axis = plt.subplot(1, 1, 1, sharex=ax[0], sharey=ax[0])
--- a/skimage/color/tests/test_colorconv.py
+++ b/skimage/color/tests/test_colorconv.py
@@ -47,6 +47,7 @@ class TestColorconv(TestCase):
     img_rgba = np.array([[[0, 0.5, 1, 0],
                           [0, 0.5, 1, 1],
                           [0, 0.5, 1, 0.5]]]).astype(float)
+    img_stains = img_as_float(img_rgb) * 0.3
 
     colbars = np.array([[1, 1, 0, 0, 1, 1, 0, 0],
                         [1, 1, 1, 1, 0, 0, 0, 0],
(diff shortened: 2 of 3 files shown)


skimage/color/tests/test_colorconv.py
```diff
@@ -183,32 +184,33 @@ def test_xyz_rgb_roundtrip(self):
...
+        img_out = separate_stains(img_out, bro_from_rgb)
+        assert_equal(img_as_ubyte(img_out), img_in)
+
+    # BRO<->RGB roundtrip with float image
+    def test_bro_rgb_roundtrip_float(self):
+        from skimage.color.colorconv import bro_from_rgb, rgb_from_bro
+        img_in = self.img_stains
+        img_out = combine_stains(img_in, rgb_from_bro)
+        img_out = separate_stains(img_out, bro_from_rgb)
+        assert_array_almost_equal(img_out, img_in)
```
grlee77, line 213: The new RGB/BRO tests look good and it doesn't seem these functions were previously being tested. 

Ideally we would maintain some kind of test of `rgb_from_hdx` and `hdx_from_rgb`, as those don't seem to be tested elsewhere. Maybe just keep a round trip test, but without the combine/separate stains calls?

grlee77 (changes requested):
I had one comment related to test coverage, but otherwise this looks good to me. Thanks for fixing this!

skimage/color/tests/test_colorconv.py
crisluengo: `rgb_from_hdx` and `hdx_from_rgb` are constants, and so are the two arrays I replaced them with: https://github.com/scikit-image/scikit-image/blob/ff188272a0ea509cb4d6025d1906c494914a4304/skimage/color/colorconv.py#L485

There's a long list of these arrays there. These arrays describe the RGB color of each of the stains in a common set. If you want to test that these arrays are correct, you'd have to source an image of a slide stained with the corresponding stains, apply the stain unmixing, and verify that there are few negative values in the result.

It is the functions [`separate_stains`](https://github.com/scikit-image/scikit-image/blob/ff188272a0ea509cb4d6025d1906c494914a4304/skimage/color/colorconv.py#L1391) and [`combine_stains`](https://github.com/scikit-image/scikit-image/blob/ff188272a0ea509cb4d6025d1906c494914a4304/skimage/color/colorconv.py#L1459) that we're testing here with these round-trips. Both the HED<->RGB and the BRO<->RGB round-trips test the same two functions, just with different data.

If you really want to test these functions with each of the data matrices, we could certainly do that. For the 2-stain varieties, we'd have to compare only the first two channels after the roundtrip. Or set the 3rd channel of the input to 0. I just don't think that such a test would add anything here.

crisluengo: I mean we could just verify that `rgb_from_hdx @ hdx_from_rgb` is the identity matrix to ensure that the pair of matrices would correctly reverse each other. But this is guaranteed because the one is computed as the inverse of the other.

doc/examples/color_exposure/plot_ihc_color_separation.py
```diff
@@ -59,12 +62,15 @@
...
 # Rescale hematoxylin and DAB signals and give them a fluorescence look
-h = rescale_intensity(ihc_hed[:, :, 0], out_range=(0, 1))
-d = rescale_intensity(ihc_hed[:, :, 2], out_range=(0, 1))
+h = rescale_intensity(ihc_hed[:, :, 0], out_range=(0, 1),
+                      in_range=(0, np.percentile(ihc_hed[:, :, 0], 99)))
+d = rescale_intensity(ihc_hed[:, :, 2], out_range=(0, 1),
+                      in_range=(0, np.percentile(ihc_hed[:, :, 2], 99)))
+
+# Put the two channels into an RGB image as green and blue channels
 zdh = np.dstack((np.zeros_like(h), d, h))
```
rfezzani, line 74:
```suggestion
zdh = np.dstack((null, d, h))
```

skimage/color/tests/test_colorconv.py
grlee77: Sorry for the noise on this comment. I was indeed thinking these were functions and not just the transform matrices. This seems fine as-is, then.

crisluengo: No need to apologize. I appreciate your thoroughness. Interactions I've had with this project have all given me a really good impression. Keep up the good work!

alexdesiqueira:
This looks great, thank you very much @crisluengo!
@grlee77 @rfezzani any idea on why `linux-cp3.7-mini-req-optional-deps` is failing? Does it concern this PR?

grlee77:
> @grlee77 @rfezzani any idea on why linux-cp3.7-mini-req-optional-deps is failing? Does it concern this PR?

I saw these kind of failures on my recent PRs too. I think #5175 should fix them

alexdesiqueira:
Thank you again @crisluengo, we really appreciate your help!

crisluengo:
@alexdesiqueira, glad I could contribute a bit.

alessiamarcolini:
Hello maintainers :) any idea on when this fix is going to be incorporated in a stable release?

jni:
Hi @alessiamarcolini! We are working to release 0.19 very soon, but realistically I think it might be a few more weeks still. Apologies for the long delay! We are trying to line things up for 1.0 so we are taking some extra time to get things right for the ~last 0.x release...

mkcor:
I just noticed that, in the current dev version, the Eosin channel is gone:
eosin_gone
cf. https://github.com/scikit-image/scikit-image/issues/5164#issue-776407593

alexdesiqueira:
@mkcor IIRC that is expected for that input image, according to the code we proposed at the time. Please let me know if this does not seem right.

mkcor:
@alexdesiqueira right. There's even an inline comment to specify this. I think it could/should be made clearer in the narrative part of the tutorial, so it's more educational. The reader is left *inferring* that should there be any eosin-stained tissues, they would appear in the red channel of the stain-separated image...

It's not clear that it would make sense to feature an "Eosin" subplot, since immunohistochemical staining involves two dyes, hematoxylin and DAB. Apparently, H&E (hematoxylin and eosin) is another popular stain in histopathology, also involving two dyes (not all three at once).

I'm reading https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4354574/ and then I'll submit a small PR.

alexdesiqueira:
> I'm reading https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4354574/ and then I'll submit a small PR.

@mkcor thank you!

crisluengo:
Yes, that is right. Slides are either Hematoxylin and Eosin, or Hematoxylin and DAB, I have never seen a slide with all three stains.

I assume the “HED” color space mixes all three colors so it can handle either of the two stain combinations, and also to keep the concept of converting 3 channels into 3 channels, like all the other color space conversions do.

I agree that the documentation could be clearer on this.

You could take any H&E image to add to this section of the tutorial. <https://www.google.com/search?q=h%26e&tbm=isch&hl=en-us&chips=q:h%26e,g_1:staining:TGg8zTdutgw%3D&client=safari&prmd=mnixv&sa=X&ved=2ahUKEwih99a8u530AhXEIH0KHV0xA-MQ4lYoA3oECAEQCg&biw=320&bih=455>

mkcor:
> Yes, that is right. Slides are either Hematoxylin and Eosin, or Hematoxylin and DAB, I have never seen a slide with all three stains.

Thanks, @crisluengo.

> I assume the “HED” color space mixes all three colors so it can handle either of the two stain combinations, and also to keep the concept of converting 3 channels into 3 channels, like all the other color space conversions do.

Makes sense.

> I agree that the documentation could be clearer on this.
> 
> You could take any H&E image to add to this section of the tutorial. https://www.google.com/search?q=h%26e&tbm=isch&hl=en-us&chips=q:h%26e,g_1:staining:TGg8zTdutgw%3D&client=safari&prmd=mnixv&sa=X&ved=2ahUKEwih99a8u530AhXEIH0KHV0xA-MQ4lYoA3oECAEQCg&biw=320&bih=455

Is there a way to filter by license? Do you know yourself of an H&E image in the public domain or under CC0? The images used in [Wikipedia pages](https://en.wikipedia.org/wiki/H%26E_stain#Examples_of_H&E_stained_tissues) are typically licensed under CC BY-SA 3.0... I would like to use a CC0-licensed image, so we can add it to our [data registry](https://gitlab.com/scikit-image/data/#data).

alessiamarcolini:
> Is there a way to filter by license? Do you know yourself of an H&E image in the public domain or under CC0? The images used in Wikipedia pages are typically licensed under CC BY-SA 3.0... I would like to use a CC0-licensed image, so we can add it to our data registry.

Hey @mkcor yes , you can search for "h&e -commons" (the `-` to exclude wikipedia results), and then under search tools there is "Usage Rights", where I selected "Creative Commons licenses", giving this [URL](https://www.google.com/search?q=h%26e%20-commons&tbm=isch&hl=en-us&tbs=il:cl&client=safari&prmd=mnixv&sa=X&ved=0CAAQ1vwEahcKEwiAr7SFwa70AhUAAAAAHQAAAAAQAg&biw=1757&bih=1680).

The results are not many, but I think [this image](https://pixnio.com/de/wissenschaft/mikroskopische-aufnahmen/west-nil-krankheit/mikrophotographie-histopathologischen-verbunden-sind-westen-nil-virus) will work!

alexdesiqueira:
Thanks @alessiamarcolini!
@mkcor for the example we already have, I think it would be sufficient to present three subplots: Original, H&E, DAB — as discussed previously. How does that sound?

alexdesiqueira:
Thanks @alessiamarcolini!
Are we thinking of using another image, @mkcor? IMHO, we could solve that using the example we already have, and presenting three subplots: Original, H&E (their sum), DAB — as discussed previously. My two BRL cents, though.

alexdesiqueira:
Thanks @alessiamarcolini!
Are we thinking of using another image, @mkcor? IMHO, we could solve that using the example we already have, and presenting three subplots: Original, H&E (their sum), DAB — as discussed previously. My two BRL cents, though.
More than that, I think it's best for us to start a new issue to discuss this; else, this discussion we're having now will be lost in this PR :slightly_smiling_face:

Issue #5064, opened 2020-11-13
enable scipy grid-constant boundary handling


grlee77:
## Convert 'constant' -> 'grid-constant' when calling scipy.ndimage.interpolation functions

#1514 covered the difference in mode='constant' for scikit-image vs. scipy.ndimage. At that time, we switched the default mode away from constant, but as of SciPy 1.6, we will can also use mode 'grid-constant' in scipy.ndimage which should be equivalent to scikit-image's own constant interpolation mode.

I am just opening this issue as a future TODO as this may need to wait awhile until SciPy >=1.6 is a dependency. Specifically, at the following location:
https://github.com/scikit-image/scikit-image/blob/ed889ba3deb17cd94d1f0f9b730f6564585936ee/skimage/transform/_geometric.py#L11-L12

we should possibly add `constant='grid-constant'`.

jni:
Thanks @grlee77! I think we should detect the scipy version and add that mode translation if it's after 1.6. That can be done immediately which is nice.

jni:
Also, thanks to NEP-29, we can immediately know the date when we can remove the detection. =)

PR #5173, opened 2021-01-05
Simplify resize implementation using new SciPy 1.6 zoom option

grlee77:
## Description

This PR greatly simplifies the implementation of skimage.transform.resize, by using SciPy 1.6.0's new `grid_mode=True` option to `scipy.ndimage.zoom`. This allows us to avoid manually generating coordinates and calling `map_coordinates`.

numpy.pad -> scipy.ndimage `mode` conversion was also updated for SciPy 1.6.
closes gh-5064

Currently this new code path replaces both the `scipy.ndimage.map_coordinates` code path as well as the one relying on scikit-image's own Cython-based interpolation code. We could also potentially leave the existing Cython code path in place (it was ~10-20% faster than `ndimage.zoom` in 2D with orders 0 or 1). Let me know which is preferred.

It can be seen in the benchmarks below that `map_coordinates` is occasionally faster than zoom (see e.g. the 3D case for order=5), but this comes at cost of substantially higher memory use needed to store the full set of output coordinates.

ASV benchmarks for `resize` and `rescale` (which just calls resize internally) were added. The following table summarizes some results for this PR vs. v0.18.1 in both run time and memory usage. This PR is sometimes substantially faster (3D for order 0, 1) and occasionally a bit slower.

shape | order | v0.18.1 | this PR | memory v0.18.1 | memory PR
------|-------|---------|---------|----------------|---------------
   (2000, 4000)   |  0  |  62.7 ms    |   79.6 ms  |  160  |  157
   (2000, 4000)   |  1  |  160 ms    |   177 ms    |  160  |  158
   (2000, 4000)   |  3  |  531 ms    |   479 ms    |  160  |  166
   (2000, 4000)   |  5  |    1.48 s  |     1.04 s  |  637  |  166
   (150, 150, 150)  |  0  |  153 ms    |   32.8 ms   |  255  |  120
   (150, 150, 150)  |  1  |  230 ms    |   126 ms    |  255  |  120
   (150, 150, 150)  |  3  |  828 ms    |   896 ms    |  255  |  129
   (150, 150, 150)  |  5  |    2.44 s  |     3.04 s  |  255  |  129

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


emmanuelle:
`transform/tests/test_warps.py::test_downsize_anti_aliasing` seems to be failing consistently.

emmanuelle:
Maybe an item should be added in the TODO so that we don't forget to remove the other code when the scipy dep is bumped to 1.6 (this will probably take some time... so a reminder might be good).

grlee77:
> transform/tests/test_warps.py::test_downsize_anti_aliasing seems to be failing consistently.

I was relying on SciPy to raise on unrecognized modes, but SciPy raises a `RuntimeError` rather than the `ValueError` the test was expecting. I added a `ValueError` again here for consistency with prior behavior, but if you don't think it is important we can just remove that error check from the tests.

TODO.txt
```diff
@@ -107,3 +107,5 @@ Other
   ``skimage.morphology._util._fast_pad``.
 * When sphinx-gallery>=0.9.0, remove the thumbnail_size in
   doc/source/conf.py as the default value will be comparable (#4801).
+* When ``scipy`` is set to >= 1.16, remove legacy (i.e. non-zoom) code paths in
+``skimage.transform.resize``.
```
jni, line 111: Thanks to NEP-29, we can now do date-based reminders (🎉!): this will be done on Jan 1 2022. Should we add a separate time-based section to the todos?

jni:
Love it! I think the few specific cases where performance is worse, the tradeoff is worth it for code simplicity.

TODO.txt
sciunto: good idea, yes

jni: Oops — that should be 2023. Time flies when you're having fun! =P

grlee77: See how this latest commit reorganizing the "Other" section looks. 

We can actually do many of the 2021 items now (we already require NumPy 1.16 and we SciPy 1.1 was released in May 2018).

jni (approved):

skimage/registration/_optical_flow.py
```diff
@@ -76,7 +76,7 @@ def _tvl1(reference_image, moving_image, flow0, attachment, tightness,
                                              [1] + reference_image.ndim * [3])
 
         image1_warp = warp(moving_image, get_warp_points(grid, flow_current),
-                           mode='nearest')
```
rfezzani, line 79: :eyes: Was this a bug?

skimage/transform/_warps.py
```diff
@@ -127,26 +132,21 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
...
+        image = convert_to_float(image, preserve_range)
+        out = ndi.zoom(image, zoom_factors, order=order, mode=ndi_mode,
+                       cval=cval, grid_mode=True)
+        _clip_warp_output(image, out, order, mode, cval, clip)
+        return out
+
+    # TODO: Remove the fallback code below once SciPy >= 1.6.0 is required.
+
     # 2-dimensional interpolation
     if len(output_shape) == 2 or (len(output_shape) == 3 and
```
rfezzani, line 151: Simply to avoid an extra `return` statement.
```suggestion

    # TODO: Remove the fallback code below once SciPy >= 1.6.0 is required.

    # 2-dimensional interpolation
    elif len(output_shape) == 2 or (len(output_shape) == 3 and
```

skimage/transform/_geometric.py
```diff
@@ -6,15 +6,6 @@
 from .._shared.utils import get_bound_method_class, safe_as_int
 
 
-def _to_ndimage_mode(mode):
```
rfezzani, line 9: I see in master branch that `_to_ndimage_mode` is not used in this file but we are reimporting it from `skimage._shared.utils` :confounded:

skimage/_shared/utils.py
```diff
@@ -415,3 +417,35 @@ def _validate_interpolation_order(image_dtype, order):
...
+             "more info.").format(mode))
+    return _fix_ndimage_mode(mode_translation_dict[mode])
+
+
+def _fix_ndimage_mode(mode):
+    # SciPy 1.6.0 introduced grid variants of constant and wrap which
+    # have less surprising behavior for images. Use these when available
+    grid_modes = {'constant': 'grid-constant', 'wrap': 'grid-wrap'}
+    if NumpyVersion(scipy.__version__) >= '1.6.0' and mode in grid_modes:
+        return grid_modes[mode]
```
rfezzani, line 450:
```suggestion
    if NumpyVersion(scipy.__version__) >= '1.6.0':
        mode = grid_modes.get(mode, mode)
```

rfezzani (approved):
Thank you @grlee77, I added some suggestions/remarks, but the PR is good as is :wink:

rfezzani:
@grlee77, it looks like a rebase is necessary :)

skimage/registration/_optical_flow.py
grlee77: I think it was an unintentional use of the scipy.ndimage name rather than letting the function convert from the numpy.pad equivalent name ('edge'). The docstring to `warp` conforms to `numpy.pad` style names (i.e. 'edge' rather than its `scipy.ndimage` equivalent: 'nearest'). It just so happened that warp didn't raise an error on unrecognized modes and just passed them along unmodified.

skimage/transform/_geometric.py
grlee77: is this just a comment or are you recommending something here? (I didn't quite follow where we are "reimporting it from")

rfezzani: It was just a comment :slightly_smiling_face:

skimage/registration/_optical_flow.py
rfezzani: May be something to investigate :confused:...

Issue #5042, opened 2020-10-30
Issues with EllipseModel


rk-exxec:
## Description
Hey, I have issues with the EllipseMopdel estimate function.
It produces a completely wrong ellipse, what am I doing wrong.
This is what it looks like:
![grafik](https://user-images.githubusercontent.com/40037381/97692066-8f3a0a80-1a9f-11eb-909c-8c21a6a9688a.png)
I am detecting a contour with opencv (blue line) and then feeding that contour to the EllipseModel.

## Way to reproduce
```python
import math
import cv2
import numpy as np
from skimage.measure import EllipseModel
def evaluate_droplet(img, y_base):
    crop_img = img[:y_base,:]
    shape = img.shape
    height = shape[0]
    width = shape[1]

    bw_edges = cv2.Canny(crop_img, 76, 179)

    contours, hierarchy = cv2.findContours(bw_edges, cv2.RETR_TREE, cv2.CHAIN_APPROX_NONE)

    if len(contours) == 0:
        raise ValueError('No contours found!')

    edge = max(contours, key=cv2.contourArea)

    img = cv2.drawContours(img,edge,-1,(255,0,0),2)
    points = edge.reshape(-1,2)
    points[:,[0,1]] = points[:,[1,0]]
    ell = EllipseModel()
    if not ell.estimate(points): raise RuntimeError('Couldn\'t fit ellipse')
    x0, y0, a, b, phi = ell.params

    img = cv2.ellipse(img, (int(round(x0)),int(round(y0))), (int(round(a)),int(round(b))), int(round(math.degrees(phi))), 0, 360, (255,0,255), thickness=1, lineType=cv2.LINE_AA)

if __name__ == "__main__":
    im = cv2.imread('untitled1.png')
    try:
        drp = evaluate_droplet(im, 250)
    except Exception as ex:
        print(ex)
    cv2.imshow('Test',im)
    cv2.waitKey(0)
```

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
>>> import sys; print(sys.version)
3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:52:53) [MSC v.1927 64 bit (AMD64)]
>>> import platform; print(platform.platform())
[...]

PR #5179, opened 2021-01-10
bugfix: prevent integer overflow in EllipseModel

mark-boer:
## Description

This is a bugfix to issue #5042. It makes sure that the `EllipseModel` and `CircleModel` do not cause integer overflows

## Checklist

I have a question, while trying to implement a unit test I ran into another bug. The test is still implemented, but is marked with a `pytest.mark.skip`. How would you like to handle this?

Also, does the code need to pass any kind of static analysis? Like black or flake8?

## For reviewers

  later.

### DIFF
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -270,6 +270,9 @@ class CircleModel(BaseModel):
 
         _check_data_dim(data, dim=2)
 
+        float_type = np.promote_types(data.dtype, np.float32)
+        data = data.astype(float_type, copy=False)
+
         x = data[:, 0]
         y = data[:, 1]
 
@@ -280,7 +283,7 @@ class CircleModel(BaseModel):
         sum_xy = np.sum(x * y)
         m1 = np.stack([[np.sum(x ** 2), sum_xy, sum_x],
                        [sum_xy, np.sum(y ** 2), sum_y],
-                       [sum_x, sum_y, float(len(x))]])
+                       [sum_x, sum_y, len(x)]])
         m2 = np.stack([[np.sum(x * x2y2),
                         np.sum(y * x2y2),
                         np.sum(x2y2)]], axis=-1)
@@ -414,13 +417,16 @@ class EllipseModel(BaseModel):
         # another REFERENCE: [2] http://mathworld.wolfram.com/Ellipse.html
         _check_data_dim(data, dim=2)
 
+        float_type = np.promote_types(data.dtype, np.float32)
+        data = data.astype(float_type, copy=False)
+
         x = data[:, 0]
         y = data[:, 1]
 
         # Quadratic part of design matrix [eqn. 15] from [1]
         D1 = np.vstack([x ** 2, x * y, y ** 2]).T
         # Linear part of design matrix [eqn. 16] from [1]
-        D2 = np.vstack([x, y, np.ones(len(x))]).T
+        D2 = np.vstack([x, y, np.ones(len(x), dtype=float_type)]).T
 
         # forming scatter matrix [eqn. 17] from [1]
         S1 = D1.T @ D1
--- a/skimage/measure/fit.py
+++ b/skimage/measure/fit.py
@@ -270,6 +270,7 @@ class CircleModel(BaseModel):
 
         _check_data_dim(data, dim=2)
 
+        # to prevent integer overflow, cast data to float, if it isn't already
         float_type = np.promote_types(data.dtype, np.float32)
         data = data.astype(float_type, copy=False)
 
@@ -417,6 +418,7 @@ class EllipseModel(BaseModel):
         # another REFERENCE: [2] http://mathworld.wolfram.com/Ellipse.html
         _check_data_dim(data, dim=2)
 
+        # to prevent integer overflow, cast data to float, if it isn't already
         float_type = np.promote_types(data.dtype, np.float32)
         data = data.astype(float_type, copy=False)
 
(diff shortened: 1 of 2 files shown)


mark-boer:
Found the reason the test I wrote is failing, the input data chosen, leads to a ridiculously small eigenvalue, which in turn led to significant floating point errors in the matrix inverse. It is probably a solution to first subtract the centroid of the input data and then add this centroid to the center of the Circle. Also I found that the algebraic method used, is not actually a proper least square fit. I'll create a new issue.

mark-boer:
The behaviour could be improved by using something like this: 

```
A = np.append(xy * 2, np.ones((xy.shape[0], 1)), axis=1)
f = np.sum(xy ** 2, axis=1)
C, _, _, _ = np.linalg.lstsq(A, f, rcond=None)
center = C[0:2]
distances = scipy.spatial.minkowski_distance(center,xy)
r = np.sqrt(np.mean(distances ** 2))
```

See e.g. sphere fit i wrote a year ago, this contains 2 fits, one fast algebraic one and one iterative one:
https://github.com/mark-boer/geomfitty/blob/4f24071f524302a95144dbd91bc84fa33c5dc26b/geomfitty/fit3d.py#L42

stefanv:
Thanks for the fix, @mark-boer.  I did a quick check and it looks like vstack does type promotion already. Could you explain the rationale behind the fix (casting the data to float)?

Before merging, we should remove the failing test and address that as part of your other issue.

mark-boer:
The `vstack` on which line do you mean? The EllipseModel overflows in the following line: 
```python
S1 = D1.T @ D1
```
For more info on what goes wrong see issue  #5042.

jni:
@stefanv I think the point of the cast might be to keep things at the same float bit depth for both D1 and D2? (ie both float32 or both float64, depending on input dtype).

jni (approved):

jni:
I agree with @stefanv that ideally the skipped test should instead get added in a separate PR that fixes the issue, but I'm not that fussed if we instead just merge this as-is and remove the skip marker in a follow-up PR.

@scikit-image/core any others up for review?

Thank you @mark-boer!

mark-boer:
@jni, I changed the input data so it cast to a float of at least 32 bits. But if the user chooses to call the function with a float64, it is not cast down to a float32. If that doesn't make sense, we can also just forcibly cast it to a float64.

I will remove the failing test this evening and create a new PR with the fix to this failing test.

jni:
> I changed the input data so it cast to a float of at least 32 bits. But if the user chooses to call the function with a float64, it is not cast down to a float32. If that doesn't make sense, we can also just forcibly cast it to a float64.

Makes perfect sense, actually! I think the fact that dtype was set for the second vstack and not the first confused @stefanv upon reading the code. Took me a while as well after reading his comment. Perhaps adding a dtype= to the vstack calls, even if redundant, will make the code clearer? Or maybe just add a comment to the effect of what you just wrote?

mark-boer:
The `dtype`, is actually an argument of the `np.ones`, even though this is not really necessary , I thought that saved an additional cast.

skimage/measure/fit.py
```diff
@@ -414,13 +417,16 @@ def estimate(self, data):
...
+        data = data.astype(float_type, copy=False)
+
         x = data[:, 0]
         y = data[:, 1]
 
         # Quadratic part of design matrix [eqn. 15] from [1]
         D1 = np.vstack([x ** 2, x * y, y ** 2]).T
         # Linear part of design matrix [eqn. 16] from [1]
-        D2 = np.vstack([x, y, np.ones(len(x))]).T
+        D2 = np.vstack([x, y, np.ones(len(x), dtype=float_type)]).T
```
rfezzani, line 429: What about
```suggestion
        D2 = np.vstack([x, y, np.ones_like(x)]).T
```

mark-boer: That's even better :-)

skimage/measure/tests/test_fit.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
```
stefanv, line 2:
```suggestion
```

skimage/measure/tests/test_fit.py
```diff
@@ -211,14 +212,18 @@ def test_ellipse_model_estimate_from_data():
         [643, 926], [644, 975], [643, 655], [646, 705], [651, 664], [651, 984],
         [647, 665], [651, 715], [651, 725], [651, 734], [647, 809], [651, 825],
         [651, 873], [647, 900], [652, 917], [651, 944], [652, 742], [648, 811],
-        [651, 994], [652, 783], [650, 911], [654, 879]])
+        [651, 994], [652, 783], [650, 911], [654, 879]], dtype="int32")
```
stefanv, line 215:
```suggestion
        [651, 994], [652, 783], [650, 911], [654, 879]], dtype=np.int32)
```

skimage/measure/tests/test_fit.py
```diff
@@ -211,14 +212,18 @@ def test_ellipse_model_estimate_from_data():
...
 
     # estimate parameters of real data
     model = EllipseModel()
     model.estimate(data)
 
     # test whether estimated parameters are smaller then 1000, so means stable
-    assert_array_less(np.abs(model.params[:4]), np.array([2e3] * 4))
+    assert_array_less(model.params[:4], np.array([1e3] * 4))
```
stefanv, line 222:
```suggestion
    assert_array_less(model.params[:4], np.full(4, 1000))
```

stefanv (commented):
Thanks, this is very readable now @mark-boer.  I left some minor suggestions, but they're mostly stylistic.

jni (approved):
Just gonna approve this some more for good measure. =P

stefanv:
Thank you very much, @mark-boer

Issue #5159, opened 2020-12-27
Clarification: coordinate conventions in watershed example


harish-vnkt:
I have been frequently referring to [this watershed segmentation example](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_watershed.html#sphx-glr-auto-examples-segmentation-plot-watershed-py). The second centroid in the docs uses the naming convention `(x2, y2)` to represent the point `(44, 52)`. If you then use `mask[tuple(coords.T)]` to set labels to the centroids on the image, the labelled centroid is present at `row = 44` and `col = 52`, wheras the naming convention suggests that the indexing happen at `row = 52` and `col = 44` instead. 

Would it be better to name all coords `(r*, c*)` in the example, or is this the expected behaviour?

Issue #5087, opened 2020-11-29
Feature request: points to labels


jni:
## Description

I find myself repeating the following lines very often:

```python
coords = ... # something that gives me an (npoints, ndim) array
indices = tuple(np.transpose(np.round(coords).astype(int)))
labels = np.zeros_like(image, dtype=np.uint32)
labels[indices] = np.arange(1, len(coords) + 1)
```

This is required for example to take a set of centroids and turn them into markers for watershed.

I think it would be great for skimage to provide a utility function to do this.

emmanuelle:
Good idea. Should it be in `segmentation` or in `utils`?

jni:
A great question. I don't have a strong preference. Maybe we reopen #1608? =P

harish-vnkt:
Hi, @jni. First timer in open source and would like to work on this issue. :) 
Would it be enough to have a function that takes centroid coordinates and the image as arguments and uses [ndi.label](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.label.html#scipy.ndimage.label) to assign labels to the centroids? Also, any idea on which module this function belongs to?

jni:
@harish-vnkt Hello! I think that approach will be slower than using the direct assignment in the above code snippet.

Regarding module, I'm not sure, nor do I know what the best name is. `skimage.util.points_to_labels`? `skimage.segmentation.points_to_label_image`? Some combination of the above? But, might as well open the PR and get the ball rolling! 😊

jni:
Closed by #5180

PR #5180, opened 2021-01-11
Function for assigning labels to points

harish-vnkt:
## Description

This is a function for the feature request at #5087. It takes an array of coordinates and an image as input and returns a mask where each coordinate is assigned a unique integer label. Thanks in advance for your feedbacks. :smile: 

## Questions/clarifications due to being a first-timer

- General design question - The function deals with coordinates on 2 axes only, i.e, fails if ```coords.shape[1] != 2```. Should this be extended to a general case where the coordinates can have arbitrary number of axes as long as the indexing works, i.e, passes as long as ```coords.shape[1] == image.ndim```? @jni 
- In this function, if the ```coords``` are ```[44, 52]```, the labelling happens at ```row=42``` and ```col=52```. Is this correct behaviour? I opened a related issue #5159 to clarify this for the [Watershed Segmentation example](https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_watershed.html#:~:text=The%20watershed%20is%20a%20classical,a%20local%20topography%20(elevation).)

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/util/points_to_labels.py
```diff
@@ -0,0 +1,45 @@
...
+           [0, 0, 0, 0, 0]], dtype=uint64)
+    """
+    if not isinstance(coords, np.ndarray):
+        raise TypeError("'coords' must be an ndarray")
+
+    if coords.shape[-1] != 2 or coords.ndim != 2:
+        raise ValueError("'coords' must be of shape (n, 2)")
+
+    h, w = image.shape[:2]
+    np_indices = tuple(np.transpose(np.round(coords).astype(np.int)))
```
grlee77, line 42:
```suggestion
    np_indices = tuple(c for c in np.transpose(np.round(coords)).astype(int))
```
This suggestion avoids conversion of the array to a tuple and instead creates a tuple containing coordinate array views (thus avoiding a copy of the array values). A "tuple of arrays" format like this is what functions like `np.nonzero` return and will work the same for the indexing used below.

skimage/util/points_to_labels.py
```diff
@@ -0,0 +1,45 @@
...
+           [0, 0, 0, 0, 0],
+           [0, 0, 0, 0, 0]], dtype=uint64)
+    """
+    if not isinstance(coords, np.ndarray):
+        raise TypeError("'coords' must be an ndarray")
+
+    if coords.shape[-1] != 2 or coords.ndim != 2:
+        raise ValueError("'coords' must be of shape (n, 2)")
+
+    h, w = image.shape[:2]
```
grlee77, line 41: If we want to support use of this function on multichannel and n-dimensional images, then we should probably add a `multichannel` keyword argument (defaulting to False). This way we don't have to try and guess based on shape if the image is multichannel. If this `multichannel` argument is present then, we can get the shape of the spatial dimensions using something like:
```Python
shape  = image.shape[:-1] if multichannel else image.shape
```

This can be moved above the shape check above so that you can check `if coords.shape[-1] != len(shape)` instead of assuming 2 dimensions.

grlee77 (changes requested):
Thanks @harish-vnkt, this looks like a great start.

I do think we will want to make this n-dimensional, but fortunately that seems fairly easy to do (see comment).

To avoid ambiguity, we should add a `Notes` section to the docstring, stating that pixels are considered to fall on integer coordinate locations (starting from zero).

Other considerations are what to do with:
1.) out of range coordinates (we could either explicitly check for this or just relying on NumPy to raise an `IndexError`. Either way, we should document the behavior add a test case for it)
2.) negative coordinates -> explicitly disallow?
Another alternative to raising errors, could be clipping indices to fall within the image shape.

emmanuelle:
Thank you @harish-vnkt ! In addition to @grlee77 's comments, could you also think of a gallery example which would show how to use this new function?

skimage/util/points_to_labels.py
jni: Huh? We shouldn't use multichannel here, all we need is the output shape. ie don't pass in an image as an argument, only the shape.

skimage/util/points_to_labels.py
```diff
@@ -0,0 +1,45 @@
...
+
+def label_points(coords, image):
+    """Assign unique integer labels to coordinates on an image mask
+
+    Parameters
+    ----------
+    coords: ndarray
+        An array of 2D coordinates with shape (n, 2)
+    image: ndarray
+        Image for which the mask is created
```
jni, line 14: The second argument should only be a shape, not an image

skimage/util/points_to_labels.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Returns
+    -------
+    label_mask: ndarray
+        A 2D mask of zeroes containing unique integer labels at the `coords`
```
jni, line 19: Remove 2D

skimage/util/points_to_labels.py
```diff
@@ -0,0 +1,45 @@
...
+           [0, 0, 0, 0, 0],
+           [0, 0, 2, 0, 0],
+           [0, 0, 0, 0, 0],
+           [0, 0, 0, 0, 0]], dtype=uint64)
+    """
+    if not isinstance(coords, np.ndarray):
+        raise TypeError("'coords' must be an ndarray")
+
+    if coords.shape[-1] != 2 or coords.ndim != 2:
+        raise ValueError("'coords' must be of shape (n, 2)")
```
jni, line 39: Remove these checks, instead check only that `coords.shape[1] == len(output_shape)`, and raise a value error saying the dimensionality of the points should match the dimensionality of the image.

skimage/util/points_to_labels.py
```diff
@@ -0,0 +1,45 @@
...
+
+    h, w = image.shape[:2]
+    np_indices = tuple(np.transpose(np.round(coords).astype(np.int)))
+    label_mask = np.zeros((h, w), dtype=np.uint64)
```
jni, line 43:
```suggestion
    label_mask = np.zeros(output_shape, dtype=np.uint64)
```

jni (changes requested):
@harish-vnkt as @grlee77 said, we do want this to be nD, but it's quite a quick fix! I think between his comments and mine we can get there!

jni:
@emmanuelle instead of a new gallery example, we can use this function in this code cell:

https://scikit-image.org/docs/dev/auto_examples/applications/plot_human_mitosis.html#segment-nuclei

rfezzani (commented):
Thank you @harish-vnkt for your contribution.
I think that the file name must be more generic and hidden: `skimage/util/_label.py`. This keeps door open to adding other util functions in the same vain (who knows :wink:).
The `label_points` function may also be exposed in the `skimage.util` module (by adding it to `skimage/util/__init__.py`).

harish-vnkt:
Hi guys, sorry for the late updates. I have accounted for all of the changes mentioned in the reviews. I have gone with @jni's suggestion for the output shape argument and updated the code and tests accordingly.

skimage/util/_label.py
```diff
@@ -0,0 +1,50 @@
...
+    Notes
+    -----
+    - The labels are assigned to coordinates that are converted to integer and considered to start from 0.
+    - Coordinates that are out of range of the mask raise an IndexError.
+    - Negative coordinates raise a ValueError
+    """
+    if coords.shape[1] != len(output_shape):
+        raise ValueError("Dimensionality of points should match the output shape")
+
+    if True in (coords < 0):
```
rfezzani, line 44:
```suggestion
    if np.any(coords < 0):
```

skimage/util/_label.py
```diff
@@ -0,0 +1,50 @@
...
+        raise ValueError("Coordinates should be positive and start from 0")
+
+    np_indices = tuple(np.transpose(np.round(coords).astype(np.int)))
```
rfezzani, line 47:
```suggestion
    np_indices = tuple(np.transpose(np.round(coords).astype(np.int, copy=False)))
```

skimage/util/_label.py
```diff
@@ -0,0 +1,50 @@
...
+    label_mask = np.zeros(output_shape, dtype=np.uint64)
```
rfezzani, line 48: `labels` is may be a better suited argument name...

skimage/util/_label.py
```diff
@@ -0,0 +1,50 @@
...
+    Parameters
+    ----------
+    coords: ndarray
+        An array of N coordinates with dimension D
+    output_shape: tuple
+        The shape of the mask on which `coords` are labelled
+
+    Returns
+    -------
+    label_mask: ndarray
```
rfezzani, line 18: As already mentioned, `labels` is may be a better argument name

rfezzani (commented):
Thank you @harish-vnkt, I left some comments, this is pretty close ;)

skimage/util/_label.py
```diff
@@ -0,0 +1,50 @@
...
+    >>> mask
+    array([[0, 1, 0, 0, 0],
+           [0, 0, 0, 0, 0],
+           [0, 0, 2, 0, 0],
+           [0, 0, 0, 0, 0],
+           [0, 0, 0, 0, 0]], dtype=uint64)
+
+    Notes
+    -----
+    - The labels are assigned to coordinates that are converted to integer and considered to start from 0.
```
rfezzani, line 37:
```suggestion
    - The labels are assigned to coordinates that are converted to
      integer and considered to start from 0.
```

skimage/util/_label.py
```diff
@@ -0,0 +1,50 @@
...
+    - Coordinates that are out of range of the mask raise an IndexError.
+    - Negative coordinates raise a ValueError
+    """
+    if coords.shape[1] != len(output_shape):
+        raise ValueError("Dimensionality of points should match the output shape")
```
rfezzani, line 42:
```suggestion
        raise ValueError("Dimensionality of points should match the "
                         "output shape")
```

skimage/util/_label.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if np.any(coords < 0):
+        raise ValueError("Coordinates should be positive and start from 0")
+
+    np_indices = tuple(np.transpose(np.round(coords).astype(np.int, copy=False)))
```
rfezzani, line 47:
```suggestion
    np_indices = tuple(np.transpose(np.round(coords).astype(np.int,
                                                            copy=False)))
```

rfezzani (approved):
Thank you @harish-vnkt, I left some minor corrections to PEP8, otherwise it looks good for me :tada:

jni (approved):
Awesome! This looks ready to me. Thank you @harish-vnkt!

PR #5181, opened 2021-01-11
DOC: Fixup to linspace in example

madphysicist:
Right now the data is not sampled every half degree. Also, angles -pi/2 to pi/2 are equivalent for the purposes of the transform.

## Description

Extremely minor fixup, for technical correctness. No API changes, just doc example.

## For reviewers

  later.

### DIFF
--- a/doc/examples/edges/plot_line_hough_transform.py
+++ b/doc/examples/edges/plot_line_hough_transform.py
@@ -79,7 +79,7 @@ image[idx, idx] = 255
 
 # Classic straight-line Hough transform
 # Set a precision of 0.5 degree.
-tested_angles = np.linspace(-np.pi / 2, np.pi / 2, 360)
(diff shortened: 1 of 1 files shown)


grlee77 (approved):
Thanks. 

This makes the spacing exactly at 0.5, starting from -90 degrees and ending at 89.5 degrees. This seems like the preferable solution to incrementing 360 to 361 which would also give spacing 0.5, but end at +90 degrees, which seems duplicate with -90 degrees for line segments.

madphysicist (commented):
Why are the angles in all the examples flipped? E.g. on line 94, why `np.rad2deg(theta[-1]), np.rad2deg(theta[0])` rather than `np.rad2deg(theta[0]), np.rad2deg(theta[-1])`?

sciunto (approved):

emmanuelle (approved):

emmanuelle:
Thank you @madphysicist !

PR #5182, opened 2021-01-11
Minor fixups to Hough line transform code and examples

madphysicist:
## Description

- Without `endpoint=False`, the default ranges include both -pi/2 and +pi/2, which is redundant. The spacing is also not one degree (see #5181)
- `theta = np.pi / 2 - np.arange(180) / 180.0 * np.pi` is equivalent to `linspace(np.pi / 2, -np.pi / 2, 180, endpoint=False)`. This is backwards from what is advertised. Also, a call to `linspace` is likely easier to read than the `arange` expression.
- Why swap the x-extents on the plots? Let's take a look:

    ```
    img = np.zeros((100, 100), bool)
    img[50] = True
    img[np.arange(75), np.arange(25, 100)] = True

    tested_angles = np.linspace(-np.pi / 2, np.pi / 2, 3600, endpoint=False)
    h, theta, d = hough_line(img, theta=tested_angles)
    fig, ax = plt.subplots(1, 2)
    ax[0].imshow(np.log(1 + h),
               extent=[np.rad2deg(theta[-1]), np.rad2deg(theta[0]), d[-1], d[0]],
               cmap='gray')
    ax[1].imshow(np.log(1 + h),
             extent=[np.rad2deg(theta[0]), np.rad2deg(theta[-1]), d[-1], d[0]],
             cmap='gray')
    for _, angle, dist in zip(*hough_line_peaks(h, theta, d)):
        ax[0].plot(np.rad2deg(angle), dist, 'ro')
        ax[1].plot(np.rad2deg(angle), dist, 'ro')
    ax[0].set_title('Current (wrong)')
    ax[1].set_title('Proposed (correct)')
    plt.show()
    ```
    image

    It's safe to say that the version of image extent that lets you plot (angle, dist) correctly is the right one.

## For reviewers

  later.

### DIFF
--- a/doc/examples/edges/plot_line_hough_transform.py
+++ b/doc/examples/edges/plot_line_hough_transform.py
@@ -65,6 +65,7 @@ import numpy as np
 
 from skimage.transform import hough_line, hough_line_peaks
 from skimage.feature import canny
+from skimage.draw import line
 from skimage import data
 
 import matplotlib.pyplot as plt
@@ -74,8 +75,9 @@ from matplotlib import cm
 # Constructing test image
 image = np.zeros((200, 200))
 idx = np.arange(25, 175)
-image[idx[::-1], idx] = 255
 image[idx, idx] = 255
+image[line(45, 25, 25, 175)] = 255
+image[line(25, 135, 175, 155)] = 255
 
 # Classic straight-line Hough transform
 # Set a precision of 0.5 degree.
--- a/doc/examples/edges/plot_line_hough_transform.py
+++ b/doc/examples/edges/plot_line_hough_transform.py
@@ -30,7 +30,7 @@ appropriately.
 
 We can think of each non-zero pixel "voting" for potential line candidates. The
 local maxima in the resulting histogram indicates the parameters of the most
-probably lines. In our example, the maxima occur at 45 and 135 degrees,
+probable lines. In our example, the maxima occur at 45 and 135 degrees,
 corresponding to the normal vector angles of each line.
 
 Another approach is the Progressive Probabilistic Hough Transform [2]_. It is
--- a/doc/examples/edges/plot_line_hough_transform.py
+++ b/doc/examples/edges/plot_line_hough_transform.py
@@ -91,7 +91,7 @@ ax[0].set_title('Input image')
 ax[0].set_axis_off()
 
 ax[1].imshow(np.log(1 + h),
-             extent=[np.rad2deg(theta[-1]), np.rad2deg(theta[0]), d[-1], d[0]],
+             extent=[np.rad2deg(theta[0]), np.rad2deg(theta[-1]), d[-1], d[0]],
              cmap=cm.gray, aspect=1/1.5)
 ax[1].set_title('Hough transform')
 ax[1].set_xlabel('Angles (degrees)')
--- a/doc/source/plots/hough_tf.py
+++ b/doc/source/plots/hough_tf.py
@@ -21,7 +21,7 @@ axes[0].set_title('Input image')
 
 axes[1].imshow(
     out, cmap=plt.cm.bone,
-    extent=(np.rad2deg(angles[-1]), np.rad2deg(angles[0]), d[-1], d[0]))
+    extent=(np.rad2deg(angles[0]), np.rad2deg(angles[-1]), d[-1], d[0]))
 axes[1].set_title('Hough transform')
 axes[1].set_xlabel('Angle (degree)')
 axes[1].set_ylabel('Distance (pixel)')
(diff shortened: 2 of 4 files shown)


madphysicist:
Should I squash the commits?

sciunto (changes requested):
Thanks for the PR! No need to squash, we will do it. Perhaps, to improve the example, it would be better to have two lines at different absolute angles like 45° and -30°. What do you think?

madphysicist:
@sciunto I'm not sure that I understand what example you're referring to? My goal was to fix some small imprecisions that I noticed rather than to add anything fundamentally new. I'd be happy to add something if you like though.

sciunto:
> @sciunto I'm not sure that I understand what example you're referring to? 

The one you changed: https://scikit-image.org/docs/stable/auto_examples/edges/plot_line_hough_transform.html#sphx-glr-auto-examples-edges-plot-line-hough-transform-py

madphysicist:
@sciunto. I made the example look like this:

image

Three lines, but I think it's in keeping with what you requested.

madphysicist:
Looks like the build errors are caused by a timeout rather than an actual error

madphysicist:
If we can agree on this PR, please merge before #5183.

doc/examples/color_exposure/plot_tinting_grayscale_images.py
```diff
@@ -60,7 +60,8 @@
 
 fig, ax = plt.subplots(figsize=(5, 2))
 # Set image extent so hues go from 0 to 1 and the image is a nice aspect ratio.
-ax.imshow(all_hues, extent=(0, 1, 0, 0.2))
+ax.imshow(all_hues, extent=(0 - 0.5 / len(hue_gradient),
+                            1 + 0.5 / len(hue_gradient), 0, 0.2))
```
sciunto, line 64: This change seems unrelated to this PR. Please revert.

doc/examples/numpy_operations/plot_view_as_blocks.py
```diff
@@ -48,7 +48,7 @@
 
 l_resized = ndi.zoom(l, 2, order=3)
 ax[0].set_title("Original rescaled with\n spline interpolation (order=3)")
-ax[0].imshow(l_resized, extent=(0, 128, 128, 0),
+ax[0].imshow(l_resized, extent=(-0.5, 128.5, 128.5, -0.5),
```
sciunto, line 51: This change seems unrelated to this PR. Please revert.

doc/examples/transform/plot_radon_transform.py
```diff
@@ -75,11 +75,13 @@
...
 theta = np.linspace(0., 180., max(image.shape), endpoint=False)
 sinogram = radon(image, theta=theta, circle=True)
+dx, dy = 0.5 * 180.0 / max(image.shape), 0.5 / sinogram.shape[0]
 ax2.set_title("Radon transform\n(Sinogram)")
 ax2.set_xlabel("Projection angle (deg)")
 ax2.set_ylabel("Projection position (pixels)")
 ax2.imshow(sinogram, cmap=plt.cm.Greys_r,
-           extent=(0, 180, 0, sinogram.shape[0]), aspect='auto')
+           extent=(-dx, 180.0 + dx, -dy, sinogram.shape[0] + dy),
+           aspect='auto')
```
sciunto, line 84: This change seems unrelated to this PR. Please revert.

skimage/viewer/plugins/color_histogram.py
```diff
@@ -27,7 +27,7 @@ def _on_new_image(self, image):
         # Calculate color histogram in the Lab colorspace:
         L, a, b = self.lab_image.T
         left, right = -100, 100
-        ab_extents = [left, right, right, left]
+        ab_extents = [left - 0.5, right + 0.5, right + 0.5, left - 0.5]
```
sciunto, line 30: This change seems unrelated to this PR. Please revert. (And this module is deprecated, will be removed soon, so no need to maintain this code).

sciunto (changes requested):

madphysicist:
@sciunto I've moved the updates to #5184. The changes were keeping in the spirit of what I am trying to do here, but I can see that they are easier to process in a separate PR.

jni (approved):
:+1: nice!

sciunto (approved):
perfect!

sciunto:
thank you

stefanv:
@madphysicist @sciunto Does this mean we're finally ready for https://github.com/scikit-image/scikit-image/pull/4877 ? :)

madphysicist:
@stefanv This is not totally related, but sure.

PR #5188, opened 2021-01-18
Add nD support to several geometric transform classes

jni:
This is the second PR with more gradual changes from #3544. I've pulled out the
automatic conversion of 1D parameter vectors to transforms. Viewed in
isolation, that was definitely way too magical given our values. I've left the
implementation in as a private function for future reference. Eventually, I'd
like to add a class method, `from_parameter_vector`, that would be
well-documented for each class, but I think that should come in a future PR.

I've done my best with the Euler rotation matrix code, but that would benefit
the most from review. Since we don't have a strict xyz/zyx axis order, it's
basically ambiguous whether XYZ or ZYX Euler angles are implemented (if it's
either of those :joy:). CC @stefanv @grlee77 @matthew-brett

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -1056,9 +1056,10 @@ def _euler_rotation(axis, angle):
         The rotation matrix along axis `axis`.
     """
     i = axis
-    s, c = np.sin(angle), np.cos(angle)
-    R2 = np.array([[          c, (-1)**(i+1) * s],
-                   [(-1)**i * s,               c]])
+    s = (-1)**i * np.sin(angle)
+    c = np.cos(angle)
+    R2 = np.array([[c, -s],
+                   [s,  c]])
     Ri = np.eye(3)
     # We need the axes other than the rotation axis, in the right order:
     # 0 -> (1, 2); 1 -> (0, 2); 2 -> (0, 1).
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -19,7 +19,7 @@ def _affine_matrix_from_vector(v):
         v, (dimensionality, dimensionality + 1)
     )
     matrix = np.concatenate(
-        (part_matrix, np.eye(dimensionality + 1)[-1:]), axis=0
+        (part_matrix, np.eye(1, dimensionality + 1, dimensionality)), axis=0
     )
     return matrix
 
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -1067,7 +1067,7 @@ def _euler_rotation(axis, angle):
     return Ri
 
 
-def _euler_rotation_matrix(angles):
+def _euler_rotation_matrix(angles, axes=None):
     """Produce an Euler rotation matrix from the given angles.
 
     The matrix will have dimension equal to the number of angles given.
@@ -1076,15 +1076,19 @@ def _euler_rotation_matrix(angles):
     ----------
     angles : array of float, shape (3,)
         The transformation angles in radians.
+    axes : list of int
+        The axes about which to produce the rotation. Defaults to 0, 1, 2.
 
     Returns
     -------
     R : array of float, shape (3, 3)
         The Euler rotation matrix.
     """
+    if axes is None:
+        axes = list(range(3))
     dim = len(angles)
     R = np.eye(dim)
-    for i, angle in enumerate(angles):
+    for i, angle in zip(axes, angles):
         R = R @ _euler_rotation(i, angle)
     return R
 
@@ -1275,27 +1279,22 @@ class SimilarityTransform(EuclideanTransform):
                 self.params = matrix
                 dimensionality = matrix.shape[0] - 1
         if params:
-            if dimensionality == 2:
-                axes = ((0, 1),)
-            elif dimensionality == 3:
-                axes = ((1, 2), (0, 1), (1, 2))  # XZX Euler angles
-            else:
+            if dimensionality not in (2, 3):
                 raise ValueError('Parameters only supported for 2D and 3D.')
             matrix = np.eye(dimensionality + 1, dtype=float)
             if scale is None:
                 scale = 1
             if rotation is None:
-                rotation = (0,) if dimensionality == 2 else (0, 0, 0)
-            if np.isscalar(rotation):
-                rotation = [rotation]
+                rotation = 0 if dimensionality == 2 else (0, 0, 0)
             if translation is None:
                 translation = (0,) * dimensionality
-            for rot, ax in zip(rotation, axes):
-                R = np.eye(dimensionality + 1)
-                c, s = np.cos(rot), np.sin(rot)
-                R[ax, ax] = c
-                R[ax, ax[::-1]] = -s, s
-                matrix = R @ matrix
+            if dimensionality == 2:
+                ax = (0, 1)
+                c, s = np.cos(rotation), np.sin(rotation)
+                matrix[ax, ax] = c
+                matrix[ax, ax[::-1]] = -s, s
+            else:  # 3D rotation
+                matrix[:3, :3] = _euler_rotation_matrix(rotation)
 
             matrix[:dimensionality, :dimensionality] *= scale
             matrix[:dimensionality, dimensionality] = translation
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -9,9 +9,9 @@ from .._shared.utils import get_bound_method_class, safe_as_int
 def _affine_matrix_from_vector(v):
     """Affine matrix from linearized (d, d + 1) matrix entries."""
     nparam = v.size
-    # solve for d in: d * (d - 1) = nparam
+    # solve for d in: d * (d + 1) = nparam
     d = (1 + np.sqrt(1 + 4 * nparam)) / 2 - 1
-    dimensionality = int(d)
+    dimensionality = int(np.round(d))  # round to prevent approx errors
     if d != dimensionality:
         raise ValueError('Invalid number of elements for '
                          f'linearized matrix: {nparam}')
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -62,6 +62,10 @@ def _center_and_normalize_points(points):
 
     # if all the points are the same, the transformation matrix cannot be
     # created. We return an equivalent matrix with np.nans as sentinel values.
+    # This obviates the need for try/except blocks in functions calling this
+    # one, and those are only needed when actual 0 is reached, rather than some
+    # small value; ie, we don't need to worry about numerical stability here,
+    # only actual 0.
     if rms == 0:
         return np.full((d + 1, d + 1), np.nan), np.full_like(points, np.nan)
 
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -16,7 +16,7 @@ def _affine_matrix_from_vector(v):
         raise ValueError('Invalid number of elements for '
                          f'linearized matrix: {nparam}')
     matrix = np.eye(dimensionality + 1)
-    matrix[-1, :] = np.reshape(v, (dimensionality, dimensionality + 1))
+    matrix[:-1, :] = np.reshape(v, (dimensionality, dimensionality + 1))
     return matrix
 
 
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -6,6 +6,24 @@ import textwrap
 from .._shared.utils import get_bound_method_class, safe_as_int
 
 
+def _affine_matrix_from_vector(v):
+    """Affine matrix from linearized (d, d + 1) matrix entries."""
+    nparam = v.size
+    # solve for d in: d * (d - 1) = nparam
+    d = (1 + np.sqrt(1 + 4 * nparam)) / 2 - 1
+    dimensionality = int(d)
+    if d != dimensionality:
+        raise ValueError('Invalid number of elements for '
+                         f'linearized matrix: {nparam}')
+    part_matrix = np.reshape(
+        v, (dimensionality, dimensionality + 1)
+    )
+    matrix = np.concatenate(
+        (part_matrix, np.eye(dimensionality + 1)[-1:]), axis=0
+    )
+    return matrix
+
+
 def _center_and_normalize_points(points):
     """Center and normalize image points.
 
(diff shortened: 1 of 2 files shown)


jni:
The CI failure appears to be an un-related hang.

skimage/transform/_geometric.py
```diff
@@ -230,13 +256,20 @@ class FundamentalMatrixTransform(GeometricTransform):
...
-    def __init__(self, matrix=None):
+    def __init__(self, matrix=None, *, dimensionality=2):
         if matrix is None:
             # default to an identity transform
-            matrix = np.eye(3)
-        if matrix.shape != (3, 3):
+            matrix = np.eye(dimensionality + 1)
+        else:
+            dimensionality = matrix.shape[0] - 1
+        if matrix.shape != (dimensionality + 1, dimensionality + 1):
```
rfezzani, line 265: Shouldn't `dimensionality` be deduced from `matrix.shape` if `matrix is not None`?
I mean, when matrix is provided, we can assume that the user knows the `dimensionality`, we don't need to force him to give it as argument if `dimensionality != 2`...
I think `dimensionality` should be `None` by default...

rfezzani: Well, sorry, I need one more coffee this morning :sweat_smile:...

BTW, this situation can never happen since `dimensionality` is overwritten when `matrix is not None`.
```suggestion
```

rfezzani: Waouh, I need a bucket of coffee...

jni: :joy: I actually checked the diff cover when developing tests here so I'm fairly confident that all lines are reachable. :smiley:

skimage/transform/_geometric.py
```diff
@@ -7,6 +7,24 @@
...
+    d = (1 + np.sqrt(1 + 4 * nparam)) / 2 - 1
+    dimensionality = int(d)
+    if d != dimensionality:
+        raise ValueError('Invalid number of elements for '
+                         f'linearized matrix: {nparam}')
+    part_matrix = np.reshape(
+        v, (dimensionality, dimensionality + 1)
+    )
+    matrix = np.concatenate(
+        (part_matrix, np.eye(dimensionality + 1)[-1:]), axis=0
```
stefanv, line 23: Easier to just pop the one into place?

skimage/transform/_geometric.py
```diff
@@ -39,24 +59,30 @@ def _center_and_normalize_points(points):
...
     """
-
+    n, d = points.shape
     centroid = np.mean(points, axis=0)
 
-    rms = math.sqrt(np.sum((points - centroid) ** 2) / points.shape[0])
+    centered = points - centroid
+    rms = np.sqrt(np.sum(centered ** 2) / n)
+
+    if rms == 0:
```
stefanv, line 68: I guess this point could come pretty close to zero by itself, so is it worth doing the explicit check, or should we just let the algorithm run its course?

skimage/transform/_geometric.py
```diff
@@ -641,37 +686,25 @@ def estimate(self, src, dst):
...
-        A[:rows, 7] = - xd * ys
-        A[rows:, 3] = xs
-        A[rows:, 4] = ys
-        A[rows:, 5] = 1
-        A[rows:, 6] = - yd * xs
-        A[rows:, 7] = - yd * ys
-        A[:rows, 8] = xd
-        A[rows:, 8] = yd
+        A = np.zeros((n * d, (d+1) ** 2))
+        for ddim in range(d):
```
stefanv, line 699: This may now be worth a comment on its origin.  The first version was more obvious where it came from.

skimage/transform/_geometric.py
```diff
@@ -970,8 +1042,60 @@ def inverse(self, coords):
...
+    -------
+    Ri : array of float, shape (3, 3)
+        The rotation matrix along axis `axis`.
+    """
+    i = axis
+    s, c = np.sin(angle), np.cos(angle)
+    R2 = np.array([[          c, (-1)**(i+1) * s],
+                   [(-1)**i * s,               c]])
+    Ri = np.eye(3)
+    axes = sorted({0, 1, 2} - {axis})
```
stefanv, line 1065: Some black magic... another good candidate for a comment!

stefanv (commented):
The Guardian of the Dimensions Returns :shield:!

I appreciate you making this an incremental step; it would benefit from some implementation notes.  But the implementation itself looks sound.

Thank you!

skimage/transform/_geometric.py
rfezzani: Or should we define a tolerance parameter for numerical stability?
```python
if rms < tol:
    ...
```

stefanv: `tol` is hard to specify, so I'd say "no, if the algorithm produces nans that pass through anyway" or "yes, if @jni says we need to force the output to have a certain context".  Either way, there may be better ways to check than looking at the RMS---like looking at the condition number of the matrix being solved.

rfezzani: OK, after reading more closely, I think that @jni want to avoid these `try... except...` blocks [here](https://github.com/scikit-image/scikit-image/blob/master/skimage/transform/_geometric.py#L306) and [here](https://github.com/scikit-image/scikit-image/blob/master/skimage/transform/_geometric.py#L645).

skimage/transform/_geometric.py
```diff
@@ -7,6 +7,24 @@
                              _to_ndimage_mode)
 
 
+def _affine_matrix_from_vector(v):
+    """Affine matrix from linearized (d, d + 1) matrix entries."""
+    nparam = v.size
+    # solve for d in: d * (d - 1) = nparam
+    d = (1 + np.sqrt(1 + 4 * nparam)) / 2 - 1
```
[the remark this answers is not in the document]
jni: You're right, that should have been `d * (d + 1)` — the extra column for the translation parameters.

skimage/transform/_geometric.py
```diff
@@ -7,6 +7,24 @@
...
+    dimensionality = int(d)
```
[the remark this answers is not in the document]
jni: Well, there's no such thing as 9 parameters for the affine... It's 6 (2D), 12 (3D), 20 (4D), ... And anyway you've made a mistake typing stuff in somewhere because I get 4.0 for n=20 and 3.88748 for 19?

*But*, it's totally true that one should never trust floating point arithmetic to be exact. I'll use np.round here.

skimage/transform/_geometric.py
jni: :shrug: The alternative code is:

```python
matrix = np.concatenate(
    (part_matrix, np.zeros(dimensionality + 1), axis=0
)
matrix[-1, -1] = 1
```

So, you get an extra line of code, for something maybe *marginally* clearer? I agree that what I'm after really is the vector commonly called e_i. But I don't think that exists in NumPy. =)

skimage/transform/_geometric.py
jni: > OK, after reading more closely, I think that @jni wants to avoid these try... except... blocks

bingo! :blush: And those are only triggered when actual 0 is reached. I'll add a comment.

jni:
Ok @stefanv I think I've addressed all your questions! Thanks for the review! :tada:

skimage/transform/_geometric.py
rfezzani: It actually exists: what you are looking for is `np.eye(1, dimensionality + 1, dimensionality)` :wink:
```python
>>> np.eye(1, 4, 3)
array([[0., 0., 0., 1.]])
```

skimage/transform/_geometric.py
```diff
@@ -970,8 +1047,64 @@ def inverse(self, coords):
...
+
+    Returns
+    -------
+    Ri : array of float, shape (3, 3)
+        The rotation matrix along axis `axis`.
+    """
+    i = axis
+    s, c = np.sin(angle), np.cos(angle)
+    R2 = np.array([[          c, (-1)**(i+1) * s],
+                   [(-1)**i * s,               c]])
```
rfezzani, line 1068:
```suggestion
    s, c = ((-1) ** i) * np.sin(angle), np.cos(angle)
    R2 = np.array([[c, -s],
                   [s, c]])
```

skimage/transform/_geometric.py
```diff
@@ -970,8 +1047,64 @@ def inverse(self, coords):
...
+    Ri = np.eye(3)
+    # We need the axes other than the rotation axis, in the right order:
+    # 0 -> (1, 2); 1 -> (0, 2); 2 -> (0, 1).
+    axes = sorted({0, 1, 2} - {axis})
+    # We then embed the 2-axis rotation matrix into the full matrix.
+    # (1, 2) -> R[1:3:1, 1:3:1] = R2, (0, 2) -> R[0:3:2, 0:3:2] = R2, etc.
+    sl = slice(axes[0], axes[1] + 1, axes[1] - axes[0])
```
rfezzani, line 1075: Give me more magic
```suggestion
    args = [None, None, None]
    args[-axis] = 1 + ( axis > 0)
    sl = slice(*args)
```
:smiling_imp:

skimage/transform/_geometric.py
```diff
@@ -39,24 +59,32 @@ def _center_and_normalize_points(points):
...
 
-    matrix = np.array([[norm_factor, 0, -norm_factor * centroid[0]],
-                       [0, norm_factor, -norm_factor * centroid[1]],
-                       [0, 0, 1]])
+    part_matrix = norm_factor * np.concatenate(
+            (np.eye(d), -centroid[:, np.newaxis]), axis=1
+            )
+    matrix = np.concatenate(
+            (part_matrix, [[0,] * d + [1]]), axis=0
+            )
```
grlee77, line 80:
```suggestion
matrix = np.eye(d + 1)
matrix[:-1, -1] = -centroid[:, np.newaxis]
```

skimage/transform/_geometric.py
```diff
@@ -230,13 +258,20 @@ class FundamentalMatrixTransform(GeometricTransform):
...
+    def __init__(self, matrix=None, *, dimensionality=2):
         if matrix is None:
             # default to an identity transform
-            matrix = np.eye(3)
-        if matrix.shape != (3, 3):
+            matrix = np.eye(dimensionality + 1)
+        else:
+            dimensionality = matrix.shape[0] - 1
+        if matrix.shape != (dimensionality + 1, dimensionality + 1):
             raise ValueError("Invalid shape of transformation matrix")
```
grlee77, line 268: Can put this shape check in the else case. It will be true by construction for the matrix=None case.

```suggestion
            if matrix.shape != (dimensionality + 1, dimensionality + 1):
                raise ValueError("Invalid shape of transformation matrix")
```

skimage/transform/_geometric.py
```diff
@@ -753,45 +794,74 @@ class AffineTransform(ProjectiveTransform):
...
+    factor) transforms, it is possible to specify 3D transforms using implicit
+    parameters also.
+
     Parameters
     ----------
+    matrix : (D+1, D+1) array, optional
+        Homogeneous transformation matrix. If this matrix is provided, it is an
+        error to provide any of scale, rotation, shear, or translation.
     matrix : (3, 3) array, optional
         Homogeneous transformation matrix.
```
grlee77, line 813: remove duplicate entry

```suggestion

```

skimage/transform/_geometric.py
grlee77: That does look correct, but is pretty obscure (thus the devil emoji, I suppose!).

sorted `axes` could also be created with a generator expression:
```Python
# Get the (sorted) axes of the plane in which rotation occurs
ax1, ax2 = (ax for ax in range(3) if ax != axis)
# We then embed the 2-axis rotation matrix into the full matrix. 
sl = slice(ax1, ax2 + 1, ax2 - ax1)
```

skimage/transform/_geometric.py
```diff
@@ -970,8 +1047,64 @@ def inverse(self, coords):
...
+    axes : list of int
+        The axes about which to produce the rotation. Defaults to 0, 1, 2.
+
+    Returns
+    -------
+    R : array of float, shape (3, 3)
+        The Euler rotation matrix.
+    """
+    if axes is None:
+        axes = list(range(3))
```
grlee77, line 1098:
```suggestion
        axes = tuple(range(3))
```
It doesn't look like the list is modified later, so may as well use a tuple

skimage/transform/_geometric.py
```diff
@@ -1092,52 +1255,63 @@ class SimilarityTransform(EuclideanTransform):
...
         Homogeneous transformation matrix.
     scale : float, optional
-        Scale factor.
+        Scale factor. Implemented only for 2D and 3D.
     rotation : float, optional
         Rotation angle in counter-clockwise direction as radians.
-    translation : (tx, ty) as array, list or tuple, optional
-        x, y translation parameters.
+        Implemented only for 2D and 3D. For 3D, this is given in XZX Euler
+        angles.
```
grlee77, line 1265: XZX is not right here is it? It looks like we use axes in the order 0, 1, 2 in _euler_matrix which should be XYZ (assuming axis=0 means X?). XZX is apparently one of 12 possible conventions, but it didn't seem to be the one used here.

Another form for 3D that I find fairly intuitive is where the user specifies a unit vector corresponding to the [axis of rotation and then a single rotation angle about that vector. There is an [explicit formula for that form](https://en.wikipedia.org/wiki/Rotation_matrix#Rotation_matrix_from_axis_and_angle).

grlee77 (commented):
This is looking pretty good to me. I have left some minor comments

skimage/transform/_geometric.py
jni: > XZX is not right here is it?

Ah, yes, XZX is left over from an older version of the code. Will fix.

> Another form for 3D that I find fairly intuitive is where the user specifies a unit vector corresponding to the axis of rotation and then a single rotation angle about that vector.

Could change to that if people think that's better...

skimage/transform/_geometric.py
jni: @stefanv done, see comment above this line.

skimage/transform/_geometric.py
jni: done

skimage/transform/_geometric.py
jni: done, thanks for the suggestion @rfezzani, this is indeed clearer. =)

skimage/transform/_geometric.py
jni: Imma keep my code + comments, thanks both. :joy:

skimage/transform/_geometric.py
jni: See option c :joy: 

https://twitter.com/ashwintsrinath/status/1339001641752801280

jni: actually, we only iterate over `axes` later, so I'm just gonna leave it as `range()`!

skimage/transform/_geometric.py
jni: @grlee77 since the 3D Euler angles and the 3D vector + rotation are unambiguously distinguishable, I'm going to leave it as 3D Euler angles for now (fixed the docstring), and we can revisit in the future if someone clamors for alternate parameterizations.

skimage/transform/_geometric.py
jni: @grlee77 this turns out to be incorrect — see `norm_factor *` above. Hooray for tests and git-bisect! :blush: Reverting this change.

jni:
@scikit-image/core I believe I've addressed all the review comments — please have another look! :tada:

rfezzani (approved):
Looks good to me. Thank you @jni :tada:

jni:
Note, the optional deps failure is from #5266 and unrelated to this PR.

stefanv:
Nice work.  Thank you @jni!

PR #5196, opened 2021-01-25
Update structure_tensor docstring to include per-axis sigma.

mkcor:
## Description

@grlee77 I also included sanity checks for `sigma`. This contribution addresses https://github.com/scikit-image/scikit-image/pull/5012#discussion_r562884140.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


grlee77 (approved):
Thanks, this looks good. 

The test function could potentially be refactored to use `@pytest.mark.parametrize('ndim', [2, 3])` to reduce duplication slightly, but I think it is fine as is.

skimage/feature/corner.py
```diff
@@ -116,6 +119,14 @@ def structure_tensor(image, sigma=1, mode='constant', cval=0, order=None):
         else:
             order = 'rc'
 
+    if isinstance(sigma, tuple):
+        if len(sigma) <= 1:
+            raise TypeError('single sigma must be passed as float or int')
+        else:
+            if len(sigma) != image.ndim:
+                raise ValueError('sigma must have as many elements as image '
+                                 'has axes')
```
rfezzani, line 128: I think that the special case of length 1 iterable is not that special and that we may allow passing array_like values and not only tuples
```suggestion
    if not np.isscalar(sigma):
        sigma = tuple(sigma)
        if len(sigma) != image.ndim:
            raise ValueError('sigma must have as many elements as image '
                             'has axes')
```

mkcor: Very true, thanks @rfezzani! I'll update the test.

mkcor:
> The test function could potentially be refactored to use `@pytest.mark.parametrize('ndim', [2, 3])` to reduce duplication slightly, but I think it is fine as is.

Done 585386836d73c91fc53664ff6987cb08b6b2f245. Thanks, @grlee77!

@rfezzani please re-review if you get a chance :)

skimage/feature/corner.py
```diff
@@ -58,9 +58,12 @@ def structure_tensor(image, sigma=1, mode='constant', cval=0, order=None):
     ----------
     image : ndarray
         Input image.
-    sigma : float, optional
+    sigma : float or tuple of float, optional
```
rfezzani, line 61:
```suggestion
    sigma : float or array_like, optional
```

skimage/feature/corner.py
```diff
@@ -58,9 +58,12 @@ def structure_tensor(image, sigma=1, mode='constant', cval=0, order=None):
...
         Standard deviation used for the Gaussian kernel, which is used as a
         weighting function for the local summation of squared differences.
+        If sigma is a tuple, its length must be equal to `image.ndim` and each
```
rfezzani, line 64:
```suggestion
        If sigma is an iterable, its length must be equal to `image.ndim` and each
```

rfezzani (commented):
Sorry for the delay @mkor :relaxed:. We just need to update the doc and we are ready to go I think :wink:

mkcor:
> Sorry for the delay @mkor relaxed. We just need to update the doc and we are ready to go I think wink

@rfezzani good catch! Thanks -- fixed. I have taken the liberty of editing your suggestion.

rfezzani (approved):

rfezzani:
Merging! Thank you @mkcor :wink:

Issue #5199, opened 2021-01-28
Most skimage.measure._moments module functions don't support float32 operations


rfezzani:
## Description

Most of `skimage.measure._moments` module functions only supports double precision float operations. It is even explicitly stated in some functions' documentation that the input must be `double` or `uint8` ([`moments`](https://scikit-image.org/docs/stable/api/skimage.measure.html?highlight=moments_hu#moments), [`moments_central`](https://scikit-image.org/docs/stable/api/skimage.measure.html?highlight=moments_hu#moments-central), [`moments_coords`](https://scikit-image.org/docs/stable/api/skimage.measure.html?highlight=moments_hu#moments-coords)...). I don't understand why :confused:.

rfezzani:
Related to #4308 :sweat_smile:

grlee77:
Also related to #5205. @rfezzani, I just merged your `moments_hu` PR. Are you planning to work on the other moments functions? (I can spend some time this week looking at float32 stuff again, but don't want to duplicate effort)

rfezzani:
Hi @grlee77, and sorry for my late answer, I am too busy these days :confused:. I didn't planned to spend time on this topic, you can go ahead if you wish :wink:

PR #5200, opened 2021-01-28
Add float32 support to moments_hu

rfezzani:
## Description

Related to #5199.
Fused types are used to allow processing single precision values.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_moments.py
+++ b/skimage/measure/_moments.py
@@ -345,7 +345,8 @@ def moments_hu(nu):
     array([7.45370370e-01, 3.51165981e-01, 1.04049179e-01, 4.06442107e-02,
            2.64312299e-03, 2.40854582e-02, 4.33680869e-19])
     """
-    return _moments_cy.moments_hu(nu.astype(np.double))
+    dtype = np.float32 if nu.dtype == 'float32' else np.float64
+    return _moments_cy.moments_hu(nu.astype(dtype, copy=False))
 
 
 def centroid(image):
--- a/skimage/measure/_moments_cy.pyx
+++ b/skimage/measure/_moments_cy.pyx
@@ -2,17 +2,26 @@
 #cython: nonecheck=False
 #cython: wraparound=False
 import numpy as np
+cimport numpy as cnp
+
+from .._shared.fused_numerics cimport np_floats
+cnp.import_array()
 
 
-def moments_hu(double[:, :] nu):
-    cdef double[::1] hu = np.zeros((7, ), dtype=np.double)
-    cdef double t0 = nu[3, 0] + nu[1, 2]
-    cdef double t1 = nu[2, 1] + nu[0, 3]
-    cdef double q0 = t0 * t0
-    cdef double q1 = t1 * t1
-    cdef double n4 = 4 * nu[1, 1]
-    cdef double s = nu[2, 0] + nu[0, 2]
-    cdef double d = nu[2, 0] - nu[0, 2]
+def moments_hu(np_floats[:, :] nu):
+    if np_floats is cnp.float32_t:
+        dtype = np.float32
+    else:
+        dtype = np.float64
+
+    cdef np_floats[::1] hu = np.zeros((7, ), dtype=dtype)
+    cdef np_floats t0 = nu[3, 0] + nu[1, 2]
+    cdef np_floats t1 = nu[2, 1] + nu[0, 3]
+    cdef np_floats q0 = t0 * t0
+    cdef np_floats q1 = t1 * t1
+    cdef np_floats n4 = 4 * nu[1, 1]
+    cdef np_floats s = nu[2, 0] + nu[0, 2]
+    cdef np_floats d = nu[2, 0] - nu[0, 2]
     hu[0] = s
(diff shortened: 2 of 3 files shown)


rfezzani:
I wonder if maintaining the cython implementation of `moments_hu` worth it :confused:.

I made some tests by implementing a pure python equivalent function:
```python
def py_moments_hu(x):
    offset = x.shape[1]
    _x = x.reshape(-1)

    x03, x12, x21, x30 = _x[3: 3 * offset + 1: offset - 1]
    t0 = x30 + x12
    t1 = x21 + x03
    q0 = t0 * t0
    q1 = t1 * t1

    x02, x11, x20 = _x[2: 2 * offset + 1: offset - 1]
    n4 = 4 * x11
    s = x20 + x02
    d = x20 - x02

    hu = np.empty((7, ), x.dtype)
    hu[0] = s
    hu[1] = d * d + n4 * x11

    hu[3] = q0 + q1
    hu[5] = d * (q0 - q1) + n4 * t0 * t1

    t0 *= q0 - 3 * q1
    t1 *= 3 * q0 - q1
    q0 = x30 - 3 * x12
    q1 = 3 * x21 - x03

    hu[2] = q0 * q0 + q1 * q1
    hu[4] = q0 * t0 + q1 * t1
    hu[6] = q1 * t0 - q0 * t1

    return hu
```

I then performed some timing:
```python
In [2]: x = np.random.rand(4, 4)                                                                                   

In [3]: %timeit py_moments_hu(x)                                                                    
7.3 µs ± 37.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [4]: from skimage.measure import moments_hu                                                           

In [5]: %timeit moments_hu(x)                                                                            
4.33 µs ± 20.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

Pure python version is as expected slower then cython implementation (1.69 times), but we are talking about µseconds here...

rfezzani:
This PR replicate #4307 (I forgot about it :sweat_smile:).

grlee77:
Thanks @rfezzani, is there any benefit to this PR over the prior one? (i.e. which would you like to have reviewed?).

I am also interested in improving preservation of float32 in general. In working on `cupyimg` I found quite a few places on the python side where scikit-image converts to float64 when not strictly necessary. I didn't see a currently open issue on this, so I will open one where we can list functions that should have single precision support and keep track of open PRs needing review.

rfezzani:
> is there any benefit to this PR over the prior one?

As you already noticed, #4307 use the `dtype` parameter but this one not. Choose the option you prefer, I will close the other :wink:

skimage/measure/_moments_cy.pyx
```diff
@@ -2,17 +2,26 @@
...
-    cdef double q1 = t1 * t1
-    cdef double n4 = 4 * nu[1, 1]
-    cdef double s = nu[2, 0] + nu[0, 2]
-    cdef double d = nu[2, 0] - nu[0, 2]
+
+def moments_hu(np_floats[:, :] nu):
+    if np_floats is cnp.float32_t:
+        dtype = np.float32
+    else:
+        dtype = np.float64
```
grlee77, line 15: Can you omit this as in the other PR you opened omits this and just use dtype=nu.dtype below

skimage/measure/_moments.py
```diff
@@ -345,7 +345,8 @@ def moments_hu(nu):
     array([7.45370370e-01, 3.51165981e-01, 1.04049179e-01, 4.06442107e-02,
            2.64312299e-03, 2.40854582e-02, 4.33680869e-19])
     """
-    return _moments_cy.moments_hu(nu.astype(np.double))
+    dtype = np.float32 if nu.dtype == 'float32' else np.float64
```
grlee77, line 348: This seems fine, but I wonder if we should explicitly raise on error on complex-valued inputs? I think there are probably a ton of other places where this is just assumed and not explicitly checked, thgouh. not sure it is worth it

```
if nu.dtype.kind = 'c':
    raise TValueError("complex-valued inputs are not suppoted.")
```

grlee77 (approved):
This looks good to me. I prefer this approach to the one in #4307 (I think it is cleaner without adding a new `dtype `argument to the API)

skimage/measure/_moments_cy.pyx
rfezzani: To have access to `nu.dtype` in #4307, `nu` is declared as `cnp.ndarray` in function signature, which is [discouraged]( http://docs.cython.org/en/latest/src/userguide/memoryviews.html#comparison-to-the-old-buffer-support).

jni (approved):
I'm ok with the PR as is. Thanks @rfezzani!

Yeah maybe we should use pure Python but this is working and is faster so 🤷

skimage/measure/_moments.py
rfezzani: I have no strict opinion on this, but most of skimage functions do not explicitly check for complex input values, I am not sure it worth it :confused:

rfezzani:
Thank you @grlee77 :tada:

Issue #5195, opened 2021-01-25
`filters.gaussian` does not work on 1D input


stefanv:
```
import numpy as np
from skimage import filters

x = np.arange(50)
y = filters.gaussian(np.arange(50), sigma=(1,))

np.all(y < 1e-10)  # True
```

sciunto:
There is actually two different aspects.
* 1D case was not handled. Fixed in a PR.
* Arrays are converted to float, and `preserve_range` must be use to manipulate this conversion in the original example.

michalkrawczyk:
Hi,
I was wondering if problem don't lies in img_as_float from convert_to_float function:
I run test for np.arrange(9) with that and:
- result on `filters.gaussian(np.arange(9), sigma=(1,))`
> [9.36517148e-20 1.69488592e-19 2.71560066e-19 3.79485270e-19
 4.87890978e-19 5.96296685e-19 7.04221889e-19 8.06293363e-19
 8.82130240e-19]
 - result on filters.gaussian(np.arange(9), sigma=1, preserve_range=True)
> [0.36378461 1.06325634 2.00469952 3.00013383 4.         4.99986617
 5.99530048 6.93674366 7.63621539]
 
 I might be wrong but it seems to be ok. Going deeper:
 - convert_to_float(np.arange(9), True)
> [0. 1. 2. 3. 4. 5. 6. 7. 8.]           # OK
- convert_to_float(np.arange(9), False)
> [5.42101086e-20 1.62630326e-19 2.71050543e-19 3.79470760e-19
 4.87890978e-19 5.96311195e-19 7.04731412e-19 8.13151629e-19
 9.21571847e-19]                    # <- What is That?

jni:
@michalkrawczyk, see this page:

https://scikit-image.org/docs/dev/user_guide/data_types.html

convert to float with preserve range = False renormalises based on the assumed range of the data type, as shown in the table on the top of that page. So in your case, you start with int64.

```
In [3]: 1 / np.iinfo(np.int64).max                                         
Out[3]: 1.0842021724855044e-19
```

That's not exactly your number but it's about right. I'm not sure why part of the int64 range is truncated here, but that's the gist.

michalkrawczyk:
@jni 
I guess I found why is truncated. See skimage.util.dtype.convert:

>     # signed/unsigned int -> float
>     if kind_out == 'f':
>         # use float type that can exactly represent input integers
[...]

Issue #2613, opened 2017-04-13
SKIP: Consistency of `multichannel` parameter handling


grlee77:
## Description

This is a summary of some inspection of the codebase to see how `multichannel` is typically handled.  It seems the most common default is `multichannel=False`, but this is not always the case.  There are a few functions that default to `None` and then try to guess `True` or `False` based on the data shape (i.e. if the last axis is size 3 (or 4), assume it is multi-channel).  A couple of segmentation routines default to `multichannel=True` instead.

<details>
 <summary>List as of Apr 13, 2017 original comment</summary>

defaults to `None`:
- skimage.filters.gaussian  (nD)
- skimage.restoration.denoise_bilateral  (2D or 2D+channels)
- skimage.restoration.denoise_nl_means  (2D, 2D+channels or 3D)

defaults to `False`:
- skimage.measure.compare_ssim  (nD)
- skimage.restoration.denoise_tv_chambolle  (nD)
- skimage.restoration.denoise_wavelet  (nD)
- skimage.restoration.estimate_sigma  (nD)
- skimage.restoration.inpaint_biharmonic  (nD)
- skimage.segmentation.random_walker_segmentation  (2D, 2D+channels, 3D, 3D+channels)

defaults to `True`:
- skimage.segmentation.felzenszwalb  (only 2D or 2D+channels)
- skimage.segmentation.slic  (2D, 2D+channels, 3D, 3D+channels)

From the above, `skimage.filters.gaussian` is the only n-dimensional function that doesn't default to `multichannel=False`.  Should it be changed from `None` to `False` for consistency?

2 of 3 segmentation routines default to True, while the other defaults to False.
</details>

It would be helpful going forward to have a clear guideline of the expected behaviour for new PRs.  I believe based on feedback in my prior PRs that `multichannel=False` is the preferred default, at least for n-dimensional routines.

**Updated List as of July 2020**
Progress has been made since this issue was originally opened and a number of functions were updated to default to `False`. 

Notably, it seems that `gaussian_filter` and `difference_of_gaussians` are the only nD functions that still try to guess whether the data is multichannel based on its shape. `hog` also infers its multichannel status based on shape, but it is a 2D-only function, so that seems less problematic.

This is the current status (see details above for the original list). The values in parentheses are the number of spatial dimensions supported by the function:

**default=True**
- skimage.segmentation.felzenszwalb (2D)
- skimage.segmentation.slic:  (2D or 3D)
[...]

PR #5201, opened 2021-01-28
FIX: handle 1D array in gaussian filter

sciunto:
## Description

See #5195 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/_gaussian.py
+++ b/skimage/filters/_gaussian.py
@@ -143,7 +143,7 @@ def _guess_spatial_dimensions(image):
     Raises
     ------
     ValueError
-        If the image array has less more than four dimensions.
+        If the image array has less than four dimensions.
     """
     if image.ndim == 1:
         return 1
--- a/skimage/filters/_gaussian.py
+++ b/skimage/filters/_gaussian.py
@@ -143,7 +143,7 @@ def _guess_spatial_dimensions(image):
     Raises
     ------
     ValueError
-        If the image array has less than two or more than four dimensions.
+        If the image array has less more than four dimensions.
     """
     if image.ndim == 1:
         return 1
@@ -156,7 +156,7 @@ def _guess_spatial_dimensions(image):
     if image.ndim == 4 and image.shape[-1] == 3:
         return 3
     else:
-        raise ValueError("Expected 2D, 3D, or 4D array, got %iD." % image.ndim)
+        raise ValueError("Expected 1D, 2D, 3D, or 4D array, got %iD." % image.ndim)
 
 
 def difference_of_gaussians(image, low_sigma, high_sigma=None, *,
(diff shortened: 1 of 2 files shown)


skimage/filters/tests/test_gaussian.py
```diff
@@ -19,6 +19,11 @@ def test_negative_sigma():
                  sigma=np.asarray([-1.0, 1.0]))
 
 
+def test_1D():
+    x = np.arange(50) / 50
+    y = filters.gaussian(x, sigma=(1,))
```
rfezzani, line 24:
```suggestion
    y = gaussian(x, sigma=(1,))
```

rfezzani (commented):
Thank you @sciunto, I think that the beginning of the `gaussian` function and may be the `_guess_spatial_dimensions` error message must be refactored. But I am not sure that an error have to be raised since it is filtered by a `try ... Except ...` block...

sciunto:
> Thank you @sciunto, I think that the beginning of the `gaussian` function and may be the `_guess_spatial_dimensions` error message must be refactored. But I am not sure that an error have to be raised since it is filtered by a `try ... Except ...` block...

I would even wonder whether keeping a guess is the right thing. I thought we want to avoid as much as possible magic. Other opinions @jni @emmanuelle ?

grlee77:
> I would even wonder whether keeping a guess is the right thing. I thought we want to avoid as much as possible magic. Other opinions @jni @emmanuelle ?

We have mostly gotten rid of this type of guessing elsewhere (this function is one exception noted in the list at the top of #2613). It looks like @jni and I discussed it here: https://github.com/scikit-image/scikit-image/issues/2613#issuecomment-656720229 and decided to address it as part of #4294. Working on that issue is one of my near-term priorities for the CZI grant, so we should be able to make progress soon.

skimage/filters/_gaussian.py
```diff
@@ -143,8 +143,10 @@ def _guess_spatial_dimensions(image):
     Raises
     ------
     ValueError
-        If the image array has less than two or more than four dimensions.
+        If the image array has less more than four dimensions.
```
grlee77, line 146:
```suggestion
        If the image array has less than four dimensions.
```

grlee77 (approved):
I vote for merging this as-is and then deal with removing `_guess_spatial_dimensions` separately

skimage/filters/tests/test_gaussian.py
```diff
@@ -19,6 +19,11 @@ def test_negative_sigma():
                  sigma=np.asarray([-1.0, 1.0]))
 
 
+def test_1D():
```
rfezzani, line 22: What is this test supposed to test? Isn't it equivalent to sample from #5195? It doesn't fail with the previous version...

rfezzani (approved):
Thank you @sciunto, waiting for the green button before merging :wink:

jni:
Failure is expected from #5211, so I'll merge now. Thanks everyone!

stefanv:
Thanks for fixing this @sciunto!

The new docstring does not look right: `ValueError` "If the image array has less than four dimensions.".

PR #5204, opened 2021-01-28
single precision support in skimage.registration module

grlee77:
## Description

This PR adds single-precision support for `cross_correlate_masked` and `phase_cross_correlation`. Benchmarks were added to verify faster operation with float32 dtype. If scipy < 1.4, internal FFTs will still be performed in double precision, but the output precision should still match the precision of the input.

One other subtle change: `cross_correlate_masked` previously returned a complex-valued array, but the imaginary component was zero to within floating point precision. Now, the values returned by `ifftn` are cast to real internally and the output of this function will always be real-valued.

For the optical flow methods, `dtype=float32` was already the default. For these optical flow methods I just added dtype checks in the tests.

Benchmark results indicate substantially faster float32 operation for all but the smallest 2d case:

```
[100.00%] ··· benchmark_registration.PhaseCrossCorrelationRegistration.time_phase_cross_correlation                                                                                  ok
[100.00%] ··· ======= ============ ================ ================= ==================
              --                                                   dtype                
              ------------------------------------- ------------------------------------
               ndims   image_size   upscale_factor   numpy.complex64   numpy.complex128 
              ======= ============ ================ ================= ==================
                 2         32             1              92.1±1μs          90.2±2μs     
                 2         32             5              240±1μs           246±10μs     
                 2         32             10             264±4μs           257±2μs      
                 2        100             1              189±1μs           252±6μs      
                 2        100             5              886±50μs        1.03±0.01ms    
                 2        100             10             1.00±0ms        1.16±0.02ms    
                 3         32             1              685±10μs          914±20μs     
                 3         32             5            2.14±0.02ms       2.53±0.03ms    
                 3         32             10           2.21±0.01ms        3.80±0.8ms    
                 3        100             1             22.8±0.4ms        41.2±0.6ms    
                 3        100             5             32.0±0.3ms         54.7±1ms     
                 3        100             10             32.6±1ms          56.8±1ms     
              ======= ============ ================ ================= ==================
```

```
[ 37.50%] ··· benchmark_registration.RegistrationSuite.time_ilk                                                                                                                      ok
[ 37.50%] ··· =============== ============
                   dtype                  
              --------------- ------------
               numpy.float32   1.77±0.01s 
               numpy.float64   1.82±0.02s 
              =============== ============

[ 50.00%] ··· benchmark_registration.RegistrationSuite.time_tvl1                                                                                                                     ok
[ 50.00%] ··· =============== ============
                   dtype                  
              --------------- ------------
               numpy.float32   2.86±0.02s 
               numpy.float64   4.29±0.01s 
              =============== ============
```

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


rfezzani (commented):
That's excellent, thank you @grlee77 :clap:. I just left some suggestions :)

skimage/registration/_masked_phase_cross_correlation.py
```diff
@@ -150,11 +155,18 @@ def cross_correlate_masked(arr1, arr2, m1, m2, mode='full', axes=(-2, -1),
...
 
-    fixed_image = np.array(arr1, dtype=float)
+    fixed_image = np.asarray(arr1)
+    moving_image = np.asarray(arr2)
+    float_dtype = np.result_type(fixed_image.dtype, moving_image.dtype,
+                                 np.float32)
+    if float_dtype.kind == 'c':
+        raise ValueError("complex-valued arr1, arr2 are not supported")
+
+    fixed_image = fixed_image.astype(float_dtype)
```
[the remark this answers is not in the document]
rfezzani: Hum, after a closer look, it seems that we need a copy here, right?

skimage/registration/_masked_phase_cross_correlation.py
```diff
@@ -12,7 +12,12 @@
 import numpy as np
 from functools import partial
 
-from .._shared.fft import fftmodule, next_fast_len
+from .._shared.fft import next_fast_len
+try:
+    from scipy.fft import fftn, ifftn
+except ImportError:
+    # scipy < 1.4 does not have an fft module
+    from scipy.fftpack import fftn, ifftn
```
rfezzani, line 20: Can't this take place in `skimage._shared.fft`?

skimage/registration/_phase_cross_correlation.py
```diff
@@ -4,7 +4,8 @@
 """
 
 import numpy as np
-from .._shared.fft import fftmodule as fft
+# use scipy FFTs to preserve dtype (numpy enforces float64)
+from scipy.fftpack import fftn, ifftn, fftfreq
```
rfezzani, line 8: Shouldn't we use the same `try .. except ...` block as in `_masked_phase_cross_correlation`?

skimage/registration/_masked_phase_cross_correlation.py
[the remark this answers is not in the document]
grlee77: Yes, it is needed to prevent modification of the input. There is a test case that checks for this and will fail if copy=False is used.

skimage/registration/_phase_cross_correlation.py
```diff
@@ -201,22 +204,24 @@ def phase_cross_correlation(reference_image, moving_image, *,
...
 
     # Locate maximum
     maxima = np.unravel_index(np.argmax(np.abs(cross_correlation)),
                               cross_correlation.shape)
     midpoints = np.array([np.fix(axis_size / 2) for axis_size in shape])
 
-    shifts = np.stack(maxima).astype(np.float64)
+    float_dtype = image_product.real.dtype
+
+    shifts = np.stack([m.astype(float_dtype, copy=False) for m in maxima])
```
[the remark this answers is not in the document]
grlee77: Yeah, given that this part should not be a bottleneck for the computation, just using `.astype(float_dtype, copy=False)` should be easier to follow.

skimage/registration/_masked_phase_cross_correlation.py
grlee77: There is already a fallback there when `scipy.fft` doesn't exist is `numpy.fft` rather than `scipy.fftpack`. I think that was done there because the behavior of real-to-complex and complex-to-real transforms in `scipy.fftpack` is inconsistent with `scipy.fft` and `numpy.fft`. In this file only complex-to-complex transforms are needed, though, so we can fall back to `scipy.fftpack`.

I can add a TODO to clean this up once we bump the minimum SciPy to 1.4

skimage/registration/_phase_cross_correlation.py
grlee77: Yes, I was testing to make sure the tests passed with fftpack and forgot go back to the try/except form afterward

skimage/registration/_masked_phase_cross_correlation.py
grlee77: It looks like this fallback doesn't help in any case due to a difference in the `s` argument (CI still fails on the minimum requirements). I may go back to just using the _shared.fft modules and manually fixing the dtype

skimage/registration/_masked_phase_cross_correlation.py
```diff
@@ -150,11 +150,18 @@ def cross_correlate_masked(arr1, arr2, m1, m2, mode='full', axes=(-2, -1),
     if mode not in {'full', 'same'}:
         raise ValueError("Correlation mode '{}' is not valid.".format(mode))
 
-    fixed_image = np.array(arr1, dtype=float)
+    fixed_image = np.asarray(arr1)
+    moving_image = np.asarray(arr2)
+    float_dtype = np.result_type(fixed_image.dtype, moving_image.dtype,
+                                 np.float32)
```
grlee77, line 156: could replace later with 
```Python
float_dtype = _supported_float_type([fixed_image.dtype, moving_image.dtype])
```
using `_supported_float_type` introduced in #5219

grlee77:
I think the comments here were previously addressed

grlee77:
I just rebased on master to check that CI is passing

skimage/registration/tests/test_masked_phase_cross_correlation.py
```diff
@@ -167,7 +169,7 @@ def test_cross_correlate_masked_test_against_mismatched_dimensions():
     m1 = np.ones_like(arr1)
     m2 = np.ones_like(arr2)
 
-    with testing.raises(ValueError):
+    with pytest.rairaises(ValueError):
```
rfezzani, line 172:
```suggestion
    with pytest.raises(ValueError):

```

rfezzani (approved):
Thank you @grlee77 :wink:

mkcor (approved):

PR #5209, opened 2021-01-29
fix clip bug in resize when anti_aliasing is applied (#5202)

### LINKED ISSUE #5202: transform.resize uses anti_aliasing as default and fails to correctly clip values when anti_aliasing is used
## Description

The function transform.resize applies anti_aliasing as default when downsizing images, which is misleading as the default value for kwarg *anti_aliasing* is None.
Moreover, when applying anti_aliasing, when *scipy.ndimage.gaussian_filter* produce values outside the given input range, the algorithm fails to correctly clip the values while clip=True (default).

## Way to reproduce

```python
img = np.ones((208, 208))
transform.resize(img, (128, 128), anti_aliasing=None, clip=True).max()
```
>>> 1.0000000000000004

```python
img = np.ones((208, 208))
sigma = (208 / 128 - 1) * 0.5
ndi.gaussian_filter(img, sigma).max()
```
>>> 1.0000000000000004

## Version information
```python
```

```python
3.8.4 | packaged by conda-forge | (default, Jul 17 2020, 15:16:46) 
[GCC 7.5.0]
Linux-5.4.0-65-generic-x86_64-with-glibc2.10
scikit-image version: 0.17.2
numpy version: 1.19.5
```

abouysso:
## Description

Clipping was performed based on the values of the filtered image and not 
the input image, which can produce values outside of input values range.

Added clarification on kwarg anti_aliasing description: it is applied by 
default for non bool type inputs when downsizing.

Closes #5202

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/transform/_warps.py
```diff
@@ -172,12 +174,12 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
         tform.params[0, 1] = 0
         tform.params[1, 0] = 0
 
-        out = warp(image, tform, output_shape=output_shape, order=order,
+        out = warp(img_in, tform, output_shape=output_shape, order=order,
```
rfezzani, line 177: The proposed solution doesn't work in this case since the clipping is made in `warp` :confused:

skimage/transform/_warps.py
```diff
@@ -115,8 +116,9 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
              "from version 0.19 a ValueError will be raised instead of this "
              "warning.", FutureWarning, stacklevel=2)
 
-    factors = (np.asarray(input_shape, dtype=float) /
-               np.asarray(output_shape, dtype=float))
+    factors = np.divide(input_shape, output_shape)
+    # create copy so input values range stays accessible through image for clip
+    img_in = image.copy()
```
rfezzani, line 121: I am not sure the copy is necessary here.

rfezzani (commented):
Thank you @abouysso, I think that the 2D case is not fixed  with the proposed approach :(
Can you please also add a test?

skimage/transform/_warps.py
abouysso: True @rfezzani, thanks ! Operations later applied on img_in are not *in_place* and thus should not be able to modify image :smiley:.

skimage/transform/_warps.py
abouysso: Thanks for pointing this out ! Fixed it in fallback code by performing clipping outside of warp function :smiley:.

rfezzani: Excellent idea! The `_clip_warp_output` call can be performed outside the `if ... elif ... else ...` block once for all now :wink:

skimage/transform/_warps.py
```diff
@@ -66,8 +66,9 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
     anti_aliasing : bool, optional
         Whether to apply a Gaussian filter to smooth the image prior
         to down-scaling. It is crucial to filter when down-sampling
-        the image to avoid aliasing artifacts. If input image data
-        type is bool, no anti-aliasing is applied.
+        the image to avoid aliasing artifacts. If input data type is not bool,
+        anti_aliasing is applied by default when down-sampling. If input image
+        data type is bool, no anti-aliasing is applied.
```
rfezzani, line 71:
```suggestion
        the image to avoid aliasing artifacts. Default is True if down-scaling and
        image data type is not bool, and False otherwise.
```

This behavior can also be explicit in the code line 109:
```python
if anti_aliasing is None:
    anti_aliasing = (not image.dtype == bool
                     and np.all(out_shape < in_shape))
```

abouysso:
I ran in another issue, also related to potential out of input range values produced by `ndi.gaussian_filter`.
When kwarg `order` is set to 0 (nearest neighbors interpolation), no clipping is performed since such interpolation does not produce out of range data. Thus it seems to me that `_clip_warp_output` behavior needs to be modified in order to be able to handle such cases.

I see two ways of doing so:
- perform clipping as long as `clip` is True, even if `order` is 0.
- add `anti_aliasing` as boolean kwarg, and perform clipping if `clip` is True and either `order != 0` or `anti_aliasing`

skimage/transform/_warps.py
```diff
@@ -115,8 +115,9 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
              "from version 0.19 a ValueError will be raised instead of this "
              "warning.", FutureWarning, stacklevel=2)
 
-    factors = (np.asarray(input_shape, dtype=float) /
-               np.asarray(output_shape, dtype=float))
+    factors = np.divide(input_shape, output_shape)
+    # create copy so input values range stays accessible through image for clip
+    img_in = image
```
rfezzani, line 120: I see a case where this approach doesn't work: when image dtype is not floating point... In the case of `uint8` for example, values will not be clipped between 0 and 1 as it is supposed to be...
We must call `convert_to_float` here to fix this!

abouysso: Indeed there could still be an issue when `preserve_range` is False :smile: Otherwise values will be clipped between 0 and 255 as expected. I will correct this as you suggested !

skimage/transform/_warps.py
```diff
@@ -106,16 +107,17 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
...
         warn("Input image dtype is bool. Gaussian convolution is not defined "
              "with bool data type. Please set anti_aliasing to False or "
              "explicitely cast input image to another data type. Starting "
              "from version 0.19 a ValueError will be raised instead of this "
              "warning.", FutureWarning, stacklevel=2)
 
     factors = np.divide(input_shape, output_shape)
+    image = convert_to_float(image, preserve_range)
     # create copy so input values range stays accessible through image for clip
     img_in = image
```
rfezzani:
```suggestion
    img_in = convert_to_float(image, preserve_range)
```

rfezzani: The following `convert_to_float` calls can now be removed :wink:.

rfezzani: :pray: You already removed them :+1:

abouysso: Yes, however `convert_to_float` is called in warp, so in the 2D case of the fallback code, it will be called twice... probably better to let it like this than touch also the warp function

rfezzani: Arrays pass through `convert_to_float` without modification when their data type is floating point, so there is no major overhead by calling it twice...

abouysso: The proposition you made will not work I think: the reference used for clipping is `image` (as modified in the first commit for this issue). A problem with uint8 data can occur when `preserve_range` is False, so `image` values are in [0, 255] and `img_in` values are in [0, 1].
For having proper reference for clipping, I think `convert_to_float` should be applied to `image`, so references values are in [0,1].

rfezzani: :facepalm: you are right :)

skimage/transform/tests/test_warps.py
```diff
@@ -355,6 +355,18 @@ def test_resize_dtype():
     assert resize(x_f32, (10, 10), preserve_range=True).dtype == x_f32.dtype
 
 
+def test_resize_anti_aliasing_clip():
```
rfezzani, line 358: :+1:
I would suggest to add a test for uint8 data and we will be fine I think :wink: Thank you @abouysso

abouysso: Yes, also added a check with `preserve_range=True` and `preserve_range=False`. :smiley: :

rfezzani: That's great @abouysso, you may save some lines of code by using the decorator `pytest.mark.parametrize` like for example [here](https://github.com/scikit-image/scikit-image/blob/master/skimage/registration/tests/test_ilk.py#L36) :wink:

skimage/transform/_warps.py
```diff
@@ -106,17 +107,19 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
                          "dimensions")
 
     if anti_aliasing is None:
-        anti_aliasing = not image.dtype == bool
+        anti_aliasing = not input_type == bool
```
rfezzani, line 110:
```suggestion
        anti_aliasing = (not input_type == bool and
                         np.any(output_shape < input_shape))
```

rfezzani:
You forgot to import `pytest`, otherwise it looks really good! Thank you @abouysso :ok_hand:

rfezzani (approved):
Thank you @abouysso, it looks good! @scikit-image/core, the PR is ready for review :wink:. CI failures seems unrelated.

skimage/transform/_warps.py
```diff
@@ -633,11 +633,13 @@ def warp_coords(coord_map, shape, dtype=np.float64):
     return coords
 
 
-def _clip_warp_output(input_image, output_image, order, mode, cval, clip):
+def _clip_warp_output(input_image, output_image, order, mode, cval, clip,
```
stefanv, line 636: I don't see why `_clip_warp_output` needs to know anything about anti-aliasing.

stefanv (commented):
Thanks for the patch, @abouysso!

@rfezzani's review comments look good, and I left one additional one.

skimage/transform/_warps.py
rfezzani: This is because `_clip_warp_output` doesn't perform clipping when `order == 0` since values are not suppose to go outside input values range. But this is no more true when `anti_aliasing` is activated :confused:. @abouysso pointed this here https://github.com/scikit-image/scikit-image/pull/5209#issuecomment-769958890

stefanv: I'm missing something: the idea is that we don't *need* to clip because of order 0.  But if out of range data makes it way **in**, that should probably be an error, unless preserve range or similar is set (otherwise, what's the point of clipping).

abouysso: Exact! However out of range data can makes it way in because of `ndi.gaussian_filter` called when anti_aliasing is True. This can be considered as a bug from scipy package. Here the proposition is to use `_clip_warp_output` to handle out of range data caused by interpolation and `ndi.gaussian_filter` !

stefanv: The case we are addressing, though, is surely one that the user did not anticipate.  If they send in out-of-range values, they likely did *not* intend for values to be clipped.

rfezzani: Yes, but this behavior is controlled by the `clip`and `preserve_range` parameters... I mean, in such case, the user have to set `preserve_range` to `True` and `clip` to `False`...

abouysso: The problem could occur like this, as an example:
- user sends in data in range [0,1] for downsizing, and sets `anti_aliasing=True`, `order=0` and `clip=True`, expecting output to be also in range [0,1]
- data is first filtered with `ndi.gaussian_filter`, which can produce out of range data
- the actual resize operation is performed
- `_clip_warp_output` is called and does nothing since `order=0`
- the user get output with out of range data caused by the first step, while `clip=True`... My opinion is that is unexpected behavior.

That's why I propose that `_clip_warp_output` knows about `anti_aliasing`, to clip out of range data potentially caused by anti_aliasing operation ! :smiley:

stefanv: Why not simply always clip inside of `_clip_warp_output`, and call it only when `clip=True`?

rfezzani: That's an other option, but avoiding clipping in the special case `order == 0 or anti_aliasing` saves traversing the image twice searching for its min and max values...

stefanv: Given that the standard code path does it anyway, I'd opt for simplicity of code.

rfezzani: Previous implementation do not clip values when `order == 0`. If I got you well, you prefer to also remove this option, right?

stefanv: I think we should allow clipping whenever overflow is a possibility; so, since for order 0 overflow cannot occur, it is fine to skip the clipping entirely.

skimage/transform/_warps.py
rfezzani: @abouysso, I forgot about the suggestion above, can you please implement it?

skimage/transform/_warps.py
rfezzani: But overflow can occur when order == 0 and anti_aliasing is True like in the situation described by @abouysso https://github.com/scikit-image/scikit-image/pull/5209#discussion_r568468260

rfezzani: That's the root of the reported bug :)

rfezzani: To illustrate this:
```python
import numpy as np
img = np.ones((5, 5))
resized = resize(img, (3, 3), order=0, anti_aliasing=True, clip=True)
resized.max()  # output: 1.0000000000000004
```

stefanv: (if `clip=True`)

abouysso: Alright, thank you for your input !

skimage/transform/_warps.py
abouysso: Done !

skimage/transform/_warps.py
```diff
@@ -648,22 +650,22 @@ def _clip_warp_output(input_image, output_image, order, mode, cval, clip):
 
     Other parameters
     ----------------
-    order : int, optional
+    order : int
```
rfezzani, line 653: We no more need the `order` argument here now :wink:

abouysso: True, I'm correcting this :wink:

skimage/transform/_warps.py
abouysso: Actually this proposition does not seem to work : comparing two tuples always returns a boolean and so np.any() seems to have no effect. My guess is *that* should do what was intended:

```python
anti_aliasing = (not input_type == bool and any(x < y for x,y in zip(output_shape, input_shape)))
```

abouysso:
```
___________________________ test_boll_array_warnings ___________________________

    def test_boll_array_warnings():
        img = np.zeros((10, 10), dtype=bool)
    
        with expected_warnings(['Input image dtype is bool']):
            rescale(img, 0.5, anti_aliasing=True)
    
        with expected_warnings(['Input image dtype is bool']):
            resize(img, (5, 5), anti_aliasing=True)
    
        with expected_warnings(['Input image dtype is bool']):
>           rescale(img, 0.5, order=1)

scikit-image/skimage/transform/tests/test_warps.py:690
```

One of the test failed here, when not returning a warning while it should. I have a hard time understanding whats is the problem, since I can't reproduce the error: this function passes on my version of the code... some help or hints would be appreciated here :smiley:

grlee77:
@abouysso : regarding the failing test case, you now have the following condition for raising the error in your code

`if input_type == bool and anti_aliasing:`

but that failing test case has not set `anti_aliasing=True`, so the warning will not be raised. Assuming that it is indeed okay to use bool inputs when `anti_aliasing=False`, then this warning check should just be removed from that function.

While you are changing that, I would also rename `test_boll_array_warnings` to `test_bool_array_warnings` (I think this was a typo?)

abouysso:
> @abouysso : regarding the failing test case, you now have the following condition for raising the error in your code
> 
> `if input_type == bool and anti_aliasing:`
> 
> but that failing test case has not set `anti_aliasing=True`, so the warning will not be raised. Assuming that it is indeed okay to use bool inputs when `anti_aliasing=False`, then this warning check should just be removed from that function.
> 
> While you are changing that, I would also rename `test_boll_array_warnings` to `test_bool_array_warnings` (I think this was a typo?)

I think here the check is for the call to `_validate_interpolation_order` which should throw a warning when data type is bool and `order != 0`. When calling directly `rescale(img, 0.5, order=1)` with the proposed changes, it throws a warning (at least the first time the function is executed, then the warning is silenced). However with the multiple tests called in `test_boll_array_warnings`, it seems the warning is silenced...

Alos you are right, I can rename `test_boll_array_warnings` into `test_bool_array_warnings`, thats a typo for sure :wink:

skimage/transform/_warps.py
```diff
@@ -66,8 +66,8 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
     anti_aliasing : bool, optional
         Whether to apply a Gaussian filter to smooth the image prior
         to down-scaling. It is crucial to filter when down-sampling
-        the image to avoid aliasing artifacts. If input image data
-        type is bool, no anti-aliasing is applied.
+        the image to avoid aliasing artifacts. Default is True if down-scaling
+        and image data type is not bool, and False otherwise.
     anti_aliasing_sigma : {float, tuple of floats}, optional
         Standard deviation for Gaussian filtering to avoid aliasing artifacts.
```
mkcor, line 72:
```suggestion
        Standard deviation for Gaussian filtering used when anti-aliasing.
```
I know this is not part of the PR, but we might as well better the docstring!

skimage/transform/_warps.py
```diff
@@ -66,8 +66,8 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
     anti_aliasing : bool, optional
         Whether to apply a Gaussian filter to smooth the image prior
         to down-scaling. It is crucial to filter when down-sampling
-        the image to avoid aliasing artifacts. If input image data
-        type is bool, no anti-aliasing is applied.
+        the image to avoid aliasing artifacts. Default is True if down-scaling
```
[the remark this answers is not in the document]
rfezzani: @mkcor, `anti_aliasing` is in fact set to True internally by default when down sizing. It is one of the additions of this PR (please see line 111) :wink:

rfezzani:
@scikit-image/core, can anyone please find some time to review this, thanks :wink:

skimage/transform/_warps.py
```diff
@@ -66,10 +66,10 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
     anti_aliasing : bool, optional
         Whether to apply a Gaussian filter to smooth the image prior
         to down-scaling. It is crucial to filter when down-sampling
```
mkcor, line 68:
```suggestion
        to down-sizing. It is crucial to filter when down-sizing
```
Sorry for requesting this change affecting lines which do not strictly belong to the PR... But I find it way too confusing to juggle between 'down-sizing,' 'down-sampling,' and 'down-sampling.' Let's stick with 'down-sizing,' since the function is called 'resize' and as the PR description says: "Added clarification on kwarg anti_aliasing description: it is applied by
default for non bool type inputs when _downsizing_."

skimage/transform/_warps.py
```diff
@@ -106,17 +107,20 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
                          "dimensions")
 
     if anti_aliasing is None:
-        anti_aliasing = not image.dtype == bool
+        anti_aliasing = (not input_type == bool and
+                         any(x < y for x,y in zip(output_shape, input_shape)))
```
mkcor, line 111:
```suggestion
                         any(x < y for x, y in zip(output_shape, input_shape)))
```

skimage/transform/_warps.py
```diff
@@ -106,17 +107,20 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
...
              "with bool data type. Please set anti_aliasing to False or "
              "explicitely cast input image to another data type. Starting "
              "from version 0.19 a ValueError will be raised instead of this "
              "warning.", FutureWarning, stacklevel=2)
 
-    factors = (np.asarray(input_shape, dtype=float) /
-               np.asarray(output_shape, dtype=float))
+    factors = np.divide(input_shape, output_shape)
+    image = convert_to_float(image, preserve_range)
+    # create copy so input values range stays accessible through image for clip
```
mkcor, line 122:
```suggestion
    # create copy so input value range stays accessible through image for clip
```
No big deal, but it would be more intuitive for `img_in` to be the copy used for retaining the range of values of the input image (both because `img_in` means, I guess, 'input image' and because `img_in` lies at the LHS of the equal sign). So, I would expect `img_in` to keep aside and `image` to undergo all subsequent operations.

skimage/transform/_warps.py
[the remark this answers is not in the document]
mkcor: @rfezzani oh, thanks! Reading the PR description helps as well... :wink: I still think it's not accurate to write "Default is True" because that would mean a user could use a bool input image, not specify anti_aliasing, and fall under `if input_type == bool and anti_aliasing:` (which is not the case). I would therefore suggest the following rewording for this part of the docstring: "It is set to True when down-sizing an image of which data type is not bool."

mkcor:
Hi @abouysso,

Merging #5327 has led to a conflict in file `skimage/transform/_warps.py` for this PR. Are you comfortable resolving it (either locally in your file editor or in GitHub's online editor)?

abouysso:
> Hi @abouysso,
> 
> Merging #5327 has led to a conflict in file `skimage/transform/_warps.py` for this PR. Are you comfortable resolving it (either locally in your file editor or in GitHub's online editor)?

Thanks for pointing this out ! I'm working on it, conflict is solved locally, but new errors are raised on pytest. A quick look seems to indicate those errors comes from test functions modified in #5327 . I still have to find the issue :smile:

skimage/transform/_warps.py
```diff
@@ -106,17 +107,20 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
...
              "with bool data type. Please set anti_aliasing to False or "
-             "explicitely cast input image to another data type. Starting "
+             "explicitly cast input image to another data type. Starting "
              "from version 0.19 a ValueError will be raised instead of this "
              "warning.", FutureWarning, stacklevel=2)
 
-    factors = (np.asarray(input_shape, dtype=float) /
-               np.asarray(output_shape, dtype=float))
+    factors = np.divide(input_shape, output_shape)
+    image = convert_to_float(image, preserve_range)
```
rfezzani, line 121: This conversion to float mustn't be done when `order=0` (please see #5327 :wink:)

abouysso: Thanks for pointing this out. Conflict seems solved !

skimage/transform/_warps.py
```diff
@@ -25,7 +25,7 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
 
     Performs interpolation to up-size or down-size N-dimensional images. Note
     that anti-aliasing should be enabled when down-sizing images to avoid
-    aliasing artifacts. For down-sampling with an integer factor also see
+    aliasing artifacts. For down-sizing with an integer factor also see
```
jni, line 28:
```suggestion
    aliasing artifacts. For downsampling with an integer factor also see
```

"downsizing" has a specific meaning in English that does not match what we are doing here. :blush:

skimage/transform/_warps.py
```diff
@@ -65,13 +65,13 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
         Also see https://scikit-image.org/docs/dev/user_guide/data_types.html
     anti_aliasing : bool, optional
         Whether to apply a Gaussian filter to smooth the image prior
-        to down-scaling. It is crucial to filter when down-sampling
-        the image to avoid aliasing artifacts. If input image data
-        type is bool, no anti-aliasing is applied.
+        to down-sizing. It is crucial to filter when down-sizing
```
jni, line 68:
```suggestion
        to downsampling. It is crucial to filter when downsampling
```

skimage/transform/_warps.py
```diff
@@ -65,13 +65,13 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
...
+        the image to avoid aliasing artifacts. If not specified, it is set to
+        True when down-sizing an image of which data type is not bool.
```
jni, line 70:
```suggestion
        True when downsampling an image whose data type is not bool.
```

skimage/transform/_warps.py
```diff
@@ -65,13 +65,13 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
...
     anti_aliasing_sigma : {float, tuple of floats}, optional
-        Standard deviation for Gaussian filtering to avoid aliasing artifacts.
+        Standard deviation for Gaussian filtering used when anti-aliasing.
         By default, this value is chosen as (s - 1) / 2 where s is the
-        down-scaling factor, where s > 1. For the up-size case, s < 1, no
+        down-sizing factor, where s > 1. For the up-size case, s < 1, no
```
jni, line 74:
```suggestion
        downsampling factor, where s > 1. For the up-size case, s < 1, no
```

skimage/transform/_warps.py
```diff
@@ -133,17 +139,13 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
                 warn("Anti-aliasing standard deviation greater than zero but "
                      "not down-sampling along all axes")
         image = ndi.gaussian_filter(image, anti_aliasing_sigma,
-                                    cval=cval, mode=ndi_mode)
+                                     cval=cval, mode=ndi_mode)
```
jni, line 142:
```suggestion
                                    cval=cval, mode=ndi_mode)
```

skimage/transform/_warps.py
jni: Also, personally I dislike reassigning values in the middle of a function, so, rather than copy image to image_in then overwriting image, how about just renaming the downstream images? ie below, use `image_smoothed = ndi.gaussian_filter(image, ...)`. What do you think @abouysso?

jni (commented):
@abouysso apologies that it has taken so long to to review this... Our team members have been rather overwhelmed this year. I've left a couple of comments but this should be ready very quickly! Do you have the time/inclination to look at my comments, or would you rather we take over from here?

Thank you!

skimage/transform/_warps.py
mkcor: One line above it also reads 'down-sizing' and two lines above, 'up-size or down-size' :grimacing:

skimage/transform/_warps.py
rfezzani: That's in fact an option, but you need to define `image_smoothed` even when `anti_aliasing is False` in this case

rfezzani:
@scikit-image/core, I added this PR to 0.19 milestone because it is close and think it is fixes `skimage.transform.resize` which is a sensible function.
@abouysso, do you have bandwidth to work on this soon, otherwise I can take over if you wish :wink:

skimage/transform/_warps.py
rfezzani: I instead created `img_bounds` to save input image bounds for later clipping... Though?

skimage/transform/_warps.py
```diff
@@ -220,7 +220,6 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
                    preserve_range=preserve_range)
 
     else:  # n-dimensional interpolation
-        order = _validate_interpolation_order(input_type, order)
```
rfezzani: Good catch @grlee77

rfezzani:
:tada: Thank you @grlee77

PR #5219, opened 2021-02-04
single precision support in skimage.restoration module

grlee77:
## Description

This PR adds float32 support to `wiener`, `unsupervised_wiener`, and `rolling_ball`. Some tests for the expected dtype were added to other functions that already had `float32` support.

Here I introduced a `_float_dtype` helper in `_shared.utils` to try to keep casting to floating point types consistent. Probably #5204 could use this as well.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/_shared/utils.py
```diff
@@ -450,3 +450,31 @@ def _fix_ndimage_mode(mode):
...
+        Floating-point dtype for the image.
+    """
+    char = image.dtype.char
+    if char == 'e':
+        return np.float32
+    elif char == 'G':
+        return np.complex128
+    elif char not in 'dfDF':
+        return np.float64
+    return image.dtype
```
FirefoxMetzger, line 480: Also, other types should be cast to `np.float64`; either the docstring needs an update or the last line does to make it consistent.

skimage/restoration/_denoise.py
```diff
@@ -3,7 +3,7 @@
 from math import ceil
 from .. import img_as_float
 from ._denoise_cy import _denoise_bilateral, _denoise_tv_bregman
-from .._shared.utils import warn
+from .._shared.utils import _float_type, warn
```
FirefoxMetzger, line 6: `_float_type` appears to be unused.

skimage/restoration/tests/test_denoise.py
```diff
@@ -35,22 +36,24 @@
 astro_odd = astro[:, :-1]
 
 
-def test_denoise_tv_chambolle_2d():
+@testing.parametrize('dtype', [np.float32, np.float64])
```
FirefoxMetzger, line 39: Actual question; Why are we limiting ourselves to the two float types here? Wouldn't it be a more thorough test to check for all the types that `_float_type` cares about?

skimage/restoration/tests/test_inpaint.py
```diff
@@ -6,14 +6,16 @@
...
+@testing.parametrize('dtype', [np.float32, np.float64])
+def test_inpaint_biharmonic_2d(dtype):
+    img = np.tile(np.square(np.linspace(0, 1, 5, dtype=dtype)), (5, 1))
     mask = np.zeros_like(img)
     mask[2, 2:] = 1
     mask[1, 3:] = 1
     mask[0, 4:] = 1
     img[np.where(mask)] = 0
     out = inpaint.inpaint_biharmonic(img, mask)
+    assert out.dtype == img.dtype
```
FirefoxMetzger, line 18: This will work for the current tests but would fail if we add say `np.float16` to line 9. That is if I understand the current idea correctly to return the image in the same type it was processed in.

skimage/restoration/tests/test_inpaint.py
```diff
@@ -6,14 +6,16 @@
 from skimage._shared.testing import assert_allclose
 
 
-def test_inpaint_biharmonic_2d():
-    img = np.tile(np.square(np.linspace(0, 1, 5)), (5, 1))
+@testing.parametrize('dtype', [np.float32, np.float64])
```
FirefoxMetzger, line 9: same thought here, why not test all relevant dtypes (all floats, or all types applicable to `inpaint`)

skimage/restoration/tests/test_j_invariant.py
```diff
@@ -27,10 +28,12 @@ def test_invariant_denoise():
     assert_(denoised_mse < original_mse)
 
 
-def test_invariant_denoise_color():
+@testing.parametrize('dtype', [np.float32, np.float64])
```
FirefoxMetzger, line 31: same thought as previously: why not test all applicable types?

skimage/restoration/tests/test_j_invariant.py
```diff
@@ -27,10 +28,12 @@ def test_invariant_denoise():
...
+def test_invariant_denoise_color(dtype):
     denoised_img_color = _invariant_denoise(
-            noisy_img_color, _denoise_wavelet,
-            denoiser_kwargs=dict(multichannel=True))
+        noisy_img_color.astype(dtype), _denoise_wavelet,
+        denoiser_kwargs=dict(multichannel=True))
+    assert denoised_img_color.dtype == dtype
```
FirefoxMetzger, line 36: similar thought as above: won't this fail if `dtype` and `_float_type(dtype)` don't match?

skimage/restoration/tests/test_restoration.py
```diff
@@ -11,46 +12,60 @@
 test_img = util.img_as_float(camera())
 
 
-def test_wiener():
-    psf = np.ones((5, 5)) / 25
+@testing.parametrize('dtype', [np.float32, np.float64])
```
FirefoxMetzger, line 15: same as above: why not test all relevant dtypes?

skimage/restoration/tests/test_restoration.py
```diff
@@ -11,46 +12,60 @@
...
+def test_wiener(dtype):
+    psf = np.ones((5, 5), dtype=dtype) / 25
     data = convolve2d(test_img, psf, 'same')
     np.random.seed(0)
     data += 0.1 * data.std() * np.random.standard_normal(data.shape)
+    data = data.astype(dtype, copy=False)
     deconvolved = restoration.wiener(data, psf, 0.05)
+    assert deconvolved.dtype == dtype
```
FirefoxMetzger, line 23: similar thought as above: won't this fail if `dtype` and `_float_type(dtype)` don't match?

skimage/restoration/tests/test_restoration.py
```diff
@@ -11,46 +12,60 @@
...
 
     path = fetch('restoration/tests/camera_wiener.npy')
-    np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3)
+    atol = 1e-5 if dtype == np.float32 else 0
+    np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3,
+                               atol=atol)
 
     _, laplacian = uft.laplacian(2, data.shape)
     otf = uft.ir2tf(psf, data.shape, is_real=False)
+    assert otf.real.dtype == dtype
```
FirefoxMetzger, line 32: similar thought as above: won't this fail if `dtype` and `_float_type(dtype)` don't match?

skimage/restoration/tests/test_restoration.py
```diff
@@ -11,46 +12,60 @@
...
     deconvolved = restoration.wiener(data, otf, 0.05,
                                      reg=laplacian,
                                      is_real=False)
+    assert deconvolved.real.dtype == dtype
```
FirefoxMetzger, line 36: similar thought as above: won't this fail if `dtype` and `_float_type(dtype)` don't match?

skimage/restoration/tests/test_restoration.py
```diff
@@ -11,46 +12,60 @@
...
     data = convolve2d(test_img, psf, 'same')
     np.random.seed(0)
     data += 0.1 * data.std() * np.random.standard_normal(data.shape)
+    data = data.astype(dtype, copy=False)
     deconvolved = restoration.wiener(data, psf, 0.05)
+    assert deconvolved.dtype == dtype
 
     path = fetch('restoration/tests/camera_wiener.npy')
-    np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3)
+    atol = 1e-5 if dtype == np.float32 else 0
```
FirefoxMetzger, line 26: how come, we can afford 0 tolerance for all non-FP32 types? FP16 will definitely need a larger one. For FP64 it might work, but a hard 0 seems like a strong claim.

skimage/restoration/tests/test_restoration.py
```diff
@@ -11,46 +12,60 @@
...
-    psf = np.ones((5, 5)) / 25
+@testing.parametrize('dtype', [np.float32, np.float64])
+def test_unsupervised_wiener(dtype):
+    psf = np.ones((5, 5), dtype=dtype) / 25
     data = convolve2d(test_img, psf, 'same')
     np.random.seed(0)
     data += 0.1 * data.std() * np.random.standard_normal(data.shape)
+    data = data.astype(dtype, copy=False)
     deconvolved, _ = restoration.unsupervised_wiener(data, psf)
+    assert deconvolved.dtype == dtype
```
FirefoxMetzger, line 50: similar thought as above: won't this fail if `dtype` and `_float_type(dtype)` don't match?

skimage/restoration/tests/test_restoration.py
```diff
@@ -11,46 +12,60 @@
...
+    assert deconvolved.real.dtype == dtype
     np.testing.assert_allclose(np.real(deconvolved),
                                np.load(path),
-                               rtol=1e-3)
+                               rtol=1e-3, atol=atol)
 
 
-def test_unsupervised_wiener():
-    psf = np.ones((5, 5)) / 25
+@testing.parametrize('dtype', [np.float32, np.float64])
```
FirefoxMetzger, line 42: similar to above: is it worth testing all relevant dtypes in addition to FP32 and FP64?

skimage/restoration/tests/test_restoration.py
```diff
@@ -11,46 +12,60 @@
...
     data = convolve2d(test_img, psf, 'same')
     np.random.seed(0)
     data += 0.1 * data.std() * np.random.standard_normal(data.shape)
+    data = data.astype(dtype, copy=False)
     deconvolved, _ = restoration.unsupervised_wiener(data, psf)
+    assert deconvolved.dtype == dtype
 
     path = fetch('restoration/tests/camera_unsup.npy')
-    np.testing.assert_allclose(deconvolved, np.load(path), rtol=1e-3)
+    atol = 1e-5 if dtype == np.float32 else 0
```
FirefoxMetzger, line 53: same here: Can we really claim 0 tolerance for non-FP32 types?

skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -18,6 +19,15 @@ def test_ellipsoid_const():
     assert np.allclose(img - background, np.zeros_like(img))
 
 
+@testing.parametrize('dtype', [np.float32, np.float64])
```
FirefoxMetzger, line 22: same here: is it worth testing all relevant dtypes?

skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -18,6 +19,15 @@ def test_ellipsoid_const():
...
+def test_rolling_ball_dtype(dtype):
+    img = 155 * np.ones((100, 100), dtype=dtype)
+    kernel = ellipsoid_kernel((25, 53), 50).astype(dtype, copy=False)
+    background = rolling_ball(img, kernel=kernel)
+    assert background.dtype == img.dtype
```
FirefoxMetzger, line 27: This will always be true, because we cast `background` to `img.dtype` after applying the kernel. The question is if we want that or if it should be `background.dtype == np.float32` if img is F16.

skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -18,6 +19,15 @@ def test_ellipsoid_const():
...
+    assert np.allclose(img - background, np.zeros_like(img))
```
FirefoxMetzger, line 28: while it's not wrong, it's a bit unexpected to test for numerical correctness in a function that claims to test datatypes. If it fails because of this line, one might be lead down the wrong path and search for a ´dtype´ issue.

skimage/restoration/uft.py
```diff
@@ -390,7 +390,8 @@ def ir2tf(imp_resp, shape, dim=None, is_real=True):
     if not dim:
         dim = imp_resp.ndim
     # Zero padding and fill
-    irpadded = np.zeros(shape)
+    irpadded_dtype = imp_resp.dtype if imp_resp.dtype.kind == 'f' else float
```
FirefoxMetzger, line 393: could you explain why `_float_types` is not applicable here?

skimage/restoration/uft.py
```diff
@@ -400,9 +401,13 @@ def ir2tf(imp_resp, shape, dim=None, is_real=True):
                                shift=-int(np.floor(axis_size / 2)),
                                axis=axis)
     if is_real:
-        return fft.rfftn(irpadded, axes=range(-dim, 0))
+        out = fft.rfftn(irpadded, axes=range(-dim, 0))
     else:
-        return fft.fftn(irpadded, axes=range(-dim, 0))
+        out = fft.fftn(irpadded, axes=range(-dim, 0))
```
FirefoxMetzger, line 406:
```suggestion
func = fft.rfftn if is_real else fft.fftn
out = func(irpadded, axes=range(-dim, 0))
```

FirefoxMetzger: not sure which one is cleaner

FirefoxMetzger (commented):
Nice work 👍 

I've added a bunch of comments, but most of them are a repetition of what I commented earlier, just to put it in every place that I found that may need a change.

In general, I wonder if functions should return the same `dtype` as the input image, or if they should return the `dtype` in which the processing was done. E.g., if the input is FP16, should it return FP32 or FP16?

skimage/_shared/utils.py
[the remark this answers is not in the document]
jni: @FirefoxMetzger I don't think your code is equivalent? There's a `return image.dtype` in Greg's code that's not accounted for in your suggestion, afaict.

However, I agree that a dict much more readable. I prefer functions to have just one or two return statements, rather than a cascade of them.

skimage/_shared/utils.py
```diff
@@ -450,3 +450,31 @@ def _fix_ndimage_mode(mode):
     if NumpyVersion(scipy.__version__) >= '1.6.0':
         mode = grid_modes.get(mode, mode)
     return mode
+
+
+def _float_type(image):
```
jni, line 455: From the name alone, it's very hard to know what this function will do. I suggest it should only receive the dtype, not the image, and it should be renamed to e.g. `_supported_float_type(input_dtype)`

skimage/_shared/utils.py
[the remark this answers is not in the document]
FirefoxMetzger: Yes. if I understand the intention of the function correctly, then `image.dtype` is actually a bug. The description says that "everything else will be converted to FP64", but it actually isn't.

Either the docstring should get an update, or the behavior should change. See the comment below.

skimage/_shared/utils.py
FirefoxMetzger: I was thinking the same. At the same time, one might be tempted to do `_float_type(np.float32)`, which will not work because `np.float32` is a class, not an instance. Not sure if that argument should be given weight though.

skimage/_shared/utils.py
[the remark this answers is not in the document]
grlee77: I agree readability/style can be improved and am fine with a dictionary based approach.

The last `return image.dtype` can only be reached for characters 'dfDF' which are float32, float64, complex64 and complex128. These are the types are currently was not modifying, so I don't think it is a bug. Clearly the fact that we are having this discussion means clarity can be improved!

I originally had a much simpler function that was just `return np.promote_types(image.dtype, np.float32)` but that one converts smaller size ints to float32, but larger ints to float64 and also did not deal with downcasting float128/complex256.

grlee77: see "char not in" two lines above. Only the floating point types we want to keep ('dfDF') can reach that last line. We don't want to cast complex64 to float64 for instance.

We could potentially add an allow_complex flag to this utility that would just raise a ValueError on complex dtypes since the majority of functions in scikit-image deal with real-valued inputs.

skimage/restoration/tests/test_denoise.py
grlee77: Sure, I think we can and float16 and float128 here to make sure they are also handled as expected.

skimage/restoration/tests/test_inpaint.py
grlee77: yes, in that case we have to change to `assert out.dtype == _float_type(img.dtype)`

skimage/restoration/tests/test_rolling_ball.py
grlee77: right, thanks!

skimage/restoration/uft.py
grlee77: I think it is now and will update it. I had a different `_float_types` implementation initially that didn't match this behavior for integer types.

This is still subtly different in that for the code here float16 or float128 would be passed on unchanged, but that isn't really desirable or necessary since the FFT functions later on will end up casting to single or double precision anyways.

skimage/restoration/uft.py
grlee77: I like the suggestion here, thanks

skimage/_shared/utils.py
grlee77: Definitely like the suggested name. Passing the dtype instead is probably also clearer.

Regarding @FirefoxMetzger's concern we can just call `dtype = np.dtype(dtype)` within the function to make sure it is an instance.

skimage/_shared/utils.py
FirefoxMetzger: ah right. I missed that! my bad. As I understand you are happy with the dict; otherwise I would suggest switching the logic of the last `elif` and the final return statement to make the final return the catch-all/default statement. Just to create a more intuitive `if ... elif ... else` style.

Not sure about the flag idea. Do you think adding complexity here will make our life easier somewhere else?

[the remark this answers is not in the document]
FirefoxMetzger: dict/elif is fine with me. If we go for an `elif` version, I would suggest switching the logic of the last elif to `char in 'dDfF'` and introducing a final `else` as in 

```
if char == 'e':
    new_dtype = np.float32
elif ...
    new_dtype = ...
else:
    new_dtype = np.float64

return new_dtype
```

since that gives the function a single return statement and makes the flow more intuitive (to me anyway).

grlee77:
Here is a proposed `_support_float_type` implementation based on the reviewer suggestions. I added two features that I would like feedback on. I moved 

1.) optionally raise ValueError if the input is complex-valued
2.) handle a sequence of dtypes for use in functions with more than one relevant input (e.g. `skimage.metrics` functions that deal with comparing two images).

I added float64 to the preserved types list for clarity, even though it is not strictly necessary since it is the default when a key is not found.

```Python

new_float_type = {
    # preserved types
    np.float32().dtype.char : np.float32,
    np.float64().dtype.char : np.float64,
    np.complex64().dtype.char : np.complex64,
    np.complex128().dtype.char : np.complex128,
    # altered types
    np.float16().dtype.char : np.float32,
    'g' : np.float64      # np.float128 ; doesn't exist on windows
    'G' : np.complex128   # np.complex256 ; doesn't exist on windows
}


def _supported_float_type(dtype, allow_complex=False):
    """Return an appropriate floating-point dtype for a given dtype.

    float32, float64, complex64, complex128 are preserved.
    float16 is promoted to float32.
    complex256 is demoted to complex128.
    Other types are cast to float64.

    Paramters
    ---------
    dtype : np.dtype or Iterable of np.dtype
        The input dtype. If a sequence of multiple dtypes is provided, each
        dtype is first converted to a supported floating point type and the
        final dtype is then determined by applying `np.result_type` on the
        sequence of supported floating point types.
    allow_complex : bool, optional
        If False, raise a ValueError on complex-valued inputs.

    Retruns
    -------
    float_type : dtype
        Floating-point dtype for the image.
    """
    if isinstance(dtype, Iterable) and not isinstance(dtype, str):
        return np.result_type(*(_supported_float_type(d) for d in dtype))
    dtype = np.dtype(dtype)
    if not allow_complex and dtype.kind == 'c':
        raise ValueError("complex valued input is not supported")
    return new_float_type.get(dtype.char, np.float64)
```

skimage/restoration/tests/test_rolling_ball.py
grlee77: I have left the dtype behavior in rolling_ball as-is for now. I think this was mainly as a convenience for users providing integer valued inputs to get integer valued outputs?

skimage/restoration/tests/test_rolling_ball.py
grlee77: done

skimage/restoration/tests/test_rolling_ball.py
FirefoxMetzger: Yes. We introduced this because of integer images - I never wrote an optimized integer kernel -, but it affects all inputs equally. The general idea was "dtype in, dtype out".

I think it will be best if the behavior here regarding the return type is consistent with the other functions in the library. If that means changing the current behavior, I think it makes sense to do.

jni: I think this concern is orthogonal to the purpose if this PR, so let's just defer this decision so as not to hold the PR up unnecessarily.

jni (approved):
Thanks @grlee77! I've only made a minor and super optional code style comment.

@FirefoxMetzger identified a test that is possibly-not-comprehensive-enough [here](https://github.com/scikit-image/scikit-image/pull/5219/files#r573459114), and I also wonder whether you've overlooked that?

jni:
(But note that I'd be happy to merge this as-is in the interest of keeping momentum going with this series of PRs.)

jni:
@grlee77 do you want to pick this one up again?

skimage/restoration/deconvolution.py
```diff
@@ -263,19 +279,21 @@ def unsupervised_wiener(image, psf, reg=None, user_params=None, is_real=True,
...
+        data_spectrum = uft.ufft2(image)
 
     # Gibbs sampling
     for iteration in range(params['max_iter']):
         # Sample of Eq. 27 p(circX^k | gn^k-1, gx^k-1, y).
 
         # weighting (correlation in direct space)
         precision = gn_chain[-1] * atf2 + gx_chain[-1] * areg2  # Eq. 29
+        _rand1 = np.random.standard_normal(data_spectrum.shape)
+        _rand2 = np.random.standard_normal(data_spectrum.shape)
```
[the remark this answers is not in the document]
grlee77: I think it was just for line length. I will update it

skimage/restoration/tests/test_j_invariant.py
grlee77: Done. There was actually a missing promotion to float32 for float16 inputs!

grlee77:
> @grlee77 do you want to pick this one up again?

Sure, I have addressed the initial comments and rebased. I updated one of the non-local means test cases, which happened to give almost the same PSNR for the two cases being compared so that it has a more suitable test object than a simple square.

skimage/restoration/tests/test_restoration.py
```diff
@@ -2,7 +2,9 @@
 import pytest
 from scipy.signal import convolve2d
 from scipy import ndimage as ndi
+from skimage._shared import testing
```
rfezzani, line 5: It's a detail, but I think that we must use `@pytest.mark.parametrize` instead of its alias `@testing.parametrize`

skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -7,15 +7,19 @@
 import pytest
 
 from skimage import data, io
+from skimage._shared import testing
```
rfezzani, line 10: Same as previously here

skimage/restoration/uft.py
```diff
@@ -22,6 +22,7 @@
 
 import numpy as np
 from .._shared.fft import fftmodule as fft
+from .._shared.utils import _supported_float_type
 
 __keywords__ = "fft, Fourier Transform, orthonormal, unitary"
```
rfezzani, line 27: Same as previously :)
```suggestion
```

rfezzani (commented):
Thank you @grlee77, that's amazing! I just left really minor comments, otherwise it is ready to merge :wink:

skimage/restoration/tests/test_restoration.py
grlee77: yes, good catch

rfezzani:
Thank you @grlee77 :tada:

grlee77:
Thanks @rfezzani, I have implemented your suggestions and CI is green.

grlee77:
Nevermind, I see you merged just before I added the comment! Thanks for the detailed review.

Issue #3812, opened 2019-03-22
Regionprops for 3D surfaces and perimeters 


emmanuelle:
At the moment, in regionprops perimeter is for 2D images only, but with the marching cubes functions we have everything to compute the surface areas of connected components.

Also, for perimeters in 2D we could compute them in another way than what is done at the moment, using the contour returned by find_contours. It would give a more accurate estimation of the perimeter I think.

What do you think? It's something I could work on if there is enough interest.

jni:
Hello! See #3797 also. =)

I'm unclear what the right move is here for nomenclature: option 1:

| name      | 2D property | 3D property |
|-----------|-------------|-------------|
| area      | area        | volume      |
| perimeter | perimeter   | area        |

option 2:

| name      | 2D property    | 3D property    |
|-----------|----------------|----------------|
| perimeter | perimeter      | NotImplemented |
| area      | area           | surface area   |
| volume    | NotImplemented | volume         |

The first option matches what is happening now with area which is good. It also generalises more easily to higher dimensions. (area -> nD, perimeter -> (n-1)D). The second option, though, would probably be more intuitive to a newcomer. What do you think?

Either way, I think it's totally a good idea to add this!

emmanuelle:
Oh yes it's funny that there was this other issue not such a long time ago! 

For the nomenclature, if we want something which generalizes well to higher dimension, we should use volume and surface area even in the 2D case (see https://en.wikipedia.org/wiki/N-sphere) so I don't really like option 1. I think I would prefer volume/surface_area but I also understand why people would prefer option 2.

yxdragon:
@emmanuelle, Yes, we can count the 2d area and perimeter by contours vector. we can use JTS, A light geometry libraries! and there is a python implementation, shapely. And it is not difficult to write it only by numpy. As I wrote for ImagePy's plugin.

assume cont is the contours vector[(x1, y1), (x2, y2), ...(xn, yn)], the code is below:

```python
def area(cont):
    return np.cross(cont[1:], cont[:-1]).sum()/2

def length(cont):
    return norm(cont[1:]-cont[:-1], axis=1).sum()

def centroid(cont):
    ws = np.cross(cont[1:], cont[:-1])
    cs = (cont[1:] + cont[:-1])/3
    return (cs.T*ws).sum(axis=1)/ws.sum()
```

another useful feature, feret ratio
```python
def feret_ratio(contours):
    tri = np.array(contours)
[...]

PR #5226, opened 2021-02-08
Rectangle class for object-detection with bounding-boxes

LauLauThom:
## Description
This is a first PR of a (hopefully) a series of PR to extend `match_template` to support simultaneous detections with multiple template images (see multi-template matching [article](https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-020-3363-7)).
This first PR introduces the Rectangle class representing individual detections and providing utilities to compute the intersection between 2 rectangles and the ratio intersection/union.
This was also briefly discussed in Zulip  [thread](https://skimage.zulipchat.com/#narrow/stream/175598-general/topic/Multi-template-matching.20contribution).

I put this Rectangle class in a *metrics>overlap* initially, although I think now it could maybe be put in a module "shape" or so. 
I put some tests at the bottom of the module, I guess they should be moved to a dedicated unit test once the final location is settled.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


jni:
Thanks for this contribution, @LauLauThom! And hello! 👋 😊 

Normally, we go for a functional approach and pure NumPy arrays for data, but I've struggled with rectangle representations enough (is it the endpoint or is it the width?) that I'm happy to go for a class here. We'll have to see what other @scikit-image/core devs think though.

Having said that, I don't see a reason for using staticmethods instead of just functions, what do you think?

Additionally, I think here that we could make this a "hyperrectangle" and make the same code work for both 2D and 3D object detections, which I think would be really useful for biologists, among others. I can take a pass at it once the API has settled a bit if you don't want to tackle it.

Finally, tests should go in a dedicated test file, rather than in a main guard at the end. You can see the `tests/` subdirectory within metrics for examples.

Let us know if you want more specific guidance here, and thank you again!

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,221 @@
...
+class Rectangle():
+    """
+    Represents rectangles which position corresponds to the (r,c) coordinates for the top left corner.
+    Rectangle objects have a number of attributes:
+        - coordinates for the bottom right corner
+        - rectangle dimensions
+        - area
+    as well as utility functions :
+        - computation of the intersection of 2 rectangles
+        - equivalent area for te union of 2 rectangles.
```
jni, line 16: This should use NumPy style formatting, see https://numpy.org/doc/stable/docs/howto_document.html

and a good example:

https://github.com/scikit-learn/scikit-learn/blob/7e197fd600b72483ff2343b8d2605368d1ab24bf/sklearn/svm/_classes.py#L13

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,221 @@
...
+    """
+
+    def __init__(self, top_left, bottom_right=None, size=None):
+        """
+        Construct a rectangle using the (r,c) coordinates for the top left corner,
+        and either the coordinates of the botton right corner
+        or the rectangle size (height, width).
```
jni, line 23: Oh I see that you used NumPy here. 🤦 Sorry about that. Actually a weird funkiness of the whole thing is that the *init* docstring goes on the class, and the init typically doesn't have one. That's because `__init__` is a private method and typically does not get rendered in the documentation. It's a bit inconsistent but that's how it is. 🤷

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,221 @@
...
+    Rectangle objects have a number of attributes:
+        - coordinates for the bottom right corner
+        - rectangle dimensions
+        - area
+    as well as utility functions :
+        - computation of the intersection of 2 rectangles
+        - equivalent area for te union of 2 rectangles.
+    """
+
+    def __init__(self, top_left, bottom_right=None, size=None):
```
jni, line 19:
```suggestion
    def __init__(self, top_left, *, bottom_right=None, size=None):
```

In order to avoid confusion, I would force the arguments after `top_left` to be keyword-only. That way users will always have to document `bottom_right` or `size`.

skimage/metrics/overlap.py
LauLauThom: I was actually wondering about that 😅, in the Ipython console both are shown but it makes sense indeed to have everything in the class dosctring. I updated accordingly. 

However I also replaced the `area` and `dimensions` attributes with getter methods and I hid the `top_left` , `bottom_right`, `width` and  `height` attributes by preceding their name with an underscore. Just to prevent users to update these values doing myRectangle.topLeft = (r,c), which would invalidate the rest of the attributes. 

I though of making them properties with getter and setter but changing the position of one corner does not tell us if the users wants to also move the other corner (ie move the rectangle) or actually resize the rectangle (ie keeping the other corner in place).

skimage/metrics/overlap.py
LauLauThom: see below 😉

LauLauThom:
> Having said that, I don't see a reason for using staticmethods instead of just functions, what do you think?

I am doing a bit of java in parallel so I tend to follow the same pattern 😁. This decorator is just an indication for other developers interested in the source code no ? 
or you mean replacing `intersection_rectangles(rectangle1, rectangle2)` with something like `rectangle1.intersection(rectangle2)`

jni:
> I am doing a bit of java in parallel

I thought that might be the case 😂 

> or you mean replacing intersection_rectangles(rectangle1, rectangle2) with something like rectangle1.intersection(rectangle2)

No, I mean replacing `Rectangle.intersection_rectangles(r1, r2)` with just `intersect_rectangles(r1, r2)`.

> I have duplicated the tests to a dedicated test script in the test subdirectory, but I don't know how to actually run these 😅, so I kept my former test in the main class until now.

install pytest (or rather `pip install -r requirements/test.txt`), then run `pytest path/to/tests/test_file.py`. You should also have an [editable install](https://packaging.python.org/guides/distributing-packages-using-setuptools/#working-in-development-mode) of your working scikit-image directory.

LauLauThom:
> No, I mean replacing `Rectangle.intersection_rectangles(r1, r2)` with just `intersect_rectangles(r1, r2)`

I see, the reason I put them in the `Rectangle` class and not just in the module overlap.py was that I was thinking that the module could potentially by extended with other shapes in the future, like circles, with similar functions. So for instance there is a `Rectangle.is_intersecting(rect1, rect2)` but we could think of a `Circle.is_intersecting(c1, c2)` at some point.

I was actually thinking to rename the module from overlap.py to something more intuitive like shapes.py, or maybe to move the code to a similar module, I think I have seen a class for polygons isn't it ? 
Most methods currently cover intersection and IoU but it could be more general-purpose with methods such as `rectangle.translate` or `.rescale`...

I could run the tests, thanks for the tip, so I finally removed the if main close ;)

LauLauThom:
Hi there, 
I was just wondering if I should do something, I know the built was successful once with the current state. As far as I understood the failed tests are about the packaging, which seems to impact other PRs too or ?

Anyway, except the point below, I don't plan any further changes ;)
> I was actually thinking to rename the module from overlap.py to something more intuitive like shapes.py, or maybe to move the code to a similar module, I think I have seen a class for polygons isn't it ?

jni:
@LauLauThom

> I was just wondering if I should do something,

yes, we will want to go to a fully functional API rather than with the staticmethods, sorry I wasn't clear. :joy: I think I read your earlier comment too quickly and I understood it as "this is the rationale for why I did it this way, [... but I will still change it to the format desired for the library]" :sweat_smile: Don't we love to read what we want into things? =)

For the circle/rectangle discussion, (a) we will cross that bridge when we get to it, but (b) I would simply suggest `def rectangles_intersect(r1: Rectangle, r2: Rectangle) -> bool:`.

LauLauThom:
Hi there,
I removed the static decorators and moved the functions out of the Rectangle class 😊
Sorry for the shitty commit history though ! 😅

LauLauThom:
Hi @jni I realised that the library shapely would do most of the work of this PR 😅

The `shapely.geometry.polygon.Polygon` object can represent the rectangular detection and readily provides the intersection, union and "is_overlapping" methods which I need for the Non-Maxima suppression. And I can derive a child object from this Polygon with additional attributes to store the detection score... 

However I don't see any use of shapely within skimage so far and I can imagine you don't want to add an additional dependency. 
I am not sure what's the best way to go on then...
I was thinking to create a branch "100%-skimage" Multi-Template Matching version that is more object-oriented on my initial project repo. From there maybe we can see if it makes sense or if it's possible to integrate it in scikit-image.

jni:
Hi again @LauLauThom!

> I realised that the library shapely would do most of the work of this PR

Indeed! There's quite a few geometry libraries in Python. However, all are overkill when one wants to overlap simple rectangles! :joy: I still think this is a valuable addition to skimage.

I've made a few changes to the code and pushed to your branch, just to make things more Pythonic, match our docstring standards, etc.

I think this is ready except for minor discussions like whether we want to call the functions `is_intersecting` and `intersection_rectangle` or `intersect` and `intersect_rectangle`.

I'm also wondering whether instead of raising a ValueError, we can somehow produce a "null rectangle", e.g. top_left == bottom_right == (0, 0). We can then also add a `__bool__` method to the class, so you could do `if intersect(r1, r2):`, where the same function produces a rectangle *and* serves as a check.

@LauLauThom @scikit-image/core any thoughts on the above?

jni (approved):

skimage/metrics/overlap.py
```diff
@@ -43,34 +43,26 @@ def __init__(self, top_left, *, bottom_right=None, dimensions=None):
...
-           self._bottom_right == rectangle2._bottom_right):
+        if (self.top_left == rectangle2.top_left and
+           self.bottom_right == rectangle2.bottom_right):
             return True
 
         return False
 
-    def get_top_left(self):
-        """Return the (r,c)-coordinates for the top left corner."""
-        return self._top_left
```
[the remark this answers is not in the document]
grlee77: I see you did use that approach for `height` and `width` later

grlee77:
This looks good to me.

I had a question regarding whether or not we want to allow users to modify the rectangle's corners after creation via assignment to `top_left` or `bottom_right`? If so, perhaps we should use properties with a setter to validate that the provided value is a tuple that matches the length of the opposite corner.

jni:
> I had a question regarding whether or not we want to allow users to modify the rectangle's corners after creation via assignment to top_left or bottom_right? If so, perhaps we should use properties with a setter to validate that the provided value is a tuple that matches the length of the opposite corner.

Yeah I thought about this. I didn't really see a super-strong reason to prevent this, but it would be useful to have a getter/setter property to make sure that they are always arrays.

@grlee77 what do you think about my `__bool__` proposal allowing us to merge intersection rectangle and intersection testing?

LauLauThom:
> Indeed! There's quite a few geometry libraries in Python. However, all are overkill when one wants to overlap simple rectangles! 😂 I still think this is a valuable addition to skimage.
> 
> I've made a few changes to the code and pushed to your branch, just to make things more Pythonic, match our docstring standards, etc.

Looks great ! Thanks !
Maybe we should add the following additional checks in the constructor : 
- `np.all(top_left < bottom_right)` 
- `np.all(dimensions > 0)` 

to make sure that we don't have inverted rectangles with the bottom_right actually above or on the left of the top_left corner.
With such inverted rectangles, I think the `is_intersecting` and `intersect_rectangle` would not work as expected.

> I think this is ready except for minor discussions like whether we want to call the functions `is_intersecting` and `intersection_rectangle` or `intersect` and `intersect_rectangle`.

Shorter names sound good to me !


> I'm also wondering whether instead of raising a ValueError, we can somehow produce a "null rectangle", e.g. top_left == bottom_right == (0, 0). We can then also add a `__bool__` method to the class, so you could do `if intersect(r1, r2):`, where the same function produces a rectangle _and_ serves as a check.

I like this idea of returning a null/false rectangle when not intersecting.
Wondering  if the `__bool__` could return false for any rectangle with `top_left == bottom_right` and not for only for the (0,0) case. So for the cases when the rectangle is just a point.   
Or would someone have a use case for which these "point rectangle" would be "valid" and should evaluate to true 🤔....

grlee77:
> @grlee77 what do you think about my __bool__ proposal allowing us to merge intersection rectangle and intersection testing?

It seems reasonable to me as long as we document that capability

jni:
While fiddling with the bool class, I realised that the current implementation treats pixels as little squares, which [they are not](https://www.cs.princeton.edu/courses/archive/fall00/cs426/papers/smith95b.pdf). I would instead like to consider (as we do in most places in skimage and napari, and as matplotlib does, and as we are generally moving to make uniform) that pixels are point samples at integer coordinates, and consider the rectangles as going from point to point (floating or not). See [this image](https://github.com/napari/napari/issues/310#issuecomment-506002559) for reference.

This changes how things behave: a rectangle with top-left at (0, 0) and bottom right at (1, 1) has area 1, not 4, and it has *no* overlap with a rectangle from (1, 1) to (2, 2).

@LauLauThom would this be workable for you or do you *absolutely need* the little-square-pixel model?

LauLauThom:
> would this be workable for you or do you _absolutely need_ the little-square-pixel model?

Hello @Juan, you are right with the current little square model, thanks for the reference by the way ! Always nice to learn things on the way.
I think the float-coordinates convention should not be a problem, it would only impact this particular module .
We just have to correct the constructor and the functions accordingly.
Actually if I am not mistaking it would remove the -1/+1 offset, so even more readable 😊

The only slight drawback is that it might differ from the original implementation but I don't expect people to use both implementations, and we did not mention one or the other "pixel-model" in the paper so I am totally fine with it !
I just have to document it.

I can implement the changes if you haven't already.

jni:
@LauLauThom just did it. :blush: Could you please give this a review to make sure it meets your needs, and @scikit-image/core can we get some more eyeballs on this before merging? :eyes: :pray:

LauLauThom:
Thanks, all good for me !
I just corrected a typo in a docstring and added additional checks in the constructor as mentioned previously:

> Maybe we should add the following additional checks in the constructor :
> 
> * `np.all(top_left <= bottom_right)`
> * `np.all(dimensions >= 0)`

jni:
The test failures appear unrelated to this, see:

https://github.com/scikit-image/scikit-image/pull/5316#issuecomment-814465568

skimage/metrics/tests/test_overlap.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_constructor_bottom_corner():
+    assert tuple(rectangle3.top_left) == (0, 0)
+    assert tuple(rectangle3.dimensions) == (2, 2)
+    with testing.raises(ValueError):
+        Rectangle((2,2), bottom_right=(0, 0))
+
+def test_intersection():
+    assert not intersect(rectangle1, rectangle2)
+    assert intersect(rectangle1, rectangle3)
```
rfezzani, line 41: Shouldn't we test for the intersection value? I suggest also to test for the case where `rectangle1` in fully included in `rectangle2` and for every case that intersection is symmetric (`intersect(r1, r2) == intersect(r2, r1)`).

rfezzani (commented):
Thank you @LauLauThom and welcome :slightly_smiling_face: . @scikit-image/core, do we keep the 2D naming and doc or do we make more explicit that the functionality is nD?

skimage/metrics/tests/test_overlap.py
rfezzani: We should also test for the edge cases:
- rectangles only share a corner
- rectangles share an edge.

rfezzani:
> This changes how things behave: a rectangle with top-left at (0, 0) and bottom right at (1, 1) has area 1, not 4, and it has _no_ overlap with a rectangle from (1, 1) to (2, 2).

I bet that we will have GH issues filled related to this.
I know that pixels are not little squares (we already had such discussion in the past :wink: ), but with most graphical image manipulation tools, when you select from pixel (0, 0) to pixel (1, 1), you will get 4 pixels selected. This is more a question of "pixel count" vs "real" rectangle area.

Can't we provide both options to the user? (`Rectangle.area` and `Rectangle.pixel_count`)?

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,198 @@
...
+
+    >>> r2 = Rectangle((10, 10), dimensions=(3, 3))
+    >>> if not intersect(r1, r2):
+    ...     print('r1 and r2 are disjoint')
+    r1 and r2 are disjoint
+    """
+    if _disjoint(rectangle1, rectangle2):
+        # return the "null rectangle" if they are disjoint
+        ndim = rectangle1.ndim
+        return Rectangle((0,) * ndim, bottom_right=(0,) * ndim)
```
rfezzani, line 173: This convention may give some counter intuitive logic!
```python
r1 = Rectangle((0, 0), bottom_right=(1, 1))
r2 = Rectangle((2, 2), bottom_right=(3, 3))

r1_r2_intersection = intersect(r1, r2)  # Rectangle((0, 0), (0, 0))

intersect(r1, r2) is False
intersect(r2, r1_r2_intersection) is False
intersect(r1, r1_r2_intersection) is True #!!!!
```

rfezzani:
Here are some more thoughts about this PR:
- What about `BoundingBox` as an alternative name for the `Rectangle` class?
- Do we want to deal with sub-pixelic values? If no, the "area" vs "pixel count" problem is less controversy...
- Are negative coordinates allowed? I don't think so in the "image" context...
- I think that we need to define  a "null" rectangle different from the "point" rectangle that is an edge case of the intersection between two rectangles.

skimage/metrics/overlap.py
jni: I don't get that result @rfezzani?

```python
In [1]: from skimage.metrics import Rectangle, intersect

In [2]: r1 = Rectangle((0, 0), bottom_right=(1, 1))

In [3]: r2 = Rectangle((2, 2), bottom_right=(3, 3))

In [4]: r12 = intersect(r1, r2)

In [5]: r12
Out[5]: Rectangle((0, 0), bottom_right=(0, 0))

In [6]: bool(r12)
Out[6]: False

In [7]: bool(intersect(r1, r2))
Out[7]: False

In [8]: bool(intersect(r2, r12))
Out[8]: False

In [9]: bool(intersect(r1, r12))
Out[9]: False
```

intersections are only true if they have more than 0 area, which R((0, 0), (0, 0)) does not.

jni:
> I bet that we will have GH issues filled related to this.
> I know that pixels are not little squares (we already had such discussion in the past 😉 ), but with most graphical image manipulation tools, when you select from pixel (0, 0) to pixel (1, 1), you will get 4 pixels selected. This is more a question of "pixel count" vs "real" rectangle area.
>
> Can't we provide both options to the user? (Rectangle.area and Rectangle.pixel_count)?

I think we would have an equal number of issues relating to why a rectangle from (0, 0) to (1, 1) has area 4. It defies all logic. =) A big problem with the pixel count idea is that we don't know whether to use closed intervals or half open intervals for integer coordinates. We have already had many such discussions about the `draw` module. It's only by convention that (1, 1), (0, 1), and (1, 0) would be included, and, given that we are in Python, anyway a half-open convention is more consistent with the rest of the language (ie Python and NumPy indexing).

With floating point values, we *could* conceivably provide a `.pixel_count`, but again, I think that is better done with fractional coordinates, as then you don't have to deal with the ambiguity of closed vs open vs half-open intervals. And then you might as well draw `Rectangle((-0.5, -0.5), bottom_right=(1.5, 1.5))` and then everything is consistent!

> What about BoundingBox as an alternative name for the Rectangle class?

🤷 I'm not fussed. I can envision having more geometric primitives in the future and then BoundingBox is more constraining. Ironically! 😂 

> Do we want to deal with sub-pixelic values? If no, the "area" vs "pixel count" problem is less controversy...

I think it's *more* controversial if we don't allow subpixel values, for the reasons stated above. If you allow floating point (and the code trivially does, so we might as well), then all approaches can be expressed using offsets.

> Are negative coordinates allowed? I don't think so in the "image" context...

I don't see why we would constrain it. Nothing prevents negative coordinates in the code, and I can imagine people using it, e.g. if we start having broader pixel spacing support in scikit-image.

> I think that we need to define a "null" rectangle different from the "point" rectangle that is an edge case of the intersection between two rectangles.

As shown above, it's only an edge case if you use the "pixel is a square" model. You're just adding ⛽ to the 🔥! 😂 Having said that, it's true that if we ever want to support "points" and "lines", then we probably have to do something special for null. But I don't think that's urgent until that support becomes a requirement.

skimage/metrics/overlap.py
rfezzani: Huh... Honestly, I did not ran the code, but I expected that the corner of a rectangle is included in it... Isn't it what we want?

LauLauThom: I think @rfezzani is having a good point. 
To me it also feels that eventhough `Rectangle((0, 0), bottom_right=(0, 0))` has a 0 area it should __intersect__ with any rectangle starting at top_left (0,0).

However the overlap would indeed  be 0.

Shapely actually distinguish intersection (any point in common) and [overlap ](https://shapely.readthedocs.io/en/latest/manual.html#object.overlaps) (more than one point).
I think it would be good to have a similar approach for consistency of the python language.
This [reference](https://giswiki.hsr.ch/images/3/3d/9dem_springer.pdf) (page 3) of shapely doc seems to provide rock-solid definitions for these relations.

So should we rename the current `intersect` to `overlap` ?
Do we need a shapely's like `intersect` ? as far as I am concerned I would not need it.

rfezzani:
Well, I looked at the PR from the perspective of providing a tool for evaluating detection algorithms... That's where the "pixel count", "sub pixel" and "negative coord" topics comes...

It seems that I misunderstood the subject, but in this case, the need for the integration of such geometric primitives in an image library is less clear to me :confused:...
Maybe can we clarify the scope of this PR...

I think that we still need to find a better class name to highlight the nD support...

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,198 @@
...
+        return self.__repr__()
+
+    @property
+    def area(self):
+        """Return the rectangle area in pixels."""
+        return np.prod(self.dimensions)
+
+    @property
+    def dimensions(self):
+        """Return the (height, width) dimensions in pixels."""
```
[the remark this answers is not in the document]
LauLauThom: I think it should also NOT say "in pixels", the docstring was previously written with the little-square model in mind.
Like top-left at (0, 0) and bottom right at (1, 1) was previously dimensions (2,2) but now it corresponds to (1,1) in "point-samples".
So either just "dimensions" in the docstring, or "geometrical dimensions", or no docstring.

LauLauThom:
> > What about BoundingBox as an alternative name for the Rectangle class?
> 
> 🤷 I'm not fussed. I can envision having more geometric primitives in the future and then BoundingBox is more constraining. Ironically! 😂

Thinking about possible extension to other shapes in the future I also though that BoundingBox sounds more constraining.
However when we think of N-d shapes, Rectangle sounds limited to 2D while BoundingBox is fine for both 2D and 3D.

What about renaming the module name ? It's not just about `overlap` but more about `shapes` in general or ? 

> the need for the integration of such geometric primitives in an image library is less clear to me 😕...

Actually I mentioned in a previous comment that the job could be done with another library dedicated to shapes 😅, but since there is sometimes a need for region-of-interest in image-processing, a built-in scikit-image solution could be interesting. 


> I'm also wondering whether instead of raising a ValueError, we can somehow produce a "null rectangle", e.g. top_left == bottom_right == (0, 0). We can then also add a `__bool__` method to the class, so you could do `if intersect(r1, r2):`, where the same function produces a rectangle _and_ serves as a check.

Following @rfezzani comment, I tend to think that the `Rectangle((0,0), (0,0)` evaluating to false is maybe not a so good idea anymore, sorry @jni. 
It's not intuitive at least and _"Explicit is better than implicit"_ isn't it 😉
Especially if we think about negative coordinates domain, it's not like being at the origin of the coordinate system is a special case compared to other 0-area rectangles.

What about just returning `None` when the rectangle just don't overlap (or intersect). then we can still do 
```python
if not intersect(r1,r2):
     pass
```
Not super Object-Oriented but convenient

skimage/metrics/overlap.py
rfezzani: Thank you @LauLauThom for these inverstigations. I like the idea of using `overlap` instead of `intersect`.
What if the `overlap` function returns the overlap region area instead of a `Rectangle` object as `intersect` does?
This way, there is no need for defining `Rectangle.__bool__` and the syntax
```python
if overlap(r1, r2):
    ...
```
Is still valid.

LauLauThom: I think by returning a Rectangle we thought of being more generic, like for debugging you could actually see the intersection rectangle.
But since it's hard to find an intuitive and consistent solution for non-overlapping pair of rectangles (I though of returning None instead of a null rectangle otherwise), I would be totally fine with this simpler solution of directly returning the area.
What do you think @jni ?

rfezzani:
> I'm not fussed. I can envision having more geometric primitives in the future and then BoundingBox is more constraining. Ironically! joy

In the context of ROI definition, which other geometric primitives may we need? I only think of sphere and ellipsoid, which both are "nD" names. Bounding box, or more precisely axis aligned bounding box (AABB) is not confusing IMHO.

We may also think about more general (and cool :smiling_face_with_three_hearts: ) structures like AABB trees or octrees...

skimage/metrics/overlap.py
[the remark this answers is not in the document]
LauLauThom: 200283fdb646a2c151bc920e50e74b72e5378927 Updated the docstring for dimensions and area

LauLauThom:
> @scikit-image/core, do we keep the 2D naming and doc or do we make more explicit that the functionality is nD?

Actually about 2D/nD topic, while most of the current code is compatible with nD, there are some functions/attribute like `area` which for 3D or nD would not be super intuitive (which side of the 3D bounding box are we returning the area ?).
Same if we want a `volume` attribute it would work for 3D not 2D.

So I was thinking to rename the current `Rectangle` class as a mother class `BoundingBox` which define everything that applies to any N dimensions.
Then for 2D and 3D we could define daughter classes extending it (e.g `BoundingBox2D`), and that would contain the dimensional-specific attributes such as `area` for 2D or `volume` for 3D.

That would also give the opportunity to have dedicated docstrings for `overlap` to clarify that it returns an area for a `BoundingBox2D` and a volume for a `BoundingBox3D`. Like below.
```python
class BoundingBox():

   def overlap(r1, r2):
      pass

class BoundingBox2D(BoundingBox):
  
  def area(self):
        pass

  def overlap(r1, r2):
      """Return the overlapping area between bounding box r1 and r2."""
       super(BoundingBox).overlap(r1, r2)

 class BoundingBox3D(BoundingBox):
   
    def volume(self):
       pass

    def overlap(r1, r2):
        """Return the overlapping volume between bounding box r1 and r2."""
         super(BoundingBox).overlap(r1, r2)

**Comment by jni** [comment]
Regarding "geometry in an image library", the domains intersect (heh) so it's really not surprising that we would need *some* geometry primitives. @rfezzani we also deal with points coordinates and have some rather fancy computations (implemented by you ;) to do with point sets, because it is common to want to deal with point sets when dealing with images. Same with rectangles — indeed the goal here is to do computations such as iou and non-maximum suppression with rectangles, and both of these things occur in the context of images.

@LauLauThom I must say I really don't like the specialisation proposal. We have discussed this problem with terminology not being properly n-dimensional before [here](https://github.com/scikit-image/scikit-image/issues/3812) in the context of regionprops. My suggestion then would be to do the same here: we have a single object called a Box (though I'm still ok with a Rectangle), and it has .area, .ndim, and .volume, and the area is not defined above 2D. I would argue that .volume should evaluate to .area in 2D because then that one attribute works for nD. (afaik the 'volume' of a 4D hyperrectangle is the product of the extents across all dimensions, not just 3)

Finally, I'm a-ok with returning None in the case of fully disjoint rectangles. (though Rectangle((0, 0), (0, 0)) should still be False!) It will also evaluate to false and be strictly more informative than the current implementation.

File: skimage/metrics/overlap.py [hunk 8]
```diff
@@ -0,0 +1,206 @@
...
+"""
+from __future__ import annotations
+
+import numpy as np
+
+
+class Rectangle:
+    """Construct a rectangle consisting of top left and bottom right corners.
+
+    The contructor uses the (r,c) coordinates for the top left corner and
```
hmaarrfk, line 14:
```suggestion
    The contructor uses the ``(r, c)`` coordinates for the top left corner and
```

rfezzani:
Considering subpixel values is a good thing I think.
Supporting negative values doesn't hurt since we all agree that `Rectangle((0, 0), bottom_right=(0, 0))` mustn't be a special case.
Returning `None` for non overlapping rectangles is a good option I think.
I like the concepts of `intersect` and `overlap`, it helps us to manage the edge cases of corner and edge intersections.
What about `integral` to replace `area` or `volume`? It is valid in any dimension...
I still think that defining an `element_count` method (instead of `pixel` or `voxel` count) will be of great help...
I don't like the specialization proposal either :confused:.

Again, there is no controversy: pixels are not little squares. But we usually manipulate images in a cell centred model with piecewise constant hypothesis:
```python
>>> img = np.arange(4).reshape((2, 2))
>>> plt.imshow(img)
```
Figure_1
As you mentioned @jni 
> It's only by convention that (1, 1), (0, 1), and (1, 0) would be included

We can not ignore conventions... Conventions are good, aren't they? :wink:

jni:
> We can not ignore conventions... Conventions are good, aren't they?

Yes but they are arbitrary! I'm saying, we have to adopt *one* convention, let's adopt the half-open convention, which matches many other conventions in Python *and* matches the geometric calculation. If I look at the coordinates on the matplotlib plot above, and I draw a rectangle from (0, 0) to (1, 1), that rectangle has area/volume/integral 1, not 4! If you want to make sure you get all the little squares, you can make the rectangle go from (-0.5, -0.5) to (1.5, 1.5) — which is intuitive within *all* conventions, because with fractional values I don't have to wonder whether the borders are included.

> we have to adopt *one* convention

:point_up: yes, we can do extra effort and support 7 different conventions, but imho I think strong adherence to a smaller number of conventions and clear documentation of those conventions results in both less maintenance and less confusion from users.

jni (approved):
I'm happy with where this is now. :tada: @rfezzani imho we should discuss further tweaks in future PRs, and let @LauLauThom move on with his life. :joy: Otherwise I worry that this PR could be stuck in decision limbo for a very long time.

Thank you @LauLauThom for coming with us on this journey! :sweat_smile:

rfezzani:
I am sorry @jni and @LauLauThom , my intention is not to block this PR, but I still don't feel comfortable with it :confused:.

The zero-area rectangle edge case provides some strange logic:
- a zero-area rectangle doesn't intersect himself
-  the `intersection_over_union` of a zero-area rectangle with himself produces a division by zero.

> but imho I think strong adherence to a smaller number of conventions and clear documentation of those conventions results in both less maintenance and less confusion from users.

All right, but this is still not the case here, the behavior is not clearly documented...

LauLauThom:
I agree there is still some work to do, and additional tests to add.
I will tackle that when I have some time 😉 It's surely taking more time that I would have expected 😅 but it's good to discuss these things rather than making breaking-changes in the future, and I am also learning stuff on the way !
Thanks for your time !

jni:
> a zero-area rectangle doesn't intersect himself

huh?

```python
In [1]: from skimage.metrics import Rectangle, intersect                                                                      
In [2]: r0 = Rectangle((0, 0), bottom_right=(0, 0))                                                                           
In [3]: r1 = Rectangle((0, 0), bottom_right=(0, 0))                                                                           
In [4]: intersect(r0, r1)                                                                                                     
Out[4]: Rectangle((0, 0), bottom_right=(0, 0))

In [5]: r2 = Rectangle((1, 1), dimensions=(0, 0))                                                                             
In [6]: r3 = Rectangle((1, 1), bottom_right=(1, 1))                                                                           
In [7]: intersect(r2, r3)                                                                                                     
Out[7]: Rectangle((1, 1), bottom_right=(1, 1))
```
?

(Note that the above all evaluate to False)

> the intersection_over_union of a zero-area rectangle with himself produces a division by zero.

Ok, we could catch this, but, technically, it's basically right?

```python
In [8]: from skimage.metrics import intersection_over_union                                                                   
In [9]: intersection_over_union(r2, r3)                                                                                       
/home/jni/projects/scikit-image/skimage/metrics/overlap.py:203: RuntimeWarning: invalid value encountered in long_scalars
  return intersection.area / union_area
Out[9]: nan
```

> All right, but this is still not the case here, the behavior is not clearly documented...

This one I must agree with. =) We should definitely add a gallery example that explains basic usage and also gives details of the conventions. I'll try to work on that next.

rfezzani:
> > a zero-area rectangle doesn't intersect himself
>
> huh?

I meant `bool(intersect(r1, r1))` is `False` :slightly_smiling_face:

jni:
Ah! Indeed. Well, actually, now that we do the None thing, maybe we remove the bool cast, or have it always be True? The other calculations just work now, since `intersect(a, b).area` will be 0.

Again, I think `nan` is consistent for empty rectangles so I reckon leave it.

So the todo right now looks like
- Rectangles always evaluate to True
- add a gallery page explaining all the conventions in here (maybe some explanations go in the class docstring)

Right?

rfezzani:
> (maybe some explanations go in the class docstring)

Yes, a `Notes` section in the `Rectangle` class and the `intersect`function docstrings would be great! :wink: 
Still the nD questions :confused:

LauLauThom:
My last commit might trigger some additional discussion 😅
I created a function `overlap` distinct from `intersect`.

The reason is that while it is working with `intersect` too (overlapping rectangles always intersect), with cases when the 2 rectangles share only 1 side or 1 point in common we are creating a 0-area rectangle and still computing the union area and the ratio of areas, while we could directly tell that they don't overlap and IoU is 0.

It sounds like a extra-work for a tiny optimisation, but I am thinking that this calculation might be repeated for very large number of bounding-boxes and using the simplest brute force approach of comparing every pair of bounding-box the calculation is like scaling in factorial of (n-1) (first box compared to all other n-1 boxes, then next box compared to all except first box and itself...). 
so we might want to have this distinction to save some time

Or more generally if someone is interested in `overlap`, currently he would need to first call `intersect` and check that the area of intersection is different than 0. It's okay but we could directly tell if there is an overlap with the boolean test on the coordinates.
Shapely also does this distinction `intersect`/`overlap`.

jni:
> My last commit might trigger some additional discussion sweat_smile
> I created a function overlap distinct from intersect.

:facepalm: 

:joy: Ok, point taken, although in scikit-image we prefer clear code over speed. But I see that this would be significant (O(2-3x)), so you're probably right. However, my main issue is that I would have no idea in my head how to remember which one of intersect and overlap actually does the thing vs tests the thing. I suggest instead that you remove overlap again and make "disjoint" public — I think that `disjoint`, `intersect`, and `BoundingBox` is clearer than `overlap`, `intersect`, and `BoundingBox`. btw I would expand BBox to BoundingBox — if I was a newcomer, BBox would be a very mysterious name!

LauLauThom:
> I would expand BBox to BoundingBox — if I was a newcomer, BBox would be a very mysterious name!

Done ;)

> I suggest instead that you remove overlap again and make "disjoint" public

I put `disjoint` public too. However I am not really convinced about removing `overlap`...
The thing is `disjoint == not intersect` : the intersection includes any "contact" be it a single point, edge or overlap of the bounding-boxes. So if you have overlap you have intersection.
However the opposite is not true, 2 BoundingBox with point or edge-intersection are not overlapping.

We could remove instead `intersect` since it's redundant with `disjoint` although `intersect` might be more intuitive than `not disjoint`.

Actually the term `Intersection Over Union` is slightly misleading, since we are interested to know if there is an overlap 😅

> I would have no idea in my head how to remember which one of intersect and overlap actually does the thing

I tried to clarify it in the function docstring but maybe it could be clearer.
There are some tests showing that r2/r3 intersect but do not overalp, but it could be a good gallery example too.

jni:
> > I would have no idea in my head how to remember which one of intersect and overlap actually does the thing
> 
> I tried to clarify it in the function docstring but maybe it could be clearer.
> There are some tests showing that r2/r3 intersect but do not overalp, but it could be a good gallery example too.

Sorry, I meant to say: if I looked in the docstrings I 100% would know, but I want to do `skimage.metrics.<TAB>` and know from the list in the tab completion, roughly, what each function does. Just from the name.

... And I just saw that you exchanged the meaning of `intersect` and `overlap` compared to what was there and what I expected, so my concerns were well founded! :joy: 

Looking at the code, I don't think I understood the code and your proposal properly before. Certainly now `intersect` and `disjoint` are completely redundant. I would like to rename `overlap` back to `intersect`. And thinking more about it, I don't think the savings from not building rectangles where the corners touch (a tiny proportion in real-world cases, I'm sure) are going to be worth the API expansion. (ie my 2x-3x estimate above *might* hold — or it might *not!* — in terms of the cost of building vs not, but this will happen only for a tiny proportion of rectangles in real-world use cases, so I don't think it's a problem.)

In short, I'd like to go back to the reduced API (BoundingBox, intersect, intersection_over_union), perhaps making disjoint public, and only expand it if it is *demonstrated* to have significant real world benefits. (ie with benchmarks).

LauLauThom:
You are right about the computational benefit, I overlooked that 0-area BBox __were__ evaluated to false, so we indeed we were not computing the union area in this case.
And `overlap` for 0-area BBoxwas problematic so I removed it.

Now:
- Bboxes always evaluate to True (we agreed that above)
- The IoU checks if `intersect` returns None or a 0-area Bbox
- IoU of a BBox with itself is always 1 also for 0-area BBoxes 

I added a few more tests as well.
I think except the gallery that remains to do it looks otherwise good now.

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,293 @@
...
+#         np.any(bbox2.bottom_right <= bbox1.top_left)):
+#         return None  # below or equal contrary to disjoint: strictly below
+
+#     new_top_left = np.maximum(bbox1.top_left, bbox2.top_left)
+#     new_bottom_right = np.minimum(
+#             bbox1.bottom_right, bbox2.bottom_right
+#             )
+#     return BoundingBox(new_top_left, bottom_right=new_bottom_right)
+
+
```
rfezzani, line 257:
```suggestion
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,293 @@
...
+def intersection_over_union(bbox1, bbox2):
+    """
+    Ratio intersection over union for a pair of BoundingBoxes.
+
+    The intersection over union (IoU) ranges between 0 (no overlap) and 1 (full
+    overlap) and has no unit.
+    For 2D BoundingBoxes, the IoU corresponds to a ratio of areas.
+    For 3D BoundingBoxes, the IoU corresponds to a ratio of volumes.
+    For higher dimensions, the IoU corresponds to a ratio of the BoundingBox integrals.
```
rfezzani, line 266:
```suggestion
    For higher dimensions, the IoU corresponds to a ratio of the BoundingBox
    integrals.
```
just pep8 formatting ;)

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,293 @@
...
+
+    The IoU of a BoundingBox with itself is always 1, even for 0-integral BoundingBox.
```
rfezzani, line 268:
```suggestion
    The IoU of a BoundingBox with itself is always 1, even for 0-integral
    BoundingBox.
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,293 @@
...
+    bbox1, bbox2: BoundingBox
+        The input BoundingBoxes.
+
+    Returns
+    -------
+    iou : float
+        The intersection over union value.
+    """
+
+    if (bbox1 == bbox2): # also prevent issue with IoU of 0-integral BoundingBox with themselves
```
rfezzani, line 281:
```suggestion
    if (bbox1 == bbox2):
        # Prevent issue with IoU of 0-integral BoundingBox with themselves
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,293 @@
...
+        return 1
+
+    intersection = intersect(bbox1, bbox2)
+
+    if (intersection is None) or (intersection.integral) == 0:
+        return 0
+
+    union_integral = (bbox1.integral +
+                      bbox2.integral -
+                      intersection.integral)
```
rfezzani, line 291:
```suggestion
    union_integral = bbox1.integral + bbox2.integral - intersection.integral
```

skimage/metrics/__init__.py
```diff
@@ -1,6 +1,11 @@
 from ._adapted_rand_error import adapted_rand_error
 from ._variation_of_information import variation_of_information
 from ._contingency_table import contingency_table
+from .overlap import (
+        BoundingBox,
+        intersect,
+        intersection_over_union,
+        )
```
rfezzani, line 8:
```suggestion
        disjoint,
        )
```
`disjoint` is public now :slightly_smiling_face:

skimage/metrics/__init__.py
```diff
@@ -9,14 +14,18 @@
...
+        'mean_squared_error',
+        'normalized_mutual_information',
+        'normalized_root_mse',
+        'peak_signal_noise_ratio',
+        'structural_similarity',
+        'hausdorff_distance',
+        'hausdorff_pair',
+        'Rectangle',
+        'intersect',
+        'intersection_over_union',
```
rfezzani, line 30:
```suggestion
        'disjoint',
        'intersection_over_union',
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,293 @@
+"""
+Module defining a number of functions to quantify the overlap between shapes.
+
+For instance BoundingBoxes used for object-detection.
+"""
+from __future__ import annotations
+
```
rfezzani, line 7:
```suggestion
```

imported but not used

rfezzani (commented):
Thank you @LauLauThom, it looks very good :+1:. I just left some suggestions. Maybe should we add tests for nD support and `disjoint` and then we are all good :smile:. Thank you again :wink:

LauLauThom:
TODO:
- [ ] test with disjoint
- [ ] test with nD BoundingBox
- [ ] rename module from `overlap` to `shape` ? (what do you think ?)
- [ ] Move the module from `metrics` to a more adapted package ?

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,246 @@
...
+        return self.bottom_right[-1] - self.top_left[-1]
+
+    @property
+    def ndim(self):
+        return len(self.top_left)
+
+    def __bool__(self):
+        return True
+
+    def __eq__(self, other: BoundingBox):
```
LauLauThom, line 82: I also have test failing locally after approving the reviewed changes 😅
Somehow the BoundingBox type is not recognized. Is it something to change in the package __init__ ?

rfezzani: Hum... I thin k this is due to Postponed Evaluation of Annotations ([PEP 563](https://www.python.org/dev/peps/pep-0563/))...
You are probably using python 3.7 right?
We must add in the imports
```
import sys
if sys.version_info.major == 3 and sys.version_info.minor ==7:
    from __future__ import annotations
```
Alternatively, we may remove the type annotations :slightly_smiling_face:

jni: @rfezzani you don't need to import sys. Just `from __future__ import annotations`. Which was in there a few commits ago. ;)

rfezzani: I proposed this to avoid warnings described in the [first draft](https://mail.python.org/pipermail/python-ideas/2017-September/047031.html) of the PEP:

> Deprecation policy
> ------------------
> 
> In Python 3.7, a ``__future__`` import is required to use the described
> functionality and a ``PendingDeprecationWarning`` is raised by the
> compiler in the presence of type annotations in modules without the
> ``__future__`` import.  In Python 3.8 the warning becomes a
> ``DeprecationWarning``.  In the next version this will become the
> default behavior.

But, you are right, this seems not necessary in the last version of the PEP :).
But again, type hints are not yet mandatory in `skimage`, so we can avoid it :).

jni: In this case, I annotated it because I think it is a nice documentation when reading the function, ie we won't compare with a 4-tuple, or whatever else someone might want to do 😂

LauLauThom: I reverted the commit to keep the from future import statement for the moment.
We can always optimize it later.

LauLauThom:
Hi guys, I just added a few more tests with 3D bounding-boxes and I realize there are use cases between bounding-boxes of different dimensions which currently throw ValueErrors. For the following functions, with possible solutions
- disjoint -> not implemented ?
- intersect -> not implemented
For these 2, the intersection of a 3D BoundingBox with a 2D BoundingBox yields another 2D BoundingBox isn't it ?
But I can't find a generic nD solution that would work for any nD bounding box with another BoundingBox of p != n dimensions.

- IoU -> 0 
I think the IoU should be exclusively defined between BoundingBoxes of identical dimensions, other the ratio IoU is not homogeneous in terms of dimensions. It should always be a ratio of nD/nD (area/area, volume/volume, not area/volume for instance)

LauLauThom:
Hi all, just taking the temperature here, should we return Not implemented as suggested above and finally merge this PR 😅

stefanv:
I just took a brief scroll through the code, and I have a comment on naming: I think `skimage.measure.intersect` is confusing, given that the `skimage.measure` submodule is not merely about bounding boxes.  I would move `intersect` and similar functions either to static or class methods, whatever feels more appropriate.  `x.intersect(y)` feels like it works, but I could also see `BoundingBox.intersect(x, y)`.

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+    disjoint = (
+            np.any(bbox1.bottom_right < bbox2.top_left)
+            or np.any(bbox2.bottom_right < bbox1.top_left)
+            )
+    return disjoint
+
+
+def intersect(bbox1, bbox2):
+    """
+    Return the BoundingBox corresponding to the intersection of 2 BoundingBoxes.
```
stefanv, line 170:
```suggestion
    Return the BoundingBox corresponding to the intersection of two BoundingBoxes.
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+            )
+    return BoundingBox(new_top_left,
+                       bottom_right=new_bottom_right)
+
+
+def intersection_over_union(bbox1, bbox2):
+    """
+    Ratio intersection over union for a pair of BoundingBoxes.
+
+    The intersection over union (IoU) ranges between 0 (no overlap) and 1 (full
```
stefanv, line 219:
```suggestion
    Intersection over union (IoU) is a metric for the overlap between two bounding boxes.
    
    The IoU ranges between 0 (no overlap) and 1 (full
```

stefanv: May also be worth mentioning the Jacard index, and showing the formula from https://en.wikipedia.org/wiki/Jaccard_index

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class BoundingBox:
+    """
+    Construct an axis-aligned Bounding-Box
+    consisting of top left and bottom right corners.
+
+    The contructor uses the (r,c,..) coordinates for the top left corner and
+    either the coordinates of the botton right corner or the BoundingBox
+    dimensions (height, width,...).
```
stefanv, line 18:
```suggestion
    dimensions ``(height, width, ...)``.
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+import numpy as np
+
+
+class BoundingBox:
+    """
+    Construct an axis-aligned Bounding-Box
+    consisting of top left and bottom right corners.
+
+    The contructor uses the (r,c,..) coordinates for the top left corner and
```
stefanv, line 16:
```suggestion
    The contructor uses the ``(r, c, ...)`` coordinates for the top left corner and
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+"""
+from __future__ import annotations
+
+import numpy as np
+
+
+class BoundingBox:
+    """
+    Construct an axis-aligned Bounding-Box
+    consisting of top left and bottom right corners.
```
stefanv, line 14:
```suggestion
    Axis-aligned bounding box.
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+    The contructor uses the (r,c,..) coordinates for the top left corner and
+    either the coordinates of the botton right corner or the BoundingBox
+    dimensions (height, width,...).
+
+    BoundingBoxes can have 2 (rectangle), 3 (3D BoundingBox) or more dimensions.
```
stefanv, line 20:
```suggestion
    BoundingBoxes can have two (rectangle), three (rectangular prism), or more dimensions.
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+    Parameters
+    ----------
+    top_left : array-like of ints or floats
+        (r,c)-coordinates for the top left corner of the BoundingBox.
+
+    bottom_right : array-like of ints or floats, optional
+        (r,c)-coordinates for the bottom right corner of the BoundingBox.
+
+    dimensions : array-like of ints or floats, optional
+        dimensions of the BoundingBox (height, width). The default is None.
```
stefanv, line 31:
```suggestion
        Dimensions of the BoundingBox ``(height, width, ...)``. The default is None.
        Either `bottom_right` or `dimensions` must be specified.
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+    BoundingBoxes can have 2 (rectangle), 3 (3D BoundingBox) or more dimensions.
+
+    Parameters
+    ----------
+    top_left : array-like of ints or floats
+        (r,c)-coordinates for the top left corner of the BoundingBox.
+
+    bottom_right : array-like of ints or floats, optional
+        (r,c)-coordinates for the bottom right corner of the BoundingBox.
```
stefanv, line 28:
```suggestion
        (r,c)-coordinates for the bottom right corner of the BoundingBox.
        Either `bottom_right` or `dimensions` must be specified.
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+    dimensions : array-like of ints or floats, optional
+        dimensions of the BoundingBox (height, width). The default is None.
+
+    Raises
+    ------
+    ValueError
+        If neither or both of bottom_right and dimensions are provided.
```
stefanv, line 36:
```suggestion
        If neither or both `bottom_right` and `dimensions` are provided.
```

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+    Attributes
+    ----------
+    top_left : array of int or float
+        The top left corner of the BoundingBox.
+    bottom_right : array of int or float
+        The bottom right corner of the BoundingBox.
```
stefanv, line 43: Missing dimensions and a bunch of others

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+            self.bottom_right = np.asarray(bottom_right)
+
+        elif dimensions is not None:
+            dimensions = np.asarray(dimensions)
+            if not (dimensions >= 0).all():
+                raise ValueError("Dimensions should be positive.")
+            self.bottom_right = self.top_left + dimensions
+
+    @property
+    def height(self):
```
stefanv, line 68: Only relevant for 2D; would replace this with the more general `.shape`.

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+    Parameters
+    ----------
+    top_left : array-like of ints or floats
+        (r,c)-coordinates for the top left corner of the BoundingBox.
+
+    bottom_right : array-like of ints or floats, optional
+        (r,c)-coordinates for the bottom right corner of the BoundingBox.
+
+    dimensions : array-like of ints or floats, optional
```
stefanv, line 30: It would be more typical to call this `shape`.

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+    @property
+    def width(self):
+        # use negative indexing in anticipation of nD hyperBoundingBoxes.
+        return self.bottom_right[-1] - self.top_left[-1]
+
+    @property
+    def ndim(self):
+        return len(self.top_left)
+
+    def __bool__(self):
```
stefanv, line 81: What is the point of this?

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+    def __str__(self):
+        return self.__repr__()
+
+    @property
+    def area(self):
+        """Return the area of a 2D BoundingBox."""
+        if self.ndim == 2:
+            return self.integral
+
+        raise NotImplementedError("Area is only defined for 2D.")
```
stefanv, line 110: Why not provide the area for a higher dimensional BBox as well?

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+    @property
+    def volume(self):
+        """Return the volume of a 3D BoundingBox."""
+        if self.ndim == 3:
+            return self.integral
+
+        raise NotImplementedError("Volume is only defined for 3D.")
```
stefanv, line 118: No reason to only implement volume for 3D.

skimage/metrics/overlap.py
```diff
@@ -0,0 +1,251 @@
...
+
+    @property
+    def integral(self):
```
stefanv, line 121: Not sure this is a commonly understood concept; let's stick with volume?

skimage/metrics/tests/test_overlap.py
```diff
@@ -0,0 +1,82 @@
...
+    assert intersect(r1, r5) is None
+
+
+def test_eq_operator():  # Intersection rectangle and == comparison
+    assert rect_inter13 == r3
+    assert r1 != r3D  # BoundingBoxes of different dimensions
+
+def test_eq_other_obj():
+    with testing.raises(TypeError):
+        _ = r2 == (1, 2, 3, 4)
```
stefanv, line 68:
```suggestion
        r2 == (1, 2, 3, 4)
```

stefanv (commented):
Thanks for the contribution, @LauLauThom!

I've left some comments inline.  Please also check the test coverage for the module to ensure that we're hitting all corner cases.

stefanv (changes requested):
Setting the "request changes" flag.

skimage/metrics/overlap.py
LauLauThom: indeed like array.shape. The latest commit corrects that. Thanks

skimage/metrics/overlap.py
LauLauThom: the latest commit documents the other attributes/properties

skimage/metrics/overlap.py
LauLauThom: Thanks for the feedback. You mean the area for a higher dim BBox would be the sum of all the surfaces areas ?
e.g. a 3D bounding box we would return the sum of the area for the 6 faces of the box ?
That would make sense, if we document it accordingly.

We briefly discuss this terminology issue below, see @jni  comment there 

> We have discussed this problem with terminology not being properly n-dimensional before [here](https://github.com/scikit-image/scikit-image/issues/3812) in the context of regionprops. My suggestion then would be to do the same here: we have a single object called a Box (though I'm still ok with a Rectangle), and it has .area, .ndim, and .volume, and the area is not defined above 2D. I would argue that .volume should evaluate to .area in 2D because then that one attribute works for nD. (afaik the 'volume' of a 4D hyperrectangle is the product of the extents across all dimensions, not just 3

Currently, area is only defined for 2D and volume for 3D.

PR #5227, opened 2021-02-10
See if latest Ubuntu image fixes QEMU CPU detection issue

stefanv:


stefanv:
This is in response to the CPython 3.9 failure observed in https://github.com/scikit-image/scikit-image/pull/5197.

It looks like the test suite is now passing.  This could either be because I switched to Ubuntu 20.04 with newer QEmu, or because  [NumPy upgraded OpenBLAS](https://github.com/numpy/numpy/pull/18196).

Whatever the case, it seems to be good news.

hmaarrfk (approved):

jni (approved):

jni:

PR #5228, opened 2021-02-10
WIP: Decorators for helping with the multichannel->channel_axis transition

grlee77:
This PR is related to #4294. It introduces two new decorators to help with the proposed transition from `multichannel` to `channel_axis`:

1.) `deprecate_multichannel_kwarg` is a decorated based on `deprecate_kwarg`, but additionally takes care of converting values from:
 multichannel=False -> channel_axis=None
 multichannel=True -> channel_axis=(-1,)

2.) `channel_as_last_axis` can be used to help automate shuffling of axes to/from the last position. This will not be needed for functdions capable of handling arbitrary channel positions, but is a quick way to adapt many existing functions that all assume channels correspond to the last axis.

For now I just applied these to the existing `montage` function as a concrete example. I want to get initial feedback before extending the approach further. Right now only a single channel axis is supported, but the `channel_as_last_axis` decorator could also be adapted to handle more general cases.


skimage/util/tests/test_montage.py
```diff
@@ -21,15 +21,16 @@ def test_montage_simple_gray():
     assert_array_equal(arr_out, arr_ref)
 
 
-def test_montage_simple_rgb():
+def test_montage_simple_rgb(channel_pos=-1):
```
emmanuelle, line 24: not sure what `channel_pos` is used for here

emmanuelle:
Thanks @grlee77 the approach looks sound to me. So `deprecate_multichannel_kwarg` would be used just during the transition to 1.0 but the `channel_as_last_axis` decorator could stay for a longer time? With the latter decorator and the use of `np.moveaxis`, do I understand correctly that for an array with channels corresponding to the leading axis, `channel_axis=0` + this decorator would lead to memory-contiguous arrays being used in functions such as `denoise_tv_bregman` etc. which do 
```
channel_in = np.ascontiguousarray(image[..., c:c+1])
```
?

skimage/util/tests/test_montage.py
grlee77: It should not be there. I originally started modifying this test class, but later decided to leave this one using `multichannel` and add a new `test_montage_simple_rgb_channel_axes` test case instead to test the `channel_axis` argument.

grlee77:
>  So deprecate_multichannel_kwarg would be used just during the transition to 1.0 but the channel_as_last_axis decorator could stay for a longer time? 

Yes, this way we don't have to make solutions for every functions individually to support alternative channel positions. If there are cases where it makes sense to handle the channels in a different manner, then those functions can implement that internally and stop using the `channel_as_last_axis` decorator. 

> do I understand correctly that for an array with channels corresponding to the leading axis, channel_axis=0 + this decorator would lead to memory-contiguous arrays being used in functions such as denoise_tv_bregman etc. which do

Yes, If the input array is C-contiguous with channels as the first axis then after `np.moveaxis` to the last position the overall array is no longer C-contiguous, but the values for a specific channel will still be C-contiguous. Minimal example

```Python
import numpy as np

a = np.ones((3, 256, 256))
a_reordered = np.moveaxis(a, 0, -1)
# overall array is now no-longer C-contiguous
print(a_reordered.flags)
# but invidual channels are C-contiguous
print(a_reordered[..., 0].flags)
"""

**Review by emmanuelle — APPROVED** [review]
Thanks @grlee77 the PR is ready as far as I'm concerned. @scikit-image/core we need another review please :-)

**Comment by stefanv** [comment]
Looks good to me; I guess we'll know more once we start using this in practice.

**Comment by grlee77** [comment]
I added a couple of additional functions to illustrate additional arguments for the `channel_as_last_axis` decorator

`match_histograms` : this one swaps channels for the first two positional arguments
`apply_parallel`: this one swaps channels for only the second positional argument

This PR now covers all `multichannel` use in `skimage.util` and `skimage.exposure`. To keep the number of changes relatively small, I would like to address other modules in subsequent PRs.

File: skimage/exposure/histogram_matching.py [hunk 2]
```diff
@@ -19,7 +20,10 @@ def _match_cumulative_cdf(source, template):
     return interp_a_values[src_unique_indices].reshape(source.shape)
 
 
-def match_histograms(image, reference, *, multichannel=False):
+@utils.channel_as_last_axis(channel_arg_positions=(0, 1))
+@utils.deprecate_multichannel_kwarg()
+def match_histograms(image, reference, *, multichannel=False,
+                     channel_axis=None):
```
jni, line 26: Since we had the foresight to make multichannel kwarg only here, how about moving it to the final position? =)

skimage/exposure/tests/test_histogram_matching.py
```diff
@@ -35,14 +35,39 @@ def test_match_histograms(self, image, reference, multichannel):
...
+                    np.abs(reference_values - matched_value)
+                ).argmin()
+                assert_almost_equal(matched_quantiles[i],
+                                    reference_quantiles[closest_id],
+                                    decimal=1)
 
+    @pytest.mark.parametrize('channel_axis', (0, 1, -1))
+    def test_match_histograms_channel_axis(self, channel_axis):
+        """Assert that pdf of matched image is close to the reference's pdf for
+        all channels and all values of matched"""
```
jni, line 60: lovely

jni (approved):
@grlee77 I've left a minor suggestion. It'd be great if the docstrings were PEP257 compliant (first line must be a one-liner), but it's in private classes so :shrug: I really don't mind. =)

grlee77:
Thanks @jni, I made the requested modifications.

jni:
Thank you @grlee77! The failed checks appeared to be due to a stalled VM.

Issue #5222, opened 2021-02-05
regionprops_table max_intensity and min_intensity are clipped to integer for a float intensity_image


yfukai:
## Description

When I use `measure.regionprops_table` with a float intensity_image, I found the `max_intensity` and `min_intensity` are clipped to the floor integer and inconsistent with the result of `measure.regionprops`. I suppose this is a bug and would like to ask for opinion.

## Way to reproduce
```python
from skimage import measure
testimg=(np.arange(0,9)*0.1).reshape((3,3))
testmask=np.zeros((3,3),dtype=np.int16)
testmask[:,1]=1

print(testimg)
print(testmask)

res1=measure.regionprops(testmask,testimg)[0]
print(res1.max_intensity,res1.min_intensity,res1.mean_intensity)
properties=["label","area","max_intensity","min_intensity","mean_intensity"]
res2=measure.regionprops_table(testmask,testimg,properties=properties)
print(res2)

""" 
the output is
[[0.  0.1 0.2]
 [0.3 0.4 0.5]
 [0.6 0.7 0.8]]
[[0 1 0]
 [0 1 0]
 [0 1 0]]
0.7000000000000001 0.1 0.4000000000000001
{'label': array([1]), 'area': array([3]), 'max_intensity': array([0]), 'min_intensity': array([0]), 'mean_intensity': array([0.4])}
"""
```

## Version information
```python
# Paste the output of the following python commands
from __future__ import print_function
import sys; print(sys.version)
import platform; print(platform.platform())
import skimage; print("scikit-image version: {}".format(skimage.__version__))
import numpy; print("numpy version: {}".format(numpy.__version__))
```

```python
# your output here
3.7.9 (default, Aug 31 2020, 12:42:55) 
[GCC 7.3.0]
Linux-3.10.0-1127.13.1.el7.x86_64-x86_64-with-debian-buster-sid
scikit-image version: 0.18.0
numpy version: 1.19.1
```

Thanks!

rfezzani:
Thank you for the report @yfukai, I confirm the bug... It is probably related to the fact that `min_intensity` and `max_intensity` are marked as `int` in [COL_DTYPES](https://github.com/scikit-image/scikit-image/blob/master/skimage/measure/_regionprops.py#L72).

yfukai:
[...]

PR #5239, opened 2021-02-16
Regionprops table dtype bugfix (#5222)

yfukai:
## Description

Fixes issue #5222 . The edge test case for `regionprops_table` with a float image is written, which fails with the original version. Then changed the `dtype` for `max(min)_intensity` to float to fix the problem. 

## Checklist

→ not written here, but consistency checked with the [API]( https://scikit-image.org/docs/stable/api/skimage.measure.html#regionprops )
→ not applicable
  existing benchmark
→ not applicable
- Unit tests
→ one added here.
→ checked with autopep8

## For reviewers

  later.


jni (approved):
Great, thank you @yfukai!

I personally would change `.astype(np.double)` to `.astype(self.intensity_image.dtype)`, but I don't think this is very important and could be done in a follow-up PR.

yfukai:
Thank you @jni for your comment! Yeah I also considered that option, but leaving `dtype` free with the `intensity_image` seems to break the test `test_column_dtypes_correct` as the `dtype` is fixed at `COL_DTYPES`. I found it somehow consistent with [the document](https://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.regionprops) since `dtype` are noted as `float`, but I would appreciate if you could improve by follow-up PRs!

jni:
> leaving dtype free with the intensity_image seems to break the test test_column_dtypes_correct as the dtype is fixed at COL_DTYPES

Yeah, I thought this might be the case. This is then ready to merge as far as I'm concerned, we just need another core dev to sign off on it. Thank you again!

mkcor (approved):
Thank you for fixing this!

mkcor:
Merging as the CI failures are timeouts unrelated to the PR.

skimage/measure/tests/test_regionprops.py
```diff
@@ -575,6 +577,17 @@ def test_regionprops_table():
...
 
+def test_regionprops_table_equal_to_original():
+    regions = regionprops(SAMPLE, INTENSITY_FLOAT_SAMPLE)
+    out_table = regionprops_table(SAMPLE, INTENSITY_FLOAT_SAMPLE,
+                                  properties=COL_DTYPES.keys())
+
+    for prop in COL_DTYPES.keys():
+        for i in range(len(regions)):
+            if not isinstance(regions[i][prop], Iterable):
+                assert regions[i][prop] == out_table[prop][i]
```
rfezzani, line 588: Can't we test all the properties here?
```suggestion
        for i, reg in enumerate(regions):
            assert np.array_equal(reg[prop], out_table[prop][i])
```

skimage/measure/_regionprops.py
```diff
@@ -398,15 +398,17 @@ def local_centroid(self):
 
     @property
     def max_intensity(self):
-        return np.max(self.intensity_image[self.image], axis=0)
+        return np.max(self.intensity_image[self.image], axis=0)\
+                 .astype(np.double)
```
rfezzani, line 402: Is this cast necessary?

rfezzani (commented):
Sorry if my review comes too late :confused:. It can be addressed in the coming PRs. Thank you @yfukai

yfukai:
Hi @rfezzani , thank you for your review! I'm sorry this is my first PR and I'm still not used to the reviewing system. I'll reply your comments and check I can update my commit!

mkcor:
@yfukai you did everything right! You don't have to update any past commit. Please create a new PR (starting from current master) to submit further changes.

skimage/measure/_regionprops.py
yfukai: Unfortunately yes for now, since leaving `dtype` free breaks the consistency of the `dtype` with `COL_DTYPES` and makes the test `test_column_dtypes_correct` fail. I guess one might need a bit more of code dev to remove this cast.

skimage/measure/tests/test_regionprops.py
yfukai: Yeah one can, but we need some trick since iterable properties such as `regions[i]["bbox"]` are stored as `out_table["bbox-0"][i]`, `out_table["bbox-1"][i]` etc., for example. (`_props_to_dict` does the job, but I think I shouldn't use it in the test itself as it also casts numbers). I think I was a bit lazy to skip iterable properties, and I will also add some tests for iterable cases and make a new PR!

yfukai: I updated the code to include iterable properties( pull request #5245 ). It's a little bit dirty hack but hopefully it suffices!

Issue #3665, opened 2019-01-18
Why is skimage.restoration.inpaint_biharmonic to slow?


mleraa:
System specs:
Macbook Pro, 8 GB RAM, Intel HD Graphics, 
```
Python 3.6.3
>>> cv2.__version__
'4.0.0'
>>> skimage.__version__
'0.14.2'
```

I am trying to delete the watermarks from a few images. I have tried out the `cv2.inpaint` and `skimage.restoration.inpaint_biharmonic` and I saw some big differences at the result and at the performance.

General approach:
[![enter image description here][1]][1]

Original Image:
[![enter image description here][2]][2]

Mask:
[![enter image description here][3]][3]

Skimage approach:

    import numpy as np
    import skimage.io
    import skimage.restoration
    import skimage.exposure
    import time
    
    img = skimage.io.imread('x')
    msk = skimage.io.imread('x')
    start_time = time.time()
    msk = skimage.exposure.rescale_intensity(msk, in_range='image', out_range=(0,1))
    newimg = skimage.restoration.inpaint_biharmonic(img, msk, multichannel=True)
    print("--- %s seconds ---" % (time.time() - start_time))
    plt.imshow(newimg)
    plt.show()
[![enter image description here][5]][5]

Cv2 approach:

    import numpy as np
    import cv2 as cv
    import matplotlib.pyplot as plt
    
    import time
    
    img = cv.imread('x')
    mask = cv.imread('x',0)
    start_time = time.time()
    dst = cv.inpaint(img,mask,3,cv.INPAINT_TELEA)
    print("--- %s seconds ---" % (time.time() - start_time))
    plt.imshow(cv2.cvtColor(dst, cv2.COLOR_BGR2RGB))

[![enter image description here][4]][4]

Performance:

 - cv2: 0.04929995536804199 seconds 
 - skimage: 116.02286505699158 seconds

Result:
At the front end the deleting is harder on opencv than on skimage. With skimage the deleting is thus more successful, because it looks smoother than opencvs deleting.

**I would like to use skimage as it gives better results. But opencv is faster. What can I do?**

**Disclaimer: This process is for educational purposes only and will not be published, further processed or redistributed**

  [1]: https://i.stack.imgur.com/OLzvm.png
  [2]: https://i.stack.imgur.com/nvyY0.jpg
  [3]: https://i.stack.imgur.com/yLmI7.png
  [4]: https://i.stack.imgur.com/DaHYC.png
[...]

PR #5240, opened 2021-02-17
Fast biharmonic  inpainting

### LINKED ISSUE #2008: Faster inpainting

grlee77:
## Description

This PR accelerates `inpaint_biharmonic` by orders of magnitude through multiple means. I thought I would take a quick look at this and ended up spending quite a few hours!

The actual results should be the same as before. The existing tests were not super extensive, so I added some new ones to try and help make sure I haven't accidentally broken anything.

ASV benchmarks were added and the gallery example was updated to add additional types of degradations.

resolves #2008, #3665

While writing up the PR here, I also found an older PR (#3489) from 2018 that seems to have never been completed.
supersedes #3489

### Benchmark results

For the ASV benchmarks added in this PR, an acceleration factor of >150 (grayscale) or >400 (RGB) was observed. The type of case tested in the benchmark with quite a few small defects was particularly inefficient previously and the gains will be somewhat less on images with only a few isolated defects.

**v0.18.1**:
```
· Discovering benchmarks
[  0.00%] ·· Benchmarking existing-py_home_lee8rx_miniconda3_envs_skimage_dev_bin_python
[ 25.00%] ··· benchmark_restoration.Inpaint.time_inpaint_grey               7.51±0s
[ 50.00%] ··· benchmark_restoration.Inpaint.time_inpaint_rgb               23.0±0s
```

**Current PR**:
```
· Discovering benchmarks
[  0.00%] ·· Benchmarking existing-py_home_lee8rx_miniconda3_envs_skimage_dev_bin_python
[ 25.00%] ··· benchmark_restoration.Inpaint.time_inpaint_grey               45.4±0ms
[ 50.00%] ··· benchmark_restoration.Inpaint.time_inpaint_rgb                53.8±0ms
```

With OpenCV, I got computation times of:
103 ms for uint8 RGB input and mode `cv2.INPAINT_TELEA`
295 ms for float32 input and mode `cv2.INPAINT_TELEA` (have to run each channel individually when using float32)
19.4 ms for uint8 RGB input and mode `cv2.INPAINT_NS`
31.9 ms for float32 input and mode `cv2.INPAINT_NS` (have to run each channel individually when using float32)

So we are now comparable to the performance of the algorithms in openCV, although the results of the method here look visually better to me!

## Implementation Details

The algorithm was not changed, but efficiency improvements were made via a combination of many changes, which I think will make this a bit challenging to review. I have summarized most of the key changes below:

1.) Split into sequential processing of locations within 2 pixels of the image boundary vs. all interior pixels. This allows using simpler logic for the interior pixels (e.g. no need to check that we are within bounds and can always use an identical kernel)

2.) For the boundary locations, the small, asymmetric biharmonic kernels are cached to avoid repeated computations of the same kernels.

3.) The interior pixel loop was simplified and converted to Cython making it much more efficient.
3b.) To enable this, I predetermine sizes for sparse matrix row, column and data vectors rather than dynamically extending lists.
3c.) Rather than using n-dimensional indexing, 1d offsets of all non-zero entries in the biharmonic kernel are precomputed outside the inner loops. This allows just doing simple 1d indexing over the kernel values and offsets.

4.) Processing of channels was moved within `_inpaint_biharmonic_single_region`. This was done because the sparse matrix only needs to be created once and can then be reused for each channel.  In the multichannel case, `spsolve` is called once for all channels which ends up being substantially faster as well.

5.) spsolve is called using `permc_spec='MMD_ATA'` because this column permutation option was empirically measured to be about 30% faster than using the default setting.

6.) I added a `split_into_regions` keyword-only argument, defaulting to False. When True, the prior approach of processing labeled regions individually is used. I found doing this was always slower than processing all regions in one go, but I kept the option because it should require less memory. When False, no call to `label` is used and the system corresponding to the whole image is solved at once.

7.) If using `split_into_regions=True`, bounding boxes as determined by `regionprops_table` are now used to allow processing only local regions around each labeled region in `_inpaint_biharmonic_single_region`.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


benchmarks/benchmark_restoration.py
```diff
@@ -148,3 +153,50 @@ def time_rollingball_threads(self, threads):
         restoration.rolling_ball(data.coins(), radius=100, num_threads=threads)
     time_rollingball_threads.params = range(0, 9)
     time_rollingball_threads.param_names = ["threads"]
+
+
+class Inpaint(object):
+    """Benchmark Rolling Ball algorithm."""
```
emmanuelle, line 159:
```suggestion
    """Benchmark inpainting algorithm."""
```

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,184 @@
...
-from ..measure import label
+from ..measure import label, regionprops_table
+from ._inpaint import _build_matrix_inner
 
 
 def _get_neighborhood(nd_idx, radius, nd_shape):
-    bounds_lo = (nd_idx - radius).clip(min=0)
-    bounds_hi = (nd_idx + radius + 1).clip(max=nd_shape)
+    bounds_lo = tuple(max(p - radius, 0) for p in nd_idx)
+    bounds_hi = tuple(min(p + radius + 1, s) for p, s in zip(nd_idx, nd_shape))
```
emmanuelle, line 17: are the values the same after the change or not? I'm not sure I understand the reason of the change here.

emmanuelle:
This is a very impressive PR and I think we can keep your paragraph of explanations for documenting the different ways of improving performance! Out of curiosity, among the various improvements, were some more crucial than others in terms of performance gain? 

I have started reviewing the PR but I will need more time to complete the review, I'll start again tomorrow.

skimage/restoration/inpaint.py
grlee77: The behavior is the same, but it avoids overhead of NumPy calls on short length 2 or 3 arrays. Originally this function was potentially being called for each value in the mask so it could add up over thousands of calls. After the loop reorganization, it is now only being called for mask locations that are near the boundary, so I doubt it is very critical to keep this change. 

Here is a short example:
```Python
radius = 2
nd_idx = np.asarray([64, 83])

%timeit tuple(max(p - radius, 0) for p in nd_idx)
# 1.48 µs ± 26 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit (nd_idx - radius).clip(min=0)
# 7.99 µs ± 151 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
so the overhead is only around 6.5 microseconds per call on my machine, but it was being called thousands of times. Similar changes were also realized outside of the function as well since the function now returns tuples instead of small arrays.

grlee77:
> Out of curiosity, among the various improvements, were some more crucial than others in terms of performance gain?

Using a line profiler, I initially saw that only a few percent of the overall time was spent in `spsolve` and nearly all time was in building the sparse matrices. Since the matrices were the same for each channel, moving channels to the inner loop directly reduced this overhead by the number of channels, so that was an easy way to accelerate things for multichannel images. Not repeating kernel computations within the loop for non-edge voxels was another large improvement. Otherwise, just many small changes to reduce overhead in the loops like using raveled offsets added up to give substantial additional improvement. 

Cythonizing the non-boundary voxel computations definitely gave another substantial increase on top of what I was able to get in Python alone (maybe a factor of 5 or so? I don't haven exact pre/post Cython number, but it was a pretty big improvement).

Only after all of the above, when `spsolve` got to the point where it accounted for the majority of the computation time, was it worth messing with things like changing the column permutation mode used by `spsolve` itself.

grlee77:
CI failures seem related to the switch to `main` (see https://github.com/scikit-image/scikit-image/pull/5243#issuecomment-781636645)

skimage/restoration/_inpaint.pyx
```diff
@@ -0,0 +1,74 @@
...
+                unknown_idx += 1
+            else:
+                cval = coef_vals[o]
+                for ch in range(nchannels):
+                    data_known[known_idx, ch] -= cval * out_flat[loc, ch]
+                n_known += 1
+        if n_known > 0:
+            row_idx_known[known_idx] = row_idx
+            known_idx += 1
+            # print(f"row_idx={row_idx}, unknown_idx={unknown_idx}, known_idx={known_idx}")
```
emmanuelle, line 72:
```suggestion
```

skimage/restoration/_inpaint.pyx
```diff
@@ -0,0 +1,74 @@
...
+
+    row_idx = row_start
+    known_idx = known_start_idx
+    unknown_idx = unknown_start_idx
+    for i in range(npix):
+        n_known = 0
+        for o in range(num_offsets):
+            loc = center_i[i] + raveled_offsets[o]
+            cval = coef_vals[o]
+            # print(f"    loc={loc}")
```
emmanuelle, line 58:
```suggestion
```

emmanuelle (approved):
Great PR, thank you very much! I read through all the new code and it seems valid to me. That said, our code (and even after these improvements) is hard to understand, in particular the principle of the algorithm is never explained. I had to read the first paper referenced in the docstring to understand its principle, probably the code would be easier to maintain with more explanations about the algorithm. This could of course be left to a future PR, but since you @grlee77 dived into the algorithm, now could also be a good time... But I won't hold this PR since it is such a big improvement in terms of performance. (I just suggested to remove lines that had been commented out)

emmanuelle:
As for the CI, would merging main solve the issue?

grlee77:
> As for the CI, would merging main solve the issue?

I rebased on master just now, so we will see! Thanks for reviewing this difficult PR. I will see about a follow-up adding a few comments about the kernel used and what the sparse matrix represents.

grlee77:
The truncated extent kernels generated for the boundary pixels are equivalent to applying the full kernel extent to the same pixel location with symmetric extension beyond the image boundary.

If we pad the mask width a two pixel wide boundary of zeros and pad the image with a 2-pixel wide boundary using mode='symmetric', then I think we can just remove the separate boundary pixel loop and use only the simpler Cython loop. This should allow simplifying the code substantially, so I will try that.

A good minimal example of the formation of the "reduced" linear system as used in this method is shown in section 4.2 and Fig. 5 of the paper linked below. The only difference in the linear system there is that it is for the Laplacian `[[0, -1, 0], [-1, 4, -1], [0, -1, 0]]` kernel rather than the larger Laplacian of Laplacian one used in the method here. Note also that due to the symmetric boundary conditions and tiny matrix size the value of `4` never appears in the matrix shown there.
https://doi.org/10.1016/j.patcog.2010.08.004
(preprint: http://www.math.uni-sb.de/service/preprints/preprint269.pdf)

emmanuelle:
@grlee77 thank you very much for thinking of a variant which could make the code easier to read. Please tell us if you want to try this as part of this PR or as a follow-up.

grlee77:
> @grlee77 thank you very much for thinking of a variant which could make the code easier to read. Please tell us if you want to try this as part of this PR or as a follow-up.

I tried it for a little while, but couldn't get it to handle the boundaries equivalently without shrinking the kernels as in the current code. Given that, I will leave it as-is for now.

grlee77:
I updated one of the literature reference since it has now been published. 

> our code (and even after these improvements) is hard to understand, in particular the principle of the algorithm is never explained

I think it is a bit difficult to explain. I find the notation in the publications to be confusing at times, but tried to add some comments on the linear system being solved to `_inpaint_biharmonic_single_region`. Do you think it helps?

skimage/restoration/inpaint.py
```diff
@@ -40,7 +40,36 @@ def _get_neigh_coef(shape, center, coef_cache={}):
...
+    """Solve a (sparse) linear system corresponding to biharmonic inpainting.
+
+    This function creates a linear system of the form:
+
+    ``A @ u = b``
+
+    where ``A`` is a sparse matrix, ``b`` is a vector enforcing smoothness and
+    boundary constraints and ``u`` is the vector of inpainted values to be
+    (uniquely) determined by solving the linear system.
+
```
emmanuelle, line 52: Thanks! These explanations really help. Optionally, one additional sentence like:
"The linear system is obtained from a biharmonic equation Mx = 0 where x is the vector of all image pixels, and decomposing M and x into blocks corresponding respectively to known values and pixels to be inpainted."

skimage/restoration/inpaint.py
```diff
@@ -156,6 +185,7 @@ def _inpaint_biharmonic_single_region(image, mask, out, neigh_coef_full,
     # set use_umfpack to False so float32 data is supported
     result = spsolve(matrix_unknown, rhs, use_umfpack=False,
                      permc_spec='MMD_ATA')
+    1 / 0
```
emmanuelle, line 188: I don't understand this addition. `1/0`

grlee77: I did that as a quick way to cause a failure for Ipython debug magic (poor man's breakpoint!) and just forgot to remove it! I was just double-checking that the `matrix_unknown` and `rhs` dimensions matched the description above.

skimage/restoration/inpaint.py
```diff
@@ -126,25 +265,87 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+    channel_stride_bytes = out.strides[-2]
+
+    # offsets to all neighboring non-zero elements in the footprint
+    offsets = tuple(c - radius for c in coef_idx)
+
+    # determine per-channel intensity limits
+    limits = []
+    for ch in range(out.shape[-1]):
+        known_points = image[..., ch][~mask]
+        limits.append((np.min(known_points), np.max(known_points)))
```
rfezzani, line 306: The `limits` computation that I suggested in my previous comment may take place here.

rfezzani (commented):
Thank you @grlee77! That's an impressive amount of work :clap:. Here is my first round of comments, I think that I need some more reading, but it looks very good :wink:

skimage/restoration/_inpaint.pyx
```diff
@@ -0,0 +1,72 @@
...
+ # Can't use "const np_floats[::1]", but instead must declare fused const types
+ # See: https://github.com/cython/cython/issues/1772
+
+cdef fused const_float_1d_c:
+    const float[::1]
+    const double[::1]
+
+cdef fused const_floating_2d_c:
+    const float[:, ::1]
+    const double[:, ::1]
```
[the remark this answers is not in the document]
grlee77: That would work, but I now realize that we may want to just not declare them const to avoid additional unneeded variants of the function from being generated. We really only need two variants: one for float32 and one for float64. However since `np_floats` and `const_floats` each fuse two types, we get C functions generated for all 4 possible combinations. Example:

```C
static Py_ssize_t __pyx_fuse_0_0__pyx_f_7skimage_11restoration_8_inpaint__build_matrix_inner(Py_ssize_t, Py_ssize_t, Py_ssize_t, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, int __pyx_skip_dispatch); /*proto*/
static Py_ssize_t __pyx_fuse_0_1__pyx_f_7skimage_11restoration_8_inpaint__build_matrix_inner(Py_ssize_t, Py_ssize_t, Py_ssize_t, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, int __pyx_skip_dispatch); /*proto*/
static Py_ssize_t __pyx_fuse_1_0__pyx_f_7skimage_11restoration_8_inpaint__build_matrix_inner(Py_ssize_t, Py_ssize_t, Py_ssize_t, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, int __pyx_skip_dispatch); /*proto*/
static Py_ssize_t __pyx_fuse_1_1__pyx_f_7skimage_11restoration_8_inpaint__build_matrix_inner(Py_ssize_t, Py_ssize_t, Py_ssize_t, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, __Pyx_memviewslice, int __pyx_skip_dispatch); /*proto*/
```

If I just used `np_floats` everywhere this would not happen.

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,213 @@
...
+    radius = neigh_coef_full.shape[0] // 2
+
+    edge_mask = np.ones(mask.shape, dtype=bool)
+    edge_mask[(slice(2, -2),) * mask.ndim] = 0
+    boundary_mask = edge_mask * mask
+    center_mask = ~edge_mask * mask
+
+    boundary_pts = np.where(boundary_mask)
+    boundary_i = np.where(boundary_mask.ravel())[0]
+    center_i = np.where(center_mask.ravel())[0]
```
rfezzani, line 84:
```suggestion
    boundary_i = np.flatnonzero(boundary_mask)
    center_i = np.flatnonzero(center_mask)
```

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,213 @@
...
+    designed to avoid discontinuity with the known values.
+    """
 
-    # Iterate over masked points
-    for mask_pt_n, mask_pt_idx in enumerate(mask_pts):
+    n_channels = out.shape[-1]
+    radius = neigh_coef_full.shape[0] // 2
+
+    edge_mask = np.ones(mask.shape, dtype=bool)
+    edge_mask[(slice(2, -2),) * mask.ndim] = 0
```
rfezzani, line 78: Are the hard coded 2s here are intentional or is it supposed to be `radius`?

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,213 @@
...
+    structure = neigh_coef_full != 0
+    tmp = ndi.convolve(mask, structure, output=np.uint8, mode='constant')
+    nnz_matrix = tmp[mask].sum()
+
+    # Need to estimate the number of zeros for the right hand side vector.
+    # The computation below will slightly overestimate the true number of zeros
+    # due to edge effects (the kernel itself gets shrunk in size near the
+    # edges, but that isn't accounted for here). We can trim any excess entries
+    # later.
+    n_mask = mask.sum()
```
rfezzani, line 103: `np.count_nonzero` is much faster then `sum`:

```python
In [1]: x = np.random.rand(500, 500) > 0.5

In [2]: %timeit np.count_nonzero(x)                                                                     
9.96 µs ± 71.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [3]: %timeit x.sum()                                                                                 
196 µs ± 150 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
```suggestion
    n_mask = np.count_nonzero(mask)
```
:wink:

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,213 @@
...
+    # set use_umfpack to False so float32 data is supported
+    result = spsolve(matrix_unknown, rhs, use_umfpack=False,
+                     permc_spec='MMD_ATA')
+    if result.ndim == 1:
+        result = result[:, np.newaxis]
 
-    result = result.ravel()
+    # Handle enormous values on a per-channel basis
+    for ch in range(n_channels):
+        result[..., ch] = np.clip(result[..., ch], *limits[ch])
```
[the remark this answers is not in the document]
grlee77: Thanks! We can also move this clipping to the outer `inpaint_biharmonic` function so it will only be called once instead of on each region individually.

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,213 @@
...
+    tmp = ndi.convolve(mask, structure, output=np.uint8, mode='constant')
+    nnz_matrix = tmp[mask].sum()
+
+    # Need to estimate the number of zeros for the right hand side vector.
+    # The computation below will slightly overestimate the true number of zeros
+    # due to edge effects (the kernel itself gets shrunk in size near the
+    # edges, but that isn't accounted for here). We can trim any excess entries
+    # later.
+    n_mask = mask.sum()
+    nnz_rhs_vector_max = n_mask - (tmp == structure.sum()).sum()
```
rfezzani, line 104: Same here
```suggestion
    n_struct = np.count_nonzero(structure)
    nnz_rhs_vector_max = n_mask - np.count_nonzero(tmp == n_struct)
```

skimage/restoration/inpaint.py
grlee77: Yes, it should be `radius`. (`radius` is always 2 for the biharmonic kernel used in the publication implemented here, though!).

skimage/restoration/inpaint.py
```diff
@@ -126,25 +265,87 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+        mask_labeled *= mask
+
+        # create slices for cropping to each labeled region
+        props = regionprops_table(mask_labeled)
+        ndim = len(img_baseshape)
+
+        bbox_slices = list(
+            zip(*[_get_ax_slices(props, ax, ndim, img_baseshape[ax])
+                  for ax in range(ndim)])
+        )
```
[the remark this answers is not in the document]
grlee77: I refactored this a bit to try to make it more readable in e0d63e7

Perhaps we should consider modifying the `regionprops` API to allow returning slices of this type for the bounding boxes (although not in this PR!)? I am not a heavy user of `regionprops`, though, so perhaps there is a good rational/use case for the way it is currently.

skimage/restoration/inpaint.py
```diff
@@ -126,25 +264,83 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+
+    # offsets to all neighboring non-zero elements in the footprint
+    offsets = tuple(c - radius for c in coef_idx)
+
+    # determine per-channel intensity limits
+    known_points = image[~mask]
+    limits = (known_points.min(axis=0), known_points.max(axis=0))
+
+    if split_into_regions:
+        for idx_region in range(1, num_labels + 1):
```
rfezzani, line 302: The two `if split_into_regions:` blocks can be merged I think:

```suggestion
    # stride for the last spatial dimension
    channel_stride_bytes = out.strides[-2]

    # offsets to all neighboring non-zero elements in the footprint
    offsets = tuple(c - radius for c in coef_idx)

    # determine per-channel intensity limits
    known_points = image[~mask]
    limits = (known_points.min(axis=0), known_points.max(axis=0))

    if split_into_regions:
        # Split inpainting mask into independent regions
        mask = mask.astype(bool, copy=False)
        kernel = ndi.morphology.generate_binary_structure(mask.ndim, 1)
        mask_dilated = ndi.morphology.binary_dilation(mask, structure=kernel)
        mask_labeled, num_labels = label(mask_dilated, return_num=True)
        mask_labeled *= mask

        bbox_slices = _get_boundary_box_slices(mask_labeled, num_labels,
                                               radius)

        for idx_region in range(1, num_labels + 1):
```

skimage/restoration/inpaint.py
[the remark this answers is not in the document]
rfezzani: Can't we directly use `scipy.ndimage.find_objects` here instead of `regionprops`?

skimage/restoration/inpaint.py
grlee77: I got a bit confused by this suggested addition. Is `M` here supposed to be the same as `A` in the docstring? I think the Mx=0 constraint only applies to the inpainted pixels, not "all image pixels"?

skimage/restoration/inpaint.py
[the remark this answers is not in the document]
grlee77: Yes, thanks! I had forgotten about that function. We just have to expand the bounding boxes it returns by `radius`:
```Python

        bbox_slices = ndi.find_objects(mask_labeled)
        # expand object bounding boxes by the kernel radius
        bbox_slices = [
            tuple(slice(max(sl.start - radius, 0),
                        min(sl.stop + radius, size))
                  for sl, size in zip(bb_slice, mask_labeled.shape))
            for bb_slice in bbox_slices
        ]
```

skimage/restoration/inpaint.py
```diff
@@ -126,25 +252,88 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+            for bb_slice in bbox_slices
+        ]
+
+        for idx_region in range(1, num_labels + 1):
+
+            # extract only the region surrounding the label of interest
+            roi_sl = bbox_slices[idx_region - 1]
+            mask_region = mask_labeled[roi_sl] == idx_region
+            # add slice for axes
+            roi_sl = tuple(list(roi_sl) + [slice(None)])
```
rfezzani, line 301:
```suggestion
            roi_sl += (slice(None), )
```

skimage/restoration/inpaint.py
```diff
@@ -126,25 +252,88 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+    neigh_coef_full, coef_idx, coef_vals = _get_neigh_coef(coef_shape,
+                                                           coef_center,
+                                                           coef_cache={},
+                                                           dtype=out.dtype)
+
+    # stride for the last spatial dimension
+    channel_stride_bytes = out.strides[-2]
+
+    # offsets to all neighboring non-zero elements in the footprint
+    offsets = tuple(c - radius for c in coef_idx)
```
rfezzani, line 272: Isn't `coef_idx` a `numpy.array`?
```suggestion
    offsets = coef_idx - radius
```

skimage/restoration/inpaint.py
```diff
@@ -126,25 +252,88 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+            otmp = out[roi_sl].copy()
+
+            # compute raveled offsets for the ROI
+            ostrides = tuple(s // channel_stride_bytes
+                             for s in otmp[..., 0].strides)
+            raveled_offsets = tuple(
+                ax_off * ax_stride
+                for ax_off, ax_stride in zip(offsets, ostrides)
+            )
+            raveled_offsets = functools.reduce(operator.add, raveled_offsets)
```
rfezzani, line 312:
```suggestion
            raveled_offsets = np.multiply(offsets, ostrides).sum()
```

It is more readable and faster on my machine:
```python
In [31]: x = rand(5)                                                                                     

In [32]: y = rand(5)                                                                                     

In [33]: import functools                                                                                

In [34]: import operator                                                                                 

In [35]: functools.reduce(operator.add, tuple(a*b for a, b in zip(x, y)))                                
Out[35]: 0.5815679532148804

In [36]: np.multiply(x, y).sum()                                                                         
Out[36]: 0.5815679532148804

In [37]: %timeit functools.reduce(operator.add, tuple(a*b for a, b in zip(x, y)))                        
3.01 µs ± 32.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [38]: %timeit np.multiply(x, y).sum()                                                                 
2.25 µs ± 17.5 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

skimage/restoration/inpaint.py
```diff
@@ -126,25 +252,88 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+    else:
+        ostrides = tuple(s // channel_stride_bytes
+                         for s in out[..., 0].strides)
+
+        # compute raveled offsets for output image
+        raveled_offsets = tuple(
+            ax_off * ax_stride
+            for ax_off, ax_stride in zip(offsets, ostrides)
+        )
+        raveled_offsets = functools.reduce(operator.add, raveled_offsets)
```
rfezzani, line 329:
```suggestion
        # compute raveled offsets for output image
        raveled_offsets = np.multiply(offsets, ostrides)
```

skimage/restoration/inpaint.py
rfezzani: Thank you for this finding @grlee77, I can confirm how `np.clip` is surprisingly slow, but I found that `np.maximum` do perfectly the job here
```python
In [39]: radius = 2                                                                                      

In [40]: nd_idx = np.asarray([64, 83])                                                                   

In [41]: %timeit tuple(max(p - radius, 0) for p in nd_idx)                                               
2.12 µs ± 13.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [42]: %timeit (nd_idx - radius).clip(min=0)                                                           
10.3 µs ± 59.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [43]: %timeit np.maximum(nd_idx - radius, 0)                                                          
1.26 µs ± 2.77 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,200 @@
...
-from ..measure import label
+from ..measure import label, regionprops_table
+from ._inpaint import _build_matrix_inner
 
 
 def _get_neighborhood(nd_idx, radius, nd_shape):
-    bounds_lo = (nd_idx - radius).clip(min=0)
-    bounds_hi = (nd_idx + radius + 1).clip(max=nd_shape)
+    bounds_lo = tuple(max(p - radius, 0) for p in nd_idx)
+    bounds_hi = tuple(min(p + radius + 1, s) for p, s in zip(nd_idx, nd_shape))
```
rfezzani, line 17:
```suggestion
    bounds_lo = np.maximum(nd_idx - radius, 0)
    bounds_hi = np.minimum(nd_idx + radius + 1, nd_shape)
```

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,200 @@
...
+        # Create (truncated) biharmonic coefficients ndarray
+        coef_shape = tuple(hi - lo for lo, hi in zip(b_lo, b_hi))
+        coef_center = tuple(p - lo for p, lo in zip(mask_pt_idx, b_lo))
+        neigh_coef, coef_idx, coefs = _get_neigh_coef(coef_shape,
+                                                      coef_center,
+                                                      coef_cache,
+                                                      dtype=out.dtype)
+
+        # compute corresponding 1d indices into the mask
+        coef_idx = [c + b for c, b in zip(coef_idx, b_lo)]
```
rfezzani, line 136:
```suggestion
        coef_idx += b_lo
```

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,200 @@
...
+    data_known = np.zeros((nnz_rhs_vector_max, n_channels), dtype=out.dtype)
+    row_idx_unknown = np.empty(nnz_matrix, dtype=np.intp)
+    col_idx_unknown = np.empty(nnz_matrix, dtype=np.intp)
+    data_unknown = np.empty(nnz_matrix, dtype=out.dtype)
+
+    # cache the various small, non-square Laplacians used near the boundary
+    coef_cache = {}
+
+    # Iterate over masked points near the boundary
+    mask_flat = mask.ravel()
```
rfezzani, line 118:
```suggestion
    mask_flat = mask.reshape(-1)
```
See the note section [here](https://numpy.org/doc/stable/reference/generated/numpy.ravel.html)

skimage/restoration/_inpaint.pyx
```diff
@@ -0,0 +1,60 @@
...
+cpdef Py_ssize_t _build_matrix_inner(
+    # starting index offsets
+    Py_ssize_t row_start,
+    Py_ssize_t known_start_idx,
+    Py_ssize_t unknown_start_idx,
+    # input arrays
+    const Py_ssize_t[::1] center_i,
+    const Py_ssize_t[::1] raveled_offsets,
+    np_floats[::1] coef_vals,
+    const short[::1] mask_flat,
```
rfezzani, line 20: why not `char` here?

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,200 @@
...
+
+    # Call an efficient Cython-based implementation for all interior points
+    row_start = mask_pt_n + 1
+    known_start_idx = idx_known
+    unknown_start_idx = idx_unknown
+    nnz_rhs = _build_matrix_inner(
+        # starting indices
+        row_start, known_start_idx, unknown_start_idx,
+        # input arrays
+        center_i, raveled_offsets, coef_vals, mask_flat.astype(np.int16),
```
rfezzani, line 162: We can save some memory here if we use `np.uint8` and `char` instead of `short` in `_build_matrix_inner` signature
```suggestion
        center_i, raveled_offsets, coef_vals, mask_flat.astype(np.uint8),
```

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,200 @@
+import functools
+import operator
```
rfezzani, line 3:
```suggestion
```

If you agree with my proposed changes, these imports are no more needed :wink:

skimage/restoration/inpaint.py
```diff
@@ -126,25 +252,88 @@ def inpaint_biharmonic(image, mask, multichannel=False):
     image = skimage.img_as_float(image)
     mask = mask.astype(bool)
```
rfezzani, line 253:
```suggestion
    mask = mask.astype(bool, copy=False)
```
We may avoid a copy here :wink:

skimage/restoration/inpaint.py
```diff
@@ -126,25 +252,88 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+    # offsets to all neighboring non-zero elements in the footprint
+    offsets = tuple(c - radius for c in coef_idx)
+
+    # determine per-channel intensity limits
+    known_points = image[~mask]
+    limits = (known_points.min(axis=0), known_points.max(axis=0))
+
+    if split_into_regions:
+        # Split inpainting mask into independent regions
+        mask = mask.astype(bool, copy=False)
```
rfezzani, line 280:
```suggestion
```
I see that this conversion is already made in the beginning of the function.

skimage/restoration/_inpaint.pyx
```diff
@@ -0,0 +1,60 @@
...
+        for o in range(num_offsets):
+            loc = center_i[i] + raveled_offsets[o]
+            cval = coef_vals[o]
+            if mask_flat[loc]:
+                data_unknown[unknown_idx] = cval
+                row_idx_unknown[unknown_idx] = row_idx
+                col_idx_unknown[unknown_idx] = loc
+                unknown_idx += 1
+            else:
+                cval = coef_vals[o]
```
rfezzani, line 52:
```suggestion
        Py_ssize_t ch
        Py_ssize_t row_idx, known_idx, unknown_idx, loc, n_known
        np_floats cval
        int nchannels = data_known.shape[1]

    row_idx = row_start
    known_idx = known_start_idx
    unknown_idx = unknown_start_idx

    for c_i in center_i:
        n_known = 0
        for offset, cval in zip(raveled_offsets, coef_vals):
            loc = c_i + offsets
            if mask_flat[loc]:
                data_unknown[unknown_idx] = cval
                row_idx_unknown[unknown_idx] = row_idx
                col_idx_unknown[unknown_idx] = loc
                unknown_idx += 1
            else:
```

skimage/restoration/inpaint.py
```diff
@@ -126,25 +252,88 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
-            mask_region = mask_labeled == idx_region
-            _inpaint_biharmonic_single_channel(mask_region,
-                out[..., idx_channel], limits)
+    # Create biharmonic coefficients ndarray
+    radius = 2
+    coef_shape = (2 * radius + 1,) * mask.ndim
+    coef_center = (radius,) * mask.ndim
+    neigh_coef_full, coef_idx, coef_vals = _get_neigh_coef(coef_shape,
+                                                           coef_center,
+                                                           coef_cache={},
```
rfezzani, line 265:
```suggestion
```

skimage/restoration/_inpaint.pyx
grlee77, line 20:
```suggestion
    const char[::1] mask_flat,
```

skimage/restoration/inpaint.py
grlee77: Most likely I should just not use an empty dict in the function signature for `_get_neigh_coef` and default to `coef_cache=None` instead). Otherwise the dict persists across function calls and passing an empty dict like this was needed to avoid side effects!

skimage/restoration/inpaint.py
grlee77: actually, I had to revert some of these changes since coef_idx is a tuple of arrays as returned by `numpy.where`

skimage/restoration/_inpaint.pyx
```diff
@@ -0,0 +1,57 @@
...
+        int nchannels = data_known.shape[1]
+
+    row_idx = row_start
+    known_idx = known_start_idx
+    unknown_idx = unknown_start_idx
+
+    for c_i in center_i:
+        n_known = 0
+        for offset, cval in zip(raveled_offsets, coef_vals):
+            loc = c_i + offset
```
grlee77, line 43: This ends up being less efficient than the prior code, probably due to use of zip so I have reverted part of the Cython changes.

I did keep the mask as `char` change, though!

skimage/restoration/inpaint.py
rfezzani: You are right, simply not defining a default value for `coef_cache` may do the trick...

skimage/restoration/inpaint.py
rfezzani: Of course! You are right, I did not noticed that :confused:....
Can't we try to exploit the power of `numpy` by converting `coef_idx` to `np.array`?
```python
coef_idx = np.array(np.where(...))
```
Moreover, these tuple manipulations make the code more difficult to follow :confused:

grlee77: Yes, I have now gone back and used arrays for clarity (see 3edd0ab). We need tuples for the kernel cache dictionary keys, but otherwise arrays are fine.

grlee77: One subtle aspect is that I could not follow your suggestion to do inplace
```Python
coef_idx += b_lo[:, np.newaxis]
```
Because this ends up modifying the value for `coef_idx` in the `coef_cache` dictionary itself, so that the next time that same kernel is encountered called the offsets will be wrong!

skimage/restoration/inpaint.py
```diff
@@ -124,27 +254,75 @@ def inpaint_biharmonic(image, mask, multichannel=False):
...
+
+    # determine per-channel intensity limits
+    known_points = image[~mask]
+    limits = (known_points.min(axis=0), known_points.max(axis=0))
+
+    if split_into_regions:
+        # Split inpainting mask into independent regions
+        kernel = ndi.generate_binary_structure(mask.ndim, 1)
+        mask_dilated = ndi.binary_dilation(mask, structure=kernel)
+        mask_labeled, num_labels = label(mask_dilated, return_num=True)
```
rfezzani, line 285: `num_labels` is no more used :tada: 
```suggestion
        mask_labeled = label(mask_dilated)
```

grlee77: Good catch. I also forgot to remove the `functools` and `operator` imports. will update it now

grlee77:
Hi @soupault, we have two reviewers participating now, but let me know if you want to also review prior to merge (I can't remember if I had requested your review or if you had marked it because you planned to review?).

Thanks to both @emmanuelle and @rfezzani for the helpful feedback.

skimage/restoration/inpaint.py
```diff
@@ -1,79 +1,204 @@
...
-        neigh_coef = np.zeros(b_hi - b_lo)
-        neigh_coef[tuple(mask_pt_idx - b_lo)] = 1
-        neigh_coef = laplace(laplace(neigh_coef))
+        # Create (truncated) biharmonic coefficients ndarray
+        coef_shape = tuple(b_hi - b_lo)
+        coef_center = tuple(nd_idx - b_lo)
+        neigh_coef, coef_idx, coefs = _get_neigh_coef(coef_shape,
+                                                      coef_center,
+                                                      coef_cache,
+                                                      dtype=out.dtype)
```
rfezzani, line 134: What about removing the cache management from `_get_neigh_coef`?
I see here that the `neigh_coef` variable is not used in this function, so what about something like this:
```suggestion
        coef_idx, coefs = coef_cache.get((coef_shape, coef_center), (None, None)
        if coef_idx is None:
            _ , coef_idx, coefs = _get_neigh_coef(coef_shape,
                                                  coef_center
                                                  dtype=out.dtype)
            coef_cache[(coef_shape, coef_center)] = (coef_idx, coefs)
```
Doing so, the cache mechanism is explicit where it is necessary and the `_get_neigh_coef` is simplified... What do you think @grlee77?

grlee77: Yeah, that looks good.

rfezzani (approved):
Nice! Thank you @grlee77 for your reactivity. I see nothing more to say :smile:. Let's wait for @soupault feedback now :wink:

emmanuelle:
@soupault can we merge the PR? We have two approvals already.

alexdesiqueira:
Thank you @grlee77! I'll go out on a limb here and merge, since we have two approvals and we're starting to have conflicts already.
Let us know if you would like something different @soupault; we can discuss that in a new PR.

PR #5241, opened 2021-02-18
run Haar application gallery example in single-threaded mode

grlee77:
## Description

I have noticed occasional timeouts on Azure when running the gallery examples on CI. When testing locally, `plot_haar_extraction_selection_classification.py` alone accounts for the majority of the time.

On my system with 10 cores, single-threaded operation is nearly an order of magnitude faster than running multi-threaded. This is probably due to relatively low computational complexity of the features leading to memory access times dominating the performance.

If the run time on Azure when checking the gallery examples is substantially reduced, we can consider merging this PR as is. If not, then another option to approximately halve the example time would be to reduce the number of images used for training from 200 to 100.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_haar_extraction_selection_classification.py
+++ b/doc/examples/applications/plot_haar_extraction_selection_classification.py
@@ -73,7 +73,7 @@ feature_types = ['type-2-x', 'type-2-y']
 X = delayed(extract_feature_image(img, feature_types) for img in images)
 # Compute the result
 t_start = time()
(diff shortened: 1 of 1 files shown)


grlee77:
I am going to close this because it does not seem to be the cause of the timeouts. Looking back through some other successful logs, this test seems to take < 2 minutes. When timeouts occur it seems to be more due to excessive time in installing the packages into the environment.

PR #5243, opened 2021-02-18
Rename `master` to `main` throughout

stefanv:


grlee77:
Thanks, @stefanv

It looks like there is still some reference to master in the gitwash docs:

````
Warning, treated as error:
/home/runner/work/scikit-image/scikit-image/doc/source/gitwash/development_workflow.rst:39:Unknown target name: "deleting master on github".
```

stefanv:
It looks like the redirect on the GitHub side is not followed by Pooch, so we may have to push a dummy master branch up with the data in it.

grlee77:
> It looks like the redirect on the GitHub side is not followed by Pooch, so we may have to push a dummy master branch up with the data in it.

Didn't the line you added setting `version_dev='main'` for pooch fix this? It seems to fetch things fine from main now when I try it locally. Is your main concern that currently open PRs would require a rebase after this is merged? (I did see failure to fetch in #5240 after pushing a change there, but that PR does not have the fix used here).

stefanv:
> Didn't the line you added setting `version_dev='main'` for pooch fix this? It seems to fetch things fine from main now when I try it locally. Is your main concern that currently open PRs would require a rebase after this is merged? (I did see failure to fetch in #5240 after pushing a change there, but that PR does not have the fix used here).

My concern was that existing installations of skimage would try and fetch the files from the old location and fail.

grlee77:
> My concern was that existing installations of skimage would try and fetch the files from the old location and fail.

I think this should only happy for users with a development installation, but not for tagged versions. 

I just activated a conda environment with v0.17.2 installed and ran

```Python
from skimage._shared.testing import fetch                               
import numpy as np                                                      

ref_data = dict(np.load(fetch("data/rank_filter_tests.npz")))   
```

The output indicates that it gets downloaded from the release branch, which still works okay:
```Downloading file 'data/rank_filter_tests.npz' from  https://github.com/scikit-image/scikit-image/raw/v0.17.x/skimage/data/rank_filter_tests.npz'```

grlee77 (approved):
It looks good to me and CI is passing now

stefanv:
Excellent, thanks for investigating @grlee77!

jni:
:tada:

hmaarrfk:

Carreau:
> @meeseeksdev backport to v0.18.x

Can you retry ?

PR #5245, opened 2021-02-20
updated regionprops_table test for iterable properties

yfukai:
## Description

Associated with pull request #5239 and issue #5222 . I updated test `test_regionprops_table_equal_to_original` so that they compare not only non-iterable properties but also iterable ones of `regionprops` and `regionprops_table`.

## Checklist

→ not applicable
→ not applicable
  existing benchmark
→ not applicable
- Unit tests
→ one modified here
→ checked with autopep8

## For reviewers

  later.

### DIFF
--- a/skimage/measure/tests/test_regionprops.py
+++ b/skimage/measure/tests/test_regionprops.py
@@ -1,5 +1,4 @@
 import math
-from collections.abc import Iterable
 
 import numpy as np
 from numpy import array
@@ -580,12 +579,20 @@ def test_regionprops_table():
 def test_regionprops_table_equal_to_original():
     regions = regionprops(SAMPLE, INTENSITY_FLOAT_SAMPLE)
     out_table = regionprops_table(SAMPLE, INTENSITY_FLOAT_SAMPLE,
-                                  properties=COL_DTYPES.keys())
+                                  properties=COL_DTYPES.keys(),
+                                  separator="-")
 
-    for prop in COL_DTYPES.keys():
-        for i in range(len(regions)):
-            if not isinstance(regions[i][prop], Iterable):
-                assert regions[i][prop] == out_table[prop][i]
+    for prop, dtype in COL_DTYPES.items():
+        for i, reg in enumerate(regions):
+            rp = reg[prop]
+            if np.isscalar(rp) or prop in OBJECT_COLUMNS or dtype is np.object_:
+                assert_array_equal(rp, out_table[prop][i])
+            else:
+                shape = rp.shape if isinstance(rp, np.ndarray) else (len(rp),)
+                for ind in np.ndindex(shape):
+                    modified_prop = "-".join(map(str, (prop,) + ind))
+                    loc = ind if len(ind) > 1 else ind[0]
+                    assert_equal(rp[loc], out_table[modified_prop][i])
 
 
 def test_regionprops_table_no_regions():
--- a/skimage/measure/tests/test_regionprops.py
+++ b/skimage/measure/tests/test_regionprops.py
@@ -585,7 +585,9 @@ def test_regionprops_table_equal_to_original():
     for prop, dtype in COL_DTYPES.items():
         for i, reg in enumerate(regions):
             rp = reg[prop]
-            if np.isscalar(rp) or prop in OBJECT_COLUMNS or dtype is np.object_:
+            if np.isscalar(rp) or \
+                    prop in OBJECT_COLUMNS or \
+                    dtype is np.object_:
                 assert_array_equal(rp, out_table[prop][i])
             else:
                 shape = rp.shape if isinstance(rp, np.ndarray) else (len(rp),)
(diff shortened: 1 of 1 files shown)


mkcor (commented):
Excellent, thank you! With the breakdown introduced by `separator="-"`, I'm not seeing any instance where `i > 0`... Please correct me if I'm missing something!
It seems to me that all properties are covered by this main for loop:
```py
    for prop, dtype in COL_DTYPES.items():
        for reg in regions:
            rp = reg[prop]
            if np.isscalar(rp) or \
                    prop in OBJECT_COLUMNS or \
                    dtype is np.object_:
                assert_array_equal(rp, out_table[prop][0])
            else:
                shape = rp.shape if isinstance(rp, np.ndarray) else (len(rp),)
                for ind in np.ndindex(shape):
                    modified_prop = "-".join(map(str, (prop,) + ind))
                    loc = ind if len(ind) > 1 else ind[0]
                    assert_equal(rp[loc], out_table[modified_prop][0])
```

yfukai:
Hi, thank you for your comment @mkcor ! Yeah you're right, `SAMPLE` has only one region with the label `1`, so `i>0` doesn't appear in the loop for now. However, I have to note that I think the modification you suggested would result in an error when someone add label `2` to `SAMPLE`. Instead we may be able to fix the index to `0` as follows to make the code a bit simpler:
```python
reg=regions[0]
for prop, dtype in COL_DTYPES.items():
    rp = reg[prop]
    if np.isscalar(rp) or \
            prop in OBJECT_COLUMNS or \
            dtype is np.object_:
        assert_array_equal(rp, out_table[prop][0])
    else:
        shape = rp.shape if isinstance(rp, np.ndarray) else (len(rp),)
        for ind in np.ndindex(shape):
            modified_prop = "-".join(map(str, (prop,) + ind)) 
            loc = ind if len(ind) > 1 else ind[0]
            assert_equal(rp[loc], out_table[modified_prop][0])
```

mkcor:
@yfukai ok, very well. This way your test is robust to the case where `SAMPLE` would have more than one region. Fair enough! Then I'm happy to approve the code as is.

mkcor (approved):

yfukai:
Wonderful, thanks!

skimage/measure/tests/test_regionprops.py
```diff
@@ -580,12 +579,22 @@ def test_regionprops_table():
...
     regions = regionprops(SAMPLE, INTENSITY_FLOAT_SAMPLE)
     out_table = regionprops_table(SAMPLE, INTENSITY_FLOAT_SAMPLE,
-                                  properties=COL_DTYPES.keys())
-
-    for prop in COL_DTYPES.keys():
-        for i in range(len(regions)):
-            if not isinstance(regions[i][prop], Iterable):
-                assert regions[i][prop] == out_table[prop][i]
+                                  properties=COL_DTYPES.keys(),
+                                  separator="-")
```
rfezzani, line 583:
```suggestion
                                  properties=COL_DTYPES.keys())
```

isn't `"-"` the default value for the `separator` argument?

skimage/measure/tests/test_regionprops.py
```diff
@@ -580,12 +579,22 @@ def test_regionprops_table():
...
+
+    for prop, dtype in COL_DTYPES.items():
+        for i, reg in enumerate(regions):
+            rp = reg[prop]
+            if np.isscalar(rp) or \
+                    prop in OBJECT_COLUMNS or \
+                    dtype is np.object_:
```
rfezzani, line 590:
```suggestion
            if np.isscalar(rp) or prop in OBJECT_COLUMNS:
```
The `dtype is np.object_` condition is not necessary.

rfezzani (commented):
Thank you @yfukai! The PR is really good. I just left two small suggestions :wink:

skimage/measure/tests/test_regionprops.py
yfukai: Oh thanks, I see! This is actually (dirty) copy from the _regionprops.py logic itself. I'll try to rewrite it in a bit more DRY way.

skimage/measure/tests/test_regionprops.py
yfukai: Exactly I noticed it is, thank you for pointing this out!

jni (approved):

jni:
Awesome, thank you again @yfukai!

yfukai:
Thanks for merging @jni , but I think I still haven't dealt with @rfezzani 's second comment. Maybe I will make some new PR for that at some point!

jni:
Oops! :sweat_smile: I saw the comment and the commit and made an assumption. :grimacing: Future PRs are absolutely welcome! I think this is already a major improvement anyway. =)

rfezzani:
:smile: no worry, it is not a big deal! The PR is already in really good shape, but as @jni said, future PRs are welcome! Thank you again @yfukai :wink:

PR #5247, opened 2021-02-23
Added support for binary images in is_low_contrast to fix imshow of binary images

MadsDyrmann:
## Description

imshow stopped working when plotting boolean images, since np.percentile no longer support boolean arrays. This PR makes sure that is_low_contrast does not fail when the input is a boolean array.

This is a small test-script that would fail before this change:

```
import numpy as np
from skimage.io import imshow

im = np.ones((10,10), dtype=np.bool)

imshow(im)
```

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -647,6 +647,10 @@ def is_low_contrast(image, fraction_threshold=0.05, lower_percentile=1,
         if image.shape[2] == 3:
             image = rgb2gray(image)
 
+    if image.dtype==np.bool:
+        return image.any() and (image==False).any()
+
+
     dlimits = dtype_limits(image, clip_negative=False)
     limits = np.percentile(image, [lower_percentile, upper_percentile])
     ratio = (limits[1] - limits[0]) / (dlimits[1] - dlimits[0])
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -647,7 +647,7 @@ def is_low_contrast(image, fraction_threshold=0.05, lower_percentile=1,
         if image.shape[2] == 3:
             image = rgb2gray(image)
 
-    if image.dtype == np.bool:
+    if image.dtype == bool:
         return image.any() and (image == False).any()
 
 
--- a/skimage/exposure/exposure.py
+++ b/skimage/exposure/exposure.py
@@ -647,8 +647,8 @@ def is_low_contrast(image, fraction_threshold=0.05, lower_percentile=1,
         if image.shape[2] == 3:
             image = rgb2gray(image)
 
-    if image.dtype==np.bool:
-        return image.any() and (image==False).any()
+    if image.dtype == np.bool:
+        return image.any() and (image == False).any()
 
 
     dlimits = dtype_limits(image, clip_negative=False)


jni:
Thanks @MadsDyrmann!

Very happy to merge this fix. Do you want to add a test to make sure it doesn't break in the future?

Having said all this, in the medium term (~1y) we aim to pull out the matplotlib dependency and make sure the focus of scikit-image is on image processing, not display or IO. So you should probably start migrating your downstream code to start using plt.imshow directly.

mkcor:
> So you should probably start migrating your downstream code to start using `plt.imshow` directly.

May I recommend [plotly.express.imshow](https://plotly.com/python/imshow/) as an alternative? With `plt.imshow` I ran into the [issue](https://github.com/matplotlib/matplotlib/issues/18627) of being unable to specify value ranges for multichannel data. See this [gallery example](https://scikit-image.org/docs/dev/auto_examples/applications/plot_3d_structure_tensor.html) for basic `px.imshow` usage.

jni:
@mkcor well, plotly plots it in a web browser, which some people are allergic to—*cough* *cough* *cough*. :stuck_out_tongue_winking_eye:

mkcor:
@jni just an alternative... :stuck_out_tongue_winking_eye:

skimage/exposure/exposure.py
```diff
@@ -647,6 +647,10 @@ def is_low_contrast(image, fraction_threshold=0.05, lower_percentile=1,
         if image.shape[2] == 3:
             image = rgb2gray(image)
 
+    if image.dtype == bool:
+        return image.any() and (image == False).any()
+
```
emmanuelle, line 652: Maybe I'm missing something, but if a boolean image has both True and False values, then the function should return False, shouldn't it? I have the impression that the returned value should be
```suggestion
return not(image.any() and (image == False).any())
```

emmanuelle: anyway a test (as @jni suggested) could either prove me that I'm wrong :-) or check that the behaviour is as expected

rfezzani: When `image.dtype == bool`, `is_low_contrast` is `True` when the image is constant: all filled with `True` or with `False`, _ie_ when `False` and `True` are not both in `image`.

So I think that you are right @emmanuelle!

BTW, I think that this test must take place at the beginning of the function, even before checking for image dimensions, because `rgb2gray` will convert the image to floating point...

emmanuelle:
@MadsDyrmann are the different suggestions clear for you? Thanks again for the PR!

mkcor:
Hi @MadsDyrmann,

To add a test, edit file `skimage/exposure/tests/test_exposure.py`: Line 749, right after test function `test_is_low_contrast()`, you can define another test, say,  `test_is_low_contrast_bool_image()` and use the example you came up with, `im = np.ones((10,10), dtype=np.bool)`.

Hope that helps!

alexdesiqueira:
Hi everyone,
@MadsDyrmann, would you like to continue working on this PR? We could help you finishing it, if you'd prefer.
Thank you very much for your help so far!

grlee77:
@MadsDyrmann, I have resumed this PR in #5467 so am closing this one as outdated.

Please confirm your name and email address either here or in that PR if you want to be attributed as a co-author on the commits.

Issue #4821, opened 2020-07-03
regionprops: rename `equivalent_diameter` -> `equivalent_diameter_area`


maxfrei750:
There exists actually a whole variety of equivalent diameters. Therefore, IMHO, it would be nice to have a more specific name (e.g.  `equivalent_diameter_area`).

jni:
Thanks @maxfrei750 !

Could you enumerate a few of the options? My impression is that this will not bring a big enough benefit to justify the API break unless we also provide others. 

However, when we release 1.0 we will be changing quite a few things and this might be a good candidate for that update. I’ll tag it for that.

maxfrei750:
Admittedly, the distinction is more important for the 3d objects:
* surface area equivalent diameter
* projection area equivalent diameter
* volume equivalent diameter

than for 2d objects:

* area equivalent diameter 
* perimeter equivalent diameter

mkcor:
@maxfrei750 so, shouldn't the renaming be `equivalent_diameter` --> `area_equivalent_diameter`?

maxfrei750:
@mkcor We had a similar situation for `feret_diameter_max` vs. `max_feret_diameter` (see https://github.com/scikit-image/scikit-image/pull/4820). We ended up using the former, since then all Feret diameters are grouped together. I would make the same argument for `equivalent_diameter_area`. The "area" is more like  an index.

mkcor:
Fixed by #5254

PR #5254, opened 2021-03-01
Introduce a consistent naming scheme for measure.regionprops

### LINKED ISSUE #5213: regionprops: consistent naming scheme
## Description
It would be nice to have a consistent naming scheme for the properties retrieved by the `measure.regionprops` function.

For instance, there are `feret_diameter_max` and `max_intensity`. It would be nice to have `intensity_max` instead. 

Other candidates are:
```
mean_intensity
min_intensity
bbox_area
convex_area
convex_image
filled_area
filled_image
intensity_image
major_axis_length
max_intensity
mean_intensity
min_intensity
minor_axis_length
weighted_centroid
weighted_local_centroid
weighted_moments
weighted_moments_central
weighted_moments_hu
weighted_moments_normalized
```

Since these changes would not be backwards compatible, they could be targeted for version 1.0.

Of course, it would also be a possibility, to rename `feret_diameter_max` to `max_feret_diameter`. However, this would prevent similar properties (i.e. all the different kinds of areas, intensities, moments and images) to be grouped together based on their name (see the discussion in https://github.com/scikit-image/scikit-image/pull/4820).

maxfrei750:
## Description

Closes #5213.

## For reviewers

  later.


emmanuelle:
@maxfrei750 thanks for the PR! Is the PR a proof of concept that should be merged for 1.0 or do you intend to have it merged ASAP? In the latter case we should keep the old names as well for compatibility (and just return the new prop of course).

maxfrei750:
@emmanuelle In #5213 @jni proposed to merge this in version 1.0, so I'd keep it that way, if you agree.

maxfrei750:
@emmanuelle Does this PR need any changes to be merged in 1.0?

grlee77:
The code here looks good to me and I think the naming scheme is better than before. I think the question is whether we want to provide a transition period in 0.19 where the old names still work, but warn users that the names will change in 1.0. @scikit-image/core what do you think?

If we do just switch the names in 1.0, we should add a summary table somewhere in the release notes to help users with the transition. Most likely we can write a short guide to the main API changes in 1.0 outside of this PR.

grlee77:
This was a related naming change discussed in #4821. While we are already changing things here, wouldn't it make sense to do this at the same time?

maxfrei750:
@grlee77 Good point. If the renaming proposed in #4821 is wanted, then I can add this here. So the question is: Is it wanted?

VolkerH:
> 
> If we do just switch the names in 1.0, we should add a summary table somewhere in the release notes to help users with the transition. Most likely we can write a short guide to the main API changes in 1.0 outside of this PR.

Maybe a utililty function that can translate old names to new names would be useful for transitioning old code.

stefanv:
> > If we do just switch the names in 1.0, we should add a summary table somewhere in the release notes to help users with the transition. Most likely we can write a short guide to the main API changes in 1.0 outside of this PR.
> 
> Maybe a utililty function that can translate old names to new names would be useful for transitioning old code.

It may be easiest to just leave the old names in place, but hide them (or not advertise them).

maxfrei750:
@grlee77 Thanks for your contribution and especially for catching my mistake with the `PROPS` dictionary. That definitively could have caused some mayhem  for people using older versions. :sweat_smile:

skimage/measure/_regionprops.py
```diff
@@ -16,36 +16,52 @@
...
-    'ConvexImage': 'convex_image',
+    'ConvexImage': 'image_convex',
+    'convex_image': 'image_convex',
     'Coordinates': 'coords',
     'Eccentricity': 'eccentricity',
     'EquivDiameter': 'equivalent_diameter',
     'EulerNumber': 'euler_number',
     'Extent': 'extent',
     # 'Extrema',
     'FeretDiameterMax': 'feret_diameter_max',
```
maxfrei750, line 41: The legacy name of `FeretDiameterMax` was `FeretDiameter`. `FeretDiameterMax` already introduced in https://github.com/scikit-image/scikit-image/pull/4820 and I think it breaks backwards compatibility. Now would be a good time to fix it.

grlee77: Thanks, I have now made this change

grlee77:
I see that #4821 also suggested `equivalent_diameter` -> `equivalent_diameter_area`. If we are going to make that change (seems reasonable to me), then let's do it here as well.

maxfrei750:
@grlee77 Good point! Thanks for pointing that out. Feel free to incorporate this change, or I will do it myself, as soon as I find the time.

grlee77:
> @grlee77 Good point! Thanks for pointing that out. Feel free to incorporate this change, or I will do it myself, as soon as I find the time.

I will go ahead and add it now

maxfrei750:
@grlee77 Thanks!

grlee77 (approved):

grlee77:
Note that when [recently discussing various API-related things](https://github.com/scikit-image/meeting-notes/blob/main/2021/july-api-meetings.md), there was a point brought up that "max intensity", etc. read as more clearly than "intensity max", etc. While this is true, I don't personally think that outweighs the benefit of the more coherent organization proposed here (see also discussion in #5213)

@scikit-image/core, please speak out in the next week or so if there are still any strong objections to making this change. If we are going to do it, let's get it in for 0.19.

We had also discussed whether or not to drop the undocumented legacy property names at some point, but that is more suited to a separate PR if we decide to go that route.

stefanv:
I agree that the more logical/hierarchical organization is worth the slight reduction in "intuitive" readability here.  I.e., `intensity_max` better than `max_intensity`.

I am -0 on removing the legacy names—not a big deal.

mkcor:
Hi @grlee77,

Thanks for the reminder through https://github.com/scikit-image/scikit-image/issues/5169#issuecomment-895241439. I don't have strong objections either way. So I'm happy to follow @stefanv on this one.

skimage/measure/_regionprops.py
```diff
@@ -397,36 +421,36 @@ def intensity_image(self):
         )
         return self._intensity_image[self.slice] * image
 
-    def _intensity_image_double(self):
-        return self.intensity_image.astype(np.double)
+    def _image_intensity_double(self):
```
mkcor, line 424: Since this is an internal function, I guess the renaming wasn't necessary... Fine, though.

PR #5257, opened 2021-03-02
Improve salt and pepper noise performance

andrewnags:
## Description

While using scikit-image in a highly parallel environment, I realized that the salt-and-pepper mode of random noise generation was consuming a large amount of memory, and taking longer than expected.  I traced the issue to calls of the form `np.random.choice([True, False], ...)` for deciding whether each pixel will be flipped, and if so whether to make it "salt" or "pepper".  

While this procedure works, it seems that `np.random.choice` is aimed more at cases where the user needs to decide among a greater number of classes, and so consumes much more memory (and takes longer) than required.  In the 2-class case, checking whether a real number drawn from a `Unif[0, 1]` distribution is less than the desired probability `p` suffices, and is quicker and consumes less memory.

For a few, small images the improvement may be negligible, but for large images, or many images in parallel, the improvement is quite significant.

At the cost of (some) speed, the memory usage could be further reduced by decomposing the desired shape into smaller chunks and processing those sequentially (because `float`s consume more memory than `bool`s), but this also makes the code messier, so I chose not to go this route.  However if the trade-off is deemed worth it, I can make this modification.

On a single 5000x5000 image, peak memory consumption decreased by ~25%, and the time decreased by over 50%.

~~As a side-effect, the old method did not take into account the random seed provided by the user, whereas the new method does.~~ (I didn't realize that numpy was seeded globally.  My mistake!)

I realized through the tests a downside to this method: because it relies on the statistics of drawing from a probability distribution, the proportions (of pixels flipped, or of salt vs. pepper) will have greater variance, whereas it seems `np.random.choice` tries to match the desired proportions as exactly as possible.  I still think this is a worthwhile trade-off, as for larger samples (and most images tend to be larger than the test cases) the proportions should converge to the desired values, and for the purposes of image processing I don't think the *exact* proportions are as important (although I don't have that much experience, so I'm happy to be told otherwise).  Furthermore since the cases of `p == 0` and `p == 1` are handled separately, those proportions will be exact.

## Checklist

  - Done.
  - No new features.
  existing benchmark
  - Added `./benchmarks/benchmark_noise.py`.
- Unit tests
  - Covered by existing unit tests for random noise.  However, I did have to increase their tolerance somewhat to account for the more probabilistic nature of the method.
  - Passes `flake8` without errors or warnings.

## For reviewers

  later.


skimage/util/noise.py
```diff
@@ -5,6 +5,36 @@
 __all__ = ['random_noise']
 
 
+def choose(p, shape):
+    """
+    Function to return Bernoulli trials at a given probability
```
sciunto, line 10: You can write "Bernoulli trials at a given probability of a given size." on a single line

skimage/util/noise.py
```diff
@@ -5,6 +5,36 @@
 __all__ = ['random_noise']
 
 
+def choose(p, shape):
```
sciunto, line 8:
```suggestion
def _bernoulli(p, shape):
```
I would rather see this function as a private one.

skimage/util/noise.py
```diff
@@ -172,10 +202,8 @@ def random_noise(image, mode='gaussian', seed=None, clip=True, **kwargs):
         out = image.copy()
         p = kwargs['amount']
         q = kwargs['salt_vs_pepper']
-        flipped = np.random.choice([True, False], size=image.shape,
-                                   p=[p, 1 - p])
-        salted = np.random.choice([True, False], size=image.shape,
-                                  p=[q, 1 - q])
+        flipped = choose(p, image.shape)
```
sciunto, line 205:
```suggestion
        flipped = _bernoulli(p, image.shape)
```

skimage/util/noise.py
```diff
@@ -172,10 +202,8 @@ def random_noise(image, mode='gaussian', seed=None, clip=True, **kwargs):
...
+        salted = choose(q, image.shape)
```
sciunto, line 206:
```suggestion
        salted = _bernoulli(q, image.shape)
```

sciunto (commented):
minor comments, nice PR. Thank you

sciunto (approved):

andrewnags:
Thank you!

skimage/util/noise.py
```diff
@@ -5,6 +5,35 @@
 __all__ = ['random_noise']
 
 
+def _bernoulli(p, shape):
+    """
+    Bernoulli trails at a given probability of a given size.
```
grlee77, line 10:
```suggestion
    Bernoulli trials at a given probability of a given size.
```

benchmarks/benchmark_noise.py
```diff
@@ -0,0 +1,26 @@
+# See "Writing benchmarks" in the asv docs for more information.
```
grlee77, line 1: possibly rename this file to `benchmark_util.py` since most other benchmark files are currently named based on the module they correspond to.

skimage/util/noise.py
```diff
@@ -5,6 +5,35 @@
...
+    Returns
+    -------
+    out : ndarray[bool]
+        The results of Bernoulli trials in the given `size` where success
+        occurs with probability `p`.
+    """
+    if p == 0:
+        return np.zeros(shape, dtype=bool)
+    if p == 1:
+        return ~np.zeros(shape, dtype=bool)
```
grlee77, line 33:
```suggestion
        return np.ones(shape, dtype=bool)
```

skimage/util/tests/test_random_noise.py
```diff
@@ -24,7 +24,7 @@ def test_salt():
 
     # Ensure approximately correct amount of noise was added
     proportion = float(saltmask.sum()) / (cam.shape[0] * cam.shape[1])
-    assert 0.11 < proportion <= 0.15
+    assert 0.11 < proportion <= 0.16
```
grlee77, line 27: I wonder why such a low lower bound was used here and elsewhere? I realize it was this way prior to this PR, but it just seems odd that the range is so asymmetric around the requested amount.

An alternative could be to use a variable for `amount` and then set the tolerance as some fraction of that. 
```Python
assert abs(amount - proportion) < 0.05 * amount
```

grlee77: I tried it in a couple of test cases and it even passes for `0.01 * amount`

grlee77 (dismissed):
Thanks, @andrewnags. I have some minor comments, but overall this looks great.

skimage/util/tests/test_random_noise.py
grlee77: Or maybe just a simple fixed 0.01 rather than as a fraction of amount.

andrewnags: I'm not sure why the intervals were setup that way to begin with, but I symmetrized the tolerance used for the salt and pepper noise tests.  I believe the tests failed on some machines when I set the tolerance too low, but I've set it to `1e-2` for now, except for the test I believe failed on some machines at that tolerance (`5e-2`).

jni (approved):

jni:
Thank you @andrewnags, this is a very nice contribution. I agree with all your arguments above. :blush:

PR #5262, opened 2021-03-09
Add textbook-like tutorial on measuring fluorescence at nuclear envelope.

mkcor:
## Description

This draft PR fits the overall goal of offering more gallery examples for the life sciences (see #4601).

Here I'm following this amazing suggestion by @haesleinhuepf: https://forum.image.sc/t/looking-for-life-scientists-to-collaborate-on-scikit-image-tutorials/49073/15

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


jni:
:wave: @mkcor this is very exciting. =) I don't think for this tutorial it's needed because the nuclei are not close together (?), but I just want to give a shout-out to @VolkerH's [`segmentation.expand_labels`](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_expand_labels.html#sphx-glr-auto-examples-segmentation-plot-expand-labels-py) which is like dilation but smarter. =)

mkcor:
> @mkcor this is very exciting. =) I don't think for this tutorial it's needed because the nuclei are not close together (?), but I just want to give a shout-out to @VolkerH's [`segmentation.expand_labels`](https://scikit-image.org/docs/stable/auto_examples/segmentation/plot_expand_labels.html#sphx-glr-auto-examples-segmentation-plot-expand-labels-py) which is like dilation but smarter. =)

Thank you for the great suggestion, @jni! I decided to go for @VolkerH's `segmentation.expand_labels` even though it's not *necessary* with dataset. Actually (reassuringly, maybe), these two results are equal:

```py
expand = segmentation.expand_labels(label)
dilate = morphology.dilation(label)
```

(where `label` is the labelled image after blurring, thresholding, filling in holes, and keeping only the one nucleus of interest) and yield this difference, when subtracting `morphology.erosion(label)`:
difference_1px

The rim is then about 2px-thick (I zoomed in with Plotly to check). But I tried reproducing the [original processing](https://media.springernature.com/original/springer-static/image/chp%3A10.1007%2F978-3-030-22386-1_2/MediaObjects/467591_1_En_2_Fig3_HTML.png) faithfully, and it's straightforward with `segmentation.expand_labels`'s keyword argument:
 ```py
expand = segmentation.expand_labels(label, distance=4)
```
where I can just expand more, as much as I like! :smile:
The resulting rim is about 5px-thick.

File: doc/examples/applications/plot_fluorescence_nuclear_envelope.py [hunk 1]
```diff
@@ -0,0 +1,214 @@
...
+
+ch0t0 = dat[0, 0, :, :]
+
+#####################################################################
+# Segment the nucleus rim
+# =======================
+
+smooth = filters.gaussian(ch0t0, sigma=2)
+
+thresh = smooth > 0.1
```
haesleinhuepf, line 73: In [the original workflow](https://github.com/NEUBIAS/neubias-springer-book-2020/blob/master/Ch2%20Measurements%20of%20Intensity%20Dynamics%20at%20the%20Periphery%20of%20the%20Nucleus/Code/code_final.ijm#L30) Otsu's threshold method was used. Is there a reason why we're using a fixed constant threshold here?

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,214 @@
...
+
+fill = ndi.binary_fill_holes(thresh)
```
haesleinhuepf, line 75: Also this step is not in [the original workflow](https://github.com/NEUBIAS/neubias-springer-book-2020/blob/master/Ch2%20Measurements%20of%20Intensity%20Dynamics%20at%20the%20Periphery%20of%20the%20Nucleus/Code/code_final.ijm#L33). I could imagine that it's not necessary if one uses Otsu's thresholding method...

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,214 @@
...
+
+#####################################################################
+# Although it is not strictly necessary, let us discard the other nucleus part
+# visible in the bottom right-hand corner.
```
haesleinhuepf, line 79: In ImageJ macro, we're [using the function of the Particle Analyzer](https://github.com/NEUBIAS/neubias-springer-book-2020/blob/master/Ch2%20Measurements%20of%20Intensity%20Dynamics%20at%20the%20Periphery%20of%20the%20Nucleus/Code/code_final.ijm#L33) for removing objects that touch the image border. How about using [clear_border](https://scikit-image.org/docs/stable/api/skimage.segmentation.html#skimage.segmentation.clear_border)? I could imagine that the script becomes then more applicable also to other images and use-cases.

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,214 @@
...
+
+label = measure.label(fill)
+regions = measure.regionprops_table(label, properties=('label', 'area'))
+label_small = regions['label'][np.argmin(regions['area'])]
+label[label == label_small] = 0  # background value
+
+expand = segmentation.expand_labels(label, distance=4)
```
haesleinhuepf, line 86: In the original workflow, [we erode the mask](https://github.com/NEUBIAS/neubias-springer-book-2020/blob/master/Ch2%20Measurements%20of%20Intensity%20Dynamics%20at%20the%20Periphery%20of%20the%20Nucleus/Code/code_final.ijm#L43) and [dilate it](https://github.com/NEUBIAS/neubias-springer-book-2020/blob/master/Ch2%20Measurements%20of%20Intensity%20Dynamics%20at%20the%20Periphery%20of%20the%20Nucleus/Code/code_final.ijm#L41), which means in ImageJ to take one pixel as radius.

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,214 @@
...
+#####################################################################
+# To begin with, let us consider the first channel of the first image.
+
+ch0t0 = dat[0, 0, :, :]
+
+#####################################################################
+# Segment the nucleus rim
+# =======================
+
+smooth = filters.gaussian(ch0t0, sigma=2)
```
haesleinhuepf, line 71: In the [original workflow sigma=1.5](https://github.com/NEUBIAS/neubias-springer-book-2020/blob/master/Ch2%20Measurements%20of%20Intensity%20Dynamics%20at%20the%20Periphery%20of%20the%20Nucleus/Code/code_final.ijm#L28)

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,214 @@
...
+    mask = get_mask(dat[i, 0, :, :])
+    props = measure.regionprops_table(
+        mask,
+        intensity_image=dat[i, 1, :, :],
+        properties=('label', 'area', 'intensity_mean')
+    )
+    assert props['label'] == 1
+    intensity_total = props['area'] * props['intensity_mean']
+    fluorescence_change.append(intensity_total)
+
```
haesleinhuepf, line 199: You could mention here that the following part is not in the original workflow. But as we are in python, we can do plotting and image analysis together in one script. :-)

haesleinhuepf (commented):
Hi @mkcor ,

super cool to see this evolving, thanks for working on this! As I translated the workflow at some point as well, I was curious and took a look at your implementation. I found some differences compared to the original implementation. I don't want to be picky; I was just wondering if you made those changes intentionally and if so, why. :-) Also did you compare quantitative measurements with the measurements generated with your script? You find quantitative measurements from the original workflow in ImageJ for example [here](https://github.com/NEUBIAS/neubias-springer-book-2021/blob/master/Ch05_GPU-accelerating_ImageJ_Macro_image_processing_workflows_using_CLIJ/benchmarking_perlis/original_analyis_0.xls) and from an [alternative implementation here](https://github.com/NEUBIAS/neubias-springer-book-2021/blob/master/Ch05_GPU-accelerating_ImageJ_Macro_image_processing_workflows_using_CLIJ/benchmarking_perlis/clij_analyis_0.xls).

Also one more minor comment: I personally prefer variable names such as "input_image" over "dat". Also "channel_0_image" could be more self-explanatory than "ch0t0". But maybe I'm just too used to writing `longVariableNamesInJava` ;-)

Great work otherwise! Many biologists will appreciate this example in scikit-image and the detailed explanation of the used commands.

Thanks again!

Best,
Robert

mkcor:
@haesleinhuepf @jni please review: https://2809-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/applications/plot_fluorescence_nuclear_envelope.html

The CI failures come from
```py
dat = imageio.volread('http://cmci.embl.de/sampleimages/NPCsingleNucleus.tif')
```
where `urllib.error.URLError: <urlopen error timed out>` -- I'll try and reach out to NeuBIAS to see if we could include this dataset in scikit-image's data registry.

mkcor:
Hi @haesleinhuepf,

Wow! You've been lightning fast and our comments have crossed! Thank you so much for your review. Please *be* picky! :wink: 

> Also one more minor comment: I personally prefer variable names such as "input_image" over "dat".

I hear you. In this case, would you agree with either `image_stack` or `image_sequence`?

haesleinhuepf:
> `image_sequence`

This sounds great! :-)

mkcor:
> Also "channel_0_image" could be more self-explanatory than "ch0t0".

@haesleinhuepf how about `t_0_channel_0_img`?

haesleinhuepf:
> @haesleinhuepf how about `t_0_channel_0_img`?

Sure. Sounds good :-)

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: Thanks for pointing me to the ImageJ Macro source! I admit I didn't look it up; I followed the [book chapter](https://link.springer.com/chapter/10.1007%2F978-3-030-22386-1_2) mostly qualitatively, but let's try and get as close as possible quantitatively...

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: Hmm, the [conclusion](https://link.springer.com/chapter/10.1007%2F978-3-030-22386-1_2#Fig8) reads "Here, to summarize the analysis in this chapter, we plot the changes in the total fluorescence intensity over time using ImageJ Macro `code_plotResults.ijm` (◘ Fig. 2.8). The code appears after the paragraph below."

So, I get that it's not part of *the* 'original workflow' but that's not because ImageJ would be limiting; is it?

Alternatively, I could change my introduction to make it explicit that I'm following this book chapter in particular, which itself is based on the original workflow.

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: Good point, it's much better to infer the threshold:
otsu

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: Right, well, this was always a bit of a mystery to me... Looking at [steps c) and d)](https://link.springer.com/chapter/10.1007%2F978-3-030-22386-1_2#Fig3), I've wondered how the small holes were handled: Does dilation automatically take care of filling them?

I don't see what difference using Otsu's thresholding method would make (see above https://github.com/scikit-image/scikit-image/pull/5262#discussion_r713413342)... Even with multi-Otsu thresholding (defaulting to 3 classes), and keeping the lower threshold, there are still small holes inside the main region:
multi_otsu

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: Great point! Your suggestion is much more readable and much more elegant than what I had; done 3c99e28b6e.

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: Ok, fixed in 02c282357d. Does the 5ish-px-looking thickness in the final processing [step](https://link.springer.com/chapter/10.1007%2F978-3-030-22386-1_2#Fig3) only come from the rendering/printing then?

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: Fixed in 02c282357d.

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: Fixed in 02c282357.

> "Reassuringly, we find the expected result: The total fluorescence intensity at the nuclear envelope increases 1.3-fold in the initial five time points, and then becomes roughly constant."

Now that I'm more consistent with the original workflow, the overall result is preserved but, comparing with the [textbook plot](https://link.springer.com/chapter/10.1007%2F978-3-030-22386-1_2#Fig8), we can see sharper dips (below 1.2) for time points 7 and 12:
original_wkfl

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,217 @@
...
+#####################################################################
+# can be recovered from:
+
+props['area'] * props['intensity_mean']
+
+#####################################################################
+# Iterate the measurement for each time point
+# ===========================================
+# Let us write a function for the first processing step (segmentation of
+# nucleus rim).
```
jni, line 170: Personally, and this is a personal preference and happy to be overridden, I would make a 3D volume where each timepoint has a different segment value. We can then use `regionprops_table` directly with `label` and `intensity_mean` and plot them directly.

To do this "cleanly" we'd probably need to use a "flat" structuring element and only use dilate/erode instead of expand_labels:

```
footprint = np.stack((np.zeros((3, 3)), np.ones((3, 3)), np.zeros((3, 3)))
```

I'm not sure it's better, I just like avoiding the for-loop computing a single regionprop each time. 😂

jni (approved):
awesome @mkcor! The plotly integration with the docs was just *made* for this example! 😃 

I've made a potentially terrible suggestion, take it or leave it. 😜

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: @jni @grlee77 please see my nD porting at 5002298d002d (there's just Otsu's threshold which I couldn't use in nD).

mkcor:
Dear @FirefoxMetzger and @almarklein,

Help! :sweat_smile: 

Up until imageio v2.10.1, the dataset I'm using would be read as expected by `volread`:
imageio_2 10 1

But since imageio v2.10.3, the first two dimensions got stacked:
imageio_2 10 3

Was this change in behaviour intended? Thank you!

mkcor:
Dear reviewers,

Please discard commits 99bc77a, 3c3031a, and c467b50. I was trying to fix the following CI failures:

* warning treated as error which appeared with f4d6d49, corresponding to imageio's version bumping to 2.10.1,
* broken docs as of 6403382, corresponding to imageio's version bumping to 2.10.3.

The latest passing CI was a6dbe91, corresponding to imageio v2.9.0 (which is the same version I'm running locally in my virtualenv).

Note that:
https://github.com/scikit-image/scikit-image/blob/c073f0e90f33851426bc0b10408279b41a780d95/requirements/default.txt#L5

The actual update is 5002298, where the read-only code block is for plotting this 3D interactive view of the 15-frame stack, making it clear that the border-touching object we want to clear is located at the edge with greatest (x, y) values:
interactive_3d_fill_seq

FirefoxMetzger:
> Dear @FirefoxMetzger and @almarklein,
> 
> Help! 😅
> 
> Up until imageio v2.10.1, the dataset I'm using would be read as expected by `volread`: imageio_2 10 1
> 
> But since imageio v2.10.3, the first two dimensions got stacked: imageio_2 10 3
> 
> Was this change in behaviour intended? Thank you!

@mkcor Could you share the above lines as text so that I could copy them and see what happens for me locally?

My first thought here was that the old shape of `(15, 2, 180, 183)` for the `volread` call looks odd, but I also don't know the layout of the TIFF so the best I can do is guess. Are these slices of channel-first images, aka. `(stack, channel, height, width)`? If it is not channel-first data, is it gray-scale? In this case, I would expect `mvolread` to be more appropriate, because the data would be something like `(time, stack, height, width)`... This is again me speculating though.

It could be regression on something that isn't covered by our test suite, but I would have to investigate it more and actually see the file. In this case, I would probably want to track this as a dedicated issue over at imageio, so that we can create a PR against it, etc.

mkcor:
Hi @FirefoxMetzger,

For sure!
```py
image_sequence = imageio.volread('http://cmci.embl.de/sampleimages/NPCsingleNucleus.tif')
print(f'shape: {image_sequence.shape}')
```
The code is from this very PR (tab 'Files changed') and, for more context, you can view a rendered version of the tutorial [here](https://2937-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/applications/plot_fluorescence_nuclear_envelope.html). The dataset is `[time, channel, row, column]` and `volread` was doing a great job at reading it until v2.9.0 included; I wasn't aware of `mvolread` so I'll look it to see if it's more suitable!

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
grlee77: > there's just Otsu's threshold which I couldn't use in nD

More specifically, you mean that you want a different threshold per timepoint as opposed to a global one over all time points? The function is nD so you could get a single threshold in one call, but you need to call it like you have it now if you want a unique threshold per timepoint.

grlee77: You could also reshape the data to 2D where timepoints is the first axis and then call using NumPy's apply_along_axis. It may be clearer to just leave it as you have it now, though?

```Python
np.apply_along_axis(threshold_otsu, 1, smooth_seq.reshape((a.shape[0], -1)))
```

FirefoxMetzger:
@mkcor Thanks for sharing. I looked into the file and it is indeed a regression (or a bug in v2.9.0's imread).

What is the copyright on this image? If possible, I'd like to use it for a regression test in ImageIO, because we currently don't seem to have a test TIFF in this particular shape (channel-first stored in different planes), which is why this made it through CI.

I opened https://github.com/imageio/imageio/issues/681 over at ImageIO to track this and to avoid polluting this PR with the discussion :)

mkcor:
Hi @FirefoxMetzger,

Thanks for looking into this issue so promptly.

> What is the copyright on this image?

I'm currently communicating with the copyright holder about possibly licensing this image dataset under CC0: In principle, he's open to it. :smiley: 

I'm not sure I can tag him here... @andreaboni? Andrea Boni is @boniandr on Twitter. I'll send him a follow-up email right now!

> If possible, I'd like to use it for a regression test in ImageIO, because we currently don't seem to have a test TIFF in this particular shape (volume of channel-first images where each channel is stored on a different plane), which is why this made it through CI.

That would be great. I'll keep you posted.

> I opened [imageio/imageio#681](https://github.com/imageio/imageio/issues/681) over at ImageIO to track this and to avoid polluting this PR with the discussion :)

Thank you. I'll see you there!

> I forgot to respond to this x). `mvolread` is for multi-volume-read so it isn't what you'd want to use here. Also, it may not be worthwhile habituating to using it, since we will soon depreciate all of the above in favor of a simple `iio.imread` which will return a ndimage. So, in the future, you should be able to do
> 
> ```python
> img = iio.v3.imread('http://cmci.embl.de/sampleimages/NPCsingleNucleus.tif')
> img.shape # (15, 2, 180, 183)
> ```

Ok, good to know. Thanks for sharing!

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
mkcor: > > there's just Otsu's threshold which I couldn't use in nD
> 
> More specifically, you mean that you want a different threshold per timepoint as opposed to a global one over all time points? The function is nD so you could get a single threshold in one call, but you need to call it like you have it now if you want a unique threshold per timepoint.

Yes, exactly. Sorry, I should have been more specific in my assertion. I just double-checked the original workflow again and, to be faithful to it, we want to compute a threshold for each timepoint.

mkcor: > You could also reshape the data to 2D where timepoints is the first axis and then call using NumPy's apply_along_axis. It may be clearer to just leave it as you have it now, though?
> 
> ```python
> np.apply_along_axis(threshold_otsu, 1, smooth_seq.reshape((smooth_seq.shape[0], -1)))
> ```

I like it! I have added it for pedagogical purposes: f3bb06658d4a5.

FirefoxMetzger:
Just to leave an update here, the ImageIO bug will be fixed in https://github.com/imageio/imageio/pull/682 , which is ready to merge as soon as we have approval to use the image in our CI suite (and prevent regression) :)

mkcor:
Following the rolling out of the [fix](https://github.com/imageio/imageio/pull/682) (thank you, @FirefoxMetzger!), I have reverted b82d4615e4e, and all tests are passing! :tada: 

Should I pin imageio's version to exclude `2.10.1` and `2.10.3` though? I think @hmaarrfk would be the best person to ask... :innocent:

FirefoxMetzger:
> Should I pin imageio's version to exclude 2.10.1 and 2.10.3 though? I think @hmaarrfk would be the best person to ask... 😇

If TIFF usage is _really_ important then that might make sense. In this case, I would exclude 2.10.1 - 2.10.4 , because all these versions suffer the regression fixed in 2.10.5. Overall, it may make more sense to just ask people to upgrade to the latest version if they run into this problem, because that is what they _should_ do anyway. 

ImageIO doesn't maintain multiple versions. All fixes are applied to `master`, and there is no back-porting. Our versioning and releasing is fully automated (currently runs nightly if there are changes), and you (as a human) shouldn't associate any meaning to the version number (beyond semver). The version number is literally just some number that machines should use when asked to set up a historic version of ImageIO, e.g., because you want to run a script from 4 years ago and we have since rolled out a new version major with deprecations.

mkcor:
Once this PR https://gitlab.com/scikit-image/data/-/merge_requests/15 is merged, I'll add the image dataset to our data registry, and update this PR to fetch the dataset from there.

grlee77:
> Once this PR https://gitlab.com/scikit-image/data/-/merge_requests/15 is merged, I'll add the image dataset to our data registry, and update this PR to fetch the dataset from there.

Great, I just merged both of your pending gitlab PRs.

mkcor:
This is ready for final review, thanks!

/cc @hmaarrfk :wink:

mkcor:
@grlee77 I've added the `0.19.2` milestone so this new tutorial ships along with #6087 but, actually, I'm not exactly sure this kind of addition belongs in a patch release... :thinking:

grlee77:
> @grlee77 I've added the 0.19.2 milestone so this new tutorial ships along with #6087 but, actually, I'm not exactly sure this kind of addition belongs in a patch release... thinking

Okay, I will review this soon. I think it should be fine to add in 0.19.2 since it is just supplementing docs and not changing behavior of existing functions.

rfezzani (commented):
Thank you @mkcor, I added a small comment concerning the step labels in the first figure's titles, suggested to tighten figure's layout and proposed few minor code style modifications.

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,287 @@
...
+ax[1, 0].set_title('c-2) Keep one nucleus')
+
+ax[1, 1].imshow(dilate, cmap=plt.cm.gray)
+ax[1, 1].set_title('d) Dilate')
+
+ax[1, 2].imshow(erode, cmap=plt.cm.gray)
+ax[1, 2].set_title('e) Erode')
+
+ax[1, 3].imshow(mask, cmap=plt.cm.gray)
+ax[1, 3].set_title('f) Nucleus Rim')
```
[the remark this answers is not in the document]
mkcor: You're absolutely right, @rfezzani! Done in 5eaf48c43. Thank you for your review :pray:

grlee77 (approved):
This looks pretty nice @mkcor. I suggested some minor efficiency improvements that hopefully don't detract from readability. I will go ahead and approve as they are suggestions and don't change the final result.

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,294 @@
...
+dilate = morphology.binary_dilation(clear)
+
+erode = morphology.binary_erosion(clear)
+
+#####################################################################
+# Finally, we subtract the eroded from the dilated to get the nucleus rim
+#(step ``f)``).
+
+mask = dilate.astype(int) - erode.astype(int)
+mask.dtype
```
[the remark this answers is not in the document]
mkcor: I think the `~` operator is standard enough not to be confusing, but I would keep/add the description about, morally, subtracting the eroded from the dilated.

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,294 @@
...
+#     thresh_values = np.apply_along_axis(filters.threshold_otsu,
+#                                         axis=1,
+#                                         arr=smooth_seq.reshape(n_z, -1))
+#
+# We use the following flat structuring element for morphological
+# computations:
+
+footprint = np.stack((np.zeros((3, 3)),
+                      ndi.generate_binary_structure(2, 1),
+                      np.zeros((3, 3))))
```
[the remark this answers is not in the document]
mkcor: Good to know, thanks!

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,294 @@
...
+dilate_seq = morphology.binary_dilation(clear_seq, footprint=footprint)
+erode_seq = morphology.binary_erosion(clear_seq, footprint=footprint)
+mask_seq = dilate_seq.astype(int) - erode_seq.astype(int)
+
+#####################################################################
+# Let us give each mask (corresponding to each time point) a different label,
+# running from 1 to 15.
+
+mask_list = [mask * lab for lab, mask in enumerate(mask_seq, 1)]
+mask_sequence = np.stack(mask_list)
```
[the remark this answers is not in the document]
mkcor: I think that the inline comment "add one for the background" is misleading, because we don't label the background *in addition* to the other regions...(?)  We just wrote that we wanted labels running from 1 to 15, so it's just that we need to call `np.arange(1, 16)` (and not `np.arange(15)`, which would give values ranging from 0 to 14).

doc/examples/applications/plot_fluorescence_nuclear_envelope.py
```diff
@@ -0,0 +1,294 @@
...
+fig.tight_layout()
+
+#####################################################################
+# Measure the total intensity
+# ===========================
+# The mean intensity is readily available as a region property in a labeled
+# image.
+
+props = measure.regionprops_table(
+    mask,
```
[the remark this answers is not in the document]
mkcor: Because we 'know' that `np.uint8` is the dtype with smallest size and smallest scalar kind which can hold 0's and 1's?

mkcor:
The CI failure is unrelated to this PR. It comes from another gallery example, namely the one on [image stitching](https://scikit-image.org/docs/stable/auto_examples/registration/plot_stitching.html):
```bash
Warning, treated as error:
/home/runner/work/scikit-image/scikit-image/doc/examples/registration/plot_stitching.py failed to execute correctly: Traceback (most recent call last):
[...]
  File "/home/runner/work/scikit-image/scikit-image/doc/examples/registration/plot_stitching.py", line 110, in <listcomp>
    trfm_list = [measure.ransac((dst, src),
AttributeError: 'NoneType' object has no attribute 'params'
```
where the full code is:
```py
trfm_list = [measure.ransac((dst, src),
                            transform.EuclideanTransform, min_samples=3,
                            residual_threshold=2, max_trials=100)[0].params
             for dst in matching_corners]
```
But it's weird that the list of `measure.ransac` calls would return `None` just for Python 3.9 (3.8 and 3.10 being fine).

/cc @rfezzani

grlee77:
> But it's weird that the list of measure.ransac calls would return None just for Python 3.9 (3.8 and 3.10 being fine).

Let me repeat the failing case and see if it resolves itself. I'm pretty sure I've seen that same failure happen at least once before. Maybe we need to fix a random seed or something?

grlee77 (approved):
LGTM, thanks!

grlee77:
@meeseeksdev backport to v0.19.x

PR #5263, opened 2021-03-09
Add ellipsoid_coords function

ksugar:
## Description

This PR adds the `draw3d.ellipsoid_coords` function that is analogous to the `draw.ellipse` function.
This function returns the voxel coordinates of an ellipsoid, while the existing function `draw3d.ellipsoid` returns the 3d ndarray that renders an ellipsoid.
This function accepts the rotations in each axis (i.e. `rot_x`, `rot_y` and `rot_z`).
It also supports the anisotropic resolution by introducing the `spacing` parameter.

A variation of this function is implemented and used in the following preprint.

> Tracking cell lineages in 3D by incremental deep learning. Ko Sugawara, Cagri Cevrim, Michalis Averof
> bioRxiv 2021.02.26.432552; doi: https://doi.org/10.1101/2021.02.26.432552

Original implementation used in the preprint:
https://github.com/elephant-track/elephant-server/blob/v0.1.0/elephant-core/elephant/util/ellipsoid.py

## Checklist

I am not sure if I should add the example.
  existing benchmark
not applicable
- [x] Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/draw/draw3d.py
+++ b/skimage/draw/draw3d.py
@@ -1,6 +1,6 @@
 import numpy as np
 from scipy.special import (ellipkinc as ellip_F, ellipeinc as ellip_E)
-from skimage.transform._geometric import _euler_rotation
+from scipy.spatial.transform import Rotation
 
 
 def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
@@ -159,8 +159,9 @@ def ellipsoid_coords(center, axis_lengths, *, shape=None, angles=None,
             f'len(axis_lengths) should be 3 but got {len(axis_lengths)}')
     axis_lengths = np.array(axis_lengths)
 
-    R = np.eye(3)
-    if angles is not None:
+    if angles is None:
+        R = np.eye(3)
+    else:
         if len(angles) != 3:
             raise ValueError(
                 'len(angles) should be 3 but got',
@@ -172,13 +173,14 @@ def ellipsoid_coords(center, axis_lengths, *, shape=None, angles=None,
             or axes[0] == axes[1]
                 or axes[1] == axes[2]):
             raise ValueError(f'axes: {axes} is invalid')
-        if intrinsic:
-            for i, angle in zip(axes, angles):
-                R = _euler_rotation(i, angle) @ R
-            R = R.T
-        else:
-            for i, angle in zip(axes, angles):
-                R = _euler_rotation(i, angle).T @ R
+        axes_str = 'ZYX' if intrinsic else 'zyx'
+        seq = ''
+        for axis in axes:
+            seq += axes_str[axis]
+
+        # Generate a rotation matrix. The order of the elements needs to be
+        # reversed so that it follows the (z, y, x) order.
+        R = Rotation.from_euler(seq, angles).as_matrix()[::-1, ::-1]
 
     if spacing is None:
         spacing = np.ones(3)
--- a/skimage/draw/draw3d.py
+++ b/skimage/draw/draw3d.py
@@ -63,27 +63,253 @@ def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
     return arr
 
 
-def ellipsoid_coords(center, axis_lengths, *, shape=None,
-                     rotation=(0., 0., 0.), spacing=(1., 1., 1.)):
-    """Generate coordinates of voxels within ellipsoid.
+def _rotate_axes_x(angle):
+    r"""Rotate the axes of a coordinate system about the x-axis.
+
+    .. math::
+
+        R(\theta) = \begin{bmatrix}
+                        \cos\theta & -\sin\theta & 0 \\
+                        \sin\theta &  \cos\theta & 0 \\
+                                 0 &           0 & 1 \\
+                    \end{bmatrix}
 
     Parameters
     ----------
-    d, r, c : float
-        Center coordinate of ellipsoid. The values will be divided by the
-        spacing parameter inside the function.
-    d_radius, r_radius, c_radius : float
-        Radii of ellipsoid. The values will be divided by the spacing
-        parameter inside the function.
-    shape : tuple, optional
+    angle : float
+        Angle to rotate the axes about the x-axis.
+
+    Returns
+    -------
+    rotation_matrix : (3, 3) ndarray of float
+        Rotation matrix to rotate the axes about the x-axis by `angle`.
+        The order of dimension is supposed to be ``(z, y, x)``.
+    """
+
+    sin, cos = np.sin(angle), np.cos(angle)
+    return np.array([[cos, -sin, 0],
+                     [sin, cos, 0],
+                     [0, 0, 1]])
+
+
+def _rotate_axes_y(angle):
+    r"""Rotate the axes of a coordinate system about the y-axis.
+
+    .. math::
+
+        R(\theta) = \begin{bmatrix}
+                         \cos\theta & 0 & \sin\theta \\
+                                  0 & 1 &          0 \\
+                        -\sin\theta & 0 & \cos\theta \\
+                    \end{bmatrix}
+
+    Parameters
+    ----------
+    angle : float
+        Angle to rotate the axes about the y-axis.
+
+    Returns
+    -------
+    rotation_matrix : (3, 3) ndarray of float
+        Rotation matrix to rotate the axes about the y-axis by ``angle``.
+        The order of dimension is supposed to be ``(z, y, x)``.
+    """
+
+    sin, cos = np.sin(angle), np.cos(angle)
+    return np.array([[cos, 0, sin],
+                     [0, 1, 0],
+                     [-sin, 0, cos]])
+
+
+def _rotate_axes_z(angle):
+    r"""Rotate the axes of a coordinate system about the z-axis.
+
+    .. math::
+
+        R(\theta) = \begin{bmatrix}
+                        1 &          0 &           0 \\
+                        0 & \cos\theta & -\sin\theta \\
+                        0 & \sin\theta &  \cos\theta \\
+                    \end{bmatrix}
+
+    Parameters
+    ----------
+    angle : float
+        Angle to rotate the axes about the z-axis.
+
+    Returns
+    -------
+    rotation_matrix : (3, 3) ndarray of float
+        Rotation matrix to rotate the axes about the z-axis by ``angle``.
+        The order of dimension is supposed to be ``(z, y, x)``.
+    """
+
+    sin, cos = np.sin(angle), np.cos(angle)
+    return np.array([[1, 0, 0],
+                     [0, cos, -sin],
+                     [0, sin, cos]])
+
+
+def _rotate_axes(about, angle):
+    """Rotate the axes of a coordinate system.
+
+    Parameters
+    ----------
+    about : {'x', 'y', 'z'},
+        Axis to rotate about.
+
+    angle : float
+        Angle to rotate the axes.
+
+    Returns
+    -------
+    rotation_matrix : (3, 3) ndarray of float
+        Rotation matrix to rotate the axes about the z-axis by ``angle``.
+        The order of dimension is supposed to be ``(z, y, x)``.
+    """
+
+    if about == 'x':
+        return _rotate_axes_x(angle)
+    elif about == 'y':
+        return _rotate_axes_y(angle)
+    elif about == 'z':
+        return _rotate_axes_z(angle)
+    else:
+        raise ValueError(f'about={about} is not supported.')
+
+
+def _angles_to_rotmat(angles, order='zxz', is_intrinsic=True):
+    r"""Calculate a rotation matrix from the Euler angles.
+
+    The Euler angles are defined based on ``order`` and ``is_intrinsic``.
+    The following example shows the ``zxz`` intrinsic rotations.
+
+    .. math::
+
+        R(\phi) = \begin{bmatrix}
+                      1 &        0 &         0 \\
+                      0 & \cos\phi & -\sin\phi \\
+                      0 & \sin\phi &  \cos\phi \\
+                  \end{bmatrix}
+
+        R(\theta) = \begin{bmatrix}
+                        \cos\theta & -\sin\theta & 0 \\
+                        \sin\theta &  \cos\theta & 0 \\
+                                 0 &           0 & 1 \\
+                    \end{bmatrix}
+
+        R(\psi) = \begin{bmatrix}
+                      1 &         0 &         0 \\
+                      0 &  \cos\psi & -\sin\psi \\
+                      0 &  \sin\psi &  \cos\psi \\
+                  \end{bmatrix}
+
+        \\
+
+        R(\phi, \theta, \psi) = (R(\psi)R(\theta)R(\phi))^\intercal
+
+    Parameters
+    ----------
+    angles : (3,) ndarray of float
+        Euler angles ``(phi, theta, psi)`` in this order.
+
+    order : str, length 3, optional (default 'zxz')
+        Order of rotations. Each character should be one of ``{'x', 'y', 'z'}``
+        and there should be no adjacent repeating characters.
+
+    is_intrinsic : bool optional (default True)
+        Specify if the rotations are intrinsic (``True``) or extrinsic
+        (``False``).
+
+        - *intrinsic*: rotations occur about the rotating coordinate system
+        - *extrinsic*: rotations occur about the original coordinate system
+
+    Returns
+    -------
+    rotation_matrix : (3, 3) ndarray of float
+        Rotation matrix calculated from the Euler angles.
+    """
+
+    if (not isinstance(order, str) or len(order) != 3 or order.strip('xyz')
+            or order[0] == order[1] or order[1] == order[2]):
+        raise ValueError(f'order: {order} is invalid')
+    angles %= np.pi
+    phi, theta, psi = angles
+    rotmat_phi = _rotate_axes(order[0], phi)
+    rotmat_theta = _rotate_axes(order[1], theta)
+    rotmat_psi = _rotate_axes(order[2], psi)
+    if is_intrinsic:
+        rotmat = (rotmat_psi @ rotmat_theta @ rotmat_phi).T
+    else:
+        rotmat = rotmat_psi.T @ rotmat_theta.T @ rotmat_phi.T
+    return rotmat
+
+
+def _is_valid_rotation_matrix(matrix):
+    r"""Determine if a rotation matrix is valid.
+
+    A rotation matrix is an orthogonal matrix whose determinant equals 1.
+
+    .. math::
+
+        RR^\intercal = I
+        \\
+        \det R = 1
+
+    Parameters
+    ----------
+    matrix : (3, 3) ndarray of float
+        Matrix to be evaluated.
+
+    Returns
+    -------
+    is_valid : bool
+        ``True`` if a matrix is a rotation matrix, ``False`` otherwise.
+    """
+
+    return (np.allclose(matrix @ matrix.T, np.identity(3)) and
+            np.allclose(np.linalg.det(matrix), 1))
+
+
+def ellipsoid_coords(center, axis_lengths, *, shape=None, rotation_angles=None,
+                     rotation_order='zxz', is_intrinsic=True,
+                     rotation_matrix=None, spacing=None):
+    r"""Generate coordinates of voxels within ellipsoid.
+
+    Parameters
+    ----------
+    center : (3,) array-like of float
+        Center coordinate of ellipsoid. The order is (z, y, x). The values will
+        be divided by ``spacing`` inside the function.
+    axis_lengths : (3,) array-like of float
+        Axis lengths of ellipsoid. The order is (z, y, x). The values will be
+        divided by ``spacing`` inside the function.
+    shape : tuple of int, length 3, optional
         Image shape which is used to determine the maximum extent of output
         pixel coordinates. This is useful for ellipsoids which exceed the
-        image size.
-        By default the full extent of the ellipsoid is used.
-    rot_z, rot_y, rot_x : float, optional (default 0.)
-        Rotation for each axis.
-    spacing : (3,) array-like of float, optional (default (1., 1., 1.))
-        Spacing in (z, y, x) spatial dimensions in this order.
+        image size. By default the full extent of the ellipsoid is used.
+    rotation_angles : (3,) array-like of float, optional (default None)
+        Rotation angles to rotate the ellipsoid. The order of rotations is
+        specified by ``rotation_order``. Ignored if ``rotation_matrix`` is
+        specified. No rotation will be applied if both ``rotation_angles``
+        and ``rotation_matrix`` are ``None``.
+    rotation_order : str, length 3, optional (default 'zxz')
+        Order of rotations. Each character should be one of ``{'x', 'y', 'z'}``
+        and there should be no adjacent repeating characters.
+    is_intrinsic : bool optional (default True)
+        Specify if the rotations are intrinsic (``True``) or extrinsic
+        (``False``).
+
+        - *intrinsic*: rotations occur about the rotating coordinate system
+        - *extrinsic*: rotations occur about the original coordinate system
+    rotation_matrix : (3, 3) array-like of float, optional (default None)
+        Rotation matrix to rotate the ellipsoid. If both ``rotation_angles``
+        and ``rotation_matrix`` are specified, ``rotation_matrix`` has priority
+        over ``rotation_angles``. No rotation will be applied if both
+        ``rotation_angles`` and ``rotation_matrix`` are ``None``.
+    spacing : (3,) array-like of float, optional (default None)
+        Spacing in each spatial dimension. The order is (z, y, x). If
+        ``spacing`` is ``None``, ``1.`` is set to all dimensions.
 
     Returns
     -------
@@ -91,30 +317,93 @@ def ellipsoid_coords(center, axis_lengths, *, shape=None,
         Voxel coordinates of ellipsoid.
         May be used to directl index into an array, e.g.
         ``img[dd, rr, cc] = 1``
+
+    Raises
+    ------
+    ValueError
+        If the length of ``center`` is not 3.
+    ValueError
+        If the length of ``axis_length`` is not 3.
+    ValueError
+        If the length of ``rotation_angles`` is not 3.
+    ValueError
+        If the length of ``spacing`` is not 3.
+    ValueError
+        If ``rotation_matrix`` is invalid.
+    ValueError
+        If ``rotation_order`` is invalid.
+
+    Notes
+    ------
+    ``rotation_angles`` are defined as the Euler angles based on
+    ``rotation_order`` and ``is_intrinsic``. The following example shows the
+    ``zxz`` intrinsic rotations.
+
+    .. math::
+
+        R(\phi) = \begin{bmatrix}
+                      1 &        0 &         0 \\
+                      0 & \cos\phi & -\sin\phi \\
+                      0 & \sin\phi &  \cos\phi \\
+                  \end{bmatrix}
+
+        R(\theta) = \begin{bmatrix}
+                        \cos\theta & -\sin\theta & 0 \\
+                        \sin\theta &  \cos\theta & 0 \\
+                                 0 &           0 & 1 \\
+                    \end{bmatrix}
+
+        R(\psi) = \begin{bmatrix}
+                      1 &        0 &         0 \\
+                      0 & \cos\psi & -\sin\psi \\
+                      0 & \sin\psi &  \cos\psi \\
+                           \end{bmatrix}
+
+        \\
+
+        R(\phi, \theta, \psi) = (R(\psi)R(\theta)R(\phi))^\intercal
+
     """
 
-    center = np.array([d, r, c])
-    radii = np.array([d_radius, r_radius, c_radius])
+    if len(center) != 3:
+        raise ValueError(f'len(center) should be 3 but got {len(center)}')
+    center = np.array(center)
 
-    # calculate a rotation matrix
-    rot_z %= np.pi
-    rot_y %= np.pi
-    rot_x %= np.pi
-    sin_rot_z, cos_rot_z = np.sin(rot_z), np.cos(rot_z)
-    sin_rot_y, cos_rot_y = np.sin(rot_y), np.cos(rot_y)
-    sin_rot_x, cos_rot_x = np.sin(rot_x), np.cos(rot_x)
-    rotation_z = np.array([[1, 0, 0],
-                           [0, cos_rot_z, sin_rot_z],
-                           [0, -sin_rot_z, cos_rot_z]])
-    rotation_y = np.array([[cos_rot_y, 0, -sin_rot_y],
-                           [0, 1, 0],
-                           [sin_rot_y, 0, cos_rot_y]])
-    rotation_x = np.array([[cos_rot_x, sin_rot_x, 0],
-                           [-sin_rot_x, cos_rot_x, 0],
-                           [0, 0, 1]])
-    rotation = rotation_z @ rotation_y @ rotation_x
+    if len(axis_lengths) != 3:
+        raise ValueError(
+            f'len(axis_lengths) should be 3 but got {len(axis_lengths)}')
+    axis_lengths = np.array(axis_lengths)
 
-    if len(spacing) != 3:
+    if rotation_matrix is not None:
+        if rotation_matrix.shape != (3, 3):
+            raise ValueError(
+                'rotation_matrix should have the shape (3, 3) but got',
+                rotation_matrix.shape)
+        elif not _is_valid_rotation_matrix(rotation_matrix):
+            raise ValueError(
+                'rotation_matrix should be an orthogonal matrix whose '
+                'determinant equals 1.')
+        rotmat = rotation_matrix
+    elif rotation_angles is not None:
+        if len(rotation_angles) != 3:
+            raise ValueError(
+                'len(rotation_angles) should be 3 but got',
+                len(rotation_angles))
+        if (not isinstance(rotation_order, str)
+            or len(rotation_order) != 3
+            or rotation_order.strip('xyz')
+            or rotation_order[0] == rotation_order[1]
+                or rotation_order[1] == rotation_order[2]):
+            raise ValueError(f'order: {rotation_order} is invalid')
+        rotation_angles = np.array(rotation_angles)
+        rotmat = _angles_to_rotmat(rotation_angles, order=rotation_order,
+                                   is_intrinsic=is_intrinsic)
+    else:
+        rotmat = np.eye(3)
+
+    if spacing is None:
+        spacing = np.ones(3)
+    elif len(spacing) != 3:
         raise ValueError(f'len(spacing) should be 3 but got {len(spacing)}')
     spacing = np.array(spacing)
     scaled_center = center / spacing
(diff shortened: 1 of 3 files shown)


skimage/draw/draw3d.py
```diff
@@ -63,6 +63,112 @@ def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
     return arr
 
 
+def ellipsoid_coords(d, r, c, d_radius, r_radius, c_radius, shape=None,
+                     rot_z=0., rot_y=0., rot_x=0., spacing=(1., 1., 1.)):
```
jni, line 67: Thanks for the contribution @ksugar! I'm a fan of your work. :blush: One thing that we're trying to do in skimage more generally is make our API n-dimensional-friendly, even if the underlying function is not *yet* n-dimensional. So, in this case, I would suggest:

```suggestion
def ellipsoid_coords(center, axis_lengths, *, shape=None,
                     rotation=(0., 0., 0.), spacing=(1., 1., 1.)):
```

I would then add a note in the docstring that the rotation is in ZYX Euler angles, and that this only supports 3D for now.

Also, I have euler matrix creation implemented in #5188, maybe there's the opportunity to build this on top of that one... I'll try to finish that off! btw if you have a bit of experience with 3D geometry/transforms I would totally appreciate a review there! :sweat_smile:

jni:
Thanks for the update @ksugar! Now all that's missing is updating the tests to match :grimacing: :joy:, and as far as I'm concerned it's ready to go! :tada:

jni (approved):
Pending getting the tests working, I'm happy with the code here. Thanks @ksugar! @scikit-image/core, can someone else take a look? :pray:

ksugar:
Hi @jni , thank you for your review! I am working on updating the docstrings and the tests.
Actually, I also want to make the code more general.
In my case, it is helpful if the function accepts the rotation matrix itself, rather than the Euler angles only.
I will introduce helper functions that calculate the rotation matrix from the Euler angles and the yaw-pitch-roll system.
The usage would be like this:
```
dd, rr, cc = ellipsoid_coords((5., 5., 5.), (3., 3., 3.), rotation=euler_to_rotation(0., 0., np.pi / 2))
``` 
Could you wait a little until I update the code in that way? Thanks!

jni:
@ksugar all good. I wonder though, whether a rotation matrix is too many degrees of freedom — indeed, a matrix may or may not be a rotation matrix. I'm wondering whether we should support *both* APIs: `rotation=<angles>`, and `rotation_matrix=<angles>`. What do you think?

ksugar:
@jni You have a point! Then, how about the interface like this?
```
def ellipsoid_coords(..., rotation=<angles>, rotation_matrix=<angles>, geometry='euler', ...):
```
, where `rotation` will be ignored if `rotation_matrix` is defined. `geometry` will support  `euler` or `ypr` (yaw-pitch-roll) at the moment. The `ypr` option would be helpful if the user wants to rotate the ellipsoid along the original axes.

:memo: Additional note: The motivation to support `rotation_matrix` is based on my personal use case. I obtained the rotation matrix after some user interactions on GUI, which can be decomposed to the angles in multiple ways as you mentioned. It would not be a good idea to decompose and re-compose the rotation matrix.

emmanuelle:
Hi @ksugar just a question about the API: I'm not sure why we need two functions here, `ellipsoid` and `ellipsoid_coords`? Wouldn't it be possible to adapt the `ellipsoid` function with more parameters instead?

ksugar:
Hi @emmanuelle `ellipsoid_coords` returns the voxel coordinates of an ellipsoid, while`ellipsoid` returns the 3d ndarray that renders an ellipsoid. This behavior of `ellipsoid` is different from the `draw` functions implemented in 2d (e.g. `ellipse`, `circle`, `disk`), which return the pixel coordinates. It would be possible to integrate `ellipsoid_coords` with `ellipsoid` by introducing a flag parameter but I am not sure if it is a good idea for two reasons. First, each function returns different type of values (3d ndarray or coordinates). Second, `ellipsoid` already has a flag `levelset` that determines how to render the ellipsoid and an additional flag can make the interface and the code complicated.

ksugar:
@emmanuelle @jni Please give me advice or comments on whether it would be better to combine `ellipsoid_coords` and `ellipsoid` or to keep them separated. Thanks!

jni:
@ksugar you're absolutely right that ellipsoid was the wrong API for draw, and I don't think we should harmonize the APIs. I think medium term we can deprecate ellipsoid and replace it with ellipsoid coords, but that deprecation can happen after this PR. @emmanuelle do you agree?

@ksugar 

> , where rotation will be ignored if rotation_matrix is defined. geometry will support euler or ypr (yaw-pitch-roll) at the moment. The ypr option would be helpful if the user wants to rotate the ellipsoid along the original axes.

My suggestion is that we leave the `geometry` option out of this PR and maybe try to add it to an upcoming PR. Maybe even the `rotation_matrix` option. It's good to keep PRs small and focused and not doing too many things or too many fancy things with the API, because complicated APIs tend to generate more discussion and PRs can get stalled for a long time. I think if you only have `rotation` we'll get it in very quickly — and `rotation_matrix` is a non-breaking change that can be discussed independently. ypr is yet another extra thing that will generate even more discussion.

Other than that, `rotation` being ignored if `rotation_matrix` is not None is pretty standard and I totally support that API. So up to you whether you want to do that in this PR or a separate one.

ksugar:
@jni thank you for your feedback and suggestions! I see your point that it is better to start from the minimal implementation. I will leave the `rotation_matrix` for the next PR. 

I found that we need to specify the order of rotations (e.g. `z-x-z`) to define the Euler angles. It is also required to specify if the rotations are intrinsic or extrinsic, which determine if the rotations occur about the rotating axes or the original axes. Actually, these factors made me confused to introduce the `geometry` option. Now, I think what I actually need to do is specifying these two parameters, `rotation_order` and `is_intrinsic`. I implemented them and updated the docstrings and tests. I also renamed `rotation` to `rotation_angles` so that it will not conflict with `rotation_matrix` in future.

jni:
Wow, lots of updates, thanks @ksugar! 😅

Some further comments:

- my personal preference would be to name the arguments `angles=`, `order=`, and `intrinsic=`. Perhaps someone else from @scikit-image/core can weigh in. The public API is the most important thing to get right, everything else can come later!
- this is probably a bit more controversial and arguable, but I just want to throw it out there: I would suggest that "order" should contain axes by *number* rather than name, since NumPy arrays don't actually have an order. So `order='zyx'` would actually be `order=(0, 1, 2)`. This adds an extra mental operation for people coming from the mathematics side, but it *removes* the potential for confusing axis ordering — right now someone must dig into the docstring to find that 0 = z, 1 = y, and 2 = x. Ahead of time, people probably expect that the axes are ordered as xyz, so they need to jump a mental hoop anyway: might as well make it one that aligns with our data.
- (totally optional) now that #5188 is merged, I wonder if you can extend the implementation there ([`skimage.transform._geometric._euler_rotation_matrix(angles, axes):`](https://github.com/scikit-image/scikit-image/blob/bc57d2cb1f9e1ac108f649227d996e574875bfdf/skimage/transform/_geometric.py#L1047-L1101) to take in `intrinsic` and use that. What do you think?

rfezzani:
I :+1: @jni suggested changes, I would also make `order` default to `None` corresponding to the `(0, 1, 2)`.

ksugar:
Hi @jni thank you for your feedback! [`_euler_rotation(axis, angle)`](https://github.com/scikit-image/scikit-image/blob/bc57d2cb1f9e1ac108f649227d996e574875bfdf/skimage/transform/_geometric.py#L1047-L1075) In your PR [#5188](https://github.com/scikit-image/scikit-image/pull/5188) generates a base rotation matrix in a smart way! It's amazing! I would like to adopt your function instead of my current implementation.

I would like to confirm about the  [`_euler_rotation_matrix(axis, angle)`](https://github.com/scikit-image/scikit-image/blob/bc57d2cb1f9e1ac108f649227d996e574875bfdf/skimage/transform/_geometric.py#L1078-L1101). The order of rotations is different from what I expected.
```
R = np.eye(dim)
for i, angle in zip(axes, angles):
    R = R @ _euler_rotation(i, angle)
```
will generate `R = R_0 @ R_1 @ R_2`. If it is applied to a vector (e.g. `R @ V`), the rotations will be applied in the order `R_2` -> `R_1` -> `R_0` (from right to left), which is opposite from my expectation.
```
R = np.eye(3)
if intrinsic:
    for i, angle in zip(axes, angles):
        R = _euler_rotation(i, angle) @ R
    R = R.T
else:
    for i, angle in zip(axes, angles):
        R = _euler_rotation(i, angle).T @ R
```
works perfectly in my case, where the transpose operation `.T` is required to convert from the rotations for axes to for positions.

I would also like to use the term `axes` instead of `order` to avoid complication.

ksugar:
@rfezzani I will follow the same convention as @jni, thanks for your feedback!

jni:
Hahaha "your function is amazing! btw, it is also wrong." :joy: 

I think you are almost certainly right @ksugar! :facepalm: Would be great if @grlee77, @stefanv or @rfezzani could check the above, as I'm a total newcomer to 3D rotations and they make my head hurt. :confused: But by my reading, we did get it wrong in #5188, even after much discussion. :joy: 

Thanks @ksugar!

ksugar:
@jni actually your code is great :+1:  I'm learning a lot of techniques from your code!
I agree that dealing with 3D rotations is a tough work... `ellipsoid_coords()` is now working as a test component of the PR [#5188](https://github.com/scikit-image/scikit-image/pull/5188). It's nice we work on the complementary topics at the same time!

If this change is approved by the members and `_euler_rotation_matrix(angles, axes)` is updated, I will replace the following part from
```
if intrinsic:
    for i, angle in zip(axes, angles):
        R = _euler_rotation(i, angle) @ R
    R = R.T
```
to
```
if intrinsic:
    R = _euler_rotation_matrix(angles, axes).T
```

jni:
kind ping to @stefanv @grlee77 @rfezzani, maybe @JDWarner, if someone could check our math for 3D Euler rotations it would be much appreciated! See discussion starting at https://github.com/scikit-image/scikit-image/pull/5263#issuecomment-799167248

rfezzani:
At first, I agreed with @ksugar, rotations should applied from left to right. But to convince myself, I wanted to compare `skimage` result with `open3D`:
```python
>>> import copy
>>> import open3d as o3d
>>> from skimage.transform._geometric import _euler_rotation_matrix
>>> angles = (np.pi/2,0,np.pi/4)
>>> mesh = o3d.geometry.TriangleMesh.create_coordinate_frame()
>>> mesh_r_o3d = copy.deepcopy(mesh).translate((2,0,0))
>>> mesh_r_skimage = copy.deepcopy(mesh).translate((4,0,0))
>>> mesh_r_o3d.rotate(mesh.get_rotation_matrix_from_xyz(angles))
TriangleMesh with 1134 points and 2240 triangles.
>>> mesh_r_skimage.rotate(_euler_rotation_matrix(angles))
TriangleMesh with 1134 points and 2240 triangles.
>>> o3d.visualization.draw_geometries([mesh, mesh_r_o3d, mesh_r_skimage])
>>> np.array_equmesh.get_rotation_matrix_from_xyz(angles)
np.array_equal(  np.array_equiv(  
>>> np.array_equal(mesh.get_rotation_matrix_from_xyz(angles), _euler_rotation_matrix(angles))
True
```
The results are equal!
Capture du 2021-03-25 11-32-07

jni:
@rfezzani Thank you! Just to be absolutely clear, you're using the function from this PR, right? 😅 

@ksugar do you want to make those changes you proposed above? Then we can probably get this merged! 🎉

rfezzani:
> @rfezzani Thank you! Just to be absolutely clear, you're using the function from this PR, right?

No @jni, I am using `main` branch :confused:

ksugar:
@rfezzani @jni it is great to compare the output with Open3D! I checked the source code of Open3D and its implementations are exactly same as what we can find in the `main` branch. [Eigen::Matrix3d RotationMatrix{X, Y, Z}(double radians)](https://github.com/intel-isl/Open3D/blob/v0.12.0/cpp/open3d/utility/Eigen.cpp#L265-L284) is equivalent to [skimage.transform._geometric._euler_rotation(axis, angle)](https://github.com/scikit-image/scikit-image/blob/main/skimage/transform/_geometric.py#L1047-L1075). [Eigen::Matrix3d Geometry3D::GetRotationMatrixFromXYZ](https://github.com/intel-isl/Open3D/blob/v0.12.0/cpp/open3d/geometry/Geometry3D.cpp#L151-L156) is equivalent to [skimage.transform._geometric._euler_rotation_matrix(angles, axes)](https://github.com/scikit-image/scikit-image/blob/main/skimage/transform/_geometric.py#L1078-L1101).

I would like to confirm the assumptions in the implementations.

1. Will `_euler_rotation(axis, angle)` and `_euler_rotation_matrix(angles, axes)` rotate points or standard basis vectors?
2. Is the `axis`/`axes` ordering `0 = x, 1 = y, and 2 = z` or ` 0 = z, 1 = y, and 2 = x`?

If `_euler_rotation(axis, angle)` and `_euler_rotation_matrix(angles, axes)` rotate points and the assumed ordering is  `0 = x, 1 = y, and 2 = z`, they are the same as Open3D and I think the implementations in the `main` branch is perfect!

On the other hand, I was assuming that the functions rotate standard basis vectors and the ordering is  `0 = x, 1 = y, and 2 = z`, which I think is the reason of incompatibility. I am happy with following the convention implemented in Open3D.

About the order of matrices in calculation, as far as I understand, the following explains it.
If we define `X`, `Y` and `Z` as the rotation matrices that rotate standard basis vectors, a `X>Y>Z` rotation matrix to rotate points is described below. To convert a rotation matrix from for standard basis vectors to for points, we need to transpose it.
 
```
R = (Z @ Y @ X).T = X.T @ Y.T @ Z.T
```

If we define `X'`, `Y'` and `Z'` as the rotation matrices that rotate points, the above equation can be described as below.

```
R = (Z @ Y @ X).T = X.T @ Y.T @ Z.T = X' @ Y' @ Z'
```

stefanv:
I don't have the background knowledge to comment on correctness. @jni are you able to do one more round of review so we can get this in?

I wonder if one of our resident scipy rotation experts—@pmla, @nmayorov, @adbugger, or @evbernardes—can help out?

evbernardes:
Hey there, @stefanv! If I can help you guys out with something, would be a pleasure!

What would you like to have checked out?

evbernardes:
I'm not familiar with `scikit-image`, so sorry if I'm mistaken about something! Here is just an idea, though:

Since `scipy` is already one of the requirements of `scikit-image`, and `scipy` is already used elsewhere in the code, wouldn't it sense to maybe take advantage of the `spatial.transform.Rotation` class in `scipy`?

One advantage is that it would be up to the user to decide how to enter a rotation (matrix, quaternions, angles, etc) and all the complicated 3D rotation-related maths are already implemented and well tested.

skimage/draw/draw3d.py
```diff
@@ -63,6 +64,178 @@ def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
...
+            axes = range(3)
+        if (len(axes) != 3
+            or not set(axes).issubset(range(3))
+            or axes[0] == axes[1]
+                or axes[1] == axes[2]):
+            raise ValueError(f'axes: {axes} is invalid')
+        if intrinsic:
+            for i, angle in zip(axes, angles):
+                R = _euler_rotation(i, angle) @ R
+            R = R.T
```
evbernardes, line 178: Since $(R_2 R_1)^T = R_1^T R_2^T$, this:
```python
for i, angle in zip(axes, angles):
    R = _euler_rotation(i, angle) @ R
R = R.T
```
Should be equivalent to:
```python
for i, angle in zip(axes, angles):
    R = R @ _euler_rotation(i, angle).T
```
Which would be a bit more consistent with the case where `intrinsic == False`.

I imagine there's a good reason why these rotations are inverted/transposed?

skimage/draw/draw3d.py
```diff
@@ -63,6 +64,178 @@ def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
...
+        else:
+            for i, angle in zip(axes, angles):
+                R = _euler_rotation(i, angle).T @ R
```
evbernardes, line 181: Alternatively, since `_euler_rotation_matrix` already implements extrinsic rotations, this could all be implemented as:
```python
if intrinsic:
    R = _euler_rotation_matrix(-angles[::-1], axes[::-1])
else:
    R = _euler_rotation_matrix(-angles, axes)
```
Where the minus sign is equivalent to the transpose.

evbernardes:
By the way, these comments are a bit old and possibly outdated, but I'll comment on them anyways:

> will generate R = R_0 @ R_1 @ R_2. If it is applied to a vector (e.g. R @ V), the rotations will be applied in the order R_2 -> R_1 -> R_0 (from right to left), which is opposite from my expectation.

This is exactly what I would expect, though: because of composition and how matrix products are defined, the first rotation is the rightmost matrix. 

Otherwise, applying the rotation $R_2 R_1$ to a vector $\pmb v$ would not be the same as first applying a rotation $R_1$ and then later applying the rotation $R_2$.

evbernardes:
Moreover, I'm not sure what this means:

> If we define `X`, `Y` and `Z` as the rotation matrices that rotate standard basis vectors, a `X>Y>Z` rotation matrix to rotate points is described below. To convert a rotation matrix from for standard basis vectors to for points, we need to transpose it.

Possibly some definition I have no idea about, if that's the case I'm sorry! But I don't see the relationship between rotating "points" with using the transpose of a rotation matrix (which is equivalent to the inverse rotation)

evbernardes:
> Is the axis/axes ordering `0 = x`, `1 = y`, and `2 = z` or  0 = z, 1 = y, and 2 = x?

I'm pretty sure its `0 = x`, `1 = y`, and `2 = z`, by looking at the implementation, since this:
```python
_euler_rotation(2, np.pi/4)
```
Gives the following ([a counterclockwise rotation of `np.pi / 4` radians around the `z` axis](https://en.wikipedia.org/wiki/Rotation_matrix#In_three_dimensions)):
```python
array([[ 0.70710678, -0.70710678,  0.        ],
       [ 0.70710678,  0.70710678,  0.        ],
       [ 0.        ,  0.        ,  1.        ]])
```

ksugar:
Hi @evbernardes , thank you for your feedback and comments.

I would like to revisit the order of the rotation steps.
What output do we expect when we call the following function?

```python
_euler_rotation_matrix(angles=(np.pi/4, np.pi/3, np.pi/2), axes=(0, 1, 2))
```

In this case, I expect to have the rotation matrix that works as follows:

1. rotate `np.pi/4` along the axis 0 (x-axis)
2. rotate `np.pi/3` along the axis 1 (y-axis)
3. rotate `np.pi/2` along the axis 2 (z-axis)

However, [the current implementation](https://github.com/scikit-image/scikit-image/blob/de46b0fb265fadb75e796836c73fb00923cad7a3/skimage/transform/_geometric.py#L1119-L1142) will give the rotation matrix that works in the oposite order:

1. rotate `np.pi/2` along the axis 2 (z-axis)
2. rotate `np.pi/3` along the axis 1 (y-axis)
3. rotate `np.pi/4` along the axis 0 (x-axis)

I assumed that the first elements in the list would apply first, but if the latter case is more common in the community then I will follow that convention.

evbernardes:
Hello, @ksugar! I think the confusion comes from a misunderstanding on how **intrinsic rotations** are defined, rather than the order of the rotations. Let me explain, but first, some definitions:

Let's define $\pmb e_1$, $\pmb e_2$ and $\pmb e_3$ as three different **constant** axes in the fixed frame, and $\theta_1$, $\theta_2$ and $\theta_3$ the corresponding rotation angles. Let's also define:
```math
R(\theta, \, \pmb e)
```
And to simplify the notation, we state:
```math
\begin{align}
R_1 = R(\theta_1, \, \pmb e_1)\\
R_2 = R(\theta_2, \, \pmb e_2)\\
R_3 = R(\theta_3, \, \pmb e_3)
\end{align}
```

A rotation of an angle $\theta$ around an axis $\pmb e$.

> In this case, I expect to have the rotation matrix that works as follows:
> 
>     1. rotate `np.pi/4` along the axis 0 (x-axis)
> 
>     2. rotate `np.pi/3` along the axis 1 (y-axis)
> 
>     3. rotate `np.pi/2` along the axis 2 (z-axis)

And this is exactly what happens. Take a $\pmb v$. Let's apply now the rotations, first rotation $R_1$:
```math
\pmb v' = R_1 \pmb v
```
Now the second:
```math
\begin{align}
\pmb v'' 
&= R_2 \pmb v' \\
&= R_2 (R_1 \pmb v)
\end{align}
```
And finally the third:
```math
\begin{align}
\pmb v''' 
&= R_3 \pmb v'' \\
&= R_3 (R_2 (R_1 \pmb v))
\end{align}
```
By the associativity of the matrix product, the full rotation matrix is:
```math
R_{\text{e}} = R_3 R_2 R_1
```
From right to left, which is exactly how it's implemented:
```python
    for i, angle in zip(axes, angles):
        R = R @ _euler_rotation(i, angle)
    return R
```
> However, [the current implementation](https://github.com/scikit-image/scikit-image/blob/de46b0fb265fadb75e796836c73fb00923cad7a3/skimage/transform/_geometric.py#L1119-L1142) will give the rotation matrix that works in the oposite order:
> 
>     1. rotate `np.pi/2` along the axis 2 (z-axis)
> 
>     2. rotate `np.pi/3` along the axis 1 (y-axis)
> 
>     3. rotate `np.pi/4` along the axis 0 (x-axis)
> 
> 
> I assumed that the first elements in the list would apply first, but if the latter case is more common in the community then I will follow that convention.

As I said, I think this confusion comes from a misunderstanding of how intrinsic rotations are defined (which is a pretty common confusion, I think this is very badly explained everywhere...)

I'll explain it better on another comment because this one is getting long.

evbernardes:
Sorry for the long comment! But since I absolutely couldn't find an explanation of this on the internet somehow, I decided it would be nice to have the development here.

> I assumed that the first elements in the list would apply first, but if the latter case is more common in the community then I will follow that convention.

In this case, I would say that the better alternative is to accept both intrinsic and extrinsic rotations.

And the absolute best, in my opinion, would be to take advantage of the fact that Scipy is an internal dependency and use instances of `Rotation` as an input, this way everyone can decide how they want to define their rotations (this would probably be a big API break though!)

For example, as you expressed in an old comment, you mostly use rotation matrices, if I'm not mistaken, while I almost exclusively use quaternions...

skimage/draw/draw3d.py
evbernardes: My apologies, apparently `_euler_rotation_matrix` implements **intrinsic rotations** instead, so:
```python
if intrinsic:
    R = _euler_rotation_matrix(-angles, axes)
else:
    R = _euler_rotation_matrix(-angles[::-1], axes[::-1])
```

ksugar:
@evbernardes Thank you for your kind explanation.
I now understand the difference between intrinsic and extrinsic rotations.

> And the absolute best, in my opinion, would be to take advantage of the fact that Scipy is an internal dependency and use instances of Rotation as an input, this way everyone can decide how they want to define their rotations (this would probably be a big API break though!)

I agree with the idea of using [scipy.spatial.transform.Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html).

I will update the code using it.

evbernardes:
> @evbernardes Thank you for your kind explanation.
> I now understand the difference between intrinsic and extrinsic rotations.

It's my pleasure! Don't hesitate if you need something else.

> I agree with the idea of using [scipy.spatial.transform.Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html).
> 
> I will update the code using it.

I also think this is the most practical way!

 I wonder if someone might have something to say against using an instance of a class from another library as an input, though (maybe @stefanv ?)

If that's the case, then I'd say using rotation matrices would be a pretty good middle ground.

skimage/draw/draw3d.py
```diff
@@ -63,6 +64,178 @@ def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
...
+    axes : iterable of int, length 3, optional (default None)
+        Axes about which to produce the rotations represented by ``{0, 1, 2}``,
+        where ``0 = z``, ``1 = y`` and ``2 = x``. There should be no adjacent
+        repeating numbers in the list. If not specified, defaults to 0, 1, 2.
+    intrinsic : bool optional (default True)
+        Specify if the rotations are intrinsic (``True``) or extrinsic
+        (``False``).
+
+        - *intrinsic*: rotations occur about the rotating coordinate system
+        - *extrinsic*: rotations occur about the original coordinate system
```
stefanv, line 96:
```suggestion
        - *intrinsic*: successive rotations occur about the rotating coordinate system
        - *extrinsic*: successive rotations occur about the original coordinate system
```

skimage/draw/draw3d.py
stefanv: This is counter-intuitive; order of angles and axes simply gets inverted, which changes from rotation around the original axes to that of the object? Not saying it's wrong, just that I don't get it :)

evbernardes:
@stefanv It's defined as a rotation around non-constant axes, but the maths simplify to an inversion!

I did a little demonstration on the comments if you're curious in seeing why it happens.

evbernardes:
I still don't get why the transpose/inverse rotations are used and what's the relationship of it to "rotation of points", so maybe this should be checked as well.

evbernardes:
I accidentally deleted my last comment, so here it is again:

I still don't see why the transposed/inverse rotations are used though, or what's the relationship of it with "rotations of points".

Maybe this should be checked again too?

ksugar:
@evbernardes regarding the following point, what I thought was drawn in the attached figure.

> I still don't see why the transposed/inverse rotations are used though, or what's the relationship of it with "rotations of points".

If we rotate the axes with a rotation matrix $R$, the coordinate of the point in the new space $(x', y', z')$ can be expressed with the coordinate in the original space $(x, y, z)$ as below? I am not quite sure if I understand it. I would be grateful if you could tell me if I have misunderstood something.

image

evbernardes:
@ksugar if understood correctly, it depends how you are defining the rotation matrix. A rotation matrix is just a representation of a rotation linear transformation, that could be define either from the inertial frame to the body frame, or from the body frame to the inertial frame.

I think by putting an inversion inside of the implementation could add some confusion, and I'd argue this is one of the reasons while just taking a matrix directly (or a Scipy Rotation instance) is a better solution.

evbernardes:
I'm more used to defining the rotations from the body frame to the inertial frame.

This means that, for example, defining a normal vector to some body along the body z axis, it would always look like (0 , 0, 1) in the body frame, and applying the rotation to it would express how you'd see it in the inertial frame.

evbernardes:
I forgot to mention: if the rotation matrix as the linear transformation that transforms one vector to the rotated version of the same vector, a unit vector that defines an axis is still just a vector, so there's nothing special about it, I think. And to rotate a point, depends also on how you define the point, but points can also be described by displacement vectors.

Hope that makes sense with your application!

jarrodmillman (approved):

stefanv:
@ksugar Are you happy with where this is at, given @evbernardes's comments above? I think we are ready to merge, and can get this into the 0.21 release.

ksugar:
@stefanv thank you for the updates and I'm glad to hear that this PR is going to be merged soon!
Based on the feedback from @evbernardes , I looked into [scipy.spatial.transform.Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html) and found that it's suitable for `draw3d.ellipsoid_coords`. For the moment, I just replaced some lines by using [scipy.spatial.transform.Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.from_euler.html). Please have a look at [the latest commit](https://github.com/scikit-image/scikit-image/pull/5263/commits/04c88f58ffb5bdfde84c962deae1240453747618).
Supporting the full features of [scipy.spatial.transform.Rotation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.from_euler.html) will be one of the future tasks.
I'm happy with merging either at [the latest commit](https://github.com/scikit-image/scikit-image/pull/5263/commits/04c88f58ffb5bdfde84c962deae1240453747618) or at [the previous one](https://github.com/scikit-image/scikit-image/pull/5263/commits/dba5fb24ff69aacd637af3ee4e0acf23b766f2ed).

stefanv:
This could make a nice example for the gallery, if you're up for it.

Meanwhile, I'll push a few touch-ups and get this merged.

stefanv:
Ran into some more questions; also @jni may have thoughts:

- Is it correct to rescale the center with spacing?
- Does it make sense to have spacing in this function?
- Introducing axis names here is rather confusion; most of us understand NumPy axis conventions, but they're not tied to X, Y, and Z.
- Is the `axis` argument necessary? Can't those angles be set to zero instead?
- Can we simplify the code by operating on all axes in a loop, or by vectorizing the operation? See, e.g.:

```python
    d_lim, r_lim, c_lim = np.ogrid[:float(bounding_shape[0]),
                                   :float(bounding_shape[1]),
                                   :float(bounding_shape[2])]
    d_org, r_org, c_org = scaled_center - upper_left_bottom
    d_rad, r_rad, c_rad = axis_lengths
    conversion_matrix = R.T @ np.diag(spacing)
    d, r, c = (d_lim - d_org), (r_lim - r_org), (c_lim - c_org)
    distances = (
        ((d * conversion_matrix[0, 0]
          + r * conversion_matrix[0, 1]
          + c * conversion_matrix[0, 2]) / d_rad) ** 2 +
        ((d * conversion_matrix[1, 0]
          + r * conversion_matrix[1, 1]
          + c * conversion_matrix[1, 2]) / r_rad) ** 2 +
        ((d * conversion_matrix[2, 0]
          + r * conversion_matrix[2, 1]
          + c * conversion_matrix[2, 2]) / c_rad) ** 2
    )
```

Ideally, we would not be repeating the code for three axes.

So, while we're close on this PR, I'm going to take this off the 0.21 milestone, and we can get it into 0.22.  We're on an increased release cadence now, so shouldn't take too long.

evbernardes:
Happy I was able to help! Here are some extra thoughts:

>     * Introducing axis names here is rather confusion; most of us understand NumPy axis conventions, but they're not tied to X, Y, and Z.

I think there are two advantages of using axis by their names `x, y, z`: 

1. There is no confusion between not knowing if its (0, 1, 2) or (1, 2, 3)
2. It's the convention already used by SciPy. 
 
If the focus on Scikits is to be extensions of SciPy, I'd argue that consistency with SciPy should be a priority over consistency with NumPy, don't you think?

skimage/draw/draw3d.py
```diff
@@ -63,6 +64,165 @@ def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
...
+            or axes[0] == axes[1]
+                or axes[1] == axes[2]):
+            raise ValueError(f'axes: {axes} is invalid')
+
+        axes_str = 'ZYX' if intrinsic else 'zyx'
+        seq = ''.join(axes_str[axis] for axis in axes)
+
+        # Generate a rotation matrix. The order of the elements needs to be
+        # reversed so that it follows the (z, y, x) order.
+        R = Rotation.from_euler(seq, angles).as_matrix()[::-1, ::-1]
```
evbernardes, line 167: What is this inversion doing exactly? And why is `axes_str` defined as `ZYX` instead of `XYZ`? 
Are there two definitions just inverting one another?

skimage/draw/draw3d.py
```diff
@@ -63,6 +64,165 @@ def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
...
+        lower_right_top = np.minimum(lower_right_top, np.array(shape[:3]) - 1)
+
+    bounding_shape = lower_right_top - upper_left_bottom + 1
+
+    d_lim, r_lim, c_lim = np.ogrid[:float(bounding_shape[0]),
+                                   :float(bounding_shape[1]),
+                                   :float(bounding_shape[2])]
+    d_org, r_org, c_org = scaled_center - upper_left_bottom
+    d_rad, r_rad, c_rad = axis_lengths
+    conversion_matrix = R.T @ np.diag(spacing)
```
evbernardes, line 199: And why is the inverse of the rotation used here?

skimage/draw/draw3d.py
```diff
@@ -63,6 +64,165 @@ def ellipsoid(a, b, c, spacing=(1., 1., 1.), levelset=False):
...
+    if len(center) != 3:
+        raise ValueError(f'len(center) should be 3 but got {len(center)}')
+    center = np.array(center)
+
+    if len(axis_lengths) != 3:
+        raise ValueError(
+            f'len(axis_lengths) should be 3 but got {len(axis_lengths)}')
+    axis_lengths = np.array(axis_lengths)
+
+    if angles is None:
```
evbernardes, line 147: Btw, you don't have to use rotation matrices. You can just define the `Rotation` object directly and use `Rotation.apply()` which is a lot more efficient.

stefanv:
> If the focus on Scikits is to be extensions of SciPy, I'd argue that consistency with SciPy should be a priority over consistency with NumPy, don't you think?

I think NumPy is making a mistake here. When working with arrays, we understand what axes 0, 1, and 2 are; but after two decades if doing that, if you ask me which ones represent X, Y, and Z, I wouldn't know. It depends on how you view your array?

skimage/draw/draw3d.py
stefanv: Typically, axis 0 corresponds not to X but to Z. Think, e.g., of an image represented as an array: `(M, N)` means `(Y, X)`, upside down, sort of. You see the confusion :)

skimage/draw/draw3d.py
stefanv: Probably because of the above.

skimage/draw/draw3d.py
evbernardes: I see, this is because of the application for images then!

evbernardes:
> > If the focus on Scikits is to be extensions of SciPy, I'd argue that consistency with SciPy should be a priority over consistency with NumPy, don't you think?
> 
> I think SciPy is making a mistake here. When working with arrays, we understand what axes 0, 1, and 2 are; but after two decades if doing that, if you ask me which ones represent X, Y, and Z, I wouldn't know. It depends on how you view your array?

I see, personally I'd say it's best to have it consistently, since they are most likely being used together. But ofc, this is up to you

evbernardes:
I still think it's best to use rotation matrices directly (or the rotation object) as an input though, since it introduces a possible very inefficient step of first decomposing the rotation into ZYX angles and then converting it to rotation matrices again

skimage/draw/draw3d.py
stefanv: I woke up last night, realizing that I had lied to you. It gets more complicated: scikit-image, in a regrettable but somewhat rational early API decision, chose to adopt the XYZ coordinate system for transformations. So that's layered on top of everything. We'll fix that in v2.

evbernardes: > I woke up last night, realizing that I had lied to you.

Good to know this doesn't only happen to me!

evbernardes: Here's an idea, for a future implementation.

What about a new `scikit.Rotation` class that inherits from `scipy.spatial.transform.Rotation` class, but only changes the `apply` method in order to invert which axes to use (XYZ or ZXY), and `as_euler`/`from_euler` to use your preferred standard of `0,1,2` instead of `X,Y,Z`?

This shouldn't be too time-consuming to implement with inheritance and would allow everyone to use whatever they prefer using (matrices, quaternions or angles).

matthew-brett:

<details>

## Rationale

</details>

PR #5264, opened 2021-03-10
Fix skimage.measure.label segfault

### LINKED ISSUE #5260: label() segfault with arrays of shape (n,0,m) on linux + anaconda
## Description

Hello!

I know this is a very weirdly shaped array to be labeling, but it was unintentional :~)

I haven't seen any similar issues, but sorry if this is a duplicate. #1479 seems similar but the code there runs without problems for me.

Also checked on windows and it doesn't happen there btw. Also, I'm using anaconda..

## Way to reproduce
```python
import numpy as np
from skimage.measure import label

a = np.zeros((300,300,300))
a = a[:,0:0,:]

label(a)
```

## Version information
```python
3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0]
Linux-5.8.0-44-generic-x86_64-with-glibc2.10
scikit-image version: 0.17.2
numpy version: 1.19.2
```

```python
# your output here
Segmentation fault (core dumped)
```

rfezzani:
## Description

Fixes #5260.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_ccomp.pyx
+++ b/skimage/measure/_ccomp.pyx
@@ -281,7 +281,7 @@ def _get_swaps(shp):
     swaps = []
 
     # Dimensions where the array is "flat"
-    ones = np.where(shp == 1)[0][::-1]
+    ones = np.where(shp <= 1)[0][::-1]
     # The other dimensions
     bigs = np.where(shp > 1)[0]
     # We now go from opposite directions and swap axes if an index of a flat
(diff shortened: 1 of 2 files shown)


jni (approved):
Thanks @rfezzani! My initial instinct was to aim for putting the check in the Python function, but that still would leave us open to segfaults e.g. if someone internally used the Cython for some other reason. So, this is strictly better I reckon. :+1:

jni:
The optional deps test failure is from a new dask release and not related to this PR.

grlee77 (approved):

grlee77:
Thanks for the quick fix, @rfezzani

PR #5265, opened 2021-03-10
Deprecate input kwarg in label

rfezzani:
## Description

Related to #4152.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/measure/_label.py
+++ b/skimage/measure/_label.py
@@ -30,8 +30,8 @@ def _label_bool(image, background=None, return_num=False, connectivity=None):
         return result[0]
 
 
-@deprecate_kwarg({"input": "image"}, removed_version="1.0")
-def label(image, background=None, return_num=False, connectivity=None):
+@deprecate_kwarg({"input": "label_image"}, removed_version="1.0")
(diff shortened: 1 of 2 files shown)


skimage/measure/_label.py
```diff
@@ -29,7 +30,8 @@ def _label_bool(image, background=None, return_num=False, connectivity=None):
         return result[0]
 
 
-def label(input, background=None, return_num=False, connectivity=None):
+@deprecate_kwarg({"input": "image"}, removed_version="1.0")
+def label(image, background=None, return_num=False, connectivity=None):
```
jni, line 34: I think this is actually a `label_image`... =)

jni (approved):
@rfezzani good move in general but I think the terminology we settled on is `label_image` for this kind of image. =) Could you update the PR for that? Other than that :+1: from me.

skimage/measure/_label.py
rfezzani: Done, Thank you @jni :wink:

grlee77 (approved):

PR #5269, opened 2021-03-12
Mention possible filters in radon_transform -> filtered-back-projection

kolibril13:
## Description
As this section https://scikit-image.org/docs/dev/auto_examples/transform/plot_radon_transform.html#reconstruction-with-the-filtered-back-projection-fbp explains that one can use filters, I think it would be nice to explicitly mention these filters.
Further, it might be nice to show an image of the filters, e.g. like this:
image
which was produced by this code:
```py
import numpy as np
from scipy.fftpack import fft

def frequency_response(size,filter_name):
    
    n = np.concatenate((np.arange(1, size / 2 + 1, 2, dtype=int),
                        np.arange(size / 2 - 1, 0, -2, dtype=int)))
    f = np.zeros(size)
    f[0] = 0.25
    f[1::2] = -1 / (np.pi * n) ** 2

    # Computing the ramp filter from the fourier transform of its
    # frequency domain representation lessens artifacts and removes a
    # small bias as explained in [1], Chap 3. Equation 61
    fourier_filter = 2 * np.real(fft(f))         # ramp filter
    if filter_name == "ramp":
        pass
    elif filter_name == "shepp-logan":
        # Start from first element to avoid divide by zero
        omega = np.pi * np.fft.fftfreq(size)[1:]
        fourier_filter[1:] *= np.sin(omega) / omega
    elif filter_name == "cosine":
        freq = np.linspace(0, np.pi, size, endpoint=False)
        cosine_filter = np.fft.fftshift(np.sin(freq))
        fourier_filter *= cosine_filter
    elif filter_name == "hamming":
        fourier_filter *= np.fft.fftshift(np.hamming(size))
    elif filter_name == "hann":
        fourier_filter *= np.fft.fftshift(np.hanning(size))
    elif filter_name is None:
        fourier_filter[:] = 1

filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']

for ix, f in enumerate(filters):
    response = frequency_response(2000,f)
    plt.plot(response, label= f)
plt.legend()
plt.show()
```
the code is basically this function:
https://github.com/scikit-image/scikit-image/blob/944760c6f925fa8e8dbb968c2aa100e1797b2c48/skimage/transform/radon_transform.py#L128-L181 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -98,8 +98,7 @@ plt.show()
 # back projection is among the fastest methods of performing the inverse
 # Radon transform. The only tunable parameter for the FBP is the filter,
 # which is applied to the Fourier transformed projections. It may be used to
-# suppress high frequency noise in the reconstruction. ``skimage`` provides a
-# few different options for the filter.
+# suppress high frequency noise in the reconstruction. ``skimage`` provides 'ramp', 'shepp-logan', 'cosine', 'hamming', 'hann' as filters.
 
 from skimage.transform import iradon
 
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -104,12 +104,15 @@ plt.show()
 import matplotlib.pyplot as plt
 from skimage.transform.radon_transform import _get_fourier_filter
 
-filters = ["ramp", "shepp-logan", "cosine", "hamming", "hann"]
-for f in filters:
+filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']
+
+for ix, f in enumerate(filters):
     response = _get_fourier_filter(2000, f)
-    plt.plot(response, label=f)
+    plt.plot(response, label= f)
+
+plt.xlim([0, 1000])
+plt.xlabel('frequency')
 plt.legend()
-plt.xlabel("position in pixel")
 plt.show()
 
 ######################################################################
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -98,7 +98,8 @@ plt.show()
 # back projection is among the fastest methods of performing the inverse
 # Radon transform. The only tunable parameter for the FBP is the filter,
 # which is applied to the Fourier transformed projections. It may be used to
-# suppress high frequency noise in the reconstruction. ``skimage`` provides 'ramp', 'shepp-logan', 'cosine', 'hamming', 'hann' as filters.
+# suppress high frequency noise in the reconstruction. ``skimage`` provides 
+# 'ramp', 'shepp-logan', 'cosine', 'hamming', 'hann' as filters.
 
 from skimage.transform import iradon
 
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -103,7 +103,7 @@ plt.show()
 
 from skimage.transform import iradon
 
-reconstruction_fbp = iradon(sinogram, theta=theta, circle=True)
+reconstruction_fbp = iradon(sinogram, theta=theta, filter_name='ramp', circle=True)
 error = reconstruction_fbp - image
 print(f"FBP rms reconstruction error: {np.sqrt(np.mean(error**2)):.3g}")
 
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -108,7 +108,7 @@ filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']
 
 for ix, f in enumerate(filters):
     response = _get_fourier_filter(2000, f)
-    plt.plot(response, label= f)
+    plt.plot(response, label=f)
 
 plt.xlim([0, 1000])
 plt.xlabel('frequency')
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -117,7 +117,8 @@ plt.show()
 
 from skimage.transform import iradon
 
-reconstruction_fbp = iradon(sinogram, theta=theta, filter_name='ramp', circle=True)
+reconstruction_fbp = iradon(sinogram, theta=theta, filter_name='ramp',
+                            circle=True)
 error = reconstruction_fbp - image
 print(f"FBP rms reconstruction error: {np.sqrt(np.mean(error**2)):.3g}")
 
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -98,8 +98,8 @@ plt.show()
 # back projection is among the fastest methods of performing the inverse
 # Radon transform. The only tunable parameter for the FBP is the filter,
 # which is applied to the Fourier transformed projections. It may be used to
-# suppress high frequency noise in the reconstruction. ``skimage`` provides the filters
-# 'ramp', 'shepp-logan', 'cosine', 'hamming', and 'hann':
+# suppress high frequency noise in the reconstruction. ``skimage`` provides
+# the filters 'ramp', 'shepp-logan', 'cosine', 'hamming', and 'hann':
 
 import matplotlib.pyplot as plt
 from skimage.transform.radon_transform import _get_fourier_filter
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -112,8 +112,8 @@ plt.legend()
 plt.xlabel("position in pixel")
 plt.show()
 
-# Applying the inverse radon transformation with the 'ramp' filter can be
-# seen here:
+######################################################################
+# Applying the inverse radon transformation with the 'ramp' filter, we get:
 
 from skimage.transform import iradon
 
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -104,10 +104,10 @@ plt.show()
 import matplotlib.pyplot as plt
 from skimage.transform.radon_transform import _get_fourier_filter
 
-filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']
+filters = ["ramp", "shepp-logan", "cosine", "hamming", "hann"]
 for f in filters:
-    response = _get_fourier_filter(2000,f)
-    plt.plot(response, label= f)
+    response = _get_fourier_filter(2000, f)
+    plt.plot(response, label=f)
 plt.legend()
 plt.xlabel("position in pixel")
 plt.show()
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -112,7 +112,8 @@ plt.legend()
 plt.xlabel("position in pixel")
 plt.show()
 
-## Applying the inverse radon transformation with the 'ramp' filter can be seen here:
+# Applying the inverse radon transformation with the 'ramp' filter can be
+# seen here:
 
 from skimage.transform import iradon
 
--- a/doc/examples/transform/plot_radon_transform.py
+++ b/doc/examples/transform/plot_radon_transform.py
@@ -98,8 +98,21 @@ plt.show()
 # back projection is among the fastest methods of performing the inverse
 # Radon transform. The only tunable parameter for the FBP is the filter,
 # which is applied to the Fourier transformed projections. It may be used to
-# suppress high frequency noise in the reconstruction. ``skimage`` provides 
-# 'ramp', 'shepp-logan', 'cosine', 'hamming', 'hann' as filters.
+# suppress high frequency noise in the reconstruction. ``skimage`` provides the filters
+# 'ramp', 'shepp-logan', 'cosine', 'hamming', and 'hann':
+
+import matplotlib.pyplot as plt
+from skimage.transform.radon_transform import _get_fourier_filter
+
+filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']
+for f in filters:
+    response = _get_fourier_filter(2000,f)
+    plt.plot(response, label= f)
+plt.legend()
+plt.xlabel("position in pixel")
+plt.show()
+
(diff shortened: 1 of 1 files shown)


doc/examples/transform/plot_radon_transform.py
```diff
@@ -98,8 +98,7 @@
 # back projection is among the fastest methods of performing the inverse
 # Radon transform. The only tunable parameter for the FBP is the filter,
 # which is applied to the Fourier transformed projections. It may be used to
-# suppress high frequency noise in the reconstruction. ``skimage`` provides a
-# few different options for the filter.
+# suppress high frequency noise in the reconstruction. ``skimage`` provides 'ramp', 'shepp-logan', 'cosine', 'hamming', 'hann' as filters.
```
kolibril13, line 101:
```suggestion
# suppress high frequency noise in the reconstruction. ``skimage`` provides 
# 'ramp', 'shepp-logan', 'cosine', 'hamming', 'hann' as filters.
```

rfezzani:
Excellent suggestion @kolibril13, thank you! The figure you generated is also nice, you can add it to your PR :wink:

kolibril13:
Thanks for the feedback!
Regarding the way to include the image I see three approaches:
## 1.
Adding the `_get_fourier_filter` to `__all__` and then adding a short script:
```
filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']
for ix, f in enumerate(filters):
    response = frequency_response(2000,f)
    plt.plot(response, label= f)
plt.legend()
plt.xlabel("Position in Pixel")
plt.show()
```
But that would make changes to the core library, which might not be wanted (although I think having access to the `frequency_response` function might be useful in other cases as well.)
## 2.
3. Copying the complete code from get_fourier_filter to this example, like this:
```py
import numpy as np
from scipy.fftpack import fft

def frequency_response(size,filter_name):
    
    n = np.concatenate((np.arange(1, size / 2 + 1, 2, dtype=int),
                        np.arange(size / 2 - 1, 0, -2, dtype=int)))
    f = np.zeros(size)
    f[0] = 0.25
    f[1::2] = -1 / (np.pi * n) ** 2

    # Computing the ramp filter from the fourier transform of its
    # frequency domain representation lessens artifacts and removes a
    # small bias as explained in [1], Chap 3. Equation 61
    fourier_filter = 2 * np.real(fft(f))         # ramp filter
    if filter_name == "ramp":
        pass
    elif filter_name == "shepp-logan":
        # Start from first element to avoid divide by zero
        omega = np.pi * np.fft.fftfreq(size)[1:]
        fourier_filter[1:] *= np.sin(omega) / omega
    elif filter_name == "cosine":
        freq = np.linspace(0, np.pi, size, endpoint=False)
        cosine_filter = np.fft.fftshift(np.sin(freq))
        fourier_filter *= cosine_filter
    elif filter_name == "hamming":
        fourier_filter *= np.fft.fftshift(np.hamming(size))
    elif filter_name == "hann":
        fourier_filter *= np.fft.fftshift(np.hanning(size))
    elif filter_name is None:
        fourier_filter[:] = 1

filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']

for ix, f in enumerate(filters):
    response = frequency_response(2000,f)
    plt.plot(response, label= f)
plt.legend()
plt.show()
````
I don't like this option too much, because it would display a lot of extra code in this section, and further, it needs to be changed when frequency_response would change in future.
## 3. 
Adding the image as a static file here: https://github.com/kolibril13/scikit-image/tree/patch-1/doc/source/_static

**Comment by mkcor** [comment]
Hi @kolibril13,

You can readily import internal function `_get_fourier_filter`:

```py
from skimage.transform.radon_transform import _get_fourier_filter
```

kolibril13:
> Hi @kolibril13,
> 
> You can readily import internal function `_get_fourier_filter`:
> 
> ```python
> from skimage.transform.radon_transform import _get_fourier_filter
> ```

An elegant solution, thanks! I've now updated the pr, it is now ready for review.

doc/examples/transform/plot_radon_transform.py
```diff
@@ -98,12 +98,25 @@
 # back projection is among the fastest methods of performing the inverse
 # Radon transform. The only tunable parameter for the FBP is the filter,
 # which is applied to the Fourier transformed projections. It may be used to
-# suppress high frequency noise in the reconstruction. ``skimage`` provides a
-# few different options for the filter.
+# suppress high frequency noise in the reconstruction. ``skimage`` provides the filters
+# 'ramp', 'shepp-logan', 'cosine', 'hamming', and 'hann':
```
rfezzani, line 102:
```suggestion
# suppress high frequency noise in the reconstruction. ``skimage`` provides
# the filters 'ramp', 'shepp-logan', 'cosine', 'hamming', and 'hann':
```

doc/examples/transform/plot_radon_transform.py
```diff
@@ -98,12 +98,25 @@
...
+
+filters = ["ramp", "shepp-logan", "cosine", "hamming", "hann"]
+for f in filters:
+    response = _get_fourier_filter(2000, f)
+    plt.plot(response, label=f)
+plt.legend()
+plt.xlabel("position in pixel")
+plt.show()
+
+## Applying the inverse radon transformation with the 'ramp' filter can be seen here:
```
rfezzani, line 115:
```suggestion
# Applying the inverse radon transformation with the 'ramp' filter can be
# seen here:
```

doc/examples/transform/plot_radon_transform.py
```diff
@@ -98,12 +98,25 @@
...
 
 from skimage.transform import iradon
 
-reconstruction_fbp = iradon(sinogram, theta=theta, circle=True)
+reconstruction_fbp = iradon(sinogram, theta=theta, filter_name='ramp', circle=True)
```
rfezzani, line 119:
```suggestion
reconstruction_fbp = iradon(sinogram, theta=theta, filter_name='ramp',
                            circle=True)
```

rfezzani (commented):
Just to address PEP8 complaints :wink:

rfezzani (approved):
Thank you @kolibril13!

doc/examples/transform/plot_radon_transform.py
```diff
@@ -98,12 +98,27 @@
...
+filters = ["ramp", "shepp-logan", "cosine", "hamming", "hann"]
+for f in filters:
+    response = _get_fourier_filter(2000, f)
+    plt.plot(response, label=f)
+plt.legend()
+plt.xlabel("position in pixel")
+plt.show()
+
+# Applying the inverse radon transformation with the 'ramp' filter can be
+# seen here:
```
mkcor, line 116:
```suggestion
######################################################################
# Applying the inverse radon transformation with the 'ramp' filter, we get:
```
This will prevent the two figures of this section from getting squished together!

mkcor (approved):
Approving up the rendering suggestion I made, after looking at the [rendered web page](https://1533-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/transform/plot_radon_transform.html#reconstruction-with-the-filtered-back-projection-fbp).

PS: When using the GitHub interface, note that, instead of committing them one by one, you can add several code review suggestions to a single batch and commit the latter at once. Well, here I only made one suggestion anyway. Thanks, @kolibril13!

kolibril13:
Before merging, I have one thing that seems to be a bit weird:
The frequency filters are supposed to reduce the low frequencies, hence the minimum of the filter function should be in the middle, as it can be seen for example here:
image
https://www.researchgate.net/figure/a-Ram-Lak-Hamming-and-Hanning-filters-in-frequency-domain-b-Ram-Lak-Hamming-and_fig1_274733291
I get the same result when I shift the x values of the filter functions by half of its total length.:
```py
from skimage.transform.radon_transform import _get_fourier_filter

filters = ["ramp", "shepp-logan", "cosine", "hamming", "hann"]
for f in filters:
    response = _get_fourier_filter(1000, f)
    response=np.roll(response, int(len(response)/2))
    plt.plot(response, label=f)
plt.legend()
plt.xlabel("frequency domain")
plt.show()
```
image

But I don't know the reason for this.
Maybe this might be related to this line, and that the Fourier transform of the image is not shifted.
https://github.com/scikit-image/scikit-image/blob/944760c6f925fa8e8dbb968c2aa100e1797b2c48/skimage/transform/radon_transform.py#L291

mkcor:
Good catch, @kolibril13! I would think that the shift should happen upstream, i.e., when constructing the actual Fourier filter, so when defining function `_get_fourier_filter()`: `fourier_filter = fftmodule.fftshift(fourier_filter)` should replace
https://github.com/scikit-image/scikit-image/blob/a69d49272145d9839df75f3ee398c0bbde32f448/skimage/transform/radon_transform.py#L165
and be added after
https://github.com/scikit-image/scikit-image/blob/a69d49272145d9839df75f3ee398c0bbde32f448/skimage/transform/radon_transform.py#L169
Let me look at the existing tests...

mkcor:
Hi @kolibril13,

I checked and, actually, the code is correct. The ramp filter, which is the default filter, is a high-pass filter: Its value is zero (minimum) at zero (lowest frequency). In the picture you attached, the middle value on the x-axis must be zero.  

Please refer to these references for more details:

* https://www.mathworks.com/help/images/ref/iradon.html
* https://homepages.inf.ed.ac.uk/rbf/CVonline/LOCAL_COPIES/AV0405/HAYDEN/Slice_Reconstruction.html

Specify the frequency range when displaying the plot:
```py
from skimage.transform.radon_transform import _get_fourier_filter

filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']

for ix, f in enumerate(filters):
    response = _get_fourier_filter(2000, f)
    plt.plot(response, label= f)

plt.xlim([0, 1000])
plt.xlabel('frequency')
plt.legend()
plt.show()
```
filters

doc/examples/transform/plot_radon_transform.py
```diff
@@ -98,12 +98,27 @@
...
+import matplotlib.pyplot as plt
+from skimage.transform.radon_transform import _get_fourier_filter
+
+filters = ["ramp", "shepp-logan", "cosine", "hamming", "hann"]
+for f in filters:
+    response = _get_fourier_filter(2000, f)
+    plt.plot(response, label=f)
+plt.legend()
+plt.xlabel("position in pixel")
+plt.show()
```
kolibril13, line 113:
```suggestion
import matplotlib.pyplot as plt
from skimage.transform.radon_transform import _get_fourier_filter

filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']

for ix, f in enumerate(filters):
    response = _get_fourier_filter(2000, f)
    plt.plot(response, label= f)

plt.xlim([0, 1000])
plt.xlabel('frequency')
plt.legend()
plt.show()
```

kolibril13:
> I checked and, actually, the code is correct. The ramp filter, which is the default filter, is a high-pass filter: Its value is zero (minimum) at zero (lowest frequency). In the picture you attached, the middle value on the x-axis must be zero.

Thanks for clarifying, that sounds logical to me, and I updated the pr according to your limits with `plt.xlim([0, 1000])`
But still, I am confused by the range  `plt.xlim([1000, 2000]):`
image
Should it not be something like this?:
```py
from skimage.transform.radon_transform import _get_fourier_filter

filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']

for ix, f in enumerate(filters):
    response = _get_fourier_filter(2000, f)
    plt.plot(response, label= f)

plt.xlim([-1000, 1000])
plt.xlabel('frequency')
plt.legend()
plt.show()
```
image where the left side of the k-space should be filled as well?

doc/examples/transform/plot_radon_transform.py
```diff
@@ -98,12 +98,30 @@
...
+# the filters 'ramp', 'shepp-logan', 'cosine', 'hamming', and 'hann':
+
+import matplotlib.pyplot as plt
+from skimage.transform.radon_transform import _get_fourier_filter
+
+filters = ['ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']
+
+for ix, f in enumerate(filters):
+    response = _get_fourier_filter(2000, f)
+    plt.plot(response, label= f)
```
mkcor, line 111:
```suggestion
    plt.plot(response, label=f)
```

doc/examples/transform/plot_radon_transform.py
```diff
@@ -98,12 +98,30 @@
...
+plt.show()
+
+######################################################################
+# Applying the inverse radon transformation with the 'ramp' filter, we get:
 
 from skimage.transform import iradon
 
-reconstruction_fbp = iradon(sinogram, theta=theta, circle=True)
+reconstruction_fbp = iradon(sinogram, theta=theta, filter_name='ramp',
+                            circle=True)
```
mkcor, line 124:
```suggestion
reconstruction_fbp = iradon(sinogram, theta=theta, filter_name='ramp')
```
I know it's not related to this PR specifically, but I don't see the point of specifying `circle=True`, since it's the default value for this parameter *and* the narrative parts of the example do not refer to this parameter. Poking @rfezzani just in case.

mkcor:
Thank you for updating the PR, @kolibril13! I just found some last minor details to fix... :wink: 

> Should it not be something like this?
> where the left side of the k-space should be filled as well?

Well, it *is* like this! The filter is of size 2000 (cutoff frequency). So the spectrum is infinitely repeating every 2000. The ramp filter is the absolute value function on its bandwidth (see my previous reply). Since the filter is symmetric *and* periodic, it is enough to represent it on `[0, 1000]` to show all the information.

kolibril13:
@mkcor :
Thanks for merging and your follow up pr!
I am glad I could contribute to this amazing project.

I just realized that "None" is also a filter option (https://github.com/scikit-image/scikit-image/blob/358ce872732faa18250ea6e540495348ccaa6735/skimage/transform/radon_transform.py#L178) for the inverse radon, maybe one can add this for completeness, however it won't be too important.
Here would be the included idea:
```py
from skimage.transform.radon_transform import _get_fourier_filter

filters = [None,'ramp', 'shepp-logan', 'cosine', 'hamming', 'hann']

for ix, f in enumerate(filters):
    response = _get_fourier_filter(2000, f)
    if f is None:
        plt.plot(response, label="None")
    else:
        plt.plot(response, label=f)

plt.xlim([0, 1000])
plt.xlabel('frequency')
plt.legend()
plt.show()
```
image

PR #5271, opened 2021-03-14
Watershed: consider connectivity when calculating markers.

### LINKED ISSUE #5084: Question: Why is default structuring element of 1-connectivity used by ndimage.label for getting the default markers(labeled local minima) for watershed segmentation?
## Description

In the implemented watershed.py segmentation, one can pass a connectivity argument, when `local_minima` is called by watershed (when `markers=none`) it passes the same connectivity value onto the local minima function, this generates the `markers_bool`, which then is used by `ndimage.label()` to generate the labeled markers. `ndimage.label` can take a structure argument and it defaults to 1-connectivity structure if none are given. IMO it is more intuitive to have `ndimage.label` also use the same connectivity which is passed by the user to the watershed function instead of always using the default 1-connectivity.
Below is the line in the source I am referring to:
https://github.com/scikit-image/scikit-image/blob/663e9e4b0f2ba88249bfef45bc65700555a3f24b/skimage/segmentation/_watershed.py#L76

Instead of the above would it not be better to have?
```python
if connectivity==2:
   markers = ndi.label(markers_bool,
                       structure = [[1,1,1],
                                   [1,1,1],
                                   [1,1,1]])[0]
```

divyank0:
## Description

here, we pass the user supplied parameter connectivity to ndi.label function instead of using default parameter value. 
Changes also include the test case to cover the difference that may arise due to the change in code. 

closes #5084 

## For reviewers

  later.

### DIFF
--- a/skimage/segmentation/_watershed.py
+++ b/skimage/segmentation/_watershed.py
@@ -73,7 +73,8 @@ def _validate_inputs(image, markers, mask, connectivity):
             raise ValueError(message)
     if markers is None:
         markers_bool = local_minima(image, connectivity=connectivity) * mask
-        markers = ndi.label(markers_bool)[0]
+        footprint = ndi.generate_binary_structure(markers_bool.ndim, connectivity)
+        markers = ndi.label(markers_bool, structure=footprint)[0]
     elif not isinstance(markers, (np.ndarray, list, tuple)):
         # not array-like, assume int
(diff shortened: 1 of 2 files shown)


rfezzani (approved):
Thank you for this contribution @divyank0, very clean! :wink:

jni (approved):
Perfect! 🎉 Thanks @divyank0!

PR #5276, opened 2021-03-18
Add data files for new illuminants.

mkcor:
## Description

Following up with #5234 -- only adding the new data here. 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


mkcor:
The current test fail is not specific to this PR.

@hmaarrfk @BierretA I have manually cherry-picked the new (non-duplicate) data files from #5234.

mkcor:
@hmaarrfk please review/approve ;)

hmaarrfk (approved):
It seems that adding the files to the registry itself causes test to fail.

I noticed that the azure pipelines all build off the build wheels.

They will fail without a pull request like this.

hmaarrfk:
I'm kinda ok with this, but I think that merging https://github.com/scikit-image/scikit-image/pull/5283 would have caused fewer failures.

I can't really inspect the azure failures, but in either case, it is clear that the build system has stopped that PR from going through.

I would be ok just merging the main PR as is.

mkcor:
> It seems that adding the files to the registry itself causes test to fail.

Yes, because the files don't live in `main` yet! So, if we merge this PR first, then we can go on with the data registry update (via a PR which will pass CI).

mkcor:
> I'm kinda ok with this, but I think that merging #5283 would have caused fewer failures.

@hmaarrfk "would cause" :wink: Let's fix the CI script first, of course!

mkcor:
Let's move forward with this, now that #5283 is merged! /cc @grlee77 @jni

PR #5280, opened 2021-03-20
Fix dtype info in documentation for watershed.

### LINKED ISSUE #5278: Error in documentation for watershed
The [documentation for watershed](https://scikit-image.org/docs/dev/api/skimage.segmentation.html#skimage.segmentation.watershed) says

> image : ndarray (2-D, 3-D, …) of integers
> Data array where the lowest value points are labeled first.

But when I look into the code and the [official example](https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_watershed.html), it seems that watershed is actually dealing with float arrays, not integer arrays. I think this might be a mistake in the documentation. If not, please explain how so. If so, I would be happy if someone - who knows how - could correct the documentation.

mkcor:
## Description

Fixes #5278

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/segmentation/_watershed.py
+++ b/skimage/segmentation/_watershed.py
@@ -98,7 +98,7 @@ def watershed(image, markers=None, connectivity=1, offset=None, mask=None,
 
     Parameters
     ----------
-    image : ndarray (2-D, 3-D, ...) of integers
+    image : ndarray (2-D, 3-D, ...)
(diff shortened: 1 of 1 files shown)


grlee77 (approved):
Looks good, thanks @mkcor (and @Make42!)

PR #5281, opened 2021-03-20
Improved gaussian filter and convert_to_float(#5195)

### LINKED ISSUE #5195: `filters.gaussian` does not work on 1D input

michalkrawczyk:
## Description

Propositon of fix #5195.
Since one point of problem, which is with not handling 1D array, was already resolved - This is mostly fix of normalizing in img_to_float() from int types according to [Data Conversion Rules](https://msdn.microsoft.com/en-us/library/windows/desktop/dd607323%28v=vs.85%29.aspx)

Also changed preserve_range from filters.gaussian to True. The values in Gaussian Filter shoudn't be normalized by default - It will only give misleading results (Like in Issue [#5195](https://github.com/scikit-image/scikit-image/issues/5195) itself)

## For reviewers

- Check filters.gaussian() for 1D array ( e.g np.arange(9)), especially with preserve_range=False
- Check util.dtype.img_to_float() with some 1D, ... ,nD values
- Check if current usages of filters.gaussian() really need preserve_range==False (Optional)


skimage/util/dtype.py
```diff
@@ -313,7 +313,7 @@ def _convert(image, dtype, force_copy=False, uniform=False):
         computation_type = _dtype_itemsize(itemsize_in, dtype_out,
                                            np.float32, np.float64)
 
-        if kind_in == 'u':
+        if kind_in == 'u' or kind_in == 'i':
```
mkcor, line 316: From how the code unit is written, the case of signed ints is handled by the `else` statement. A good way to express the changes you mean would be to update or add tests so they do exactly what you're looking for, then make the changes in the codebase so that the new tests pass (TDD, for test-driven development).

skimage/filters/_gaussian.py
```diff
@@ -10,7 +10,7 @@
 
 
 def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
-             multichannel=None, preserve_range=False, truncate=4.0):
+             multichannel=None, preserve_range=True, truncate=4.0):
```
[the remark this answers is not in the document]
michalkrawczyk: I quite didn't see point of doing that here. See part of current docstring:

>     preserve_range : bool, optional
        Whether to keep the original range of values. Otherwise, the input
        image is converted according to the conventions of ``img_as_float``.
        Also see
        
 But maybe it should be rewritten to be more specific. If you consider that as necessery - I may do that

skimage/util/dtype.py
michalkrawczyk: Good point. It was handled by else statement, but that was one of the main problems in [#5281](https://github.com/scikit-image/scikit-image/pull/5281#issue-597330331) . I'll make change in unit test

michalkrawczyk:
A bit changed conversion at dtype since according to Direct3D most negative value should be not calculated by formula, but just changed to -1.0f. That made test to fail (Also added few more tests)

michalkrawczyk:
It seems like some test just failed because some functions are basing at gaussian with preserve_range=False.
I actually start to wonder if that usages should be rewritten to preserve_range=False or just get back to preserve_range on default set to True and leave note at documentation.

skimage/filters/tests/test_gaussian.py
```diff
@@ -62,8 +62,25 @@ def test_multichannel():
 
 
 def test_preserve_range():
+    """
+    Testing preserve_range param
+    """
+    ones = np.array([[1, 1], [1, 1]], dtype=np.int64)
+    filtered_ones = gaussian(ones, preserve_range=False)
+    assert(np.all(filtered_ones == filtered_ones[0]))
+
```
mkcor, line 71: Then I think it would make more sense to go along these lines:
```py
filtered_preserved = filters.gaussian(ones, preserve_range=True)
filtered_default = filters.gaussian(ones)
assert testing.assert_array_equal(filtered_default, filtered_preserved)
assert np.all(filtered_default == 1.)
```
What do you think?

skimage/filters/tests/test_gaussian.py
```diff
@@ -62,8 +62,25 @@ def test_multichannel():
...
     img = np.array([[10.0, -10.0], [-4, 3]], dtype=np.float32)
-    gaussian(img, 1, preserve_range=True)
+    gaussian(img, 1)
+
+
+def test_1d_ok():
+    """
+    Testing Gaussian Filter for 1D array.
+    With all positive integer values and only one zero, it should filter all values to be at least more than 0.1
```
mkcor, line 79:
```suggestion
    """Test Gaussian filter on 1D array.
    With seven positive integer values and only one zero, all filtered values should be greater than 0.1
```
Reference: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

skimage/filters/tests/test_gaussian.py
```diff
@@ -62,8 +62,25 @@ def test_multichannel():
...
+    """
+    nums = np.arange(7)
+    filtered = gaussian(nums)
+    assert(np.all(filtered > 0.1))
```
mkcor, line 83:
```suggestion
    assert np.all(filtered > 0.1)
```

skimage/filters/tests/test_gaussian.py
```diff
@@ -142,4 +159,4 @@ def test_dog_invalid_sigma2():
     with testing.raises(ValueError):
         difference_of_gaussians(image, 3, 2)
     with testing.raises(ValueError):
-        difference_of_gaussians(image, (1, 5), (2, 4))
+        difference_of_gaussians(image, (1, 5), (2, 4))
```
mkcor, line 162: Please keep EOL.

mkcor:
Since this PR changes the default value of a function's parameter, it will affect existing code. I guess we need some kind of warning message referring to the next release version, right? @scikit-image/core

alexdesiqueira:
>  I guess we need some kind of warning message referring to the next release version, right?

Hi @michalkrawczyk @mkcor, please refer to [our deprecation cycle policy](https://scikit-image.org/docs/dev/contribute.html#deprecation-cycle) in this case.
Thank you for your work!

skimage/filters/tests/test_gaussian.py
michalkrawczyk: Old habbit from C.  Sorry

skimage/filters/tests/test_gaussian.py
michalkrawczyk: Not exactly. I will rewrite that to be more clear

skimage/filters/tests/test_gaussian.py
michalkrawczyk: For now it is not necessary. 
But assuming that some changes may occur in the future - I would add that too

michalkrawczyk:
Done. @mkcor May I ask you to have a look now?

skimage/filters/tests/test_gaussian.py
```diff
@@ -62,8 +64,24 @@ def test_multichannel():
...
+
     img = np.array([[10.0, -10.0], [-4, 3]], dtype=np.float32)
-    gaussian(img, 1, preserve_range=True)
+    gaussian(img, 1)
+
+
+def test_1d_ok():
+    """Testing Gaussian Filter for 1D array.
+    With any array consisting of positive integers and only one zero - it should filter all values
+    to be at least more than 0.1
```
mkcor, line 80: "at least" and "more than" are redundant :wink: 
I recommend going for the mathematical standard way of saying (i.e., "greater than")

skimage/util/tests/test_dtype.py
```diff
@@ -191,3 +191,19 @@ def test_subclass_conversion():
         x = x.astype(dtype)
         y = _convert(x, np.floating)
         assert y.dtype == x.dtype
+
+
+def test_int_to_float():
+    """
+    Check Normalization when casting img_as_float from int types to float
```
mkcor, line 198: Please fix one-line docstring style

skimage/util/tests/test_dtype.py
```diff
@@ -191,3 +191,19 @@ def test_subclass_conversion():
...
+    """
+    int_list = np.arange(9, dtype=np.int64)
+    converted = img_as_float(int_list)
+    assert(np.allclose(converted, int_list * 1e-19, atol=0.0, rtol=0.1))
```
mkcor, line 202:
```suggestion
    assert np.allclose(converted, int_list * 1e-19, atol=0.0, rtol=0.1)
```
`assert` is a keyword in Python

skimage/filters/_gaussian.py
```diff
@@ -96,6 +98,11 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
     >>> filtered_img = gaussian(image, sigma=1, multichannel=True)
 
     """
+    if preserve_range is None:
+        msg = ('The default value of preserve_range will change '
+               'to "True" in version 1.0')
```
mkcor, line 103:
```suggestion
               'to True in version 1.0')
```
@scikit-image/core can you please have a look and say if you agree with this API change?

skimage/filters/_gaussian.py
```diff
@@ -39,9 +39,11 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
         the function will attempt to guess this, and raise a warning if
         ambiguous, when the array has shape (M, N, 3).
     preserve_range : bool, optional
-        Whether to keep the original range of values. Otherwise, the input
-        image is converted according to the conventions of ``img_as_float``.
-        Also see
+        If True (by default) - keep the original range of values.
```
stefanv, line 42: We will only be able to set this to True after the deprecation period.

skimage/filters/_gaussian.py
```diff
@@ -96,6 +98,11 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
     >>> filtered_img = gaussian(image, sigma=1, multichannel=True)
 
     """
+    if preserve_range is None:
+        msg = ('The default value of preserve_range will change '
+               'to True in version 1.0')
+        warn(msg, DeprecationWarning)
+        preserve_range = True
```
stefanv, line 105: This should be False for now

skimage/util/dtype.py
```diff
@@ -321,6 +321,15 @@ def _convert(image, dtype, force_copy=False, uniform=False):
             # DirectX uses this conversion also for signed ints
             # if imin_in:
             #     np.maximum(image, -1.0, out=image)
+        elif kind_in == 'i':
```
stefanv, line 324: @scikit-image/core Can anyone verify that this part of the patch is correct?

jni:
Hi all, sorry to jump in really late here but I don't think we should take this leap. `filters.gaussian` will be one of our most used functions and will now be obnoxiously noisy. Additionally, there is a question of consistency with the rest of the library — a quick search shows that all our functions use `preserve_range=False`, except `radon_transform` *and* `non_local_means`, *for which we are in the middle of a deprecation to change it to False!*

In short, although I do support setting preserve_range=True everywhere, it should be done as a concerted effort in the road to 1.0, *not* as a piecemeal, whatever is most convenient for a specific function at a specific time approach.

michalkrawczyk:
@jni Yeah I do see that actually almost all your current libraries use `preserve_range=False`. 
The problem is that library requires more `preserved_range=False` while users requires more `preserved_range=True` and misleading results from False are more considered as bug. 
It's not actually more convienient - it's more likely how gaussian filter should be in practice, without any 'additional features' by default.

skimage/util/dtype.py
michalkrawczyk: @stefanv What you're  actually concern about in this part? Maybe I should give you some details, so you would be able to answer by yourself?

jni:
Indeed! As I said above I think we should make the change eventually. But when we move to 1.0 we have a chance to do it without the noise. Unfortunately we can't just change the output of the function without warning as that would break many, many users' existing code.

skimage/util/dtype.py
stefanv: Yes, this is some of the most finicky code in the library, so it is helpful to have a very good sense of what is happening.

michalkrawczyk: First of all: As I see kind_in == 'i' stand for all signed integer types.
Before that signed integer types went all to else block where they get first added 0.5 before normalizing to range [-1.0 ; 1.0].
That for first look seems to be good trick to get off problems of their ranges since e.g int8 has [-128 ; 127] and imax_out will assume as its value 127. 
But in that way all normalized signed integers cannot achieve exact 0 value, which for future calculations will lead to lots of problems and unnecessary lost features. Also note as adding 0.5 will affect more normalization as number of bytes of base int will drop and for int8 it especially hits hard.
I've performed conversion from DirectX documentation. It is still a bit leaky cause -2^bites and -(2^bites - 1) will both be normalized to -1.0, but that won't affect as much and rarely will be noticable

michalkrawczyk: Imagine for example that you got single x = 0 with type int8.
with else block you will first have x = 0.5 since addition
then it will be multiplied to x = 2 and divided to get x = 2/255.
You see that something is wrong, but when you have e.g int64 you'll get 2/9223372036854775807 which is small enough to fool you.

With added elif you get for int8 x = 0/127 and for int64 x = 0/9223372036854775807 which is still x=0.

skimage/filters/_gaussian.py
```diff
@@ -96,6 +98,11 @@ def gaussian(image, sigma=1, output=None, mode='nearest', cval=0,
     >>> filtered_img = gaussian(image, sigma=1, multichannel=True)
 
     """
+    if preserve_range is None:
+        msg = ('The default value of preserve_range will change '
+               'to True in version 1.0')
+        warn(msg, DeprecationWarning)
+        preserve_range = False
```
grlee77, line 105: I think the `change_default_value` decorator utility @rfezzani created can be used to do this. To use it add the following import near the top of the file:
```Python
from .._shared.utils import change_default_value
```

Then rather than changing the value to None or raising the warning here, all you need is a 1 line decorator above the function

```Python
@change_default_value('preserve_range', new_value=True, changed_version='1.0')
```

skimage/filters/tests/test_gaussian.py
grlee77: I will mark this suggestion as resolved since we are not immediately changing the default to True

skimage/util/dtype.py
grlee77: Thanks @michalkrawczyk, this change does indeed look consistent with the Direct X document you linked to and seems reasonable to me.

grlee77 (changes requested):
I requested a minor change to use an existing decorator to handle the deprecation. Otherwise this looks good to me.

skimage/filters/_gaussian.py
michalkrawczyk: @grlee77 Thanks , that looks good. You may check now

By the way I would like to suggest to add this decorator to [Deprecation Cycle Documentation](https://scikit-image.org/docs/dev/contribute.html#deprecation-cycle)

michalkrawczyk:
Module cannot be found: 'skimage.restoration._inpaint' after merge with main

grlee77:
> Module cannot be found: 'skimage.restoration._inpaint' after merge with main

I think this is because we recently added a new Cython function for inpainting. After you merged main, you would have had to rebuild scikit-image in order to avoid this error.

I went ahead and ran this locally and fixed up some issues related to warnings in 528a88f. This is basically just either providing the `preserve_range` keyword or use `expected_warnings` to catch the deprecation warnings so that the test suite can run without raising any of these warnings.

The cosmetic changes in 0fe4242 are just for consistency with the style @rfezzani and I had been using in several other recent PRs.

grlee77 (approved):
Thanks @michalkrawczyk, this looks good to me now

grlee77:
The CI errors seen here are unrelated to this PR (I saw them in a few other cases recently as well)

alexdesiqueira:
Thank you @michalkrawczyk!

jni:
Hi everyone, I'm really sorry about this, but I think we should revert this. I probably should have been more explicit in my comment above: I don't think we should have this in 0.19. I think 1.0 should `preserve_range=True` everywhere *without the keyword*, and we should do that with a release-wide warning, not piecemeal warnings + kwargs everywhere. Again, the Gaussian filter is among the most used functions in scikit-image. I think it's a major mistake to force everyone using it to add a keyword argument to it that should probably be removed eventually anyway.

Since @stefanv has also expressed concerns about some of the deprecations we're making, I think we should probably have a focused meeting soon about our upcoming API choices *and how we will go about making the changes.*

stefanv:
> Since @stefanv has also expressed concerns about some of the deprecations we're making, I think we should probably have a focused meeting soon about our upcoming API choices _and how we will go about making the changes._

Thanks for bringing this up, Juan.  Would you be willing to set up that meeting and write up a paragraph or two about the challenge faced to bring everyone up to speed?

grlee77:
> Thanks for bringing this up, Juan. Would you be willing to set up that meeting and write up a paragraph or two about the challenge faced to bring everyone up to speed?

I just opened #5439 to summarize some of the things I am aware of. Many are relatively easy to implement once we have a decision on the desired direction.

PR #5282, opened 2021-03-20
Stick with usual convention for 3D spatial coords

### LINKED ISSUE #5275: Consider renaming x/y/z to plane/row/col in the 3D structure tensor example
## Description

I didn't find any discussion in #5012 about axis naming, but to me it seems a bit muddled... We normally would call the vertical axis Y and the horizonal axis X, but in [the 3D structure tensor they are inverted](https://scikit-image.org/docs/dev/auto_examples/applications/plot_3d_structure_tensor.html).

Since we are striving to use planes/rows/columns throughout the library, I'm wondering @mkcor @grlee77 @rfezzani whether you would support changing the wording in that example to use that nomenclature. Happy to provide the PR for this if there is consensus.

mkcor:
## Description

Fixes #5275

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_3d_structure_tensor.py
+++ b/doc/examples/applications/plot_3d_structure_tensor.py
@@ -6,8 +6,8 @@ Estimate anisotropy in a 3D microscopy image
 In this tutorial, we compute the structure tensor of a 3D image.
 For a general introduction to 3D image processing, please refer to
 :ref:`sphx_glr_auto_examples_applications_plot_3d_image_processing.py`.
-The data we use here are sampled from an image of kidney tissue by Genevieve
-Buckley in confocal fluorescence microscopy (more details at [1]_ under
+The data we use here are sampled from an image of kidney tissue obtained by
+confocal fluorescence microscopy (more details at [1]_ under
 ``kidney-tissue-fluorescence.tif``).
 
 .. [1] https://gitlab.com/scikit-image/data/#data
--- a/doc/examples/applications/plot_3d_structure_tensor.py
+++ b/doc/examples/applications/plot_3d_structure_tensor.py
@@ -42,7 +42,7 @@ print(f'dtype: {data.dtype}')
 # channel, which leaves us with a 3D single-channel image. What is the range
 # of values?
 
-n_plane, n_row, n_col, n_chan = data.shape
+n_plane, n_Y, n_X, n_chan = data.shape
 v_min, v_max = data[:, :, :, 1].min(), data[:, :, :, 1].max()
 print(f'range: ({v_min}, {v_max})')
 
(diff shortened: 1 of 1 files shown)


doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -53,7 +53,7 @@
     data[n_plane // 2, :, :, 1],
     zmin=v_min,
     zmax=v_max,
-    labels={'x': 'Y', 'y': 'X', 'color': 'intensity'}
+    labels={'x': 'row', 'y': 'col', 'color': 'intensity'}
```
jni, line 56: it's x=col, y=row. =)

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -79,8 +79,8 @@
 #
 #     import plotly.graph_objects as go
 #
-#     (n_Z, n_Y, n_X) = sample.shape
-#     Z, Y, X = np.mgrid[:n_Z, :n_Y, :n_X]
+#     (n_Z, n_X, n_Y) = sample.shape
+#     Z, X, Y = np.mgrid[:n_Z, :n_X, :n_Y]
```
jni, line 83: I think this was ok before?

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -105,15 +105,16 @@
     sample[0, :, :],
     zmin=v_min,
     zmax=v_max,
-    labels={'x': 'Y', 'y': 'X', 'color': 'intensity'},
+    labels={'x': 'row', 'y': 'col', 'color': 'intensity'},
```
jni, line 108: As above: x=col, y=row

jni (changes requested):
@mkcor thank you for the fix! :blush: I think some rows/cols got confused (x=cols and y=rows) but this is almost there!

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: Aww there we go!! It's because x=col and y=row that I loaded the dataset with the following dimension order: [z, y, x, c] = [plane, row, column, channel] :relieved: 

Ok, so I'll revert 58eec1cea6afa66c134cc19097d3ec7c0f39459d (where I introduced the inversion) and proceed with the expected renaming :sweat_smile:

doc/examples/applications/plot_3d_structure_tensor.py
mkcor: Right, reverted with 0da52ca.

mkcor:
@jni thanks for reviewing!

jni (approved):

jni:
Thank you @mkcor! Since this is a doc fix I'm gonna go ahead and merge. If anyone finds further issues we can always update again. 😊

PR #5283, opened 2021-03-22
Fixup test for INSTALL_FROM_SDIST

hmaarrfk:
## Description
The previous test for `INSTALL_FROM_SDIST` always failed due to a bad boolean test.

I guess in bash, whatever non empty string is provided returns true. Therefore, you should explicitely test for truthyness.

```bash
(test) ✔ ~/git/sk/scikit-image [data-files-illuminants ↑·1|✚ 1…1] 
23:28 $ bash test.sh 
false
(test) 23:28 $ INSTALL_FROM_SDIST=0 bash test.sh 
true
(test) ✔ ~/git/sk/scikit-image [data-files-illuminants ↑·1|✚ 1…1] 
23:28 $ INSTALL_FROM_SDIST=1 bash test.sh 
true
(test) ✔ ~/git/sk/scikit-image [data-files-illuminants ↑·1|✚ 1…1] 
23:28 $ cat test.sh 
if [[ $INSTALL_FROM_SDIST ]]; then
    echo true
else
    echo false
fi

```

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -85,7 +85,7 @@ jobs:
             set -ex
             source tools/github/before_install.sh
             python setup.py sdist
-            if [[ $INSTALL_FROM_SDIST ]]; then
+            if [[ $INSTALL_FROM_SDIST != "0" ]]; then
                 pip uninstall cython -y;
                 pip install dist/scikit-image-*.tar.gz;
             else
@@ -144,7 +144,7 @@ jobs:
           # Problem with SimpleITK and py 3.9
           - python-version: 3.9
             OPTIONAL_DEPS: 1
-    
+
     env:
         BUILD_DOCS: 1
         TEST_EXAMPLES: 0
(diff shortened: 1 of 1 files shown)


mkcor (approved):
Thanks for fixing this!

.github/workflows/tests.yml
```diff
@@ -84,8 +84,9 @@ jobs:
         run: |
             set -ex
             source tools/github/before_install.sh
+            set -ex
```
grlee77, line 87: do we need "set -ex" a second time?

grlee77 (approved):
Thanks @hmaarrfk. I had a question about duplicate "set -ex" calls. Is the second one just to make sure it is still set after sourcing the script on the previous line?

.github/workflows/tests.yml
hmaarrfk: becuase the source'd script removes it.... it was tricky to debug should i add a comment?

grlee77: Okay, that is what I suspected. Let's go ahead and get this in as is

hmaarrfk:

Issue #5223, opened 2021-02-05
Add new illuminants for color conversion


BierretA:
## Description

For users of the R programming language, color space conversions can be performed using the standard function grDevices::convertColor, which is part of R-core.
The white points for this function are:
|Name|x|y||X|Y|Z|
|-|-|-|-|-|-|-|
|A|0.44757|0.40745||1.0984661|1.0000000|0.3558228|
|B|0.34842|0.35161||0.9909274|1.0000000|0.8531327|
|C|0.31006|0.31616||0.980706|1.000000|1.182249|
|D50| 0.34574|0.35867||0.9639501|1.0000000|0.8241280|
|D55|0.33250|0.34761)||0.9565317|1.0000000|0.9202555|
|D65|0.3137|0.3291||0.9532057|1.0000000|1.0853844|
|E|1/3|1/3||1|1|1|

White points B and C are not implemented in skimage.
D50, D55 and D65 white points in R do not match the XYZ coordinates of the illuminants in skimage colorconv.py. 

Adding "B" and "C" white points and adding "R" as a new observer for the "D50", "D55" and "D65" illuminants would improve compatibilty with color conversions performed in R.

grlee77:
Hi @BierretA, is this something you are interested in contributing or is this just a feature request? It sounds reasonable, but I am not so familiar with the `color` module, specifically. It would be good if any other core devs herecould chime in on this.

There are also some related projects like [colour](https://colour.readthedocs.io/en/latest/index.html) and `[colorspacious](https://colorspacious.readthedocs.io/en/latest/tutorial.html), that provide some functionality we currently don't have. I am not sure if they have an implementation matching the issue raised here or not, though. There was some past talk in #1991 about how scikit-image could possibly port any unique functionality here to one of those and then use it here as a dependency. I don't think anyone is actively working on that at the moment, though.

BierretA:
I am new to open source but I would be happy to help. If my suggestion is of interest, I can try doing a PR. Adding a few illuminants should be easy enough.

grlee77:
Great, we are happy to help new contributors.

You may want to wait a few days for any other feedback before starting the work.

jni:
I'm in the same boat as @grlee77: don't know much about colorspaces, but very happy to help new contributors! 😊 The feature makes sense, I don't think there'll be much resistance from the other core devs.

FirefoxMetzger:
[...]

Issue #5277, opened 2021-03-20
How to update data registry is not documented yet.


mkcor:
Thank you for investigating @mkcor.  When we're done, it would be good to add a note about this process to the developer guide too.

_Originally posted by @stefanv in https://github.com/scikit-image/scikit-image/issues/5267#issuecomment-802152923_

PR #5308, opened 2021-03-31
Add new illuminants to data registry.

mkcor:
## Description

This PR is a follow-up on #5276, replacing #5234 by breaking it down into sequential PRs which pass CI.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


grlee77 (approved):
This looks good, thanks for helping get everything working with Pooch @mkcor. It looks like @BierretA is still be credited here as well, so that is good.

There is one other suggestion in #5223 about allowing user-supplied observer and illuminants. I think we can merge this PR as is now and close #5223, but then open a new feature request issue just for that idea so it doesn't get lost.

mkcor:
> This looks good, thanks for helping get everything working with Pooch @mkcor.

Sure, then I need to document it (#5277).

> It looks like @BierretA is still be credited here as well, so that is good.

Yes! I cherry-picked what I could from #5234 and also used Git's `Co-authored-by` feature. Sorry, I hadn't ported everything from #5234 when I opened this PR; thanks for reviewing so promptly, @grlee77! It should be good now.

I don't know whether @BierretA would endorse my choice, but I didn't want duplicate files in our data registry: Actually, I caught it for `luv_array_...` but not for `lab_array_...` (noticing right now as I'm typing), but basically the [illuminant, observer] combination [D75, R] is the [same as [D75, 2]](https://github.com/scikit-image/scikit-image/pull/5308/files#diff-a5b79edf6268ee850675946c603a0c03fac862574ab9ac2a661031049ea628e3R441-R443), while the [E, 10] and [E, R] combinations are the [same as [E, 2]](https://github.com/scikit-image/scikit-image/pull/5308/files#diff-a5b79edf6268ee850675946c603a0c03fac862574ab9ac2a661031049ea628e3R444-R446).

> There is one other suggestion in #5223 about allowing user-supplied observer and illuminants. I think we can merge this PR as is now and close #5223, but then open a new feature request issue just for that idea so it doesn't get lost.

Thanks for pointing this out, I didn't have it in mind. Sounds good!

BierretA:
Hello @mkcor ,
Sorry I could not answer faster, I do not have access to my usual computer since last week.

> I don't know whether @BierretA would endorse my choice, but I didn't want duplicate files in our data registry: Actually, I caught it for `luv_array_...` but not for `lab_array_...` (noticing right now as I'm typing), but basically the [illuminant, observer] combination [D75, R] is the [same as [D75, 2]](https://github.com/scikit-image/scikit-image/pull/5308/files#diff-a5b79edf6268ee850675946c603a0c03fac862574ab9ac2a661031049ea628e3R441-R443), while the [E, 10] and [E, R] combinations are the [same as [E, 2]](https://github.com/scikit-image/scikit-image/pull/5308/files#diff-a5b79edf6268ee850675946c603a0c03fac862574ab9ac2a661031049ea628e3R444-R446).

I completely agree with your your modifications, avoiding duplicate files is better. What you did for luv array tests is good, and the same should be done for lab array tests.

mkcor:
Hi @BierretA!

> I completely agree with your modifications, avoiding duplicate files is better. What you did for luv array tests is good, and the same should be done for lab array tests.

Very good, so I'll update the `lab` data files to avoid duplicates, and the data registry and `lab` tests accordingly.

Still, does it make sense to have built-in duplicates, I mean in the illuminants' [definition](https://github.com/scikit-image/scikit-image/pull/5308/files#diff-a5b79edf6268ee850675946c603a0c03fac862574ab9ac2a661031049ea628e3R441-R443)? It looks like we should edit `skimage/color/colorconv.py` as well. I would need to dig into the [R docs](https://www.rdocumentation.org/packages/grDevices/versions/3.6.2/topics/convertColor) and references therein to double-check, but I would be surprised to find out that, say, [D75, R] would be the same as [D75, 2] *by definition*. Would you happen to know, off the top of your head, where we can find the info? Thank you!

mkcor:
Ok, illuminant E gives equal weight to all wavelengths by definition, so its tristimulus values `(X, Y, Z)` are `(1.0, 1.0, 1.0)` regardless of the observer's field of view.

I haven't found a reference for it, but I imagine that, for illuminant D75, the R observer happens to be the standard observer.

mkcor:
Ok, now I'm slightly more familiar with the world of illuminants, I'm at peace with this PR :relaxed:

BierretA:
> Still, does it make sense to have built-in duplicates, I mean in the illuminants' [definition](https://github.com/scikit-image/scikit-image/pull/5308/files#diff-a5b79edf6268ee850675946c603a0c03fac862574ab9ac2a661031049ea628e3R441-R443)? It looks like we should edit `skimage/color/colorconv.py` as well. I would need to dig into the [R docs](https://www.rdocumentation.org/packages/grDevices/versions/3.6.2/topics/convertColor) and references therein to double-check, but I would be surprised to find out that, say, [D75, R] would be the same as [D75, 2] _by definition_. Would you happen to know, off the top of your head, where we can find the info? Thank you!

Illuminants A, B, C and E in R were indeed defined to be the same as the illuminants with a 2° observer. Unfortunately, the information is not clear in the R-documentation can only be found in the source code (it used to be accessible from https://www.rdocumentation.org/packages/grDevices/versions/3.6.2, but it looks like it was recently changed?).

The value for illuminant D75 is missing in R. I added it here to keep the keys as '2', '10', and 'R' for every illuminant (I chose the same value as the 2° observer, like illuminants A/B/C/E). If having different keys only for D75 is not a problem, this should be removed.

mkcor:
Thank you, @BierretA! I sort of wrapped my head around these illuminants and I'm enjoying the learning process.

> The value for illuminant D75 is missing in R. I added it here to keep the keys as '2', '10', and 'R' for every illuminant (I chose the same value as the 2° observer, like illuminants A/B/C/E). If having different keys only for D75 is not a problem, this should be removed.

Either approach makes sense, but I like it this way. It's a 'natural' extension, I would say.
 
> Illuminants A, B, C and E in R were indeed defined to be the same as the illuminants with a 2° observer.

Ok, good to know: They are defined to be the same in terms of tristimulus values `(X, Y, Z)`, as we can see in `skimage/color/colorconv.py`. But then... shouldn't file `"color/tests/data/lab_array_a_r.npy"` end up being the same as `"color/tests/data/lab_array_a_2.npy"` (and so on)?

Now, I have to look into the failing CI tests... :sweat_smile:

BierretA:
> Ok, good to know: They are defined to be the same in terms of tristimulus values (X, Y, Z), as we can see in skimage/color/colorconv.py. But then... shouldn't file "color/tests/data/lab_array_a_r.npy" end up being the same as "color/tests/data/lab_array_a_2.npy" (and so on)?

Yes. And actually, we can remove them (for illuminants A, B and C) and update the tests accordingly. I made the modifications on my [fork](https://github.com/BierretA/scikit-image/tree/update-data-registry) but I cannot push it directly here?

mkcor:
Hi @BierretA,

Thank you so much for your responsiveness!

> I made the modifications on my [fork](https://github.com/BierretA/scikit-image/tree/update-data-registry) but I cannot push it directly here?

That's because 'here' is my branch, so it's expected that you wouldn't have write (push) permission on it.

I would suggest that we @grlee77 @hmaarrfk merge this current PR first; even if further cleanup can be made, it makes sense and it's formally consistent as is. Then you @BierretA can submit a PR with the additional cleanup.

> Yes. And actually, we can remove them (for illuminants A, B and C) and update the tests accordingly.

Ok, the illuminants are the same by definition, but `"color/tests/data/lab_array_a_r.npy"`, i.e,
```r
array([[[  53.23288,   62.2834 ,   43.65972]],

       [[   0.     ,    0.     ,    0.     ]],

       [[ 100.     ,  -23.54141,  -90.37643]],

       [[  32.30259,   65.6623 , -194.2238 ]],

       [[  46.22836,  -61.89611,   23.96719]]])
```
 is 'different' from `"color/tests/data/lab_array_a_2.npy"`, i.e.,
```r
array([[[  53.23288179,   62.27968785,   43.66165206]],

       [[   0.        ,    0.        ,    0.        ]],

       [[ 100.        ,  -23.54631417,  -90.36903339]],

       [[  32.30258667,   65.65947843, -194.21673374]],

       [[  46.22835703,  -61.89823312,   23.96931025]]])
```
because it uses fewer decimals. In your cleanup PR, would you be able to replace the new R illuminants with their double-precision version? This could also be a third, separate PR, no stress.

PS: Failing CI tests were unrelated and have been fixed in the meantime.

BierretA:
> In your cleanup PR, would you be able to replace the new R illuminants with their double-precision version? This could also be a third, separate PR, no stress.

Sure. I will wait for this PR to be merged and make a new one.

alexdesiqueira:
Thank you @BierretA @mkcor!

PR #5317, opened 2021-04-07
DOC: Attempt at autoformatting docstrings in morphology/*

Carreau:
Numpydoc is specific about the meaning of spaces arround : and a few
other things, while it may look almost identical in sphinx depending on
themes it is semantically different.

I'm attempting to write an docstring autoreformatter thetry to fix
common mistakes, and fix a couple of other things like typo in parameter
names, remove parameters that are still documented, and try to get a
uniform style.

You'll see in the diff
 - add / removal of spaces
 - add / removal of blank lines
 - indent 4 spaces always.
 - Capitalisation of section names
 - removal of `border_value` parameter which is indeed not a function
    argument.

In morphology/* this hasn't found any other issue it can fix, and the
only other non-automatically fixable issues are missing documentation
for `dtype` parameters.

## For reviewers

  later.

### DIFF
--- a/skimage/morphology/_deprecated.py
+++ b/skimage/morphology/_deprecated.py
@@ -37,9 +37,9 @@ def watershed(image, markers=None, connectivity=1, offset=None, mask=None,
     out: ndarray
         A labeled matrix of the same type and shape as markers
 
-    See also
+    See Also
     --------
-    skimage.segmentation.random_walker: random walker segmentation
+    skimage.segmentation.random_walker : random walker segmentation
         A segmentation algorithm based on anisotropic diffusion, usually
         slower than the watershed but with good results on noisy data and
         boundaries with holes.
--- a/skimage/morphology/_skeletonize.py
+++ b/skimage/morphology/_skeletonize.py
@@ -34,7 +34,7 @@ def skeletonize(image, *, method=None):
     skeleton : ndarray
         The thinned image.
 
-    See also
+    See Also
     --------
     medial_axis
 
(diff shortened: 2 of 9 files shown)


sciunto (approved):

sciunto:
Great! Thank you for trying on our source files. The failure is related to #5318

skimage/morphology/convex_hull.py
```diff
@@ -155,6 +155,7 @@ def convex_hull_object(image, *, connectivity=2):
     ----------
     image : (M, N) ndarray
         Binary input image.
+
```
grlee77, line 158: I think this extra space is unexpected?

grlee77 (approved):
This is great, thanks! I commented on one inserted space that looked strange to me. 

For context, I think these were generated with @Carreau's [velin](https://github.com/Carreau/velin)

skimage/morphology/convex_hull.py
grlee77: It does not cause any problem in the rendered doc: https://1635-2014929-gh.circle-artifacts.com/0/doc/build/html/api/skimage.morphology.html#skimage.morphology.convex_hull_object

sciunto: I interpreted it as a convention to split between arguments before * and after. Just a guess that I found elegant.

Carreau: I don't believe there are written rules about this; wether to put blank lines between items varies across projects; currently my auto formatter has some heuristics, in particular if one items description have multiple paragraph (which is the case here for connectivity) it will add one line between all items.

It is quite hard to no modify whitespace as is does Docstring -> AST-Like repr -> back to text, tso initial whitespace information is lost.

Carreau:
I reran it with the `--compact` option to keep blank line to a minimal, this removed the line you pointed to.

PR #5322, opened 2021-04-07
avoid importing or calling functions from ndimage submodules

grlee77:
## Description

This is a maintenance PR to avoid use of submodules of ndimage that are not technically part of the public API (for example, `scipy.ndimage.morphology` could be renamed `scipy.ndimage._morphology` upstream, although in practice I am not sure this will actually happen given how much downstream code it might break)

Only the top-level `scipy.ndimage` module is considered public API:
see: https://docs.scipy.org/doc/scipy/reference/api.html#api-definition
and recent discussion in https://github.com/cupy/cupy/pull/5047#issuecomment-814463488

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/_unsharp_mask.py
+++ b/skimage/filters/_unsharp_mask.py
@@ -1,5 +1,5 @@
 import numpy as np
-from scipy.ndimage.filters import gaussian_filter
+from scipy.ndimage import gaussian_filter
 from skimage import img_as_float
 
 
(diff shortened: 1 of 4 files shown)


sciunto:
good to know!

sciunto (approved):

mkcor (approved):

PR #5338, opened 2021-04-20
run face classification gallery example with a single thread

grlee77:
## Description

The example modified in this PR takes >5 minutes in local testing with the Dask `'threads'` scheduler. Changing to `'single-threaded'` resulted in completion in 46 seconds.

The intention of the PR is to reduce the time required for generating the gallery on CI and hopefully reduce the number of Azure timeouts we have been seeing recently.

If this also works well on CI, should we just remove use of Dask from the example? It seems like it complicates things unnecessarily if we can't actually benefit from multiple workers for this task.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/applications/plot_haar_extraction_selection_classification.py
+++ b/doc/examples/applications/plot_haar_extraction_selection_classification.py
@@ -73,7 +73,7 @@ feature_types = ['type-2-x', 'type-2-y']
 X = delayed(extract_feature_image(img, feature_types) for img in images)
 # Compute the result
 t_start = time()
-X = np.array(X.compute(scheduler='threads'))
+X = np.array(X.compute(scheduler='single-threaded'))
 time_full_feature_comp = time() - t_start
 
 # Label images (100 faces and 100 non-faces)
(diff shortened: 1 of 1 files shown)


grlee77:
On CI, it did not seem to make a big difference. Maybe this is because Azure was already limiting the jobs to a single core? (I see 80 s for the full feature set computation in this PR vs 87 s in a previous PR I looked at.

For some reason, Azure will not show me the log in the cases where a timeout occurred ([example with timeout during gallery generation](https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=5669&view=logs&jobId=5e84a1d3-1354-51e2-40cf-26996967c2ef)), but I would like to know at what point it gets stuck in the times when it fails. In that failure case timeout occurred 30 minutes into the gallery generation, but in some other jobs that have succeeded in the past, it finishes in <13 minutes.

grlee77:
For f72a3cf:
Python3.8 gallery testing completed in 11:44
Python3.7 gallery testing timed out at 55 minutes, but I couldn't see the log to see where this occasional hang occurs.

Now trying to split up the doc/gallery builds and add timeout on individual tasks to see if that helps get logs in case of failure.

grlee77:
I have now moved some CI-related changes discussed above to #5343. This PR now only contains the minimal change to the example script.

jni (approved):
I'm happy with this but at the same time it looks like it's masking a deeper GIL issue... Perhaps we should create an issue about this?

alexdesiqueira:
Thank you Greg!

PR #5343, opened 2021-04-21
enable pip and skimage.data caching on Azure

grlee77:
## Description

This PR enables some caching of downloaded files on Azure as well as some other minor modifications. This was split of from #5338 where the caching behavior was originally verified. That PR will focus only on the change to the face classification example while this one will have the CI changes.

The goal here is to help avoid the somewhat common timeouts that have been seen recently in our Azure CI jobs involving either the documentation or gallery example builds.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -183,7 +183,7 @@ jobs:
 
       # Run example applications
       for f in doc/examples/*/*.py; do
-        $PYTHON "${f}"
+        $PYTHON -W ignore:Matplotlib:UserWarning "${f}"
         if [ $? -ne 0 ]; then
           exit 1
         fi
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,5 +1,9 @@
 # Azure Pipelines configuration file for Continuous Integration
 # for building the package and running the tests under Windows.
+variables:
+  PIP_CACHE_DIR: $(Pipeline.Workspace)/.pip
+  # data cache dir determined via: print(pooch.os_cache('scikit-image'))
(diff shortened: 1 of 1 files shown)


grlee77:
If this looks good to others, we can open an issue to enable similar caching on GitHub Actions.

azure-pipelines.yml
```diff
@@ -48,121 +51,146 @@ jobs:
...
+      displayName: Cache pip packages
+
+    - bash: |
+        set -ex
+        PYTHON="$(python.pythonLocation)\\python.exe"
+
+        # Update pip
+        $PYTHON -m pip install -U pip setuptools wheel
+        # print out the pip cache dir followed by the variable defined above
+	# to confirm that they match
```
jni, line 84:
```suggestion
	    # to confirm that they match
```

jni (commented):
@grlee77 the diff is pretty huge, can you point out the relevant bits? I expect you didn't actually change every line?

grlee77:
> @grlee77 the diff is pretty huge, can you point out the relevant bits? I expect you didn't actually change every line?

If you look at the first three commits individually it is much smaller. 

The last commit is only indentation and can be dropped (at one point I saved in VS Code while using an Azure yml validator plugin and it made these changes, but they are not truly necessary).

jni (approved):

jni:
Great, thanks, that was much easier! :sweat_smile: 

There's one minor indentation fix remaining, otherwise LGTM. I'll merge tomorrow if there's no further comments.

grlee77:
I modified the last commit and the indentation should be consistent now. I think if you look at the Azure CI this time around you will see things being retrieved from the caches.

grlee77:
Thanks for reviewing @jni. I went ahead and merged to see if we end up getting fewer CI failures in practice

PR #5344, opened 2021-04-21
single precision support in moments functions

### LINKED ISSUE #5199: Most skimage.measure._moments module functions don't support float32 operations

grlee77:
## Description
closes #5199
This PR continues the work started in #5200.

The first commit here is a cherry-pick of the `_supported_float_dtype` function introduced in #5219. We can rebase here once that PR is merged.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/utils.py
+++ b/skimage/_shared/utils.py
@@ -3,6 +3,7 @@ import functools
 import numbers
 import sys
 import warnings
+from collections.abc import Iterable
 
 import numpy as np
 from numpy.lib import NumpyVersion
@@ -579,3 +580,47 @@ def _fix_ndimage_mode(mode):
     if NumpyVersion(scipy.__version__) >= '1.6.0':
         mode = grid_modes.get(mode, mode)
     return mode
+
+
+new_float_type = {
+    # preserved types
+    np.float32().dtype.char: np.float32,
+    np.float64().dtype.char: np.float64,
+    np.complex64().dtype.char: np.complex64,
+    np.complex128().dtype.char: np.complex128,
+    # altered types
+    np.float16().dtype.char: np.float32,
+    'g': np.float64,      # np.float128 ; doesn't exist on windows
+    'G': np.complex128,   # np.complex256 ; doesn't exist on windows
+}
+
+
+def _supported_float_type(input_dtype, allow_complex=False):
+    """Return an appropriate floating-point dtype for a given dtype.
+
+    float32, float64, complex64, complex128 are preserved.
+    float16 is promoted to float32.
+    complex256 is demoted to complex128.
+    Other types are cast to float64.
+
+    Paramters
+    ---------
+    input_dtype : np.dtype or Iterable of np.dtype
+        The input dtype. If a sequence of multiple dtypes is provided, each
+        dtype is first converted to a supported floating point type and the
+        final dtype is then determined by applying `np.result_type` on the
+        sequence of supported floating point types.
+    allow_complex : bool, optional
+        If False, raise a ValueError on complex-valued inputs.
+
+    Retruns
+    -------
+    float_type : dtype
+        Floating-point dtype for the image.
+    """
+    if isinstance(input_dtype, Iterable) and not isinstance(input_dtype, str):
+        return np.result_type(*(_supported_float_type(d) for d in input_dtype))
+    input_dtype = np.dtype(input_dtype)
+    if not allow_complex and input_dtype.kind == 'c':
+        raise ValueError("complex valued input is not supported")
+    return new_float_type.get(input_dtype.char, np.float64)
--- a/skimage/measure/_moments.py
+++ b/skimage/measure/_moments.py
@@ -1,5 +1,5 @@
 import numpy as np
-from .._shared.utils import check_nD
+from .._shared.utils import _supported_float_type, check_nD
 from . import _moments_cy
 import itertools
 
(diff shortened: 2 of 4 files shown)


skimage/_shared/utils.py
```diff
@@ -579,3 +580,47 @@ def _fix_ndimage_mode(mode):
...
+def _supported_float_type(input_dtype, allow_complex=False):
+    """Return an appropriate floating-point dtype for a given dtype.
+
+    float32, float64, complex64, complex128 are preserved.
+    float16 is promoted to float32.
+    complex256 is demoted to complex128.
+    Other types are cast to float64.
+
+    Paramters
+    ---------
```
jni, line 607:
```suggestion
    Parameters
    ----------
```

skimage/_shared/utils.py
```diff
@@ -579,3 +580,47 @@ def _fix_ndimage_mode(mode):
...
+    input_dtype : np.dtype or Iterable of np.dtype
+        The input dtype. If a sequence of multiple dtypes is provided, each
+        dtype is first converted to a supported floating point type and the
+        final dtype is then determined by applying `np.result_type` on the
+        sequence of supported floating point types.
+    allow_complex : bool, optional
+        If False, raise a ValueError on complex-valued inputs.
+
+    Retruns
```
jni, line 616:
```suggestion
    Returns
```

jni (approved):
Great, thank you @grlee77! I've suggested a couple of typo fixes, but this is otherwise fine by me. :+1:

grlee77:
I don't think it matters if we merge this one or #5219 first. I will just rebase the other as needed

skimage/measure/tests/test_moments.py
```diff
@@ -10,6 +10,7 @@
 from skimage._shared import testing
 from skimage._shared.testing import (assert_equal, assert_almost_equal,
                                      assert_allclose)
+from skimage._shared.utils import _supported_float_type
```
rfezzani, line 13: Since we use `_supported_float_type` in the tests below to provide the expected output type, may be it should be tested also. What do you think?

skimage/measure/_moments.py
```diff
@@ -240,10 +242,13 @@ def moments_central(image, center=None, order=3, **kwargs):
...
-    calc = image.astype(float)
+    float_dtype = _supported_float_type(image.dtype)
+    calc = image.astype(float_dtype, copy=False)
     for dim, dim_length in enumerate(image.shape):
-        delta = np.arange(dim_length, dtype=float) - center[dim]
-        powers_of_delta = delta[:, np.newaxis] ** np.arange(order + 1)
+        delta = np.arange(dim_length, dtype=float_dtype) - center[dim]
+        powers_of_delta = np.stack(
+            [delta ** c for c in range(order + 1)], axis=-1
+        )
```
rfezzani, line 251: Why this change? :slightly_smiling_face:

rfezzani (commented):
Thank you @grlee77, I just left a suggestion, otherwise it looks good to me :wink:

skimage/measure/_moments.py
grlee77: I was curious which was faster, but it is not very important to change this.

I timed it just now and it seems like new ways is faster for sizes larger than ~256, but the old way is faster for smaller sizes. Either way, I don't think this is a bottleneck, so I am fine with reverting this change if you prefer.


```Python
import numpy as np

def p1(delta, order=3):
    return delta[:, np.newaxis] ** np.arange(order + 1)

def p2(delta, order=3):
    return np.stack([delta[:, np.newaxis] ** o for o in np.arange(order + 1)], axis=-1)

for n in [10, 100, 256, 1000, 10000, 1000000]:
    delta = np.random.randn(n)
    print(f"\nSize {n}:")
    %timeit p1(delta)
    %timeit p2(delta)
```

```

Size 16:
3.07 µs ± 124 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
13.3 µs ± 448 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

Size 64:
7.93 µs ± 293 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
15 µs ± 178 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

Size 256:
18.4 µs ± 264 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
18.6 µs ± 424 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

Size 1024:
65 µs ± 1.06 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
31.9 µs ± 816 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

Size 4096:
250 µs ± 1.95 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
94.1 µs ± 2.88 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

Size 16384:
1 ms ± 11.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
374 µs ± 8.29 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

Size 1048576:
68.7 ms ± 934 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
34.6 ms ± 725 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

skimage/measure/tests/test_moments.py
grlee77: Sure, I have added tests now

skimage/measure/_moments.py
rfezzani: the `order` parameter is usually small and previous implementation is one-line... Unless you think it is less readable, I suggest to keep previous version :wink:

grlee77: okay, the change has been reverted!

grlee77:
@rfezzani, please merge if this looks good to you know

jni:
@grlee77 looks like the last change killed the CI...

grlee77:
> @grlee77 looks like the last change killed the CI...

right, it should be fixed now

rfezzani:
:tada: sorry @grlee77 for being late :confused:! Thank you again :wink:

Issue #4057, opened 2019-07-25
Add feature encoding algorithms (Fisher vector, VLAD, and BoW) 


DavidTorpey:
I was wondering if it would be possible to include the three main popular feature encoding / pooling / quantisation algorithms into scikit-image, namely, Fisher vectors, VLAD, and BoW. They are very popular, time-tested algorithms with many applications.

I am happy to submit this feature as well.

stefanv:
VLAD would complement DAISY well, I think. Are the other features applicable to images? You mention scikit-learn, but this is scikit-image.

grlee77:
I am not very familiar with this area, but are there any patent / IP concerns that could make it difficult to incorporate these into scikit-image?  I believe our policy is not to distribute any patented algorithms so users don't have to worry about IP issues.

We recently removed seam carving for this reason and I thought the reason we didn't have SIFT was also due to patent issues. (Although for SIFT it looks like the patent is from 1999, so it may be expired?).

stefanv:
For SIFT, we have Censure, which is arguably better. But if the patent expired we could look into it.

DavidTorpey:
No, I don't think there are IP or patents concerns.

SIFT is separate to these encoding techniques. These encoding techniques such as VLAD and Fisher vectors simply encode a set of descriptors. These descriptors are left to the user to define, and can be anything from SIFT, SURF, ORB, KAZE, or any other descriptor at all.

DavidTorpey:
The SIFT patent expires in March next year as far as I'm aware. But again, that does not affect the implementation and/or inclusion of the above algorithms in scikit-image.

stefanv:
KAZE looks neat; we should try and get that in as well, if it is not patent encumbered.

stefanv:
Wrt feature encoding, perhaps scikit-learn is indeed the more obvious home. Have you asked them? If they can't or won't, we can think about how to fit it into scikit-image.

DavidTorpey:
Yeah, I don't think KAZE is patented.

Yeah, I've asked them. They would rather not include it there since it doesn't fit into the pipelines paradigm?

I would be happy to submit these algos.

stefanv:
Go for it. We'll need a bit of justification, probably in the form of a gallery example, but otherwise I think it should be fine.

DavidTorpey:
Okay perfect, no problem.

DavidTorpey:
[...]

PR #5349, opened 2021-04-23
Implement Fisher vectors in scikit-image

DavidTorpey:
## Description

Implementation of Fisher vectors as per [this issue](https://github.com/scikit-image/scikit-image/issues/4057).

## For reviewers

  later.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -70,6 +70,10 @@ Function: skimage/_shared/version_requirements.py:is_installed
 Copyright: 2009-2011 Pierre Raybaut
 License: MIT
 
+File: skimage/feature/fisher_vector.py
+Copyright: 2014 2014 Dan Oneata
+License: MIT
+
 
 License: BSD-2-Clause
 
--- a/doc/examples/features_detection/plot_fisher_vector.py
+++ b/doc/examples/features_detection/plot_fisher_vector.py
@@ -21,7 +21,7 @@ Please note that scikit-learn is required to run this example.
 from matplotlib import pyplot as plt
 import numpy as np
 from sklearn.datasets import load_digits
-from sklearn.metrics import classification_report, plot_confusion_matrix
+from sklearn.metrics import classification_report, ConfusionMatrixDisplay
 from sklearn.model_selection import train_test_split
 from sklearn.svm import LinearSVC
 
@@ -68,5 +68,11 @@ predictions = svm.predict(testing_fvs)
 
 print(classification_report(test_targets, predictions))
 
-plot_confusion_matrix(svm, testing_fvs, test_targets)
+ConfusionMatrixDisplay.from_estimator(
+    svm,
+    testing_fvs,
+    test_targets,
+    cmap=plt.cm.Blues,
+)
+
 plt.show()
--- a/doc/examples/features_detection/plot_fisher_vector.py
+++ b/doc/examples/features_detection/plot_fisher_vector.py
@@ -3,7 +3,7 @@
 Fisher vector feature encoding
 ===============================
 
-A Fisher vector is an image feature encoding and quantisation technique that
+A Fisher vector is an image feature encoding and quantization technique that
 can be seen as a soft or probabilistic version of the popular
 bag-of-visual-words or VLAD algorithms. Images are modelled using a visual
 vocabulary which is estimated using a K-mode Gaussian mixture model trained on
--- a/doc/examples/features_detection/plot_fisher_vector.py
+++ b/doc/examples/features_detection/plot_fisher_vector.py
@@ -1,0 +1,72 @@
+"""
+===============================
+Fisher vector feature encoding
+===============================
+
+A Fisher vector is an image feature encoding and quantisation technique that
+can be seen as a soft or probabilistic version of the popular
+bag-of-visual-words or VLAD algorithms. Images are modelled using a visual
+vocabulary which is estimated using a K-mode Gaussian mixture model trained on
+low-level image features such as SIFT or ORB descriptors. The Fisher vector
+itself is a concatenation of the gradients of the Gaussian mixture model (GMM)
+with respect to its parameters - mixture weights, means, and covariance
+matrices.
+
+In this example, we compute Fisher vectors for the digits dataset in
+scikit-learn, and train a classifier on these representations.
+
+Please note that scikit-learn is required to run this example.
+"""
+
+from matplotlib import pyplot as plt
+import numpy as np
+from sklearn.datasets import load_digits
+from sklearn.metrics import classification_report, plot_confusion_matrix
+from sklearn.model_selection import train_test_split
+from sklearn.svm import LinearSVC
+
+from skimage.transform import resize
+from skimage.feature import fisher_vector, ORB, learn_gmm
+
+
+data = load_digits()
+images = data.images
+targets = data.target
+
+# Resize images so that ORB detects interest points for all images
+images = np.array([resize(image, (80, 80)) for image in images])
+
+# Compute ORB descriptors for each image
+descriptors = []
+for image in images:
+    detector_extractor = ORB(n_keypoints=5, harris_k=0.01)
+    detector_extractor.detect_and_extract(image)
+    descriptors.append(detector_extractor.descriptors.astype('float32'))
+
+# Split the data into training and testing subsets
+train_descriptors, test_descriptors, train_targets, test_targets = \
+    train_test_split(descriptors, targets)
+
+# Train a K-mode GMM
+k = 16
+gmm = learn_gmm(train_descriptors, n_modes=k)
+
+# Compute the Fisher vectors
+training_fvs = np.array([
+    fisher_vector(descriptor_mat, gmm)
+    for descriptor_mat in train_descriptors
+])
+
+testing_fvs = np.array([
+    fisher_vector(descriptor_mat, gmm)
+    for descriptor_mat in test_descriptors
+])
+
+svm = LinearSVC().fit(training_fvs, train_targets)
+
+predictions = svm.predict(testing_fvs)
+
+print(classification_report(test_targets, predictions))
+
+plot_confusion_matrix(svm, testing_fvs, test_targets)
+plt.show()
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -3,7 +3,6 @@ sphinx-gallery>=0.11
 numpydoc>=1.5
 sphinx-copybutton
 pytest-runner
-scikit-learn
 matplotlib>=3.5
 dask[array]>=2022.9.2
 pandas>=1.5
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -16,3 +16,4 @@ ipywidgets
 ipykernel
 plotly>=5.10
 kaleido
+scikit-learn>=0.24.0
--- a/skimage/feature/fisher_vector.py
+++ b/skimage/feature/fisher_vector.py
@@ -80,6 +80,7 @@ def learn_gmm(descriptors, *, n_modes=32, gm_args=None):
     Examples
     --------
     >>> import pytest
+    >>> _ = pytest.importorskip('sklearn')
     >>> from skimage.feature import fisher_vector
     >>> rg = np.random.Generator(np.random.PCG64())
     >>> sift_for_images = [rg.standard_normal((10, 128)) for _ in range(10)]
@@ -196,6 +197,7 @@ def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
     Examples
     --------
     >>> import pytest
+    >>> _ = pytest.importorskip('sklearn')
     >>> from skimage.feature import fisher_vector, learn_gmm
     >>> sift_for_images = [np.random.random((10, 128)) for _ in range(10)]
     >>> num_modes = 16
--- a/skimage/feature/fisher_vector.py
+++ b/skimage/feature/fisher_vector.py
@@ -1,0 +1,262 @@
+"""
+fisher_vector.py - Implementation of the Fisher vector encoding algorithm
+
+This module contains the source code for Fisher vector computation. The
+computation is separated into two distinct steps, which are called separately
+by the user, namely:
+
+learn_gmm: Used to estimate the GMM for all vectors/descriptors computed for
+           all examples in the dataset (e.g. estimated using all the SIFT
+           vectors computed for all images in the dataset, or at least a subset
+           of this).
+
+fisher_vector: Used to compute the Fisher vector representation for a
+               single set of descriptors/vector (e.g. the SIFT
+               descriptors for a single image in your dataset, or
+               perhaps a test image).
+
+Reference: Perronnin, F. and Dance, C. Fisher kernels on Visual Vocabularies
+           for Image Categorization, IEEE Conference on Computer Vision and
+           Pattern Recognition, 2007
+
+Origin Author: Dan Oneata (Author of the original implementation for the Fisher
+vector computation using scikit-learn and NumPy. Subsequently ported to
+scikit-image (here) by other authors.)
+"""
+import numpy as np
+
+
+class FisherVectorException(Exception):
+    pass
+
+
+class DescriptorException(FisherVectorException):
+    pass
+
+
+def learn_gmm(descriptors, *, n_modes=32, gm_args=None):
+    """Estimate a Gaussian mixture model (GMM) given a set of descriptors and
+    number of modes (i.e. Gaussians). This function is essentially a wrapper
+    around the scikit-learn implementation of GMM, namely the
+    :func:`sklearn.mixture.GaussianMixture` class.
+
+    Due to the nature of the Fisher vector, the only enforced parameter of the
+    underlying scikit-learn class is the covariance_type, which must be 'diag'.
+
+    There is no simple way to know what value to use for `n_modes` a-priori.
+    Typically, the value is usually one of ``{16, 32, 64, 128}``. One may train
+    a few GMMs and choose the one that maximises the log probability of the
+    GMM, or choose `n_modes` such that the downstream classifier trained on
+    the resultant Fisher vectors has maximal performance.
+
+    Parameters
+    -----------
+    descriptors : np.ndarray (N, M) or list [(N1, M), (N2, M), ...]
+        List of NumPy arrays, or a single NumPy array, of the descriptors
+        used to estimate the GMM. The reason a list of NumPy arrays is
+        permissible is because often when using a Fisher vector encoding,
+        descriptors/vectors are computed separately for each sample/image in
+        the dataset, such as SIFT vectors for each image. If a list if passed
+        in, then each element must be a NumPy array in which the number of
+        rows may differ (e.g. different number of SIFT vector for each image),
+        but the number of columns for each must be the same (i.e. the
+        dimensionality must be the same).
+    n_modes : int
+        The number of modes/Gaussians to estimate during the GMM estimate.
+    gm_args : dict
+        Keyword arguments that can be passed into the underlying scikit-learn
+        :func:`sklearn.mixture.GaussianMixture` class.
+
+    Returns
+    -------
+    gmm : :func:`sklearn.mixture.GaussianMixture`
+        The estimated GMM object, which contains the necessary parameters
+        needed to compute the Fisher vector.
+
+    References
+    -----------
+    .. [1] https://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html
+
+    Examples
+    --------
+    >>> from skimage.feature import fisher_vector
+    >>> rg = np.random.Generator(np.random.PCG64())
+    >>> sift_for_images = [rg.standard_normal((10, 128)) for _ in range(10)]
+    >>> num_modes = 16
+    >>> # Estimate 16-mode GMM with these synthetic SIFT vectors
+    >>> gmm = learn_gmm(sift_for_images, n_modes=num_modes)
+    """
+
+    try:
+        from sklearn.mixture import GaussianMixture
+    except ImportError:
+        raise ImportError(
+            'scikit-learn is not installed. Please ensure it is installed in '
+            'order to use the Fisher vector functionality.'
+        )
+
+    if not isinstance(descriptors, (list, np.ndarray)):
+        raise DescriptorException(
+            'Please ensure descriptors are either a NumPY array, '
+            'or a list of NumPy arrays.'
+        )
+
+    d_mat_1 = descriptors[0]
+    if isinstance(descriptors, list) and not isinstance(d_mat_1, np.ndarray):
+        raise DescriptorException(
+            'Please ensure descriptors are a list of NumPy arrays.'
+        )
+
+    if isinstance(descriptors, list):
+        expected_shape = descriptors[0].shape
+        ranks = [len(e.shape) == len(expected_shape) for e in descriptors]
+        if not all(ranks):
+            raise DescriptorException(
+                'Please ensure all elements of your descriptor list '
+                'are of rank 2.'
+            )
+        dims = [e.shape[1] == descriptors[0].shape[1] for e in descriptors]
+        if not all(dims):
+            raise DescriptorException(
+                'Please ensure all descriptors are of the same dimensionality.'
+            )
+
+    if not isinstance(n_modes, int) or n_modes <= 0:
+        raise FisherVectorException(
+            'Please ensure n_modes is a positive integer.'
+        )
+
+    if gm_args:
+        has_cov_type = 'covariance_type' in gm_args
+        cov_type_not_diag = gm_args['covariance_type'] != 'diag'
+        if has_cov_type and cov_type_not_diag:
+            raise FisherVectorException('Covariance type must be "diag".')
+
+    if isinstance(descriptors, list):
+        descriptors = np.vstack(descriptors)
+
+    if gm_args:
+        has_cov_type = 'covariance_type' in gm_args
+        if has_cov_type:
+            gmm = GaussianMixture(
+                n_components=n_modes, **gm_args
+            )
+        else:
+            gmm = GaussianMixture(
+                n_components=n_modes, covariance_type='diag', **gm_args
+            )
+    else:
+        gmm = GaussianMixture(n_components=n_modes, covariance_type='diag')
+
+    gmm.fit(descriptors)
+
+    return gmm
+
+
+def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
+    """Compute the Fisher vector given some descriptors/vectors,
+    and an associated estimated GMM.
+
+    Parameters
+    -----------
+    descriptors : np.ndarray
+        NumPy array of the descriptors to use to compute the Fisher vector
+        representation for.
+    gmm : sklearn.mixture.GaussianMixture
+        An estimated GMM object, which contains the necessary parameters needed
+        to compute the Fisher vector.
+    improved : bool, default=False
+        Flag denoting whether to compute improved Fisher vectors or not.
+        Improved Fisher vectors are L2 and power normalized. Power
+        normalization is simply f(z) = sign(z) pow(abs(z), alpha) for some
+        0 <= alpha <= 1.
+    alpha : float, default=0.5
+        The parameter for the power normalization step. Ignored if
+        improved=False.
+
+    Returns
+    -------
+    fisher_vector : np.ndarray
+        The computation Fisher vector, which is given by a concatenation of the
+        gradients of a GMM with respect to its parameters (mixture weights,
+        means, and covariance matrices). For D-dimensional input descriptors or
+        vectors, and a K-mode GMM, the Fisher vector dimensionality will be
+        2KD + K. Thus, its dimensionality is invariant to the number of
+        descriptors/vectors.
+
+    References
+    -----------
+    .. [1] Perronnin, F. and Dance, C. Fisher kernels on Visual Vocabularies
+           for Image Categorization, IEEE Conference on Computer Vision and
+           Pattern Recognition, 2007
+    .. [2] Perronnin, F. and Sanchez, J. and Mensink T. Improving the Fisher
+           Kernel for Large-Scale Image Classification, ECCV, 2010
+
+    Examples
+    --------
+    >>> from skimage.feature import fisher_vector, learn_gmm
+    >>> sift_for_images = [np.random.random((10, 128)) for _ in range(10)]
+    >>> num_modes = 16
+    >>> # Estimate 16-mode GMM with these synthetic SIFT vectors
+    >>> gmm = learn_gmm(sift_for_images, n_modes=num_modes)
+    >>> test_image_descriptors = np.random.random((25, 128))
+    >>> # Compute the Fisher vector
+    >>> fv = fisher_vector(test_image_descriptors, gmm)
+    """
+    try:
+        from sklearn.mixture import GaussianMixture
+    except ImportError:
+        raise ImportError(
+            'scikit-learn is not installed. Please ensure it is installed in '
+            'order to use the Fisher vector functionality.'
+        )
+
+    if not isinstance(descriptors, np.ndarray):
+        raise DescriptorException(
+            'Please ensure descriptors is a NumPy array.'
+        )
+
+    if not isinstance(gmm, GaussianMixture):
+        raise FisherVectorException(
+            'Please ensure gmm is a sklearn.mixture.GaussianMixture object.'
+        )
+
+    if improved and not isinstance(alpha, float):
+        raise FisherVectorException(
+            'Please ensure that the alpha parameter is a float.'
+        )
+
+    num_descriptors = len(descriptors)
+
+    mixture_weights = gmm.weights_
+    means = gmm.means_
+    covariances = gmm.covariances_
+
+    posterior_probabilities = gmm.predict_proba(descriptors)
+
+    # Statistics necessary to compute GMM gradients wrt its parameters
+    pp_sum = posterior_probabilities.mean(axis=0, keepdims=True).T
+    pp_x = posterior_probabilities.T.dot(descriptors) / num_descriptors
+    pp_x_2 = posterior_probabilities.T.dot(
+        np.power(descriptors, 2)
+    ) / num_descriptors
+
+    # Compute GMM gradients wrt its parameters
+    d_pi = pp_sum.squeeze() - mixture_weights
+
+    d_mu = (pp_x - pp_sum * means).ravel()
+
+    d_sigma_t1 = pp_sum * np.power(means, 2)
+    d_sigma_t2 = pp_sum * covariances
+    d_sigma_t3 = 2 * pp_x * means
+    d_sigma = (-pp_x_2 - d_sigma_t1 + d_sigma_t2 + d_sigma_t3).ravel()
+
+    # Concatenate GMM gradients to form Fisher vector representation
+    fisher_vector = np.hstack((d_pi, d_mu, d_sigma))
+
+    if improved:
+        fisher_vector = fisher_vector / np.linalg.norm(fisher_vector)
+        fisher_vector = \
+            np.sign(fisher_vector) * np.power(np.abs(fisher_vector), alpha)
+
+    return fisher_vector
--- a/skimage/feature/fisher_vector.py
+++ b/skimage/feature/fisher_vector.py
@@ -159,7 +159,7 @@ def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
 
     Parameters
     -----------
-    descriptors : np.ndarray
+    descriptors : np.ndarray, shape=(n_descriptors, descriptor_length)
         NumPy array of the descriptors to use to compute the Fisher vector
         representation for.
     gmm : sklearn.mixture.GaussianMixture
@@ -244,19 +244,25 @@ def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
     # Compute GMM gradients wrt its parameters
     d_pi = pp_sum.squeeze() - mixture_weights
 
-    d_mu = (pp_x - pp_sum * means).ravel()
+    d_mu = pp_x - pp_sum * means
 
     d_sigma_t1 = pp_sum * np.power(means, 2)
     d_sigma_t2 = pp_sum * covariances
     d_sigma_t3 = 2 * pp_x * means
-    d_sigma = (-pp_x_2 - d_sigma_t1 + d_sigma_t2 + d_sigma_t3).ravel()
+    d_sigma = -pp_x_2 - d_sigma_t1 + d_sigma_t2 + d_sigma_t3
+
+    # Apply analytical diagonal normalization
+    sqrt_mixture_weights = np.sqrt(mixture_weights)
+    d_pi /= sqrt_mixture_weights
+    d_mu /= np.expand_dims(sqrt_mixture_weights, -1) * covariances
+    d_sigma /= np.expand_dims(np.sqrt(2 * mixture_weights), -1) * covariances ** 2
 
     # Concatenate GMM gradients to form Fisher vector representation
-    fisher_vector = np.hstack((d_pi, d_mu, d_sigma))
+    fisher_vector = np.hstack((d_pi, d_mu.ravel(), d_sigma.ravel()))
 
     if improved:
-        fisher_vector = fisher_vector / np.linalg.norm(fisher_vector)
         fisher_vector = \
             np.sign(fisher_vector) * np.power(np.abs(fisher_vector), alpha)
+        fisher_vector = fisher_vector / np.linalg.norm(fisher_vector)
 
     return fisher_vector
--- a/skimage/feature/fisher_vector.py
+++ b/skimage/feature/fisher_vector.py
@@ -160,8 +160,8 @@ def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
     Parameters
     -----------
     descriptors : np.ndarray, shape=(n_descriptors, descriptor_length)
-        NumPy array of the descriptors to use to compute the Fisher vector
-        representation for.
+        NumPy array of the descriptors for which the Fisher vector
+        representation is to be computed.
     gmm : sklearn.mixture.GaussianMixture
         An estimated GMM object, which contains the necessary parameters needed
         to compute the Fisher vector.
@@ -254,8 +254,8 @@ def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
     # Apply analytical diagonal normalization
     sqrt_mixture_weights = np.sqrt(mixture_weights)
     d_pi /= sqrt_mixture_weights
-    d_mu /= np.expand_dims(sqrt_mixture_weights, -1) * covariances
-    d_sigma /= np.expand_dims(np.sqrt(2 * mixture_weights), -1) * covariances ** 2
+    d_mu /= sqrt_mixture_weights[:, np.newaxis] * np.sqrt(covariances)
+    d_sigma /= np.sqrt(2) * sqrt_mixture_weights[:, np.newaxis] * covariances
 
     # Concatenate GMM gradients to form Fisher vector representation
     fisher_vector = np.hstack((d_pi, d_mu.ravel(), d_sigma.ravel()))
--- a/skimage/feature/fisher_vector.py
+++ b/skimage/feature/fisher_vector.py
@@ -80,8 +80,6 @@ def learn_gmm(descriptors, *, n_modes=32, gm_args=None):
     Examples
     --------
     >>> import pytest
-    >>> pytest.importorskip('sklearn')
-
     >>> from skimage.feature import fisher_vector
     >>> rg = np.random.Generator(np.random.PCG64())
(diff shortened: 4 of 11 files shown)


skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,208 @@
...
+           Pattern Recognition, 2007
+
+Origin Author: Dan Oneata (Author of the original implementation for the Fisher
+vector computation using scikit-learn and NumPy. Subsequently ported to
+scikit-image (here) by other authors.)
+"""
+import typing
+
+import numpy as np
+from sklearn.mixture import GaussianMixture
```
grlee77, line 29: I think this `GaussianMixture` import would need to be inside of `learn_gmm` so that we can have scikit-learn as an optional rather than required dependency.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,208 @@
...
+
+Reference: Perronnin, F. and Dance, C. Fisher kernels on Visual Vocabularies
+           for Image Categorization, IEEE Conference on Computer Vision and
+           Pattern Recognition, 2007
+
+Origin Author: Dan Oneata (Author of the original implementation for the Fisher
+vector computation using scikit-learn and NumPy. Subsequently ported to
+scikit-image (here) by other authors.)
+"""
+import typing
```
grlee77, line 26: We don't currently have typing in skimage. I know there is some interest, but I haven't followed the discussion too closely (see #4794, #3153) and don't think we have fully sorted out the details of types to use for images, etc.

grlee77 (commented):
I haven't reviewed this properly yet, just a couple of initial comments.

skimage/feature/fisher_vector.py
grlee77: prior discussion: https://github.com/scikit-image/scikit-image/issues/4057#issuecomment-515235811

DavidTorpey: Will do.

Also, CI is failing with the following error:
`ValueError: expected parenthesized list: '-learn( (>= 0.18.1))'`

I'm not sure how to fix that as I think setup.py in this case is expecting module names (i.e. the dash in scikit-learn is not being accepted)? @grlee77

requirements/default.txt
```diff
@@ -6,3 +6,4 @@ pillow>=5.4.0,!=7.1.0,!=7.1.1
 imageio>=2.4.1
 tifffile>=2019.7.26
 PyWavelets>=1.1.1
+scikit-learn(>=0.18.1)
```
grlee77, line 9: Pretty sure this should just be:
`scikit-learn>=0.18.1`

but it should be in `requirements/optional.txt` instead of this file

I think the package name must match the name of the PyPI package and the `-` in the name should not be a problem (for example we have a package `sphinx-gallery` in our requirements/docs.txt)

skimage/feature/fisher_vector.py
grlee77: I added a comment about this there. If you make that change of moving it to requirements/optional.txt, you will have to avoid this top-level import here for CI to pass since only some CI runs install the optional dependencies.

requirements/default.txt
DavidTorpey: I originally had it like that, but it was failing with the same error, so I tried this.

I'll put it in `requirements/optional.txt` anyway.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,203 @@
...
+compute_fisher_vector: Used to compute the Fisher vector representation for a
+                       single set of descriptors/vector (e.g. the SIFT
+                       descriptors for a single image in your dataset, or
+                       perhaps a test image).
+
+Reference: Perronnin, F. and Dance, C. Fisher kernels on Visual Vocabularies
+           for Image Categorization, IEEE Conference on Computer Vision and
+           Pattern Recognition, 2007
+
+Origin Author: Dan Oneata (Author of the original implementation for the Fisher
```
stefanv, line 22: What is the license on that code?  We should include it in LICENSE.txt

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,203 @@
...
+    >>> from skimage.feature import fisher_vector
+    >>> sift_for_images = [np.random.random((10, 128)) for _ in range(10)]
+    >>> num_modes = 16
+    >>> # Estimate 16-mode GMM with these synthetic SIFT vectors
+    >>> gmm = fisher_vector.learn_gmm(sift_for_images, num_modes)
+    >>> test_image_descriptors = np.random.random((25, 128))
+    >>> # Compute the Fisher vector
+    >>> fv = fisher_vector.compute_fisher_vector(test_image_descriptors, gmm)
+    """
+    from sklearn.mixture import GaussianMixture
```
stefanv, line 163: Since sklearn is an optional dependency (should be added as such in `requirements/optional.txt`), we should catch this importerror, print a sensible explanation, and then probably re-raise.

stefanv:
Thanks for this contribution, David!

skimage/feature/fisher_vector.py
DavidTorpey: It's the MIT licence. I have spoken with the Dan Oneata and he has given permission to use his original implementation.

stefanv: OK, that's good.  You can see how we've dealt with external BSD licenses here: https://github.com/scikit-image/scikit-image/blob/main/LICENSE.txt#L60

requirements/default.txt
DavidTorpey: @grlee77  Do you have any suggestions on how to prevent CI from failing since sklearn is now an optional import? Even if I catch and reraise the import error, this won't help with CI failing.

skimage/feature/tests/test_fisher_vector.py
```diff
@@ -0,0 +1,169 @@
+import pytest
+
+import numpy as np
+
+from skimage.feature.fisher_vector import (
+    learn_gmm, compute_fisher_vector, FisherVectorException,
+    DescriptorException
+)
+
```
grlee77, line 9: In order to skip everything in this file when scikit-learn is not installed, you could add

`pytest.importorskip('sklearn')` at the top of this file. Then pytest will mark all of these tests as skipped instead of failures on environments where scikit-learn has not been installed.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,216 @@
...
+    >>> sift_for_images = [np.random.random((10, 128)) for _ in range(10)]
+    >>> num_modes = 16
+    >>> # Estimate 16-mode GMM with these synthetic SIFT vectors
+    >>> gmm = fisher_vector.learn_gmm(sift_for_images, num_modes)
+    """
+
+    try:
+        from sklearn.mixture import GaussianMixture
+    except ImportError:
+        print(
```
grlee77, line 82:
```suggestion
        raise ImportError(
```
This should still raise an error. We can just skip the tests for the CI jobs where optional dependencies are not installed. See [pytest.importorskip](https://docs.pytest.org/en/stable/reference.html#pytest-importorskip).

DavidTorpey: Added

skimage/feature/fisher_vector.py
DavidTorpey: @grlee77 Done

skimage/feature/fisher_vector.py
DavidTorpey: @grlee77 Removed

requirements/default.txt
DavidTorpey: @grlee77 Done. I moved to optional dependencies and added the pytest.importorskip

skimage/feature/fisher_vector.py
DavidTorpey: @stefanv Done. I moved to optional dependencies and printed a sensible explanation when the import fails

skimage/feature/tests/test_fisher_vector.py
DavidTorpey: @grlee77 Done as per above

skimage/feature/fisher_vector.py
DavidTorpey: Done. I moved to optional dependencies and printed a sensible explanation when the import fails

DavidTorpey:
@grlee77 @stefanv I've added an example to the docs, and fixed everything else (I think!). Thanks

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+
+class FisherVectorException(Exception):
+    pass
+
+
+class DescriptorException(FisherVectorException):
+    pass
+
+
+def learn_gmm(descriptors, num_modes, **kwargs):
```
stefanv, line 37: I presume this function is used only internall?  If so, rename to `_learn_gmm`.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+    """Function to estimate a GMM given a set of descriptors and number of
+    modes (i.e. Gaussians). This function is essentially a wrapper around
+    the scikit-learn implementation of GMM, namely the
+    sklearn.mixture.GaussianMixture class.
+
+    Due to the nature of the Fisher vector, the only enforce parameter of the
```
stefanv, line 43: "the only enforce" -- enforceable?

stefanv: It is not clear what this means.  Does this refer to whatever goes into kwargs?  I would also name `**kwargs` more explicitly: `**gm_args` or something like that.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+
+class DescriptorException(FisherVectorException):
+    pass
+
+
+def learn_gmm(descriptors, num_modes, **kwargs):
+    """Function to estimate a GMM given a set of descriptors and number of
+    modes (i.e. Gaussians). This function is essentially a wrapper around
+    the scikit-learn implementation of GMM, namely the
+    sklearn.mixture.GaussianMixture class.
```
stefanv, line 41: @scikit-image/core Any idea how to format `sklearn.mixture.GaussianMixture` so that it gets linked to in the docs?  I suspect we don't have this set up.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+
+    Due to the nature of the Fisher vector, the only enforce parameter of the
+    underlying scikit-learn class is the covariance_type, which must be 'diag'.
+
+    Parameters
+    -----------
+    descriptors : np.ndarray or list
```
stefanv, line 48: Can you provide some shape information here?

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+        List of NumPy arrays, or a single NumPy array, of the descriptors
+        to use to estimate the GMM. Reason a list of NumPy arrays is
+        permissible is because often when using a Fisher vector encoding,
+        descriptors/vectors are computed separately for each sample/image in
+        the dataset, such as SIFT vectors for each image.
+    num_modes : int
```
stefanv, line 54: Does this always need to be specified, or is there also a way to make a reasonable guess?

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+        The number of modes/Gaussians to estimate during the GMM estimate.
+    **kwargs : dict
```
stefanv, line 56: gm_args

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+    Parameters
+    -----------
+    descriptors : np.ndarray
+        NumPy array of the descriptors to use to compute the Fisher vector
+        representation for.
+    gmm : sklearn.mixture.GaussianMixture
+        An estimated GMM object, which contains the necessary parameters needed
+        to compute the Fisher vector.
+    improved : bool, default=False
+        Flag denoting when to compute improved Fisher vectors or not. Improved
```
stefanv, line 144:
```suggestion
        Flag denoting whether to compute improved Fisher vectors or not. Improved
```

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+        Fisher vectors are L2 and power normalized.
```
stefanv, line 145: What does power normalized mean?  You can either expand this explanation, or add a Notes section to the docstring.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+def compute_fisher_vector(descriptors, gmm, improved=False):
+    """Function to compute the Fisher vector given some descriptors/vectors,
+    and an associated estimated GMM.
+
+    Parameters
+    -----------
+    descriptors : np.ndarray
+        NumPy array of the descriptors to use to compute the Fisher vector
+        representation for.
+    gmm : sklearn.mixture.GaussianMixture
```
stefanv, line 140: Why is this not computed on behalf of the user?  I.e., why not have a single `fisher_vector` function call that does the whole thing?  Is it common, e.g., to compute the GMM and to re-use it several times?

stefanv (commented):
Thanks @DavidTorpey, I did a more thorough review now.

My main question is around the API: does it make sense to provide the user with a single entry point, say `fisher_vector(s)`, that goes straight from input to desired output?

skimage/feature/fisher_vector.py
grlee77: I think just using
```
:func:`sklearn.mixture.GaussianMixture`
```
should work.

Cross-linking to docs from other projects requires the `sphinx.ext.intersphinx` extension to be enabled in `conf.py`. It looks like we already have it there and also have scikit-learn already present in the intersphinx mapping defined there:
https://github.com/scikit-image/scikit-image/blob/481ea168687f208da4cce9ef307ec635be43ba06/doc/source/conf.py#L387-L397

stefanv: Awesome, thanks Greg.

skimage/feature/fisher_vector.py
DavidTorpey: @stefanv He granted permission over email, and the original GitHub gist has no associated license. He mentioned in the email that we can consider the code under the MIT license.

What would I need to do in this case? Add the full MIT license to the LICENSE.txt file?

requirements/docs.txt
```diff
@@ -16,3 +16,4 @@ myst-parser
 ipywidgets
 plotly>=4.14.0
 kaleido
+scikit-learn>=0.24.0
```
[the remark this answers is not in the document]
DavidTorpey: Only from 0.24.0 is sklearn.metrics.plot_confusion_matrix (used in the docs) available. GaussianMixture (used in the actual Fisher vector implementation) was available prior to that. I could standardise on 0.24.0 if suits?

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,223 @@
...
+        descriptors = np.vstack(descriptors)
+
+    gmm = GaussianMixture(n_components=num_modes, covariance_type='diag')
+
+    gmm.fit(descriptors)
+
+    return gmm
+
+
+def compute_fisher_vector(descriptors, gmm, improved=False):
```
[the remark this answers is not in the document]
DavidTorpey: Will change.

skimage/feature/fisher_vector.py
DavidTorpey: Nope. The standard way to implement is Fisher vector is to separate the GMM computation from the actual Fisher vector computation. The Fisher vector just depends on the derivatives of some generative probability model. Another reason to keep them separate is so you can compute the Fisher vector for new examples/images without needing to re-estimate the GMM.

vl-feat is probably the most popular implementation of the Fisher vector, and it also separates the GMM computation from the Fisher vector computation: https://www.vlfeat.org/overview/encodings.html

skimage/feature/fisher_vector.py
DavidTorpey: I meant "enforced". I mean that the user can choose to override any of the default arguments of the underlying GaussianMixture class, except for the `diag` argument, as this must be set to `diag` for a Fisher vector.

I will change to `**gm_args`

skimage/feature/fisher_vector.py
DavidTorpey: Will do

skimage/feature/fisher_vector.py
DavidTorpey: We could set a reasonable default of 32 or 64, as these are common values. However, if we wanted to make a reasonable guess, we could perhaps run a small grid search over this value and choose the GMM with maximum log prob. However, I'm not sure this would be the best approach, and I'm unaware of any other heuristics for this. Let me know your thoughts

skimage/feature/fisher_vector.py
DavidTorpey: I'll expand in here

skimage/feature/fisher_vector.py
DavidTorpey: It is more common to compute is once and reuse. It's basically quantising all your training descriptors into GMM parameters. You will also need the GMM to compute the Fisher vectors for testing examples/images. It's the way vl-feat's Fisher vector implementation works as well (i.e. the separation of GMM from FV). See my reply [above](https://github.com/scikit-image/scikit-image/pull/5349#discussion_r626755166) for more detail.

requirements/docs.txt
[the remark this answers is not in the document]
stefanv: That works, thanks.

skimage/feature/fisher_vector.py
stefanv: OK, got it.

skimage/feature/fisher_vector.py
stefanv: Yeah, maybe that would be better suited as an example.  Not critical.

stefanv: Although, just adding that idea into the docstring may be helpful already; it gave me an idea of how I would start.

skimage/feature/fisher_vector.py
stefanv: Could you please forward me that email?

And then exactly as you say: add the MIT license with his name to the LICENSE.txt file.  You'll see in that file how to specify which filename it applies to.

skimage/feature/fisher_vector.py
DavidTorpey: I'll add this into the docstring.

skimage/feature/fisher_vector.py
DavidTorpey: Sent the email.

No problem, I'll add that to the LICENSE.txt file.

skimage/feature/fisher_vector.py
DavidTorpey: Where should I add this line to? @stefanv @grlee77 Thanks

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,234 @@
...
+
+class DescriptorException(FisherVectorException):
+    pass
+
+
+def learn_gmm(descriptors, num_modes, **gm_args):
+    """Function to estimate a GMM given a set of descriptors and number of
+    modes (i.e. Gaussians). This function is essentially a wrapper around
+    the scikit-learn implementation of GMM, namely the
+    sklearn.mixture.GaussianMixture class.
```
grlee77, line 41:
```suggestion
    :func:`sklearn.mixture.GaussianMixture` class.
```

This should just work. We can check the docs generated by the CircleCI job to verify this

skimage/feature/fisher_vector.py
grlee77: I made a suggestion. If you accept that we can check if it is working as expected

skimage/feature/fisher_vector.py
DavidTorpey: Great, thank you

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,240 @@
...
+computation is separated into two distinct steps, which are called separately
+by the user, namely:
+
+learn_gmm: Used to estimate the GMM for all vectors/descriptors computed for
+           all examples in the dataset (e.g. estimated using all the SIFT
+           vectors computed for all images in the dataset, or at least a subset
+           of this).
+
+fisher_vector: Used to compute the Fisher vector representation for a
+                       single set of descriptors/vector (e.g. the SIFT
```
stefanv, line 14: Check indentation

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,240 @@
...
+def learn_gmm(descriptors, num_modes, **gm_args):
+    """Function to estimate a GMM given a set of descriptors and number of
+    modes (i.e. Gaussians). This function is essentially a wrapper around
+    the scikit-learn implementation of GMM, namely the
+    :func:`sklearn.mixture.GaussianMixture` class.
+
+    Due to the nature of the Fisher vector, the only enforced parameter of the
+    underlying scikit-learn class is the covariance_type, which must be 'diag'.
+
+    There is no simple way to know what value to set for num_modes a-priori.
```
stefanv, line 46:
```suggestion
    There is no simple way to know what value to use for `num_modes` a-priori.
```

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,240 @@
...
+    Typically values are usually one of {16, 32, 64, 128}. One may train a few
```
stefanv, line 47:
```suggestion
    Typically, the values is one of ``{16, 32, 64, 128}``. One may train a few
```

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,240 @@
...
+    GMMs and choose the one that maximises the log probability of the GMM, or
+    choose num_modes such as the downstream classifier trained on the resultant
```
stefanv, line 49:
```suggestion
    choose `num_modes` such that the downstream classifier trained on the resultant
```

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,240 @@
...
+    Fisher vectors has maximal performance.
+
+    Parameters
+    -----------
+    descriptors : np.ndarray (N, M) or list [(N1, M), (N2, M), ...]
+        List of NumPy arrays, or a single NumPy array, of the descriptors
+        to use to estimate the GMM. Reason a list of NumPy arrays is
```
stefanv, line 56:
```suggestion
        used to estimate the GMM. The reason a list of NumPy arrays is
```

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,240 @@
...
+
+class FisherVectorException(Exception):
+    pass
+
+
+class DescriptorException(FisherVectorException):
+    pass
+
+
+def learn_gmm(descriptors, num_modes, **gm_args):
```
stefanv, line 37: Let's set this to 16 or some reasonable default, and then use Python 3 keyword only arguments:

```
learn_gmm(descriptors, *, num_modes, gm_args)
```

Above, I also suggest that gm_args is made a separate keyword argument so that we can later add other keyword args to this function.

It looks like we use the convention `n_modes` elsewhere, so if you don't mind let's change to that.

stefanv (commented):
I left some last minor suggestion for tweaks, but overall I'm happy with this now.

Aftr those changes, if another team member can do a final look-over, I think we're ready for merge.

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

requirements/default.txt
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/tests/test_fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

requirements/docs.txt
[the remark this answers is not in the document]
DavidTorpey: Done

skimage/feature/fisher_vector.py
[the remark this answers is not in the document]
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,243 @@
...
+        the dataset, such as SIFT vectors for each image. If a list if passed
+        in, then each element must be a NumPy array in which the number of
+        rows may differ (e.g. different number of SIFT vector for each image),
+        but the number of columns for each must be the same (i.e. the
+        dimensionality must be the same).
+    n_modes : int
+        The number of modes/Gaussians to estimate during the GMM estimate.
+    gm_args : dict
+        Keyword arguments that can be passed into the underlying scikit-learn
+        GaussianMixture class.
```
stefanv, line 68:
```suggestion
        :func:`sklearn.mixture.GaussianMixture` class.
```
```suggestion
        GaussianMixture class.
```

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Returns
+    -------
+    gmm : sklearn.mixture.GaussianMixture
```
stefanv, line 72: Not 100% sure if the :func: trick will work here.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,243 @@
...
+        NumPy array of the descriptors to use to compute the Fisher vector
+        representation for.
+    gmm : sklearn.mixture.GaussianMixture
+        An estimated GMM object, which contains the necessary parameters needed
+        to compute the Fisher vector.
+    improved : bool, default=False
+        Flag denoting whether to compute improved Fisher vectors or not.
+        Improved Fisher vectors are L2 and power normalized. Power
+        normalization is simply f(z) = sign(z) pow(abs(z), alpha) for some
+        0 <= alpha <= 1.
```
stefanv, line 162: Where is the alpha specified?

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,243 @@
...
+
+    Returns
+    -------
+    fisher_vector : np.ndarray
+        The computation Fisher vector, which is given by a concatenation of the
+        gradients of a GMM wrt its parameters (mixture weights, means, and
```
stefanv, line 168:
```suggestion
        gradients of a GMM with regards to its parameters (mixture weights, means, and
```
```suggestion
        gradients of a GMM wrt its parameters (mixture weights, means, and
```

stefanv (approved):
One question about the alpha in the normalization, but otherwise good to me.  Approving.

Thank you for this contribution @DavidTorpey!

skimage/feature/fisher_vector.py
DavidTorpey: I'll try it out

skimage/feature/fisher_vector.py
DavidTorpey: The vast majority of the time it is alpha=0.5 (i.e. square root) for Fisher vectors. However, I'll add it as a keyword argument that the user can override a 0.5 default if they wish.

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

DavidTorpey: Yeah, haha no worries I did notice. I actually meant "with respect to".

DavidTorpey: Hey @stefanv Is there anything else I should do for this PR to be merged, or just a waiting game now?

stefanv:
@grlee77 Are you happy to merge?

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,261 @@
...
+    else:
+        gmm = GaussianMixture(n_components=n_modes, covariance_type='diag')
+
+    gmm.fit(descriptors)
+
+    return gmm
+
+
+def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
+    """Function to compute the Fisher vector given some descriptors/vectors,
```
mkcor, line 156:
```suggestion
    """Compute the Fisher vector given some descriptors/vectors,
```

LICENSE.txt
```diff
@@ -79,3 +79,25 @@ skimage/_shared/version_requirements.py:is_installed:
     LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
     OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
     WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+skimage/feature/fisher_vector.py
+
+    Copyright 2014 Dan Oneata
```
mkcor, line 85:
```suggestion
    Copyright (C) 2014 Dan Oneata
```

doc/examples/features_detection/plot_fisher_vector.py
```diff
@@ -0,0 +1,71 @@
...
+
+In this example, we compute Fisher vectors for the digits dataset in
+scikit-learn, and train a classifier on these representations.
+
+Please note that scikit-learn is required to run this example.
+"""
+
+from matplotlib import pyplot as plt
+from sklearn.datasets import load_digits
+import numpy as np
```
mkcor, line 22:
```suggestion
import numpy as np
from sklearn.datasets import load_digits
```

doc/examples/features_detection/plot_fisher_vector.py
```diff
@@ -0,0 +1,71 @@
...
+===============================
+Fisher vector feature encoding
+===============================
+
+A Fisher vector is an image feature encoding and quantisation technique that
+can be seen as a soft or probabilistic version of the popular
+bag-of-visual-words or VLAD algorithms. Images are modelled using a visual
+vocabulary which is estimated using a K-mode Gaussian mixture model trained on
+low-level image features such as SIFT or ORB descriptors. The Fisher vector
+itself is a concatenation of the gradients of the Gaussian mixture model with
```
mkcor, line 11:
```suggestion
itself is a concatenation of the gradients of the Gaussian mixture model (GMM) with
```

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,261 @@
...
+class FisherVectorException(Exception):
+    pass
+
+
+class DescriptorException(FisherVectorException):
+    pass
+
+
+def learn_gmm(descriptors, *, n_modes=32, gm_args=None):
+    """Function to estimate a GMM given a set of descriptors and number of
```
mkcor, line 38:
```suggestion
    """Estimate a Gaussian mixture model (GMM) given a set of descriptors and number of
```

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,261 @@
...
+        needed to compute the Fisher vector.
+
+    References
+    -----------
+    .. [1] https://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html
+
+    Examples
+    --------
+    >>> from skimage.feature import fisher_vector
+    >>> sift_for_images = [np.random.random((10, 128)) for _ in range(10)]
```
mkcor, line 83: Let's check if this is in line with #5357 :eyes:

skimage/feature/tests/test_fisher_vector.py
```diff
@@ -0,0 +1,205 @@
...
+from skimage.feature.fisher_vector import (
+    learn_gmm, fisher_vector, FisherVectorException,
+    DescriptorException
+)
+
+
+def test_gmm_wrong_descriptor_format_1():
+    """
+    Test that DescriptorException is raised when wrong type for descriptions
```
mkcor, line 15:
```suggestion
    """Test that DescriptorException is raised when wrong type for descriptions
```
Reference: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

skimage/feature/tests/test_fisher_vector.py
```diff
@@ -0,0 +1,205 @@
...
+    is passed.
+    """
+
+    with pytest.raises(DescriptorException):
+        learn_gmm('completely wrong test', n_modes=1)
+
+
+def test_gmm_wrong_descriptor_format_2():
+    """
+    Test that DescriptorException is raised when descriptors are of different
```
mkcor, line 25:
```suggestion
    """Test that DescriptorException is raised when descriptors are of different
```

mkcor (commented):
Thank you, @DavidTorpey! I only started reviewing your great PR, will finish later.

skimage/feature/fisher_vector.py
DavidTorpey: Done

LICENSE.txt
DavidTorpey: Done

doc/examples/features_detection/plot_fisher_vector.py
DavidTorpey: Done

doc/examples/features_detection/plot_fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/fisher_vector.py
DavidTorpey: Done

skimage/feature/tests/test_fisher_vector.py
DavidTorpey: Done

skimage/feature/tests/test_fisher_vector.py
DavidTorpey: Done

DavidTorpey:
Hi @mkcor . Thanks for the review. I have attended to your comments.

DavidTorpey:
Hi. Please let me know if there is anything else. Thanks @mkcor @stefanv

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,262 @@
...
+    d_sigma_t3 = 2 * pp_x * means
+    d_sigma = (-pp_x_2 - d_sigma_t1 + d_sigma_t2 + d_sigma_t3).ravel()
+
+    # Concatenate GMM gradients to form Fisher vector representation
+    fisher_vector = np.hstack((d_pi, d_mu, d_sigma))
+
+    if improved:
+        fisher_vector = fisher_vector / np.linalg.norm(fisher_vector)
+        fisher_vector = \
+            np.sign(fisher_vector) * np.power(np.abs(fisher_vector), alpha)
```
grlee77, line 260: There is summary of the algorithm (Algorithm 1) in this overview article by the same group: [Image Classification with the Fisher Vector: Theory and Practice](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwj3weStuprxAhXFG80KHcSCBmUQFnoECAMQAA&url=https%3A%2F%2Fhal.inria.fr%2Fhal-00830491v2%2Fdocument&usg=AOvVaw2jdg_Uf8o3lkJvCYBICysM)

It looks to me like there may be some subtle differences here vs. what is described:
1.) the Fisher vector computations look similar, but I do not see the divisors shown in Eqs. 31-33?
2.) it looks like the normalizations in section 3 of the algorithm are applied in the opposite order?

I have basically no prior expertise with this type of feature detection algorithm, so it may just be my misunderstanding. I see [here that the specific normalization may not matter](https://gist.github.com/danoneata/9927923/#gistcomment-2581199)?. I see from the comments there that there is actually a different reference (Krapac2011) that may correspond more closely to the specific implementation here.

The code for the implementation and the example you provided look good to me. I just don't have a great way of fully judging the correctness here and was curious whether you have tested if the implementation vs. some other reference implementation of the algorithm to see if performance is similar?

grlee77 (commented):
This looks good to me aside from a few remaining small clarifications. It is hard for me to review this any more closely due to lack of expertise in this area.

skimage/feature/fisher_vector.py
DavidTorpey: Yes, you are correct that the implementation as it stands does not include those normalization factors from Sanchez et al. The reason being is that prior to the "improved" normalizations (L2 and power normalization), a final transformation is required for the Fisher vectors using the (inverse of the) FIM (Fisher information matrix) as the transformation matrix to each Fisher vector.

Since computing the inverse of the FIM is often intractable in practise, Sanchez et al. propose 3 approximations for this step. The first is approximate the FIM as the identity matrix, which equates to no normalization (as in the current implementation). The other two are an analytical, closed-form approximation of the FIM, and an empirical estimate of the FIM using all Fisher vectors in the training set. Since the latter empirical approximation of the FIM requires all training Fisher vectors, it is somewhat more cumbersome and less user-friendly to compute for such an implementation in scikit-image.

I have therefore added the analytical approximation normalizing scaling factors to the code (as in Sanchez et al.).

A while back I used both this implementation of the Fisher vector, as well as vlfeat's implementation for some research I was doing. I recall them performing similarly, but I eventually opted for this sklearn implementation as it suitable my Python ecosystem better at the time. I don't have those exact metrics anymore, though.

DavidTorpey: I have also swapped the order of the L2 and power normalization.

skimage/feature/fisher_vector.py
```diff
@@ -0,0 +1,261 @@
...
+
+def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
+    """Function to compute the Fisher vector given some descriptors/vectors,
+    and an associated estimated GMM.
+
+    Parameters
+    -----------
+    descriptors : np.ndarray
+        NumPy array of the descriptors to use to compute the Fisher vector
+        representation for.
```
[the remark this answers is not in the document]
DavidTorpey: Done

doc/examples/features_detection/plot_fisher_vector.py
```diff
@@ -0,0 +1,71 @@
+"""
+===============================
+Fisher vector feature encoding
+===============================
+
+A Fisher vector is an image feature encoding and quantisation technique that
```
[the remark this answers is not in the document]
DavidTorpey: Done

skimage/feature/fisher_vector.py
```diff
@@ -244,19 +244,25 @@ def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
...
 
     # Concatenate GMM gradients to form Fisher vector representation
-    fisher_vector = np.hstack((d_pi, d_mu, d_sigma))
+    fisher_vector = np.hstack((d_pi, d_mu.ravel(), d_sigma.ravel()))
 
     if improved:
-        fisher_vector = fisher_vector / np.linalg.norm(fisher_vector)
         fisher_vector = \
             np.sign(fisher_vector) * np.power(np.abs(fisher_vector), alpha)
+        fisher_vector = fisher_vector / np.linalg.norm(fisher_vector)
```
grlee77, line 266: If I make this change in isolation from the other analytical diagonal normalization above, the accuracy in the example is approximately the same, so this part seems fine either way.

skimage/feature/fisher_vector.py
```diff
@@ -244,19 +244,25 @@ def fisher_vector(descriptors, gmm, *, improved=False, alpha=0.5):
...
     d_sigma_t2 = pp_sum * covariances
     d_sigma_t3 = 2 * pp_x * means
-    d_sigma = (-pp_x_2 - d_sigma_t1 + d_sigma_t2 + d_sigma_t3).ravel()
+    d_sigma = -pp_x_2 - d_sigma_t1 + d_sigma_t2 + d_sigma_t3
+
+    # Apply analytical diagonal normalization
+    sqrt_mixture_weights = np.sqrt(mixture_weights)
+    d_pi /= sqrt_mixture_weights
+    d_mu /= np.expand_dims(sqrt_mixture_weights, -1) * covariances
+    d_sigma /= np.expand_dims(np.sqrt(2 * mixture_weights), -1) * covariances ** 2
```
grlee77, line 258: Thanks. This normalization seems to have broken the performance in the example script. The macro avg accuracy decreased from ~0.72 to ~0.17. 

I think there is an error in that d_mu should be using `sqrt(covariances)` rather than `covariances`
and d_sigma should be using `covariances` instead of `covariances ** 2`.
If I make those changes the accuracy in the demo improves, but only to ~0.63 which is still lower than what it was before.

grlee77:
```suggestion
    sqrt_mixture_weights = np.sqrt(mixture_weights)
    d_pi /= sqrt_mixture_weights
    d_mu /= sqrt_mixture_weights[:, np.newaxis] * np.sqrt(covariances)
    d_sigma /= np.sqrt(2) * sqrt_mixture_weights[:, np.newaxis] * covariances
```

grlee77 (commented):
Thanks for updating this @DavidTorpey. When I tried running the provided example, the result was better prior to the changes in the most recent commit. I think I identified one source of error, but that still did not fully resolve the issue.

stefanv:
@grlee77 @lagru @DavidTorpey It would be sad to have all this good work go to waste. Do any of you have the bandwidth to push through the final changes?

stefanv:
We'd really like to get this into the next release; any chance @DavidTorpey @grlee77 we can make those last tweaks?

skimage/feature/fisher_vector.py
DavidTorpey: Okay great, yeah could also be a function of the data distribution for sure. Thanks

stefanv:
I've rebased the branch on `main` to make it easier to update.

stefanv:
For some reason, the plot example freezes up on my computer. Something to do with threading and sklearn? So, I made a guess on how to fix the example, but we'll see.

stefanv:
We should probably ensure that the example can execute in a few seconds, either way.

DavidTorpey:
@grlee77 @stefanv I think I've addressed everything,  and tests and docs are now passing. 

Please let me know if there's anything else

Thank you!

stefanv:
It's a bit annoying that we have to litter `pytest.importorskip` in the docstrings, but I don't see an easy alternative right now.

Thank you, @DavidTorpey!

stefanv:
:man_dancing: Thanks @DavidTorpey for the feature and @grlee77 for reviewing.

DavidTorpey:
Thanks @stefanv (and @grlee77 ) for the all the assistance in the process!

PR #5350, opened 2021-04-23
DEV: fix CircleCI caching

tupui:
Caching is not properly working as can be seen in the logs. For instance in the [following](https://app.circleci.com/pipelines/github/scikit-image/scikit-image/1638/workflows/c72c0029-071a-4b03-8247-f9d737d34b13/jobs/1740), only the data part is restored.
This is because the list is not a list of paths, but a list of [cascading fallbacks](https://circleci.com/docs/2.0/caching/#example-caching-configuration).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,8 @@ jobs:
       - restore_cache:
           keys:
               - data-v1-{{ checksum "skimage/data/_registry.py" }}
+      - restore_cache:
+          keys:
               - packages-v1-{{ checksum "requirements/all_requirements" }}
       - run:
           name: install dependencies and package


grlee77:
Thanks @tupui! I restarted the CircleCI a second time here and can verify that the [packages cache is now being found](https://app.circleci.com/pipelines/github/scikit-image/scikit-image/1658/workflows/9e9f65d8-b395-4296-86bb-1ef3cbd477cc/jobs/1766)

Also, based on the existing code here, I realized I should probably have also added `skimage/data/_registry.py` to the data cache key in #5343.

grlee77 (approved):
This looks good to me. The Azure failure is a common timeout that will hopefully become less common in the future now that #5343 was merged.

tupui:
Great thanks!

alexdesiqueira:
Thank you @tupui! How about **Test / macos-cp3.7 (pull_request)**, @grlee77; any ideas?

grlee77:
> Thank you @tupui! How about Test / macos-cp3.7 (pull_request), @grlee77; any ideas?

I am guessing that is just a fluke, but can restart just to verify

alexdesiqueira:
You're right @grlee77, it ran nicely now. Thanks!

PR #5354, opened 2021-04-26
single precision support in skimage.filters

grlee77:
## Description

This PR contains the commits from #5353, so that one should be reviewed prior to this (float32 support Hessian functions in skimage.feature were required in the ridge filters implementation here).

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/filters/lpi_filter.py
```diff
@@ -4,15 +4,16 @@
...
-from .._shared.utils import check_nD
+from .._shared.utils import _supported_float_type, check_nD
 
 eps = np.finfo(float).eps
 
 
 def _min_limit(x, val=eps):
-    mask = np.abs(x) < eps
-    x[mask] = np.sign(x[mask]) * eps
+    mask = np.abs(x) < val
```
rfezzani, line 15: :eyes: good catch :clap:

skimage/filters/lpi_filter.py
```diff
@@ -195,7 +200,7 @@ def inverse(data, impulse_response=None, filter_params={}, max_gain=2,
         filt = predefined_filter
 
     F, G = filt._prepare(data)
-    _min_limit(F)
+    _min_limit(F, val=np.finfo(float).eps)
```
rfezzani, line 203:
```suggestion
    _min_limit(F, val=eps)
```
`eps` is already defined as a global variable. But I am not against your writing, in this case I suggest removing the global declaration of `eps`, just for consistency :wink:

skimage/filters/lpi_filter.py
```diff
@@ -238,7 +243,7 @@ def wiener(data, impulse_response=None, filter_params={}, K=0.25,
...
```
rfezzani, line 246: Same here

rfezzani (approved):
Excellent, thank you @grlee77.

grlee77:
Hi @rfezzani, thanks for reviewing this previously. I fixed conflicts and addressed your comment about the global `eps` variable.

Since the last time you reviewed this, I added single precision support to the three `threshold_*` functions that return arrays (see commits 42c9a94 and 5ebcf52).  

Also, it could go in a separate standalone PR if preferred, but there are a couple of minor fixes to prior single precision PRs in fbb1df9 (a dtype, not an image should be passed to `_supported_float_type`).

rfezzani:
Thank you @grlee77, a second approval and we are good to go :wink: !

grlee77:
This is the only one of the single precision PRs that has not yet been merged.  It has one approval now and the changes made here are consistent with the already merged PRs in this area. Given that, I would be inclined to merge it as is rather than ask others to spend substantial time doing additional reviews. Are there any objections to this?

rfezzani:
> ... Are there any objections to this?

Not from my side ^^

PR #5355, opened 2021-04-26
Fix cython warnings

grlee77:
## Description

Each of these three commits addresses a warning raised by Cython. Specifically:

```
warning: skimage\_shared\fast_exp.pyx:7:0: Overriding cdef method with def method.

warning: skimage\feature\_cascade.pyx:737:49: Not all members given for struct 'Detection'

warning: skimage\filters\rank\core_cy_3d.pyx:47:23: Index should be typed for more efficient access
```

The change in the 3D rank filter code likely improves performance by typing the loop variables/indices, but I have not tried to measure this.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/feature/_cascade.pyx
+++ b/skimage/feature/_cascade.pyx
@@ -734,7 +734,6 @@ cdef class Cascade:
 
                     if result:
 
-                        new_detection = Detection()
                         new_detection.r = current_row
                         new_detection.c = current_col
                         new_detection.width = current_width
--- a/skimage/feature/_cascade.pyx
+++ b/skimage/feature/_cascade.pyx
@@ -734,6 +734,7 @@ cdef class Cascade:
 
                     if result:
 
+                        new_detection = Detection()
                         new_detection.r = current_row
                         new_detection.c = current_col
                         new_detection.width = current_width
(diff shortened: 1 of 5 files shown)


skimage/feature/_cascade.pyx
```diff
@@ -734,7 +734,6 @@ cdef class Cascade:
 
                     if result:
 
-                        new_detection = Detection()
```
jni, line 737: Does this mean that the declaration above instantiates it?

jni (approved):
Thank you @grlee77!

skimage/feature/_cascade.pyx
grlee77: I am not sure what it does (I didn't find documentation calling the struct like this in a quick search). I guess taking a look at the generated C could would be needed to see what it actually does

grlee77: I am thinking about reverting this change just to be safe. This occurs within a parallel range (`prange`) loop and the old notation resulted in another temporary struct being marked as thread private in the OpenMP loop.

I did not realize that there is only a single test case using this function, so I am a bit wary that this may break things!

grlee77: When you assign to a variable within a `prange` loop, Cython will automatically mark the variable as private.

grlee77: I have reverted the change to be more confident that multiple threads will not be assigning to a common struct, creating thread conflicts.

alexdesiqueira:
Thank you Greg!

PR #5357, opened 2021-04-26
MAINT/ENH: refactor np.random.x to use np.random.Generator

### LINKED ISSUE #5342: ENH/DOC: remove references to np.random.RandomState in favour of np.random.Generator
### Description

Once NumPy minimal version is set to 1.17 #5340, I propose to do a PR to remove from the documentation/code any reference to `RandomState` and wrong (not best practice) usage of random number generators.

We went through this in SciPy (doc part for now). More on the PR https://github.com/scipy/scipy/pull/13778

### Additional context

Since NumPy 1.17, the [canonical way](https://numpy.org/doc/stable/reference/random/index.html) to create random numbers is:

```python
import numpy as np
from numpy.random import default_rng
rng = default_rng()
```
And If a seed is to be used. User should not use a common value such as 0 or 123456. This is to prevent bias (everyone is using the same values) and ensure sufficient entropy for the generator. More [details](https://numpy.org/doc/stable/reference/random/bit_generators/index.html#seeding-and-entropy). The good practice (from NumPy's doc) is to generate a seed like this:

```python
from numpy.random import SeedSequence
print(SeedSequence().entropy)
```

Then you can use the value like
```python
rng = default_rng(my_seed_generated_and_not_common_value)
```

tupui:
## Description

Closes #5342

Remove usage of `np.random.RandomState` in code and examples in favour of `np.random.Generator`. Only kept in tests as this is still the recommended way.
Also this change all direct calls to `np.random.x` to use a `np.random.Generator` (also recommended).

## For reviewers

  later.


tupui:
@jni I started to work on this. I did a first pass at everything. I pushed so that this can get feedbacks and also to see the CI.

skimage/util/noise.py
```diff
@@ -201,15 +210,14 @@ def random_noise(image, mode='gaussian', seed=None, clip=True, **kwargs):
         out = image.copy()
         p = kwargs['amount']
         q = kwargs['salt_vs_pepper']
-        flipped = _bernoulli(p, image.shape)
-        salted = _bernoulli(q, image.shape)
+        flipped = _bernoulli(p, image.shape, seed)
+        salted = _bernoulli(q, image.shape, seed)
```
grlee77, line 214: _bernoulli expects `rng` rather than seed? I think that is causing some of the test failures

grlee77:
I see that in a couple of places a new argument was added:
1.) added `seed` kwarg to `medial_axis`
2.) added `random_seed` kwarg to `unsupervised_wiener`
I think these should probably be keyword-only arguments (i.e. `,*, seed=None`)

it seems we were not very consistent in naming the argument `seed` vs. `random_seed` across the library (I think I also saw `sample_seed` in one place). Making this more consistent is something to consider separately for the 1.0 API.

skimage/feature/brief.py
```diff
@@ -112,7 +118,7 @@ class BRIEF(DescriptorExtractor):
     """
 
     def __init__(self, descriptor_size=256, patch_size=49,
-                 mode='normal', sigma=1, sample_seed=1):
+                 mode='normal', sigma=1, sample_seed=None):
```
grlee77, line 121: why change the default here?

grlee77: If we leave it `sample_seed=1` I think this would give different random values than with the legacy `RandomState(1)`, but it would at least still be the same values each time it is run?

tupui:
> I see that in a couple of places a new argument was added:
> 
> 1.) added `seed` kwarg to `medial_axis`
> 
> 2.) added `random_seed` kwarg to `unsupervised_wiener`
> 
> I think these should probably be keyword-only arguments (i.e. `,*, seed=None`)
> 
> 
> 
> it seems we were not very consistent in naming the argument `seed` vs. `random_seed` across the library (I think I also saw `sample_seed` in one place). Making this more consistent is something to consider separately for the 1.0 API.
> 
> 

Oh yes sorry i forgot to mention that. I can make them keyword only. Let me know the name of the keyword I should use.

skimage/feature/brief.py
tupui: I would leave the decision to you here as I don't have any background here 😅. I just checked the doctests we're passing here, which is the case. In general it's not a good idea to have seeds set in stones but here it might be a special case.

grlee77: Let's see what others think. I have never used the feature detection module much myself.

jni:
@tupui Gah, looks like something went wrong with a rebase...?

tupui:
Yes, will fix it. I had master which was in front of the changes I wanted to get from the NumPy bump PR.

skimage/_shared/tests/test_utils.py
```diff
@@ -260,6 +260,6 @@ def test_decorated_channel_axis_shape(channel_axis):
     else:
         assert size == x.shape[channel_axis]
 
-
+
```
jni: Please revert this...

skimage/feature/brief.py
```diff
@@ -1,15 +1,14 @@
...
 from scipy.ndimage import gaussian_filter
 
-from .util import (DescriptorExtractor, _mask_border_keypoints,
+from skimage.feature.util import (DescriptorExtractor, _mask_border_keypoints,
                    _prepare_grayscale_input_2D)
 
-from .brief_cy import _brief_loop
-from .._shared.utils import check_nD
+from skimage.feature.brief_cy import _brief_loop
+from skimage._shared.utils import check_nD
```
jni, line 8: Revert these changes, they should be relative imports.

skimage/feature/brief.py
jni: Yeah tbh this would be a breaking change, I agree that it should not have been 1 to begin with but now we are sort of stuck with it. I'd revert it, and leave the decision of what to do with it to a later issue.

skimage/measure/fit.py
```diff
@@ -735,8 +736,8 @@ def ransac(data, model_class, min_samples, residual_threshold,
     >>> x = xc + a * np.cos(t)
     >>> y = yc + b * np.sin(t)
     >>> data = np.column_stack([x, y])
-    >>> np.random.seed(seed=1234)
-    >>> data += np.random.normal(size=data.shape)
+    >>> rng = np.random.default_rng()
+    >>> data += rng.normal(size=data.shape)
```
jni, line 740: Can we perhaps continue setting the seed here (and below) so that the doctest can still be run? In some cases it's unclear whether things are suitably tested without the doctest...!

skimage/util/noise.py
```diff
@@ -5,7 +5,7 @@
 __all__ = ['random_noise']
 
 
-def _bernoulli(p, shape):
+def _bernoulli(p, shape, seed):
```
jni, line 8:
```suggestion
def _bernoulli(p, shape, *, random_state):
```

jni (commented):
Thanks @tupui! I've left some minor comments but generally 👍 on the PR. Should be easy to get this over the line. 🙏

doc/examples/applications/plot_rank_filters.py
```diff
@@ -66,7 +66,8 @@
 from skimage.filters.rank import median
 from skimage.morphology import disk, ball
 
-noise = np.random.random(noisy_image.shape)
+rng = np.random.default_rng()
+noise = rng.random(noisy_image.shape)
```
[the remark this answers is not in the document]
tupui: Haha totally, I also still need to check NumPy's doc from time to time because I am using the wrong methods.

skimage/feature/brief.py
tupui: I reverted to 1 then.

skimage/measure/fit.py
tupui: Do you have a CI job for the doctests?

We had some discussions for SciPy (and still are having some on Astropy) and we came up with a monkeypatching of the line `>>> rng = np.random.default_rng()` to include a seed. To me this is no practical but it works. Here it does not seems that you have much examples which would really need a seed. Hence I would advice not to use any. And if there are things that are tested in the examples which are not in the tests, then we should simply move these to tests. In the end it's double the work for you maintainers to maintain 2 test suites otherwise.

jni: Yes, the doctests are configured run by default. imho the double effort is in the other direction — if we add tests in addition to doctests (which no longer become doctests), then we have to maintain them in two places.

tupui: I was more thinking about moving than duplicating. IMO doctests should only serve to check that examples are still aligned with the implementation.

Anyway, what I can propose is to either put a seed when really necessary and comment with a `DO NOT COPY`. So that you could still use them for testing. For the seed itself, I would not use the same for all examples and I would get it using `np.random.SeedSequence`. Would that be ok?

jni: Yup, that all sounds fine! Sometimes an example usage is close enough to what a test would be that you can kill two birds with one stone — and we do this liberally at skimage. :blush: :grimacing:

tupui:
I fixed the tests. I just have the following issues. For the first 2, I guess I need to update the files? For the last 2 I am not sure why this crashes.

Can you help with that @jni?

I will have a look at the examples now.

```bash
_______________________________ test_3d_vs_fiji ________________________________

        img_f = io.imread(fetch("data/_blobs_3d_fiji_skeleton.tif"))

___________________________ test_unsupervised_wiener ___________________________
    
________________________________ test_localvar _________________________________

passing but no stable with different seeds

_________________________________ test_speckle _________________________________

scikit-image/skimage/util/tests/test_random_noise.py:143: AssertionError

_________________________________ test_poisson _________________________________

scikit-image/skimage/util/tests/test_random_noise.py:154: AssertionError
```

grlee77:
> Can you help with that @jni?

I fixed the test_speckle and test_poisson and regenerated the data for `test_unsupervised_wiener`. For `test_3d_vs_fiji`, I assume the original reference data was generated using Fiji? I didn't try updating that one here.

grlee77:
I think the only failing tests now are the ones related to the change in the sha256 for the 3 files that were updated. In some of the CI logs these are being downloaded from the `main` branch rather than this one, which is causing the mismatch:

for example:
```
Downloading file 'data/_blobs_3d_fiji_skeleton.tif' from 'https://github.com/scikit-image/scikit-image/raw/main/skimage/data/_blobs_3d_fiji_skeleton.tif' to '/home/runner/.cache/scikit-image/main'
```
so it is downloading the old version of the file, but comparing it to the sha256 for the new one

I assume this will just work once we merge this branch.

tupui:
Thanks you @grlee77! This is great. Is there anything else I should do?

skimage/data/_binary_blobs.py
```diff
@@ -30,7 +33,7 @@ def binary_blobs(length=512, blob_size_fraction=0.1, n_dim=2,
     Examples
     --------
     >>> from skimage import data
-    >>> data.binary_blobs(length=5, blob_size_fraction=0.2, seed=1)
+    >>> data.binary_blobs(length=5, blob_size_fraction=0.2)  # doctest: +SKIP
```
hmaarrfk, line 36: Is this really the better way?

hmaarrfk:
i'm curious how default_rng will change between versions of numpy. Is it guaranteed to be the same, or will it vary as things evolve.

tupui:
> i'm curious how default_rng will change between versions of numpy. Is it guaranteed to be the same, or will it vary as things evolve.

`np.random.RandomState` is guaranteed to be the same but has known issues and is only recommended for testing purposes. See [NEP19](https://numpy.org/neps/nep-0019-rng-policy.html). It is unlikely that `default_rng` will change often, but they keep it open so that they could fix bugs or improve the generators. Things that got intractable with the old architecture.

skimage/data/_binary_blobs.py
tupui: In this case I could use a `seed` generated randomly and add `# do not copy` as done in other examples (this is not perfect but we agreed in previous comments that this was an OK compromise). I guess I forgot this one. I can do this if preferred.

hmaarrfk: hmm ok. yeah i would rather people not copy the seed. I see this kind of copy from examples stuff too often.

skimage/data/_binary_blobs.py
```diff
@@ -18,9 +18,12 @@ def binary_blobs(length=512, blob_size_fraction=0.1, n_dim=2,
     volume_fraction : float, default 0.5
         Fraction of image pixels covered by the blobs (where the output is 1).
         Should be in [0, 1].
-    seed : int, optional
-        Seed to initialize the random number generator.
-        If `None`, a random seed from the operating system is used.
+    seed : {None, int, `numpy.random.Generator`}, optional
```
hmaarrfk, line 21: This is a new feature and should be highlighted. I Think we things like `.. versionadded::`

tupui: Indeed, also not clear what to choose as per #5359. And probably I should add it as keyword argument only with `..., *, seed`. What should I put for the version?

skimage/measure/fit.py
tupui: @hmaarrfk we had this discussion. So in the example you noted, I missed to do this. Let me know what you choose so I modify it accordingly if needed.

skimage/data/_binary_blobs.py
grlee77: Yes, please make any added parameters keyword-only. 

Use `0.19` for the versionadded statement.

tupui: ok thanks @grlee77. And for the naming?

grlee77:
I think the only other task aside from the versionadded and keyword-only stuff mentioned above is adding something about this to `doc/release/release_dev.rst`. The new keywords can be mentioned under the "New Features" section. We should probably also mention that seeds will not give identical results to prior releases due to the change in the underlying generator used.

To help avoid conflicts you may need to merge master to this branch first since a lot of entries left over from the 0.18 release were recently removed from that file in #5367.

skimage/data/_binary_blobs.py
grlee77: I don't think we reached any conclusion in #5359 on the preferred name, but I am personally leaning toward `random_state`. Unless there are objections, we could use that for any newly added parameter here. I wouldn't rename any of the pre-existing parameters in this PR. Let's leave that to a potential follow-up.

tupui: To be clear, in this case I did not add the parameter but just the documentation and called behind `default_rng ` instead of `RandomState`. But there are other places where I added the parameter. Correct me if I am wrong, but I should just add versionadded info to the new places but not here.

tupui:
@grlee77 I think this is ready on my side. I addressed the points you listed and merged master.

grlee77 (approved):
Thanks @tupui, it looks good to me.

grlee77:
@hmaarrfk or @jni, can one of you take a quick look at this one again and merge if it seems okay to you? 

The CI failures here look like they are because I updated a few of the test files and pooch is using the version from main instead of the ones from this branch.

doc/release/release_dev.rst
```diff
@@ -15,6 +15,10 @@ https://scikit-image.org
 New Features
 ------------
 
+- Added a new keyword only parameter ``random_state`` to
```
mkcor, line 18:
```suggestion
- Add a new keyword-only parameter ``random_state`` to
```

skimage/data/_binary_blobs.py
```diff
@@ -18,9 +18,12 @@ def binary_blobs(length=512, blob_size_fraction=0.1, n_dim=2,
...
+        If `seed` is None the `numpy.random.Generator` singleton is used.
```
mkcor, line 22:
```suggestion
        If `seed` is None, the `numpy.random.Generator` singleton is used.
```

skimage/draw/_random_shapes.py
```diff
@@ -351,9 +352,13 @@ def random_shapes(image_shape,
         If `True`, allow shapes to overlap.
     num_trials : int, optional
         How often to attempt to fit a shape into the image before skipping it.
-    random_seed : int, optional
-        Seed to initialize the random number generator.
-        If `None`, a random seed from the operating system is used.
+    random_seed : {None, int, `numpy.random.Generator`}, optional
+        If `random_seed` is None the `numpy.random.Generator` singleton is
```
mkcor, line 356:
```suggestion
        If `random_seed` is None, the `numpy.random.Generator` singleton is
```

skimage/feature/brief.py
```diff
@@ -34,7 +33,14 @@ class BRIEF(DescriptorExtractor):
     mode : {'normal', 'uniform'}, optional
         Probability distribution for sampling location of decision pixel-pairs
         around keypoints.
-    sample_seed : int, optional
+    sample_seed : {None, int, `numpy.random.Generator`}, optional
+        If `sample_seed` is None the `numpy.random.Generator` singleton is
```
mkcor, line 37:
```suggestion
        If `sample_seed` is None, the `numpy.random.Generator` singleton is
```

skimage/feature/haar.py
```diff
@@ -257,11 +256,15 @@ def draw_haar_like_feature(image, r, c, width, height,
     max_n_features : int, default=None
         The maximum number of features to be returned.
         By default, all features are returned.
-    random_state : int, RandomState instance or None, optional
-        If int, random_state is the seed used by the random number generator;
-        If RandomState instance, random_state is the random number generator;
-        If None, the random number generator is the RandomState instance used
-        by `np.random`. The random state is used when generating a set of
+    random_state : {None, int, `numpy.random.Generator`}, optional
+        If `random_state` is None the `numpy.random.Generator` singleton is
```
mkcor, line 260:
```suggestion
        If `random_state` is None, the `numpy.random.Generator` singleton is
```

skimage/future/graph/graph_cut.py
```diff
@@ -100,11 +99,15 @@ def cut_normalized(labels, rag, thresh=0.001, num_cuts=10, in_place=True,
         The maximum possible value of an edge in the RAG. This corresponds to
         an edge between identical regions. This is used to put self
         edges in the RAG.
-    random_state : int, RandomState instance or None, optional
-        If int, random_state is the seed used by the random number generator;
-        If RandomState instance, random_state is the random number generator;
-        If None, the random number generator is the RandomState instance used
-        by `np.random`. The random state is used for the starting point
+    random_state : {None, int, `numpy.random.Generator`}, optional
+        If `random_state` is None the `numpy.random.Generator` singleton is
```
mkcor, line 103:
```suggestion
        If `random_state` is None, the `numpy.random.Generator` singleton is
```

skimage/measure/fit.py
```diff
@@ -704,11 +703,13 @@ def ransac(data, model_class, min_samples, residual_threshold,
         where the probability (confidence) is typically set to a high value
         such as 0.99, e is the current fraction of inliers w.r.t. the
         total number of samples, and m is the min_samples value.
-    random_state : int, RandomState instance or None, optional
-        If int, random_state is the seed used by the random number generator;
-        If RandomState instance, random_state is the random number generator;
-        If None, the random number generator is the RandomState instance used
-        by `np.random`.
+    random_state : {None, int, `numpy.random.Generator`}, optional
+        If `random_state` is None the `numpy.random.Generator` singleton is
```
mkcor, line 707:
```suggestion
        If `random_state` is None, the `numpy.random.Generator` singleton is
```

skimage/morphology/_skeletonize.py
```diff
@@ -371,6 +370,15 @@ def medial_axis(image, mask=None, return_distance=False):
         value in `mask` are used for computing the medial axis.
     return_distance : bool, optional
         If true, the distance transform is returned as well as the skeleton.
+    random_state : {None, int, `numpy.random.Generator`}, optional
+        If `random_state` is None the `numpy.random.Generator` singleton is
```
mkcor, line 374:
```suggestion
        If `random_state` is None, the `numpy.random.Generator` singleton is
```

skimage/morphology/tests/test_skeletonize_3d.py
```diff
@@ -175,7 +175,7 @@ def test_two_hole_image():
 
 def test_3d_vs_fiji():
     # generate an image with blobs and compate its skeleton to
-    # the skeleton generated by FIJI
+    # the skeleton generated by FIJI (Plugins>Skeleton->Skeletonize)
```
mkcor, line 178:
```suggestion
    # the skeleton generated by FIJI (Plugins > Skeleton > Skeletonize)
```
or "(Plugins>Skeleton>Skeletonize)" or "(Plugins->Skeleton->Skeletonize)"

skimage/restoration/deconvolution.py
```diff
@@ -163,6 +163,15 @@ def unsupervised_wiener(image, psf, reg=None, user_params=None, is_real=True,
     clip : boolean, optional
        True by default. If true, pixel values of the result above 1 or
        under -1 are thresholded for skimage pipeline compatibility.
+    random_state : {None, int, `numpy.random.Generator`}, optional
+        If `random_state` is None the `numpy.random.Generator` singleton is
```
mkcor, line 167:
```suggestion
        If `random_state` is None, the `numpy.random.Generator` singleton is
```

skimage/segmentation/_quickshift_cy.pyx
```diff
@@ -33,7 +33,14 @@ def _quickshift_cython(np_floats[:, :, ::1] image, np_floats kernel_size,
         Higher means fewer clusters.
     return_tree : bool
         Whether to return the full segmentation hierarchy tree and distances.
-    random_seed : int
+    random_seed : {None, int, `numpy.random.Generator`}, optional
+        If `random_seed` is None the `numpy.random.Generator` singleton
```
mkcor, line 37:
```suggestion
        If `random_seed` is None, the `numpy.random.Generator` singleton
```

skimage/transform/_hough_transform.pyx
```diff
@@ -346,7 +347,13 @@ def _probabilistic_hough_line(cnp.ndarray img, Py_ssize_t threshold,
         Increase the parameter to merge broken lines more aggressively.
     theta : 1D ndarray, dtype=double
         Angles at which to compute the transform, in radians.
-    seed : int, optional
+    seed : {None, int, `numpy.random.Generator`, optional}
+        If `seed` is None the `numpy.random.Generator` singleton is used.
```
mkcor, line 351:
```suggestion
        If `seed` is None, the `numpy.random.Generator` singleton is used.
```

skimage/util/noise.py
```diff
@@ -57,8 +59,15 @@ def random_noise(image, mode='gaussian', seed=None, clip=True, **kwargs):
                       images.
         - 'speckle'   Multiplicative noise using out = image + n*image, where
                       n is Gaussian noise with specified mean & variance.
-    seed : int, optional
-        If provided, this will set the random seed before generating noise,
+    seed : {None, int, `numpy.random.Generator`}, optional
+        If `seed` is None the `numpy.random.Generator` singleton is
```
mkcor, line 63:
```suggestion
        If `seed` is None, the `numpy.random.Generator` singleton is
```

mkcor (commented):
Thank you, @tupui! Would you mind taking care of the (now) conflicting files? Sorry this has been dragging on...

grlee77:
The conflicts are related to other float32 PRs of mine that were merged, so I can help resolve them here.

grlee77:
One issue that came up in resolving conflicts is related to the following lines in unsupervised_wiener:

```Python
    rng = np.random.default_rng(random_state)
```
```Python
        _rand1 = rng.standard_normal(data_spectrum.shape)
        _rand1 = _rand1.astype(float_type, copy=False)
        _rand2 = rng.standard_normal(data_spectrum.shape)
        _rand2 = _rand2.astype(float_type, copy=False)
        excursion = np.sqrt(0.5 / precision) * (_rand1 + 1j * _rand2)
```

Here it should be more efficient to pass a `dtype` argument to `standard_normal`. 
```Python
        _rand1 = rng.standard_normal(data_spectrum.shape, dtype=float_type)
        _rand2 = rng.standard_normal(data_spectrum.shape, dtype=float_type)
        excursion = np.sqrt(0.5 / precision) * (_rand1 + 1j * _rand2)
```

However, this results in a different stream of random values for float32 vs. float64, which breaks comparison to the existing float64 reference data in `test_unsupervised_wiener`.

Here I used `astype` after generating float64 random values so that we could reuse the same reference data, but perhaps we should also store separate reference data for the float32 case instead?

tupui:
Thanks @grlee77 for updating the PR.
Sorry I cannot help myself until end of June. I am away without a computer.

mkcor:
> However, this results in a different stream of random values for float32 vs. float64, which breaks comparison to the existing float64 reference data in `test_unsupervised_wiener`.

@grlee77 ok.

> Here I used astype after generating float64 random values so that we could reuse the same reference data, but perhaps we should also store separate reference data for the float32 case instead?

Can we do this (in a subsequent PR) by parametrizing `test_unsupervised_wiener` with `[np.float32, np.float64]`?

mkcor:
> Can we do this (in a subsequent PR) by parametrizing `test_unsupervised_wiener` with `[np.float32, np.float64]`?

Oh, it looks like `test_unsupervised_wiener` could be parametrized(?) with `[np.float16, np.float32, np.float64]` from (screenshot of failing "linux-cp3.7-optional-deps" tests):
linux_cp37_optnl_deps
? Although I'm not seeing this in the PR...

I understand that the first two tests  (`data/tests/test_data.py::test_download_all_with_pooch` and `morphology/tests/test_skeletonize_3d.py::test_3d_vs_fiji`) fail because of hash mismatches, but what about `test_unsupervised_wiener`?

grlee77:
> I understand that the first two tests (data/tests/test_data.py::test_download_all_with_pooch and morphology/tests/test_skeletonize_3d.py::test_3d_vs_fiji) fail because of hash mismatches, but what about test_unsupervised_wiener?

This failure is for the same reason (the data files used in the tests had to be updated since the random values generated within `unsupervised_wiener` are different after this PR).

mkcor (approved):

tupui:
Thanks everyone for your help! Great community 😃

PR #5358, opened 2021-04-26
Upgrading Pillow version to 6.1.0

alexdesiqueira:
## Description

Dependabot is showing some issues on Pillow < 8.1.1. Maybe upgrading won't hurt?

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -2,7 +2,7 @@ numpy>=1.16.5
 scipy>=1.2.3
 matplotlib>=3.0.3
 networkx>=2.2
-pillow>=5.4.0,!=7.1.0,!=7.1.1
+pillow>=8.1.2
 imageio>=2.4.1
 tifffile>=2019.7.26
 PyWavelets>=1.1.1
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -2,7 +2,7 @@ numpy>=1.16.5
 scipy>=1.2.3
 matplotlib>=3.0.3
 networkx>=2.2
-pillow>=8.1.2
+pillow>=6.1.0,!=7.1.0,!=7.1.1
 imageio>=2.4.1
 tifffile>=2019.7.26
 PyWavelets>=1.1.1
--- a/TODO.txt
+++ b/TODO.txt
@@ -102,7 +102,7 @@ Other (2022)
 ------------
 * Remove conditional import of ``scipy.fft`` in ``skimage._shared.fft`` once
   the minimum supported version of ``scipy`` reaches 1.4.
-* Remove pillow version related warning for CVE when pillow > 7.1.0 in
+* Remove pillow version related warning for CVE when pillow > 8.1.2 in
   `skimage/io/_plugins/pil_plugin.py` and `skimage/io/collection.py`.
 
 Other (2023)
--- a/TODO.txt
+++ b/TODO.txt
@@ -92,9 +92,6 @@ Other (2021)
   vectorized version of ``np.linspace``.
 * Once NumPy is set to >= 1.17, consider removing
   ``skimage.morphology._util._fast_pad``.
-* Remove pillow version related warning for MPO file format in
-  `io._plugins.pil_plugin.imread` when upgrading pillow min version to
-  6.0.0
 * Remove '--verify-repo=none' in tlmgr calls in ``tools/travis/osx_install.sh``,
   when texlive 2020 is available.
 
--- a/skimage/io/_plugins/pil_plugin.py
+++ b/skimage/io/_plugins/pil_plugin.py
@@ -49,11 +49,6 @@ def imread(fname, dtype=None, img_num=None, **kwargs):
             return pil_to_ndarray(im, dtype=dtype, img_num=img_num)
     else:
         im = Image.open(fname)
-        if im.format == 'MPO' and LooseVersion(pil_version) < '6.0.0':
-            warnings.warn("You are trying to read a MPO image. "
-                          "To ensure a good support of this format, "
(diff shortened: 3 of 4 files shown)


alexdesiqueira:
According to the discussion in #5340, where @jni says:
> ... we use NEP-29 as a guide for how long to keep our minimal dependencies. 

This way, we have:

```
❯ nep29 pillow
| version |    date    |
|---------|------------|
|  8.2.0  | 2021-04-01 |
|  8.1.0  | 2021-01-02 |
|  8.0.0  | 2020-10-14 |
|  7.2.0  | 2020-06-30 |
|  7.1.0  | 2020-04-01 |
|  7.0.0  | 2020-01-02 |
|  6.2.0  | 2019-10-01 |
|  6.1.0  | 2019-07-03 |
```

So we could update Pillow to `6.1.0` in July. On the other hand, [Dependabot](https://github.com/scikit-image/scikit-image/security/dependabot/requirements/default.txt/Pillow/open) says:

> Pillow before 8.1.1 allows attackers to cause a denial of service (memory consumption) because the reported size of a contained image is not properly checked for a BLP container, and thus an attempted memory allocation can be very large.

... and other scary warnings. For now, I'll leave this open for discussion, if you'd like. Else, please close it :slightly_smiling_face:

sciunto:
we mainly have PIL as a direct dependency for io collections, and indirectly from imageio. A way to attack this issue is to remove direct PIL imports. THis is something I investigated, with the help of @jni. It appears that we have a piece of code to load series of tiff images and series of animated gif images. Afaik, the main stopper is that imageio doesn't have the capability to read animated gif and we delegate this to PIL. I don't know how much reading animated gif is important. 

I believe this point of view can help in the debate as well.

jni:
Yeah, I've always considered security "above my pay grade". This all just makes me think I was right. 😂 And also about delegating all io to imageio.

At any rate: I think the NEP-29 guideline is just that — a guideline, and it can be overridden in cases where it makes sense. If pillow 8.1.2 is available on conda-forge, Anaconda defaults, etc, I'm happy to bump up the dependency.

Another alternative is to defer our imports of pillow and raise a warning when it is imported if the version predates 8.1.2.

I'm ok with either option, so I'm going to approve this and let the next person decide. ;)

jni (approved):

sciunto:
In the past, we decided to set a warning.

alexdesiqueira:
> Another alternative is to defer our imports of pillow and raise a warning when it is imported if the version predates 8.1.2.

@sciunto:
> In the past, we decided to set a warning.

How would that warning be?

alexdesiqueira:
> Another alternative is to defer our imports of pillow and raise a warning when it is imported if the version predates 8.1.2.

@sciunto:

> In the past, we decided to set a warning.

How would be that warning?

sciunto:
See https://github.com/scikit-image/scikit-image/pull/4861 ;)

requirements/default.txt
```diff
@@ -2,7 +2,7 @@ numpy>=1.16.5
 scipy>=1.2.3
 matplotlib>=3.0.3
 networkx>=2.2
-pillow>=5.4.0,!=7.1.0,!=7.1.1
+pillow>=8.1.2
```
sciunto, line 5: THen, I guess you want to relax the min version

sciunto:
> Maybe we could use imageio's alternative; they would deal with Pillow, then. How does that sound?

Yes, I was reading imageio's doc and didn't see that support.

sciunto (changes requested):
see above for the min req

sciunto:
> > Maybe we could use imageio's alternative; they would deal with Pillow, then. How does that sound?
> 
> Yes, I was reading imageio's doc and didn't see that support. But, if it works, then definitely yes.

We can try this in a different PR.

requirements/default.txt
alexdesiqueira: > see above for the min req
>  THen, I guess you want to relax the min version

Sorry @sciunto, I didn't get it :slightly_smiling_face:

sciunto: Don't you want to set >=6.1.0 or so? otherwise, the warning doesn't make sense anymore.

alexdesiqueira: I expected it to be >=8.1.2, but I modified the warning with the latest I'd like it to be >=8.1.2, but I updated the CVE numbers on the warning to reflect the latest issues. Isn't it the way to do here?
What would you prefer? To set up >=6.1.0?

alexdesiqueira: I expect it to be >=8.1.2, but I updated the CVE numbers on the warning to reflect the latest issues. Isn't it the way to do here?
What would you prefer? To set up >=6.1.0?

sciunto: If you choose > 8.1.2, no need to have the warning, because the users will be forced to upgrade. If we do not want to force the upgrade, then warning and >6.1.0. THat's my understanding.

alexdesiqueira: @sciunto got it. I thought the warning would check if the machine has that version, and would ask to update based on that.
Warning off, then? :slightly_smiling_face:

hmaarrfk:
I feel like security issues are for system integrators to worry about. Problems and bugs come and go (well they come and they come). 

security issues would bump us to bleeding edge.

alexdesiqueira:
@hmaarrfk you have a point. I'll put 6.1.0 in this one, then, and leave the latest warnings; this way we'll get a full version up, at least

sciunto (approved):

sciunto:
Thank you @alexdesiqueira

Issue #1133, opened 2014-09-03
Color conversion code is very slow


stefanv:
As reported by @holtzhau, the skimage color conversion code is approx. 25x slower than that in OpenCV:

``` python
import time
import cv2
from skimage import data, color

image = data.chelsea()

t = time.time()
lab_image = cv2.cvtColor(image, cv2.COLOR_BGR2Lab)
t_cv = time.time() - t

t = time.time()
lab = color.rgb2lab(image)
t_sk = time.time() - t

print('cv:', t_cv)
print('skimage:', t_sk)
print('factor:', t_sk / t_cv)
```

Results:

```
cv: 0.00397992134094
skimage: 0.0965170860291
factor: 24.2510034146
```

jmetz:
I'm getting a factor of only 3 slower than cv2 with your gist code (https://gist.github.com/blink1073/b9a22d7110acad78b625) @blink1073 did some other dependency get improved in the meantime?? 

```
0.0774719715118
0.0082700252533
0.00317692756653
```

blink1073:
It could be that newer versions of numpy are faster at finding the `einsum`.

jmetz:
I'm using Numpy 1.9.2... so fairly recent version, but not bleeding edge. Do you think that's close enough to merit adding your `einsum` version and closing this issue?

blink1073:
I ran it again using Numpy 1.10 on Python 2.7 and got the following:

```
0.0372121334076
0.00720596313477
0.00090503692627
```

I still think it is not work complicating the existing code base.

jni:
> I still think it is not work complicating the existing code base.

@blink1073 can you elaborate on this? Lately I'm getting a bit annoyed at Python performance. I'm ok with staying at say 2x the optimum. But ~10x starts to seem sloppy.

blink1073:
I just mean my implementation was not nearly as straightforward as what we currently have.  I really do think Cython is the right answer here, I just haven't had a chabce to try it.

JDWarner:
I'm fine with @blink1073's Gist approach. It narrows the gap significantly.

Reality check: is the difference between 0.001 and 0.007 seconds actually going to matter for any real-world use case?

I'm all for enhancing performance, but also strongly believe in the mantra that we should spend out effort wisely. If this is essentially never going to be a limiting factor in a computation, instead of getting bogged down here we should focus our time on riper targets in the package.

jni:
[...]

PR #5362, opened 2021-04-28
faster conversion between RGB and HSV colorspaces

grlee77:
## Description

The current code for `rgb2hsv` and `hsv2rgb` is pretty inefficient. This PR uses Cython to accelerate these two functions, while also reducing the peak memory required. Related, to #1133, but this PR only improves two of the functions in the module.

For the included benchmarks, I get ~20 times faster computation and the memory used by for hsv2rgb is several times smaller. The cause of the large memory footprint in the current hsv2rgb implementation is the following lines where several temporary arrays matching the size of the input image are created:
https://github.com/scikit-image/scikit-image/blob/821c7f27f21c5fcf6a881a0640b94ebfd751665c/skimage/color/colorconv.py#L315-L322

## Note
The calls to `reshape(-1, 3)` are used to collapse multiple spatial dimensions into a single spatial axis to loop over in the Cython code. The output of the Cython function is then restored to the original shape.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


grlee77:
I wrote simple CUDA kernel code for all color conversions in [cuCIM](https://github.com/rapidsai/cucim). This is done by JIT compiling elementwise kernels for each color conversion ([example for separate_stains](https://github.com/rapidsai/cucim/blob/600d6ab62d305daa53832b4ac99d1597e20a3644/python/cucim/src/cucim/skimage/color/colorconv.py#L1631-L1648)), giving a large acceleration over the CPU code.

The cases involving HSV were the ones where the relative performance difference was largest, so I have backported a CPU equivalent here for that one.

grlee77:
Actually, these functions are also a reasonable use case for [Pythran](https://pythran.readthedocs.io/en/latest/). I tried this just out of interest and the following is what the Pythran code for `rgb2hsv_inner` that I came up with looks like:

```Python
#pythran export rgb2hsv_inner(float64[:, 3] order (C))
#pythran export rgb2hsv_inner(float32[:, 3] order (C))
def rgb2hsv_inner(rgb):
    hsv = np.empty_like(rgb)
    n = rgb.shape[0]
    for i in range(n):
        minv = rgb[i, :].min()
        maxv = rgb[i, :].max()
        delta = maxv - minv
        if delta == 0.0:
            hsv[i, :2] = 0.0
        else:
            hsv[i, 1] = delta / maxv
            if rgb[i, 0] == maxv:
                hsv[i, 0] = (rgb[i, 1] - rgb[i, 2]) / delta
            elif rgb[i, 1] == maxv:
                hsv[i, 0] = 2.0 + (rgb[i, 2] - rgb[i, 0]) / delta
            elif rgb[i, 2] == maxv:
                hsv[i, 0] = 4.0 + (rgb[i, 0] - rgb[i, 1]) / delta
            hsv[i, 0] /= 6.0
            hsv[i, 0] -= floor(hsv[i, 0])
        hsv[i, 2] = maxv
    return hsv
```

It is pretty similar to the Cython case, but a bit simpler. This Pythran version was actually slightly faster than the Cython code here on small images, but ~25% slower on large ones. I am not sure what is the underlying cause for the difference.

skimage/color/_colorconv.pyx
```diff
@@ -0,0 +1,81 @@
...
+
+
+def rgb2hsv_inner(np_floats[:, ::] rgb):
+    cdef:
+        np_floats[:, ::] hsv = np.empty_like(rgb)
+        np_floats delta, minv, maxv, tmp
+        Py_ssize_t i, n, ch
+
+    n = rgb.shape[0]
+    for i in range(n):
```
rfezzani, line 19: What about `prange` here?

rfezzani (commented):
Thank you @grlee77 :wink:

rfezzani:
Pythran version loops twice over the channels looking for min and max value while in your implementation there is a unique traversal. This may explain the performance gain :wink:

skimage/color/_colorconv.pyx
grlee77: I can try it. There may not be enough computation for it to provide a benefit, though.

skimage/color/colorconv.py
```diff
@@ -225,48 +226,17 @@ def rgb2hsv(rgb):
...
-    out[..., 2] = out_v
+    rgb = _prepare_colorarray(rgb)
 
-    # # remove NaN
-    out[np.isnan(out)] = 0
+    shape = rgb.shape
+    if not rgb.flags.c_contiguous:
+        rgb = np.ascontiguousarray(rgb)
+    rgb = rgb.reshape(-1, 3)
+    hsv = rgb2hsv_inner(rgb).reshape(shape)
```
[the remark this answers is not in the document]
grlee77: yeah, probably so. I could also switch back to `out`, but I think it is a bit clearer to just use `hsv`

grlee77:
> Pythran version loops twice over the channels looking for min and max value while in your implementation there is a unique traversal.

I thought that too, but tried changing it and didn't see much difference

skimage/color/colorconv.py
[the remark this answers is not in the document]
rfezzani: I have no strong preference :wink:

skimage/color/_colorconv.pyx
grlee77: I tried `prange` and did see some improvements by a factor of 2-5 for 10 cores (20 threads) on large images. For some reason the `rgb2hsv` function became much slower than before if setting OMP_NUM_THREADS=1 before running the benchmarks though! That seems odd to me, but I wasn't able to quickly diagnose the source of the problem so I have left it without `prange` for now.

skimage/color/_colorconv.pyx
```diff
@@ -0,0 +1,83 @@
...
+            else:
+                hsv[i, 1] = delta / maxv
+                if rgb[i, 0] == maxv:
+                    hsv[i, 0] = (rgb[i, 1] - rgb[i, 2]) / delta
+                elif rgb[i, 1] == maxv:
+                    hsv[i, 0] = 2.0 + (rgb[i, 2] - rgb[i, 0]) / delta
+                elif rgb[i, 2] == maxv:
+                    hsv[i, 0] = 4.0 + (rgb[i, 0] - rgb[i, 1]) / delta
+                hsv[i, 0] /= 6.0
+                hsv[i, 0] -= floor(<double>hsv[i, 0])
```
rfezzani, line 40: I am not sure that this will work but what about defining a `_floor` function specific to `np_floats`?
```
if np_floats is cnp.float32_t:
    _floor = floorf
else:
    _floor = floor
```
it may avoid this cast to double precision when the input is single precision...

grlee77: I think we would need to a C99 flag (or `language='c++'`) to ensure `floorf` is available. I think that used to be a problem for older MSVC versions, but is probably fine now. 

Cython 0.29.x doesn't define floorf in it's `libc.math`, but the current 3.0 alpha does. That doesn't really matter, though, since we can always just add it ourselves with: 
```Cython
cdef extern from "<math.h>" nogil:
    float floorf(float)
```

skimage/color/_colorconv.pyx
```diff
@@ -0,0 +1,83 @@
...
+
+
+def hsv2rgb_inner(np_floats[:, ::] hsv):
+    cdef:
+        np_floats[:, ::] rgb = np.empty_like(hsv)
+        np_floats f, v, p
+        Py_ssize_t i, n, hi, rem
+
+    n = rgb.shape[0]
+    with nogil:
```
hmaarrfk, line 52: Are you able to add a reference for the implementation of the algorithm here? That would allow future people to continue to optimize it.

grlee77: yes, I have added references in 0fab7ef

jonahpearl:
The skimage version is still pretty slow -- I tried just grabbing OP's suggestion but it doesn't seem any faster. Any suggested fixes? Or is the current skimage version actually a sped-up implementation, and it was even slower before? Thanks!

matthew-brett:

<details>

## Rationale

</details>

PR #5370, opened 2021-05-03
increase cibuildwheel verbosity

leej3:
## Description

Currently cibuildwheel is failing for cpython 3.8. I believe this is due to a failure to resolve a numpy version, deduced by the fact that the build fails when trying to import numpy. The source of the failed import is hard to infer from the CI log though because of the default verbosity of cibuildwheel: increasing this will likely make debugging easier in the future.

## For reviewers

  later.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -48,6 +48,7 @@ jobs:
           CIBW_MANYLINUX_I686_IMAGE: ${{ matrix.cibw_manylinux }}
           CIBW_TEST_REQUIRES: pytest pooch pytest-localserver pytest-faulthandler
           CIBW_TEST_COMMAND: pytest --pyargs skimage
+          CIBW_BUILD_VERBOSITY: 2
       - uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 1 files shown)


jni (approved):
I'm ok with this, but I'll let @grlee77 and/or @hmaarrfk confirm in case there's reasons we wouldn't want this that I'm unaware of...

hmaarrfk:
Should we use the global env variables to avoid repeating the same thing on all jobs?

grlee77:
If I understand @hmaarrfk's suggestion properly, it is just to place once at the top (prior to the start of the `jobs:` section)?

```yml
env:
    CIBW_BUILD_VERBOSITY: 2
```

hmaarrfk:
Yes, I believe that is correct.

hmaarrfk:
```
env:
    CIBW_BUILD_VERBOSITY: 2
    CIBW_TEST_REQUIRES: pytest pooch pytest-localserver pytest-faulthandler
    CIBW_TEST_COMMAND: pytest --pyargs skimage
```

We might as well move a few other stuff up there.

PR #5371, opened 2021-05-03
[WIP] change numpy version in pyproject.toml

leej3:
## Description

Will likely be made redundant by #5340 but some fix for cibuildwheel is
required prior to release for CI to pass for cibuildwheel with cpython 3.8   

## For reviewers

  later.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ requires = [
     # default numpy requirements
     "numpy==1.16.5; python_version=='3.6' and platform_machine!='aarch64' and platform_python_implementation != 'PyPy'",
     "numpy==1.16.5; python_version=='3.7' and platform_machine!='aarch64' and platform_python_implementation != 'PyPy'",
-    "numpy==1.17.3; python_version=='3.8' and platform_machine!='aarch64' and platform_python_implementation != 'PyPy'",
+    "numpy>='1.17.0,!=1.18.0'; python_version=='3.8' and platform_machine!='aarch64' and platform_python_implementation != 'PyPy'",
     "numpy==1.19.3; python_version=='3.9' and platform_python_implementation != 'PyPy'",
 
     # First PyPy versions for which there are numpy wheels


jni:
@leej3 do you have a reference for why this would work? That line appears to have been lifted [straight from scipy](https://github.com/scipy/scipy/blob/6e5c9ee0dbc5e86d7a6901b2296f1c9c1a07177c/pyproject.toml#L31) so I'm reluctant to change it without some justification...

leej3:
>  do you have a reference for why this would work?

No, hopefully #5370 will give a bit more information as to why cibuildwheel is failing for cpython3.8. The failure seems to be somewhat intermittent. Hopefully with some more detailed logging of the failure I can update this more precisely to get the CI to pass.

> so I'm reluctant to change it without some justification

Agreed, that should be the default strategy. That i can tell, they do not use cibuildwheel though, which is where the failure is occurring.

hmaarrfk:
I think the consensus is that we should pin hard here. The pin as provided would allow version 1.19 to be installed

leej3:
Sounds good. I'm just trying to decipher the error. I misunderstood what was triggering the cibuildwheel workflow. I have now triggered it on my fork with more verbosity. It is still not apparent to me what the various stages of cibuildwheel are and why one of the environments is missing numpy for the failing python version. The correct version of numpy (the current hard-pinned version) is being installed into a virtual environment. The cython build environment doesn't seem to have numpy though. If I figure it out I will submit a fix... most likely without using this pull request at all. 

i don't know if CI on my fork is publically accessible:
https://github.com/leej3/scikit-image/runs/2491382299?check_suite_focus=true

hmaarrfk:
It could be a "manylinux2014" thing. But is this happening for aarch only correct?

leej3:
Oh you are correct. So I was seeing numpy installed in preceding 3.8 builds and it is not installed for aarch64. Based on the pyproject.toml it seems the minimum numpy wheel for aarch64 that should be used is 1.19.2, in which case the error would just be solved by explicitly specifying that. As in the following would allow the build to pass :
```
"numpy==1.19.2; python_version=='3.8' and platform_machine=='aarch64'"
```

I didn't realize none of the conditional numpy version specifications are being used.

PR #5372, opened 2021-05-03
improved single precision support in skimage.transform

grlee77:
## Description

The warp, pyramid and radon transform functions all already supported float32, but this PR expands the tests a bit and makes sure float16 gets promoted to float32 rather than failing.

For `integral_image`, I changed it to promote any floating-point types to double precision for better accuracy with `numpy.cumsum`. (As far as I know, [cumsum does not use the higher accuracy pairwise summation that some NumPy sums use](https://github.com/numpy/numpy/issues/9329#issuecomment-311888469)).  We had seen issues with limited precision of integral images in the past with non-local means and had changed the Cython code internal to that function to always use doubles for the integral image computation.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/transform/integral.py
```diff
@@ -27,7 +27,13 @@ def integral_image(image):
            ACM SIGGRAPH Computer Graphics, vol. 18, 1984, pp. 207-212.
 
     """
-    S = image
+    if image.real.dtype.kind == 'f':
```
rfezzani, line 30: I am not sure to understand this change. Is it a bug fix? Otherwise it is a breaking change :confused:

rfezzani (commented):
Thank you @grlee77.

rfezzani:
@grlee77, the type promotion to double in `integral` is opposed to the strategy of preserving input image dtype :confused:

skimage/transform/integral.py
grlee77: It can be fine to keep the precision of the input if the image is not too large, but these cumulative sums can become very inaccurate for large image sizes. The basic issue is that the cumulative sum may be very large and then you are doing addition of relatively small values to very large values which leads to inaccuracy. The same effect is still present in float64, but just won't be as noticeable until larger image sizes are reached. 

There was a concrete example using the integral images implemented internally to non-local means ([see here](https://github.com/scikit-image/scikit-image/pull/2878#issuecomment-346344773) and in the subsequent comment). In that case the error is highly visible in the image for float32 when the input is shape 600x1000x1000, but not visually noticeable for float64.

One compromise to avoid a behavior change would be to do the cumsum in float64, but then cast the final sum to the dtype of the input?

grlee77:
> @grlee77, the type promotion to double in integral is opposed to the strategy of preserving input image dtype

It currently only preserves float dtypes. Apparently np.cumsum already uses int64 for, e.g. uint8 inputs. Otherwise you would be pretty much guaranteed to get integer overflows! I am just arguing that for accuracy sake, we should be promoting to float64. We could still restore the final output to the input float type, though, as a compromise.

grlee77:
Another simple example. Suppose the image is somewhat large, but has float16 dtype. In that case it is also likely to encounter floating point overflow:
```Python
integral_image(np.ones((2048, 2048), dtype=np.float16))
```
gives
```
array([[1.000e+00, 2.000e+00, 3.000e+00, ..., 2.046e+03, 2.047e+03,
        2.048e+03],
       [2.000e+00, 4.000e+00, 6.000e+00, ..., 4.092e+03, 4.094e+03,
        4.096e+03],
       [3.000e+00, 6.000e+00, 9.000e+00, ..., 7.500e+03, 7.504e+03,
        7.508e+03],
       ...,
       [2.046e+03, 4.092e+03, 6.136e+03, ...,       inf,       inf,
              inf],
       [2.047e+03, 4.094e+03, 6.140e+03, ...,       inf,       inf,
              inf],
       [2.048e+03, 4.096e+03, 6.144e+03, ...,       inf,       inf,
              inf]], dtype=float16)
```

rfezzani:
I understand the overflow problem, we can then consider this is a bug fix. May be should we document this behaviour...

skimage/transform/integral.py
grlee77: I have reverted this dtype change for integral_image and will raise a separate PR related to that instead

grlee77:
> I understand the overflow problem, we can then consider this is a bug fix. May be should we document this behaviour...

I reverted the change to default dtype in `integral_image`. We can revisit in a separate PR.

skimage/transform/tests/test_warps.py
```diff
@@ -374,14 +378,17 @@ def test_resize_dtype():
     assert resize(x_f32, (10, 10), preserve_range=True).dtype == x_f32.dtype
 
 
-def test_swirl():
-    image = img_as_float(checkerboard())
+@testing.parametrize('dtype', [np.float32, np.float64])
```
mkcor, line 381:
```suggestion
@testing.parametrize('dtype', [np.float16, np.float32, np.float64])
```
Why not include float16, knowing that it's expected to be promoted to float32?

skimage/transform/tests/test_warps.py
```diff
@@ -620,27 +646,33 @@ def test_log_polar_mapping():
     assert np.allclose(coords, ground_truth)
 
 
-def test_linear_warp_polar():
+@testing.parametrize('dtype', [np.float32, np.float64])
```
mkcor, line 649: Same question here.

skimage/transform/tests/test_warps.py
```diff
@@ -620,27 +646,33 @@ def test_log_polar_mapping():
...
+    image = image.astype(dtype, copy=False)
     warped = warp_polar(image, radius=25)
+    assert warped.dtype == _supported_float_type(dtype)
     profile = warped.mean(axis=0)
     peaks = peak_local_max(profile)
     assert np.alltrue([peak in radii for peak in peaks])
 
 
-def test_log_warp_polar():
+@testing.parametrize('dtype', [np.float32, np.float64])
```
mkcor, line 664: And here.

mkcor (approved):
Thanks, @grlee77! I wasn't sure why float16 wasn't systematically included in the floating-point types when testing with parameters, so please let me know for my personal understanding, but this PR is good to go! :shipit:

skimage/transform/tests/test_warps.py
grlee77: Thanks! I have added float16 to the cases you pointed out.

rfezzani (approved):
Merging :tada: Thank you @grlee77

PR #5374, opened 2021-05-03
add numpy version specification on aarch for cpython 3.8

leej3:
Correct fix cibuildwheel with cpython 3.8 on aarch64 as described in https://github.com/scikit-image/scikit-image/pull/5371

## For reviewers

  later.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ requires = [
     # wheels require fixes contained in numpy 1.19.2
     "numpy==1.19.2; python_version=='3.6' and platform_machine=='aarch64'",
     "numpy==1.19.2; python_version=='3.7' and platform_machine=='aarch64'",
+    "numpy==1.19.2; python_version=='3.8' and platform_machine=='aarch64'",
 
     # default numpy requirements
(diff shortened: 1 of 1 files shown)


grlee77 (approved):
Looks good.

I was initially wondering if 3.9 should also be here, but It seems it is already handled by another case below which does not exclude aarch64.

leej3:
> but It seems it is already handled by another case below which does not exclude aarch64

Yes. I considered using a `>=` but I don't have a sense of which failure is more confusing... not having a numpy version defined or automatically defining a numpy version as new python versions are added to the list.

Unfortunately leej3:cibuildwheel does not trigger the cibuildwheel actions. This passed on my fork though where it does get triggered because there the branch is  just named cibuildwheel.
https://github.com/leej3/scikit-image/actions/runs/806955121

hmaarrfk:
thx

jni:
Does this need backporting to v0.18.x?

hmaarrfk:
yes. THe backport command was removed from the PR requests. So I always forget what it is....

jni:

PR #5382, opened 2021-05-06
Added ND butterworth filter

din14970:
## Description

In high resolution electron microscopy, FFT based filters are commonly used to enhance features in periodic images. One such common filter is a high pass 2D Butterworth filter, which masks low spatial frequencies (usually background) with minimal introduction of artifacts. The Butterworth filter is implemented for 1D signals in scipy but not for 2D arrays. I use scikit-image quite often but was missing this filter, so I hope I can contribute it. For images of the general sort this can be either a low-pass or high-pass filter for spatial frequencies, or it can be combined to create a bandpass filter.

On the astronaut image:
```
butterworth(rgb2gray(astronaut()), cutoff_frequency_ratio = 0.01, high_pass=False)
```
 
image

```
butterworth(rgb2gray(astronaut()), cutoff_frequency_ratio = 0.1, high_pass=True)
```
image

Not sure whether I put it in a good place and whether it requires its own example

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,88 @@
...
+    lsize = np.array([nx, ny]) * factor
+    # start and stop is to ensure center of mask aligns with center of fft
+    gx = np.arange(-(nx - 1) // 2, (nx - 1) // 2 + 1) / lsize[0]
+    gy = np.arange(-(ny - 1) // 2, (ny - 1) // 2 + 1) / lsize[1]
+    qxa, qya = np.meshgrid(gx, gy)
+    q2 = np.sqrt(np.square(qxa) + np.square(qya))
+    wfilt = 1 / np.sqrt(1 + np.power(q2, order))
+    if high_pass:
+        wfilt = 1 - wfilt
+    return wfilt
```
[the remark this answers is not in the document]
din14970: I thought about this but was unsure how to best approach this and what to consider an n-dimensional image. In an RGB image, the channels should each independently be 2D Fourrier transformed and filtered, the channel dimension should not be considered a spatial dimension. But in a gray level 3D image of course the third dimension should be considered. Perhaps then I need a `multichannel` keyword arg. I'll see if I can make a decent implementation.

grlee77:
> I thought about this but was unsure how to best approach this and what to consider an n-dimensional image. In an RGB image, the channels should each independently be 2D Fourrier transformed and filtered, the channel dimension should not be considered a spatial dimension. But in a gray level 3D image of course the third dimension should be considered. Perhaps then I need a multichannel keyword arg. I'll see if I can make a decent implementation.

Historically we have used `multichannel=True` to indicate that the last axis contains channels rather than a spatial dimension. However we are currently in the process of deprecating that and moving to a `channel_axis` keyword argument. This argument should default to `channel_axis=None`, meaning all axes are spatial. It is sufficient to support only `channel_axis=None` and `channel_axis=-1` if the general case would be more of a pain to implement. We can just add a [channels_as_last_axis decorator](https://github.com/scikit-image/scikit-image/blob/main/skimage/_shared/utils.py#L220) to automatically do a temporary shift of channels to the last axis as needed.

din14970:
Ok, was a bit more complicated than I expected to make channel work for any axis so it looks a bit hacky. N dimensional images are supported and any axis index can be passed as channel index. By default it is assumed all axes are spatial dimensions, a `channel_axis` MUST be passed for color images. Any number of channels are supported. While refactoring I also slightly improved the mask generation function. 

For large n-dimensional images the creation of the Fourier mask may pose a memory issue due to the reliance on meshgrid. More memory friendly may be a "loopy" implementation but for this I would need numba as I'm not familiar with cython. Are there any plans to allow numba as a dependency?

din14970:
Also, I feel like a large portion of the `butterworth` filter could be recycled for other types of FFT masks. Perhaps if there is interest in the future this could be pulled out and generalized.

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
+import numpy as np
+import scipy.fft as fft
+from skimage.exposure import rescale_intensity
+from skimage import dtype_limits
```
grlee77, line 4:
```suggestion

from .._shared.fft import fft_module as fft
from ..exposure import rescale_intensity
from ..utils import dtype_limits
```

The change to the fft import is needed until we drop support for SciPy < 1.4 (it falls back to numpy.fft when scipy.fft is unavailable).

The others just changes the style to use relative imports for consistency with the rest of the library

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def _get_ND_butterworth_filter(shape, factor, order, high_pass):
+    """
+    Get a N-dimensional Butterworth mask
```
grlee77, line 9:
```suggestion
    """Create a N-dimensional Butterworth mask.
```

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+    ----------
+    shape : tuple of ints
+        dimensions of the n-dimensional image and mask
+    factor : float
+        fraction of width and height of the filter where the cutoff should be
+    order : float
+        controls the slope in the cutoff region
+    high_pass : bool
+        whether the filter is high pass (low frequencies attenuated) or
+        low pass (high frequencies are attenuated)
```
grlee77, line 21:
```suggestion
    shape : tuple of int
        Shape of the n-dimensional image and mask.
    factor : float
        Fraction of width and height of the filter where the cutoff should be.
    order : float
        Controls the slope in the cutoff region.
    high_pass : bool
        Whether the filter is high pass (low frequencies attenuated) or
        low pass (high frequencies are attenuated).
```

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+
+    Returns
+    -------
+    wfilt : numpy array of shape
+        The filter/fft mask
```
grlee77, line 26:
```suggestion
    wfilt : ndarray
        The filter/fft mask.
```

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    ranges = []
+    for d in shape:
+        # start and stop ensures center of mask aligns with center of fft
+        axis = np.arange(-(d - 1) // 2, (d - 1) // 2 + 1) / (d * factor)
+        ranges.append(axis**2)
+    # q2 = euclidian distance squared grid
```
grlee77, line 33:
```suggestion
    # q2 = squared Euclidean distance grid
```

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
+import numpy as np
```
grlee77, line 1:
```suggestion
import functools

import numpy as np
```

added functools for use below

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+    wfilt : numpy array of shape
+        The filter/fft mask
+    """
+    ranges = []
+    for d in shape:
+        # start and stop ensures center of mask aligns with center of fft
+        axis = np.arange(-(d - 1) // 2, (d - 1) // 2 + 1) / (d * factor)
+        ranges.append(axis**2)
+    # q2 = euclidian distance squared grid
+    q2 = np.sum(np.meshgrid(*ranges, indexing="ij"), axis=0)
```
grlee77, line 34:
```suggestion
    q2 = functools.reduce(
        np.add, np.meshgrid(*ranges, indexing='ij', sparse=True)
    )
```
This is a more efficient way to do the same thing (in terms of both memory usage and computation). This way all arrays returned by meshgrid are size 1 on all but 1 axis and no large array with `len(shape + 1)` dimensions has to be created.

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+    # division of order by 2 to avoid additional square root on q2
+    wfilt = 1 / np.sqrt(1 + np.power(q2, order/2))
+    if high_pass:
+        wfilt = 1 - wfilt
+    return fft.ifftshift(wfilt)
```
grlee77, line 39: I suppose one could gain some efficiency by applying ifftshift to the 1d ranges arrays rather than doing an nd shift here at the end.

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+    Wireless Engineer 7.6 (1930): 536-541.
+    """
+    fft_shape = (image.shape if channel_axis is None
+                 else np.delete(image.shape, channel_axis))
+    wfilt = _get_ND_butterworth_filter(fft_shape, cutoff_frequency_ratio,
+                                       order, high_pass)
+    butterfilt = np.empty(image.shape, dtype=np.float64)
+    if channel_axis is None:
+        im_fft = fft.fftn(image)
+        butterfilt[:] = np.real(fft.ifftn(wfilt * im_fft))
```
grlee77, line 98: We can leave it to a potential later follow-up, but I think we can probably use `rfftn` and `irfftn` for efficiency when the image is real-valued. This would require also modifying the coordinate ranges in the `_get_ND_butterworth_filter` appropriately.

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+    else:
+        abs_channel = channel_axis % image.ndim
+        post = image.ndim - abs_channel - 1
+        for i in range(image.shape[channel_axis]):
+            sl = (slice(None),)*abs_channel + (i,) + (slice(None),)*post
+            im_fft = fft.fftn(image[sl])
+            butterfilt[sl] = np.real(fft.ifftn(wfilt * im_fft))
```
grlee77, line 105: We can also do this without a loop if we just insert a dimension 1 axis in wfilt and then set the `axes` argument of fftn/ifftn to include all axes except `channel_axis`.

grlee77: The current looping way would save some memory by using smaller intermediate arrays, but probably at cost of performance.

skimage/filters/tests/test_fft_based.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+import numpy as np
+from skimage.filters import butterworth
+
+
+def test_butterworth_2D():
+    im = np.zeros((4, 4))
+    filtered = butterworth(im)
+    assert filtered.shape == im.shape
+    assert np.all(im == filtered)
```
grlee77, line 10: Let me think about other possible tests for correctness.

The current tests are testing that things run without crashing for a range of dimensions / channel axes, but aren't really checking any property of the filtered image or comparing to a reference result.

grlee77 (changes requested):
Thanks @din14970. I left a first round of review comments here. A number of just simple style suggestions for the docstrings.

Note that you don't have to commit each suggestion individually, but can add them all to a single batch from the "Files changed" tab here on GitHub. Alternatively, you can manually apply them offline and push a commit the traditional way.

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,108 @@
...
+        If there is a channel dimension, provide the index here. If None
+        (default) then all axes are assumed spatial dimensions.
+    preserve_range : bool, optional
+        Whether to keep the original range of values. Otherwise the input
+        image is returned with dtype float with values scaled between 0 and 1.
+
+    Returns
+    -------
+    result : numpy array
+        The Butterworth-filtered image
```
grlee77, line 76:
```suggestion
    """Apply a Butterworth filter to enhance high or low frequency features.

    This filter is defined in the Fourier domain.

    Parameters
    ----------
    image : (M[, N[, ..., P]][, C]) ndarray
        Input image.
    cutoff_frequency_ratio : float, optional
        Determines the position of the cut-off relative to the
        shape of the FFT.
    high_pass : bool, optional
        Whether to perform a high pass filter. If False, a low pass filter is
        performed.
    order : float, optional
        Order of the filter which affects the slope near the cut-off.
        Higher order means steeper slope in frequency space.
    channel_axis : int, optional
        If there is a channel dimension, provide the index here. If None
        (default) then all axes are assumed spatial dimensions.
    preserve_range : bool, optional
        Whether to keep the original range of values. Otherwise the input
        image is returned with dtype float with values scaled between 0 and 1.

    Returns
    -------
    result : numpy array
        The Butterworth-filtered image.
```

skimage/filters/_fft_based.py
din14970: done, however I had to change it to `fftmodule` and `..util`

skimage/filters/_fft_based.py
din14970: This is a cool and elegant trick, and indeed with sparse we won't run into any memory issues. Thanks!

skimage/filters/_fft_based.py
din14970: yep, good suggestion, done

skimage/filters/_fft_based.py
din14970: It's a good suggestion, I've implemented it now as you without a loop. Not sure how it compares in terms of performance, but in my opinion it looks better since I can recycle the same ifft/fft statement.

din14970: I still have this odd slice construct, which I now use to insert an axis. Not sure if there is a more pythonic way to do this.

skimage/filters/_fft_based.py
din14970: I generalized it now so that also complex images can be passed in. If a real image is passed then `rfftn/irfftn` are used, otherwise `fftn/ifftn`. To allow for this I also made a small edit in `_get_ND_butterworth_filter` to create a mask of the right dimensions.

skimage/filters/tests/test_fft_based.py
din14970: Yes I didn't have any good ideas on this. I thought first of constructing a dummy FFT by creating a zero array and then initializing a few pixels to 1, inverse fourrier transforming, then passing this image into the bw filter, then fourrier transforming it back. Unfortunately this no longer resembles the original FFT. I guess if one wisely constructs the FFT it should work?

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,134 @@
...
+    # for real image FFT, halve the last axis
+    if real:
+        limit = d // 2 + 1
+        ranges[-1] = ranges[-1][:limit]
+    # q2 = squared Euclidian distance grid
+    q2 = functools.reduce(
+            np.add, np.meshgrid(*ranges, indexing="ij", sparse=True)
+            )
+    # division of order by 2 to avoid additional square root on q2
+    wfilt = 1 / np.sqrt(1 + np.power(q2, order / 2))
```
grlee77, line 44:
```suggestion
    wfilt = 1 / np.sqrt(1 + np.power(q2, order))
```


Assuming this `order` variable corresponds to `n` in the first equation of the [Butterworth wikipedia article](https://en.wikipedia.org/wiki/Butterworth_filter), then it seems this should be `order` rather than `order / 2`? I realize that you have avoided taking a sqrt, but it seems this would only reduce the term from (`2 * order`) to `order` and not to `order / 2`

grlee77:
Nice job updating this to use the rfftn and irfftn functions!

In searching for references to use of this function in image processing, I sometimes saw the filter defined without any square root in the denominator. For example [here](https://www.geeksforgeeks.org/matlab-butterworth-lowpass-filter-in-image-processing/), or in John C. Russ's "Image Processing Handbook" (3rd. ed.), where the filter equation is listed as:
`H = 1 / (1 + C * (R / R0) ** (2 * n))` where it is stated that C is usually either 1.0 or 0.414.

Other sources such as the Wikipedia article and [this one](https://www.electronics-notes.com/articles/radio/rf-filters/butterworth-formula-equations-calculations.php) or [this one](https://www.electronics-tutorials.ws/filter/filter_8.html) seem to agree with the form implemented here. 

Perhaps we should make the exact equation explicit in the Notes of the docstring here to avoid potential confusion over these differing conventions

jni (approved):
The nd implementation is gorgeous @din14970! regarding the memory concerns for large nD images, we typically leave that to later generations to address. 😂 It's much better to have a working and slow implementation merged to get further improvements down the line as needed.

There's a couple of comments from @grlee77 that need addressing (`order`/`sqrt` conventions), as well as the improved testing idea, but other than that this is ready for me.

Regarding testing, in many situations, a test is hard to define from first principles. In those cases, we typically use one or more of the following options:

- Run this for 1D arrays and for "2D multichannel" arrays (ie a series of 1D arrays), and compare it to the output from SciPy.
- Run this on a known image (e.g. a tiny crop `data.camera()`), check that the output makes sense, save it as a .npy file, and check against that. There's a few examples in e.g. `skimage/restoration/tests`. Make this small so that they don't take up too much space. Converting to uint8 is also a useful trick for compression.
- Do the same as above but compare to a 3rd party library.

din14970:
@grlee77 I tried to dig into this a bit but am still not entirely sure.

I thought first it might be a dimension issue. The references where there is a square root in the denominator always pertain to 1D signals; whenever it concerns 2D signals there is no square root. I couldn't find any reference explaining the difference however. So my hypothesis was that one could extrapolate to any dimension with `1 / (1 + d ** (2*n)) ** (k/2)` with `k` the number of spatial dimensions. To test this out I made the following example:

```python
import scipy.fft as fft
import numpy as np

def funk(R):
    return R

w1 = 0.1 
order = 8
# 1D
x = np.arange(100)
y1 = funk(x)
ify1 = butterworth(y1, w1, False, order, 0.5)
ify1b = butterworth(y1, w1, False, order, 0.5)

# 2D
X1, X2 = np.meshgrid(x, x, sparse=True)
R = np.sqrt(X1**2 + X2**2)
y2 = funk(R)
ify2 = butterworth(y2, w1, False, order, 0.5)
ify2b = butterworth(y2, w1, False, order, 1)

# 3D
X1, X2, X3 = np.meshgrid(x, x, x, sparse=True)
R = np.sqrt(X1**2 + X2**2 + X3**2)
y3 = funk(R)
ify3 = butterworth(y3, w1, False, order, 0.5)
ify3b = butterworth(y3, w1, False, order, 1.5)

fig, ax = plt.subplots(ncols = 3, figsize=(15, 5))
ax[0].plot(x, y1, label="1D")
ax[0].plot(x, y2[0], label="2D")
ax[0].plot(x, y3[0,0], label="3D")
ax[0].legend()
ax[0].set_title("Original function")

ax[1].plot(x, ify1, label="1D")
ax[1].plot(x, ify2[0], label="2D")
ax[1].plot(x, ify3[0,0], label="3D")
ax[1].legend()
ax[1].set_title("Butter filter exponent 0.5")

ax[2].plot(x, ify1b, label="1D")
ax[2].plot(x, ify2b[0], label="2D")
ax[2].plot(x, ify3b[0,0], label="3D")
ax[2].legend()
ax[2].set_title("Butter filter adjusted exponent")
```
Basically I generate some N-D images based on some function of R on a 1D, 2D and 3D grid. Looking along a single axis, all these functions look identical (figure 1). I then apply the butterworth filter and plot again the result along one dimension. I modified the butterworth function to also take the parameter `exponent`: `wfilt = 1 / (1 + np.power(q2, order))**exponent`. The result:

afbeelding

The middle image is where the exponent is 0.5 for all dimensions, the right one is where the exponent is adjusted according my hypothesis. Basically it looks like it doesn't matter very much. I looked into different functions, different orders, different w0 but couldn't find a clear reason to use one or the other.

Therefore, since the primary use-case for this thing would be to highlight/remove certain frequencies in an image, and since a user will have to tune parameters on a case by case basis anyway, I propose we go with exponent 1 in the denominator for simplicity and speed. The filter has sufficient flexibility with the `order` and `cutoff` parameters. 

@jni I think I'll go with the second approach, though this breaks down if the filter is ever changed for some reason (i.e. people want the square root in the denominator)

din14970:
@grlee77 @jni the tests are failing because it can't find the test files I added; locally it works fine. What is the proper way to do this?

skimage/filters/tests/test_fft_based.py
```diff
@@ -0,0 +1,58 @@
+import pytest
+import os
+import numpy as np
+from .. import butterworth
+from skimage.data import astronaut, coins
```
mkcor, line 5:
```suggestion
from skimage.data import astronaut, coins
from .. import butterworth
```

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,140 @@
...
+    order : float, optional
+        Order of the filter which affects the slope near the cut-off.
+        Higher order means steeper slope in frequency space.
+    channel_axis : int, optional
+        If there is a channel dimension, provide the index here. If None
+        (default) then all axes are assumed to be spatial dimensions.
+    preserve_range : bool, optional
+        Whether to keep the original range of values. Otherwise the output
+        image is returned with dtype float with values scaled between 0 and 1.
+        If a complex valued image was passed, no rescaling is performed.
```
mkcor, line 80:
```suggestion
        Whether to keep the original range of values. Defaults to False. If False,
        image is returned with dtype float with values scaled between 0 and 1.
        If a complex-valued image is passed, no rescaling is performed.
```

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,140 @@
...
+    -------
+    result : ndarray
+        The Butterworth-filtered image.
+
+    Notes
+    -----
+    A band-pass filter can be achieved by combining a high pass and low
+    pass filter.
+
+    The literature contains multiple conventions for the  functional form of
```
mkcor, line 92:
```suggestion
    The literature contains multiple conventions for the functional form of
```

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,140 @@
...
+    the Butterworth filter. Here it is implemented as the n-dimensional form of
+
+    .. math::
+        \frac{1}{1-\left(\frac{f}{c*f_{max}}\right)^{2*n})
+
+    with :math:`f` the absolute value of the spatial frequency, :math:`c` the
+    `cutoff_frequency_ratio` and :math:`n` the `order` modeled after [2]
```
mkcor, line 99:
```suggestion
    `cutoff_frequency_ratio` and :math:`n` the `order` modeled after [2]_
```

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,140 @@
...
+    Apply a high pass and low pass Butterworth filter to a grayscale and
+    color image respectively:
+
+    >>> from skimage.data import camera, astronaut
+    >>> from skimage.filters import butterworth
+    >>> high_pass = butterworth(camera(), 0.07, True, 8)
+    >>> low_pass = butterworth(astronaut(), 0.01, False, 4, channel_axis=-1)
+
+    Reference
+    ---------
```
mkcor, line 112:
```suggestion
    References
    ----------
```

skimage/filters/tests/test_fft_based.py
jni: Actually, 

```suggestion
from skimage.data import astronaut, coins
from skimage.filters import butterworth
```

We use absolute imports in our tests.

jni:
@din14970 apologies for the long delay in responding.

If you look at the restoration tests, they use [`skimage._shared.testing.fetch`](https://github.com/scikit-image/scikit-image/blob/00875e67733a950f8f2477ac77a38712f14e62f3/skimage/restoration/tests/test_restoration.py#L7) to grab the data, example [here](https://github.com/scikit-image/scikit-image/blob/00875e67733a950f8f2477ac77a38712f14e62f3/skimage/restoration/tests/test_restoration.py#L38). To be honest I don't actually understand that infrastructure very well, but, when in Rome! So I think try copying that pattern and 🤞!

din14970:
@jni I tried a couple combinations but it doesn't work. It looks to me like somehow the files already need to be accepted in the official repo so they can be downloaded by pooch. Also adding them to this registry file doesn't work. Would be really helpful if someone who knows this system could take a look. Since tests initially worked for me locally but don't work on Azure I think it'll be hard for me to solve the issue without additional information.

mkcor:
> @jni I tried a couple combinations but it doesn't work. It looks to me like somehow the files already need to be accepted in the official repo so they can be downloaded by pooch. Also adding them to this registry file doesn't work. Would be really helpful if someone who knows this system could take a look. Since tests initially worked for me locally but don't work on Azure I think it'll be hard for me to solve the issue without additional information.

@din14970 yes, absolutely; I've been meaning to document this for a while, but haven't got around to doing it yet. To ensure that tests pass (notably, `data/tests/test_data.py::test_download_all_with_pooch`), you want to

1. submit a PR containing only the new data files and,
2. once this first PR merged, submit another PR updating the data registry (and typically adding tests and/or code making use of the new data files).

For example, this is what I did with

1. #5276 followed by:
2. #5308

If you don't mind, you could open a new PR with `astronaut_butter.npy` and `coins_butter.npy`; I'll review and approve; then you can merge `main` into this branch and tests should pass.

din14970:
@mkcor thanks a lot, this is the information I was missing! Seems like a bit of a strange system though, especially if some mistake is discovered later in the code but the test files are already in main. I opened #5415 to add the test data files. In retrospect it seems it might be easier to just hard code the expected result arrays into the tests. Anyway let's try this way, once it's merged I will merge main locally and force push here.

skimage/filters/__init__.py
```diff
@@ -21,13 +21,15 @@
 from ._sparse import correlate_sparse
 from ._unsharp_mask import unsharp_mask
 from ._window import window
+from ._fft_based import (butterworth)
```
mkcor, line 24:
```suggestion
from ._fft_based import butterworth
```

skimage/filters/_fft_based.py
mkcor: @din14970 the import order/grouping that @grlee77 suggested follows a convention: https://www.python.org/dev/peps/pep-0008/#imports

mkcor:
> @mkcor thanks a lot, this is the information I was missing! Seems like a bit of a strange system though, especially if some mistake is discovered later in the code but the test files are already in main.

Yes, I admit it is a little strange (https://github.com/scikit-image/scikit-image/pull/5234#issuecomment-803909564). But it's originally intended for adding actual ('interesting' and meaningful) datasets rather than toy data samples used only in unit tests.

> I opened #5415 to add the test data files. In retrospect it seems it might be easier to just hard code the expected result arrays into the tests.

Right, especially since the data files are tiny. Anyway, PR #5415 is in and you can now move forward.

> Anyway let's try this way, once it's merged I will merge main locally and force push here.

Why would you have to force push? :thinking:

skimage/filters/_fft_based.py
din14970: woops, I made the changes locally and didn't see that when I was making corrections. I also added an empty line below the numpy import as per the PEP8 rule. I found some other files in the repo that do not strictly adhere to these rules though ;)

din14970:
> Right, especially since the data files are tiny. Anyway, PR #5415 is in and you can now move forward.

If I make any more contributions in the future then I will keep this in mind.

> Why would you have to force push?

My bad I got a bit confused, usually when things get messy with files I tend to rebase, in this case with just a merge of course force is not necessary.

din14970:
Something somewhere is wrong with the docs, if anyone can find the issue easily please let me know as I've been staring at this for quite some time already. I will stare at it some more but not today. 

When I build local it looks like this:

image

The errors I get:

```
_fft_based.py:docstring of skimage.filters._fft_based.butter
worth:10: WARNING: Block quote ends without a blank line; unexpected un
indent.

_fft_based.py:docstring of skimage.filters._fft_based.butter
worth:39: WARNING: Error in "rubric" directive:
no content permitted.

.. rubric:: Notes

    A band-pass filter can be achieved by combining a high pass and low
    pass filter.

    The literature contains multiple conventions for the functional for
m of
    the Butterworth filter. Here it is implemented as the n-dimensional
 form of

    .. math::
_fft_based.py:docstring of skimage.filters._fft_based.butter
worth:55: WARNING: Error in "rubric" directive:
no content permitted.

.. rubric:: References

    .. [R158eb86722ba-1] Butterworth, Stephen. "On the theory of filter
 amplifiers."
           Wireless Engineer 7.6 (1930): 536-541.
    .. [R158eb86722ba-2] Russ, John C., et al. "The image processing ha
ndbook."
           Computers in Physics 8.2 (1994): 177-178.
```

grlee77:
> Something somewhere is wrong with the docs, if anyone can find the issue easily please let me know 

I will take a look.

mkcor:
> When I build local it looks like this:

It looks like your screenshot doesn't match the PR version, because the latter reads "Here it is implemented as the n-dimensional form of" before the math formula and "with :math:`f` the absolute value ..." after it (as in the error you report), not "Here, \(f\) the absolute value..."

din14970:
@mkcor yes sorry about that, I was messing around with it locally because it was a bit faster to iterate and I haven't pushed up the changes. The gist of it looked the same:
* "This filter is in the Fourrier domain" and "The input image" are strangly formated  -- probably related to the indentation block error
* The `Notes` doesn't render, neither does the math equation
* `References` doesn't render

grlee77:
The doc errors were due to LaTeX formatting issues, but this isn't that clear from the message. The solution is using `\\` to get `\` in the docstring's LaTeX code. There was also a typo in the equation with a `)` instead of `}` at the end of the line.

grlee77:
f53349f closes #5416, which was initially preventing me from building the docs locally. It is unrelated to this PR, though, so perhaps I should open a standalone PR for it

din14970:
@grlee77 awesome thanks a lot for looking into this! Weird because I also considered this but I saw multiple examples e.g. https://github.com/scikit-image/scikit-image/blob/178e05f17855859d6661ebb6e04d8ddcb50c1cc5/skimage/metrics/simple_metrics.py#L200 where a single `\` appears to be enough. Anyway with these changes it seems to be working!

grlee77:
Yes, but I think those cases start the docstring with `r"""` instead of `"""` (i.e. it is a Python raw string). We could also take that approach here to make the LaTeX more readable.

skimage/filters/_fft_based.py
```diff
@@ -0,0 +1,144 @@
...
+        abs_channel = channel_axis % image.ndim
+        post = image.ndim - abs_channel - 1
+        sl = ((slice(None),) * abs_channel + (np.newaxis,) +
+              (slice(None),) * post)
+        wfilt = wfilt[sl]
+    if is_real:
+        butterfilt = fft.irfftn(wfilt * fft.rfftn(image, axes=axes),
+                                s=fft_shape, axes=axes)
+        out_range = dtype_limits(image) if preserve_range else (0.0, 1.0)
+        butterfilt = rescale_intensity(butterfilt, out_range=out_range)
```
grlee77, line 140:
```suggestion
        if not preserve_range:
            image = img_as_float(image)        
        butterfilt = fft.irfftn(wfilt * fft.rfftn(image, axes=axes),
                                s=fft_shape, axes=axes)
```

I would have instead, expected behavior like the above for `preserve_range`. The current code will rescale the range for pretty much all inputs even when preserve_range is True. For example, floating point inputs will not have their range preserved.

@jni, thoughts? If we change this here we will need to change the data files used in the tests.

skimage/filters/tests/test_fft_based.py
grlee77: I came up with another basic test case that just verified highpass vs. lowpass behavior, but want to resolve the one remaining issue regarding `preserve_range` first.

skimage/filters/_fft_based.py
jni: Gah. I think you're right @grlee77. 🤦

din14970: sorry for replying so late, didn't have time for a couple days. I think I implemented it in this way because the fft + ifft may introduce image intensities outside the original data-type range. For example if you put in a uint8 image, you may get negative intensities. This is because when you remove some spatial frequencies, you might get "jumps" in places where there are edges. My original interpretation of `preserve_range` was that the output image should fit in the same dtype range as the input image so that it could also be cast as the same dtype. This is why I do the rescaling after the filter. But I now see your point that this has the potentially unwanted side effect of stretching the histogram to the extrema, so probably your solution is better. Note however that in your implementation, it is still possible to get an image that is not rescaled to a range [0, 1] or [-1, 1] if `preserve_range=False`, but I don't see another appropriate alternative that won't stretch the histogram. I suggest the following:

* I change the implementation to what @grlee77 suggests
* I add some clarification in the `Notes` section about `preserve_range` that it is possible to still get some outlier intensities
* I remove the test data files and hardcode new test data into the test files.

Alternatively, since the option doesn't seem to make much sense anyway I remove the `preserve_range` argument. Any vote on this?

din14970:
I decided to get rid of `preserve_range` as it does not make sense in my opinion. Even if input is scaled between 0-1, the output still will not be. It just adds confusion to what the argument is supposed to do, and depending on what users want it is easier if they just rescale the image intensity before or after the filter. I removed the test files and just copied the expected arrays into the tests. I merged again with main, now something is failing again.

grlee77:
> I merged again with main, now something is failing again.

These seem unrelated and should be resolved by #5434

grlee77:
I am okay with having these relatively small arrays in-line in the test file, but let's see what other maintainers think. I updated setup.cfg to ask it to ignore some whitespace and line-length issues that would be reported during style checking.

Please see the two new test cases added in 0a769f7 and see if they look reasonable to you. I wanted to test something with non-zero input that would at least verify that the spectrum of the output had been modified in a roughly high-pass or low-pass nature. The specific 0.1 and 0.9 thresholds are basically arbitrary and not tuned to the specific filter settings.

din14970:
Thanks for adding some tests, they are definitely valuable. I just left two minor comments which may just be misunderstandings on my part.

mkcor:
@din14970 CI should pass if you merge `main` now :+1:

skimage/filters/tests/test_fft_based.py
```diff
@@ -0,0 +1,250 @@
...
+    im = np.zeros((3, 4, 5, 6), dtype=dtype)
+    filtered = butterworth(im, channel_axis=chan)
+    assert filtered.shape == im.shape
+
+
+def test_butterworth_correctness_bw():
+    small = coins()[180:190, 260:270]
+    filtered = butterworth(small,
+                           cutoff_frequency_ratio=0.2,
+                           )
```
mkcor, line 98:
```suggestion
                           cutoff_frequency_ratio=0.2)
```
so it's consistent with line 121?

mkcor:
> I am okay with having these relatively small arrays in-line in the test file, but let's see what other maintainers think.

I am also okay with this practice. Technically, this choice reverts #5415 but this is fine and was contemplated in the first place.

> I updated setup.cfg to ask it to ignore some whitespace and line-length issues that would be reported during style checking.

Wonderful!

skimage/filters/tests/test_fft_based.py
din14970: done

grlee77:
Okay, the one failure here is unrelated and seems to be a new warning from the just released scipy 1.7.0. I can make a new PR to fix it

```
E                   ValueError: Unexpected warning: scipy.linalg.pinv2 is deprecated since SciPy 1.7.0, use scipy.linalg.pinv instead
```

grlee77 (approved):
Thanks for sticking with this @din14970, it looks good to me now.

mkcor (approved):

din14970:
@grlee77 @mkcor @jni thanks a lot also for your thorough reviews, suggestions and improvements!

PR #5385, opened 2021-05-11
rename greycomatrix and greycoprops. prefer "gray" in documentation

### LINKED ISSUE #2604: SKIP: gray vs grey
## Description

Recently, I saw a student stuck after typing

```
io.imread('image.jpg', as_gray=True)
```
Silently, it loads as a RGB image.

Indeed, the API is `as_grey`. The problem is that our API is not homogeneous. It seems that US spelling is in majority in our library. I would suggest to either:
* make sure that both gray and grey are working
* raise a warning

Any suggestion @scikit-image/core ?

grlee77:
## Description

partially resolves #2604 (following the trend of previous work in the color module for rgb2grey->rgb2gray, etc)

The first commit deals initiates a deprecation for:
`greycomatrix` -> `graycomatrix`
`greycoprops` -> `graycoprops`

My main question for that commit is the version to set for the deprecation. I have it as 1.0, but am fine with increasing it to a subsequent release if we want to give at least 2 releases before removing the old functions.

The remaining commits involve consistently using gray rather than grey in the documentation.

I can address the remaining uses of grey in functions and filenames in the morphology module in a separate PR.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


mkcor:
> My main question for that commit is the version to set for the deprecation. I have it as 1.0, but am fine with increasing it to a subsequent release if we want to give at least 2 releases before removing the old functions.

I find that it's okay for this particular change. I would be like @JDWarner on this one :wink:

mkcor (approved):

alexdesiqueira:
Looks great. Thank you @grlee77!

alexdesiqueira:
Looks great! Thank you @grlee77

alexdesiqueira:
Looks great! Thank you @grlee77!

PR #5392, opened 2021-05-13
Integral_image: promote floating point inputs to double by default

grlee77:
## Description

This PR was split off from #5372 as it is a change in existing behavior. To avoid surprises due to poor accuracy of floating point cumulative sums, I propose we default to double precision when computing integral images. See additional details in prior comment: https://github.com/scikit-image/scikit-image/pull/5372#discussion_r625460770.

The user can override the default by manually specifying the `dtype` argument.

One question is whether we can consider this a bug fix or if we should introduce a deprecation cycle before making the switch. The new output should be strictly an improvement from an accuracy standpoint, but output dtype will be larger for `float16`, `float32` or `complex64` inputs.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/integral.py
+++ b/skimage/transform/integral.py
@@ -37,7 +37,7 @@ def integral_image(image, *, dtype=None):
     """
     if dtype is None and image.real.dtype.kind == 'f':
         # default to at least double precision cumsum for accuracy
-        dtype = np.promote_types(image.dtype, float)
+        dtype = np.promote_types(image.dtype, np.float64)
         image = image.astype(dtype, copy=False)
 
     S = image
--- a/skimage/transform/integral.py
+++ b/skimage/transform/integral.py
@@ -23,8 +23,8 @@ def integral_image(image, *, dtype=None):
 
     Notes
     -----
-    For better accuracy and to avoid potential overflow, the datatype of the
-    output may differ from the input when the default dtype of None is used.
+    For better accuracy and to avoid potential overflow, the data type of the
+    output may differ from the input's when the default dtype of None is used.
     For inputs with integer dtype, the behavior matches that for
     :func:`numpy.cumsum`. Floating point inputs will be promoted to at least
     double precision. The user can set `dtype` to override this behavior.
--- a/skimage/transform/integral.py
+++ b/skimage/transform/integral.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 
-def integral_image(image):
+def integral_image(image, *, dtype=None):
     r"""Integral image / summed area table.
 
     The integral image contains the sum of all elements above and to the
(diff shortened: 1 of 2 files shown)


skimage/transform/integral.py
```diff
@@ -21,15 +21,28 @@ def integral_image(image):
...
+
     References
     ----------
     .. [1] F.C. Crow, "Summed-area tables for texture mapping,"
            ACM SIGGRAPH Computer Graphics, vol. 18, 1984, pp. 207-212.
 
     """
+    if dtype is None and image.real.dtype.kind == 'f':
+        # default to at least double precision cumsum for accuracy
+        dtype = np.promote_types(image.dtype, float)
```
jni, line 40: I think on some Windows builds, "float" will default to float32. Should we specify np.float64 here?

jni (approved):

skimage/transform/integral.py
grlee77: That is the case for `int` on 32-bit, but I thought float was usually double precision? Definitely not a problem to just use np.float64 to make it unambiguous, though

skimage/transform/integral.py
```diff
@@ -21,15 +21,28 @@ def integral_image(image):
     S : ndarray
         Integral image/summed area table of same shape as input image.
 
+    Notes
+    -----
+    For better accuracy and to avoid potential overflow, the datatype of the
```
mkcor, line 26:
```suggestion
    For better accuracy and to avoid potential overflow, the data type of the
```
Not sure, actually, but copying https://scikit-image.org/docs/dev/user_guide/data_types.html. Or "dtype" directly.

skimage/transform/integral.py
```diff
@@ -21,15 +21,28 @@ def integral_image(image):
...
+    output may differ from the input when the default dtype of None is used.
```
mkcor, line 27:
```suggestion
    output may differ from the input's when the default dtype of None is used.
```
or "that of the input"

mkcor:
> One question is whether we can consider this a bug fix or if we should introduce a deprecation cycle before making the switch.

I'm happy to consider it a bug fix, since the change fixes floating-point cumulative sums which used to be off. Does @rfezzani agree?

> The new output should be strictly an improvement from an accuracy standpoint, but output dtype will be larger for [...] complex64 inputs.

Which (larger) output dtype do `complex64` inputs yield? I guess `complex128` where the complex number is represented by two 64-bit floats?

grlee77:
> Which (larger) output dtype do complex64 inputs yield? I guess complex128 where the complex number is represented by two 64-bit floats?

Yes, this is correct.

rfezzani:
> > One question is whether we can consider this a bug fix or if we should introduce a deprecation cycle before making the switch.
> 
> I'm happy to consider it a bug fix, since the change fixes floating-point cumulative sums which used to be off. Does @rfezzani agree?
> 
> > The new output should be strictly an improvement from an accuracy standpoint, but output dtype will be larger for [...] complex64 inputs.
> 
> Which (larger) output dtype do `complex64` inputs yield? I guess `complex128` where the complex number is represented by two 64-bit floats?

Sorry for not being available these last weeks :pray:...

:+1: I agree @mkcor, and thank you @juan for merging this :wink: and thank you again @grlee77 :tada:

PR #5394, opened 2021-05-14
Speed up find_local_max when given a finite num_peaks

klaussfreire:
## Description

Optimizes _shared.coord.ensure_spacing as used in
feature.peak.find_local_max when given num_peaks

There's no need to process all candidates, which could
far outnumber the requested maximum, since candidates
in find_local_max are sorted by intensity. So only return
the first few matches and ignore the rest.

The resulting speedup can be huge in a number of cases

## For reviewers

  later.

### DIFF
--- a/skimage/_shared/coord.py
+++ b/skimage/_shared/coord.py
@@ -2,7 +2,7 @@ import numpy as np
 from scipy.spatial import cKDTree, distance
 
 
-def _ensure_spacing(coord, spacing, p_norm):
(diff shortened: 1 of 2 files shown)


jni (approved):
Cool, thanks @klaussfreire! We've definitely been having performance issues with that function so this is much appreciated. The changes are completely internal (no API changes), which also makes it easy to get it in.

mkcor (approved):
Nice!

PR #5397, opened 2021-05-14
Increase automation in release process.

leej3:
## Description

Hi all,  @jni tasked us with drafting up a more automated release strategy. @viniciusdc has worked through this and we were wondering if we could get some feedback on it for any desired changes.

The overall objective is to automate as much as possible. Everything should now proceed upon the push of a tag. In short this will trigger the following (see workflow [here](https://github.com/viniciusdc/scikit-image/actions/runs/842727666)):

- A build of all the wheels
- Testing as part of cibuildwheels (pytest --pyargs skimage for now)
- Use twine to upload wheels and source
- Create the github release
 

Some things that can be added:

- Once people are happy with this we can revert the second commit so that the workflow executes correctly under the scikit-image organization.
- GPG signing binaries using twine (thoughts/preferences for this are welcome)
- We have not done anything with the documentation for now (we will likely switch to town-crier for this so didn’t want to spend too much time on it).
- Currently we are uploading to testpypi but it would be great to get confirmation that it works seamlessly for the standard release before merging… some secrets will need to be set for github actions. 

## For reviewers

  later.


.github/workflows/make_release.yml
```diff
@@ -0,0 +1,231 @@
...
+          pip install 'numpy>=1.16.5,!=1.18.0'
+      
+      - uses: actions/download-artifact@v2
+        id: download
+        with:
+          name: wheels
+          path: ./dist
+    
+      - name: Publish the source distribution on PyPI
+        if: success()
```
jni, line 214: imho we should push with/without success. We can manage failing tests by releasing rcs before the real thing. We've often had sporadic CI failures that we know don't actually affect the library by and large, and we want to release anyway. So it's great to *know* whether there are failing tests, but it should not stop the upload. imho! @scikit-image/core?

jni:
Other than my one comment, imho this is ready and I'm ready for it! 😂

grlee77 (commented):
Thanks @leej3 and @viniciusdc, this looks pretty nice!

So the top part of the file is a copy of `cibuildwheel.yml` (with changes only to the `on:` section and setting `fast-fail` set to True) and the new part is the `deploy` job in the bottom part of the file. It looks pretty readable to me, so I think other maintainers will be able to make small modifications as needed.

My main question is what happens to the existing `cibuildwheel.py`. Do we have to also keep it for the other `on:` conditions? 

If we are going to have both, I think we should add a comment in each advising devs to make sure that the wheel building jobs stay in sync between the two. Also, should the action triggering on `tags` be removed from `cibuildwheel.py` now that this case is covered by `make_release.py`?

.github/workflows/make_release.yml
hmaarrfk: I agree that we shouldn't stop releases on failure. you summarized it well.

.github/workflows/make_release.yml
```diff
@@ -0,0 +1,231 @@
+name: Create Release
+on:
+  push:
+    tags:
+      - 'v*'
+env:
+  CIBW_BUILD_VERBOSITY: 2
+  CIBW_TEST_REQUIRES: pytest pooch pytest-localserver pytest-faulthandler
```
stefanv, line 8: This comes from `requirements/test.txt` I presume; can we install these from that file in a step, or do they need to be explicitly re-specified here?

.github/workflows/make_release.yml
```diff
@@ -0,0 +1,231 @@
...
+  CIBW_TEST_COMMAND: pytest --pyargs skimage
+
+
+jobs:
+  build_linux_37_and_above_wheels:
+    if: github.repository_owner == 'viniciusdc'
+    name: Build python ${{ matrix.cibw_python }} wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: true
```
stefanv, line 18: Curious if we want fail-fast here, or whether we should push through.

.github/workflows/make_release.yml
```diff
@@ -0,0 +1,231 @@
...
+          CXX: /usr/bin/clang++
+          CPPFLAGS: "-Xpreprocessor -fopenmp"
+          CFLAGS: "-Wno-implicit-function-declaration -I/usr/local/opt/libomp/include"
+          CXXFLAGS: "-I/usr/local/opt/libomp/include"
+          LDFLAGS: "-Wl,-rpath,/usr/local/opt/libomp/lib -L/usr/local/opt/libomp/lib -lomp"
+
+      - name: Build wheels for CPython (MacOS)
+        if: matrix.os == 'macos-latest'
+        run: |
+          brew install libomp
```
stefanv, line 96: Already installed previously?

stefanv (commented):
Thank you for this @leej3 @viniciusdc!

Currently, the Mac/Linux/Windows jobs are duplicated.  Do you think it would be very messy to pull them into one, with conditionals for the various OS-specific bits?  Or is this done for some other reason, such as better parallel builds?

leej3 (commented):
Thanks for the feedback all. We'll resolve these details in the next few days.

Regarding the duplication of cibuildwheel.yml we think the cleanest approach would be to merge them and call the file wheels_tests_and_release.yml or something like that. The cibuildwheel jobs will be triggered by tags or the cibuildwheel branch, the deploy job would only be triggered by the version tags. Let us know if that's not ok.

leej3:
I believe we've addressed all the comments now. 

I think it would be best to squash all of these commits and having  a single commit to harmonize with main (largely just replace "viniciusdc" with "scikit-image"). Let me know what you would like though... superceding with another PR so as to not erase the context of the comments here might also be desirable.

@viniciusdc, could you check the apveyor failure... possibly just something I introduced when I refactored.

setup.py
```diff
@@ -14,8 +14,8 @@
 from numpy.distutils.command.build_ext import build_ext
 
 
-DISTNAME = 'scikit-image'
-DESCRIPTION = 'Image processing in Python'
```
sciunto, line 18: :confused:

setup.py
```diff
@@ -14,8 +14,8 @@
...
+DISTNAME = 'scikit-image-testviniver'
+DESCRIPTION = 'Just a test'
```
sciunto, line 18: :confused:

sciunto (dismissed):

setup.py
jni: @leej3 ☝️ this is the cause of the appveyor failure, should go away after that is removed.

I suggest simply removing those lines and we can squash and merge the PR here — no need to manually squash the commits.

.github/workflows/wheel_tests_and_release.yml
```diff
@@ -183,8 +174,53 @@ jobs:
...
+        name: Install Python
+        with:
+          python-version: '3.7'
+      
+      - name: Install Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install twine
+          pip install beautifulsoup4
+          pip install 'numpy>=1.16.5,!=1.18.0'
```
stefanv, line 202: Where does this set of dependencies come from?  Can we specify these in a requirements file?

stefanv (commented):
This looks very nice and tidy; thank you!

One question on the side.

.github/workflows/wheel_tests_and_release.yml
viniciusdc: `beautifulsoup4` was intended to be a required dependency to install the previously used `wheel_uploader`, so it can be safely removed. The numpy requirements in there, seems to be related to an issue of incompatibility in the environment. But as it has already been fixed, we can remove that depency as well. I will test those removals and commit here once finished. Thanks @stefanv for pointing this out.

viniciusdc: HI @stefanv , I removed the bs4 dependency as it was needless, but the numpy restriction in there refers to the execution of `setup.py` in the next step. As it's removal causes 
```python
File "setup.py", line 14, in <module>
 ```
As we move forward we might need a separate requirements file for this, but right now I think that we can maintain this way to give more visibility. what do you think?

stefanv: I guess what had me confused is that this is available in any version of NumPy; so can we install NumPy from the skimage requirements file instead?

hmaarrfk: Hint: `-r requirements/build.txt`

leej3: Yes, that looks easiest. Thanks. Sorry for the delay.

hmaarrfk: no delay. sometimes people implementing features are more in tune with challenges that arrise with what might seem like an "obvious" approach. I'm glad to know the suggestion works.

leej3:
Is this ok to be merged?

stefanv (approved):
Looks good to me, thanks @leej3!

alexdesiqueira:
It seems that @leej3 dealed with @sciunto's review already. Merging; please let us know if there's something else you would like to see, François, and we deal with that on another PR.
Thank you very much @leej3!

jni:
❤️!!!

grlee77:

PR #5400, opened 2021-05-17
release the GIL in denoise_tv_bregman and denoise_bilateral

grlee77:
## Description

I tried using `apply_parallel` with the various `denoise_*` functions in `skimage.restoration`. Most showed a substantial speedup, but `denoise_tv_bregman` and `denoise_bilateral` showed worse performance when calling with `apply_parallel` than without. This turns out to be because the GIL is not being released by these two functions.

As in #5399, this PR just adds a `with nogil` context block into each of these functions so the GIL will be released, allowing multithreading. I confirmed locally that this leads to the expected improvement.

The somewhat large diff is misleading as nearly all line differences are just the change in indentation.

## Timing

For example for `denoise_tv_bregman` on an input of shape (2048, 4096) it takes:
normal call (single thread):  2.33 s
call with apply_parallel (prior to this PR, GIL not released): 3.2 s
call with apply_parallel (after this PR): 0.53 s

The result for `denoise_bilateral` is very similar.

### DIFF
--- a/skimage/restoration/_denoise_cy.pyx
+++ b/skimage/restoration/_denoise_cy.pyx
@@ -12,7 +12,7 @@ from .._shared.fused_numerics cimport np_floats
 
 cnp.import_array()
 
-cdef inline Py_ssize_t Py_ssize_t_min(Py_ssize_t value1, Py_ssize_t value2):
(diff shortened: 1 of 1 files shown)


jni (approved):

alexdesiqueira:
Thank you @grlee77!

Issue #5259, opened 2021-03-04
PolynomialTransform.estimate() potentially uses more memory than it needs to


aleksejs-fomins:
See [forum discussion](https://forum.image.sc/t/basic-polynomialtransform-estimate-blows-up-with-memory-overflow/49659)

jni:
tldr, if we use scipy.sparse.linalg.svds to estimate the polynomial transform coefficients, instead of np.linalg.svd, we save massive amounts of memory for a potentially small runtime cost. We could either make this optional or just change it altogether.

jni:
I should say: the runtime cost is small in the general case, but in the case where you instantiate a massive NumPy array, sparse is much much faster.

rfezzani:
I am :+1: for `svds` :wink:

abigailstone:
Hi - I'm an open source newbie looking for a good first issue! I'd be interested in tackling this if nobody's started working on it yet. 
I have a quick clarification question first: there's some talk in that forum post about a `preserve_memory` flag for sparse `svds`, is that the preferred solution or should the sparse completely replace the existing `svd`?

jni:
Hi @abigailstone, and welcome! :blush: Given the potential gains, my preference is to just replace it altogether. Which also has the advantage of being easier to implement! :blush:

jni:
(And I think @rfezzani's comment above means at least one other core dev agrees!)

rfezzani:
Your are guessing well @jni :slightly_smiling_face:

ChenFred98:
Hi, I am also new to the whole open source community and I would love to be able to help for this project ! I'm not too familiar with the way problems should be tackled when someone is already working on it, should I get in contact with @abigailstone to see how the issue-fixing is going or should I consider that I can work on the issue as long as it is open ?

PR #5402, opened 2021-05-20
Use sparse svd from scipy to save memory usage

patelvyom:
## Description
Fix issue #5259 by using sparse SVD by default as suggested to save memory utilisation.
 

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -1449,7 +1449,7 @@ class PolynomialTransform(GeometricTransform):
         A[rows:, -1] = yd
 
         # scipy sparse svd uses less memory
-        _, _, V = svds(A)
+        _, _, V = svds(A, return_singular_vectors = True)
 
         # solution is right singular vector that corresponds to smallest
         # singular value
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -1448,7 +1448,7 @@ class PolynomialTransform(GeometricTransform):
         A[:rows, -1] = xd
         A[rows:, -1] = yd
 
-        #scipy sparse svd uses less memory
+        # scipy sparse svd uses less memory
         _, _, V = svds(A)
 
         # solution is right singular vector that corresponds to smallest
--- a/skimage/transform/_geometric.py
+++ b/skimage/transform/_geometric.py
@@ -2,6 +2,7 @@ import math
 import numpy as np
 from scipy import spatial
 from scipy.sparse.linalg import svds
+from scipy.linalg import svd
 import textwrap
 
 from .._shared.utils import get_bound_method_class, safe_as_int
@@ -1447,9 +1448,13 @@ class PolynomialTransform(GeometricTransform):
 
         A[:rows, -1] = xd
         A[rows:, -1] = yd
-
-        # scipy sparse svd uses less memory
-        _, _, V = svds(A, return_singular_vectors = True)
+        
+        if A.shape[1] <= A.shape[0]:
+            # scipy sparse svd uses less memory
+             _, _, V = svds(A, k=1, which='SM', return_singular_vectors='vh')
+        else:
+            # use full SVD in the overdetermined case
+            _, _, V = np.linalg.svd(A, full_matrices=True)
 
         # solution is right singular vector that corresponds to smallest
         # singular value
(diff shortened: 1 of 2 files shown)


skimage/transform/_geometric.py
```diff
@@ -1447,7 +1448,8 @@ def estimate(self, src, dst, order=2):
         A[:rows, -1] = xd
         A[rows:, -1] = yd
 
-        _, _, V = np.linalg.svd(A)
+        # scipy sparse svd uses less memory
+        _, _, V = svds(A, return_singular_vectors = True)
```
mkcor, line 1452:
```suggestion
        _, _, V = svds(A, return_singular_vectors=True)
```
Are you an R programmer? :wink: It looks like `return_singular_vectors=True)` is the default anyway, so there's no need to specify it, I would say. Can you run the test suite locally?

vyompatel060500: After running the tests I think I figured out the issue. scipy.sparse.linalg.svds() does not guarantee the order of returned singular vectors and hence it may fail the assumption in some cases. I am looking for other alternatives.  If you have any ideas it'll be great.

grlee77, line 1452:
```suggestion

        if A.shape[1] <= A.shape[0]:
            # scipy sparse svd uses less memory
            _, _, V = svds(A, k=1, which='SM', return_singular_vectors='vh')
        else:
            # use full SVD in the overdetermined case
            _, _, V = np.linalg.svd(A, full_matrices=True)
```

I played around with this and the `svds` call as above should return only the smallest singular value. This failed two test cases, though. In the overdetermined case, `V[-1]` is not the same `V[-1]` as when using a reduced SVD because `V.shape[0]` is greater than the number of singular values in A. If I restrict to using `svds` only in the non-overdetermined cases as above, then I only saw one precision-related failure in `test_polynomial_default_order`. It passes if I set `decimal=5` in the call `assert_almost_equal`.

I'm not sure if there is a better solution (I am not a linear algebra guru), but maybe this helps? I would be a little concerned that we don't have a ton of test cases, though, so I can't be super confident on the extent that precision changes might be an issue in real-world use cases.

grlee77: clarification: when always using the `svds` path two test cases fail. With the if/else as suggested above there is only the one precision-related failure that can be resolved by changing to `decimal=5` in the test.

grlee77:
This PR now passes all existing tests, but as commented at https://github.com/scikit-image/scikit-image/pull/5402#pullrequestreview-681017965, the new code path may have somewhat lower precision than the full SVD that was being used previously. I don't have a good way to guage whether this is likely to be a problem in practical applications. 

@jni, what do you think? A more conservative option would be to have a keyword-only parameter such as `svd_mode` that could be set to either `'full'` or `'sparse'` so that users could force fallback to the full SVD if there are precision issues. If precision is okay as-is, though, I would prefer not to add the additional parameter.

skimage/transform/_geometric.py
```diff
@@ -1446,8 +1448,13 @@ def estimate(self, src, dst, order=2):
...
         A[rows:, -1] = yd
-
-        _, _, V = np.linalg.svd(A)
+        
+        if A.shape[1] <= A.shape[0]:
+            # scipy sparse svd uses less memory
+             _, _, V = svds(A, k=1, which='SM', return_singular_vectors='vh')
+        else:
+            # use full SVD in the overdetermined case
+            _, _, V = np.linalg.svd(A, full_matrices=True)
```
ahojnnes, line 1457: We are importing scipy.linalg.svd above but are using np.linalg.svd here.

skimage/transform/_geometric.py
```diff
@@ -1446,8 +1448,13 @@ def estimate(self, src, dst, order=2):
 
         A[:rows, -1] = xd
         A[rows:, -1] = yd
-
-        _, _, V = np.linalg.svd(A)
+        
+        if A.shape[1] <= A.shape[0]:
+            # scipy sparse svd uses less memory
+             _, _, V = svds(A, k=1, which='SM', return_singular_vectors='vh')
```
ahojnnes, line 1454: In the sparse case, you might want to consider setting up A as a sparse matrix to further reduce memory usage.

matthew-brett:

<details>

## Rationale

</details>

PR #5420, opened 2021-06-08
Add no-reference perceptual blur metric.

mkcor:
@ladretp (Patricia Ladret) would like to submit this contribution: It's a function which estimates the strength of overall blur in an image. It is available in Matlab and it is quite popular, so we might as well have it in scikit-image!

Reference: https://doi.org/10.1117/12.702790

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


mkcor:
I didn't notice the second failure (23b8d30) locally, because I only ran a subset of the test suite (`$ pytest skimage/metrics`).

I should have noticed from the first failure (c5ba0f4) that I couldn't get away from the circular import issue:
'blur_effect' under `metrics` would depend on 'sobel_h' and 'sobel_v' from `filters` but those depend on 'laplacian' from `restoration` which in turn depends on 'mean_squared_error' from `metrics` :grimacing: 

Therefore, I decided to move the new function under `filters` (a6bbac4) -- after all, the blur effect is often related to the use of filters, so I think it makes sense that `blur_effect` would fall under `skimage.filters`. In addition, all `skimage.metrics` functions take two images as inputs (measuring the distance between the two images according to the given metric), so I think our new `blur_effect` function belongs to a different family.

/cc @ladretp

mkcor:
@ladretp I did 74f488b because "DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself." Also, `np.float64` is the default dtype (https://numpy.org/doc/stable/reference/generated/numpy.ones.html).

grlee77:
> I should have noticed from the first failure (c5ba0f4) that I couldn't get away from the circular import issue:
'blur_effect' under metrics would depend on 'sobel_h' and 'sobel_v' from filters but those depend on 'laplacian' from restoration which in turn depends on 'mean_squared_error' from metrics grimacing

Another way around the circular imports is to import the needed Sobel functions internally within `blur_effect` rather than at the top of the file.

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,85 @@
...
+
+    if image.ndim == 3:
+        image = rgb2gray(image)
+    image = img_as_float(image)
+    shape = image.shape
+
+    h = np.ones((h_size, 1)) / h_size
+
+    # vertical blur
+    ver = convolve(image, h)
```
grlee77, line 63:
```suggestion
    # vertical blur
    ver = ndi.uniform_filter1d(image, h_size, axis=0)
```

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,85 @@
...
+    im_blur = np.abs(sobel_h(ver))
+
+    T = np.fmax(np.zeros_like(image), im_sharp - im_blur)
+    M1 = np.sum(im_sharp[2:shape[0] - 1, 2:shape[1] - 1])
+    M2 = np.sum(T[2:shape[0] - 1, 2:shape[1] - 1])
+
+    By = np.abs((M1 - M2)) / M1
+
+    # horizontal blur
+    hor = convolve(image, h.transpose())
```
grlee77, line 74:
```suggestion
    hor = ndi.uniform_filter1d(image, h_size, axis=1)
```

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,85 @@
+import numpy as np
+from scipy.ndimage import convolve
```
grlee77, line 2:
```suggestion
import scipy.ndimage as ndi
```

skimage/filters/_blur_effect.py
grlee77: `uniform_filter` should be a bit more efficient than `convolve`. It uses a different implementation that reuses the same computations for the overlapping parts of the window on adjacent pixels (i.e. If the window size was 100, when you shift to the next adjacent pixel you just need to subtract a single value at one edge and add one value at the other edge rather than re-evaluating all 100 values in the window).

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,85 @@
...
+        image = rgb2gray(image)
+    image = img_as_float(image)
+    shape = image.shape
+
+    h = np.ones((h_size, 1)) / h_size
+
+    # vertical blur
+    ver = convolve(image, h)
+    im_sharp = np.abs(sobel_h(image))
+    im_blur = np.abs(sobel_h(ver))
```
grlee77, line 65: I prefer `sobel` with axis=0 or axis=1 to `sobel_h` or `sobel_v`. I think that will be less confusing and is easier to see how we could refactor to support n-dimensions.

grlee77 (changes requested):
I just started looking at this, but it seems that conceptually it should also work in n-dimensions? I left some initial review comments related to this

mkcor:
> Another way around the circular imports is to import the needed Sobel functions internally within `blur_effect` rather than at the top of the file.

Oh, within the function? Makes sense, thank you!

Thank you for reviewing, @grlee77 -- I'll include your suggestions and add a little demo (gallery example) as well.

mkcor:
> I just started looking at this, but it seems that conceptually it should also work in n-dimensions? I left some initial review comments related to this

Good point, let me check. I'll probably add `channel_axis=None` in the function's signature then.

mkcor:
I have rebased (fcabedf) because the email address I initially used for Patricia Ladret @ladretp was obsolete.

mkcor:
@grlee77 I added 82a16cc to avoid this 7-year-old warning message:
https://github.com/scikit-image/scikit-image/blob/9072919423e7aaad02e1664d02203607b9bd0bb2/skimage/filters/_gaussian.py#L113-L115
but maybe we'll update it in a follow-up PR to #5286? Am I right in taking the approach that the input image is grayscale by default (`channel_axis=None`), so the user should use the `channel_axis` parameter explicitly whenever the image is multichannel?

I know that we also have `channel_axis=-1` defaults throughout the API: My understanding is that this is the case for functions which typically apply to multichannel images, where `channel_axis=-1` means that the channel dimension is the final one by default (as opposed to any other dimension, not so much as opposed to the image being single-channel). I would like to make sure I'm in line with the latest choices in terms of API / philosophy.

skimage/filters/_blur_effect.py
mkcor: :tada: nD support: 09c0379e1121619229 please review :wink:

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,80 @@
...
+    h_size : int, optional
+        Size of the re-blurring filter. Default is 11.
+    channel_axis : int or None, optional
+        If None, the image is assumed to be grayscale (single-channel).
+        Otherwise, this parameter indicates which axis of the array
+        corresponds to color channels.
+
+    Returns
+    -------
+
```
grlee77, line 31:
```suggestion
```

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,80 @@
...
+       no-reference perceptual blur metric" Proc. SPIE 6492, Human Vision and
+       Electronic Imaging XII, 64920I (2007)
+       https://hal.archives-ouvertes.fr/hal-00232709
+       :DOI:`10.1117/12.702790`
+    """
+    # number of non-channel axes
+    n_axes = image.ndim
+
+    if channel_axis is not None:
+        if not isinstance(channel_axis, int) or channel_axis >= image.ndim:
```
grlee77, line 59: This should also error on `channel_axis < -image.ndim`, but I'm not sure we need to check these at all or just let `np.moveaxis` handle it? It seems `moveaxis` would return a `TypeError` for non-integer `channel_axis` and a `numpy.AxisError` if the axis is out of range.

The main benefit of us raising the error is that we can make a more user-friendly error message that mentions `channel_axis` by name instead of, for example:
```Python
np.moveaxis(np.random.randn(16, 32), 1.5, -1)

  File "<ipython-input-53-0100d5f06b0f>", line 1, in <module>
    np.moveaxis(np.random.randn(16, 32), 1.5, -1)

  File "<__array_function__ internals>", line 5, in moveaxis

  File "/home/lee8rx/mambaforge/envs/skimage_dev/lib/python3.9/site-packages/numpy/core/numeric.py", line 1461, in moveaxis
    source = normalize_axis_tuple(source, a.ndim, 'source')

  File "/home/lee8rx/mambaforge/envs/skimage_dev/lib/python3.9/site-packages/numpy/core/numeric.py", line 1391, in normalize_axis_tuple
    axis = tuple([normalize_axis_index(ax, ndim, argname) for ax in axis])

TypeError: 'float' object is not iterable
```

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,80 @@
...
+    image = img_as_float(image)
+    shape = image.shape
+    B = []
+
+    for a in range(n_axes):
+        filt_im = ndi.uniform_filter1d(image, h_size, axis=a)
+        im_sharp = np.abs(sobel(image, axis=a))
+        im_blur = np.abs(sobel(filt_im, axis=a))
+        T = np.fmax(np.zeros_like(image), im_sharp - im_blur)
+        slices = [slice(2, shape[a] - 1) for a in range(n_axes)]
```
grlee77, line 75:
```suggestion
        slices = [slice(2, shape[ax] - 1) for ax in range(n_axes)]
```
`a` was already used for the outer for loop variable, so it seems safer to use a separate name here

Where does this slice excluding a single edge voxel come from? Is it due to the "radius" of the sobel kernel?

grlee77 (changes requested):
This looks good to me aside from a few minor comments.

Nice use of plotly to be able to scroll through the blurred images.

The tests that are here look good. If there is a reference implementation that is commonly used in the literature that we want to ensure a close match to (I think you mentioned a Matlab implementation) we could try to save a result from that third party library for comparison to ensure we give consistent results with previously published studies.

ladretp:
Dear Gregory

I'm one of the author of this paper and I propose this function here 
because now I'm using python scikit-image, -learn etc...

  The matlab file is here:

https://fr.mathworks.com/matlabcentral/fileexchange/24676-image-blur-metric?s_tid=srchtitle

BUT and that's why I wanted to rewrite the function, this one was not 
written by one of us (the authors) but by a nice unknown person who did 
not even warn us (we discovered the function here by chance). This 
function does not include the treatment of the edges of the image, and 
especially the derivative is calculated very roughly, that's why I 
wanted to use Sobel. Moreover, and this is one of the justified 
criticisms of some users, one cannot change the size of the  filter 
(size_h) inside the function and when the image has very important blur 
it does not give convincing results...

I'm very sorry Marianne is implementing for me in scikit-image this 
function because I don't know the github process ..., thanks to her!

Patricia

Le 10/06/2021 à 22:28, Gregory R. Lee a écrit :

grlee77:
Thanks for the context, @ladretp! Based on this, let's not try to make a comparison to the existing Matlab implementation. I am fine with the existing tests as-is.

skimage/filters/_blur_effect.py
mkcor: Good point! I've decided to combine both, so the error starts with a message that mentions `channel_axis`: 752f5d7448f89b143214c9556141cfca26a61497 What do you think?

skimage/filters/_blur_effect.py
```diff
@@ -54,13 +54,17 @@ def blur_effect(image, h_size=11, channel_axis=None):
...
+        try:
             # ensure color channels are in the final dimension
             image = np.moveaxis(image, channel_axis, -1)
-            image = rgb2gray(image)
+        except np.AxisError:
+            print('channel_axis must be one of the image array dimensions')
+            raise
+        except TypeError:
+            print('channel_axis must be an integer')
+            raise
```
grlee77, line 65: Thanks, this looks fine to me. I forget which version of NumPy first introduced `np.AxisError`, but I think it is earlier than our minimum supported version (indeed CI passed for the minimal dependencies case).

grlee77 (approved):

skimage/filters/_blur_effect.py
mkcor: Right, our minimum supported version for NumPy is `1.17.0` and `AxisError` was already available in `1.13.0`: https://numpy.org/doc/stable/release/1.13.0-notes.html?highlight=axiserror#compatibility-notes

doc/examples/filters/plot_blur_effect.py
```diff
@@ -0,0 +1,83 @@
...
+# blur metric applies to single-channel images.
+
+image = data.astronaut()
+image = color.rgb2gray(image)
+
+#####################################################################
+# Let us blur this image with a series of uniform filters of increasing size.
+
+N = 32
+blurred_images = [ndi.uniform_filter(image, size=k) for k in range(2, N, 2)]
```
rfezzani, line 45:
```suggestion
blurred_images = [ndi.uniform_filter(image, size=k) for k in range(2, 32, 2)]
```
`N` is not really a good parameter name and it is only used here, I think that we can avoid it ;)

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,82 @@
...
+    Compute a metric that indicates the strength of blur in an image (0 for no
+    blur, 1 for maximal blur).
+
+    Parameters
+    ----------
+    image : ndarray
+        RGB or grayscale nD image. The input image is converted to grayscale
+        before computing the blur metric.
+    h_size : int, optional
+        Size of the re-blurring filter. Default is 11.
```
rfezzani, line 23:
```suggestion
        Size of the re-blurring filter.
```
We usually don't specify the default value in the docstring.

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,82 @@
...
+            # ensure color channels are in the final dimension
+            image = np.moveaxis(image, channel_axis, -1)
+        except np.AxisError:
+            print('channel_axis must be one of the image array dimensions')
+            raise
+        except TypeError:
+            print('channel_axis must be an integer')
+            raise
+        image = rgb2gray(image)
+        n_axes = n_axes - 1
```
rfezzani, line 67: Can't we define `n_axes` after the `channel_axis` management below to avoid this?
```suggestion
    n_axes = image.ndim
```

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,82 @@
...
+
+    for a in range(n_axes):
+        filt_im = ndi.uniform_filter1d(image, h_size, axis=a)
+        im_sharp = np.abs(sobel(image, axis=a))
+        im_blur = np.abs(sobel(filt_im, axis=a))
+        T = np.maximum(0, im_sharp - im_blur)
+        slices = [slice(2, shape[ax] - 1) for ax in range(n_axes)]
+        M1 = np.sum(im_sharp[tuple(slices)])
+        M2 = np.sum(T[tuple(slices)])
+        B.append(np.abs((M1 - M2)) / M1)
```
rfezzani, line 80: `slices` can be computed outside the `for` loop, allowing to give a better argument name then `a` :wink: 
```suggestion
    slices = tuple([slice(2, s - 1) for s in shape])
    for ax in range(n_axes):
        filt_im = ndi.uniform_filter1d(image, h_size, axis=ax)
        im_sharp = np.abs(sobel(image, axis=ax))
        im_blur = np.abs(sobel(filt_im, axis=ax))
        T = np.maximum(0, im_sharp - im_blur)
        M1 = np.sum(im_sharp[slices])
        M2 = np.sum(T[slices])
        B.append(np.abs((M1 - M2)) / M1)
```

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,82 @@
...
+
+    return np.max(B), B
```
rfezzani, line 82: What about introducing a `metric` parameter which is a reduction function (with default to `np.max`), to be applied to `B`?
This allows to use what ever reduction to `B` (`mean`, `std`...), to return plain `B` when it is set to `None` and avoids returning `B` along with its maximum as it is made here?
```suggestion
    return B if metric is None else metric(B)
```

skimage/filters/_blur_effect.py
```diff
@@ -0,0 +1,82 @@
...
+    ----------
+    .. [1] Frederique Crete, Thierry Dolmiere, Patricia Ladret, and Marina
+       Nicolas "The blur effect: perception and estimation with a new
+       no-reference perceptual blur metric" Proc. SPIE 6492, Human Vision and
+       Electronic Imaging XII, 64920I (2007)
+       https://hal.archives-ouvertes.fr/hal-00232709
+       :DOI:`10.1117/12.702790`
+    """
+    # number of non-channel axes
+    n_axes = image.ndim
```
rfezzani, line 54:
```suggestion
```

Can be declared after the following block :wink:

rfezzani (commented):
Thank you @mkcor, I left few suggestions :wink: .

doc/examples/filters/plot_blur_effect.py
mkcor: Good catch! Thanks.

skimage/filters/_blur_effect.py
mkcor: Right, the upfront definition of `n_axes` was legacy from earlier versions.

skimage/filters/_blur_effect.py
mkcor: Right!

skimage/filters/_blur_effect.py
mkcor: Ok.

skimage/filters/_blur_effect.py
mkcor: Plain `B` is available as the second returned value, so the user can always access it and aggregate it:
```py
_, plain_B = blur_effect(my_image)
my_blur_metric = np.mean(plain_B)
```
for example. The 'default' to `np.max` is based on the referenced paper: https://hal.archives-ouvertes.fr/hal-00232709/document where the metric is validated against human perception.

rfezzani: Yes, I see that plain ``B` is returned, and that's what I find not elegant! Most users will use only one of the two returned values...
So I though about something similar to the `reduce_func` argument in [skimage.measure.profile_line](https://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.profile_line)....

mkcor: Ok, I see! I admit that returning multiple values isn't the most elegant... Let me add an optional `reduce_func` then!

rfezzani: Excellent, thank you @mkcor :wink:

mkcor: Here it is: 62dd792f0870b2f2d8cf33b1d61bf2a18e5515a8 :slightly_smiling_face: 
Thank you for the thorough review, @rfezzani, it's been really helpful! :muscle:

rfezzani (commented):
Thank you @mkcor, the code looks find to me, but I wonder if this function mustn't take place in `skimage.measure` instead of `skimage.filter`...

mkcor:
> Thank you @mkcor, the code looks fine to me, but I wonder if this function mustn't take place in `skimage.measure` instead of `skimage.filter`...

Oh, good question! We are *measuring* the strength of blur indeed... Somehow I didn't think of `skimage.measure` when [moving away](https://github.com/scikit-image/scikit-image/pull/5420#issuecomment-857609073) from `skimage.metrics`! I'm totally open to moving the blur metric function under `skimage.measure`; do you think the gallery example can still fall under `doc/examples/filters/`?

rfezzani:
Sorry, I didn't noticed the move you made from `metrics` to `filters` :confused:... But yes, I think that `measure` is a better place for this function :wink: 
If you decide to move `blur_effect` from `filter` to `measure`, there is no point to keep the gallery example in the `doc/examples/filters` directory since the `filters` module will no more be imported in the script....

mkcor:
> If you decide to move `blur_effect` from `filter` to `measure`, there is no point to keep the gallery example in the `doc/examples/filters` directory since the `filters` module will no more be imported in the script.... But this is not a big deal :)

Well, I would tend to have a more 'semantic' approach to the gallery organization. After all, there is no strict match (bijection) between the `skimage` submodules and the gallery's subdirectories:  https://github.com/scikit-image/scikit-image/tree/main/doc/examples

I don't think this example could fall under `applications/` nor `features_detection/`... definitely not `edges/` nor `segmentation/`... Most gallery examples importing `measure` seem to be found under `transform/`: Do you think it would make sense there?

rfezzani:
It is really not a big deal, the gallery example can stay without any hurt in `doc/examples/filters` :wink:

mkcor:
> Another way around the circular imports is to import the needed Sobel functions internally within blur_effect rather than at the top of the file.

In the end, that's what I did in 766eb182e3e9733f7221141768a9a145e9a7e46d ;)

rfezzani (approved):
Thank you @mkcor, merging :tada:

PR #5424, opened 2021-06-10
Add benchmark checks to CI

jaimergp:
## Description

This PR provides a new CI workflow that runs the current commit against the PR target branch. The different strategies are discussed in the conversation below, which conclusions driven by data points collected in my fork.

It also adds a `asv check` command to the standard CI suite.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


jaimergp:
Noting here that some of the benchmarks included with `asv dev` will result in memory allocation errors. It takes 30 minutes to run a quick pass. Might only be worth it for _some_ entries in the job matrix. I'll let the team guide that decision... thoughts?

grlee77:
Thanks, I just merged #5426 and manually started the CI here. We will see if it autostarts on the next commit now...

jaimergp:
That seemed to do the trick, thanks @grlee77!

grlee77:
FYI, if you are seeing warnings about a deprecated `multichannel` argument, those will be resolved by #5427 for the 0.19 release, see:
https://github.com/scikit-image/scikit-image/pull/5427/commits/2c6e7215f0e43ce96cd08c2e9b8c3c4738f87e95

Basically, we changed the API to allow specification of the axis containing "channels" for color images rather than using a boolean `multichannel` argument. For 0.19 both arguments are still present, but the user will get a warning if `multichannel` is used. To be able to benchmark against older releases, I added this `_channel_kwarg` helper function

jaimergp:
I'm going to aim for relative performance measurements in GH Actions. I've seen several projects being _ok_ with this approach so at least the catch regressions on PRs. We won't get a history of performance, but that won't be possible without dedicated hardware.

I've merged a similar [PR in my own fork](https://github.com/jaimergp/scikit-image/actions/runs/928770526) that will run the benchmarks every 6h for a week for the same two commits. I will then collect the results (uploaded via artifacts) and compare how reliable they are. We should expect the same relative differences (+- some statistical deviation) in all runs if the approach is stable enough.

I'll ping you when I have some numbers next week!

grlee77:
> I've merged a similar PR in my own fork that will run the benchmarks every 6h for a week for the same two commits. I will then collect the results (uploaded via artifacts) and compare how reliable they are. We should expect the same relative differences (+- some statistical deviation) in all runs if the approach is stable enough.

Nice. This will be interesting to see!

jaimergp:
This notebook (attached) will help us assess the stability of the GHA CI for benchmarking purposes. We still need to wait a bit to collect more data points, though, but it looks ok so far!

[timeseries.ipynb.zip](https://github.com/scikit-image/scikit-image/files/6642074/timeseries.ipynb.zip)

jaimergp:
During the weekend we have collected 15 independent attempts. See analysis in [this gist](https://gist.github.com/jaimergp/aa4f059c14e394c4089b320cb8b51b1a).

Reliability is okayish, but it does give some false positives here and there... Let's see how it behaves over the weekdays instead of the weekend.

jaimergp:
Interim update, looks like we are obtaining behaviours within the expectations of "stability"?

image

[Updated notebook here](https://gist.github.com/jaimergp/aa4f059c14e394c4089b320cb8b51b1a).

grlee77:
Yeah, this looks like we would be able to detect substantial performance regressions (>20 % difference or so). 

One concern is with the duration it takes to run this. It looks like close to two hours when I looked at your branch. This is a little concerning because we would ideally expand the benchmarks significantly (many functions are not currently covered), so we may run into timeout issues at some point. Do you know how many averages ASV is doing by default? Perhaps it is higher than necessary given the other sources of variability when running via GHA. I know there is a `--quick` flag that forces only a single repetition, but was wondering if there is a way to still do more than 1, but less than whatever the default is?

Have you considered a way to only run this upon request (e.g. by a comment such as `run benchmarks` posted by a maintainer in the PR). This would preserve CI resources for cases like documentation PRs or newly introduced functions, where this comparison would not be helpful.

jaimergp:
Thanks for the feedback!

With respect to the runtime:

* At least 30 mins is spent building scikit-image twice (baseline vs contender).
* The `repeat` [attribute](https://asv.readthedocs.io/en/stable/benchmarks.html#timing-benchmarks) can be tuned with the `-a` flag. By default:

> When not provided (repeat set to 0), the default value is (1, 10, 20.0) if processes==1 and (1, 5, 10.0) otherwise.

* Running the benchmarks on request is definitely doable but will require some tuning with the triggers and conditionals (which can be overridden by the PR submitter, but I think that's ok since it's not like we are providing a self-hosted machine or anything). We can think of comments (which require an ok-list) or, even better, labels (which are permission-restricted already). The results can be posted as a comment.

jaimergp:
Results from the week are ready. Looks like it's stable enough! I suggest a default threshold ratio of `1.4` for now:

image (4)

***

Whether we are sticking to the default running configuration (2h per run) or not, I don't think we'll get it down to less than an hour, so I guess we'll have to manually run the workflow with one of these options:

* Have GHA listen for specific comments coming from a maintainer.
* Have GHA listen for a [label added to the PR](https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label). Re-run by removing and re-adding.
* Use `workflow_dispatch` to have a manual button on the `Actions` tab. You can choose the branch, but I am not sure this includes forks/PRs.

I'd strongly suggest using the labels approach. Only maintainers can add labels, so authentication is not done on the GHA side, but GH UI. I don't know how this is reported in the UI, but I guess we'll have the GHA bot post a comment with a link to the logs.

For now, I am going to add a couple of commits to a PR in my fork to experiment with the labels approach.

jaimergp:
I had to do some research for the auto-comment capabilities and decided on having two jobs handling this, because two kinds of event are needed due to permission issues. That's why you'll see `Benchmark / Linux` and `Benchmark / Report`.

Why are two jobs needed?

* `$GITHUB_TOKEN` has read-only permissions on PRs opened from a fork if the `pull_request` event is used as a trigger. This prevents the workflow from posting a comment in the PR.
* However, the variant `pull_request_target` trigger gives `GITHUB_TOKEN` some write capabilities, enough for posting comments. The downside is that the repo is cloned in a different way (no merged commit), and `asv` dislikes that. Also, it's a security risk. As a result, jobs triggered by `pull_request_target` won't run in the workflow file they are defined in has been modified as part of the PR (that's the case here). I guess I'll have to test in a PR against my fork.

Both jobs will be triggered (eventually) when a label `run-benchmarks` is added to the PR. `Linux` will run the benchmark suite, and `Report` will post a comment with the URL to the logs. Note that it will post the link once the job _starts_ not _ends_. I'll need to investigate if we can add "job is done" comments two under this dual event system.

I'll keep you updated!

jaimergp:
Ok, the dual trigger thing [kind of works](https://github.com/jaimergp/comment-pr-fork/pulls), and I also realized the auto comment is probably not critical, because it does show up in the UI, so we might just get rid of it for simplicity. Getting the URL of a job triggered in the same workflow+push but different event requires [cumbersome JS](https://github.com/jaimergp/comment-pr-fork/blob/main/.github/workflows/demo.yml#L33-L72) that I am not too proud of...

grlee77:
> At least 30 mins is spent building scikit-image twice (baseline vs contender).

Maybe [Ccache](https://ccache.dev/) can help with this? The majority of files are unmodified in any given PR so I would guess it should reduce compile times substantially.

Using Ccache (and caching `$CCACHE_DIR` across runs on CI) for our GHA and Azure CI test runs might be worth pursuing in a separate PR. I tried looking if NumPy and SciPy use it and I do see it in [this Azure config from SciPy](https://github.com/scipy/scipy/blob/master/ci/azure-travis-template.yaml). Both also use a `runtests.py` which prepends `/usr/lib/ccache` and `/usr/local/lib/ccache` to the path.

grlee77:
Using a label instead of a comment seems fine to me.

jaimergp:
* CCache is implemented and running. Seems to work nicely: running a single test, everything takes 13mins!
* Since compilation is not _that_ slow, I have looked into the runtimes of the cron logs to find the slowest tests:

image (5)


Top 25:

```
benchmark_restoration.RollingBall.time_rollingball_ndim                0 days 00:02:00.076000
benchmark_transform_warp.WarpSuite.time_same_type                      0 days 00:01:59.113000
benchmark_transform_warp.WarpSuite.time_to_float64                     0 days 00:01:58.896000
benchmark_restoration.RollingBall.time_rollingball_threads             0 days 00:01:57.447000
benchmark_util.NoiseSuite.time_salt_and_pepper                         0 days 00:01:47.096000
benchmark_restoration.RollingBall.time_rollingball                     0 days 00:01:24.114000
benchmark_rank.Rank3DSuite.time_3d_filters                             0 days 00:01:00.180000
benchmark_restoration.RollingBall.time_rollingball_nan                 0 days 00:00:58.531000
benchmark_restoration.RestorationSuite.time_denoise_nl_means_f32       0 days 00:00:54.205000
benchmark_restoration.RestorationSuite.time_denoise_nl_means_f64       0 days 00:00:52.591000
benchmark_interpolation.InterpolationResize.time_resize                0 days 00:00:52.134000
benchmark_interpolation.InterpolationResize.time_rescale               0 days 00:00:51.970000
benchmark_restoration.RollingBall.peakmem_rollingball                  0 days 00:00:35.032000
benchmark_restoration.RestorationSuite.peakmem_denoise_nl_means_f32    0 days 00:00:27.259000
benchmark_restoration.RestorationSuite.peakmem_denoise_nl_means_f64    0 days 00:00:26.460000
benchmark_transform_warp.ResizeLocalMeanSuite.time_resize_local_mean   0 days 00:00:20.623000
benchmark_segmentation.MaskSlicSegmentation.time_mask_slic             0 days 00:00:19.403000
benchmark_util.NoiseSuite.peakmem_salt_and_pepper                      0 days 00:00:19.038000
benchmark_filters.MultiOtsu.time_threshold_multiotsu                   0 days 00:00:16.150000
benchmark_filters.FiltersSobel3D.time_sobel_3d                         0 days 00:00:14.881000
benchmark_rank.RankSuite.time_filter                                   0 days 00:00:14.124000
benchmark_registration.RegistrationSuite.time_tvl1                     0 days 00:00:13.546000
benchmark_restoration.DeconvolutionSuite.time_richardson_lucy_f64      0 days 00:00:12.587000
benchmark_interpolation.InterpolationResize.peakmem_resize             0 days 00:00:12.309000
benchmark_segmentation.SlicSegmentation.time_slic_basic                0 days 00:00:12.283000
dtype: timedelta64[ns]
```

I think we can optimize some those slow ones by setting the relevant `repeat` and `number` attributes only in those (or maybe reducing the params matrix).

jaimergp:
I was confused by the timings between this PR and my fork, since this one takes way LONGER. So I dug the artifacts for this one and guess what:

```
benchmark_transform_warp.ResizeLocalMeanSuite.time_resize_local_mean   0 days 00:43:20.942000
benchmark_restoration.RollingBall.time_rollingball                     0 days 00:02:19.772000
benchmark_restoration.RollingBall.time_rollingball_ndim                0 days 00:02:00.094000
benchmark_restoration.RollingBall.time_rollingball_threads             0 days 00:01:56.193000
benchmark_transform_warp.WarpSuite.time_same_type                      0 days 00:01:51.384000
benchmark_transform_warp.WarpSuite.time_to_float64                     0 days 00:01:50.124000
benchmark_util.NoiseSuite.time_salt_and_pepper                         0 days 00:01:43.611000
benchmark_restoration.RollingBall.time_rollingball_nan                 0 days 00:01:08.216000
benchmark_restoration.RollingBall.peakmem_rollingball                  0 days 00:01:04.001000
benchmark_rank.Rank3DSuite.time_3d_filters                             0 days 00:00:55.299000
benchmark_interpolation.InterpolationResize.time_rescale               0 days 00:00:50.096000
benchmark_interpolation.InterpolationResize.time_resize                0 days 00:00:50.084000
benchmark_restoration.RestorationSuite.time_denoise_nl_means_f32       0 days 00:00:46.976000
benchmark_restoration.RestorationSuite.time_denoise_nl_means_f64       0 days 00:00:45.181000
benchmark_restoration.RestorationSuite.peakmem_denoise_nl_means_f32    0 days 00:00:24.417000
benchmark_restoration.RestorationSuite.peakmem_denoise_nl_means_f64    0 days 00:00:22.713000
benchmark_util.NoiseSuite.peakmem_salt_and_pepper                      0 days 00:00:18.089000
benchmark_segmentation.MaskSlicSegmentation.time_mask_slic             0 days 00:00:17.864000
benchmark_filters.MultiOtsu.time_threshold_multiotsu                   0 days 00:00:14.151000
benchmark_filters.FiltersSobel3D.time_sobel_3d                         0 days 00:00:14.055000
benchmark_rank.RankSuite.time_filter                                   0 days 00:00:13.759000
benchmark_registration.RegistrationSuite.time_tvl1                     0 days 00:00:12.190000
benchmark_registration.RegistrationSuite.time_ilk                      0 days 00:00:11.888000
benchmark_segmentation.MaskSlicSegmentation.time_slic_basic            0 days 00:00:11.297000
benchmark_restoration.DeconvolutionSuite.time_richardson_lucy_f64      0 days 00:00:11.286000
Name: 0, dtype: timedelta64[ns]
```

43 mins in a single test? This is the one I fixed the function signature... Any idea why this particular is so slow? I am going to add a `timeout = 120` attribute for now, but this needs to be investigated.

jaimergp:
> I am going to add a timeout = 300 attribute for now, but this needs to be investigated.

Note to self: attributes only apply per permutation. So the timeout is for each individual combination of params. There's no global timeout (class-wide). 

Fortunately, one can skip tests by raising `NotImplementedError` in `setup`, so I am going to trim the matrix a bit (specially given that 90 out of 384 tests for `ResizeLocalMeanSuite` won't run due to insufficient memory). This should bring the time down substantially, but we need to decide which tests are skippable in this one.

For now, I've chosen these:

```python
class ResizeLocalMeanSuite:
    params = ([np.uint8, np.uint16, np.float32, np.float64],
              [128, 512, 1024, 2048],
              [128, 512, 1024],
              [2, 3],
              [2, 3],
              [True, False]
              )
    param_names = ['dtype', 'shape_in', 'shape_out', 'ndim_in',
                   'ndim_out', 'grid_mode']

    def setup(self, dtype, shape_in, shape_out, ndim_in, ndim_out, grid_mode):
        is_slow = any((dtype == np.float64, shape_in == 2048,
                       shape_out == 1024, ndim_in == 3, ndim_out == 3))
        if os.environ.get("SKIP_SLOW") == "1" and is_slow:
            m = "Skipping in CI because of memory usage and time constraints"
            raise NotImplementedError(m)
...
```

jaimergp:
With `ccache` and less entries in the `ResizeLocalMeanSuite` the suite takes 1h30-1h45 to run! I am running some other configuration tests in the fork to see if we can reduce it further.

By default, `asv` uses two processes, interleaved (not parallel), to reduce the measurement noise. I wonder how important this is in our setting and if we can do with one pass instead of two. I am also trying with no interleaving to see how much overhead that adds, and if it's worth it reliability wise.

grlee77:
Yeah, it is definitely not reasonable to spend so much more time on `ResizeLocalMeanSuite`, so we should address that by reducing cases and/or sizes. 

I would suggest testing three shapes (512, 512), (2048, 2048) and (192, 192, 192). This function always promotes to a floating point internally so all int types end up as float64 anyway. Given that, I would reduce dtypes to just float32 and float64. `grid_mode` should not make any substantial difference so I wouldn't parameterize over it. I am happy to make a PR for that separate from this one if you like.

I think a general rule of thumb should be to try testing with both an image small enough to fit in L2 cache of the CPU and one that is too large to fit in that cache because those cases can behave differently, but we shouldn't need to test a huge range of shapes and dtypes.

jaimergp:
Excellent!

I have reduced the params matrix in those tests that were taken more than two minutes each. That should bring the time down a bit. I am also checking how much we can gain from using a single process instead of two, and whether that involves a sensitivity reduction. So far, my tests reveal that a single process takes ~1h, while the default two-process scheme sits around ~1h45. I don't have enough data points to comment about sensitivity, but it looks okayish so far.

jaimergp:
I've run some tests to see how important is to use two subsequent passes/processes for the reliability vs spent time compromises. These are the preliminary results after 11 data points:

## Duration

Mean of 11 runs

* Two processes, interleaved (default):  01:53:42
* Two processes, not interleaved: 01:43:24
* Single process: 01:06:55

## Reliability

Two processes, interleaved (default)

image

Two processes, not interleaved

image

Single process

image

## Preliminary conclusions

A single-process run takes ~45min less to finish, with some added deviation in the ratios. That said, the observed data points never went over 1.5. Is this a compromise the `scikit-image` team is willing to make for the sake of speed or would you rather have a longer, subtly more stable measurement?

jaimergp:
I am investigating now if doing two replicas of a single-process job brings the best out of both worlds. Let's see! Each single-process job takes <60mins!

jaimergp:
Some more data points from the weekend:

Default (two processes interleaved):

image


Single process:

image


***

Nothing new. A single process is less reliable but, in my opinion, still good enough for the purposes here. Since we are running two replicas in parallel (and we can set more, if needed), I'd **only consider a regression when the same test has been caught by all replicas**. Otherwise, it is probably a false positive.

Would the team be ok with that approach? If that's the case, I will change the triggers so it only works on labels and `workflow_dispatch`.

grlee77:
Hi @scikit-image/core, @jaimergp has been doing some great detective work on the reliability of running benchmarks on cloud hardware.

I think the primary questions now are (Jaime, please correct/clarify if I missed something!): 

1.) Based on the variability plotted in https://github.com/scikit-image/scikit-image/pull/5424#issuecomment-868323266, we cannot reliably detect changes < 25% using non-dedicated hardware, but larger changes (>50% or so) should be fairly easy to detect. We can run the benchmarks in a single process in ~ 1 hour and if we run two of these in parallel we can rule out most false positives. Does this seem reasonable to the team?

2.) Jaime proposes to run these only when a maintainer adds a specific label to the PR. This avoids excessive CI usage from needlessly running benchmarks on, for example, documentation PRs

stefanv:
That approach sounds good to me.

We could consider running it once a week, since we're only catching big regressions anyway?

jaimergp:
> We could consider running it once a week

Like on a cronjob? I was thinking about this too. There's a risk that little regressions are not caught individually as they occur in PRs, but that the accumulated slowdown with respect to the last release might be significant. To account for that, I think it'd be interesting to run a weekly benchmark from current `main` wrt to the last release. Would that make sense in addition to the per-PR checks triggered by labels?

jaimergp:
Something to consider about having two single-process replicas vs a single two-process job: the spread is quite different.

image

image

In some cases, the single process went above 2.5! Although this variation should be diminished by the replicas, maybe the maintainers would prefer to have a single, more reliable job running, even if it's for longer time each?

grlee77:
> In some cases, the single process went above 2.5! Although this variation should be diminished by the replicas, maybe the maintainers would prefer to have a single, more reliable job running, even if it's for longer time each?

Yeah, having a single, longer job should be fine given that we aren't going to only run it weekly or on request. What is the maximum allowed job duration? (i.e. if we expand the benchmarks going forward how likely are we to exceed it)

jaimergp:
GHA will allow you 6h per job, which should be more than enough. If it ever reaches that limit, we can split the benchmark in several jobs so they deal with a handful of tests each.

jaimergp:
Ok team, I think you can try the label triggers now!

Add a label named `run-benchmark` to this PR and check whether the Benchmark job runs!

stefanv:
It worked!  How do I look at the artifact via asv?  It may be worth adding a README.md into the artifact zip file with instructions.

jni:
Super amazing work @jaimergp! I agree with @stefanv that some instructions would be great. 😃 

By the way, I just removed/added the label and that worked too! So great! 🎉 🚀

jni (approved):

jaimergp:
> How do I look at the artifact via asv? It may be worth adding a README.md into the artifact zip file with instructions.

Got it! I'll add some instructions!

jaimergp:
@jni @stefanv -- I added some instructions [here](https://github.com/jaimergp/scikit-image/blob/ci-benchmark-check/benchmarks/README_CI.md). This file will be included in the artifact, together with the CI logs and the JSON databases. Let me know if that's clear enough or if you need something else!

stefanv:
> @jni @stefanv -- I added some instructions [here](https://github.com/jaimergp/scikit-image/blob/ci-benchmark-check/benchmarks/README_CI.md). This file will be included in the artifact, together with the CI logs and the JSON databases. Let me know if that's clear enough or if you need something else!

This is... AMAZING?!  Really, really nicely done.  Thank you!

stefanv (approved):

benchmarks/README_CI.md
```diff
@@ -0,0 +1,122 @@
...
+    * `ratio=1.0` -> performance didn't change.
+    * `ratio<1.0` -> PR made it slower.
+    * `ratio>1.0` -> PR made it faster.
+
+Due to the sensitivity of the test, we cannot guarantee that false positives are not produced. In practice, values between `(0.7, 1.5)` are to be considered part of the measurement noise. When in doubt, running the benchmark suite one more time will provide more information about the test being a false positive or not.
+
+## Running the benchmarks on GitHub Actions
+
+1. On a PR, add the label `run-benchmark`.
+2. The CI job will be started. Checks will appear if the usual dashboard panel above the comment box.
```
grlee77, line 25:
```suggestion
2. The CI job will be started. Checks will appear in the usual dashboard panel above the comment box.
```

grlee77 (approved):
Thanks @jaimergp, the new docs are both clear and comprehensive. I had just one suggestion to correct a typo.

jaimergp:
Thanks for the kind words @stefanv @grlee77! I think we should run the benchmarks one more time before the merge so we can check the contents of the artifact. Can you please add the label one more time? Thanks!

grlee77:
> Can you please add the label one more time? Thanks!

Done

jaimergp:
Ok, confirmed! The files are there as expected! 🥳

grlee77:
Hi @jaimergp, I am seeing many more Azure CI failures after this PR. There seem to be two separate issues related to this commit:
 https://github.com/scikit-image/scikit-image/commit/2fa66e57f7417314adbcfec6a20e3de8dcffcaa6#diff-7915b9b726a397ae7ba6af7b9703633d21c031ebf21682f3ee7e6a4ec52837a5

Removing the condition on running the gallery examples has resulted in all jobs timing out now, although the timeouts should be resolved if #5446 is merged. 

The second issue is that a couple of the cases fail to build wheels for (dependencies of) `sphinx-gallery`. I am not sure of the best fix for that one: (see [here](https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=6233&view=logs&jobId=0c528f75-1409-556f-1999-9f155ac37a3a&j=0c528f75-1409-556f-1999-9f155ac37a3a&t=f5e66b6e-42ae-5fb8-4ac0-6e2189f512a2))

The specific error is
```
  error: failed to run custom build command for `pyo3 v0.13.2`
  
  Caused by:
    process didn't exit successfully: `C:\Users\VSSADM~1\AppData\Local\Temp\pip-install-vc714uet\pywinpty_c3cbae2a690e484a91f2810baf619595\target\release\build\pyo3-93fbd3a6215435b6\build-script-build` (exit code: 1)
    --- stderr
    Error: "Your Rust target architecture (64-bit) does not match your python interpreter (32-bit)"
```

It is probably easiest to just restore the conditional running of the gallery examples, but let me know what you think.

jaimergp:
Oh, wow, that's totally an accident on my end. I shouldn't have removed that conditional. Will open a PR to revert now.

Not sure about the 2nd issue.

grlee77:
> Not sure about the 2nd issue.

Yeah, I am not familiar with that package it is trying to build, but it seems to be something pulled in by sphinx-gallery. It might work if we selected a 64-bit Python interpreter, but I don't know. I think those failure cases will be bypassed by restoring the conditional in any case.

PR #5425, opened 2021-06-10
Fix duplicate test function names in test_unsharp_mask.py

grlee77:
## Description

This PR fixes a copy-paste error introduced in one of my `channel_axis` PRs. There were two test cases named `test_unsharp_masking_with_different_ranges_deprecated`, but the first was supposed to be named `test_unsharp_masking_with_different_ranges` because it is using the non-deprecated arguments.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/filters/tests/test_unsharp_mask.py
+++ b/skimage/filters/tests/test_unsharp_mask.py
@@ -60,9 +60,8 @@ def test_unsharp_masking_with_different_radii(radius, shape,
               ((3, 13, 17), 0)])
 @parametrize("offset", [-5, 0, 5])
 @parametrize("preserve", [False, True])
-def test_unsharp_masking_with_different_ranges_deprecated(shape, offset,
-                                                          channel_axis,
-                                                          preserve):
+def test_unsharp_masking_with_different_ranges(shape, offset, channel_axis,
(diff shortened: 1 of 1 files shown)


skimage/filters/tests/test_unsharp_mask.py
```diff
@@ -60,9 +60,8 @@ def test_unsharp_masking_with_different_radii(radius, shape,
               ((3, 13, 17), 0)])
 @parametrize("offset", [-5, 0, 5])
 @parametrize("preserve", [False, True])
-def test_unsharp_masking_with_different_ranges_deprecated(shape, offset,
-                                                          channel_axis,
-                                                          preserve):
+def test_unsharp_masking_with_different_ranges(shape, offset, channel_axis,
```
grlee77, line 63: The other test with the duplicate name was the one at line 87 below.

grlee77: `pytest` would have been running only one of the two rather than both!

mkcor (approved):

rfezzani (approved):
Good catch @grlee77 , thank you :wink:

PR #5426, opened 2021-06-11
Benchmarks: fix ResizeLocalMeanSuite.time_resize_local_mean signature

jaimergp:
## Description

Running `asv check` will fail on current `main` due to this incomplete function signature. `time_*` functions need (apparently) to accept all the parameters specified in the class (a total of six, in this case). They can also admit an arbitrary number of arguments with `*args`, which is the proposed fix.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


grlee77 (approved):
Looks good. current CI won't test this, so will go ahead and merge now

PR #5433, opened 2021-06-18
Rename `label` variable to circumvent link issue in docs.

### LINKED ISSUE #5414: Incorrect link for `label` in regionprops gallery example
## Description

In the regionprops gallery example:

https://scikit-image.org/docs/0.18.x/auto_examples/segmentation/plot_regionprops.html

The "label" function links to Python's int documentation. I think that's because later in the example label (the function measure.label) gets redefined as `label = props[i].label`, which is indeed an int.

Probably this needs a fix in sphinx-gallery but in the meantime perhaps renaming some variables would help.

mkcor:
## Description

Fixes #5414 -- thanks for catching this, @jni 

Another way of avoiding this issue is to import submodules rather than functions; the link is correct when calling `measure.label`. Personally, I prefer writing
```py
from skimage import draw, measure, transform
```
in gallery examples, so functions are prefixed with their submodule and the reader becomes familiar with the code structure...

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/examples/segmentation/plot_regionprops.py
+++ b/doc/examples/segmentation/plot_regionprops.py
@@ -104,14 +104,14 @@ properties = ['area', 'eccentricity', 'perimeter', 'mean_intensity']
 # For each label, add a filled scatter trace for its contour,
 # and display the properties of the label in the hover of this trace.
 for index in range(1, labels.max()):
-    label = props[index].label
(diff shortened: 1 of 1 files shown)


grlee77 (approved):

grlee77:
This looks good. The failures seen here are unrelated and resolved by #5433

PR #5438, opened 2021-06-21
Document the multichannel->channel_axis transition in the release notes

grlee77:
## Description

I put the most detail under the API section, but there were also deprecation and new feature components.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/doc/release/release_dev.rst
+++ b/doc/release/release_dev.rst
@@ -15,6 +15,10 @@ https://scikit-image.org
 New Features
 ------------
 
+- Added support for processing images with channels located along any array
+  axis. This is in contrast to previous releases where channels were required
+  to be the last axis of an image. See more info on the new ``channel_axis``
+  argument under the API section of the release notes.
 - Added a new keyword only parameter ``random_state`` to
   ``morphology.medial_axis`` and ``restoration.unsupervised_wiener``.
 - Seeding random number generators will not give the same results as the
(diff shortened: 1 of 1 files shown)


mkcor (approved):
Wonderful! I appreciate your writing style, @grlee77: It is very clear.

mkcor:
By the way, I wonder if we should turn
https://github.com/scikit-image/scikit-image/blob/dad624b0f1063d653a1d7e7beba080312cfd59b6/skimage/color/colorconv.py#L746
into
```py
def rgb2gray(rgb, channel_axis=-1)
```
'simply' adding a call to `np.moveaxis` at the beginning of the function... :thinking:

grlee77:
> def rgb2gray(rgb, channel_axis=-1)

Good point. The prior PRs deprecated `multichannel`, replacing it with `channel_axis`. However, we probably want to also add a new `channel_axis` PR to any functions that always assumed a color image for input and thus did not have a `multichannel` argument previously. This would include many functions in `skimage.color`, but I'm not sure if there are also others outside of that package?

rfezzani:
> By the way, I wonder if we should turn
> 
> https://github.com/scikit-image/scikit-image/blob/dad624b0f1063d653a1d7e7beba080312cfd59b6/skimage/color/colorconv.py#L746
> 
> into
> 
> ```python
> def rgb2gray(rgb, channel_axis=-1)
> ```
> 
> 'simply' adding a call to `np.moveaxis` at the beginning of the function... thinking

It's funny, I though about that when reviewing your PR @mkcor #5420 ([here](https://github.com/scikit-image/scikit-image/pull/5420/files#diff-86a281d5e54219571aa714be839b3d6d1044d45918761c000cbd8ecd153d2590R56)) :smile:

mkcor:
> It's funny, I though about that when reviewing your PR @mkcor #5420 (here) :smile:

@rfezzani I know!! :wink:

mkcor:
> [...] but I'm not sure if there are also others outside of that package?

I don't know right off the bat. I'll open an issue so we can keep track.

By the way, this PR is a documentation one, so I guess I can go ahead and merge without a second approval.

Issue #5428, opened 2021-06-16
Utility function for cropping images from a bounding box


sofroniewn:
## Feature Request

This is a feature request for a utility function for cropping images from a bounding box.

Note this need comes up a lot, particularly for 2D images - see this discussion on [image.sc](https://forum.image.sc/t/mask-and-crop-with-napari/54008/3?u=sofroniewn) for an example, including prototype code. It would be nice to have a full nD solution though. Possibly something that did broadcasting appropriately too (i.e. cropping a 3D image with a 2D rectangle broadcasts across the "missing" dimension). @jni suggested I open an issue here to track.

I'm thinking something sort of like the following but for the full nD case.

```python
def create_box(data):
    """Create integer box

    Parameters
    ----------
    data : (N, 2) array
        Points around which the box is created.

    Returns
    -------
    box : (2, 2) array
        Integrer extrema of the box.
    """
    min_val = data.min(axis=0)
    max_val = data.max(axis=0)
    tl = np.array([min_val[0], min_val[1]])
    br = np.array([max_val[0], max_val[1]])
    box = np.round(np.array([tl, br])).astype(int)
    return box


def bounding_box_crop(image, rectangle):
    """Create integer box

    Parameters
    ----------
    image : (N, M) array
        2D image.
    rectangle : (4, 2) array or (2, 2) array
        Rectangle for cropping.

    Returns
    -------
    cropped_image : (U, V) array
        Cropped 2D image.
    """
    min_val, max_val = create_box(rectangle)
    return image[min_val[0]: max_val[0], min_val[1]: max_val[1]]
```

Note this is different from the existing https://scikit-image.org/docs/dev/api/skimage.util.html#skimage.util.crop which has a slightly different API.

A quick google reveals this is an often requested utility function so many people might benefit from it

Alfa-Shashank:
Hello guys, I would like to work on this, is this issue still open?

tarun-bisht:
Hi, I am intrested in solving this issue. Is this still open

grlee77:
@Alfa-Shashank , @tarun-bisht.  Thank you for the interest, but there is a PR for this in progress at #5499. 

In general, if you go to the "Pull requests" tab there is a search box where you can try to determine if a particular feature already has a pull request open. Since #5499 references this issue, you can also see it listed here (just before @Alfa-Shashank's comment).

tarun-bisht:
@grlee77 thanks for clarifying

[...]

PR #5441, opened 2021-06-24
Bounding Box crop module issue 5428

ArpithaRaviSeetala:
Bounding Box crop module issue #5428
A newfile module - crop_bounding_box.py added. This file contains two
methods. One, to locate the users downloads folder and create the
Cropped folder if necessary. Two, a method to crop parts of the image
in the bounding box. All the cropped images are then saved in the folder
created in the download_path() method.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.

### DIFF
--- a/skimage/util/__init__.py
+++ b/skimage/util/__init__.py
@@ -16,7 +16,6 @@ from ._invert import invert
 from ._montage import montage
 from ._map_array import map_array
 from ._label import label_points
-from .crop_bounding_box import bounding_box_crop, download_path
 
 
 @functools.wraps(np.pad)
@@ -49,6 +48,4 @@ __all__ = ['img_as_float32',
            'invert',
            'unique_rows',
            'label_points',
-           'bounding_box_crop',
-           'download_path',
            ]
--- a/skimage/util/__init__.py
+++ b/skimage/util/__init__.py
@@ -16,6 +16,7 @@ from ._invert import invert
 from ._montage import montage
 from ._map_array import map_array
 from ._label import label_points
+from .crop_bounding_box import bounding_box_crop, download_path
 
 
 @functools.wraps(np.pad)
@@ -48,4 +49,6 @@ __all__ = ['img_as_float32',
            'invert',
            'unique_rows',
            'label_points',
+           'bounding_box_crop',
+           'download_path',
            ]
--- a/skimage/util/crop_bounding_box.py
+++ b/skimage/util/crop_bounding_box.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+"""Bounding boxes cropping module for any image."""
+
+import cv2
+from pathlib import Path
+import os
+
+
+def download_path():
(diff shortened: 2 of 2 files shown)


ArpithaRaviSeetala:
Hello everyone, I am new to this platform. Can anybody help me why the "continuous-integration" checks don't pass even after I reverted the changes back to the state when all the checks have passed successfully. Can anybody help me with what happens in a "continuous integration" check.

grlee77:
> Can anybody help me with what happens in a "continuous integration" check.

In this case, I wouldn't worry about it. The failure I see here is a timeout on Azure that has been annoying me for some time. Something occasionally stalls during the example gallery build, but I have so far been unable to diagnose a root cause. If I restart it, it may pass.

ArpithaRaviSeetala:
> 
> 
> > Can anybody help me with what happens in a "continuous integration" check.
> 
> In this case, I wouldn't worry about it. The failure I see here is a timeout on Azure that has been annoying me for some time. Something occasionally stalls during the example gallery build, but I have so far been unable to diagnose a root cause. If I restart it, it may pass.

Thanks a lot for the clarification. Please do let me know in-case I can do something to re-run the checks from my end.

grlee77:
Thank you for making the PR @ArpithaRaviSeetala, but I see that this implementation is OpenCV-based. That may be good from a performance perspective, but will limit it to the 2D case only (n-dimensional support was part of the request in #5428 and something we make an effort to support where possible)

The bigger issue is that OpenCV is not one of our required dependencies for scikit-image, so cannot be used as the basis for this function. We could debate about having an optional, fast code path using OpenCV for users who have it installed, but there would still need a fallback implementation for users without OpenCV.

ArpithaRaviSeetala:
> 
> 
> Thank you for making the PR @ArpithaRaviSeetala, but I see that this implementation is OpenCV-based. That may be good from a performance perspective, but will limit it to the 2D case only (n-dimensional support was part of the request in #5428 and something we make an effort to support where possible)
> 
> The bigger issue is that OpenCV is not one of our required dependencies for scikit-image, so cannot be used as the basis for this function. We could debate about having an optional, fast code path using OpenCV for users who have it installed, but there would still need a fallback implementation for users without OpenCV.

Thanks a lot for the feedback @grlee77 . I will look into the points mentioned and try to come up with another approach which also includes solution for the n-dimensional case.

grlee77:
Thank you for opening the PR @ArpithaRaviSeetala, but I am going to close this since there is now another PR for this feature that has an nD implementation (#5499). Feel free to review/comment on that one or work on another feature.

ArpithaRaviSeetala:
> 
> 
> Thank you for opening the PR @ArpithaRaviSeetala, but I am going to close this since there is now another PR for this feature that has an nD implementation (#5499). Feel free to review/comment on that one or work on another feature.

Okay @grlee77 . Sure. Thank you very much.

Issue #5439, opened 2021-06-22
Meta-issue: pending API changes for 2.0


grlee77:
*The "1.0" concept since changed to "2.0", so replace all 1.0 below with 2.0*

## Description

This issue is just to summarize some of the API considerations we need to make a decision on for the 1.0 release. Please edit and add any additional items I may have forgotten (most of these I copied over from the scikit-image API 1.0 "project" under the Projects tab).

### Module deprecations

- remove skimage.viewer:  issue #?
could then close >10 issues: https://github.com/scikit-image/scikit-image/issues?q=is%3Aissue+is%3Aopen+viewer

- remove most code in skimage.io, leaving only a thin imageio wrapper: #5036

### Function/module moves

- Update RAG functions and move them out of skimage.future: #1696
additional issues related to RAG: https://github.com/scikit-image/scikit-image/issues?q=is%3Aissue+is%3Aopen+RAG

### Parameter removal
- Juan proposed removing `preserve_range` altogether in 1.0 (always using the preserve_range=True behavior)
https://github.com/scikit-image/scikit-image/pull/5281#issuecomment-864646679

### parameter/function renaming

- Need to decide on random_seed vs. seed vs. sample_seed: #5359

- Need to decide on n_jobs vs. workers vs. num_workers vs. ncpu, etc: #4876

- Need to decide on max_iter vs. niter vs. iterations.  Also selem vs. structure vs. footprint: #4154

- label_image vs labels as a parameter name: issue #?

- Harmonizing structure element API to use radius or similar name: #4536

- Rename img_as_* to rescale_to_* or convert_to_*:  #1234

- Many regionprops are renamed in #5254 (corresponding to issue #4821)
The old names are still recognized so this should not be a breaking change

### Misc
- audit usage of img_as_float: #3373
Requires some work to further verify which functions have scale-dependent default parameters, etc.
 
- Fixes for data types and other common sources of errors: #3009
involves more prominent and extended version of https://scikit-image.org/docs/stable/user_guide/data_types.html

jni:
@grlee77 a fantastic list! I would make one modification and two additions:

- for 1.0 I think we should only have one name per region property. We can probably rethink the whole API anyway so that e.g. you get a dict rather than a list where the index has no relation to the label of that region. Regionprops is probably the most useful part of the library so let's think big!
- All functions that take in x/y (transposed) coordinates should take axis-aligned (aka rc) coordinates instead. Example: warps and geometric transformations.
[...]

PR #5445, opened 2021-06-26
rename selem to footprint everywhere

grlee77:
## Description

In the discussion in #4154 it was decided that `footprint` should be less ambiguous than `selem` for use in all rank filters and morphology functions.

This PR makes that change

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


skimage/morphology/selem.py
```diff
@@ -1,366 +1,13 @@
...
+    square, rectangle, diamond, disk, cube, octahedron, ball, octagon, star
+)
+
+warnings.warn(
+    "The `skimage.morphology.selem` module is deprecated and will be removed "
+    "in scikit-image 1.0. The functions formerly in "
+    "`skimage.morphology.selem` should be imported directly from "
+    "`skimage.morphology` instead.",
+    FutureWarning, stacklevel=2
+)
```
jni, line 13: Lovely.

I wonder though whether we should not instead encourage directly using `from skimage.morphology import footprints; footprint = footprints.square(5)`. To me this addresses the point in #5439 about grouping functions according to "purpose and equivalence classes". `morphology.square()` is also less descriptive than `footprints.square`.

Perhaps something to add to a dedicated issue for discussion (tracked from #5439) rather than hold up this PR.

jni: (But, a quick fix here would be to remain agnostic in the prescription: instead of "should be imported from skimage.morphology instead", we could write "skimage.morphology.selem has been moved to skimage.morphology.footprints.") (And yes I have a mild preference for "footprints" over "footprint", because the latter is easily confused with a variable/kwarg name.)

skimage/morphology/tests/test_gray.py
```diff
@@ -266,17 +267,26 @@ def test_discontiguous_out_array():
...
 
 
 def test_deprecated_import():
     msg = "Importing from skimage.morphology.grey is deprecated."
-    with testing.expected_warnings([msg]):
+    with expected_warnings([msg]):
         from skimage.morphology.grey import erosion
+
+
+@pytest.mark.parametrize(
```
jni, line 286: Gotta love this function 😂

jni (approved):
Awesome. 🙏 I've left one minor comment about the future of the API, but happy to merge this as-is and leave that discussion for later.

jni:
Note: if there are no further updates I'll merge this by the end of the week.

skimage/measure/_label.py
```diff
@@ -21,8 +21,8 @@ def _label_bool(image, background=None, return_num=False, connectivity=None):
             f'be in [1, ..., {image.ndim}]. Got {connectivity}.'
         )
 
-    selem = _resolve_neighborhood(None, connectivity, image.ndim)
-    result = ndimage.label(image, structure=selem)
+    footprint = _resolve_neighborhood(None, connectivity, image.ndim)
+    result = ndimage.label(image, structure=footprint)
```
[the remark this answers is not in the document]
grlee77: No, but all instances of structure as a `kwarg` seem to be when calling `scipy.ndimage` functions. I doubt there will be much enthusiasm for renaming those upstream (at least there was resistance to renaming the first argument from `input` to `image` in the past).

skimage/morphology/selem.py
grlee77: I was not sure if this `selem`/`footprint` module within `filters`, `morphology`, etc. was really intended to be public or not. Specifically, these functions are [grouped in with all other morphology functions in the docs](https://scikit-image.org/docs/stable/api/skimage.morphology.html).

I am fine with renaming to `footprints.py` and importing from there, but wonder if we can perhaps follow up later to make these logical groupings (`binary` vs. `gray` vs. `footprints`) clearer in the docs.

grlee77:
Thanks for reviewing @jni. I made the suggested rename and warning changes. As far as I could tell, the uses of `structure` arise of `scipy.ndimage` functions, not ours.

Two other things were also updated since you reviewed:
1.) renamed test_selem.py -> test_footprints.py
2.) There were still tons of mentions of "structuring element" in the docstrings! These now refer to footprint

skimage/measure/_label.py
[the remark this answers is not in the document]
jni: Oh my apologies, I thought this was `measure.label`. 🤦

skimage/morphology/selem.py
jni: Yes, following things up later is totally fine, but that's why I'm saying, let's not be prescriptive in the deprecation message, because we might change our mind during those discussions.

jni:
Great. Totally happy with the changes now. I'll merge this in <24h if no one has stepped in to review+merge before then.

rfezzani (approved):
Thank you @grlee77, merging :wink:

PR #5446, opened 2021-06-26
disable calls to plotly.io.show when running on Azure

grlee77:
## Description

It has been a common occurance lately for the CI job on Azure that involves gallery testing to timeout.

I have a suspicion that this is somehow related to Plotly, but haven't confirmed that for sure yet.

In this PR, I try defining environment variable: `SKIMAGE_DISABLE_PLOTLY_SHOW` to disable it

This environment variable is only exported during Gallery testing on Azure, so users should still get the plots popping up in their browser if running the scripts from the command line.

## Checklist

  existing benchmark
- Unit tests

## For reviewers

  later.


grlee77:
This seems to have worked. I manually started the Azure job a second time [here](https://dev.azure.com/scikit-image/scikit-image/_build/results?buildId=6188&view=results) to check that we didn't just get lucky.

@mkcor, the timeouts that have been common on Azure in recent weeks seem to have been related to the plotly.io.show calls I had asked you to add to the examples. This PR bypasses those when running the Azure gallery testing job. I tried this with a gallery example, but let me know if you are aware of anything else in Plotly itself.

grlee77:
A second option would be replacing calls to show() with `fig.write_html("<filename>")`, but this would leave a files on disk that the user would have to manually go and open with a browser if they wanted to see it, so I am not a big fan of that.

jni:
@grlee77 The problem I have with this is that it makes the gallery examples uglier, for something that is a *mild* problem from the developer side. I'm not sure on balance it is a good change. I think I would prefer writing a tool in our `/tools` folder that would crawl all our examples and comment-out lines with `plotly.io.show`. What do you think? (And maybe it doesn't need a dedicated script, just a grep/find/sed one-liner, but my shell-fu is not good enough for that!)

grlee77:
I agree that it makes those examples more ugly. I reverted the environment variable and instead use a one line sed-based solution in 0e2dc2a.

grlee77:
CI is green for the `sed`-based version as well.

The only change to an example script now is adding the show commands to `doc/examples/applications/plot_3d_structure_tensor.py ` where they were missing.

grlee77:
ping @jni

azure-pipelines.yml
```diff
@@ -189,16 +189,17 @@ jobs:
...
-          echo "skipped"
-        else
-          $PYTHON -W ignore:Matplotlib:UserWarning "${f}"
-          if [ $? -ne 0 ]; then
-            exit 1
-          fi
+        # Comment out any plotly.io.show() calls before running the example.
+        # Plotly opens a web browser, which often seemed to cause the CI to
+        # hang and timeout.
+        sed -i 's/plotly.io.show/# plotly.io.show/g' ${f}
```
mkcor, line 195: I like this non-invasive workaround :slightly_smiling_face:

azure-pipelines.yml
```diff
@@ -189,16 +189,17 @@ jobs:
...
+
+        $PYTHON -W ignore:Matplotlib:UserWarning "${f}"
+        if [ $? -ne 0 ]; then
+          exit 1
         fi
       done
-    condition:
+    condition: eq(variables['TEST_EXAMPLES'], 'true')
```
mkcor, line 202: I guess this will disappear upon merging `main`

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -143,7 +144,7 @@
 # Let us view the spatial distribution of the eigenvalues in the X-Y plane
 # where the maximum eigenvalue is found (i.e., ``Z = coords[1]``).
 
-px.imshow(
+fig1 = px.imshow(
```
mkcor, line 147: There are two other `px.imshow` calls prior to this one!

doc/examples/applications/plot_3d_structure_tensor.py
```diff
@@ -188,14 +189,18 @@
...
-px.imshow(
+fig3 = px.imshow(
     subplots,
     zmin=v_min,
     zmax=v_max,
     facet_col=2,
     labels={'facet_col': 'longitudinal'}
 )
 
+plotly.io.show(fig1)
```
mkcor, line 200: This call needs to happen right after `fig1 = ...`. Otherwise only `fig3` is showing: https://2292-2014929-gh.circle-artifacts.com/0/doc/build/html/auto_examples/applications/plot_3d_structure_tensor.html#sphx-glr-auto-examples-applications-plot-3d-structure-tensor-py

mkcor:
> @mkcor, the timeouts that have been common on Azure in recent weeks seem to have been related to the plotly.io.show calls I had asked you to add to the examples.

I think these timeouts would happen (sometimes) even without the plotly.io.show calls: Plotly figures by themselves can make the generated HTML files [pretty heavy](https://github.com/scikit-image/scikit-image/pull/4953#issuecomment-686486012). When building the example gallery locally, I've noticed the slowdown caused by Plotly figures, regardless of the renderer.

> This PR bypasses those when running the Azure gallery testing job.

Okay.

> I tried this with a gallery example, but let me know if you are aware of anything else in Plotly itself.

Not really; I haven't followed closely beyond https://github.com/plotly/plotly.py/pull/1577 by @emmanuelle. I didn't know of `plotly.io.to_html` and `plotly.io.write_html` before you pointed out the second option (shame on me).

I know that, here,
https://github.com/scikit-image/scikit-image/blob/6ff11f6ff31e53c1dffa0cd65b03bf7b5ee8d668/doc/source/conf.py#L155
we want 'sphinx_gallery_png' rather than 'sphinx_gallery' so that gallery thumbnails display well. The other [renderer candidates](https://github.com/plotly/plotly.py/blob/17b7c27f9e17b413b6958835ad53dc1aefb90834/packages/python/plotly/plotly/io/_renderers.py#L407) don't seem fitting.

doc/examples/applications/plot_3d_structure_tensor.py
grlee77: Oh, thanks. I had only run from the command line and forgot to check the built docs.

azure-pipelines.yml
grlee77: yeah, it does not show up in the diff anymore. this got modified unintentionally during the recent CI-based benchmarking PR, but we later restored it.

grlee77:
@mkcor: I have addressed the comments. CI is still running, but the relevant Azure jobs have already passed.

mkcor (approved):

mkcor:
> @mkcor: I have addressed the comments. CI is still running, but the relevant Azure jobs have already passed.

Yay! Thanks. Since this is a CI+docs PR, can I go ahead and merge?

grlee77:
We addressed Juan's original objection to the environment variable, so I say go for it!

